PR #454, opened 2021-12-01
Add scalar type promotion for mul and div

dan-garvey:
Mirrors the sub logic to provide type promotion for elementwise ops.

### DIFF
--- a/e2e_testing/torchscript/elementwise.py
+++ b/e2e_testing/torchscript/elementwise.py
@@ -363,6 +363,8 @@ def RsubModule_noalpha_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 4))
     
 # ==============================================================================
+
+
 class ElementwiseMulScalarModule(torch.nn.Module):
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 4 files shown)


e2e_testing/torchscript/type_promotion.py
```diff
@@ -34,6 +34,51 @@ def TypePromotionSameCategoryDifferentWidthModule_basic(module, tu: TestUtils):
         torch.randint(10, [4]))
 
 
+class TypePromotionSameCategoryDifferentWidthMulModule(torch.nn.Module):
```
cathyzhyi, line 37: The tests in this file are for testing type promotion infrastructure.  Mul and Div are ops using the infrastructure and need to be moved to basic.py.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1512,7 +1512,13 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
       return nullptr;
     }
-    return b.create<arith::MulFOp>(loc, payloadArgs[0], payloadArgs[1]);
+    AtenMulTensorOp::Adaptor adaptor(operands);
+    Type dtype = converter->convertType(mul.getType())
+                     .cast<RankedTensorType>()
+                     .getElementType();
+    Value lhs = convertScalarToDtype(b, loc, payloadArgs[0], dtype);
+    Value rhs = convertScalarToDtype(b, loc, payloadArgs[1], dtype);
+    return b.create<arith::MulFOp>(loc, lhs, rhs);
```
cathyzhyi, line 1521: Can you refer to https://github.com/llvm/mlir-npcomp/blob/5475d02a0d03b2bda7ef9cc240773041d22a926b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp#L1485 to add support for integer type as well.

cathyzhyi (changes requested):

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
dan-garvey: should I leave div alone for now? It seems like for integer operands both inputs should be promoted to float32:

https://pytorch.org/docs/stable/generated/torch.div.html

cathyzhyi: ha, I see. yea, we can deal with div later.

e2e_testing/torchscript/elementwise.py
```diff
@@ -378,7 +380,31 @@ def forward(self, x):
 @register_test_case(module_factory=lambda: ElementwiseMulScalarModule())
 def ElementwiseMulScalarModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 4))
-    
+
+
+
+class ElementwiseMulScalarPromotionModule(torch.nn.Module):
```
cathyzhyi, line 386: `MulTensor => `MulScalar`test. 
`Promotion` => `TypePromotion` for consistency with the test below. IMO, we could remove `TypePromotion` because the core functionality of type promotion are tested in the type_promotion.py. The focus of these tests are checking the op itself is lowered correctly.

e2e_testing/torchscript/elementwise.py
```diff
@@ -553,7 +579,32 @@ def forward(self, x):
 def ElementwiseDivScalarModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 4))
 
+
+class ElementwiseDivScalarTypePromotionModule(torch.nn.Module):
```
cathyzhyi, line 583: same here

cathyzhyi (changes requested):

e2e_testing/torchscript/elementwise.py
dan-garvey: Given that a test for these ops already exist, and a test for type promotion already exists, are these tests just redundant?

cathyzhyi: hmm, I didn't find any mul.Tensor test but only a mul.Scalar test. Note, `torch.mul` could endup with different variant depending on the operand types. I think adding one test for `mul.Tensor` is good enough. Adding two tests for different types would be even better because Int and float are using different code paths.

cathyzhyi (approved):

PR #460, opened 2021-12-03
Bug fixes that pops up when updating generatedAten ops td

pashu123:
There is an op name change that requires trivial changes.
Also, some of the warning has been fixed.


cathyzhyi:
@Eliasj42 FYI, we need to use `https://github.com/llvm/torch-mlir/blob/main/build_tools/update_torch_ods.sh` when updating the ods.

cathyzhyi (approved):
Thanks for the fix!

PR #461, opened 2021-12-04
Update external llvm.

pashu123:
The external llvm is updated to point to
https://reviews.llvm.org/rG966b72098363d44adf2882b9c34fcdbe344ff913.
Some of the changes wrt. NamedAttr has been addressed.


pashu123:
@sjarus Wondering why the tosa test case starts passing after the llvm-external update? Do you have any idea?

sjarus:
> @sjarus Wondering why the tosa test case starts passing after the llvm-external update? Do you have any idea?

The previous failure was in TOSA to LinAlg. It performs a reshape that could not accommodate dynamic shapes right. I haven't looked at the latest flow, but improvements in the backend below TOSA should not be surprising here. It's good news, right ? :)

sjarus:
As further background, there's an ongoing task to fix tosa.reshape dialect form to accommodate dynamic shape expression well enough to unlock full dynamism in the reduction operator lowering. The elementwise binary op case is related to it .

PR #463, opened 2021-12-07
Add aten::nll_loss_backward op

pashu123:
The lowering of aten::nll_loss_backward op has been added
from torch to linalg dialect. The changes has been made as
a part of -torch-convert-to-linalg pass.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/Utils/TorchUpstream.h
+++ b/include/torch-mlir/Dialect/Torch/Utils/TorchUpstream.h
@@ -70,6 +70,15 @@ struct ResultTypeState {
 ScalarType result_type(const ResultTypeState &in_state);
 ScalarType promote_skip_undefined(ScalarType a, ScalarType b);
 
+//===----------------------------------------------------------------------===//
+// These constants control the reduction behavior of the loss functions.
+// None, Mean and Sum corresponds to "do not reduce", "Mean of losses", and "sum
+// of losses" respectively.
+// Source:
+// https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/core/Reduction.h
+//===----------------------------------------------------------------------===//
+enum Reduction { None, Mean, Sum, END };
+
 } // namespace torch_upstream
 } // namespace torch
 } // namespace mlir
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -19,6 +19,7 @@
 #include "mlir/IR/Matchers.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionDialect.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h"
@@ -28,6 +29,7 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 using namespace mlir::torch::TorchConversion;
+using namespace mlir::torch::torch_upstream; // For ScalarType and type
 
 // -----------------------------------------------------------------------------
 // Patterns (as this grows, it should be organized into multiple files)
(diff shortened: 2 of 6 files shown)


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -966,6 +971,27 @@ ChangeResult TypeAnalyzer::visitAtenNllLossForwardOp(
...
+      ValueKnowledge::getNotNonePessimisticValueState(op.getContext());
+
+  knowledge.dtype = self.dtype;
+  int64_t reduction;
+  unsigned resultRank = self.sizes.size();
+
+  if (self.hasSizes &&
+      matchPattern(op.reduction(), m_TorchConstantInt(&reduction))) {
+    // reduction == 1 means reduce 1st dim.
+    resultRank = reduction == 1 ? resultRank - 1 : resultRank;
```
ramiro050, line 988: I don't think this is true. From the definition

https://github.com/pytorch/pytorch/blob/617a3bd944320d2d40bae78aefbfcdebf5255a59/aten/src/ATen/native/LossNLL.cpp#L110

`NllLossBackward` returns a tensor of the same size as the `self` argument.

Also, the visitor for `NllLossForward`, where this calculation is also used, does not seem right to me. The case of `reduction == 2` is not handled, and it is not clear to me that subtracting one from the result rank results in a scalar, as specified by the documentation

https://pytorch.org/docs/stable/generated/torch.nn.NLLLoss.html

ramiro050 (changes requested):
Thanks for this! A few comments

lib/Dialect/Torch/Transforms/RefineTypes.cpp
pashu123: Thanks for the comments, I will take a look at this. I guess I have made some mistakes.

pashu123: So, the shape of output is the same as the shape of the self in the case of NLL loss backward and I have modified the case of NLL loss forward taking into account both reductions.

pashu123: So, the shape of output is the same as the shape of the self in the case of NLL loss backward and I have modified the case of NLL loss forward taking into account both reductions.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1263,6 +1263,107 @@ class ConvertAtenNllLossForwardOp
...
+                  // grad_input[i][j] = (j == target[i]) ? grad_output[i] : 0
+                  Value cmpEq = rewriter.create<arith::CmpIOp>(
+                      loc, arith::CmpIPredicate::eq, indJ, indTarget);
+
+                  // The target index shouldn't be equal to `ignoreIndex`.
+                  Value cmpNEq = rewriter.create<arith::CmpIOp>(
+                      loc, arith::CmpIPredicate::ne, ignoreIndexVal, indTarget);
+                  Value finalPredicate =
+                      rewriter.create<arith::AndIOp>(loc, cmpEq, cmpNEq);
+                  Value negate =
```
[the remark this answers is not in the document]
pashu123: Updated.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
pashu123: scalar -> 0d tensor?

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1323,6 +1328,108 @@ class ConvertAtenNllLossForwardOp
...
+  LogicalResult
+  matchAndRewrite(AtenNllLossBackwardOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+    Location loc = op->getLoc();
+    Value input = adaptor.self();
+    Value target = adaptor.target();
+    Value weight = adaptor.weight();
+    Value grad_output = adaptor.grad_output();
```
cathyzhyi, line 1350: nit: gradOuput

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1323,6 +1328,108 @@ class ConvertAtenNllLossForwardOp
...
+                      loc, rewriter.getIndexType(), args[0]);
+                  Value indJ = rewriter.create<linalg::IndexOp>(loc, 1);
+
+                  // The final result is given by:
+                  // grad_input[i][j] = (j == target[i]) ? -grad_output[i] : 0
+                  Value cmpEq = rewriter.create<arith::CmpIOp>(
+                      loc, arith::CmpIPredicate::eq, indJ, indTarget);
+
+                  // The target index shouldn't be equal to `ignoreIndex`.
+                  Value cmpNEq = rewriter.create<arith::CmpIOp>(
```
cathyzhyi, line 1415: nit: cmpNe

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -328,6 +328,11 @@ static Value buildUnitNormalCdf(OpBuilder &b, Location &loc, Value x) {
   return buildNormalCdf(b, loc, x, zero, one);
 }
 
+// These constants control the reduction behavior of the loss functions.
+// None, Mean and Sum corresponds to "do not reduce", "Mean of losses", and "sum
+// of losses" respectively.
+enum Reduction { None, Mean, Sum, END };
```
cathyzhyi, line 334: we can move this definition to `include/torch-mlir/Dialect/Torch/Utils/Utils.h` since it's shared by multiple files. Also, please include the `aten/src/ATen/core/Reduction.h` in the comments for future reference.

cathyzhyi (approved):

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
ramiro050: Would the file https://github.com/llvm/torch-mlir/blob/main/include/torch-mlir/Dialect/Torch/Utils/TorchUpstream.h be a better location for this `enum`, since it is code from torch upstream?

cathyzhyi: yea, you are right. that file should be a better place.

include/torch-mlir/Dialect/Torch/Utils/TorchUpstream.h
```diff
@@ -70,6 +70,13 @@ struct ResultTypeState {
 ScalarType result_type(const ResultTypeState &in_state);
 ScalarType promote_skip_undefined(ScalarType a, ScalarType b);
 
+// These constants control the reduction behavior of the loss functions.
```
cathyzhyi, line 73: nit: can you change the comment to similar format as the other comments?

pashu123: Done, thanks.

PR #464, opened 2021-12-07
[MLIR][TORCH] Add E2E support for aten.zeros op

vivekkhandelwal1:
This commit adds lowering of `aten.zeros` op.

### DIFF
--- a/e2e_testing/torchscript/basic.py
+++ b/e2e_testing/torchscript/basic.py
@@ -900,7 +900,6 @@ class NumelModule(torch.nn.Module):
         None,
         ([-1, -1, -1], torch.float32, True),
     ])
-
     def forward(self, input):
         return torch.numel(input)
 
@@ -918,10 +917,89 @@ class NumelZeroRankModule(torch.nn.Module):
         None,
         ([], torch.int64, True),
     ])
-
     def forward(self, input):
         return torch.numel(input)
 
 @register_test_case(module_factory=lambda: NumelZeroRankModule())
 def NumelZeroRankModule_basic(module, tu: TestUtils):
     module.forward(torch.randint(10,[]))
+
+
+class ZerosModuleInt2D(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.zeros(3, 4, dtype=torch.int64)
+
+@register_test_case(module_factory=lambda: ZerosModuleInt2D())
+def ZerosModuleInt2D_basic(module, tu: TestUtils):
+    module.forward()
+
+
+class ZerosModuleInt3D(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.zeros(3, 4, 5, dtype=torch.int64)
+
+@register_test_case(module_factory=lambda: ZerosModuleInt3D())
+def ZerosModuleInt3D_basic(module, tu: TestUtils):
+    module.forward()
+
+
+class ZerosModuleFloat2D(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.zeros(3, 4, dtype=torch.float32)
+
+@register_test_case(module_factory=lambda: ZerosModuleFloat2D())
+def ZerosModuleFloat2D_basic(module, tu: TestUtils):
+    module.forward()
+
+
+class ZerosModuleFloat3D(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.zeros(3, 4, 5, dtype=torch.float32)
+
+@register_test_case(module_factory=lambda: ZerosModuleFloat3D())
+def ZerosModuleFloat3D_basic(module, tu: TestUtils):
+    module.forward()
+
+
+class ZerosModuleFalsePinMemory(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.zeros(3, 4, dtype=torch.float32, pin_memory=False)
+
+@register_test_case(module_factory=lambda: ZerosModuleFalsePinMemory())
+def ZerosModuleFalsePinMemory_basic(module, tu: TestUtils):
+    module.forward()
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -3269,24 +3269,44 @@ public:
 } // namespace
 
 namespace {
-class ConvertAtenOnesOp : public OpConversionPattern<AtenOnesOp> {
-public:
-  using OpConversionPattern::OpConversionPattern;
+// Converts AtenOnesOp and AtenZerosOp.
+struct ConvertAtenOnesZerosOp : ConversionPattern {
+  ConvertAtenOnesZerosOp(TypeConverter &typeConverter, MLIRContext *context)
+      : ConversionPattern(typeConverter, MatchAnyOpTypeTag(), /*benefit=*/1,
+                          context) {}
+
   LogicalResult
-  matchAndRewrite(AtenOnesOp op, OpAdaptor adaptor,
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
                   ConversionPatternRewriter &rewriter) const override {
+    if (!isa<AtenOnesOp, AtenZerosOp>(op))
+      return rewriter.notifyMatchFailure(op,
+                                         "not a supported ones or zeros op");
+
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
       return failure();
-    Location loc = op.getLoc();
+    Location loc = op->getLoc();
+
+    SmallVector<Value, 3> opArguments;
+    int64_t elementValue;
+
+    if (AtenOnesOp onesOp = dyn_cast<AtenOnesOp>(op)) {
+      opArguments.insert(opArguments.end(),
+                         {onesOp.size(), onesOp.layout(), onesOp.pin_memory()});
+      elementValue = 1;
+    } else if (AtenZerosOp zerosOp = dyn_cast<AtenZerosOp>(op)) {
+      opArguments.insert(opArguments.end(), {zerosOp.size(), zerosOp.layout(),
+                                             zerosOp.pin_memory()});
+      elementValue = 0;
+    }
 
     // We ignore device, but add simple asserts for unimplemented kwargs
-    if (!op.layout().getType().isa<Torch::NoneType>())
+    if (!opArguments[1].getType().isa<Torch::NoneType>())
       return rewriter.notifyMatchFailure(op,
                                          "only default layout is supported");
 
     bool pinMemory = false;
-    if (!op.pin_memory().getType().isa<Torch::NoneType>() &&
-        !matchPattern(op.pin_memory(), m_TorchConstantBool(&pinMemory))) {
+    if (!opArguments[2].getType().isa<Torch::NoneType>() &&
+        !matchPattern(opArguments[2], m_TorchConstantBool(&pinMemory))) {
       return rewriter.notifyMatchFailure(
           op, "pin_memory must be constant bool or None");
     }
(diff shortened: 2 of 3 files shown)


lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3303,21 +3310,28 @@ class ConvertAtenOnesOp : public OpConversionPattern<AtenOnesOp> {
...
 
     RankedTensorType newResultType =
-        getTypeConverter()->convertType(op.getType()).cast<RankedTensorType>();
+        getTypeConverter()
+            ->convertType(op->getResult(0).getType())
+            .cast<RankedTensorType>();
     Type outElementType = newResultType.getElementType();
 
-    Value one = rewriter.create<arith::ConstantOp>(
+    int64_t elementValue = 0;
```
pashu123, line 3318: Can be replaced by `int64_t elementVal = isa<AtenOnesOp>(op) ? 1 : 0`.

e2e_testing/torchscript/basic.py
```diff
@@ -918,10 +917,89 @@ def __init__(self):
...
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.zeros(3, 4, dtype=torch.float32, pin_memory=False)
+
+@register_test_case(module_factory=lambda: ZerosModuleFalsePinMemory())
+def ZerosModuleFalsePinMemory_basic(module, tu: TestUtils):
+    module.forward()
```
vivekkhandelwal1, line 1005: This test case and the similar one for AtenOnesOp are failing. When I undo my changes, re-build, and test the case for AtenOnesOp, it passes. The original implementation https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp#L3287-L3294 and my implementation https://github.com/llvm/torch-mlir/pull/464/commits/9f9970cccbd407d4867e4facebcc5c27a86ba44b#diff-fa644837a9cf28c5b4b9d88babd95dc19fc962b2036e083e4b79a6c9672982edR3294-R3301 both are identical. I don't know why the code is behaving differently. I tried hard to debug the error but could not resolve it. One thing I found that in my implementation the following condition https://github.com/llvm/torch-mlir/pull/464/commits/9f9970cccbd407d4867e4facebcc5c27a86ba44b#diff-fa644837a9cf28c5b4b9d88babd95dc19fc962b2036e083e4b79a6c9672982edR3295-R3299 evaluates to true, while for the original, it turns out to be false.

pashu123:
@vivekkhandelwal1 Please update the submodule. The error might be due to that.

pashu123:
Also, this can be broken down into `Aten.fill_scalar op`.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3269,32 +3269,39 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
 } // namespace
 
 namespace {
-class ConvertAtenOnesOp : public OpConversionPattern<AtenOnesOp> {
-public:
-  using OpConversionPattern::OpConversionPattern;
+// Converts AtenOnesOp and AtenZerosOp
```
pashu123, line 3272: Always end comments with fullstop.

vivekkhandelwal1:
> @vivekkhandelwal1 Please update the submodule. The error might be due to that.

That I have already done.

pashu123 (changes requested):

pashu123:
> > @vivekkhandelwal1 Please update the submodule. The error might be due to that.
> 
> That I have already done.

I see the error is not due to the submodule.

vivekkhandelwal1:
> 

I think no because when you go through the definition of torch.ones and torch.zeros, there are many other arguments like pinned memory, device, and layout, which are not supported by fill_scalar op.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
vivekkhandelwal1: This will not work because the value returned by `isa<AtenOnesOp>(op)` is not of the bool type.

pashu123:
> > Also, this can be broken down into `Aten.fill_scalar op`.
> 
> I think no because when you go through the definition of torch.ones and torch.zeros, there are many other arguments like pinned memory, device, and layout, which are not supported by fill_scalar op.

Oh! Thanks for the clarification.

e2e_testing/torchscript/basic.py
vivekkhandelwal1: @cathyzhyi @ramiro050

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -369,6 +369,8 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
       return visitConstantTensorAllocOp<AtenZerosOp>(zeros);
     } else if (auto ones = dyn_cast<AtenOnesOp>(op)) {
       return visitConstantTensorAllocOp<AtenOnesOp>(ones);
+    } else if (auto ones = dyn_cast<AtenZerosOp>(op)) {
+      return visitConstantTensorAllocOp<AtenZerosOp>(ones);
```
pashu123, line 373: The call is already at line 369, no need to redefine.

vivekkhandelwal1: I think I added it by mistake. Thanks!

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3269,32 +3269,39 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
...
     // We ignore device, but add simple asserts for unimplemented kwargs
-    if (!op.layout().getType().isa<Torch::NoneType>())
+    if (!operands[2].getType().isa<Torch::NoneType>())
       return rewriter.notifyMatchFailure(op,
                                          "only default layout is supported");
 
     bool pinMemory = false;
-    if (!op.pin_memory().getType().isa<Torch::NoneType>() &&
-        !matchPattern(op.pin_memory(), m_TorchConstantBool(&pinMemory))) {
+    if (!operands[4].getType().isa<Torch::NoneType>() &&
```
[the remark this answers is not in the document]
vivekkhandelwal1: That I also know. I want to ask that the same code is working for the original implementation of AtenOnesOp and passing all the test cases, while my implementation, even though it's the same, is failing for the AtenOnesOp test case as well. Please go through https://github.com/llvm/torch-mlir/pull/464#discussion_r764084184; I have mentioned the issue in detail here.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3269,32 +3269,39 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
...
+                                         "not a supported ones or zeros op");
+
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
       return failure();
-    Location loc = op.getLoc();
+    Location loc = op->getLoc();
 
     // We ignore device, but add simple asserts for unimplemented kwargs
-    if (!op.layout().getType().isa<Torch::NoneType>())
+    if (!operands[2].getType().isa<Torch::NoneType>())
```
ramiro050, line 3290: This might be the cause of the error you are seeing. Although it may seem like the `operands` array are equivalent to the arguments one would get by doing `op.argumentName()`, they are actually different! The `operands` array represent the arguments to the operation but with types converted to the types expected by the next dialect. For example, if your op takes as an argument a `!torch.vtensor<[1,2],f32>`, the corresponding argument in the operands list will have type `!tensor<1x2xf32>`. On the other hand, the operands that you get by doing `op.argumentName()` are still at the torch dialect level, so their types are `torch` types.

In this particular case, because you are comparing to a `torch` type, you should be using the argument by doing `op.argumentName()`. The same applies for every time you pattern match below.

Rules of thumb:

Use `op.argumentName()` when:
- you're extracting values or checking if something is a `torch` type

Use `operands` array when:
- you're extracting type information expecting a tensor type (such as when you calculate rank or result element type and you cast to `RankedTensorType` along the way)
- you want your operand to continue living after the `TorchToLinalg` conversion is over (This is what ops that modify tensors do, they pass an element of the `operands` list to the `linalg.generic` calculation because the tensor should still exist after the conversion is over)

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
vivekkhandelwal1: Thanks, but `op` here is a pointer to the AtenOp, due to which we can't do `op.argumentName()`. Please correct me if I'm wrong.

ramiro050: That's correct. There are two approaches that come to mind for handling this. One is to do what the `ConvertReductionOp` does, which is to dynamic cast the op and then extract the information that is specific to the op, as well as define information that is specific for the op (such as the element value used to fill the tensor). 

The other approach, if `aten.ones` and `aten.zeros` had the exact same interface, would be to use a template. However, because here you have code that requires dynamic knowledge to determine the element value used to fill the tensor, I would go with the first approach.

ramiro050: Actually, a template is also a viable option because you can pass the fill value as a template argument:

```cpp
template<typename OpTy, int FillVal>
```

Feel free to choose whichever you find more appropriate

ramiro050: A third option is to keep the conversion classes separate, but write a helper function for the common calculations that both op-converters can share

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3269,32 +3269,56 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
...
+
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
       return failure();
-    Location loc = op.getLoc();
+    Location loc = op->getLoc();
+
+    SmallVector<Value, 3> opArguments;
+    int64_t elementValue;
+
+    if (isa<AtenOnesOp>(op)) {
```
pashu123, line 3292: You can do the check as `if(AtenOnesOp onesOp = dyn_cast<AtenOnesOp>(op))`, similary for `AtenZeros`.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3269,32 +3269,56 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
...
+      AtenOnesOp onesOp = dyn_cast<AtenOnesOp>(op);
+      opArguments.push_back(onesOp.size());
```
pashu123, line 3294: How about inserting them in one single line rather than 3 push_backs?

vivekkhandelwal1:
@ramiro050, please review this PR.

pashu123 (approved):
LGTM.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3269,32 +3269,52 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
...
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
       return failure();
-    Location loc = op.getLoc();
+    Location loc = op->getLoc();
+
+    SmallVector<Value, 3> opArguments;
+    int64_t elementValue;
+
+    if (AtenOnesOp onesOp = dyn_cast<AtenOnesOp>(op)) {
+      opArguments.insert(opArguments.end(),
```
ramiro050, line 3293: nit: It might make things clearer to use `opArguments = {...};` (or if that doesn't work, you can do `opArguments.assign({...});` The argument `opArguments.end()` makes it seem as if there could be other arguments already in the vector.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3269,32 +3269,52 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
...
+      elementValue = 1;
+    } else if (AtenZerosOp zerosOp = dyn_cast<AtenZerosOp>(op)) {
+      opArguments.insert(opArguments.end(), {zerosOp.size(), zerosOp.layout(),
+                                             zerosOp.pin_memory()});
+      elementValue = 0;
+    }
 
     // We ignore device, but add simple asserts for unimplemented kwargs
-    if (!op.layout().getType().isa<Torch::NoneType>())
+    if (!opArguments[1].getType().isa<Torch::NoneType>())
```
ramiro050, line 3303: I'm not sure how I feel about using this approach. It feels like we are losing a bit of readability because now we don't know what we are getting the type of without going back to the definition of `opArguments`. You could define variables

```
Value size;
Value layout;
// etc
```

and then assign to them in the `if-else` rather than using the vector.

ramiro050 (changes requested):
LGTM! I just have a couple of stylistic comments

PR #465, opened 2021-12-08
Add support for passing & returning memref of bool types

pashu123:
Support for passing memref of bool types as a function argument
and return type is added in the ref-backend.

### DIFF
--- a/e2e_testing/torchscript/basic.py
+++ b/e2e_testing/torchscript/basic.py
@@ -1003,3 +1003,57 @@ class ZerosModuleFalsePinMemory(torch.nn.Module):
 @register_test_case(module_factory=lambda: ZerosModuleFalsePinMemory())
 def ZerosModuleFalsePinMemory_basic(module, tu: TestUtils):
     module.forward()
+
+
+class BoolTensorReturnFalseModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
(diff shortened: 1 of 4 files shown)


e2e_testing/torchscript/basic.py
```diff
@@ -925,3 +925,54 @@ def forward(self, input):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1], torch.bool, True),
+    ])
+
+    def forward(self,a):
```
silvasean, line 939: nit: format this file (I use autopep8)

e2e_testing/torchscript/basic.py
```diff
@@ -925,3 +925,54 @@ def forward(self, input):
 @register_test_case(module_factory=lambda: NumelZeroRankModule())
 def NumelZeroRankModule_basic(module, tu: TestUtils):
     module.forward(torch.randint(10,[]))
+
+class BoolReturnFalseModule(torch.nn.Module):
```
silvasean, line 929: You might want to call these "BoolTensor" instead of just "Bool", since there is also returning of `!torch.bool` (python scalar booleans) eventually.

silvasean (approved):
Thanks!

e2e_testing/torchscript/basic.py
pashu123: Done, thanks.

PR #466, opened 2021-12-08
[MLIR][TORCH] Add E2E support for aten.index_select op

vivekkhandelwal1:
This commit adds lowering of `aten.index_select` op.

### DIFF
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -3437,6 +3437,82 @@ public:
 };
 } // namespace
 
+namespace {
+// Let's say we have an input tensor: initialized with some random values of
+// size [4, 5, 6]. An index tensor (always 1-d): [0, 2] of size [2], and an
+// integer argument dim = 1. The size of the output tensor will be [4, 2, 6].
+// The approach is as follows:
+//
+// for i in range(input.size[0])
+//    for j in range(index.size[0])
+//       for k in range(input.size[2])
+//          indexValue = index[j]
+//          output[i,j,k] = input[i,indexValue,k]
+
+class ConvertAtenIndexSelectOp : public OpConversionPattern<AtenIndexSelectOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenIndexSelectOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    Location loc = op.getLoc();
+    Value input = adaptor.self();
+    Value indices = adaptor.index();
+    RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+    RankedTensorType resultType = getTypeConverter()
+                                      ->convertType(op->getResult(0).getType())
+                                      .cast<RankedTensorType>();
+    Type elementType = resultType.getElementType();
+    unsigned inputRank = inputType.getRank();
+
+    int64_t dimInt;
+    if (!matchPattern(op.dim(), m_TorchConstantInt(&dimInt)))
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3394,6 +3394,87 @@ class ConvertAtenNumelOp : public OpConversionPattern<AtenNumelOp> {
 };
 } // namespace
 
+namespace {
```
pashu123, line 3397: Please add doc-comments and write the pseudocode of the algorithm, it will be easier to follow.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3394,6 +3394,87 @@ class ConvertAtenNumelOp : public OpConversionPattern<AtenNumelOp> {
...
+                                      .cast<RankedTensorType>();
+    Type elementType = resultType.getElementType();
+    unsigned inputRank = inputType.getRank();
+
+    int64_t dimInt;
+    if (!matchPattern(op.dim(), m_TorchConstantInt(&dimInt)))
+      return op->emitError("unimplemented: dim is not constant");
+
+    Value initTensor;
+    if (inputType.hasStaticShape() && indicesType.hasStaticShape()) {
```
silvasean, line 3423: why is there a special case for static shape? We generally avoid that (except possibly for static size-1 dimensions) in TorchToLinalg.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3394,6 +3394,87 @@ class ConvertAtenNumelOp : public OpConversionPattern<AtenNumelOp> {
...
+
+    Value finalRes =
+        rewriter
+            .create<linalg::GenericOp>(
+                loc, resultType, ValueRange{input, indices}, initTensor,
+                /*indexingMaps=*/indexingMaps,
+                /*iteratorTypes=*/iteratorTypes,
+                [&](OpBuilder &b, Location loc, ValueRange args) {
+                  Value indexInput =
+                      rewriter.create<linalg::IndexOp>(loc, dimInt);
```
silvasean, line 3461: This payload doesn't seem right to me. index_select is basically a "gather", so I would expect the payload to contain a tensor.extract op.

silvasean (changes requested):

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
vivekkhandelwal1: Here we have `args[0]`, an element of the `input` tensor, and we are using it. That's why we didn't require an extract tensor op.

pashu123: I think so with `tensor_extract` op you have to potential to remove the last `iter_var`. So your algorithm would look something like: 
``` 
for i in range(input.size[0]): 
       for j in range (index.size[0]):
           for k in range(input.size[2]):
               index_val = index[j]
               output[i][j][k] = tensor.extract(input, i, index_val, j)
```
Remember, for `tensor.extract` op the accesses don't have to be affine whereas for the payload args it has to be affine.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3437,6 +3437,93 @@ class ConvertAtenNumelOp : public OpConversionPattern<AtenNumelOp> {
 };
 } // namespace
 
+namespace {
+// Converts AtenIndexSelectOp.
+//
```
pashu123, line 3442: This can be removed as it's obvious from the class definition.

pashu123 (changes requested):

silvasean (approved):
Looks great!

pashu123 (approved):
LGTM!

PR #467, opened 2021-12-08
[TORCH][MLIR] Add E2E support for [`aten.gt.Scalar`|`aten.where.self`]

Shukla-Gaurav:
This commit adds lowering of `aten.gt.Scalar` and `aten.where.self` as a
part of element-wise ops lowering.


silvasean (approved):
Thanks!

PR #468, opened 2021-12-08
Linalg lowering for aten.conv2d(bias=True)

ljfitz:
Previously aten.conv2d was only lowered if there was no bias.
Here lowering is extended to support bias.


silvasean:
Thanks!

PR #471, opened 2021-12-09
Add where, gt, bucketize and reshape ops to Torch dialect

harsh-nod:
This patch adds the where, gt, bucketize and reshape
ops to the Torch dialect. These ops are present in the histogram
calibration module.

TEST: Successfully lowers ops to Torch dialect in histogram module.


ramiro050 (approved):
LGTM!

PR #472, opened 2021-12-10
Add `aten.gt.Tensor` op

pashu123:
`aten.gt.Tensor` op has been added in torch dialect and the
lowering of the op has been done to the linalg dialect.

### DIFF
--- a/e2e_testing/torchscript/elementwise.py
+++ b/e2e_testing/torchscript/elementwise.py
@@ -343,6 +343,42 @@ class ElementwiseGtScalarModule(torch.nn.Module):
 def ElementwiseGtScalarModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 5))
 
+class ElementwiseGtFloatTensorModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+        ([-1], torch.float32, True),
+    ])
+    def forward(self, x, y):
+        return torch.gt(x, y)
+
+
+@register_test_case(module_factory=lambda: ElementwiseGtFloatTensorModule())
+def ElementwiseGtFloatTensorModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 5), tu.rand(5))
+
+class ElementwiseGtIntTensorModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.int64, True),
+        ([-1], torch.int64, True),
+    ])
+    def forward(self, x, y):
+        return torch.gt(x, y)
+
+
+@register_test_case(module_factory=lambda: ElementwiseGtIntTensorModule())
+def ElementwiseGtIntTensorModule_basic(module, tu: TestUtils):
+    module.forward(torch.randint(10, (3, 5)), torch.randint(10, (5,)))
+
 # ==============================================================================
 
 
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -1669,6 +1669,32 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
       return b.create<arith::MulIOp>(loc, lhs, rhs);
     }
   }
+  if (auto gtTensor = dyn_cast<AtenGtTensorOp>(op)) {
+    AtenGtTensorOp::Adaptor adaptor(operands);
+    Type lhsDtype = payloadArgs[0].getType();
+    Type rhsDtype = payloadArgs[1].getType();
+
+    // TODO: Type promotion in case of different `lhsDtype` and `rhsDtype` needs
+    // to be handled.
+    if (lhsDtype != rhsDtype)
+      gtTensor.emitError("unimplemented: different lhs and rhs dtype");
+
+    Type elementalType =
+        gtTensor.self().getType().cast<BaseTensorType>().getDtype();
+
+    if (elementalType.isa<mlir::FloatType>())
+      return b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGT,
+                                     payloadArgs[0], payloadArgs[1]);
+    if (IntegerType intType = elementalType.dyn_cast<mlir::IntegerType>()) {
+      if (intType.isUnsigned())
+        return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ugt,
+                                       payloadArgs[0], payloadArgs[1]);
+      if (intType.isSigned())
+        return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sgt,
+                                       payloadArgs[0], payloadArgs[1]);
+    }
+    gtTensor.emitError("unimplemented: dtype isn't supported.");
+  }
   if (auto div = dyn_cast<AtenDivTensorOp>(op)) {
     AtenDivTensorOp::Adaptor adaptor(operands);
     Type dtype = converter->convertType(div.getType())
@@ -2070,7 +2096,7 @@ struct ConvertElementwiseOp : ConversionPattern {
              AtenSqrtOp, AtenFloorOp, AtenPowTensorScalarOp, AtenLog2Op,
              AtenRsqrtOp, AtenDivScalarOp, AtenAbsOp, AtenReciprocalOp,
              AtenBitwiseAndTensorOp, AtenGtScalarOp, AtenWhereSelfOp,
-             AtenCeilOp>(op))
+             AtenCeilOp, AtenGtTensorOp>(op))
       return rewriter.notifyMatchFailure(op, "not a supported elementwise op");
 
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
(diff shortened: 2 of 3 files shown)


lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1661,6 +1661,20 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    Type dtype = payloadArgs[0].getType();
+
+    // TODO: Currently, the dominating dtype is the lhs type whereas the
+    // bitwidth of lhs and rhs should be compared to get the resulting dtype.
+    Value lhs = convertScalarToDtype(b, loc, payloadArgs[0], dtype);
+    Value rhs = convertScalarToDtype(b, loc, payloadArgs[1], dtype);
+    if (dtype.isa<mlir::FloatType>()) {
+      return b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGT, lhs, rhs);
+    } else {
+      return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sgt, lhs, rhs);
```
ramiro050, line 1675: I don't think this is exhaustive. In particular, `torch.uint8` is not handled correctly by this line because you're using the signed comparison. I would add an `else-if` branch checking if `dtype.isSignedInteger()` (https://mlir.llvm.org/doxygen/classmlir_1_1Type.html#a50385bcb919f49adab0f8f2521dc479f), and the else can return an error message saying that the only supported types at the moment are floats and signed ints.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1661,6 +1661,20 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
       return b.create<arith::MulIOp>(loc, lhs, rhs);
     }
   }
+  if (auto gtTensor = dyn_cast<AtenGtTensorOp>(op)) {
+    AtenGtTensorOp::Adaptor adaptor(operands);
+    Type dtype = payloadArgs[0].getType();
+
+    // TODO: Currently, the dominating dtype is the lhs type whereas the
+    // bitwidth of lhs and rhs should be compared to get the resulting dtype.
```
ramiro050, line 1669: Here you have to worry not just about bit-width, but also about whether a type is a float or int. At the moment this implementation will not work correctly if the first input tensor is an int tensor and the second is a float because the float will get truncated to an int.

ramiro050 (changes requested):

e2e_testing/torchscript/elementwise.py
```diff
@@ -343,6 +343,24 @@ def forward(self, x):
...
 
+class ElementwiseGtTensorModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+        ([-1], torch.float32, True),
```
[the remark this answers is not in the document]
pashu123: I have not included it. Since there are many things that I will have to clarify regarding integer types.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
pashu123: I see, torch.int32 is seen as a signless integer in payload args. Currently, I am only supporting float types.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
pashu123: I see, I will work on the type-promotion specific to this later. But thanks for your inputs.

e2e_testing/torchscript/elementwise.py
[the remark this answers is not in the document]
pashu123: I have included it.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
pashu123: I have tried supporting the integer types as well. Please review.

e2e_testing/torchscript/elementwise.py
```diff
@@ -343,6 +343,42 @@ def forward(self, x):
...
+        ([-1, -1], torch.int32, True),
+        ([-1], torch.int32, True),
+    ])
+    def forward(self, x, y):
+        return torch.gt(x, y)
+
+
+@register_test_case(module_factory=lambda: ElementwiseGtIntTensorModule())
+def ElementwiseGtIntTensorModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 5).to(torch.int32), tu.rand(5).to(torch.int32))
```
Shukla-Gaurav, line 380: I guess `tu.rand().to(torch.int32)` will always produce 0, instead you should use `torch.randint()`.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1669,6 +1669,32 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+  if (auto gtTensor = dyn_cast<AtenGtTensorOp>(op)) {
+    AtenGtTensorOp::Adaptor adaptor(operands);
+    Type lhsDtype = payloadArgs[0].getType();
+    Type rhsDtype = payloadArgs[1].getType();
+
+    Type elementalType =
+        gtTensor.self().getType().cast<BaseTensorType>().getDtype();
+
+    if (lhsDtype != rhsDtype)
+      gtTensor.emitError("unimplemented: different lhs and rhs dtype");
```
Shukla-Gaurav, line 1681: Add a TODO here to handle type promotion.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1669,6 +1669,32 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
       return b.create<arith::MulIOp>(loc, lhs, rhs);
     }
   }
+  if (auto gtTensor = dyn_cast<AtenGtTensorOp>(op)) {
+    AtenGtTensorOp::Adaptor adaptor(operands);
+    Type lhsDtype = payloadArgs[0].getType();
+    Type rhsDtype = payloadArgs[1].getType();
+
+    Type elementalType =
+        gtTensor.self().getType().cast<BaseTensorType>().getDtype();
```
Shukla-Gaurav, line 1678: I guess you don't have to compute this `elementType`, you can use`lhsDtype` or `rhsDtype`.

Shukla-Gaurav (changes requested):
LGTM! Few minor comments to handle.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
pashu123: The problem here is signless integers; when you remain in torch dialect you have the knowledge of whether the integer is signed or not, but when you some to `linalg on tensors` world the integer becomes signless. This has to be looked into.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
pashu123: Done.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1669,6 +1669,34 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+
+    // TODO: Type promotion in case of different `lhsDtype` and `rhsDtype` needs
+    // to be handled.
+    if (lhsDtype != rhsDtype)
+      gtTensor.emitError("unimplemented: different lhs and rhs dtype");
```
Shukla-Gaurav, line 1683: Move this above the `elementType` computation for early exit.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1669,6 +1669,34 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+                 elementalType.dyn_cast<mlir::IntegerType>()) {
+      if (intType.isUnsigned())
+        return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ugt,
+                                       payloadArgs[0], payloadArgs[1]);
+      if (intType.isSigned())
+        return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sgt,
+                                       payloadArgs[0], payloadArgs[1]);
+    }
+    else
+      gtTensor.emitError("unimplemented: dtype isn't supported.");
```
Shukla-Gaurav, line 1698: You should remove `else` from here so that all the unhandled cases must fall here.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1669,6 +1669,34 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    // TODO: Type promotion in case of different `lhsDtype` and `rhsDtype` needs
+    // to be handled.
+    if (lhsDtype != rhsDtype)
+      gtTensor.emitError("unimplemented: different lhs and rhs dtype");
+
+    if (elementalType.isa<mlir::FloatType>())
+      return b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGT,
+                                     payloadArgs[0], payloadArgs[1]);
+    else if (IntegerType intType =
+                 elementalType.dyn_cast<mlir::IntegerType>()) {
```
Shukla-Gaurav, line 1689: else if -> if.

Shukla-Gaurav (approved):
Please address other minor comments. I am approving now.

Shukla-Gaurav (approved):
LGTM!

PR #473, opened 2021-12-11
[TORCH][MLIR] Add E2E support for `aten.ceil` op

Shukla-Gaurav:
This commit adds lowering of `aten.ceil` op as a part of element-wise
ops lowering.


ramiro050 (approved):
LGTM!

PR #474, opened 2021-12-12
[TORCH][MLIR] Templatize lowering of [`aten.zeros|aten.ones|aten.empty`] ops

Shukla-Gaurav:
- Templatize `aten.zeros` and `aten.ones` ops lowering.
- Add E2E support for `aten.empty` op.
- Add Integer type support in `aten.mul.Scalar` op lowering.

### DIFF
--- a/e2e_testing/torchscript/basic.py
+++ b/e2e_testing/torchscript/basic.py
@@ -637,6 +637,57 @@ def OnesModuleFalsePinMemory_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
+class EmptyIntModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return 0 * torch.empty((3, 4), dtype=torch.int64)
+
+@register_test_case(module_factory=lambda: EmptyIntModule())
+def EmptyModule_int(module, tu: TestUtils):
+    module.forward()
+
+# ==============================================================================
+
+class EmptyFloatModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.abs(torch.empty((3, 4), dtype=torch.float32)) > -1.0
+
+@register_test_case(module_factory=lambda: EmptyFloatModule())
+def EmptyModule_float(module, tu: TestUtils):
+    module.forward()
+
+
+class EmptyFalsePinMemoryModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.abs(torch.empty((3, 4), dtype=torch.float32, 
+                                     pin_memory=False)) > -1.0 
+
+@register_test_case(module_factory=lambda: EmptyFalsePinMemoryModule())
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1828,13 +1828,13 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
                      .cast<RankedTensorType>()
                      .getElementType();
-    if (!dtype.isa<mlir::FloatType>()) {
-      mulScalar.emitError("unimplemented: non-floating point dtype");
-      return nullptr;
+    Value lhs = payloadArgs[0];
+    Value rhs = convertScalarToDtype(b, loc, operands[1], dtype);
+    if (dtype.isa<mlir::FloatType>()) {
+      return b.create<arith::MulFOp>(loc, lhs, rhs);
+    } else {
```
pashu123, line 1835: Please check for type, there may be types other than float and integer. https://pytorch.org/docs/stable/tensors.html

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3417,83 +3417,83 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
...
-    Value fillOp = rewriter.create<linalg::FillOp>(loc, constantOp, outTensor)
-                       .getResult(0);
-
-    rewriter.replaceOpWithNewOp<tensor::CastOp>(op, newResultType, fillOp);
+    auto resultType =
+        typeConverter->convertType(op.getType()).template cast<RankedTensorType>();
+    Type outElemType = resultType.getElementType();
+
+    // Returns a constant of type `elemType` with value `val`.
+    auto getConstant = [&](int64_t val, Type elemType) -> Value {
```
pashu123, line 3467: This is a very generic function and can share functionalities across many functions. Instead of a lambda can be placed outside as a static function.

e2e_testing/torchscript/basic.py
```diff
@@ -637,6 +637,56 @@ def OnesModuleFalsePinMemory_basic(module, tu: TestUtils):
...
+class EmptyFalsePinMemoryModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.abs(torch.empty((3, 4), dtype=torch.float32, pin_memory=False)) > -1.0
```
pashu123, line 682: Nit: Please format your python code.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
pashu123: Add a test case where types are not the same i.e., `torch.mul(x, 4.0)` where x is torch.int64.

pashu123 (changes requested):

pashu123 (approved):
Great effort! LGTM.

Shukla-Gaurav:
@silvasean This PR uses a C++17 feature `if constexpr ()`. Should I fix this right away in a follow-up PR or Is it okay to leave it as it is? cc: @ramiro050

silvasean:
I would avoid `if constexpr`. In this case, I would recommend to not use a template for this. Just pass `Operation *op` -- you can still `dyn_cast`/`isa`.

PR #475, opened 2021-12-13
Add support for int types in gtScalar op.

pashu123:
Support for integer types in gtScalar op has been added.
The code share same logic with gtTensor op and can be merged
which is added as a TODO.

### DIFF
--- a/e2e_testing/torchscript/elementwise.py
+++ b/e2e_testing/torchscript/elementwise.py
@@ -326,7 +326,7 @@ def ElementwiseMaximumModule_basic(module, tu: TestUtils):
 # ==============================================================================
 
 
-class ElementwiseGtScalarModule(torch.nn.Module):
+class ElementwiseGtFloatScalarModule(torch.nn.Module):
     def __init__(self):
         super().__init__()
 
@@ -339,9 +339,46 @@ class ElementwiseGtScalarModule(torch.nn.Module):
         return torch.gt(x, 0.6)
 
 
-@register_test_case(module_factory=lambda: ElementwiseGtScalarModule())
-def ElementwiseGtScalarModule_basic(module, tu: TestUtils):
+@register_test_case(module_factory=lambda: ElementwiseGtFloatScalarModule())
+def ElementwiseGtFloatScalarModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 5))
+
+
+class ElementwiseGtIntScalarModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.int64, True),
+    ])
+    def forward(self, x):
+        return torch.gt(x, 10)
+
+
+@register_test_case(module_factory=lambda: ElementwiseGtIntScalarModule())
+def ElementwiseGtIntScalarModule_basic(module, tu: TestUtils):
+    module.forward(torch.randint(-10, 15, (3,4)))
+
+
+class ElementwiseGtMixed2ScalarModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.int32, True),
+    ])
+    def forward(self, x):
+        return torch.gt(x, 7)
+
+
+@register_test_case(module_factory=lambda: ElementwiseGtMixed2ScalarModule())
+def ElementwiseGtMixed2ScalarModule_basic(module, tu: TestUtils):
+    module.forward(torch.randint(-10, 15, (3,4)).to(torch.int32))
+
 
 class ElementwiseGtFloatTensorModule(torch.nn.Module):
     def __init__(self):
@@ -360,6 +397,7 @@ class ElementwiseGtFloatTensorModule(torch.nn.Module):
 @register_test_case(module_factory=lambda: ElementwiseGtFloatTensorModule())
 def ElementwiseGtFloatTensorModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 5), tu.rand(5))
+
 
 class ElementwiseGtIntTensorModule(torch.nn.Module):
     def __init__(self):
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -1720,14 +1720,33 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
   }
 
   if (auto gtScalar = dyn_cast<AtenGtScalarOp>(op)) {
-    Type dtype = gtScalar.self().getType().cast<ValueTensorType>().getDtype();
-    if (!dtype.isa<mlir::FloatType>()) {
-      gtScalar.emitError("unimplemented: non-floating point operand dtype");
-      return nullptr;
+    Type dtype = gtScalar.self().getType().cast<BaseTensorType>().getDtype();
+
+    // TODO: `gtTensor` and `gtScalar` share similar code and can be called from
+    // one static function.
+    Value otherPromoted =
+        convertScalarToDtype(b, loc, operands[1], payloadArgs[0].getType());
+
+    if (dtype.isa<mlir::FloatType>())
+      return b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGT,
+                                     payloadArgs[0], otherPromoted);
+    if (IntegerType intType = dtype.dyn_cast<mlir::IntegerType>()) {
+      if (!operands[1].getType().isa<mlir::IntegerType>()) {
+        // TODO: Promote tensor args from integer to float.
+        gtScalar.emitError(
+            "unimplemented: type promotion from tensor to scalar.");
+        return nullptr;
+      }
+
+      if (intType.isUnsigned())
+        return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ugt,
+                                       payloadArgs[0], otherPromoted);
+      if (intType.isSigned())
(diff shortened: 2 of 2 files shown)


e2e_testing/torchscript/elementwise.py
```diff
@@ -325,23 +325,40 @@ def ElementwiseMaximumModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
-
-class ElementwiseGtScalarModule(torch.nn.Module):
+class ElementwiseGtScalarFloatModule(torch.nn.Module):
```
ramiro050, line 328: nit: it seems like the convention below is to do `FloatScalar` rather than `ScalarFloat`

ramiro050 (approved):
LGTM! I just have a small style comment

e2e_testing/torchscript/elementwise.py
```diff
@@ -86,23 +86,27 @@ def ElementwiseTernaryModule_basic(module, tu: TestUtils):
 
 
 class ElementwiseWhereSelfModule(torch.nn.Module):
-    def __init__(self):
-        super().__init__()
 
-    @export
-    @annotate_args([
-        None,
-        ([-1, -1, -1], torch.float32, True),
```
ramiro050, line 95: why was this changed?

e2e_testing/torchscript/elementwise.py
```diff
@@ -86,23 +86,27 @@ def ElementwiseTernaryModule_basic(module, tu: TestUtils):
...
-        ([-1, -1], torch.float32, True),
-        ([-1], torch.float32, True),
-    ])
-    def forward(self, a, b, c):
-        return torch.where(a > 0.5, b, c)
+  def __init__(self):
```
ramiro050, line 90: nit: this file uses 4 spaces for indentation

ramiro050 (changes requested):

e2e_testing/torchscript/elementwise.py
pashu123: `a>0.5` where a is of type float32 whereas 0.5 is to type float64. This is a type mismatch and aten.gt.Scalar would throw an error since type promotion is not handled there.

pashu123: Previously, it wasn't promoting a to match type of (0.5) which is wrong, it should have promoted a.

ramiro050: Type promotion is a very confusing and messy thing 😂 when it comes to scalars that are part of a tensor operation, if the scalar is a float, it gets converted to a `float32` (the default float type) and then the promotion takes place between the tensor and the scalar.

https://github.com/pytorch/pytorch/blob/deb6989880d8a62bb45ce6c6b058bb5d2e28cf91/aten/src/ATen/native/TypeProperties.cpp#L98-L105

If the scalar is part of an op that does not involve tensors, such as `NumToTensor` then the scalar does get converted to `float64` if it was a float scalar.

Therefore, for the case where the input tensor is a float tensor, all you have to do is convert the scalar to match the type of the float tensor.

ramiro050: In general, it is not a good practice to remove a unit test once the unit test has been added. New changes should not break features that were already supported. I recommend undoing the changes you made for the float tensor types and try to get the lowering to work correctly with them

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1720,14 +1720,25 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+
+    // TODO: `gtTensor` and `gtScalar` share similar code and can be called from
+    // one static function.
+    Value otherPromoted =
+        convertScalarToDtype(b, loc, operands[1], payloadArgs[0].getType());
+
+    if (dtype.isa<mlir::FloatType>())
+      return b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGT,
+                                     payloadArgs[0], otherPromoted);
+    if (IntegerType intType = dtype.dyn_cast<mlir::IntegerType>()) {
```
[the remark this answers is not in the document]
pashu123: Same is the case when tensor is of type int32 and scalar argument is of type int64? You are losing information in this case too.

e2e_testing/torchscript/elementwise.py
```diff
@@ -339,10 +339,29 @@ def forward(self, x):
...
+        None,
+        ([-1, -1], torch.int64, True),
+    ])
+    def forward(self, x):
+        return torch.gt(x, 100)
+
+
+@register_test_case(module_factory=lambda: ElementwiseGtIntScalarModule())
+def ElementwiseGtIntScalarModule_basic(module, tu: TestUtils):
+    module.forward(torch.randint(3, (3,4)))
```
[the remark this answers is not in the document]
pashu123: Cool. I will add that.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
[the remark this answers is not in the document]
pashu123: @ramiro050 I have added more test cases. For the above case, I can add a TODO there, and that can be taken later.

e2e_testing/torchscript/elementwise.py
```diff
@@ -339,10 +339,65 @@ def forward(self, x):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.int64, True),
+    ])
+    def forward(self, x):
+        return torch.gt(x, 7.0)
```
ramiro050, line 375: Here the important test is making sure that we are not rounding the float value to an integer. So you want to make sure that `torch.gt(x, 7.7)` and `torch.gt(x, 7.3)` work correctly. (they currently will not work correctly because the lowering is turning the float into an int).

It's okay to have features not supported, as long as they result in an unimplemented error. Since the functionality you want right now is `torch.gt(int, int)`, I would just make sure in the `int` branch of the lowering that both the scalar and the tensor are integers. Then leave this case as unimplemented and for a future patch to solve.

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1720,14 +1720,25 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
   }
 
   if (auto gtScalar = dyn_cast<AtenGtScalarOp>(op)) {
-    Type dtype = gtScalar.self().getType().cast<ValueTensorType>().getDtype();
-    if (!dtype.isa<mlir::FloatType>()) {
-      gtScalar.emitError("unimplemented: non-floating point operand dtype");
-      return nullptr;
+    Type dtype = gtScalar.self().getType().cast<BaseTensorType>().getDtype();
+
+    // TODO: `gtTensor` and `gtScalar` share similar code and can be called from
```
silvasean, line 1725: One side comment thought: Would it make sense to add a rule to ReduceOpVariants to convert `aten.foo.Scalar(x, y)` to `aten.foo.Tensor(x, NumToTensor(y))`? I think we can detect this case and add a trait "ScalarVariantWithCorrespondingTensorVariant" to the ops that have this property. It would look similar to how we handle emit_with_mutating_variants in https://github.com/llvm/torch-mlir/blob/528354de84c5892d539fce17c74b2c928a645991/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py#L430

We would probably rename emit_with_mutating_variants -> emit_elementwise_op and have it detect which mutating/scalar/tensor variants exist, emit them, and add all the necessary traits.


Will let Ramiro drive this review, which seems to be going in a good direction anyway. Please submit after he gives approval.

silvasean (commented):
Will let Ramiro drive this review, which seems to be going in a good direction anyway. Please submit after he gives approval. I left a comment about generalizing this though.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
ramiro050: I think adding something like this would be really nice. We just have to be careful about the type conversions that happened when a scalar is turned into a tensor. In particular, `NumToTensor` would have the wrong behavior because, for example, it converts scalar floats to `float64`, which would mess with the type promotion when performing the tensor computation. What we want is something that behaves like the `wrapped_number` in PyTorch, which follows a different set of type conversion rules.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1720,14 +1720,32 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+        return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ugt,
+                                       payloadArgs[0], otherPromoted);
+      if (intType.isSigned())
+        return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sgt,
+                                       payloadArgs[0], otherPromoted);
     }
-    Value otherPromoted = convertScalarToDtype(b, loc, operands[1], dtype);
-    return b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGT,
-                                   payloadArgs[0], otherPromoted);
+    gtScalar.emitError("unimplemented: dtype isn't supported.");
```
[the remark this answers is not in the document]
pashu123: done

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
silvasean: +1, yeah, there are probably some subtle cases.

Issue #448, opened 2021-11-29
Tensors with dim values of 0 can exist in pytorch but break RefBackend


dan-garvey:
Not even sure if this qualifies as an issue, but if a lowering results in a rank>1 tensor with a dim of 0, the refbackend will crash. I'm not certain a case where this actually occurs in a model exists? 
Below is an example of torch code that produces such a result. 
`
a =torch.randint(10, (5,5))

print(a)
tensor([[5, 7, 2, 6, 5],
[3, 5, 2, 1, 2],
[6, 9, 6, 7, 5],
[8, 3, 9, 6, 2],
[7, 9, 9, 3, 7]])
print(a[:0,:0])
tensor([], size=(0, 0), dtype=torch.int64)
`

The error that will happen:
Exception ignored on calling ctypes callback function: <function RefBackendInvoker.__init__.<locals>.consume_return_mrf32 at 0x7fefd87993a0>
Traceback (most recent call last):                                                                                                                                                                     
`  File "/home/dan/my_gits/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py", line 46, in consume_return_mrf32
    self.result = unranked_memref_to_numpy(a, np.float32)                                                                                                                                              
  File "/home/dan/my_gits/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/runtime/np_to_memref.py", line 98, in unranked_memref_to_numpy                                       
    np_arr = np.ctypeslib.as_array(val[0].aligned, shape=val[0].shape)                                                                                                                                 
  File "/home/dan/my_gits/torch-mlir/mlir_venv/lib/python3.8/site-packages/numpy/ctypeslib.py", line 518, in as_array                                                                                  
    return asarray(obj)                                                                                                                                                                                
  File "/home/dan/my_gits/torch-mlir/mlir_venv/lib/python3.8/site-packages/numpy/core/_dtype_ctypes.py", line 106, in dtype_from_ctypes_type                                                           
    return _from_ctypes_array(t)                                                                                                                                                                       
  File "/home/dan/my_gits/torch-mlir/mlir_venv/lib/python3.8/site-packages/numpy/core/_dtype_ctypes.py", line 33, in _from_ctypes_array                                                                
    return np.dtype((dtype_from_ctypes_type(t._type_), (t._length_,)))                                                                                                                                 
[...]

PR #476, opened 2021-12-13
Small fixes for slice edge cases

dan-garvey:
Adds a check for optional type, and covers the case where end < start. Thanks to @cathyzhyi for pointing these out.

### DIFF
--- a/e2e_testing/torchscript/slice_like.py
+++ b/e2e_testing/torchscript/slice_like.py
@@ -32,7 +32,6 @@ def SliceModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
-# This Test currently xfails due to https://github.com/llvm/torch-mlir/issues/448
 class SliceOutOfUpperBoundIndexModule(torch.nn.Module):
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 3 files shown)


silvasean:
Please add tests showing the new functionality :)

dan-garvey:
Unfortunately this case goes from xfail to xfail, as it results in 0 size tensors which still crash the refbackend #448

Do we want to create tests for xfail on Optional type coverage?

silvasean:
is there a way to hack around that temporarily? for example, to cat the zero size tensor with a tensor of size 1 so the final tensor is of size 1?

dan-garvey:
Have they tried cloning you yet

e2e_testing/torchscript/slice_like.py
```diff
@@ -121,7 +126,10 @@ def __init__(self):
         ([-1, -1, -1], torch.float32, True),
     ])
     def forward(self, x):
-        return x[5:5, 3:3, -1:]
+        #TODO: remove hacky cat tensor once refbackend supports 0 size dim
```
silvasean, line 129: nit: space before TODO

silvasean (approved):
Go ahead and merge after addressing nit :)

PR #477, opened 2021-12-13
[TORCH][MLIR] Add E2E support for `aten.empty_like` op

Shukla-Gaurav:
This commit adds decomposition of `aten.empty_like` into `aten.empty`
op.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -506,6 +506,25 @@ class DecomposeAtenLayerNormOp : public OpRewritePattern<AtenLayerNormOp> {
 };
 } // namespace
 
+// Decompose .
```
pashu123, line 509: Nit: Doc comments need to be updated.

pashu123 (approved):
LGTM!

PR #478, opened 2021-12-13
[WIP][MLIR][TORCH] Add E2E support for aten.cumsum.

gprateek93:


vivekkhandelwal1:
I doubt the correctness of indexing maps in this code. They don't seem correct to me. If someone else can also look at that, it would be good.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3594,6 +3594,72 @@ class ConvertAtenIndexSelectOp : public OpConversionPattern<AtenIndexSelectOp> {
...
+        indexExprs.push_back(getAffineDimExpr(i, rewriter.getContext()) -
+                             getAffineConstantExpr(1, context));
+      } else {
+        indexExprs.push_back(getAffineDimExpr(i, rewriter.getContext()));
+      }
+    }
+    auto cumsumResultMap = AffineMap::get(inputRank, 0, indexExprs, context);
+    SmallVector<AffineMap> indexingMaps = {
+        rewriter.getMultiDimIdentityMap(inputRank), cumsumResultMap};
+    SmallVector<StringRef> iteratorTypes(inputRank, "parallel");
```
cathyzhyi, line 3631: nit: use `getParallelIteratorTypeName());`

cathyzhyi: should it be "reduction" on the `dim` dimension?

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3594,6 +3594,72 @@ class ConvertAtenIndexSelectOp : public OpConversionPattern<AtenIndexSelectOp> {
...
+    Value cumsumTensor =
+        createZeroInitTensor(rewriter, loc, resultOutputDims, elementType);
+    // llvm::errs()<<"before\n";
+    auto cumsumResult =
+        rewriter
+            .create<linalg::GenericOp>(
+                loc, cumsumTensor.getType(), input, cumsumTensor, indexingMaps,
+                iteratorTypes,
+                [&](OpBuilder &b, Location loc, ValueRange args) {
+                  Value currIndex = b.create<linalg::IndexOp>(loc, 0);
```
cathyzhyi, line 3644: I think `currIndex` should be `b.create<linalg::IndexOp>(loc, dim);`

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3594,6 +3594,72 @@ class ConvertAtenIndexSelectOp : public OpConversionPattern<AtenIndexSelectOp> {
...
+                  Value result = b.create<arith::AddFOp>(loc, args[0], args[1]);
+                  Value indexEqualToZero = b.create<arith::CmpIOp>(
```
cathyzhyi, line 3646: We need to verify from a log but my understanding is that `currIndex` would start from 1 and never equal to 0 because in such case the indexing of `cumsumTensor` would be out of bound (currIndex - 1 < 0) and linalg generic op should guarantee the indexing is never out of bound according to https://mlir.llvm.org/docs/Dialects/Linalg/.  Could you verify that?

cathyzhyi:
I would initialize the `cumsumTensor ` with input slicing 0 from `dim` dimension and the rest of the code, including the payload and indexing map mostly look good to me. 

We can using `tensor::extract_slice` to extract 0 along `dim` dimension and use `tensor::insert_slice` to initialize  `cumsumTensor `.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
gprateek93: No, the input and output dimensions are same in cumsum. There is no reduction.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
gprateek93: Yes, Thanks for pointing out this.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
gprateek93: I have verified it. It throws error ``linalg.generic' op unexpected result less than 0 at expression #0 in (d0) -> (d0 - 1)``

gprateek93:
Hi @cathyzhyi @ramiro050 @pashu123, this patch works. Please review.

e2e_testing/torchscript/basic.py
```diff
@@ -1057,3 +1057,20 @@ def forward(self, a):
...
+
+
+class CumsumModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([2, 2, 3], torch.float32, True),
```
pashu123, line 1069: Please check for dynamic shape inputs.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3734,6 +3734,90 @@ class ConvertAtenIndexSelectOp : public OpConversionPattern<AtenIndexSelectOp> {
...
+    Value firstSlice = rewriter.create<tensor::ExtractSliceOp>(
+        loc, input, offsets, sliceBounds, strides);
+    cumsumTensor = rewriter.create<tensor::InsertSliceOp>(
+        loc, firstSlice, cumsumTensor, offsets, sliceBounds, strides);
+    SmallVector<AffineMap> indexingMaps(
+        2, rewriter.getMultiDimIdentityMap(inputRank));
+    SmallVector<StringRef> iteratorTypes(inputRank, "parallel");
+
+    // Perform cumulative summation.
+    for (int i = 1; i < inputShape[dim]; i++) {
```
ramiro050, line 3784: I'm a bit worried about the code bloating that this could cause if `inputShape[dim]` is very large

ramiro050: It seems that what is making things very complicated is the fact that tensors get treated as immutable objects when passed as arguments to the `Linalg.generic` body, so you can't read back a value written by a previous iteration of the `Linalg.generic` body. I wonder if working with `memref`s might solve this issue, or maybe we need to lower this op at a lower level, such as with `scf.for` ops

silvasean:
Can we do the same thing as tf.cumsum gets lowered to linalg (first to mhlo.reduce_window, and then to linalg.generic)?

mhlo.reduce_window to linalg: https://github.com/tensorflow/mlir-hlo/blob/4f421c68959bf6b1267562edf6e1072b9a681995/lib/Dialect/mhlo/transforms/legalize_to_linalg.cc#L1907
Example lowering of cumsum to mhlo.reduce_window: https://github.com/google/iree/issues/7885

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
silvasean: We definitely don't want to lower to scf.for/memrefs directly. We need to keep it at the linalg-on-tensors abstraction level for codegen backends to be able to get good performance (e.g. parallelize the op on GPU; examples of how that is done is e.g. https://moderngpu.github.io/scan.html).

ramiro050: That's a really interesting article! Thanks for sharing :)

silvasean:
More info here: https://discord.com/channels/689900678990135345/706175614805082132/920804069562544159
Mahesh says that linalg cannot currently represent cumsum well :/

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
MaheshRavishankar: I find it hard to review without having a test here that verifies the IR being generated. I am not aware of the testing requirements in torch-MLIR, so going off just the code, its easy to miss something. Having said that, this is generate "n" reduction operations, which only works if "n" is static. I would expect you generate an `scf.for` (or at this level just normal unstructured control flow would be good enough, maybe even preferable) within which you generate the reduction. operation. Still not the best way to lower it. See related bug that came up in context of TF/MHLO https://github.com/google/iree/issues/7885 . Harsh on that bug pointed out a GEMM based lowering. That might help you in the short term.

There is a better solution point where we could represent the cummulative sum (i.e. prefix sum) operation in a more native way in MLIR. There are a few operations in this category that have been prototyped in IREE under the `linalg_ext` dialect. We'd need something similar. There is nothing in core though. The plan is to move those to ops to core but its not on the horizon.

Sorry this mostly a non-answer, but just sharing things the way I see it.

MaheshRavishankar: Also in the past, I have implemented https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda within a compiler. But this is something that should be done from MLIR lower down and not be exposed to Torch-MLIR.

silvasean: re: testing, we generally don't add lit tests for TorchToLinalg lowering, because they usually don't have much bug catching power in my experience -- they are basically just diff tests (I don't remember a single bug being caught by such a test). We do insist on thorough e2e tests to verify correctness though, as ultimately that seems to have the greatest value for catching issues at this layer. We would like to eventually move to a world where these TorchToLinalg lowerings are just executable Python snippets written in OpDSL+PyDM or similar, and individually unit tested at the Python level for execution correctness without really involving a "compiler pass" per se (really just loading up a library of functions and inlining them).

PR #482, opened 2021-12-14
Fix naming of results in ODS generator

ramiro050:
This commit fixes the naming of results in the torch ODS generator
when dealing with multiple results. In particular, this commit appends
an index to each result name to guarantee that they are all unique.

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
@@ -297,6 +297,11 @@ def raw_emit_op(operator: JitOperator, f: TextIO, *, traits: List[str],
     emitter = TextEmitter(f)
     p = lambda *args: emitter.print(*args)
     op_name, td_def_name = operator.get_mlir_names()
+
+    # Generate unique result names for ops with nameless results
+    multiple_results = len(operator.returns) > 1
+    generic_result_name = lambda i: "result" + (str(i) if multiple_results else "")
+
     p(f"def {td_def_name} : Torch_Op<{emitter.quote(op_name)}, [")
     with emitter.indent():
         with emitter.indent():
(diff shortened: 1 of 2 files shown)


Shukla-Gaurav (approved):
Thanks for fixing this :) LGTM!

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -297,6 +297,11 @@ def raw_emit_op(operator: JitOperator, f: TextIO, *, traits: List[str],
     emitter = TextEmitter(f)
     p = lambda *args: emitter.print(*args)
     op_name, td_def_name = operator.get_mlir_names()
+
+    # Generate unique names for ops that return multiple results
+    multiple_results = len(operator.returns) > 1
+    result_name = lambda i: "result" + (str(i) if multiple_results else "")
```
silvasean, line 303: nit: aten::nll_loss_forward has actual names. I think we should only use this fallback if the name is `''`.

silvasean (approved):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
ramiro050: Yeah, if the op does name the results, the `or` operation takes care of this

https://github.com/llvm/torch-mlir/blob/4fbc79d9b9370c54056d3517be62b328a067f370/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py#L329

ramiro050: I can update the name of `result_name` to `generic_result_name` to make things more clear

ramiro050: I've updated the comment to be a bit more detailed and renamed `result_name` to `generic_result_name` to make intent clearer

PR #483, opened 2021-12-15
[TORCH][MLIR] Add E2E support for `aten.eq` and `aten.lt` ops

Shukla-Gaurav:
- Added E2E support for `aten.eq.Tensor` and `aten.lt.Tensor` ops. Both
  the operands are expected to be of the same type, i.e., type promotion
  is not addressed as a part of this commit.
- Added E2E support for `aten.eq.Scalar` and `aten.lt.Scalar` ops.
  Tensor operand type to Scalar operand type promotion has not been
  handled in this commit.


e2e_testing/torchscript/elementwise.py
```diff
@@ -419,6 +419,194 @@ def ElementwiseGtIntTensorModule_basic(module, tu: TestUtils):
...
+    def forward(self, x, y):
+        return torch.eq(x, y)
+
+
+@register_test_case(module_factory=lambda: ElementwiseEqFloatTensorModule())
+def ElementwiseEqFloatTensorModule_basic(module, tu: TestUtils):
+    module.forward(torch.tensor([[1.0, 2.2, 6.0], [6.0, 2.0, 3.1]])
+                   .to(torch.float32), 
+                   torch.tensor([1.0, 2.4, 6.0]).to(torch.float32))
+
```
pashu123, line 589: Why aren't you passing random tensors?

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1710,6 +1710,54 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     }
     gtTensor.emitError("unimplemented: dtype isn't supported.");
   }
+  if (auto eqTensor = dyn_cast<AtenEqTensorOp>(op)) {
```
pashu123, line 1713: I think that `eq`, `gt`, and `lt` share a lot of code and can be refactored into one function. Maybe that can be added as a TODO.

e2e_testing/torchscript/elementwise.py
Shukla-Gaurav: There is a high possibility that two random floating-point tensors won't collide, thus producing always false result. So it's better to have unit test cases having same values at some indices.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1710,6 +1710,54 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    AtenEqTensorOp::Adaptor adaptor(operands);
+    Type lhsDtype = payloadArgs[0].getType();
+    Type rhsDtype = payloadArgs[1].getType();
+
+    // TODO: Type promotion in case of different `lhsDtype` and `rhsDtype` needs
+    // to be handled.
+    if (lhsDtype != rhsDtype)
+      eqTensor.emitError("unimplemented: lhs and rhs dtype must be same");
```
ramiro050, line 1721: After emitting an error, make sure you return `nullptr` to avoid executing the rest of the function. The same goes for the other `emitError`s below. Also, I think `gtTensor` above forgets to do this too. It would be nice if you could add return statements to those too

ramiro050 (approved):
LGTM! Thanks for the really comprehensive set of unit tests! Just a small change request below. Feel free to merge after that change

PR #484, opened 2021-12-15
[TORCH]MLIR] Fix C++17 extension warning

Shukla-Gaurav:
This is to address C++17 extension requirement warning added by #474. 

The existing implementation of `ConvertConstantTensorAllocOp<>` requires
a C++17 feature `if constexpr ()`. This commit removes the use of that
feature to support the implementation even for lower C++ versions.


ramiro050 (approved):
LGTM! The `empty` op and `ones` and `zeros` do share quite a bit of logic, but this is fine for now. Maybe in the future we can decompose `ones` and `zeros` into `empty` + `fill_Scalar` ops

PR #487, opened 2021-12-15
[tosa] Support for Torch.squeeze

sjarus:

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -526,6 +526,41 @@ LogicalResult ConvertAtenOp<AtenArgmaxOp>::matchAndRewrite(
   return success();
 }
 
+template <>
+LogicalResult ConvertAtenOp<AtenSqueezeOp>::matchAndRewrite(
+    AtenSqueezeOp op, OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+
+  Value self = adaptor.self();
+  auto selfTy = self.getType().template cast<RankedTensorType>();
+
+  if (!selfTy)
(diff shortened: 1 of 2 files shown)


sjarus:
Torch-TOSA functional unit tests also pass: 

Results PASS test_squeeze_1_f32
Results PASS test_squeeze_1x23_f32
Results PASS test_squeeze_1x4x8x29_f32
Results PASS test_squeeze_1x32x1x32_f32

sjarus:
The static e2e tests now work. The dynamic ones won't until TOSA to LinAlg pass is updated to support it - something that also touches all the reduction ops and argmax.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -526,6 +526,41 @@ LogicalResult ConvertAtenOp<AtenArgmaxOp>::matchAndRewrite(
...
+  Value self = adaptor.self();
+  auto selfTy = self.getType().template cast<RankedTensorType>();
+
+  if (!selfTy)
+    return op.emitError("Only ranked tensor types supported in TOSA argmax");
+
+  auto selfShape = selfTy.getShape();
+
+  SmallVector<int64_t> newOutputShape;
+  for (auto &dim : selfShape) {
```
silvasean, line 543: Maybe add a comment here about the assumptions w.r.t. to dimensions dynamically equal to 1 where it is not known statically.

silvasean (approved):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
sjarus: Will do. Am planning to refactor this to also support squeeze.dim so next version will also contain your suggestion. May have unsqueeze.dim as well at the time.

PR #488, opened 2021-12-16
Add e2e lowering for aten.index.put op.

pashu123:
This commit adds e2e lowering for `aten.index.put` op.
Also, fixes a bug in reduce_op_variant to check for optional types
which may contain non-value type tensors.

### DIFF
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -3723,6 +3723,142 @@ public:
 };
 } // namespace
 
+// Collapses a `tensor` of rank k into a `tensor` of rank 1.
+// Special case 0d tensor, in that case it's an expansion.
+static Value collapseTo1dTensor(OpBuilder &b, Location loc, Value tensor) {
+  auto tensorType = tensor.getType().cast<RankedTensorType>();
+  if (tensorType.getRank() == 0) {
+    Value one = b.create<arith::ConstantIndexOp>(loc, 1);
+    Value scalarVal = b.create<tensor::ExtractOp>(loc, tensor);
+    return createInitTensor(b, loc, {one}, tensorType.getElementType(),
+                            scalarVal);
+  }
+  if (tensorType.getRank() == 1)
+    return tensor;
+  SmallVector<ReassociationIndices> reassociation(1);
+  for (auto i : llvm::seq<int64_t>(0, tensorType.getRank()))
+    reassociation[0].push_back(i);
+  return b.create<linalg::TensorCollapseShapeOp>(loc, tensor, reassociation);
+}
+
+// Given `input`, `indices`, and `values` `index_put` is given by:
+// `input` is an ND tensor, `indices` is a list of tensors, `value` is an ND
+// tensor.
+// For ex: input is a 2d tensor, indices is a list containing two 1d tensors.
+//   for i in range(0, len(indices[0]))):
+//     for j in range(0, len(input.shape[0]))
+//      for k in range(0, len(input.shape[1]))
+//         if (indices[0][i] == j && indices[1][i] == k)
+//            input[j][k] = values[i];
+namespace {
+class ConvertAtenIndexPutOp : public OpConversionPattern<AtenIndexPutOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenIndexPutOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    Location loc = op.getLoc();
+    Value input = adaptor.self();
+    Value indices = op.indices();
+    Value elemVal = adaptor.values();
+    auto inputType = input.getType().cast<RankedTensorType>();
+    unsigned inputRank = inputType.getRank();
+
+    SmallVector<Value> indicesTuple;
+    if (!getListConstructElements(indices, indicesTuple)) {
+      return rewriter.notifyMatchFailure(op,
+                                         "unimplemented: indices argument is "
+                                         "not constructed from ListConstruct.");
+    }
+    unsigned numOfIndices = indicesTuple.size();
+    SmallVector<Value> indicesVal =
+        getTypeConvertedValues(rewriter, loc, getTypeConverter(), indicesTuple);
+
+    // TODO: Support broadcasting along dim.
+    if (inputRank != numOfIndices)
+      return rewriter.notifyMatchFailure(
+          op, "invalid shape: Broadcasting value along dim is not supported");
+
+    // TODO: More comprehensive checks:-
+    //   1. Since broadcasting is not implemented the number of elements in each
+    //   indexes as well as the output tensor should be same.
+    //
+    //   2. Since broadcasting is not supported and no. of elements should be
+    //   same hence the shape of tensors should be equivalent, for example:
+    //   [x,y] and [x,y,1] or [x,y,1,1] is considered equivalent but [y,x] and
+    //   [x,y] are not considered equivalent.
+    //
+    //  3. There should be an assert that the index_values lies in the bounds of
+    //  the input tensor.
+
+    // The value tensor along with all the indices are collapsed to 1d tensor.
+    SmallVector<Value> collapsedIndices;
+    for (Value val : indicesVal)
+      collapsedIndices.push_back(collapseTo1dTensor(rewriter, loc, val));
+
+    elemVal = collapseTo1dTensor(rewriter, loc, elemVal);
+
+    AffineExpr zeroIdxExpr = rewriter.getAffineDimExpr(0);
+    // `indicesExpr` points to the mapping {d0, d1...} -> {d0} i.e., the
+    // outermost dimension.
+    SmallVector<AffineExpr, 4> indicesExpr{zeroIdxExpr};
+    SmallVector<AffineExpr, 4> inputExpr;
+    for (unsigned i = 0; i < inputRank; i++)
+      inputExpr.push_back(rewriter.getAffineDimExpr(i + 1));
+
+    SmallVector<ArrayRef<AffineExpr>> indexExpr;
+    for (unsigned i = 0; i < numOfIndices; i++)
+      indexExpr.push_back(indicesExpr);
+
+    // The arguments to the linalg.generic are: inputs: (indices1, indices2 ..
+    // indicesN, input, elemVal), output = (input).
+    indexExpr.insert(indexExpr.end(),
+                     {inputExpr /* input */, indicesExpr /* elemVal */,
+                      /* input */ inputExpr});
+
+    auto indexingMaps = AffineMap::inferFromExprList(indexExpr);
+    SmallVector<StringRef> iteratorTypes(inputRank + 1, "parallel");
+
+    SmallVector<Value> genericInputs(collapsedIndices.begin(),
+                                     collapsedIndices.end());
+    genericInputs.insert(genericInputs.end(), {input, elemVal});
+
+    Value updatedInput =
+        rewriter
+            .create<linalg::GenericOp>(
+                loc, input.getType(), genericInputs, input,
+                /*indexingMaps=*/indexingMaps,
+                /*iteratorTypes=*/iteratorTypes,
+                [&](OpBuilder &b, Location loc, ValueRange args) {
+                  Value truePredicate = rewriter.create<arith::ConstantOp>(
+                      loc, rewriter.getI1Type(), rewriter.getBoolAttr(true));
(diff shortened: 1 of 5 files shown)


pashu123:
This will need https://reviews.llvm.org/D115878 to get through.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -39,7 +39,8 @@ class ConvertToImmutableTensors : public RewritePattern {
         opOperand.set(rewriter.create<CopyToValueTensorOp>(op->getLoc(),
                                                            opOperand.get()));
       } else if (auto listType = operandType.dyn_cast<ListType>()) {
-        if (!listType.getContainedType().isa<NonValueTensorType>())
+        if (!(listType.getContainedType().isa<NonValueTensorType>() ||
+              listType.getContainedType().isa<OptionalType>()))
```
cathyzhyi, line 43: need to check the OptionalType is a `NonValueTensorType ` optional type.

cathyzhyi (commented):
I think this would generate illegal instruction

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3435,6 +3435,132 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
...
+    SmallVector<Value> collapsedIndices;
+    for (Value val : indicesVal)
+      collapsedIndices.push_back(collapseTo1dTensor(rewriter, loc, val));
+
+    elemVal = collapseTo1dTensor(rewriter, loc, elemVal);
+
+    AffineExpr zeroIdxExpr = rewriter.getAffineDimExpr(0);
+    // `indicesExpr` points to the mapping {d0,d1...} -> {d0} i.e., the
+    // outermost dimension.
+    SmallVector<AffineExpr, 4> indicesExpr{zeroIdxExpr};
```
ramiro050, line 3510: nit: no need to specify size of `SmallVector`. The same goes for all the other `SmallVector`s below

> In the absence of a well-motivated choice for the number of inlined elements N, it is recommended to use SmallVector<T> (that is, omitting the N).

https://www.llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h

ramiro050 (changes requested):
Not done reviewing. But could you mention in this PR a pseudo-code version of what your algorithm is doing? This will allow me to review a bit faster

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3723,6 +3723,132 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
...
+    Value scalarVal = b.create<tensor::ExtractOp>(loc, tensor);
+    return createInitTensor(b, loc, {one}, tensorType.getElementType(),
+                            scalarVal);
+  }
+  if (tensorType.getRank() == 1)
+    return tensor;
+  SmallVector<ReassociationIndices> reassociation(1);
+  for (auto i : llvm::seq<int64_t>(0, tensorType.getRank()))
+    reassociation[0].push_back(i);
+  return b.create<linalg::TensorCollapseShapeOp>(loc, tensor, reassociation);
```
gprateek93, line 3741: Please wait for Garvey's PR to merge. You will have to change it to `<TensorCollapeShapeOp>` instead.

pashu123: I see, thanks.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
pashu123: @cathyzhyi I have added a check to see that all tensors within optionalListType are non-value tensor types. Optional case handling is added as a TODO for now.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
pashu123: ```../external/llvm-project/llvm/../mlir/include/mlir/IR/AffineMap.h:90:3: note: candidate function not viable: no known conversion from 'SmallVector<SmallVector<mlir::AffineExpr> >' to 'ArrayRef<SmallVector<mlir::AffineExpr, 4> >' for 1st argument```

pashu123: The function argument is expecting SmallVector<mlir::AffineExpr,4> instead of SmallVectorImpl.

ramiro050: If you change the type of `indexExpr` here

https://github.com/llvm/torch-mlir/blob/b51330fb5f87241057a3e3356b028b1ebcd601c8/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp#L3813

to `SmallVector<ArrayRef<AffineExpr>>` it should work.

Conversion from `SmallVector` to `ArrayRef` happens automatically, however it does not happen recursively, so after this vector was converted, the result was an `ArrayRef<SmallVector<AffineExpr>>` rather than an `ArrayRef<ArrayRef<AffineExpr>>`, as expected by the AffineMap` function.

pashu123: @ramiro050 Thanks for the fix and explanation!

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3723,6 +3723,142 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
...
+    //   [x,y] and [x,y,1] or [x,y,1,1] is considered equivalent but [y,x] and
+    //   [x,y] are not considered equivalent.
+    //
+    //  3. There should be an assert that the index_values lies in the bounds of
+    //  the input tensor.
+
+    // The value tensor along with all the indices are collapsed to 1d tensor.
+    SmallVector<Value> collapsedIndices;
+    for (Value val : indicesVal)
+      collapsedIndices.push_back(collapseTo1dTensor(rewriter, loc, val));
```
ramiro050, line 3801: Also, I'm not sure this method of collapsing will generalize well in the case where the indices tensors don't have the same number of elements. You can achieve a rank broadcasting with the `linalg.generic` maps. For example, suppose we have the following tensors:

```
a: tensor<4, 7>
i: tensor<2, 3>
j: tensor<4, 2, 3>
new_vals: tensor<3>
```

Then you can have the following maps:

```
a: (a, b, c, d, e) -> (a, b)
i: (a, b, c, d, e) -> (d, e)
j: (a, b, c, d, e) -> (c, d, e)
new_vals: (a, b, c, d, e) -> (e)
```

In other words, the domain of the maps have `a.rank + max_rank(i, j, new_vals)` dimensions. The input tensor maps to the first set of dimensions, and all other tensors map to the rest of the dimensions.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3723,6 +3723,142 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
...
+      indexExpr.push_back(indicesExpr);
+
+    // The arguments to the linalg.generic are: inputs: (indices1, indices2 ..
+    // indicesN, input, elemVal), output = (input).
+    indexExpr.insert(indexExpr.end(),
+                     {inputExpr /* input */, indicesExpr /* elemVal */,
+                      /* input */ inputExpr});
+
+    auto indexingMaps = AffineMap::inferFromExprList(indexExpr);
+    SmallVector<StringRef> iteratorTypes(inputRank + 1, "parallel");
```
ramiro050, line 3824: nit: use the function version of iterator type: `getParallelIteratorTypeName()`

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -39,68 +39,81 @@ class ConvertToImmutableTensors : public RewritePattern {
         opOperand.set(rewriter.create<CopyToValueTensorOp>(op->getLoc(),
                                                            opOperand.get()));
       } else if (auto listType = operandType.dyn_cast<ListType>()) {
-        if (!listType.getContainedType().isa<NonValueTensorType>())
```
ramiro050, line 42: Can you fix the formatting in this file? It's hard to tell which changes are new and which are not

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3723,6 +3723,142 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
...
+
+// Given `input`, `indices`, and `values` `index_put` is given by:
+// `input` is an ND tensor, `indices` is a list of tensors, `value` is an ND
+// tensor.
+// For ex: input is a 2d tensor, indices is a list containing two 1d tensors.
+//   for i in range(0, len(indices[0]))):
+//     for j in range(0, len(input.shape[0]))
+//      for k in range(0, len(input.shape[1]))
+//         if (indices[0][i] == j && indices[1][i] == k)
+//            input[j][k] = values[i];
```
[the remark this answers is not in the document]
cathyzhyi: yea, we just had a discussion about this in the channel. However, the current proposed approach would involves adding a bunch of new ops into `linalg_ext` which I suspect would not be scalable if we have a lot of such ops. Will need input from @silvasean about this.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3723,6 +3723,142 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
...
+    Value scalarVal = b.create<tensor::ExtractOp>(loc, tensor);
+    return createInitTensor(b, loc, {one}, tensorType.getElementType(),
+                            scalarVal);
+  }
+  if (tensorType.getRank() == 1)
+    return tensor;
+  SmallVector<ReassociationIndices> reassociation(1);
+  for (auto i : llvm::seq<int64_t>(0, tensorType.getRank()))
+    reassociation[0].push_back(i);
+  return b.create<linalg::TensorCollapseShapeOp>(loc, tensor, reassociation);
```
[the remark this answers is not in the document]
cathyzhyi: It's probably ok as it is because `TensorCollapseShapeOp` has the `::mlir::OpTrait::OneResult` trait.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -39,68 +39,81 @@ class ConvertToImmutableTensors : public RewritePattern {
...
-      } else if (auto optionalType = operandType.dyn_cast<OptionalType>()) {
-        // TODO: A more general way to handle the optional type is to
-        // introduce a `copy.to_optional_vtensor` op.
-        if (!optionalType.getContainedType().isa<NonValueTensorType>())
-          continue;
+          if (listConstruct.elements().empty())
+            continue;
+
+          // TODO: Handle optional type in list type.
+          if (listType.getContainedType().isa<OptionalType>())
```
cathyzhyi, line 61: nit: need {} for the body of this `if`

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
pashu123: I have added Broadcasting as a TODO for now.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3723,6 +3723,142 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
...
+// `input` is an ND tensor, `indices` is a list of tensors, `value` is an ND
+// tensor.
+// For ex: input is a 2d tensor, indices is a list containing two 1d tensors.
+//   for i in range(0, len(indices[0]))):
+//     for j in range(0, len(input.shape[0]))
+//      for k in range(0, len(input.shape[1]))
+//         if (indices[0][i] == j && indices[1][i] == k)
+//            input[j][k] = values[i];
+namespace {
+class ConvertAtenIndexPutOp : public OpConversionPattern<AtenIndexPutOp> {
```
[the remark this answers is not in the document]
pashu123: Yes sure.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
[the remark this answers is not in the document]
pashu123: This op can be better handled in the `scf` dialect.

PR #492, opened 2021-12-17
Add aten::t op

nirvedhmeshram:


powderluv:
@silvasean  can he please get CI permissions

pashu123:
@nirvedhmeshram Please clang-format.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1240,6 +1245,26 @@ ChangeResult TypeAnalyzer::visitAtenTransposeIntOp(
...
 
+ChangeResult TypeAnalyzer::visitAtenTOp(
+	AtenTOp op,
+	ArrayRef<LatticeElement<ValueKnowledge> *> operands) {
+	auto input = operands[0]->getValue();
+	auto knowledge =
+		ValueKnowledge::getNotNonePessimisticValueState(op.getContext());
+	knowledge.dtype = input.dtype;
+	knowledge.hasSizes = input.hasSizes;
+	knowledge.sizes = input.sizes;
```
pashu123, line 1256: Proceed only if `input.hasSizes`.

e2e_testing/torchscript/basic.py
```diff
@@ -1179,3 +1179,21 @@ def forward(self, a):
...
+
+# ==============================================================================
+class TModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([3, 4], torch.float32, True),
```
pashu123, line 1191: Check for dynamic dimension. [-1, -1].

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1240,6 +1245,26 @@ ChangeResult TypeAnalyzer::visitAtenTransposeIntOp(
...
+	int64_t inputRank = input.sizes.size();
+	if (inputRank>2) {
+		std::string error_message = "t() expects a tensor with <=2 dimensions, but self is " + std::to_string(inputRank) + "D";
+		llvm::report_fatal_error(error_message.c_str());
+	}
+	if (inputRank == 2) {
+		std::swap(knowledge.sizes[0], knowledge.sizes[1]);
```
pashu123, line 1263: This will give an error if dynamic dimensions are passed.

pashu123 (changes requested):
Nice work! The logic is fine. Just some refinements are needed.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -379,6 +379,31 @@ class DecomposeAtenMatmulOp : public OpRewritePattern<AtenMatmulOp> {
...
+                                PatternRewriter &rewriter) const override {
+    Value lhs = op.self();
+    int lhsRank = getTensorRank(lhs);
+    auto loc = op.getLoc();
+
+    // If both lhs and rhs ranks are 2 then map it to `aten.mm` op.
+    Value zero = rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(0));
+    if (lhsRank < 2) {
+      rewriter.replaceOpWithNewOp<AtenTransposeIntOp>(op, op.getType(), lhs, zero,zero);
+    }
```
Shukla-Gaurav, line 396: Since this is identity case, you can simply replace the op with its operand. 
`replaceOp(op, lhs)`.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1240,6 +1245,26 @@ ChangeResult TypeAnalyzer::visitAtenTransposeIntOp(
...
+	auto knowledge =
+		ValueKnowledge::getNotNonePessimisticValueState(op.getContext());
+	knowledge.dtype = input.dtype;
+	knowledge.hasSizes = input.hasSizes;
+	knowledge.sizes = input.sizes;
+	int64_t inputRank = input.sizes.size();
+	if (inputRank>2) {
+		std::string error_message = "t() expects a tensor with <=2 dimensions, but self is " + std::to_string(inputRank) + "D";
+		llvm::report_fatal_error(error_message.c_str());
+	}
```
Shukla-Gaurav, line 1261: I usually use `assert` in such cases as the expected operand must be less than or equal to rank 2.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
Shukla-Gaurav: Also, you can remove trivial braces here.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -379,6 +379,31 @@ class DecomposeAtenMatmulOp : public OpRewritePattern<AtenMatmulOp> {
...
 } // namespace
 
+namespace {
+class DecomposeAtenTOp : public OpRewritePattern<AtenTOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenTOp op,
+                                PatternRewriter &rewriter) const override {
+    Value lhs = op.self();
+    int lhsRank = getTensorRank(lhs);
```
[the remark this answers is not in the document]
ramiro050: `getTensorRank` returns an int because it uses -1 as an error value. I think it's fine to leave the int, but it would be good to add a check making sure it's non-negative

ramiro050:
Note: you can run `clang-format` on your most recent commit by doing `git clang-format HEAD~1`. This will make some new changes than you can then commit and squash

e2e_testing/torchscript/basic.py
nirvedhmeshram: Is it okay to do all three rank classes in dynamic or do I need 6 classes 3 for static and three for dynamic?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
nirvedhmeshram: Ya it passes fine in dynamic as well. Do I need to change something here?

ramiro050: It looks fine to me. if the input rank is 2, then there has to be enough elements in the array to perform the swap without any issues.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
nirvedhmeshram: I think assert will not produce an error message right? This error message is what native torch gives that's why I included it. Should I include the assert after the if statement? But there it may be redundant.

nirvedhmeshram: oh I see, I could do assert(inputRank<=2 && error_message.c_str())
Will this work?

e2e_testing/torchscript/basic.py
ramiro050: You can do all dynamic

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -379,6 +379,36 @@ class DecomposeAtenMatmulOp : public OpRewritePattern<AtenMatmulOp> {
...
+    auto loc = op.getLoc();
+
+    if (lhsRank < 2)
+      rewriter.replaceOp(op, lhs);
+    else if (lhsRank > 2 || lhsRank < 0) {
+      std::string errorMessage =
+          "t() expects a tensor with <=2 dimensions, but self is " +
+          std::to_string(lhsRank) + "D";
+      op.emitError(errorMessage.c_str());
```
silvasean, line 398: nit: patterns shouldn't emit errors. You can just `return rewriter.notifyMatchFailure(...)`

e2e_testing/torchscript/basic.py
```diff
@@ -1179,3 +1179,56 @@ def forward(self, a):
...
+        ([-1], torch.float32, True),
+    ])
+    def forward(self, lhs):
+        return torch.t(lhs)
+
+
+@register_test_case(module_factory=lambda: TModuleRank0())
+def TModuleRank0_basic(module, tu: TestUtils):
+    module.forward(tu.rand(1))
+
```
nirvedhmeshram, line 1234: Any idea why TModuleRank1_basic and TModuleRank0_basic pass unexpectedly?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -379,6 +379,36 @@ class DecomposeAtenMatmulOp : public OpRewritePattern<AtenMatmulOp> {
...
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenTOp op,
+                                PatternRewriter &rewriter) const override {
+    Value lhs = op.self();
+    int lhsRank = getTensorRank(lhs);
+    auto loc = op.getLoc();
+
+    if (lhsRank < 2)
+      rewriter.replaceOp(op, lhs);
+    else if (lhsRank > 2 || lhsRank < 0) {
```
ramiro050, line 394: the `lhsRank < 0` case will never reach here because it will go into the first `if` branch above. You should check that the rank is valid right after `getTensorRank` before doing anything else

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1240,6 +1244,26 @@ ChangeResult TypeAnalyzer::visitAtenTransposeIntOp(
   return getLatticeElement(op.getResult()).join(knowledge);
 }
 
+ChangeResult TypeAnalyzer::visitAtenTOp(
+    AtenTOp op, ArrayRef<LatticeElement<ValueKnowledge> *> operands) {
+  auto input = operands[0]->getValue();
+  auto knowledge =
+      ValueKnowledge::getNotNonePessimisticValueState(op.getContext());
+  knowledge.dtype = input.dtype;
```
silvasean, line 1252: don't use `assert` here. This function just computes the knowledge that can be inferred. It does not report errors or abort. If no knowledge can be inferred, then it should just join the pessimistic value knowledge.

silvasean (changes requested):

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/RefineTypes.cpp
nirvedhmeshram: So should I not have any checks for valid rank here? earlier I had if statements that would raise llvm::fatal_error if the rank was invalid. However, I do have checks in the decomposition pass so might not need to do any checks here?

e2e_testing/torchscript/basic.py
```diff
@@ -1179,3 +1179,56 @@ def forward(self, a):
...
+    module.forward(tu.rand(3))
+
+class TModuleRank0(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1], torch.float32, True),
```
ramiro050, line 1225: This isn't zero rank. To represent a zero rank tensor, you just do `[]`. Also, `tu.rand(1)` is not right either because this returns a rank-1 tensor. You should just create a tensor like `torch.tensor(yourFavoriteFloatValue, dtype=torch.float32)`.

ramiro050 (changes requested):

e2e_testing/torchscript/basic.py
ramiro050: New tests are set as "expected to fail" by default on the TOSA side. To add it to the "expected to pass" set, add your test to this list

https://github.com/sjarus/torch-mlir/blob/4bb8668f8db1fd2829ba9734381c7ecbf540eabe/e2e_testing/torchscript/xfail_sets.py#L24

The reason those two are passing on the TOSA side despite TOSA not yet having support for the transpose op is because for those two cases, the transpose op gets removed during decomposition, which happens before the `TorchToTosa` pass takes place.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1240,6 +1244,23 @@ ChangeResult TypeAnalyzer::visitAtenTransposeIntOp(
...
+  if (!input.hasSizes)
+    return getLatticeElement(op.getResult()).join(knowledge);
+  knowledge.hasSizes = input.hasSizes;
+  knowledge.sizes = input.sizes;
+  int64_t inputRank = input.sizes.size();
+  if (inputRank == 2) {
```
silvasean, line 1258: if input rank != 2, then I think we need to return a pessimistic result?

nirvedhmeshram: Could you please elaborate why? We check on line 1253 that it has sizes then why not pass them on so that we can use them for various cases in the decomposition pass, such as do nothing for rank 0 & 1, transpose for rank=2 and error for all others.

e2e_testing/torchscript/basic.py
```diff
@@ -1179,3 +1179,56 @@ def forward(self, a):
...
+        None,
+        ([], torch.float32, True),
+    ])
+    def forward(self, lhs):
+        return torch.t(lhs)
+
+
+@register_test_case(module_factory=lambda: TModuleRank0())
+def TModuleRank0_basic(module, tu: TestUtils):
+    module.forward(torch.tensor(7,dtype=torch.float32))
```
ramiro050, line 1233: nit: there should be a space after the comma

lib/Dialect/Torch/Transforms/RefineTypes.cpp
ramiro050: I think what @silvasean meant was that if the input is invalid, you should return a pessimistic result. You're correct that rank 0,1, and 2 are valid inputs, but anything else should result in a pessimistic result. 

The reason for this is that the functions in `RefineTypes` are expected to always return knowledge that is valid, even if it is not the most exact. For example, suppose we have an op that is just the identity operation on a tensor. If we have as an input a `!torch.tensor<[1,2,3],f32>`, refine types is allowed to return `!torch.tensor<[1,2,3],f32>`, `!torch.tensor<[?,?,?],f32>` (i.e. an infinite set of float rank-3 tensors because the expected output is part of this set), or any other set of values that contains the correct result type in it.

In the case of `aten.t`, currently if the op gets as an input `!torch.tensor<[1,2,3],f32>`, it will return that a valid output is `!torch.tensor<[1,2,3],f32>`, which is not true for this particular op. Such an input is invalid, so the correct knowledge to return is "I don't know what the output is" (the pessimistic result). 

See this comment for more info on the invariant that must be preserved by the `RefineTypes` functions:

https://github.com/llvm/torch-mlir/blob/5eed562e19916ba178f5e8836898ef2d8615e1aa/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L117-L118

ramiro050 (changes requested):
Things are looking good! Just a couple of comments below:

ramiro050 (approved):
LGTM!

PR #495, opened 2021-12-17
[MLIR][TORCH] Add E2E support for aten.index.Tensor op

vivekkhandelwal1:
This commit adds lowering of `aten.index.Tensor` op

### DIFF
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -4224,6 +4224,83 @@ public:
 };
 } // namespace
 
+namespace {
+class ConvertAtenIndexTensorOp : public OpConversionPattern<AtenIndexTensorOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenIndexTensorOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    Location loc = op.getLoc();
+    Value input = adaptor.self();
+    Value indices = op.indices();
+    SmallVector<Value> indicesTuple;
+    if (!getListConstructElements(indices, indicesTuple)) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: the indices list is not from a list construct");
+    }
+
+    SmallVector<Value> indicesVal =
+        getTypeConvertedValues(rewriter, loc, getTypeConverter(), indicesTuple);
+
+    RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+    RankedTensorType resultType = getTypeConverter()
+                                      ->convertType(op->getResult(0).getType())
+                                      .cast<RankedTensorType>();
+    Type elementType = resultType.getElementType();
+    unsigned inputRank = inputType.getRank();
+    unsigned numIndexTensors = indicesTuple.size();
+    SmallVector<Value> inputShape = getTensorSizes(rewriter, loc, input);
+
+    // Case 1 : When numIndexTensors == 1 and `input` is a 1-d tensor.
+    // TODO: generalize the implementation for other cases.
+    if (numIndexTensors == 1 && inputRank == 1) {
+      if (failed(checkNotNone(rewriter, op, indicesVal[0])))
+        return rewriter.notifyMatchFailure(op, "unimplemented None type arg");
+      unsigned resultRank =
+          indicesVal[0].getType().cast<RankedTensorType>().getRank();
+      SmallVector<Value> resultShape;
+      SmallVector<AffineExpr> indicesExpr, resultExpr;
+      SmallVector<StringRef> iteratorTypes;
+      for (unsigned i = 0; i < resultRank; i++)
+        resultShape.push_back(getDimOp(rewriter, loc, indicesVal[0], i));
+      Value initTensor =
+          rewriter.create<linalg::InitTensorOp>(loc, resultShape, elementType);
+      for (unsigned i = 0; i < resultRank; i++) {
+        indicesExpr.push_back(rewriter.getAffineDimExpr(i));
+        resultExpr.push_back(rewriter.getAffineDimExpr(i));
+        iteratorTypes.push_back(getParallelIteratorTypeName());
+      }
+      auto indexingMaps =
+          AffineMap::inferFromExprList({indicesExpr, resultExpr});
+
+      Value finalRes =
+          rewriter
+              .create<linalg::GenericOp>(
+                  loc, initTensor.getType(), ValueRange{indicesVal[0]},
+                  initTensor,
+                  /*indexingMaps=*/indexingMaps,
+                  /*iteratorTypes=*/iteratorTypes,
+                  [&](OpBuilder &b, Location loc, ValueRange args) {
+                    Value indexTarget = castIntToIndex(b, loc, args[0]);
+                    Value extractedElement =
+                        b.create<tensor::ExtractOp>(loc, input, indexTarget);
+                    b.create<linalg::YieldOp>(loc, extractedElement);
+                  })
+              .getResult(0);
+
+      rewriter.replaceOpWithNewOp<tensor::CastOp>(op, resultType, finalRes);
+      return success();
+    } else
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: support for this set of inputs not present");
+  }
+};
+} // namespace
+
 // -----------------------------------------------------------------------------
 // The pass
 // -----------------------------------------------------------------------------
@@ -4341,6 +4418,8 @@ public:
     target.addIllegalOp<AtenTensorIntOp, AtenTensorFloatOp>();
     patterns.add<ConvertAtenArangeStartStepOp>(typeConverter, context);
     target.addIllegalOp<AtenArangeStartStepOp>();
+    patterns.add<ConvertAtenIndexTensorOp>(typeConverter, context);
+    target.addIllegalOp<AtenIndexTensorOp>();
 
     if (failed(applyPartialConversion(getOperation(), target,
                                       std::move(patterns))))
--- a/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
+++ b/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
@@ -39,22 +39,34 @@ public:
         opOperand.set(rewriter.create<CopyToValueTensorOp>(op->getLoc(),
                                                            opOperand.get()));
       } else if (auto listType = operandType.dyn_cast<ListType>()) {
-        if (!listType.getContainedType().isa<NonValueTensorType>())
+        if (!(listType.getContainedType().isa<NonValueTensorType>() ||
+              listType.getContainedType().isa<OptionalType>()))
           continue;
 
         // Construct a new list whose elements are value tensors copied from
-        // the none value tensors of the original list.
+        // the non-value tensors of the original list.
         auto listConstruct =
             opOperand.get().getDefiningOp<PrimListConstructOp>();
         if (!listConstruct) {
           rewriter.cancelRootUpdate(op);
-          return rewriter.notifyMatchFailure(op, 
-              "unimplemented: list of non vtensor type not constructed "
-              "from list construct");
+          return rewriter.notifyMatchFailure(
+              op, "unimplemented: list of non vtensor type not constructed "
+                  "from list construct");
         }
 
         if (listConstruct.elements().empty())
           continue;
+
+        // TODO: Handle optional type in list type.
+        if (listType.getContainedType().isa<OptionalType>()) {
+          if (!llvm::all_of(listConstruct.elements(), [](Value val) {
+                return val.getType().isa<NonValueTensorType>();
+              }))
+            return rewriter.notifyMatchFailure(
+                op, "unimplemented: list containing optional type is not "
+                    "handled.");
+        }
+
         auto newListElements = llvm::to_vector<4>(llvm::map_range(
             listConstruct.elements(), [&](Value tensor) -> Value {
               return rewriter.create<CopyToValueTensorOp>(op->getLoc(), tensor);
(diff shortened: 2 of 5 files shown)


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -293,7 +293,7 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
           ValueKnowledge::getNotNonePessimisticValueState(op->getContext());
       if (input.hasSizes) {
         knowledge.hasSizes = true;
-        knowledge.sizes.resize(input.sizes.size(), kUnknownSize);
```
ramiro050, line 296: Does using `input.sizes.size()` not work?

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3985,6 +3985,79 @@ class ConvertAtenIndexSelectOp : public OpConversionPattern<AtenIndexSelectOp> {
...
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    Location loc = op.getLoc();
+    Value input = adaptor.self();
+    Value indices = op.indices();
+    SmallVector<Value> indicesTuple;
+    if (!getListConstructElements(indices, indicesTuple)) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: the indices are not present.");
```
ramiro050, line 4004: nit: Here there can be an error even if the indices are present, for they could be constructed using something other than a PrimListConstruct. I would rewrite the error message as:

```
unimplemented: the indices list is not from a list construct
```

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3985,6 +3985,79 @@ class ConvertAtenIndexSelectOp : public OpConversionPattern<AtenIndexSelectOp> {
...
+                                      ->convertType(op->getResult(0).getType())
+                                      .cast<RankedTensorType>();
+    Type elementType = resultType.getElementType();
+    unsigned inputRank = inputType.getRank();
+    unsigned numIndexTensors = indicesTuple.size();
+    SmallVector<Value> inputShape = getTensorSizes(rewriter, loc, input);
+
+    // Case 1 : When numIndexTensors == 1
+    // TODO: generalize the implementation for other cases.
+    if (numIndexTensors == 1 && inputRank == 1) {
```
ramiro050, line 4021: Since there are cases not handled, the `if` should end with an `else` branch that returns a `notifyMatchFailure` that says `unimplemented: ...` and describes the only cases currently supported.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3985,6 +3985,79 @@ class ConvertAtenIndexSelectOp : public OpConversionPattern<AtenIndexSelectOp> {
...
+      unsigned resultRank =
+          indicesVal[0].getType().cast<RankedTensorType>().getRank();
```
ramiro050, line 4023: Since the list of tensors is made up of optional tensors, you have to make sure the tensor is not `None` before using it to get the rank

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -39,7 +39,8 @@ class ConvertToImmutableTensors : public RewritePattern {
         opOperand.set(rewriter.create<CopyToValueTensorOp>(op->getLoc(),
                                                            opOperand.get()));
       } else if (auto listType = operandType.dyn_cast<ListType>()) {
-        if (!listType.getContainedType().isa<NonValueTensorType>())
+        if (!(listType.getContainedType().isa<NonValueTensorType>() ||
+              listType.getContainedType().isa<OptionalType>()))
```
ramiro050, line 43: Make sure to coordinate with @pashu123 on this fix, since we now have two PRs trying to fix this. See also @cathyzhyi's comment: https://github.com/llvm/torch-mlir/pull/488#pullrequestreview-834314301

ramiro050 (changes requested):
Great job getting this to work! A few comments below:

lib/Dialect/Torch/Transforms/RefineTypes.cpp
vivekkhandelwal1: No

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -289,11 +289,30 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
...
     if (auto indexTensor = dyn_cast<AtenIndexTensorOp>(op)) {
       auto input = operands[0]->getValue();
+      auto indicesList = indexTensor.indices();
       auto knowledge =
           ValueKnowledge::getNotNonePessimisticValueState(op->getContext());
+      auto listConstruct = indicesList.getDefiningOp<PrimListConstructOp>();
+      if (!listConstruct)
+        return getLatticeElement(op->getResult(0)).join(knowledge);
+
+      auto indices = llvm::to_vector<4>(llvm::map_range(
```
ramiro050, line 299: nit: since the indices could be a list of any length, we should not hard-code the initial size to 4. You should be able to do `llvm::to_vector(...)`.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -289,11 +289,30 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
...
+          listConstruct.elements(), [&](Value v) -> ValueKnowledge {
+            return getLatticeElement(v).getValue();
+          }));
+
+      for (auto index : indices)
+        knowledge = ValueKnowledge::join(knowledge, index);
```
ramiro050, line 305: What is the reason for joining with every index?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -289,11 +289,30 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
     // input.
     if (auto indexTensor = dyn_cast<AtenIndexTensorOp>(op)) {
```
ramiro050, line 290: nit: because of how complex `AtenIndexTensorOp` is, I think it would be better to extract this logic into a `visit...` function

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -289,11 +289,30 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
...
+
+      auto indices = llvm::to_vector<4>(llvm::map_range(
+          listConstruct.elements(), [&](Value v) -> ValueKnowledge {
+            return getLatticeElement(v).getValue();
+          }));
+
+      for (auto index : indices)
+        knowledge = ValueKnowledge::join(knowledge, index);
+      if (!knowledge.hasSizes)
+        return getLatticeElement(op->getResult(0)).join(knowledge);
```
ramiro050, line 307: I'm not sure if this is the correct result. From `join`ing the knowledge with the knowledge in each index, I would imagine that the `dtype` is probably set to the `dtype` of the indices, which would be the wrong result type if the op is indexing a float tensor.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -289,11 +289,30 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
...
       if (input.hasSizes) {
         knowledge.hasSizes = true;
-        knowledge.sizes.resize(input.sizes.size(), kUnknownSize);
+        // Case: If the input is a 1-d tensor and indices list size is equal
+        // to 1.
+        if (input.sizes.size() == 1 && indices.size() == 1)
+          knowledge.sizes.resize(indices[0].sizes.size(), kUnknownSize);
+        else
+          knowledge.sizes.resize(input.sizes.size(), kUnknownSize);
```
ramiro050, line 315: I don't think this is right. This `else` branch corresponds to the cases currently not being handled. Therefore, the correct knowledge is, "we don't know what the output is", which is the Pessimistic result. In general for this op, the result shape depends on the index tensors, not the input tensor

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -289,11 +289,30 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
...
+        knowledge = ValueKnowledge::join(knowledge, index);
+      if (!knowledge.hasSizes)
+        return getLatticeElement(op->getResult(0)).join(knowledge);
       if (input.hasSizes) {
         knowledge.hasSizes = true;
-        knowledge.sizes.resize(input.sizes.size(), kUnknownSize);
+        // Case: If the input is a 1-d tensor and indices list size is equal
+        // to 1.
+        if (input.sizes.size() == 1 && indices.size() == 1)
+          knowledge.sizes.resize(indices[0].sizes.size(), kUnknownSize);
```
ramiro050, line 313: is there a reason we don't set the size knowledge to `indices[0].sizes`?

ramiro050 (changes requested):
A couple of comments/questions below:

vivekkhandelwal1:
I don't think the reason behind the failing checks has anything to do with the code. Because when I tested it on my system, it was working and passing all the test cases.

ramiro050:
> I don't think the reason behind the failing checks has anything to do with the code. Because when I tested it on my system, it was working and passing all the test cases.

What checks are failing? The CI says everything passed

vivekkhandelwal1:
> > I don't think the reason behind the failing checks has anything to do with the code. Because when I tested it on my system, it was working and passing all the test cases.
> 
> What checks are failing? The CI says everything passed

Earlier it was failing. I think after I re-run the job, checks passed.

vivekkhandelwal1:
@ramiro050, Is this PR good to merge?

ramiro050:
> @ramiro050, Is this PR good to merge?

Everything looks good. The only thing missing are the changes to the `ReduceOpVariants` file. It would be nice to have that fixed before merging

vivekkhandelwal1:
@ramiro050, I have made some changes to the `ReduceOpVariants` file. Please take a look at that.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -59,6 +61,10 @@ class ConvertToImmutableTensors : public RewritePattern {
             listConstruct.elements(), [&](Value tensor) -> Value {
               return rewriter.create<CopyToValueTensorOp>(op->getLoc(), tensor);
             }));
+        // Checking if the OptionalType is a NonValueTensorType optional type.
+        if (listType.getContainedType().isa<OptionalType>() &&
+            newListElements.front().getType().isa<NonValueTensorType>())
```
ramiro050, line 66: I don't think this is right. We should be checking all the elements in the list. This should also happen before we perform the `CopyToValueTensorOp` because by the time we copy the tensor we should be sure that it is not none.  I think we should do what @cathyzhyi mentioned, which is to handle each optional tensor the same way that it is done here

https://github.com/llvm/torch-mlir/blob/28be3d84ce0735c45a215e88729ae63e5ac7f1d9/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp#L72

All the checks happen before we `CopyToValueTensor`. Note: This will more than likely result in a lot of shared code, so I think it would be good to have a helper function that takes care of checking that an optional tensor is valid for copying to value tensor and use that here and in the `else-if` branch below that handles optional types.

Also, it would be nice to have some sort of test to make sure things are indeed working as expected. You can add a test like this one

https://github.com/llvm/torch-mlir/blob/4486de5ef3f31e2ae45b1a557823b4f635694631/test/Dialect/Torch/reduce-op-variants.mlir#L70

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -39,68 +39,81 @@ class ConvertToImmutableTensors : public RewritePattern {
...
-            opOperand.get().getDefiningOp<PrimListConstructOp>();
-        if (!listConstruct) {
-          rewriter.cancelRootUpdate(op);
-          return rewriter.notifyMatchFailure(op, 
-              "unimplemented: list of non vtensor type not constructed "
-              "from list construct");
-        }
+        if (!(listType.getContainedType().isa<NonValueTensorType>() ||
+              listType.getContainedType().isa<OptionalType>()))
+            continue;
```
ramiro050, line 44: I think these are the same changes that @pashu123 added in https://github.com/llvm/torch-mlir/pull/488, so I'm copying over the comments to here:

https://github.com/llvm/torch-mlir/pull/488#discussion_r779109555
https://github.com/llvm/torch-mlir/pull/488#discussion_r779176882

An example of how the formatting is off is the `continue;` above. It should be indented two spaces from the starting column of the `if`.

ramiro050 (changes requested):

test/Dialect/Torch/reduce-op-variants.mlir
```diff
@@ -112,3 +112,25 @@ func @torch.tensor.literal() -> !torch.tensor {
...
+// CHECK:           return %[[RET]] : !torch.tensor
+// CHECK:         }
+func @convert_to_value_semantic_tensors_optional_list(%self: !torch.tensor<[5],f32>,
+                                                 %indices: !torch.tensor<[2, 3],si64>)
+                                                  -> !torch.tensor {
+    %tensor_optional_list = torch.prim.ListConstruct %indices : (!torch.tensor<[2, 3],si64>)
+    -> !torch.list<!torch.optional<!torch.tensor<[2, 3],si64>>>
+    %ret = torch.aten.index.Tensor %self, %tensor_optional_list : !torch.tensor<[5],f32>,
+              !torch.list<!torch.optional<!torch.tensor<[2, 3],si64>>> -> !torch.tensor
+    return %ret: !torch.tensor
```
ramiro050, line 135: nit: Indentation is off. Most of the functions in this file use 2 space indentation. Also, the lines 129, 130, 132, and 134, all seem to follow different indentation rules. One way to easily format MLIR code is to put it in a text file, and the run `torch-mlir-opt text-file.mlir` (however, you do have to rename the variables afterward). Running it and renaming the variables results in the following

```mlir
func @convert_to_value_semantic_tensors_optional_list(%self: !torch.tensor<[5],f32>, %indices: !torch.tensor<[2,3],si64>) -> !torch.tensor {
  %tensor_optional_list = torch.prim.ListConstruct %indices : (!torch.tensor<[2,3],si64>) -> !torch.list<!torch.optional<!torch.tensor<[2,3],si64>>>
  %ret = torch.aten.index.Tensor %self, %tensor_optional_list : !torch.tensor<[5],f32>, !torch.list<!torch.optional<!torch.tensor<[2,3],si64>>> -> !torch.tensor
  return %ret : !torch.tensor
}
```

ramiro050 (approved):
LGTM, just a small style fix request:

PR #498, opened 2021-12-17
[TORCH][MLIR] Add E2E support for `aten.reshape` op

Shukla-Gaurav:
This commit decomposes `aten.reshape` into `aten.view` op in the case of
value tensor type operand.

### DIFF
--- a/e2e_testing/torchscript/main.py
+++ b/e2e_testing/torchscript/main.py
@@ -39,7 +39,7 @@ from . import backprop
 from . import reduction
 from . import argmax
 from . import matmul
-from . import view
+from . import reshape_like
 from . import scalar
 from . import squeeze
 from . import slice_like
--- a/e2e_testing/torchscript/reshape_like.py
+++ b/e2e_testing/torchscript/reshape_like.py
@@ -123,3 +123,41 @@ class View1DFoldModule(torch.nn.Module):
 @register_test_case(module_factory=lambda: View1DFoldModule())
 def View1DFoldModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(32))
+
+# ==============================================================================
+
+class ReshapeExpandModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1], torch.float32, True),
+    ])
+
+    def forward(self, a):
+        return a.reshape(12, 32)
+
+@register_test_case(module_factory=lambda: ReshapeExpandModule())
+def ReshapeExpandModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(384))
+
+# ==============================================================================
+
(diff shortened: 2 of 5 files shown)


ramiro050 (approved):
LGTM!

dan-garvey (approved):

silvasean:
This is not a legal lowering.

https://pytorch.org/docs/stable/generated/torch.reshape.html

```
Returns a tensor with the same data and number of elements as input, but with the specified shape. When possible, the returned tensor will be a view of input. Otherwise, it will be a copy. Contiguous inputs and inputs with compatible strides can be reshaped without copying, but you should not depend on the copying vs. viewing behavior.
```

So in some cases, it needs a deep copy rather than a view.

silvasean:
I would suggest to add a test case covering the situations where reshaping must copy (e.g. when reshaping a view).

Shukla-Gaurav:
> I would suggest to add a test case covering the situations where reshaping must copy (e.g. when reshaping a non-contiguous view).

How can I check for the non-contiguous case? @silvasean @ramiro050 @cathyzhyi

cathyzhyi:
> I would suggest to add a test case covering the situations where reshaping must copy (e.g. when reshaping a non-contiguous view).

@silvasean I think we won't be able to check if the tensor is contiguous or not. We currently represent the contiguous, contiguous tensors the same way right?

silvasean:
It won't be as easy as checking the type, but I think we can probably do something. Can you add the IR snippet of how this occurs in a real program?

cathyzhyi:
> It won't be as easy as checking the type, but I think we can probably do something. Can you add the IR snippet of how this occurs in a real program?

In this snippet, %0 would have become discontiguous tensor in TorchScript.
```
func @forward(%arg0: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[6],f32> {
  %int6 = torch.constant.int 6
  %0 = torch.aten.t %arg0 : !torch.vtensor<[?,?],f32> -> !torch.vtensor<[?,?],f32>
  %1 = torch.prim.ListConstruct %int6 : (!torch.int) -> !torch.list<!torch.int>
  %2 = torch.aten.reshape %0, %1 : !torch.vtensor<[?,?],f32>, !torch.list<!torch.int> -> !torch.vtensor<[6],f32>
  return %2 : !torch.vtensor<[6],f32>
}
```

silvasean:
Oh, I didn't realize this was happening with vtensor's :) In that case it's very easy! We can just lower it to `view` op on vtensor's.

Shukla-Gaurav:
All the comments have been addressed now. @cathyzhyi @ramiro050 @silvasean Can I go ahead and merge this?

cathyzhyi (approved):
LGTM

cathyzhyi:
@Shukla-Gaurav could you rebase and double check if tests are passing now? then you can go ahead and merge it.

PR #500, opened 2021-12-23
[TORCH][MLIR] Fold trivial cases of `aten.to.dtype` and `aten.view` op

Shukla-Gaurav:
- It folds `aten.to.dtype` when the input tensor type and result type
  are exactly same.
- It folds `aten.view` when the rank of both the input tensor type and
  result type is unity.


pashu123 (approved):
LGTM!

ramiro050 (approved):
LGTM!

PR #501, opened 2021-12-23
[MLIR][TORCH] Add E2E support for torch.arange op

vivekkhandelwal1:
This commit adds lowering of `aten.arange.start_step` op.
This commit decomposes `aten.arange` and `aten.arange.start` into
`aten.arange.start_step` op.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -564,6 +564,48 @@ public:
 } // namespace
 
 namespace {
+// The `aten.arange` op is converted to `aten.arange.start_step` op.
+class DecomposeAtenArangeOp : public OpRewritePattern<AtenArangeOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenArangeOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    // The AtenArangeOp doesn't have a start and step value. Therefore we set
+    // them as default values 0 and 1, respectively.
+    Value start, step;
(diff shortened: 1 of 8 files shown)


pashu123 (changes requested):

gprateek93 (commented):
Nice work. One minor comment.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -563,6 +563,51 @@ class DecomposeAtenEmptyLikeOp : public OpRewritePattern<AtenEmptyLikeOp> {
...
+    Location loc = op.getLoc();
+    Value start, step;
+    Value end = op.end();
+    Value result = op.result();
+    BaseTensorType resultType = result.getType().cast<BaseTensorType>();
+    bool isFloatDtype =
+        resultType.hasDtype() && resultType.getDtype().isa<mlir::FloatType>();
+    // If the op is a AtenArangeStartOp then it has a start value. Otherwise we
+    // set it as default value 0.
+    if (isa<AtenArangeStartOp>(op)) {
```
ramiro050, line 595: I'm always a bit unsure about combining templates and dynamic casts. The main advantage of a template is that it allows compile time polymorphism (which can have a speed benefit), but the dynamic cast counteracts the benefit by requiring type information during runtime. 

One way of avoiding this issue is to have the template be a helper function that takes as arguments the `op` and the `start` value, then the two classes for decomposing `aten.arange` and `aten.arange.start` can call this function. In other words, we are abstracting too early. What the two ops have in common is not the decomposition, but the decomposition once the `start` value has been determined

ramiro050 (changes requested):

ramiro050 (approved):
Thanks for this. LGTM!

Issue #415, opened 2021-11-11
Shape inference on modules returns ?


denolf:
Hi,

Even for a module with just a single con2d layer, shape inference returns a set of question marks:
`func @forward(%arg0: !torch.vtensor<[1,1,64,64],f32>) -> !torch.vtensor<[?,?,?,?],f32>` 

while` @annotate_args` was used to specify the size of the input tensor.

Is there a pass available that can resolve this? Currently the pipelined `torchscript-module-to-torch-backend-pipeline` was used.

The python code for the small test is attached. 

Additionally, what is the decomposition of the `torchscript-module-to-torch-backend-pipeline`? Using `symbol-dce,torch-prepare-for-globalize-object-graph,torch-globalize-object-graph,symbol-dce,inline,torch-adjust-calling-conventions,torch-inline-global-slots` does not produce the same output.

Thanks

Kristof
[cnv1TorchMLIR.py.txt](https://github.com/llvm/torch-mlir/files/7523128/cnv1TorchMLIR.py.txt)

silvasean:
I'm in the process of revamping our shape inference to be more precise. Historically, we only needed ranks, because linalg would do the per-dimension size inference. But now that we have more backends, a more generalized approach is needed.

You can see the definition of the pipeline here: https://github.com/llvm/torch-mlir/blob/1019ddf5a0edc388666bf88cd16d8cbbf9c66d9a/lib/Dialect/Torch/Transforms/Passes.cpp#L35

denolf:
Hi Sean,

That sounds promising and interesting. Do you have any idea when that will be ready, I am eager to try it out.

Thanks

Kristof

silvasean:
I was hoping by the end of this week, but I guess that has come. I still have a lot to do on it. Give me another week.

sjarus:
@silvasean , I have what looks like a similar shape inference signature. For testing purposes, I added an ElementwiseNegModule op exactly on the lines of the Log one in e2e_testing/torchscript/elementwise.py , calling torch.neg instead of torch.log. 

It fails with this signature: 

        Exception:
        Lower Torch Backend IR -> TOSA Backend IR failed with the following diagnostics:
        error: unsupported by backend lowering: tensor with unknown rank or dtype
        note: see current operation: %0 = "torch.aten.neg"(%arg0) : (!torch.vtensor<[?,?],f32>) -> !torch.vtensor
        note: this is likely due to a missing case in RefineTypes

Looking at the generated MLIR: 

[...]

PR #502, opened 2021-12-27
Refine static shapes for return tensors of conv2d and maxpool2d

ljfitz:
In RefineTypes.cpp, the return `dtype` was being determined for conv2d and maxpool2d, but the shape was not. This code adds static shape computation then the operator arguments provide the required static details.

This may relate to/clash with other possible approaches to #415, where @silvasean talks about revamping shape inference to be more precise, but I don't know what @silvasean had in mind.

### DIFF
--- a/lib/Dialect/Torch/Transforms/RefineTypes.cpp
+++ b/lib/Dialect/Torch/Transforms/RefineTypes.cpp
@@ -824,16 +824,61 @@ ChangeResult TypeAnalyzer::visitAtenLinearOp(
   return getLatticeElement(op->getResult(0)).join(knowledge);
 }
 
+static int64_t getOutputDimForOpWithKernel(int64_t dimIn, int64_t padding,
+                                           int64_t dilation, int64_t kernelSize,
+                                           int64_t stride) {
+  return ((dimIn + 2 * padding - dilation * (kernelSize - 1) - 1) / stride) + 1;
+}
+
+template <class Op>
+std::vector<int64_t>
+computeOpWithKernelOutputShape(Op op, const ValueKnowledge &ifm,
+                               int64_t features, int64_t kernelHeight,
+                               int64_t kernelWidth) {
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -819,16 +819,63 @@ ChangeResult TypeAnalyzer::visitAtenLinearOp(
   return getLatticeElement(op->getResult(0)).join(knowledge);
 }
 
+// dim_out =
+//  floor((dim_in + 2 * padding - dilation * (kernelSize - 1) - 1) / stride) + 1
```
ramiro050, line 823: Nit: I don't think this is necessary since this information can very easily be determined from the body of the function

ramiro050 (changes requested):
Thanks for this! Everything looks good to me, I just have a small style comment below

ramiro050 (commented):
Oh also, given that the size calculations are really complicated, it would be nice to have an e2e test with static shapes to make sure our code agrees with PyTorch

https://github.com/llvm/torch-mlir/blob/main/e2e_testing/torchscript/conv.py

ljfitz:
> e2e test with static shapes

Thanks @ramiro050 for the review and the suggestion. I've added an e2e test and manually verified that the output of `torchscript-module-to-torch-backend-pipeline` has information being propagated as expected.

ramiro050 (approved):
LGTM! Thanks for adding this!

ljfitz:
> LGTM! Thanks for adding this!

And thanks for the supportive review. I see that I haven't permission to merge the the pull request. Do you know if merging will happen automatically or if I need to take some additional action?

ramiro050:
> And thanks for the supportive review. I see that I haven't permission to merge the the pull request. Do you know if merging will happen automatically or if I need to take some additional action?

Merged!

PR #503, opened 2021-12-30
Make broadcasting result shape more static

cathyzhyi:
This involes the following 2 parts:
- Change refine type to propagate more static shape info.
- Get as much static shape info as possible when creating the result
tensor when converting to linalg.

### DIFF
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -2345,7 +2345,7 @@ struct ConvertElementwiseOp : ConversionPattern {
         // undefined behavior, by doing appropriate checks against the current
         // dimension size.
         auto currentDimSize =
-            rewriter.create<tensor::DimOp>(loc, tensorOperand, size.index());
+            getDimOp(rewriter, loc, tensorOperand, size.index());
 
         // If the result size of this dimension has so far only hit the
         // statically-known-to-be-1 case above (i.e., we have not yet assigned a
@@ -2372,12 +2372,13 @@ struct ConvertElementwiseOp : ConversionPattern {
           /*dimCount=*/resultRank, /*symbolCount=*/0, exprs, getContext()));
     }
 
-    SmallVector<StringRef> iteratorTypes(resultRank, "parallel");
+    SmallVector<StringRef> iteratorTypes(resultRank,
+                                         getParallelIteratorTypeName());
     // Add the indexing map for the outs init tensor.
     indexingMaps.push_back(rewriter.getMultiDimIdentityMap(resultRank));
 
     Value initTensor = rewriter.create<linalg::InitTensorOp>(
-        loc, resultShape, resultType.getElementType());
+        loc, getAsOpFoldResult(resultShape), resultType.getElementType());
     bool hadErrorCreatingPayload = false;
     auto generic = rewriter.create<linalg::GenericOp>(
         loc, /*resultTensorTypes=*/initTensor.getType(),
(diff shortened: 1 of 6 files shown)


lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2376,8 +2376,20 @@ struct ConvertElementwiseOp : ConversionPattern {
...
+          if (auto constOp = v.getDefiningOp<arith::ConstantOp>())
+            return constOp.getValue().cast<IntegerAttr>().getInt();
+          return -1;
+        }));
+
+    for (int i = staticResultShape.size() - 1; i >= 0; i--) {
+      if (staticResultShape[i] != -1)
+        resultShape.erase(resultShape.begin() + i);
+    }
+
```
ramiro050, line 2390: question: is there a reason why this is not being handled by `dispatchIndexOpFoldResults` here

https://github.com/llvm/llvm-project/blob/9e6f88b31a7f7957a850d3541ffa759f2993ffc1/mlir/lib/Dialect/Linalg/IR/LinalgOps.cpp#L875-L876

Definition of `dispatchIndexOpFoldResults`: https://github.com/llvm/llvm-project/blob/fe68088d44f760c7d3d8ee6735d396d97cb55478/mlir/lib/Dialect/Utils/StaticValueUtils.cpp#L22

ramiro050 (changes requested):
The changes look good to me! I just have a few suggestions/questions:

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -759,6 +759,44 @@ static void fillInSizesGivenSizesList(ValueKnowledge &knowledge, Value sizes) {
...
+    for (int64_t i = 0; i < resultRank; i++) {
+      int64_t lhsDimSize = rankAdjustedSizesLhs[i];
+      int64_t rhsDimSize = rankAdjustedSizesRhs[i];
+      // Dynamic shape can't be decided at compilation.
+      if (lhsDimSize == kUnknownSize || rhsDimSize == kUnknownSize)
+        continue;
+
+      // Incompatible broadcasting shape.
+      if (lhsDimSize != rhsDimSize && lhsDimSize != 1 && rhsDimSize != 1)
+        continue;
```
ramiro050, line 793: If we've statically determined that the input tensors are of the wrong shape, is it still correct to return a shape as the final knowledge, or should the knowledge be set to the pessimistic value?

I'm trying to understand how to interpret the following expected invariant:

https://github.com/llvm/torch-mlir/blob/34562b42b1295ea5074627bfb845bab6f311737b/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L117-L118

It seems to me that when we were not explicitly handling the static size analysis done here, there existed scenarios where the invariant would be broken. In particular, in the case where the input tensors had static non-broadcastable sizes, we were still wrongly claiming to know the exact rank of the result. However, it would not be useful to just return the pessimistic result for everything. Ignoring static analysis forced us to assume that the inputs are always well formed for broadcasting.

However, now that we are performing static analysis, we can be a bit more strict about the invariant. What do you think?

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/RefineTypes.cpp
cathyzhyi: yea, that's a good point. Thanks for the thorough thought on this.

ramiro050 (approved):
LGTM! Thanks for adding this

Issue #506, opened 2022-01-03
build error: no member named 'Operator'in namespace 'torch::jit'


Innally:
I was running the whole project from the old build perfectly earlier today, 
but later I re-build the project completely because I messed up my python venv. Basically, I repeated everything I did like the first time  I built the project( I even delete the project and cloned it again to make sure the building process is the same) 
However, when I do this command: 
`cmake --build build --target tools/torch-mlir/all
`

12 errors in total I'll list them blow.
Could anyone help me with this? or is it because of the version?
 
Errors i got:
`/project/yh/gnn/torch-mlir/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/get_registered_ops.cpp:51:42: error: no member named 'Operator' in namespace 'torch::jit'
  for (const std::shared_ptr<torch::jit::Operator> &op :
                             ~~~~~~~~~~~~^
/project/yh/gnn/torch-mlir/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/get_registered_ops.cpp:52:20: error: no member named 'getAllOperators' in namespace 'torch::jit'
       torch::jit::getAllOperators()) {
       ~~~~~~~~~~~~^
/project/yh/gnn/torch-mlir/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/get_registered_ops.cpp:58:23: error: member access into incomplete type 'const c10::FunctionSchema'
      name[0] = schema.name();
                      ^
/project/yh/gnn/torch-mlir/mlir_venv/lib/python3.9/site-packages/torch/include/ATen/core/class_type.h:17:8: note: forward declaration of 'c10::FunctionSchema'
struct FunctionSchema;
       ^
/project/yh/gnn/torch-mlir/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/get_registered_ops.cpp:59:23: error: member access into incomplete type 'const c10::FunctionSchema'
      name[1] = schema.overload_name();
                      ^
/project/yh/gnn/torch-mlir/mlir_venv/lib/python3.9/site-packages/torch/include/ATen/core/class_type.h:17:8: note: forward declaration of 'c10::FunctionSchema'
struct FunctionSchema;
       ^
/project/yh/gnn/torch-mlir/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/get_registered_ops.cpp:64:33: error: member access into incomplete type 'const c10::FunctionSchema'
    record["is_vararg"] = schema.is_vararg();
                                ^
[...]

PR #507, opened 2022-01-03
Fix build error regarding missing types in torch::jit

ramiro050:
This commit adds include statements of the file
`torch/csrc/jit/ir/ir.h` for files that use types from torch::jit.

Fixes https://github.com/llvm/torch-mlir/issues/506


cathyzhyi (approved):
Thanks for the fix!

PR #508, opened 2022-01-04
Update llvm for tensor collapse/expand

dan-garvey:
EDIT: updates llvm to: d3208cf1cdc927e5e74f6f6a1a69dc548ac629e0

### DIFF
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -11,7 +11,7 @@
 
 #include "../PassDetail.h"
 #include "mlir/Dialect/Arithmetic/IR/Arithmetic.h"
-#include "mlir/Dialect/Linalg/IR/LinalgOps.h"
+#include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Math/IR/Math.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Traits.h"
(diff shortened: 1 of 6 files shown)


cathyzhyi:
Thanks for this change. I think we need to use the sha from upstream repo for the submodule sha. We might alos need to add some more fixes. I am trying in my local repo.

cathyzhyi:
Here are some more fixes https://github.com/cathyzhyi/mlir-npcomp/commit/2eff10f9254f7e03ea0d5fefe39778c59b8e9ae1

cathyzhyi:
FYI, the asm printer is also broken by https://reviews.llvm.org/D113873

dan-garvey:
> FYI, the asm printer is also broken by https://reviews.llvm.org/D113873

I'll take a look

cathyzhyi:
After some discussion with MLIR folks, people are proposing to add options to revert https://reviews.llvm.org/D113873. This change makes
```
%0 = torch.aten.size %arg0 : !torch.vtensor<[?,3],f32> -> !torch.list<!torch.int>
```
becomes
```
%0 = torch.aten.size %arg0 : !torch.vtensor<[?,3],f32> -> <!torch.int>
```
which affects the readability.
The author of the change would be back next week so I would suggest we hold off the change for now if it's not too urgent.

cathyzhyi:
FYI, this is the [fix](https://reviews.llvm.org/D116905) for the printer failures.

cathyzhyi:
@dan-garvey I will take this one as it's needed for https://github.com/llvm/torch-mlir/pull/528

cathyzhyi:
closing as it's covered by #529

PR #513, opened 2022-01-07
[MLIR][TORCH] Fix tensor literal int elem type to be signless

cathyzhyi:
The element type of tensor literal should be signless when converted to
builtin tensor types.

### DIFF
--- a/e2e_testing/torchscript/basic.py
+++ b/e2e_testing/torchscript/basic.py
@@ -602,7 +602,7 @@ class ContiguousModule(torch.nn.Module):
 @register_test_case(module_factory=lambda: ContiguousModule())
 def ContiguousModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 1))
-    
+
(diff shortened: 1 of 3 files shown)


e2e_testing/torchscript/basic.py
```diff
@@ -963,3 +963,38 @@ def forward(self, lhs):
 def TModuleRank0_basic(module, tu: TestUtils):
     module.forward(torch.tensor(7, dtype=torch.float32))
 
+class TensorLiteralModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        torch.manual_seed(0)
+        self.t = torch.randn(2, 3)
```
ramiro050, line 970: I think this should be `randint`. `randn` returns floats. Same for the second test

ramiro050 (changes requested):
A couple of comments below:

ramiro050 (approved):
LGTM!

PR #515, opened 2022-01-10
Add support for constant_pad_nd

ljfitz:
Note that to enable folding of the code coming from an example
like the ConstantPad2dStaticModule e2e test, support for other
operations had to be added/improved:
- aten::neg.int
- aten::eq.float
- aten::eq.str
- prim::Uninitialized

### DIFF
--- a/e2e_testing/torchscript/basic.py
+++ b/e2e_testing/torchscript/basic.py
@@ -284,6 +284,47 @@ def MaxPool2dModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(1, 1, 20, 20) - 0.5)
 
 
+class ConstantPad2dStaticModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.pad2d = torch.nn.ConstantPad2d((0, 1, 2, 3), -float('inf'))
+
+    @export
+    @annotate_args([
+        None,
+        ([1, 1, 20, 20], torch.float32, True),
+    ])
+    def forward(self, x):
+        return self.pad2d(x)
+
+# ==============================================================================
+
+
+@register_test_case(module_factory=lambda: ConstantPad2dStaticModule())
+def ConstantPad2dStaticModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(1, 1, 20, 20) - 0.5)
+
+
+class ConstantPadNdStaticModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([1, 1, 20, 20, 4, 4], torch.float32, True),
+    ])
+    def forward(self, x):
+        return torch.ops.aten.constant_pad_nd(x, (0, 1), -float('inf'))
+
+# ==============================================================================
+
+
+@register_test_case(module_factory=lambda: ConstantPadNdStaticModule())
+def ConstantPadNdStaticModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(1, 1, 20, 20, 4, 4) - 0.5)
+
+
 class TransposeIntModule(torch.nn.Module):
     def __init__(self):
         super().__init__()
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
@@ -45,12 +45,33 @@ struct torch_constant_int_op_binder {
     return false;
   }
 };
+
+struct torch_constant_float_op_binder {
+  APFloat *bind_value;
+
+  /// Creates a matcher instance that binds the value to bv if match succeeds.
+  torch_constant_float_op_binder(APFloat *bv) : bind_value(bv) {}
+
+  bool match(Operation *op) {
+    if (auto constantFloat = dyn_cast<Torch::ConstantFloatOp>(op)) {
+      *bind_value = constantFloat.value();
+      return true;
+    }
+    return false;
+  }
+};
 } // namespace detail
 
 /// Matches the integer stored in a `torch.constant.bool`.
 inline detail::torch_constant_int_op_binder
 m_TorchConstantInt(int64_t *bind_value) {
   return detail::torch_constant_int_op_binder(bind_value);
+}
+
+/// Matches the integer stored in a `torch.constant.float`.
+inline detail::torch_constant_float_op_binder
+m_TorchConstantFloat(APFloat *bind_value) {
+  return detail::torch_constant_float_op_binder(bind_value);
 }
 
 namespace detail {
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -275,7 +275,24 @@ static SmallVector<Value> getTypeConvertedValues(OpBuilder &b, Location loc,
 }
 
 // Helper function to get the padding tensor given the padding int values.
-// It's assumed that the padding on the low end and high end are the same.
+static Value getPaddedTensor(Operation *op, OpBuilder &b, Value &input,
+                             SmallVectorImpl<int64_t> &lowPaddingInts,
+                             SmallVectorImpl<int64_t> &highPaddingInts,
+                             Value pad) {
+  Location loc = op->getLoc();
+  Type rankedTensorType = linalg::PadTensorOp::inferResultType(
+      input.getType().cast<RankedTensorType>(), lowPaddingInts, highPaddingInts);
+  SmallVector<OpFoldResult> lowPaddings = getAsOpFoldResult(b, loc, lowPaddingInts);
+  SmallVector<OpFoldResult> highPaddings = getAsOpFoldResult(b, loc, highPaddingInts);
+  Value paddedInput = linalg::PadTensorOp::createPadScalarOp(
+      rankedTensorType, input, pad, /*low=*/lowPaddings, /*high=*/highPaddings,
+      /*packing=*/false, loc, b);
+  return paddedInput;
+}
+
+// Helper function to get the padding tensor given the padding int values.
+// It's assumed that the padding on the low end and high end are the same,
+// and that zero padding is required.
 static Value getPaddedTensor(Operation *op, OpBuilder &b, Value &input,
                              SmallVectorImpl<int64_t> &paddingInts) {
   assert(input.getType().isa<RankedTensorType>() &&
@@ -284,13 +301,7 @@ static Value getPaddedTensor(Operation *op, OpBuilder &b, Value &input,
   Value c0 = b.create<arith::ConstantOp>(
       loc,
       b.getZeroAttr(input.getType().cast<RankedTensorType>().getElementType()));
-  SmallVector<OpFoldResult> paddings = getAsOpFoldResult(b, loc, paddingInts);
-  Type ranked4DTensorType = linalg::PadTensorOp::inferResultType(
-      input.getType().cast<RankedTensorType>(), paddingInts, paddingInts);
-  Value paddedInput = linalg::PadTensorOp::createPadScalarOp(
-      ranked4DTensorType, input, c0, /*low=*/paddings, /*high=*/paddings,
-      /*packing=*/false, loc, b);
-  return paddedInput;
+  return getPaddedTensor(op, b, input, paddingInts, paddingInts, c0);
 }
 
 static Value buildNormalCdf(OpBuilder &b, Location &loc, Value x, Value mean,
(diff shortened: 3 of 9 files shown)


e2e_testing/torchscript/basic.py
```diff
@@ -284,6 +284,47 @@ def MaxPool2dModule_basic(module, tu: TestUtils):
...
+        return self.pad2d(x)
+
+# ==============================================================================
+
+
+@register_test_case(module_factory=lambda: ConstantPad2dStaticModule())
+def ConstantPad2dStaticModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(1, 1, 20, 20) - 0.5)
+
+
```
cathyzhyi, line 307: We usually put the delimiter before the testing module here. Can you also help to fix the existing one for `MaxPool2dModule_basic`

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
```diff
@@ -45,6 +45,21 @@ struct torch_constant_int_op_binder {
     return false;
   }
 };
+
+struct torch_constant_float_op_binder {
+  APFloat *bind_value;
```
cathyzhyi, line 50: Prefer to use the primitive type `double` so that user of the API can use the result number directly for calculation. Like in `AtenEqFloatOp::fold`, `lhs.compare(rhs) == APFloat::cmpEqual` can be replaced with `lhs == rhs`.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1005,6 +1034,20 @@ void PrimTupleIndexOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
 }
 
+//===----------------------------------------------------------------------===//
+// PrimUninitializedOp
+//===----------------------------------------------------------------------===//
+
+void PrimUninitializedOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                                    MLIRContext *context) {
+  patterns.add(+[](PrimUninitializedOp op, PatternRewriter &rewriter) {
+    if (! op.use_empty())
```
cathyzhyi, line 1044: We shouldn't need to add such patterns because the canonicalizer should remove used ops without side effects automatically. I think the reason this is not removed is that the op might have side affects. 

Is this op on some dead code path?

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1129,6 +1172,18 @@ OpFoldResult AtenMulIntOp::fold(ArrayRef<Attribute> operands) {
...
 }
 
+//===----------------------------------------------------------------------===//
+// AtenNegIntOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenNegIntOp::fold(ArrayRef<Attribute> operands) {
+  int64_t c;
+  bool constant = matchPattern(getOperand(), m_TorchConstantInt(&c));
+  if (constant)
```
cathyzhyi, line 1182: nit: don't need the `constant` variable.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -653,6 +655,33 @@ OpFoldResult AtenEqIntOp::fold(ArrayRef<Attribute> operands) {
                               [](int64_t a, int64_t b) { return a == b; });
 }
 
+//===----------------------------------------------------------------------===//
+// AtenEqFloatOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenEqFloatOp::fold(ArrayRef<Attribute> operands) {
```
cathyzhyi, line 662: all folders need unit tests in https://github.com/llvm/mlir-npcomp/blob/d6b6c0268cfd6ead0aa5f2f21ba02bbe8ead3180/test/Dialect/Torch/canonicalize.mlir

cathyzhyi (changes requested):
I did a first pass. Will take a closer look at torch lowering and refine types later.

lib/Dialect/Torch/IR/TorchOps.cpp
ljfitz: For the e2e example, without this a `"prim.Uninitialized"` is present unconditionally but the result is unused. I cannot track down a side-effect that would require it to remain present. Any hints as to where it make be treated as having a (pseudo) side-effect that prevents it being otherwise removed?

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2685,6 +2696,58 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
...
+    SmallVector<int64_t, 4> highPadding(/*Size=*/rank - padRank, /*Value=*/0);
+    // add requested padding - note op.pad() is higest dim first ordered pairs
+    // of low,high
+    for (uint64_t i = padRank; i > 0; --i) {
+      lowPadding.push_back(padInts[i * 2 - 2]);
+      highPadding.push_back(padInts[i * 2 - 1]);
+    }
+
+    Value castedValue =
+        rewriter.create<arith::TruncFOp>(loc, type.getElementType(), adaptor.value());
```
cathyzhyi, line 2741: the `convertScalarToDtype` helper can be used to covert the `value` type to tensor element type. 

Also, technically, the `elementType` need to be got from the converted result type, `getTypeConverter()->convertType(op.getType())`;.  This is not an issue for float because the element type of converted tensor type and the torch tensor type are the same. However, it would be different for integer tensor types.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2685,6 +2696,58 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
...
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+    Location loc = op->getLoc();
+    Value self = adaptor.self();
+    auto type = self.getType().cast<RankedTensorType>();
+    int64_t rank = type.getRank();
+
+    Type elementType = self.getType().cast<RankedTensorType>().getElementType();
+    if (!elementType.isa<mlir::FloatType>())
```
cathyzhyi, line 2714: This type limitation can be removed once the `arith::TruncFOp` below is changed to use `convertScalarToDtype`

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -953,6 +958,28 @@ ChangeResult TypeAnalyzer::visitAtenMaxPool2dOp(
   return getLatticeElement(op->getResult(0)).join(knowledge);
 }
 
+ChangeResult TypeAnalyzer::visitAtenConstantPadNdOp(
+    AtenConstantPadNdOp op,
+    ArrayRef<LatticeElement<ValueKnowledge> *> operands) {
+  auto knowledge =
+      ValueKnowledge::getNotNonePessimisticValueState(op->getContext());
+  knowledge.hasSizes = true;
+  auto &ifm = operands[0]->getValue();
```
cathyzhyi, line 967: we don't need the & as the return type of `getValue` is already reference type.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -953,6 +958,28 @@ ChangeResult TypeAnalyzer::visitAtenMaxPool2dOp(
...
+  SmallVector<int64_t, 4> padInts;
+  if (ifm.hasSizes && matchPattern(op.pad(), m_TorchConstantIntList(padInts))) {
+    knowledge.sizes = ifm.sizes;
+    uint64_t padRank = padInts.size() / 2;
+    uint64_t padOffset = knowledge.sizes.size() - padRank;
+    // op.pad() is higest dim first ordered pairs of low,high
```
cathyzhyi, line 973: nit: period at the end of comment.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
cathyzhyi: nit: not sure why is the variable name `ifm`? Can we call it `input`?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -953,6 +958,28 @@ ChangeResult TypeAnalyzer::visitAtenMaxPool2dOp(
...
+    for (uint64_t i = padRank, r = padOffset; i > 0; --i, ++r) {
+      knowledge.sizes[r] += padInts[i * 2 - 2] + padInts[i * 2 - 1];
+    }
+  } else
+    knowledge.sizes.resize(4, kUnknownSize);
```
cathyzhyi, line 978: we don't need the else branch. The `knowledge` is initialized with unknown size. If you would like to be more accurate. You could arrange the branches like
```
if (input.hasSizes) {
   knowledge.hasSizes = true;
   if (matchPattern(op.pad(), m_TorchConstantIntList(padInts))){
     fill in sizes here....
   } else {
     knowledge.sizes.resize(inputRank, kUnknownSize);
   }
}
```

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -953,6 +958,28 @@ ChangeResult TypeAnalyzer::visitAtenMaxPool2dOp(
   return getLatticeElement(op->getResult(0)).join(knowledge);
 }
 
+ChangeResult TypeAnalyzer::visitAtenConstantPadNdOp(
+    AtenConstantPadNdOp op,
+    ArrayRef<LatticeElement<ValueKnowledge> *> operands) {
+  auto knowledge =
+      ValueKnowledge::getNotNonePessimisticValueState(op->getContext());
+  knowledge.hasSizes = true;
```
cathyzhyi, line 966: this need to be deleted at this line as we might not have sizes information for the output.

e2e_testing/torchscript/basic.py
```diff
@@ -284,6 +284,47 @@ def MaxPool2dModule_basic(module, tu: TestUtils):
...
 
+class ConstantPad2dStaticModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.pad2d = torch.nn.ConstantPad2d((0, 1, 2, 3), -float('inf'))
+
+    @export
+    @annotate_args([
+        None,
+        ([1, 1, 20, 20], torch.float32, True),
```
cathyzhyi, line 295: Can we also add testing for dynamic shapes?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -953,6 +958,28 @@ ChangeResult TypeAnalyzer::visitAtenMaxPool2dOp(
...
+  auto &ifm = operands[0]->getValue();
+  SmallVector<int64_t, 4> padInts;
+  if (ifm.hasSizes && matchPattern(op.pad(), m_TorchConstantIntList(padInts))) {
+    knowledge.sizes = ifm.sizes;
+    uint64_t padRank = padInts.size() / 2;
+    uint64_t padOffset = knowledge.sizes.size() - padRank;
+    // op.pad() is higest dim first ordered pairs of low,high
+    for (uint64_t i = padRank, r = padOffset; i > 0; --i, ++r) {
+      knowledge.sizes[r] += padInts[i * 2 - 2] + padInts[i * 2 - 1];
```
cathyzhyi, line 975: need to guard the calculation with if `knowledge.sizes[r] == kunknownSize`

cathyzhyi (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
cathyzhyi: `side-effect` doesn't mean there must be uses. It could just be an op that can through exception or can allocate memory. If an op doesn't have side effect, the op definition need to contain the `NoSideEffect` trait like this one https://github.com/llvm/mlir-npcomp/blob/d6b6c0268cfd6ead0aa5f2f21ba02bbe8ead3180/include/torch-mlir/Dialect/Torch/IR/TorchOps.td#L798.  Currently, we don't support generating that trait for auto generated ops ods yet. 

However, putting all these aside,  according to https://discuss.pytorch.org/t/keyerror-prim-uninitialized/64860 `prim.Uninitialized` means the compiler believe the code can never reach this statement. It's very weird to see the `prim.Uninitialzied` on an unconditional path.

cathyzhyi: Oh, actually, it didn't say it's on dead code path and but the value is not used. Hmm, in this case we could just delete the op unconditionally for now.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
ljfitz: > we don't need the & as the return type of `getValue` is already reference type

While that holds true for `auto` and pointer types, the `&` is necessary if you want a reference rather than a copy (e.g. 
https://github.com/llvm/torch-mlir/pull/515#discussion_r781478778).

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
ljfitz: Thanks for the explanation!

lib/Dialect/Torch/IR/TorchOps.cpp
ljfitz: I'd seen that discussion and considered trying to remove it earlier than the point here. However in the incoming IR the result SSA is referenced, but after some folding and simplifications the unused `prim.Uninitialized` remains. This was the first point I could identify that it was straightforward to remove it. I'm happy to dig further, but would appreciate some pointers if you have them.

ljfitz:
Thanks @cathyzhyi for the careful review - very much appreciated. I think I've addressed everything, though I'm not sure if you find the handling of `prim.Uninitialized` acceptable for this PR or not.

As there were quite a few comments I've pushed a separate commit for now to make it easier for you to check. I'll rebase and squash when the PR is approved.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
cathyzhyi: ha, you are right.

cathyzhyi (approved):
LGTM. Thanks for the PR!

lib/Dialect/Torch/IR/TorchOps.cpp
cathyzhyi: I see. yea, we can leave it as it is then. btw, did you use `clang-format`? seems there is a space after `!`

ljfitz: I'd run it manually but it looks like I forgot to do so after a change. Do you use https://pre-commit.com/ or similar to avoid forgetting?

ljfitz:
> LGTM. Thanks for the PR!

And thank you for the thorough and supportive review. I've rebased, clang-formatted and squashed the commits. I see that I haven't permission to merge the the pull request. Do you know if merging will happen automatically or if I need to take some additional action?

PR #516, opened 2022-01-10
Add support for two return values

vivekkhandelwal1:
This commit adds support for two return values of type
memref f32 and i64.

### DIFF
--- a/lib/RefBackend/RefBackend.cpp
+++ b/lib/RefBackend/RefBackend.cpp
@@ -163,9 +163,9 @@ static LogicalResult mungeFunction(
     auto supportedFuncsEnd = supportedConsumeFuncReturnFuncs.end();
     std::string funcName = getConsumeReturnFunctionNameForReturnTypes(retTypes);
     if (supportedConsumeFuncReturnFuncs.find(funcName) == supportedFuncsEnd) {
-      op.emitError(
-          "must have one return value of memref types or scalar types "
-          "of i32, i64, f32, f64, i1, or three return values of memref f32");
+      op.emitError("must have one return value of memref types or scalar types "
+                   "of i32, i64, f32, f64, i1, or two return values of memref "
+                   "f32 and i64, or three return values of memref f32");
       isSupported = false;
     }
 
@@ -194,7 +194,8 @@ static std::set<std::string> getSupportedConsumeFuncReturnFuncs(OpBuilder &b) {
   Type f64 = b.getF64Type();
 
   SmallVector<TypeRange> supportedReturnTypes = {
-      mri1, mri32, mri64, mrf32, mrf64, i64, f32, f64, {mrf32, mrf32, mrf32}};
+      mri1, mri32, mri64, mrf32,          mrf64,
+      i64,  f32,   f64,   {mrf32, mri64}, {mrf32, mrf32, mrf32}};
 
   llvm::for_each(supportedReturnTypes, [&](TypeRange &types) {
     funcNames.insert(getConsumeReturnFunctionNameForReturnTypes(types));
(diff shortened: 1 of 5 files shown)


python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
```diff
@@ -65,6 +65,42 @@ def consume_return_f32(a):
...
+                                 consume_return_mrf64)
+
+        self.ee.register_runtime("refbackend_consume_func_return_i64",
+                                 consume_return_i64)
+
+        self.ee.register_runtime("refbackend_consume_func_return_f32",
+                                 consume_return_f32)
+
+        self.ee.register_runtime("refbackend_consume_func_return_f64",
+                                 consume_return_f64)
```
ramiro050, line 98: I think the CI error is related to these. These are already present below. You should only need to add the `consume_return_mrf32_mri64`.

lib/RefBackend/RefBackend.cpp
```diff
@@ -194,7 +194,8 @@ static std::set<std::string> getSupportedConsumeFuncReturnFuncs(OpBuilder &b) {
   Type f64 = b.getF64Type();
 
   SmallVector<TypeRange> supportedReturnTypes = {
-      mri1, mri32, mri64, mrf32, mrf64, i64, f32, f64, {mrf32, mrf32, mrf32}};
+      mri1, mri32, mri64, mrf32,          mrf64,
```
ramiro050, line 197: Don't forget to do `clang-format`

ramiro050 (changes requested):
Can you also add an e2e test like this one:

https://github.com/cathyzhyi/mlir-npcomp/blob/9240f72804f7d167d45a32659dd9c7e27d203eda/e2e_testing/torchscript/basic.py#L583

lib/RefBackend/RefBackend.cpp
vivekkhandelwal1: I have done that. This is the resultant formatting.

vivekkhandelwal1:
@ramiro050, I can't understand why CI is failing. When I tested manually, everything worked fine.

ramiro050:
> @ramiro050, I can't understand why CI is failing. When I tested manually, everything worked fine.

New e2e tests are set to "expectedly failed" on the TOSA side by default. Since your test does not need TOSA changes to pass, it is generating an error because it is "unexpectedly passing." You need to add your test to this set

https://github.com/llvm/torch-mlir/blob/ca662dc9ccfe50a2366edc6e807420cfc5efe17d/e2e_testing/torchscript/xfail_sets.py#L24

python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
```diff
@@ -65,6 +65,18 @@ def consume_return_f32(a):
...
+                          ctypes.POINTER(UnrankedMemRefDescriptor))
+        def consume_return_mrf32_mri64(arg0, arg1):
+            self.result = unranked_memref_to_numpy(
+                arg0, np.float32), unranked_memref_to_numpy(
+                    arg1,
+                    np.int64)
+
+        self.ee.register_runtime(
+            "refbackend_consume_func_return_mrf32_mri64",
+            consume_return_mrf32_mri64)
```
ramiro050, line 78: nit: I think its better to move this after 

https://github.com/llvm/torch-mlir/blob/acb50008a9d50e65b82f473ae8463c36121c969c/python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py#L110-L111

to keep the organization consistent

ramiro050 (changes requested):

vivekkhandelwal1:
@ramiro050, is this good to merge?

ramiro050 (approved):
LGTM!

PR #518, opened 2022-01-11
[MLIR][TORCH] Add E2E support for max_pool2d_with_indices op

vivekkhandelwal1:
This commit adds lowering of `max_pool2d_with_indices` op.


e2e_testing/torchscript/basic.py
```diff
@@ -1014,3 +1014,67 @@ def forward(self, a, b):
...
+    def forward(self, x):
+        return torch.ops.aten.max_pool2d_with_indices(x,
+                                                      kernel_size=[6, 8],
+                                                      stride=[2, 2],
+                                                      padding=[3, 4],
+                                                      dilation=2)
+
+@register_test_case(module_factory=lambda: MaxPool2dWithIndicesModule())
+def MaxPool2dWithIndicesModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(2, 4, 6, 8) - 0.5)
```
ramiro050, line 1038: You can specify a range as argument flags in `tu.rand`:

https://github.com/llvm/torch-mlir/blob/30df2ec71b8f7c6f5e0bd5bb26a62e70c0520b77/python/torch_mlir_e2e_test/torchscript/framework.py#L151

Same comment for the other e2e tests

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2584,6 +2584,93 @@ struct ConvertReductionOp : ConversionPattern {
...
+  if (!matchPattern(op.kernel_size(), m_TorchConstantIntList(kernelSizeInts)))
+    return rewriter.notifyMatchFailure(op, "only support kernel size ints");
+
+  Value falseValue = rewriter.create<arith::ConstantOp>(
+      loc, IntegerAttr::get(rewriter.getIntegerType(1), 0));
+  Value ceilModeFalse = rewriter.create<arith::CmpIOp>(
+      loc, arith::CmpIPredicate::eq, ceilMode, falseValue);
+  rewriter.create<AssertOp>(
+      loc, ceilModeFalse,
+      rewriter.getStringAttr("only ceil_mode false is supported"));
```
ramiro050, line 2620: Is there a reason for doing this at runtime? I think we can handle this at compile time by pattern matching on the boolean, similar to how we've done it in other ops

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2584,6 +2584,93 @@ struct ConvertReductionOp : ConversionPattern {
 };
 } // namespace
 
+// Computes maxpool2d for AtenMaxPool2dOp and AtenMaxPool2dWithIndicesOp.
+template <typename OpTy>
+static LogicalResult
+computeMaxPool2d(OpTy op, Location loc, ConversionPatternRewriter &rewriter,
+                 Value self, Value ceilMode, Value &result) {
```
ramiro050, line 2591: nit: `loc` can be computed from `op`, so it does not need to be an argument. Also, if we pattern-match on `ceilMode`, then we can use the `ceilMode = op.ceil_mode()`, which means we don't need to have that argument either.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -953,6 +959,39 @@ ChangeResult TypeAnalyzer::visitAtenMaxPool2dOp(
...
+  auto &ifm = operands[0]->getValue();
+  SmallVector<int64_t, 2> kernelSize;
+  if (!matchPattern(op.kernel_size(), m_TorchConstantIntList(kernelSize)))
+    kernelSize = SmallVector<int64_t, 2>{kUnknownSize, kUnknownSize};
+  if (ifm.hasSizes)
+    outputKnowledge.sizes = computeOpWithKernelOutputShape(
+        op, ifm, ifm.sizes[1], kernelSize[0], kernelSize[1]);
+  else
+    outputKnowledge.sizes.resize(4, kUnknownSize);
+  outputKnowledge.dtype = operands[0]->getValue().dtype;
```
ramiro050, line 980: This seems to be shared by both `AtenMaxPool2d` and `AtenMaxPool2dWithIndices`, so it should be refactored.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
cathyzhyi: ha, I think this was copied from the original code when refactoring. The thinking was runtime assert was more flexible and would work if it's not constant false but a runtime false. However, this is inconsistent with the rest of the codebase so it's better to be changed to the same as other places.

vivekkhandelwal1: Ya, I copied it from the original code. I will make the required changes.

vivekkhandelwal1:
I have not rebased it over the latest upstream changes because the newest commit creates a linker error.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2593,93 +2677,116 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
...
+// also be a 4-d tensor. Then:
+//  for i in input.size[0]:
+//    for j in input.size[1]:
+//      for k in input.size[2]:
+//        for l in input.size[3]:
+//          for p in maxpool2d.size[1]:
+//            for q in maxpool2d.size[2]:
+//              for r in maxpool2d.size[3]:
+//                if input[i, j, k, l] == maxpool2d[i, p, q, r]:
+//                  indices[i, p, q, r] = (k * input.size[3] + l)
```
ramiro050, line 2710: I don't think this algorithm is correct. Lets say the input tensor has two windows `A` and `B`. If a value `x` in `B` happens to equal the maximum of `A`, then the index stored for window `A` could end up being the index of `x`, which is in a different window. I'm also worried a bit about the performance since we are iterating over almost the entire `maxpool2d` tensor for every value in the input tensor.

One way of doing this would be to restrict the search to the window you're currently in. From the values of `k` and `l`, doing some arithmetic gymnastics, one should be able to determine the window that that point is in, allowing you to compare it to the max value for that window in the `maxpool2d` tensor. This removes the need to iterate over the `maxpool2d` tensor, and it ensures that we are only getting indices from the window we are currently examining.

The way I mentioned above is also close to how one could calculate `maxpool2d` in the first place. It seems that to get the indices we are essentially implementing a variation on `maxpool2d` by hand, which is also a bit concerning, since now `maxpool2d_indices` will take at least twice as long as regular `maxpool2d`. Ideally the extraction of the indices should be performed at the same time as the pooling, but this would require changes to linalg. For now I think it is okay to have `O(2 * maxpool2d)` performance.

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
vivekkhandelwal1: Hi @ramiro050, I have a doubt. What if the maximum value appears twice in a window? Which of the index should we choose? Also, do you think it would be easier to implement this in TMTensor? I am not sure where to implement this.

vivekkhandelwal1:
Please don't re-review the code as of now. I have just resolved the conflicts, working on fixing the issues.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
vivekkhandelwal1: Hi @ramiro050 @cathyzhyi, I have thought of the following algorithm for this:

The result of the max pool operation is `maxPool2d`, and its shape is (N, C, Hout, Wout), and the kernel size is (kH, kW). Then
```
for i in range(N):
    for j in range(C):
        for m in range(Hout):
            for n in range(Wout):
                for p in range(kH):
                    for r in range(kW):
                        if input[i, j, m + p, n + r] == maxPool2d[i, j, m, n]:
                            indices[i, j, m, n] = (m + p)*kW + (n + r)
```
This algorithm handles the padding implicitly but does not handle the strides and dilation. I am thinking of a way to get this work for the cases where dilation and strides do not equal one, which is the default value. If you have any suggestions, please do comment. I am implementing this one until I find some way for the non-default stride and dilation to get work.

vivekkhandelwal1: I implemented this algorithm and tested it just now. It works for the default dilation, default stride, and unpadded input case. I will check why it's failing for the padding case for which it should have worked.

cathyzhyi: I think to consider stride and dilation you just need to adjust the input index:
```
		index_H = m*stride[0] + p* dilation[0]
		index_W = n*stride[1] + r* dilation[1]
                if input[i, j, index_H, index_W] == maxPool2d[i, j, m, n]:
                      indices[i, j, m, n] = index_H*kW + index_W
```

vivekkhandelwal1: Able to get this working for all non-default params.

e2e_testing/torchscript/basic.py
```diff
@@ -1014,3 +1014,67 @@ def forward(self, a, b):
...
+
+
+class MaxPool2dWithIndicesStride1dModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1, -1], torch.float32, True),
```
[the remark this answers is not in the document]
vivekkhandelwal1: I don't think the current implementation of maxpool2d supports 3-d cases at present.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -27,6 +27,107 @@ using namespace mlir;
...
 
+// Checks the validity of pooling parameters and stores them in the respective
+// vector.
+template <typename OpTy>
+static LogicalResult
+checkAndGetPoolingParameters(OpTy op, ConversionPatternRewriter &rewriter,
+                             SmallVector<int64_t, 2> &kernelSizeInts,
+                             SmallVector<int64_t, 2> &strideInts,
+                             SmallVector<int64_t, 2> &paddingInts,
+                             SmallVector<int64_t, 2> &dilationInts) {
```
pashu123, line 38: Please pass these as `SmallVectorImpl<int64_t>` .

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -27,6 +27,107 @@ using namespace mlir;
...
+                                       "only support constant bool ceil_mode");
+
+  return success();
+}
+
+// Computes maxpool2d for AtenMaxPool2dOp and AtenMaxPool2dWithIndicesOp.
+template <typename OpTy>
+static LogicalResult
+computeMaxPool2d(OpTy op, ConversionPatternRewriter &rewriter, Value self,
+                 SmallVector<int64_t, 2> kernelSizeInts,
```
pashu123, line 69: Same here `SmallVectorImpl` instead of SmallVector.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -27,6 +27,107 @@ using namespace mlir;
...
+  SmallVector<Value> paddingIntValues =
+      getAsConstantIntValues(rewriter, loc, paddingInts);
+  SmallVector<Value> dilationIntValues =
+      getAsConstantIntValues(rewriter, loc, dilationInts);
+  SmallVector<Value> kernelSizeIntValues =
+      getAsConstantIntValues(rewriter, loc, kernelSizeInts);
+  SmallVector<Value> strideIntValues =
+      getAsConstantIntValues(rewriter, loc, strideInts);
+
+  Value Hout = torch_to_linalg::getOutputDimForConvOps(
```
pashu123, line 98: Rename to `hOut`.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -27,6 +27,107 @@ using namespace mlir;
...
+      rewriter, loc, H, paddingIntValues[0], dilationIntValues[0],
+      kernelSizeIntValues[0], strideIntValues[0]);
+  Value Wout = torch_to_linalg::getOutputDimForConvOps(
```
pashu123, line 101: Same here.

pashu123 (changes requested):

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -36,94 +137,180 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
...
+    for (unsigned i = 0; i < 4; i++) {
+      inputExprs.push_back(rewriter.getAffineDimExpr(i));
+      outputExprs.push_back(rewriter.getAffineDimExpr(i));
+    }
+    kernelExprs.push_back(rewriter.getAffineDimExpr(4));
+    kernelExprs.push_back(rewriter.getAffineDimExpr(5));
+
+    // Here we have six dimensions, each corresponding to N, C, Hout, Wout, kH,
+    // and kW, respectively, as described in the algorithm above.
+    SmallVector<AffineMap> indexingMaps = {
```
cathyzhyi, line 246: nit: could use `AffineMap::inferFromExprList`

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -36,94 +137,180 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
...
+        AffineMap::get(6, 0, inputExprs, op.getContext()),
+        AffineMap::get(6, 0, kernelExprs, op.getContext()),
+        AffineMap::get(6, 0, outputExprs, op.getContext())};
+    SmallVector<StringRef> iteratorTypes(6, getParallelIteratorTypeName());
```
cathyzhyi, line 250: If a max value appeared more than once in the window would the smallest index would be returned. I think the iteration along kH and kW need to be sequential because of such cases.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -36,94 +137,180 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
...
+                      b.create<arith::MulIOp>(loc, n, stride[1]);
+                  Value rTimesDilation =
+                      b.create<arith::MulIOp>(loc, r, dilation[1]);
+                  Value indexW = b.create<arith::AddIOp>(loc, nTimesStride,
+                                                         rTimesDilation);
+                  Value input = b.create<tensor::ExtractOp>(
+                      loc, paddedInput, ValueRange{i, j, indexH, indexW});
+                  Value maxVal = args[0];
+                  Value pred = b.create<arith::CmpFOp>(
+                      loc, arith::CmpFPredicate::OEQ, input, maxVal);
```
cathyzhyi, line 287: Also need to check the output doesn't contain a valid index already.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -36,94 +137,180 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
...
+                      b.create<arith::SubIOp>(loc, indexH, padding[0]);
+                  Value indexWMinusPadding =
+                      b.create<arith::SubIOp>(loc, indexW, padding[1]);
+                  Value outIndex = b.create<arith::MulIOp>(
+                      loc, indexHMinusPadding, inputShapeW);
+                  outIndex = b.create<arith::AddIOp>(loc, outIndex,
+                                                     indexWMinusPadding);
+
+                  Value result = b.create<arith::SelectOp>(
+                      loc, pred, castIndexToInt(b, loc, outIndex), args[2]);
```
cathyzhyi, line 299: nit: assign args[2] to a variable with name to improve readability.  Can you put all the assignment at the start of the block? Similar to this code here https://github.com/llvm/torch-mlir/blob/5b9bdfaf3fb56c5aed12d689e21c9884b1c5a789/lib/Conversion/TorchToLinalg/Linear.cpp#L226.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -36,94 +137,180 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
...
-            elementType.cast<mlir::FloatType>().getFloatSemantics(),
-            /*Negative*/ true));
-    Value initValue = rewriter.create<arith::ConstantOp>(loc, initialAttr);
-    Value outTensorInitialized =
-        rewriter.create<linalg::FillOp>(loc, initValue, outTensor).getResult(0);
-
-    auto stridesAttr = rewriter.getI64VectorAttr(strideInts);
-    auto dilationAttr = rewriter.getI64VectorAttr(dilationInts);
+    SmallVector<Value> resultShape(getTensorSizes(rewriter, loc, maxPool2d));
+    Value indicesTensor = createZeroInitTensor(
```
cathyzhyi, line 221: Need to init the tensor with -1 for detecting cases where max value appear multiple times in the window.

cathyzhyi (changes requested):
More consideration needs to be given to cases maximum value appear multiple times in a window. I put some comments related to this.

lib/Conversion/TorchToLinalg/Pooling.cpp
ramiro050: Also, pass as constant reference to avoid making copies

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -27,6 +27,107 @@ using namespace mlir;
...
+      kernelSizeIntValues[0], strideIntValues[0]);
+  Value Wout = torch_to_linalg::getOutputDimForConvOps(
+      rewriter, loc, W, paddingIntValues[1], dilationIntValues[1],
+      kernelSizeIntValues[1], strideIntValues[1]);
+
+  // Initialize output tensor with smallest floating point value
+  Value outTensor = rewriter.create<linalg::InitTensorOp>(
+      loc, ValueRange{N, C, Hout, Wout}, elementType);
+  auto initialAttr = rewriter.getFloatAttr(
+      elementType, APFloat::getSmallest(
```
ramiro050, line 109: This isn't right. The naming of `getSmallest` is not very good at conveying its behavior, but what it does is that it returns the number of smallest *magnitude*. So in this case it returns something very close to `-0.0`. The right function is `getLargest`, which will return the largest negative number (in magnitude) possible. 

Note: I noticed that some of the tests do include negative and positive numbers; however, it seems that on average there is always a positive number around in each window to make things work out in the end. Can you add a test with only negative numbers to make sure this logic is right?

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -27,6 +27,107 @@ using namespace mlir;
...
+  Location loc = op.getLoc();
+  Type elementType = self.getType().cast<RankedTensorType>().getElementType();
+  if (!elementType.isa<mlir::FloatType>())
+    return op.emitError("unimplemented: non-floating point type");
+
+  SmallVector<int64_t, 4> paddingIncludingNC = {0, 0};
+  paddingIncludingNC.insert(paddingIncludingNC.end(), paddingInts.begin(),
+                            paddingInts.end());
+  Value paddedInput =
+      torch_to_linalg::getPaddedTensor(op, rewriter, self, paddingIncludingNC);
```
ramiro050, line 82: I don't think this is the right variant of `getPaddedTensor` to use. From the comment on it,

https://github.com/llvm/torch-mlir/blob/fa0b24a73ce4e403b077c8d894fe7a12d97303bb/lib/Conversion/TorchToLinalg/Utils.cpp#L56-L58

the function pads the tensors with zeros, which will result in the wrong maximums when dealing with negative tensors.

You should use this one instead to specify the fill value:

https://github.com/llvm/torch-mlir/blob/fa0b24a73ce4e403b077c8d894fe7a12d97303bb/lib/Conversion/TorchToLinalg/Utils.cpp#L37

ramiro050 (changes requested):
A few drive-by comments

lib/Conversion/TorchToLinalg/Pooling.cpp
cathyzhyi: @vivekkhandelwal1 can you help to change the helper name to `getZeroPaddedTensor`?

vivekkhandelwal1:
@cathyzhyi @ramiro050 Just a gentle reminder for your review.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -36,94 +137,196 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
...
       return failure();
-    Location loc = op->getLoc();
+
     Value self = adaptor.self();
-    Value ceilMode = adaptor.ceil_mode();
+    Value maxPool2d;
 
+    SmallVector<int64_t, 2> kernelSizeInts, strideInts, paddingInts,
+        dilationInts;
+    if (failed(checkAndGetPoolingParameters<AtenMaxPool2dOp>(
```
cathyzhyi, line 146: for `max_pool2d_with_indices` and `max_pool2d` the input can be 3D. We need to guard the 3D case if we don't want to support that.

cathyzhyi (changes requested):
Can we add a testcase when there are duplicating maximum value in a window? An input tensor with all ones would be good enough.

cathyzhyi (approved):

PR #519, opened 2022-01-11
* [tosa] Support for AtenPowTensorScalarOp with constant Scalar as input

anupgangwar:

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -635,6 +635,49 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
   }
 };
 
+// FIXME(AG): This will eventually go into a Tosa*Utils file
+// Convert an fp32 scalar into tosa fp32 tensor.
+static LogicalResult
+tosaF32TensorFromTorchFloat(ConversionPatternRewriter &rewriter, Operation *op,
+                            Value torchScalarValue, Value &tosaTensor) {
+  double scalarValue;
+
+  if (!matchPattern(torchScalarValue, m_TorchConstantFloat(&scalarValue)))
+    return failure();
+
+  // Construct a tosa.const
+  tosaTensor =
+      mlir::tosa::getTosaConstTensorSingleF32(rewriter, op, scalarValue);
+
+  return success();
+}
+
+template <>
+LogicalResult ConvertAtenOp<AtenPowTensorScalarOp>::matchAndRewrite(
+    AtenPowTensorScalarOp op, OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+
+  Value self = adaptor.self();
+  auto selfTy = self.getType().template cast<RankedTensorType>();
+
+  if (!selfTy)
+    return op.emitError("Only ranked tensor types supported in TOSA Pow");
+
+  if (!selfTy.getElementType().isa<mlir::FloatType>())
+    return op.emitError("Only floating-point datatype legalization supported");
+
+  Value expTensor;
+  Value expScalar = op.exponent();
+  if (failed(tosaF32TensorFromTorchFloat(rewriter, op.getOperation(), expScalar,
+                                         expTensor)))
+    return op.emitError("Currently only scalar constants are supported for "
+                        "conversion in TOSA Pow operation");
+
+  rewriter.replaceOpWithNewOp<tosa::PowOp>(
+      op, getTypeConverter()->convertType(op.getType()), self, expTensor);
+
+  return success();
+}
 } // namespace
 
 // -----------------------------------------------------------------------------
@@ -740,6 +783,7 @@ public:
     INSERT_ATENOP_PATTERN(AtenMulTensorOp);
     INSERT_ATENOP_PATTERN(AtenDivTensorOp);
     INSERT_ATENOP_PATTERN(AtenArgmaxOp);
+    INSERT_ATENOP_PATTERN(AtenPowTensorScalarOp);
 #undef INSERT_ATENOP_PATTERN
 
     if (failed(applyPartialConversion(getOperation(), target,
--- a/lib/Dialect/TorchConversion/Transforms/Passes.cpp
+++ b/lib/Dialect/TorchConversion/Transforms/Passes.cpp
@@ -12,6 +12,7 @@
 #include "mlir/Dialect/MemRef/Transforms/Passes.h"
 #include "mlir/Dialect/StandardOps/Transforms/Passes.h"
 #include "mlir/Dialect/Tosa/Transforms/Passes.h"
+#include "mlir/Conversion/Passes.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Transforms/Passes.h"
 #include "torch-mlir/Conversion/TorchToLinalg/TorchToLinalg.h"
@@ -100,6 +101,9 @@ void TorchConversion::createTorchBackendToTosaBackendPipeline(
     // The resolution of `dim` ops tends to create identical ops. CSE them.
     pm.addNestedPass<FuncOp>(createCSEPass());
   }
+
+  // Add the ToStandard pass for lowering some ops
+  pm.addNestedPass<FuncOp>(createTosaToStandard());
 
   // Finish the type conversion from `torch` types to the types of the
   // TOSA backend contract.
--- a/e2e_testing/torchscript/xfail_sets.py
+++ b/e2e_testing/torchscript/xfail_sets.py
@@ -50,4 +50,5 @@ TOSA_PASS_SET = {
     "SqueezeDimModule_identity",
     "SqueezeDimModule_unitDim",
     "ReturnTwoTensorF32I64_basic",
+    "ElementwisePowModule_basic",
 }
--- a/lib/Dialect/TorchConversion/Transforms/VerifyTosaBackendContract.cpp
+++ b/lib/Dialect/TorchConversion/Transforms/VerifyTosaBackendContract.cpp
@@ -44,6 +44,7 @@ class VerifyTosaBackendContractPass
     target.addLegalDialect<tosa::TosaDialect>();
     target.addDynamicallyLegalOp<tensor::CastOp>(opHasLegalTypes);
     target.addDynamicallyLegalOp<arith::ExtSIOp>(opHasLegalTypes);
+    target.addDynamicallyLegalOp<arith::ConstantOp>(opHasLegalTypes);
 
     RewritePatternSet patterns(context);
     if (failed(applyFullConversion(module, target, std::move(patterns)))) {
--- a/test/Conversion/TorchToTosa/basic.mlir
+++ b/test/Conversion/TorchToTosa/basic.mlir
@@ -330,3 +330,20 @@ func @torch.aten.minimum$basic(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.v
   %0 = torch.aten.minimum %arg0, %arg1 : !torch.vtensor<[?,?],f32>, !torch.vtensor<[?,?],f32> -> !torch.vtensor<[?,?],f32>
   return %0 : !torch.vtensor<[?,?],f32>
 }
+
+// -----
+
+// CHECK-LABEL:   func @torch.aten.pow.Tensor_Scalar$basic(
+// CHECK-SAME:                                             %[[VAL_0:.*]]: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[?,?],f32> {
+// CHECK:           %[[VAL_1:.*]] = torch_c.to_builtin_tensor %[[VAL_0]] : !torch.vtensor<[?,?],f32> -> tensor<?x?xf32>
+// CHECK:           %[[VAL_2:.*]] = torch.constant.float 3.123400e+00
+// CHECK:           %[[VAL_3:.*]] = "tosa.const"() {value = dense<3.123400e+00> : tensor<f32>} : () -> tensor<f32>
+// CHECK:           %[[VAL_4:.*]] = "tosa.pow"(%[[VAL_1]], %[[VAL_3]]) : (tensor<?x?xf32>, tensor<f32>) -> tensor<?x?xf32>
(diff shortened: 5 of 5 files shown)


anupgangwar:
Tagging @silvasean and @sjarus for merging this.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,56 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
...
 };
 
+// FIXME(AG): This will eventually go into a Tosa*Utils file
+// Convert an fp32 scalar into tosa fp32 tensor.
+static LogicalResult tosaF32TensorFromTorchFloat(ConversionPatternRewriter &rewriter,
+                                        Operation *op, Value torchScalarValue,
+                                        Value &tosaTensor) {
+  double scalarValue;
+  // Check if this is an unrealized conversion cast
+  auto uConvCast = dyn_cast<mlir::UnrealizedConversionCastOp>(torchScalarValue.getDefiningOp());
```
cathyzhyi, line 645: Hmm, why do we need to check `UnrealizedConversionCastOp` here? Do you have the IR dump?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,56 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
...
+  auto uConvCastInOps = uConvCast.inputs();
+  // Ensure that there is exactly one input
+  if (uConvCastInOps.front() != uConvCastInOps.back()) op->emitError("This cast must have exactly one operand");
+  auto torchScalarValue_ = uConvCast ? uConvCastInOps.front() : torchScalarValue;
+
+  if (!matchPattern(torchScalarValue_, m_TorchConstantFloat(&scalarValue)))
```
cathyzhyi, line 651: I am not familiar with Tosa but it's a bit surprising that it can only handle the case the exp value is constant. Is it possible to also consider non constant value here?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,56 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
...
 
+// FIXME(AG): This will eventually go into a Tosa*Utils file
+// Convert an fp32 scalar into tosa fp32 tensor.
+static LogicalResult tosaF32TensorFromTorchFloat(ConversionPatternRewriter &rewriter,
+                                        Operation *op, Value torchScalarValue,
+                                        Value &tosaTensor) {
+  double scalarValue;
+  // Check if this is an unrealized conversion cast
+  auto uConvCast = dyn_cast<mlir::UnrealizedConversionCastOp>(torchScalarValue.getDefiningOp());
+  auto uConvCastInOps = uConvCast.inputs();
```
cathyzhyi, line 646: nit: how about `uConvCastInputs`?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,56 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
...
+  // Ensure that there is exactly one input
+  if (uConvCastInOps.front() != uConvCastInOps.back()) op->emitError("This cast must have exactly one operand");
```
cathyzhyi, line 648: nit: I think `uConvCastInputs.size() != 1` would be easier to read.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,56 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
...
+
+  if (!selfTy)
+    return op.emitError("Only ranked tensor types supported in TOSA Pow");
+
+  Value expScalar = adaptor.exponent();
+  auto expScalarType = expScalar.getType();
+  if (!expScalarType.dyn_cast<mlir::FloatType>()) op.emitError("Only floating point supported for exponent");
+
+  Value expTensor;
+
```
cathyzhyi, line 676: nit: empty line could be removed.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,56 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
...
+  if (failed(tosaF32TensorFromTorchFloat(rewriter, op.getOperation(), expScalar, expTensor)))
+    return op.emitError("Currently only scalar constants are supported for conversion in TOSA Pow operation");
+
+  rewriter.replaceOpWithNewOp<tosa::PowOp>(
+    op,
+    OpConversionPattern<AtenPowTensorScalarOp>::getTypeConverter()->convertType(
```
cathyzhyi, line 682: nit: `OpConversionPattern<AtenPowTensorScalarOp>::` could be removed.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,56 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
...
+
+  Value self = adaptor.self();
+  auto selfTy = self.getType().template cast<RankedTensorType>();
+
+  if (!selfTy)
+    return op.emitError("Only ranked tensor types supported in TOSA Pow");
+
+  Value expScalar = adaptor.exponent();
+  auto expScalarType = expScalar.getType();
+  if (!expScalarType.dyn_cast<mlir::FloatType>()) op.emitError("Only floating point supported for exponent");
```
cathyzhyi, line 673: `clang-format` for formatting.

cathyzhyi: need `return` before `op.emitError`

lib/Conversion/TorchToTosa/TorchToTosa.cpp
cathyzhyi, line 648: the if block need to be in a new line. Please use `clang-format` to format the code.

cathyzhyi, line 648: need `return` before `op->emitError`

cathyzhyi (changes requested):
I did a pass but I am not familiar enough with Tosa. Could @sjarus help to review as well?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
sjarus: The tosa.pow operator expects two tensor parameters. The pow value could be broadcast. 

Torch offers a TorchTensor input and TorchScalar exponent here. Is there a simple construct to express the IntegerType MLIR::Value exponent into a scalar tensor, i.e. rank-0 TensorType with IntegerType element-type ?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
anupgangwar: If you pass the MLIR in basic.mlir for '@torch.aten.pow.Tensor_Scalar', then it leads to an UnrealizedConversionCastOp which has the value attached to it. I do not have the IR dump.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
cathyzhyi: hmm, linalg can broadcast one scalar value to a tensor. However, this is done in the TorchToLinalg level.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
cathyzhyi: oh, I think I understand now. There is no direct conversion from TorchType to TosaType so the `UnrealizedConversionCastOp` was inserted. 

However, to get the original unconverted Torch::FloatType scalar  you can use `op.exponent()` instead of using `adaptor.exponent();`. The operands got from adaptor are converted types and the operands got from the original Torch op are original torch types.

anupgangwar: Thanks, works.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
anupgangwar: Conversion cast it now removed.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
anupgangwar: @cathyzhyi I will mark this as resolved.

anupgangwar:
All feedback has been incorporated.

sjarus:
LGTM. I'll let Cathy take a look and then either of us can merge.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,46 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
...
+
+  Value self = adaptor.self();
+  auto selfTy = self.getType().template cast<RankedTensorType>();
+
+  if (!selfTy)
+    return op.emitError("Only ranked tensor types supported in TOSA Pow");
+
+  Value expTensor;
+  Value expScalar = op.exponent();
+  if (failed(tosaF32TensorFromTorchFloat(rewriter, op.getOperation(), expScalar, expTensor)))
```
cathyzhyi, line 667: nit: those 2 lines look very long. Can you use `clang-format` for formatting?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,46 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
   }
 };
 
+// FIXME(AG): This will eventually go into a Tosa*Utils file
+// Convert an fp32 scalar into tosa fp32 tensor.
+static LogicalResult tosaF32TensorFromTorchFloat(ConversionPatternRewriter &rewriter,
```
cathyzhyi, line 640: nit: need to run clang-format

cathyzhyi (approved):
LGTM other than formatting.
Thanks for the change!

lib/Conversion/TorchToTosa/TorchToTosa.cpp
anupgangwar: While I have run clang-format, the generated code is less readable than the above one liner above.

anupgangwar:
Ran clang-format. Can this be merged now?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -635,6 +635,46 @@ class ConvertAtenSqueezeAllDimsOp : public ConvertAtenSqueezeOp<AtenOpT> {
...
+  return success();
+}
+
+template <>
+LogicalResult ConvertAtenOp<AtenPowTensorScalarOp>::matchAndRewrite(
+    AtenPowTensorScalarOp op, OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+
+  Value self = adaptor.self();
+  auto selfTy = self.getType().template cast<RankedTensorType>();
```
cathyzhyi, line 661: I think we missed a type check `getElementType().isa<mlir::FloatType>()` here. sry, I missed this in the previous review.

anupgangwar: 'tosaF32TensorFromTorchFloat' will fail if the 'exponent' does not parse as a float.

cathyzhyi: We'd prefer an assert to make it explicit only floating point is supported. Failure in `tosaF32TensorFromTorchFloat` is a symptom of passing in wrong input type but doesn't really indicate the root cause with out some level of investigation.

lib/Dialect/TorchConversion/Transforms/Passes.cpp
```diff
@@ -101,6 +102,9 @@ void TorchConversion::createTorchBackendToTosaBackendPipeline(
     pm.addNestedPass<FuncOp>(createCSEPass());
   }
 
+  // Add the ToStandard pass for lowering some ops
+  pm.addNestedPass<FuncOp>(createTosaToStandard());
```
ljfitz, line 106: Linking is failing for me with this change. Shouldn't `MLIRTosaToStandard` be added to `LINK_LIBS` of `lib/Dialect/TorchConversion/Transforms/CMakeLists.txt`?

anupgangwar: Can you tell me which target you are trying to build? 'check-torch-mlir' did not need this and all the checks here passed as well. @vivekkhandelwal1 created Issue #521 to track this.

ljfitz: I see it for `ninja bin/torch-mlir-lsp-server` - the same as #521. The `lib/Dialect/TorchConversion/Transforms/CMakeLists.txt` file should really list all the libraries it directly depends on so that linking is deterministic and not dependent on the chance that `cmake` will build the lining order in a way that happens to link fully. If you add `MLIRTosaToStandard` to `LINK_LIBS` you will see that `cmake` adds `libMLIRTosaToStandard.a` after `libTorchMLIRTorchConversionPasses.a`. As in this link `libTorchMLIRTorchConversionPasses.a` is the first to require `createTosaToStandard()`, this order is required. I know `libMLIRTosaToStandard.a`  appears before `libTorchMLIRTorchConversionPasses.a` too, but the linker won't look into preceding `.a` files for newly unresolved symbols.

anupgangwar: For now, I have reverted the change to Passes.cpp. Will add it to <Transforms>/CMakeLists.txt when we renable this.

PR #524, opened 2022-01-14
[tosa] Add tosa-to-standard before tosa-to-linalg pass

sjarus:


sjarus:
This pass is needed because TOSA-to-LinAlg depends on TOSA and Standard dialect content, with tosa.const being converted to arith.constants by tosa-to-standard. Without it, TOSA-to-LinAlg fails stating it can't recognize tosa.const() . This update fixes @anupgangwar's earlier AtenPowOp implementation in the process.

cathyzhyi (approved):
LGTM

PR #458, opened 2021-12-03
[tosa] Add Torch reduction operators

sjarus:
- Supports variants with multiple dims, one dim, all dime
- Leverages legalize_common and legalize_utils code from
TensorFlow-TOSA work

Signed-off-by: Suraj Sudhir <suraj.sudhir@arm.com>

PR #525, opened 2022-01-14
[tosa] Add more common utility functions

sjarus:
- Common code as TF repository, being moved to MLIR core.
- Will support further legalizations to be published.


sjarus:
Adds more of the existing legalization utility code originally set up in https://github.com/llvm/torch-mlir/pull/458. 

These are common code from the TF->TOSA legalization path with no changes. They support TOSA legalization construction, and will in near future move to MLIR core as due to this RFC: https://llvm.discourse.group/t/rfc-moving-tosa-construction-lib-into-mlir-core/4961

sjarus:
Build and checks are clean, proceeding to merge so I can unblock @anupgangwar .

PR #527, opened 2022-01-17
Fix op signatures to follow latest torch nightly version.

gprateek93:


cathyzhyi:
The changes in python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py would be very hard for other people to maintain. The commit history of this file would also be confusing. Shall we move the emitting ops into a separate python file that can be autogenerated from a script so that if the ops signature changes upstream people could just run a script and update the ops?

cathyzhyi (changes requested):
The changes in python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py would be very hard for other people to maintain. The commit history of this file would also be confusing. Shall we move the emitting ops into a separate python file that can be autogenerated from a script so that if the ops signature changes upstream people could just run a script and update the ops?

cathyzhyi:
Sry, I needed to revert #520 in #529. Could you reenable it when delivering this PR?

PR #528, opened 2022-01-17
Importing `linalg_ext` dialect

cathyzhyi:


silvasean:
Before we do this, I'd like for us to reflect to the community more the thinking that has gone into adding it.

I'm also really really hesitant at this point to add anything IREE-specific. Arguably linalg_ext is not, but I see IREEInput dialect being added here as well.

stellaraccident:
Naming aside, I think everything in there is in some state of "would be good to have upstream" if we had places for it (or more mileage). I personally would look at this pragmatically: these are things that may not be upstream for a long time, since all of this is so new, but some form of it should be to have a full story. I suspect that using it might help converge some of that. The name here is more an artifact of things needing a unique name than an indication that this is an "iree thing". The best case here is we get some more proofs on some of this while we're all still experimental and there is more pull to define the upstream story.

stellaraccident:
Also, as an aside, we newly have even the Tensorflow frontends slimmed down to only depend on things from here. This really does represent things we have pragmatically found that mlir is missing for building frontends but that needs more consensus to move on.

silvasean:
+1 on all the comments Stella, but I really want to avoid outsiders seeing this as an "IREE thing", even if "IREE" basically just a uuid in this context (perception matters here, sadly, which is why we deleted the iree input dialect in the first place). I wonder if we can grab a generic name like "mlutil" or something for these (please suggest better name :) ). Ignoring that, does anybody know why we need the iree input dialect here at all? Shouldn't linalg_ext be standalone?

cathyzhyi:
> Ignoring that, does anybody know why we need the iree input dialect here at all? Shouldn't linalg_ext be standalone?

I copy & pasted from another place and can remove the inputDialect.

cathyzhyi:
`mlutils` sounds good to me. I can replace all the `iree` with `mlutils` where it's possible.

stellaraccident:
Feel free to add your own standalone dialect with ops that match what you want, and i can copy it down to iree and convert in. Just promise me that whatever you do has a chance of going upstream eventually, and please make sure it follows llvm style, is simple and easily copyable. 

At some point, we do need this stuff upstream.

PR #529, opened 2022-01-18
Update external/llvm-project

cathyzhyi:
- Add `qualified` to ods because of https://reviews.llvm.org/D113873 and https://reviews.llvm.org/D116905
- Needed to revert https://github.com/llvm/torch-mlir/pull/520 as it was based on an old torch version. https://github.com/llvm/torch-mlir/pull/527 will bring this back with a better design.
- Change ConvertAtenCatOp to use more accurate tensor shape info and as much static info as possible to pass `tensor.insert_slice` verification code added by https://reviews.llvm.org/D114715
- Other minor fixes


pashu123 (approved):
LGTM!

PR #533, opened 2022-01-21
Adding E2E tests for histogram binning calibration

nirvedhmeshram:


pashu123:
NIT- Python format is required.

nirvedhmeshram:
> NIT- Python format is required.

fixed

pashu123 (approved):
LGTM!

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,103 @@
...
+        self.positive_weight = torch.tensor([0.4])
+        self.bin_ctr_in_use_after = 0
+        self.bin_ctr_weight_value = 0.9995
+        self.oneminusbin_ctr_weight_value = 0.0005
+        self._iteration = 0
+
+    @export
+    @annotate_args([
+        None,
+        ([5000], torch.int32, True),
```
cathyzhyi, line 47: Any reason static shape is used here?

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,103 @@
...
+        isvalid = torch.gt(curr_bin_num_examples,
+                           self.bin_ctr_in_use_after)  # GT
+        calibrated_prediction_data = torch.where(
+            isvalid, curr_segment_value, origin_prediction.float())  # Conditional
+        return calibrated_prediction_data, bin_ids_data
+
+
+@register_test_case(module_factory=lambda: HistogramBinningCalibrationByFeature())
+def HBC_staticsize(module, tu: TestUtils):
+    num_bins: int = 5000
```
cathyzhyi, line 86: Unused variable here.

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,103 @@
+
```
silvasean, line 1: nit: remove space

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,103 @@
...
+    segment_offsets: Tensor = torch.cat(
+        (torch.tensor([0]), segment_offsets), 0)
+    num_values: int = int(torch.sum(segment_lengths).item())
+    segment_values: Tensor = torch.randint(
+        0,
+        num_segments,
+        (num_values,),
+    )
+    segment_values = torch.cat(
+        (segment_values, torch.zeros(5000-segment_values.numel())), 0)
```
ramiro050, line 102: nit: is this 5000 related to `num_bins` or `num_logits`? If so, it would be nice to use the variable

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,103 @@
...
+        dense_segment_value = torch.zeros(logit.numel(), dtype=torch.int32)
+        validoffsets = torch.gt(
+            segment_lengths[1:self._num_bins+1], segment_lengths[0:self._num_bins])
+        gathered_segment_values = (
+            segment_value[segment_lengths[0:self._num_bins].long()]+1).int()
+        dense_segment_value = torch.where(
+            validoffsets, gathered_segment_values, dense_segment_value)
+        zeros = torch.empty_like(
+            dense_segment_value, dtype=torch.int32).fill_(0)  # ConstantFill
+        isnotvalid = torch.gt(dense_segment_value, self._num_segments)  # GT
```
silvasean, line 63: is this autogenerated? can you mention how it is generated?

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,103 @@
...
+        calibrated_prediction_data = torch.where(
+            isvalid, curr_segment_value, origin_prediction.float())  # Conditional
+        return calibrated_prediction_data, bin_ids_data
+
+
+@register_test_case(module_factory=lambda: HistogramBinningCalibrationByFeature())
+def HBC_staticsize(module, tu: TestUtils):
+    num_bins: int = 5000
+    num_segments: int = 42
+    num_logits = 5000
```
ramiro050, line 88: These variables are also inside the `module`. It might be better to make the `module` variables public (i.e. change their name from `_num_segments` -> `num_segments`, then here you can just do `module.num_segments`. That way we are not defining the same hard coded values multiple times

ramiro050 (changes requested):

e2e_testing/torchscript/histogram_binning_calibration.py
nirvedhmeshram: I used autopep8 
` autopep8 -i e2e_testing/torchscript/histogram_binning_calibration.py`

e2e_testing/torchscript/histogram_binning_calibration.py
nirvedhmeshram: Yes this was because with dynamic shapes an older version of the op was having trouble running correctly on IREE with CUDA backend. It was making a grid of 79x64 with static shapes but with dynamic shapes the grid was 5000x64.

silvasean: Our test cases shouldn't be influenced by the details of specific backends. So I think that making it dynamic would be better.

e2e_testing/torchscript/histogram_binning_calibration.py
nirvedhmeshram: Should I mention how I generated the  formatting as a comment in the file or is this resolved?

e2e_testing/torchscript/histogram_binning_calibration.py
nirvedhmeshram: Nice suggestion. I am however getting an error when trying this. 
```TypeError: randint() received an invalid combination of arguments - got (int, _Tracer, tuple), but expected one of:```
This is the offending line
```    segment_values: Tensor = torch.randint(
        0,
        module.num_segments,
        (num_values,),
    )```

Comment (ramiro050) [reply e2e_testing/torchscript/histogram_binning_calibration.py:86]: Hm, I see. We'll probably have to leave it as is then

Comment (cathyzhyi) [reply e2e_testing/torchscript/histogram_binning_calibration.py:86]: +1 on avoiding same hardcoded number in multiple places. As a compromise we could still define a global variable which is hardcoded as 5000 and let both places to refer to that.

Comment (nirvedhmeshram) [reply e2e_testing/torchscript/histogram_binning_calibration.py:86]: actually I think I understand, i will assign from the global variable to a member variable of the object. I will check it out to make sure the values make it into the IR

File: e2e_testing/torchscript/histogram_binning_calibration.py (hunk 5 above)
Comment (nirvedhmeshram) [reply e2e_testing/torchscript/histogram_binning_calibration.py:68]: Ahh, this is behind some layers that I dont have access to but I manually edited a lot of the original script anyway. @powderluv do you want to ellaborate on this?

**Comment by nirvedhmeshram** [comment]
I am going to change this commit to add another case with static size. We are essentially trying to support the FBGEMM benchmark mentioned here https://github.com/pytorch/FBGEMM/blob/main/fbgemm_gpu/bench/histogram_binning_calibration_benchmark.py#L95-L98
Therefore we can have a dynamic case and the static case that is of specific interest for this op

File: e2e_testing/torchscript/histogram_binning_calibration.py (hunk 5 above)
Comment (powderluv) [reply e2e_testing/torchscript/histogram_binning_calibration.py:68]: That is nothing special and is just comments. Here is the source of the source file from @harsh-nod :

https://github.com/pytorch/FBGEMM/pull/804/files

Comment (silvasean) [reply e2e_testing/torchscript/histogram_binning_calibration.py:68]: Can we remove those autogenerated comments at least? I don't think they add much value.

Comment (nirvedhmeshram) [reply e2e_testing/torchscript/histogram_binning_calibration.py:68]: Yes, I can do that

**Comment by nirvedhmeshram** [comment]
@silvasean @cathyzhyi Please let me know if you have any feedback for this PR

File: e2e_testing/torchscript/histogram_binning_calibration.py [hunk 7]
```diff
@@ -0,0 +1,190 @@
...
+    segment_values: Tensor = torch.randint(
+        0,
+        num_segments,
+        (num_values,),
+    )
+    segment_values = torch.cat(
+        (segment_values, torch.zeros(num_logits-segment_values.numel())), 0)
+    return segment_values.int(), segment_offsets.int(), input_data_cpu
+
+# HistogramBinningCalibrationByFeature
```
silvasean, line 39: nit: remove this comment

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,190 @@
...
+@register_test_case(module_factory=lambda: HistogramBinningCalibrationByFeature())
+def HBC_basic(module, tu: TestUtils):
+    segment_value, segment_lengths, logit = make_HBC_inputs()
+    module.forward(segment_value, segment_lengths, logit)
+
+# This is a specialized case to support the FBGEMM input size of 5000
+# https://github.com/pytorch/FBGEMM/blob/main/fbgemm_gpu/bench/histogram_binning_calibration_benchmark.py
+
+
+class HistogramBinningCalibrationByFeatureStaticSize(torch.nn.Module):
```
silvasean, line 119: Based on the prior discussion, it seems that the only reason to have the static shape test was a bug in IREE. I don't think that's a strong enough reason to keep a duplicated test in our test suite. So can you remove the static shape version of the test?

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,190 @@
...
+import torch
+
+from torch_mlir_e2e_test.torchscript.framework import TestUtils
+from torch_mlir_e2e_test.torchscript.registry import register_test_case
+from torch_mlir_e2e_test.torchscript.annotations import annotate_args, export
+
+
+# ==============================================================================
+# define global variables
+num_segments = 42
```
silvasean, line 15: nit: NUM_SEGMENTS

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,190 @@
...
+                           self.bin_ctr_in_use_after)
+        calibrated_prediction_data = torch.where(
+            isvalid, curr_segment_value, origin_prediction.float())
+        return calibrated_prediction_data, bin_ids_data
+
+
+@register_test_case(module_factory=lambda: HistogramBinningCalibrationByFeature())
+def HBC_basic(module, tu: TestUtils):
+    segment_value, segment_lengths, logit = make_HBC_inputs()
+    module.forward(segment_value, segment_lengths, logit)
```
silvasean, line 113: you might want to mention the actual shapes here in a comment, since it isn't obvious from the code.

silvasean (changes requested):

e2e_testing/torchscript/histogram_binning_calibration.py
silvasean: I would recommend that downstreams that really care about this test running on IREE carry a local patch to make this statically shaped until the iree issues are resolved.

nirvedhmeshram: Thanks for the feedback. Sounds like a good plan.

silvasean (approved):

e2e_testing/torchscript/histogram_binning_calibration.py
```diff
@@ -0,0 +1,104 @@
...
+
+import torch
+
+from torch_mlir_e2e_test.torchscript.framework import TestUtils
+from torch_mlir_e2e_test.torchscript.registry import register_test_case
+from torch_mlir_e2e_test.torchscript.annotations import annotate_args, export
+
+
+# ==============================================================================
+# define global variables
```
silvasean, line 14: nit: `# Global parameters.`

silvasean (approved):

nirvedhmeshram:
Looks like all feedback is incorporated. I will go ahead and merge. Thank you.

PR #537, opened 2022-01-25
Add decomposition/lowering of `aten.max.dim` op.

pashu123:
In order to facilitate lowering of `aten.max.dim` op, a custom op i.e.,
`cutom.aten.max.dim` is introduced which makes the decomposition easy.

### DIFF
--- a/e2e_testing/torchscript/reduction.py
+++ b/e2e_testing/torchscript/reduction.py
@@ -122,3 +122,97 @@ class ReduceMeanDtypeModule(torch.nn.Module):
 @register_test_case(module_factory=lambda: ReduceMeanDtypeModule())
 def ReduceMeanDtypeModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 4, 5).to(torch.float64))
+
+# ==============================================================================
+
+class ReduceMaxAlongDim(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float64, True),
+    ])
+    def forward(self, a):
+        return torch.ops.aten.max(a, 1)[0]
+
+
+@register_test_case(module_factory=lambda: ReduceMaxAlongDim())
+def ReduceMaxAlongDim_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5).to(torch.float64))
+
+# ==============================================================================
+
+class ReduceMaxAlongDimNegative(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float64, True),
+    ])
+    def forward(self, a):
+        return torch.ops.aten.max(a, 1)[0]
+
+
+@register_test_case(module_factory=lambda: ReduceMaxAlongDimNegative())
+def ReduceMaxAlongDimNegative_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5, low=-10, high=10).to(torch.float64))
+
+# ==============================================================================
+
+class ReduceMaxKeepDim(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float64, True),
+    ])
+    def forward(self, a):
+        return torch.ops.aten.max(a, 1, keepdim=True)[1]
+
+
+@register_test_case(module_factory=lambda: ReduceMaxKeepDim())
+def ReduceMaxKeepDim_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5).to(torch.float64))
+
+# ==============================================================================
+
+class ReduceMaxKeepDimReturnBoth(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, a):
+        return torch.ops.aten.max(a, 1, keepdim=True)
+
+@register_test_case(module_factory=lambda: ReduceMaxKeepDimReturnBoth())
+def ReduceMaxKeepDimReturnBoth_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5, low=-10, high=-5))
+
+# ==============================================================================
+
+class ReduceMaxAllDims(torch.nn.Module):
+
+  def __init__(self):
+    super().__init__()
+
+  @export
+  @annotate_args([
+      None,
+      ([-1, -1, -1], torch.float32, True),
+  ])
+  def forward(self, a):
+    return torch.ops.aten.max(a)
+
+@register_test_case(module_factory=lambda: ReduceMaxAllDims())
+def ReduceMaxAllDims_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5, low=-10, high=-5))
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -2126,6 +2126,13 @@ static Value createLinalgNeutralElementForReduceOp(OpBuilder &b, Location loc,
   if (isa<AtenSumOp, AtenSumDimIntListOp>(op) &&
       elementType.isa<mlir::FloatType>())
     return b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
+  if (isa<AtenMaxOp>(op) && elementType.isa<mlir::FloatType>())
+    return b.create<arith::ConstantOp>(
+        loc, b.getFloatAttr(
+                 elementType,
+                 APFloat::getLargest(
+                     elementType.cast<mlir::FloatType>().getFloatSemantics(),
+                     /*Negative=*/true)));
 
   op->emitError("unimplemented lowering in "
                 "createLinalgNeutralElementForReduceOp");
@@ -2141,6 +2148,11 @@ static Value createLinalgPayloadCalculationForReduceOp(
         convertScalarToDtype(b, loc, payloadArgs[0], resultElementType);
     Value result = payloadArgs[1];
     return b.create<arith::AddFOp>(loc, self, result);
+  } else if (isa<AtenMaxOp>(op) && resultElementType.isa<mlir::FloatType>()) {
+    Value self =
+        convertScalarToDtype(b, loc, payloadArgs[0], resultElementType);
+    Value result = payloadArgs[1];
+    return b.create<arith::MaxFOp>(loc, self, result);
   }
   op->emitError("unimplemented lowering in "
                 "createLinalgPayloadCalculationForReduceOp");
@@ -2148,68 +2160,56 @@ static Value createLinalgPayloadCalculationForReduceOp(
 }
 
 namespace {
-// Aten argmax lowering represents the ArgMax op as an linalg.indexed_generic
+// Aten maxdim lowering represents the MaxDim op as an linalg.indexed_generic
 // op, producing two output buffers.
 //
-// The first output buffer contains the index of the found maximum value. It is
-// initialized to 0 and is resulting integer type.
+// The first output buffer contains the maximum value found. It is initialized
+// to the minimum representable value of the input element type.
 //
-// The second output buffer contains the maximum value found. It is initialized
-// to the minimum representable value of the input element type. After being
-// populated by indexed_generic, this buffer is disgarded as only the index is
-// requested.
+// The second output buffer contains the index of the found maximum value. It is
+// initialized to 0 and is resulting integer type.
 //
 // The indexed_generic op updates both the maximum value and index if the
 // current value exceeds the running max.
-class ConvertAtenArgmaxOp : public OpConversionPattern<AtenArgmaxOp> {
+class ConvertAtenMaxDimOp : public OpConversionPattern<AtenMaxDimOp> {
 public:
-  using OpConversionPattern<AtenArgmaxOp>::OpConversionPattern;
+  using OpConversionPattern<AtenMaxDimOp>::OpConversionPattern;
 
   LogicalResult
-  matchAndRewrite(AtenArgmaxOp argmaxOp, OpAdaptor adaptor,
+  matchAndRewrite(AtenMaxDimOp maxDimOp, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
 
-    Location loc = argmaxOp.getLoc();
+    Location loc = maxDimOp.getLoc();
     Value input = adaptor.self();
-    RankedTensorType resultType =
+    RankedTensorType valResultType =
         getTypeConverter()
-            ->convertType(argmaxOp.getResult().getType())
+            ->convertType(maxDimOp.getResult(0).getType())
+            .cast<RankedTensorType>();
+    RankedTensorType idxResultType =
+        getTypeConverter()
+            ->convertType(maxDimOp.getResult(1).getType())
             .cast<RankedTensorType>();
     RankedTensorType inputType = input.getType().cast<RankedTensorType>();
-    Type outElementType = resultType.getElementType();
-    if (!outElementType.isa<IntegerType>())
+    Type idxElementType = idxResultType.getElementType();
+    if (!idxElementType.isa<IntegerType>())
       return rewriter.notifyMatchFailure(
-          argmaxOp,
-          "aten.arg_max to linalg.* requires integer-like result type");
+          maxDimOp,
+          "aten.max_dim to linalg.* requires integer-like result type");
 
     bool keepDim = false;
-    if (!matchPattern(argmaxOp.keepdim(), m_TorchConstantBool(&keepDim)))
+    if (!matchPattern(maxDimOp.keepdim(), m_TorchConstantBool(&keepDim)))
       return failure();
 
     int64_t dim;
-    if (!matchPattern(argmaxOp.dim(), m_TorchConstantInt(&dim))) {
-      if (!argmaxOp.dim().getType().isa<Torch::NoneType>())
-        return rewriter.notifyMatchFailure(
-            argmaxOp,
-            "aten.arg_max to linalg.* requires int or NoneType value for Dim");
-      // For pytorch, if the value of Dim is None, argmax
-      // returns the index of the max value of the flattened input tensor,
-      // so here we flatten the input tensor.
-      SmallVector<ReassociationIndices> reassociation(1);
-      for (auto i : llvm::seq<int64_t>(0, inputType.getRank()))
-        reassociation[0].push_back(i);
-      input = rewriter.create<tensor::CollapseShapeOp>(argmaxOp->getLoc(),
-                                                       input, reassociation);
-      // Becomes 0 for flattened tensor.
-      dim = 0;
-      // Recast to fix shape.
-      inputType = input.getType().cast<RankedTensorType>();
-    }
+    if (!matchPattern(maxDimOp.dim(), m_TorchConstantInt(&dim)))
+      return rewriter.notifyMatchFailure(
+          maxDimOp, "aten.max_dim to linalg.* requires int value for Dim");
+
     Type inElementType = inputType.getElementType();
     if (!inElementType.isa<mlir::FloatType>()) {
       return rewriter.notifyMatchFailure(
-          argmaxOp,
-          "aten.arg_max to linalg.* requires Float input element type");
+          maxDimOp,
+          "aten.max_dim to linalg.* requires Float input element type");
     }
 
     // Constant op to account for the reduction along dim.
@@ -2224,7 +2224,7 @@ public:
     }
     // First fill the output buffer for the index.
     Value filledTensorIdx =
-        createZeroInitTensor(rewriter, loc, resultShape, outElementType);
+        createZeroInitTensor(rewriter, loc, resultShape, idxElementType);
 
     // Second fill the output buffer for the running max.
     Value initTensorMax =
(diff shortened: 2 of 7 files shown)


lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2583,6 +2596,17 @@ struct ConvertReductionOp : ConversionPattern {
         if (dim < inputType.getRank())
           dimSet.insert(dim);
       }
+    } else if (auto maxDimOp = dyn_cast<CustomAtenMaxDimOp>(op)) {
+      auto tensorOperand = operands[0];
+      auto inputType = tensorOperand.getType().cast<RankedTensorType>();
+      if (!matchPattern(maxDimOp.keepdim(), m_TorchConstantBool(&keepDim)))
+        return failure();
```
ramiro050, line 2603: nit: use `notifyMatchFailure` to include an error message

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -154,3 +154,20 @@ func @torch.aten.arange.start() -> !torch.vtensor<[?],si64> {
...
+// CHECK-LABEL: func @decompose_aten_max_dim
+// CHECK-SAME:              (%[[INP:.*]]: !torch.vtensor<[?,?,?],f64>) -> !torch.vtensor<[?,1,?],si64> {
+// CHECK:      %[[CST1:.*]] = torch.constant.int 1
+// CHECK:      %[[TRUE:.*]] = torch.constant.bool true
+// CHECK:      %[[MAX_DIM:.*]] = torch.custom.aten.max.dim %[[INP]], %[[CST1]], %[[TRUE]] :
+// CHECK-SAME: !torch.vtensor<[?,?,?],f64>, !torch.int, !torch.bool -> !torch.vtensor<[?,1,?],f64>
+// CHECK:      %[[ARG_MAX:.*]] = torch.aten.argmax %[[INP]], %[[CST1]], %[[TRUE]] : 
+// CHECK-SAME: !torch.vtensor<[?,?,?],f64>, !torch.int, !torch.bool -> !torch.vtensor<[?,1,?],si64>
+// CHECK:      return %[[ARG_MAX]] : !torch.vtensor<[?,1,?],si64>
+func @decompose_aten_max_dim(%arg0: !torch.vtensor<[?,?,?],f64>) -> !torch.vtensor<[?,1,?],si64> {
```
ramiro050, line 168: nit: the function name does not match the style used in this file

e2e_testing/torchscript/reduction.py
```diff
@@ -122,3 +122,63 @@ def forward(self, a):
...
+  def __init__(self):
+    super().__init__()
+
+  @export
+  @annotate_args([
+      None,
+      ([-1, -1, -1], torch.float64, True),
+  ])
+  def forward(self, a):
+    return torch.max(a, 1)[0]
```
ramiro050, line 139: Is there a reason we can't return the two tensors?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -354,6 +354,15 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
       Type dtype = operands[0]->getValue().dtype;
       return visitReductionAlongDimIntOp(anyDim, anyDim.dim(), anyDim.keepdim(),
                                          dtype, operands);
+    } else if (auto maxDim = dyn_cast<AtenMaxDimOp>(op)) {
+      Type firstResDtype = operands[0]->getValue().dtype;
+      Type secResDtype =
```
ramiro050, line 359: nit: I would spell out `second` fully to match the `first` variable

ramiro050 (commented):
Good job! I just have a few comments and questions

cathyzhyi (approved):

e2e_testing/torchscript/reduction.py
cathyzhyi, line 139: nit: It's better to use `torch.ops.aten.max.dim` directly so that it's more obvious that's the variant being tested.

pashu123: We don't have the support for the combination of memref<f64> and memref<i64> in the refbackend.  We do have memref of f32 and i64 which I have added as a third test case.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2125,6 +2125,13 @@ static Value createLinalgNeutralElementForReduceOp(OpBuilder &b, Location loc,
   if (isa<AtenSumOp, AtenSumDimIntListOp>(op) &&
       elementType.isa<mlir::FloatType>())
     return b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
+  if (isa<CustomAtenMaxDimOp>(op) && elementType.isa<mlir::FloatType>())
+    return b.create<arith::ConstantOp>(
+        loc, b.getFloatAttr(
+                 elementType,
+                 APFloat::getSmallest(
+                     elementType.cast<mlir::FloatType>().getFloatSemantics(),
+                     true)));
```
cathyzhyi, line 2134: nit: comment /*Negative=*/

include/torch-mlir/Dialect/Torch/IR/CustomAtenOps.td
```diff
@@ -0,0 +1,34 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+// This file contains custom aten ops to facilitate easy decomposition and
+// should be used conservatively.
+//===----------------------------------------------------------------------===//
+
+def Torch_CustomAtenMaxDimOp
```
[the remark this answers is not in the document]
pashu123: I am worried about the third point explained here: https://pytorch.org/docs/stable/generated/torch.amax.html?highlight=max#torch.amax. Rest I think is fine.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2583,6 +2597,17 @@ struct ConvertReductionOp : ConversionPattern {
...
       }
+    } else if (auto maxDimOp = dyn_cast<CustomAtenMaxDimOp>(op)) {
+      auto tensorOperand = operands[0];
+      auto inputType = tensorOperand.getType().cast<RankedTensorType>();
+      if (!matchPattern(maxDimOp.keepdim(), m_TorchConstantBool(&keepDim)))
+        return failure();
+      int64_t dim;
+      if (!matchPattern(maxDimOp.dim(), m_TorchConstantInt(&dim)))
+        return failure();
+      dim = dim >= 0 ? dim : dim + inputType.getRank();
```
cathyzhyi, line 2608: nit: use toPositiveDim and isValidDim helpers

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -489,6 +489,34 @@ class DecomposeAtenMeanOp : public OpRewritePattern<AtenMeanOp> {
...
+    Value keepDim = op.keepdim();
+    Value values = op.values();
+    Value indices = op.indices();
+    BaseTensorType valuesTensorType = values.getType().cast<BaseTensorType>();
+    BaseTensorType indicesTensorType = indices.getType().cast<BaseTensorType>();
+
+    Value maxResult = rewriter.create<CustomAtenMaxDimOp>(loc, valuesTensorType,
+                                                          input, dim, keepDim);
+    Value indicesResult = rewriter.create<AtenArgmaxOp>(loc, indicesTensorType,
+                                                        input, dim, keepDim);
```
cathyzhyi, line 513: It's not efficient to decompose to `CustomAtenMaxDimOp` and `AtenArgmaxOp` since the second op calculates both max values and indices in one `linalg.generic` and drop the max value result. The calculation for `CustomAtenMaxDimOp` is redundant. It's probably better to decompose `AtenArgmaxOp` into `aten.max.dim` instead.

pashu123: I didn't get it. The second result of `aten.max.dim` op actually calculates `argmax` and will be required.

silvasean:
I'd generally like to avoid adding custom ops like this unless absolutely needed. And in this case, it's fairly redundant to decompose it into a max + argmax (both are O(N)). It would be better to decompose it into something like:

```
index = argmax(t)
value = t[index]
```
or something like that.

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -154,3 +154,20 @@ func @torch.aten.arange.start() -> !torch.vtensor<[?],si64> {
...
+// CHECK-LABEL: func @decompose.aten.max.dim
+// CHECK-SAME:              (%[[INP:.*]]: !torch.vtensor<[?,?,?],f64>) -> !torch.vtensor<[?,1,?],si64> {
+// CHECK:      %[[CST1:.*]] = torch.constant.int 1
+// CHECK:      %[[TRUE:.*]] = torch.constant.bool true
+// CHECK:      %[[MAX_DIM:.*]] = torch.custom.aten.max.dim %[[INP]], %[[CST1]], %[[TRUE]] :
+// CHECK-SAME: !torch.vtensor<[?,?,?],f64>, !torch.int, !torch.bool -> !torch.vtensor<[?,1,?],f64>
+// CHECK:      %[[ARG_MAX:.*]] = torch.aten.argmax %[[INP]], %[[CST1]], %[[TRUE]] : 
+// CHECK-SAME: !torch.vtensor<[?,?,?],f64>, !torch.int, !torch.bool -> !torch.vtensor<[?,1,?],si64>
+// CHECK:      return %[[ARG_MAX]] : !torch.vtensor<[?,1,?],si64>
+func @decompose.aten.max.dim(%arg0: !torch.vtensor<[?,?,?],f64>) -> !torch.vtensor<[?,1,?],si64> {
```
ramiro050, line 168: nit: the convention in this file is to name the function after the op that is being decomposed. So it should be: `@torch.aten.max.dim`

include/torch-mlir/Dialect/Torch/IR/CustomAtenOps.td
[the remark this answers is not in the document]
ramiro050: I'm gonna resolve this since Yi and Sean have suggested better approaches for decomposition

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ramiro050: This might be helpful:

https://github.com/llvm/torch-mlir/blob/cadea678e5e2cfc62a1625af6e50cac34144e710/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp#L2156-L2159

Its better to use the switch the ops `Argmax` and `MaxDim`. `MaxDim` can be lowered in `TorchToLinalg` using the code that is already there for `Argmax`, and `Argmax` can be moved to the `DecomposeComplexOps` pass and decompose into `MaxDim`

silvasean: I like this, because then we can special-case when the "indices" have no uses and lower to a regular max reduction, because that can be lowered more efficiently.

cathyzhyi: The current implementation of `Argmax` can be mostly reused as it calculates max values and indices already but dropped the max values https://github.com/llvm/mlir-npcomp/blob/d6b6c0268cfd6ead0aa5f2f21ba02bbe8ead3180/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp#L2237.

pashu123: @silvasean @ramiro050 @cathyzhyi   We can't decompose aten_arg_max to aten_max_dim because the dim can be NoneType in the former(which means reduce along all dims). I have converted the argMaxOp rewrite pattern into a static function which both `argMax` as well as `maxDim` op can call.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2140,162 +2148,225 @@ static Value createLinalgPayloadCalculationForReduceOp(
...
-// populated by indexed_generic, this buffer is disgarded as only the index is
-// requested.
-//
-// The indexed_generic op updates both the maximum value and index if the
-// current value exceeds the running max.
-class ConvertAtenArgmaxOp : public OpConversionPattern<AtenArgmaxOp> {
+static std::pair<Value, Value>
+createArgMaxKernel(OpBuilder &b, Location loc, Value input,
+                   Optional<int64_t> dim, bool keepDim,
+                   RankedTensorType idxResType, RankedTensorType valResType) {
```
ramiro050, line 2171: nit: I think its better to flip the order of these two result types so that they match the order of the types of the output

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2140,162 +2148,225 @@ static Value createLinalgPayloadCalculationForReduceOp(
...
-// The second output buffer contains the maximum value found. It is initialized
-// to the minimum representable value of the input element type. After being
-// populated by indexed_generic, this buffer is disgarded as only the index is
-// requested.
-//
-// The indexed_generic op updates both the maximum value and index if the
-// current value exceeds the running max.
-class ConvertAtenArgmaxOp : public OpConversionPattern<AtenArgmaxOp> {
+static std::pair<Value, Value>
+createArgMaxKernel(OpBuilder &b, Location loc, Value input,
```
ramiro050, line 2169: This function should be called `createAtenMaxDimKernel`

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2140,162 +2148,225 @@ static Value createLinalgPayloadCalculationForReduceOp(
...
-        inElementType,
-        APFloat::getLargest(
-            inElementType.cast<mlir::FloatType>().getFloatSemantics(), true));
-
-    Value fillValueMax =
-        rewriter.create<arith::ConstantOp>(loc, fillValueMaxAttr);
-    Value filledTensorMax =
-        rewriter.create<linalg::FillOp>(loc, fillValueMax, initTensorMax)
-            .result();
+    std::pair<Value, Value> argMaxResult = createArgMaxKernel(
```
ramiro050, line 2312: rename variable to `maxDimResult`

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
ramiro050: I know, I just meant that because this kernel calculation is made for `maxDimOp`, it should use the same order of arguments that `maxDimOp` has

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
pashu123: @cathyzhyi @ramiro050 I have made the required changes, please re-review.

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -154,3 +154,34 @@ func @torch.aten.arange.start() -> !torch.vtensor<[?],si64> {
...
+// CHECK-LABEL: func @torch.aten.argmax.reduceall(
+// CHECK-SAME:   %[[INP:.*]]: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[],si64> {
+// CHECK:        %[[NONE:.*]] = torch.constant.none
+// CHECK:        %[[FALSE:.*]] = torch.constant.bool false
+// CHECK:        %[[CST0:.*]] = torch.constant.int 0
+// CHECK:        %[[CST1:.*]] = torch.constant.int 1
+// CHECK:        %[[FLATTEN:.*]] = torch.aten.flatten.using_ints %[[INP]], %[[CST0]], %[[CST1]] : !torch.vtensor<[?,?],f32>, !torch.int, !torch.int -> !torch.vtensor<[?],f32>
+// CHECK:        %[[VAL:.*]], %[[IND:.*]] = torch.aten.max.dim %[[FLATTEN]], %[[CST0]], %[[FALSE]] : !torch.vtensor<[?],f32>, !torch.int, !torch.bool -> !torch.vtensor<[],f32>, !torch.vtensor<[],si64>
+// CHECK:        return %[[IND]] : !torch.vtensor<[],si64>
+func @torch.aten.argmax.reduceall(%arg0: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[],si64> {
```
ramiro050, line 182: nit: This should be named `torch.aten.argmax$reduce_all`

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2141,64 +2148,69 @@ static Value createLinalgPayloadCalculationForReduceOp(
...
                 "createLinalgPayloadCalculationForReduceOp");
   return nullptr;
 }
 
 namespace {
-// Aten argmax lowering represents the ArgMax op as an linalg.indexed_generic
+// Aten maxdim lowering represents the MaxDim op as an linalg.indexed_generic
 // op, producing two output buffers.
 //
 // The first output buffer contains the index of the found maximum value. It is
```
ramiro050, line 2166: the first output buffer is the values tensor

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2141,64 +2148,69 @@ static Value createLinalgPayloadCalculationForReduceOp(
...
 // initialized to 0 and is resulting integer type.
 //
 // The second output buffer contains the maximum value found. It is initialized
```
ramiro050, line 2169: the second output buffer is the index tensor

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2126,6 +2126,13 @@ static Value createLinalgNeutralElementForReduceOp(OpBuilder &b, Location loc,
   if (isa<AtenSumOp, AtenSumDimIntListOp>(op) &&
       elementType.isa<mlir::FloatType>())
     return b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
+  if (isa<AtenMaxOp>(op) && elementType.isa<mlir::FloatType>())
+    return b.create<arith::ConstantOp>(
+        loc, b.getFloatAttr(
+                 elementType,
+                 APFloat::getSmallest(
```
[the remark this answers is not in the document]
pashu123: Thanks for this.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
pashu123: Done

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -325,6 +325,56 @@ class DecomposeAten_LogSoftmaxBackwardDataOp
...
+            .getWithSizesAndDtype(indicesTensorType.getSizes(),
+                                  inputType.getDtype())
+            .cast<BaseTensorType>();
+
+    // If the dim type is `NoneType` i.e. reduce along all the dimensions.
+    // `AtenMaxDimOp` doesn't support dim as `NoneType` so first the input
+    // tensor is flattened to 1d tensor and then the reduction happens on the
+    // 0th dimension.
+    if (dim.getType().isa<Torch::NoneType>()) {
+      BaseTensorType tempType =
```
ramiro050, line 357: nit: change `tempType` to something more descriptive: `flattenType`

ramiro050 (commented):
A few small comments:

e2e_testing/torchscript/reduction.py
```diff
@@ -122,3 +122,97 @@ def forward(self, a):
...
+  @annotate_args([
+      None,
+      ([-1, -1, -1], torch.float32, True),
+  ])
+  def forward(self, a):
+    return torch.ops.aten.max(a)
+
+@register_test_case(module_factory=lambda: ReduceMaxAllDims())
+def ReduceMaxAllDims_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5, low=-10, high=10))
```
ramiro050, line 218: So the case that we really care about is having a tensor of only negative numbers away from zero because if there is even a single non-negative number, we won't be able to catch an incorrectly initialized tensor (regarding the `getLargest` and `getSmallest` functions), since `max(-infinity, pos_val) == max(0, pos_val)`. Since this tensor has 60 elements and each is equally likely to be positive or negative, it is almost guaranteed to have at least one positive number. 

I would do something like `low=-10, high=-5`

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
ramiro050: I think this suggestion was missed

ramiro050 (approved):
LGTM! I just have a couple of comments

e2e_testing/torchscript/reduction.py
pashu123: I see, thanks.

e2e_testing/torchscript/reduction.py
```diff
@@ -122,3 +122,97 @@ def forward(self, a):
...
+        return torch.ops.aten.max(a, 1)[0]
+
+
+@register_test_case(module_factory=lambda: ReduceMaxAlongDim())
+def ReduceMaxAlongDim_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5).to(torch.float64))
+
+# ==============================================================================
+
+class ReduceMaxAlongDimNegative(torch.nn.Module):
```
cathyzhyi, line 147: @pashu123  Noticing this when reviewing another PR. The test name is ambiguious. Do you mean the element is negative or the dim is negative?

PR #538, opened 2022-01-25
[Torch][Linalg] Add basic support for RNG

cathyzhyi:
This PR include the following pieces:

- Add torch `Generator` type. `Generator` type is converted to i64 in
 refbackend type converter.
 - Add seed managment support for the default global generator.
 `torch_c.getNextSeed` op is used to get the seed. On refbackend, the
 `torch_c.getNextSeed` is lowered to load/store from [0] of global
 variable `default_generator` memref<i64> in `InsertRngGlobals` pass.
 - Add `aten.uniform_` and testing as an example op RNG ops. Add
 `torch.pseudo.aten.uniform` op. It has the same operands and return as
 the `aten.uniform_` from the op registry except for value semantics.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedAtenOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedAtenOps.td
@@ -1325,6 +1325,22 @@ def Torch_AtenThresholdBackwardOp : Torch_Op<"aten.threshold_backward", [
   let assemblyFormat = "$grad_output `,` $self `,` $threshold attr-dict `:` qualified(type($grad_output)) `,` qualified(type($self)) `,` qualified(type($threshold)) `->` qualified(type($result))";
 }
 
+def Torch_AtenUniform_Op : Torch_Op<"aten.uniform_", [
+    AllowsTypeRefinement
+  ]> {
+  let summary = "Generated op for `aten::uniform_ : (Tensor, float, float, Generator?) -> (Tensor)`";
+  let arguments = (ins
+    AnyTorchTensorType:$self,
+    Torch_FloatType:$from,
+    Torch_FloatType:$to,
+    TorchOptionalGeneratorType:$generator
+  );
+  let results = (outs
+    AnyTorchTensorType:$result
+  );
+  let assemblyFormat = "$self `,` $from `,` $to `,` $generator attr-dict `:` qualified(type($self)) `,` qualified(type($from)) `,` qualified(type($to)) `,` qualified(type($generator)) `->` qualified(type($result))";
+}
+
 def Torch_AtenTriuOp : Torch_Op<"aten.triu", [
     AllowsTypeRefinement,
     HasValueSemantics
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
@@ -912,4 +912,23 @@ def Torch_OverwriteTensorOp : Torch_Op<"overwrite.tensor", [
   }];
 }
 
+// The corresponding without underscore variant for `torch.aten.uniform_`
+// doesn't exist in the pytorch ops registry. Add it here.
+def Torch_PseudoAtenUniformOp: Torch_Op<"pseudo.aten.uniform", [
+    AllowsTypeRefinement,
+    HasValueSemantics,
+  ]> {
+  let summary = "`uniform op : (Tensor, float, float, Generator?) -> (Tensor)`";
+  let arguments = (ins
+    AnyTorchTensorType:$self,
+    Torch_FloatType:$from,
+    Torch_FloatType:$to,
+    TorchOptionalGeneratorType:$generator
+  );
+  let results = (outs
+    AnyTorchTensorType:$result
+  );
+  let assemblyFormat = "$self `,` $from `,` $to `,` $generator attr-dict `:` type($self) `,` type($from) `,` type($to) `,` type($generator) `->` type($result)";
+}
+
 #endif // TORCH_OPS
--- a/include/torch-mlir/Dialect/Torch/IR/TorchTypes.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchTypes.td
@@ -243,6 +243,10 @@ def Torch_DeviceType : Torch_Type<"Device", "Device"> {
   let summary = "Torch device";
 }
 
+def Torch_GeneratorType : Torch_Type<"Generator", "Generator"> {
+  let summary = "Torch Generator for producing pseudo-random numbers";
+}
+
 def Torch_BoolType : Torch_Type<"Bool", "bool"> {
   let summary = "Torch BoolType";
   let description = [{
@@ -384,6 +388,8 @@ def TorchOptionalStringType:
       OptionalOf<Torch_StringType, "Optional torch Str type">;
 def TorchOptionalDeviceType:
       OptionalOf<Torch_DeviceType, "Optional torch device type">;
+def TorchOptionalGeneratorType:
+      OptionalOf<Torch_GeneratorType, "Optional torch Generator type">;
 
 def IsListTypePred : CPred<"$_self.isa<::mlir::torch::Torch::ListType>()">;
 class ListOf<list<Type> allowedTypes, string descr> :
@@ -430,6 +436,7 @@ def AnyTorchType : AnyTypeOf<[
   Torch_BoolType,
   Torch_DictType,
   Torch_DeviceType,
+  Torch_GeneratorType,
   Torch_ListType,
   Torch_LinearParamsType,
   Torch_NumberType,
--- a/include/torch-mlir/RefBackend/Passes.td
+++ b/include/torch-mlir/RefBackend/Passes.td
@@ -18,6 +18,12 @@ def MungeCallingConventions : Pass<"refback-munge-calling-conventions", "ModuleO
   let dependentDialects = ["memref::MemRefDialect"];
 }
 
+def InsertRngGlobals: Pass<"refback-insert-rng-globals", "ModuleOp"> {
+  let summary = "Insert global variables and sequence to get the next global seed for RNG ops";
+  let constructor = "mlir::torch::RefBackend::createInsertRngGlobalsPass();";
+  let dependentDialects = ["memref::MemRefDialect"];
+}
+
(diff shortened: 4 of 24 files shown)


include/torch-mlir/Dialect/Torch/IR/GeneratedAtenOps.td
```diff
@@ -1297,6 +1297,23 @@ def Torch_AtenThresholdBackwardOp : Torch_Op<"aten.threshold_backward", [
   let assemblyFormat = "$grad_output `,` $self `,` $threshold attr-dict `:` qualified(type($grad_output)) `,` qualified(type($self)) `,` qualified(type($threshold)) `->` qualified(type($result))";
 }
 
+def Torch_AtenUniform_Op : Torch_Op<"aten.uniform_", [
+    IsTrailingUnderscoreInplaceVariant,
```
silvasean, line 1301: I don't think this should have IsTrailingUnderscoreInplaceVariant, because that will cause ReduceOpVariants to convert it to `aten.uniform` which doesn't exist.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -912,4 +912,21 @@ def Torch_OverwriteTensorOp : Torch_Op<"overwrite.tensor", [
   }];
 }
 
+def Torch_RngUniformOp : Torch_Op<"rng_uniform", [
+    AllowsTypeRefinement,
+    HasValueSemantics,
+  ]> {
+  let summary = "`uniform op : (Tensor, float, float, Generator?) -> (Tensor)`";
```
silvasean, line 919: Can you explain what you mean by having this op "as the building
block of other torch rng ops"?

include/torch-mlir/Dialect/Torch/IR/TorchTypes.td
```diff
@@ -243,6 +243,10 @@ def Torch_DeviceType : Torch_Type<"Device", "Device"> {
   let summary = "Torch device";
 }
 
+def Torch_GeneratorType : Torch_Type<"Generator", "Generator"> {
+  let summary = "Torch Generator for producing pseudo random numbers";
```
silvasean, line 247: nit: pseudo-random

include/torch-mlir/Dialect/Torch/Utils/Utils.h
```diff
@@ -19,6 +19,15 @@ namespace Torch {
 int64_t toPositiveDim(int64_t dim, int64_t inputRank);
 bool isValidDim(int64_t dim, int64_t inputRank);
 bool getListConstructElements(Value v, SmallVectorImpl<Value> &elems);
+constexpr StringRef getDefaultGeneratorGlobalVarName() {
```
silvasean, line 22: This should be a detail of backends. It shouldn't be part of the Torch Dialect.

include/torch-mlir/RefBackend/Passes.td
```diff
@@ -18,6 +18,12 @@ def MungeCallingConventions : Pass<"refback-munge-calling-conventions", "ModuleO
   let dependentDialects = ["memref::MemRefDialect"];
 }
 
+def InsertRngGlobals: Pass<"insert-rng-globals", "ModuleOp"> {
```
silvasean, line 21: nit: refback-insert-rng-globals

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -4308,6 +4308,147 @@ class ConvertAtenIndexTensorOp : public OpConversionPattern<AtenIndexTensorOp> {
...
 
+// Helper function for creating FuncOp given function name and function type.
+static FuncOp findOrCreateGeneratorFunc(OpBuilder &b, Operation *op,
+                                        StringRef fnName, FunctionType fnTy) {
+  auto module = op->getParentOfType<ModuleOp>();
+  auto funcOp = module.lookupSymbol<FuncOp>(fnName);
+  if (funcOp)
+    return funcOp;
+
+  OpBuilder::InsertionGuard guard(b);
```
silvasean, line 4319: TorchToLinalg is a FuncOp pass, so it cannot mutate the outer module like this. I would recommend adding an op `torch_c.get_next_seed : () -> i64` and then have InsertRngGlobals basically create a function `getNextSeed` (signature () -> i64) and a global `currentSeed` (one-element memref of i64).

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -4308,6 +4308,147 @@ class ConvertAtenIndexTensorOp : public OpConversionPattern<AtenIndexTensorOp> {
...
+                  Value updateScaled =
+                      b.create<arith::MulFOp>(loc, updateFloat, scale);
+                  Value res = b.create<arith::AddFOp>(loc, updateScaled, min);
+                  b.create<linalg::YieldOp>(loc, res);
+                })
+            .getResult(0);
+
+    // Set the next seed by recaculating the temp from the last iteration of
+    // `linalg.generic`.
+    Value temp = initialSeed;
```
silvasean, line 4432: TorchToLinalg shouldn't need to be concerned with updating the seed. The internal implementation that a backend chooses for `torch_c.get_next_seed` is responsible for that.

lib/RefBackend/RefBackend.cpp
```diff
@@ -22,6 +22,7 @@
 #include "mlir/Dialect/StandardOps/IR/Ops.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "set"
```
silvasean, line 24: nit: (not for this PR): should be `#include <set>` and be near `<numeric>` below.

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -157,6 +158,16 @@ def nans(self, *sizes):
...
 
 
+class ResultComparisonType(Enum):
+    """ A enum class indicating how the test result is compared with the golden result.
+    """
+    # For a tensor type result, every element must be close to the golden result.
+    DEFAULT = 1
+    # For a tensor type result, the mean and standard deviation of all elements
+    # are close to the golden result.
+    STATISTIC = 2
```
silvasean, line 168: I don't think the test framework should know about this. I think it would be better to let the test itself call `torch.std`/etc. and return those values instead of the raw random values. The reason I think that is better is that it will be test-specific which statistics make sense to compare.

silvasean (commented):
there's a lot of random reformattings of Python code in this PR, which is sometimes confusing. Can you revert those? An easy way to do it is `git checkout -p HEAD~`

include/torch-mlir/Dialect/Torch/IR/GeneratedAtenOps.td
silvasean: Oh, I see that you updated that pattern to handle this specially.

silvasean: I asked @bdhirsh about this in a chat.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
cathyzhyi: I meant other ops like `bernoulli` could be decomposed to this op.

include/torch-mlir/Dialect/Torch/IR/GeneratedAtenOps.td
cathyzhyi: oh, I forgot to run `./build_tools/update_torch_ods.sh`. This trait should be removed.

include/torch-mlir/Dialect/Torch/Utils/Utils.h
cathyzhyi: Removed all the related function from include/torch-mlir/Dialect/Torch/Utils/Utils.h.

python/torch_mlir_e2e_test/torchscript/framework.py
cathyzhyi: Updated testing as suggested.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -117,6 +117,30 @@ class ConvertToImmutableTensors : public RewritePattern {
...
+// Reduce Ops without value semantics but the corresponding without trailing
+// underscore variant doesn't exist.
+namespace {
+class ReduceNonValueSemanticOps : public RewritePattern {
+public:
+  ReduceNonValueSemanticOps(MLIRContext *context)
+      : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {}
+  LogicalResult matchAndRewrite(Operation *op,
+                                PatternRewriter &rewriter) const override {
+    Operation *newOp;
```
ramiro050, line 129: nit: This line is not needed. You can do `Operation *newOp = rewriter....`, that way all the code related to `newOp` is together

lib/RefBackend/RefBackend.cpp
```diff
@@ -163,9 +168,11 @@ static LogicalResult mungeFunction(
...
     std::string funcName = getConsumeReturnFunctionNameForReturnTypes(retTypes);
     if (supportedConsumeFuncReturnFuncs.find(funcName) == supportedFuncsEnd) {
-      op.emitError("must have one return value of memref types or scalar types "
-                   "of i32, i64, f32, f64, i1, or two return values of memref "
-                   "f32 and i64, or three return values of memref f32");
+      op.emitError(
+          "must have one return value of memref types or scalar types "
+          "of i32, i64, f32, f64, i1, or two return values of memref "
+          "f32 and i64, or two return values of memref f32, or two "
+          "return values of memref f64, or three return values of memref f32");
```
ramiro050, line 175: nit: I wonder if it would be more clear to simply list the actual types rather than describe them. Such as

> Supported return types: 
> mri1, mri32, mri64, mrf32, mrf64, i64, f32, f64, 
> (mrf32, mri64), (mrf32, mrf32), (mrf64, mrf64),
> (mrf32, mrf32, mrf32)

This will make it a bit easier to keep this comment up to date and accurate. At the moment it gives the wrong information in the first sentence because values of `i32` and `i1` are not supported, only memrefs.

ramiro050 (approved):
Awesome work! LGTM, I just have a few minor comments

PR #543, opened 2022-01-27
* [tosa] Support for AtenReshapeOp

anupgangwar:
* [tosa] Support for AtenBatchNormOp

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -1760,6 +1760,161 @@ LogicalResult ConvertAtenOp<AtenConv2dOp>::matchAndRewrite(
   return success();
 }
 
+template <>
+LogicalResult ConvertAtenOp<AtenReshapeOp>::matchAndRewrite(
+    AtenReshapeOp op, OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+
+  auto self = adaptor.self();
+
+  auto selfTy = self.getType().template cast<RankedTensorType>();
+  if (!selfTy)
+    return op.emitError("Only ranked tensor types supported in TOSA Reshape");
+
+  // Check that at most one dimension is -1
+  SmallVector<int64_t> newShape;
+  if (!matchPattern(op.shape(), m_TorchConstantIntList(newShape)))
+    return op.emitError("Only constant shape supported in TOSA Reshape");
+
+  int auto_sz = 0;
+  for (auto s : newShape)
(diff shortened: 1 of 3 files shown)


anupgangwar:
```
Summary:
    Passed: 62
    Expectedly Failed: 189
```
Tagging @cathyzhyi for review.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -1760,6 +1760,166 @@ LogicalResult ConvertAtenOp<AtenConv2dOp>::matchAndRewrite(
...
+  return success();
+}
+
+// Normalization ops perform elementwise ops of a single mean/stdev value
+// against the feature map and because input is NCHW, the rank-1 value must be
+// reshaped so it sits on the same dim as 'C'.
+static LogicalResult reshapeToNormInputDim(Operation *op,
+                                           ConversionPatternRewriter &rewriter,
+                                           TypeConverter *converter,
+                                           Type outType, const Value &toBcast,
```
cathyzhyi, line 1801: nit: don't need & when passing Value type. Value type itself is already a pointer wrapper of the storage type.

cathyzhyi (changes requested):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -1760,6 +1760,166 @@ LogicalResult ConvertAtenOp<AtenConv2dOp>::matchAndRewrite(
...
+  // op5 = mul(op4, bscale)
+  // op6 = add(op5, boffset)
+
+  auto meanType = adaptor.running_mean().getType().dyn_cast<TensorType>();
+  auto varianceType = adaptor.running_var().getType().dyn_cast<TensorType>();
+  if (!varianceType || !meanType)
+    return op.emitError("Only ranked tensor types are supported");
+
+  Value meanVal, varianceVal, weightVal, biasVal;
+  if (meanType.getNumElements() == 0) {
```
cathyzhyi, line 1864: I think this doesn't really exist right? Maybe we could just assert the getNumElements != 0?

anupgangwar: Would that be the case for varianceType as well?

cathyzhyi: Yea, I think so.

anupgangwar:
@cathyzhyi All changes have been incorporated.

cathyzhyi (approved):

PR #559, opened 2022-02-04
Add TMTensor dialect to torch-mlir

cathyzhyi:
This is intended to explore support for non-structured ops that can't
be modeled by Linalg dialect. `tm_tensor.scan` is added as the first
such op. The dialect should aim to be upstreamed in the future.


stellaraccident:
Just glancing through - thanks for doing this. Depending on the timeline of overlap, we may be open to pulling a copy of this into IREE and wiring it to the lower level to aid interop. I wouldn't want to do that long term, but if headed upstream eventually, that can provide an escape valve.

MaheshRavishankar:
Interesting. I was just told about this PR. So this is a way to generate the loop implementation for these ops. At the outset, I am happy with such an interface (you could argue would have been good to separate the tiling interface and lowering to loops interface to start with). The good thing of just lowering to loops is that its the least involved in terms of e2e support.
The issue comes when going down the IREE path. IREE will try to tile and distribute these ops. So for the few ops we have looked at, they work, but it is experimental.

I dont have an opinion on the TMTensor stuff added here. It looks like it is for the fallback case where you are not using IREE. But we want this to stay in sync with `linalg_ext` as well, and we want to be deliberate about what goes in `linalg_ext`. Anything we add to `linalg_ext` will only add to the cost involved in moving these ops to core.

external/llvm-external-projects/torch-mlir-dialects/test/lit.cfg.py
```diff
@@ -0,0 +1,69 @@
+# -*- Python -*-
```
silvasean, line 1: does this test suite get run as part of ninja check-torch-mlir-all?

external/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorBase.td
```diff
@@ -0,0 +1,40 @@
...
+
+//===----------------------------------------------------------------------===//
+// Dialect definition
+//===----------------------------------------------------------------------===//
+
+def TMTensor_Dialect : Dialect {
+  let name = "tm_tensor";
+  let cppNamespace = "::mlir::torch::TMTensor";
+  let description = [{
+    The `tm_tensor` dialect is intended to experiment more support for
```
silvasean, line 23: """
The tm_tensor (tm = torch-mlir) dialect is a temporary staging ground in the torch-mlir project for a set of widely-accepted tensor compute operations that are not well-served by existing representations in MLIR upstream. These ops are currently heavily inspired by the `linalg_ext` dialect (which itself is heavily inspired by the structured ops of the `linalg` dialect). But while `linalg_ext` is meant to power specific codegen transformations, the `tm_tensor` dialect is a much more pure "interface dialect" agnostic to any particular set of transformations applied to the operations. We simply require a way to name the specified operations for interchange between projects, without taking strong opinions on the mechanics of transformations.

The dialect does include interfaces to generate scalar reference code for the operations, which simultaneously provides a precise definition of their semantics, and aids in producing executable reference implementations of the operations.

The goal of this dialect is to eventually either be upstreamed or to be subsumed by functionality included by upstream MLIR. It should also be kept consistent with the `linalg_ext` dialect unless there is a good reason not to.
"""

silvasean (approved):
Awesome work Yi! I left a suggestion for the dialect documentation.

This is a lot more code than I imagined... but such is the boilerplate of MLIR :)

silvasean:
Good points @MaheshRavishankar. I view this as the "interface" variant of linalg_ext which is more of a "transformation" dialect. This is very similar to the split between IREE input dialect and the internal one that is very similar to it. Eventually all these stars will collide, because Torch-MLIR will need the IREE input dialect (or something similar) as well. At that point, it should be clear to upstream what the general trend is here for a set of "interface" dialects for general ML programs at this general level of abstraction, and hopefully we can get some consensus around a path forward, if not path has already been charted.

PR #561, opened 2022-02-07
Convert bool to float or integer type.

pashu123:
Conversion of torch.bool tensor type to float and integer type is
handled.


cathyzhyi (approved):
LGTM, thanks!

PR #562, opened 2022-02-07
Add aten.bernoulli decomposition.

pashu123:
aten.bernoulli is decomposed to aten.gtTensor(aten.uniform(x), x).

### DIFF
--- a/e2e_testing/torchscript/rng.py
+++ b/e2e_testing/torchscript/rng.py
@@ -42,3 +42,57 @@ def UniformModule_basic(module, tu: TestUtils):
         tu.rand(512, 1024, 4).double(),
         tu.rand(512, 256, 4).double())
 
+# ==============================================================================
+
+class BernoulliModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, a):
+        x = torch.bernoulli(a)
+        mean = torch.mean(x)
+        std = torch.std(x)
+        return  mean, std
+
+@register_test_case(module_factory=lambda: BernoulliModule())
+def BernoulliModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(256, 512, 64))
+
+class BernoulliFloatModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float64, True),
+        ([-1, -1, -1], torch.float64, True),
+        ([-1, -1, -1], torch.float64, True),
+    ])
+    def forward(self, a, b, c):
+        x = torch.ops.aten.bernoulli_(a, 0.4)
+        y = torch.ops.aten.bernoulli_(b, 0.7)
+        z = torch.ops.aten.bernoulli_(c, 0.5)
+        mean = torch.cat([
+            torch.flatten(torch.mean(x)),
+            torch.flatten(torch.mean(y)),
+            torch.flatten(torch.mean(z))
+        ])
+        std = torch.cat([
+            torch.flatten(torch.std(x)),
+            torch.flatten(torch.std(y)),
+            torch.flatten(torch.std(z))
+        ])
+        return  mean, std
+
+@register_test_case(module_factory=lambda: BernoulliFloatModule())
+def BernoulliFloatModule_basic(module, tu: TestUtils):
+    module.forward(
+        tu.rand(256, 512, 8).double(),
+        tu.rand(512, 1024, 4).double(),
+        tu.rand(512, 256, 4).double())
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -19,6 +19,7 @@
 using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
+using namespace mlir::torch::torch_upstream; // For ScalarType and type
 
 // Helper funtion to get rank of `Base tensor type`.
 // -1 is returned if the tensorRank can't be determined.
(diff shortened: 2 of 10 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -692,6 +709,57 @@ class DecomposeAtenStdOp : public OpRewritePattern<AtenStdOp> {
...
+    Location loc = op.getLoc();
+    Value self = op.self();
+    BaseTensorType inputTensorTy = self.getType().cast<BaseTensorType>();
+    // `intType` contains the corresponding integer for the dtype which is used
+    // by the aten.to.dtype op.
+    int intType = getIntForDtype(inputTensorTy.getDtype());
+    Value convertIntVal =
+        rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(intType));
+    if (!inputTensorTy.hasSizes())
+      return failure();
```
cathyzhyi, line 728: nit: better error msg.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -33,6 +35,21 @@ static int getTensorRank(Value tensor) {
   return tensorRank;
 }
 
+// Helper function to return the int_value for the given dtype.
+static int getIntForDtype(Type type) {
```
cathyzhyi, line 39: There is a very similar helper in RefineTypes https://github.com/llvm/torch-mlir/blob/127c7d8e27ede211af77392b5277ba06fb515fb7/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L35. Can you refactor the code and put the common part in Torch/Utils/Utils.h

cathyzhyi (changes requested):
Looks mostly good!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
pashu123: I think the better place would be `torch_upstream.h` since ScalarType enum is defined over there and this function also returns the same.

cathyzhyi: torch_upstream.h are for things directly copied from upstream. We shouldn't include anything specific to our project there like #include "torch-mlir/Dialect/Torch/IR/TorchOps.h" or Type.

pashu123: Oh! I see thanks for the explanation.

pashu123: Done.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -692,6 +694,103 @@ class DecomposeAtenStdOp : public OpRewritePattern<AtenStdOp> {
 };
 } // namespace
 
+// Determine uniform lower and upper bounds given probability p.
+// The lower and upper bounds lie in range [0, 1].
+static std::pair<double, double> giveUniformBounds(double p) {
+  std::pair<double, double> bounds{0.0, 1.0};
```
cathyzhyi, line 700: could you include the source for the formula.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -692,6 +694,103 @@ class DecomposeAtenStdOp : public OpRewritePattern<AtenStdOp> {
...
+      loc, rewriter.getF64FloatAttr(uniformBounds.second));
+
+  Value falseVal = rewriter.create<ConstantBoolOp>(loc, false);
+  Value noneVal = rewriter.create<ConstantNoneOp>(loc);
+
+  // Create a uniform random op with low and high set to lb and ub respectively.
+  Value uniformRandom = rewriter.create<PseudoAtenUniformOp>(
+      loc, inputType, input, lb, ub, noneVal);
+  Value gtValue =
+      rewriter.create<AtenGtTensorOp>(loc, boolType, input, uniformRandom);
```
cathyzhyi, line 741: I just realize that the uniformRandom result should be compared against `p` rather than input. `Fills each location of self with an independent sample from `Bernoulli(p)`. `Bernoulli` is `takes the value 1 with probability p and the value 0 with probability 1-p`. The comparison result should be AtenLTScalar(uniformRandom, p).

cathyzhyi (changes requested):

e2e_testing/torchscript/rng.py
```diff
@@ -42,3 +42,52 @@ def UniformModule_basic(module, tu: TestUtils):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, a):
+        x = torch.ops.aten.bernoulli_(a, 0.4)
```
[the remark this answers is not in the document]
pashu123: Done.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -692,6 +693,94 @@ class DecomposeAtenStdOp : public OpRewritePattern<AtenStdOp> {
...
+      rewriter.create<ConstantFloatOp>(loc, rewriter.getF64FloatAttr(1.0));
+
+  Value falseVal = rewriter.create<ConstantBoolOp>(loc, false);
+  Value noneVal = rewriter.create<ConstantNoneOp>(loc);
+
+  // Create a uniform random op with low and high set to lb and ub respectively.
+  Value uniformRandom = rewriter.create<PseudoAtenUniformOp>(
+      loc, inputType, input, lb, ub, noneVal);
+  Value gtValue = rewriter.create<AtenGtScalarOp>(loc, boolType, uniformRandom,
+                                                  negProb);
```
cathyzhyi, line 731: I think `value 1 with probability p` should be `< p` not `> (1-p)`

cathyzhyi (changes requested):

cathyzhyi (changes requested):
Thanks!

cathyzhyi (approved):

PR #563, opened 2022-02-07
[TORCH][MLIR] Add E2E support for `aten.native_batch_norm` op

Shukla-Gaurav:
- This commit adds support for `aten.native_batch_norm` operation.
- The current implementation only supports inference mode of
  `aten.native_batch_norm` op.

Signed-Off-By: Gaurav Shukla <gaurav@nod-labs.com>

### DIFF
--- a/lib/Dialect/Torch/Transforms/RefineTypes.cpp
+++ b/lib/Dialect/Torch/Transforms/RefineTypes.cpp
@@ -493,6 +493,8 @@ public:
       return visitAtenNllLossBackwardOp(nllBackwardOp, operands);
     } else if (auto nativeLayerNormOp = dyn_cast<AtenNativeLayerNormOp>(op)) {
       return visitAtenNativeLayerNormOp(nativeLayerNormOp, operands);
+    } else if (auto nativeBatchNormOp = dyn_cast<AtenNativeBatchNormOp>(op)) {
+      return visitAtenNativeBatchNormOp(nativeBatchNormOp, operands);
     } else if (auto constantPadNdOp = dyn_cast<AtenConstantPadNdOp>(op)) {
       return visitAtenConstantPadNdOp(constantPadNdOp, operands);
     } else if (auto indexTensorOp = dyn_cast<AtenIndexTensorOp>(op)) {
@@ -654,6 +656,9 @@ private:
       ArrayRef<LatticeElement<ValueKnowledge> *> operands);
   ChangeResult visitAtenNativeLayerNormOp(
       AtenNativeLayerNormOp op,
+      ArrayRef<LatticeElement<ValueKnowledge> *> operands);
+  ChangeResult visitAtenNativeBatchNormOp(
+      AtenNativeBatchNormOp op,
       ArrayRef<LatticeElement<ValueKnowledge> *> operands);
   ChangeResult
   visitAtenIndexTensorOp(AtenIndexTensorOp op,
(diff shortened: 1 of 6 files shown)


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1859,6 +1864,47 @@ ChangeResult TypeAnalyzer::visitAtenAddCLikeOp(
...
+  // the input tensor as well as the output tensor should be (N, C, D?, H?, W?).
+  // The running_mean, running_var, weight, and bias should be of size (C).
+  bool training = false;
+  if (matchPattern(op.training(), m_TorchConstantBool(&training)) &&
+      input.hasSizes && input.sizes.size() >= 2) {
+    batchNormKnowledge.hasSizes = true;
+    meanKnowledge.hasSizes = true;
+    invStdKnowledge.hasSizes = true;
+    batchNormKnowledge.sizes = input.sizes;
+    meanKnowledge.sizes.resize(1, 0);
```
cathyzhyi, line 1893: Is the return size rank 0 or rank 1 with size 0? if it's rank 0, I think we don't need to `resize(1, 0)`

cathyzhyi (commented):
Looks good! I just have one question and one nit.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1859,6 +1864,47 @@ ChangeResult TypeAnalyzer::visitAtenAddCLikeOp(
...
+    invStdKnowledge.sizes.resize(1, 0);
```
ramiro050, line 1894: nit: I think it would be better to directly assign the contents of the `sizes` vectors by doing `...sizes = {0};` This makes it clearer what the vector is and in the case that `sizes` for some reason already contains values, this will ensure that the vector gets set to a single 0 element (`resize(1, 0)` does not fill the value 0 if the original vector is not empty).

ramiro050 (approved):
Thanks for the comments in the decomposition. They made things very easy to follow. I just have one small comment

lib/Dialect/Torch/Transforms/RefineTypes.cpp
ramiro050: From running one of the unit tests, it seems that the result is rank 1 with size 0

```python
In [13]: _, mean, var = torch.ops.aten.native_batch_norm(torch.rand(2,5,2,3), torch.rand(5), torch.rand(5), 
                                                         torch.rand(5), torch.rand(5), training=False, momentum=0.1, eps=0.0001)

In [14]: mean.shape
Out[14]: torch.Size([0])

In [15]: var.shape
Out[15]: torch.Size([0])
```

Shukla-Gaurav: Yes, it is rank 1 size 0. :)

lib/Dialect/Torch/Transforms/RefineTypes.cpp
Shukla-Gaurav: yeah, {0} makes it more clearer and improves readability. Thanks :)

cathyzhyi (approved):

byronyi:
@Shukla-Gaurav any plan to support training mode for batch norm?

Shukla-Gaurav:
@byronyi I can take a look into it. Although there is some issue with the training mode, I will post a draft for it first. This may take a couple of days.

cathyzhyi:
@Shukla-Gaurav FYI, there is an issue  https://github.com/llvm/torch-mlir/issues/663 with regards to training and Sean had some suggestions on how to proceed in that issue.

PR #564, opened 2022-02-08
[TORCH][MLIR] Add run-time assert support in Torch-dialect

Shukla-Gaurav:
- This commit adds `aten.assert` op in the Torch dialect.
- The `aten.assert` op is lowered to `mlir::Assert` op.


include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -931,4 +931,19 @@ def Torch_PseudoAtenUniformOp: Torch_Op<"pseudo.aten.uniform", [
   let assemblyFormat = "$self `,` $from `,` $to `,` $generator attr-dict `:` type($self) `,` type($from) `,` type($to) `,` type($generator) `->` type($result)";
 }
 
+// This operation is added to support runtime assertion in the Torch Dialect.
```
cathyzhyi, line 934: I think we need more comments here to justify why this op is introduced instead of doing what TS compiler does for `torch.assert`. `torch.assert` would introduce control flow like:
```
 %cond = "torch.aten.Bool.Tensor"(%0) : (!torch.tensor) -> !torch.bool
    "torch.prim.If"(%cond) ({
      "torch.prim.If.yield"() : () -> ()
    }, {
      "torch.prim.RaiseException"(%msg) : (!torch.str) -> () 
      "torch.prim.If.yield"() : () -> () 
    }) : (!torch.bool) -> ()
```
 This new torch op is added to simplify the IR control flow (without branch). Also it makes insertion of runtime assert in the source code easier. 

We could call the op `Torch_RuntimeAssertOp`(torch.runtime.assert) to make it clear that this is not part of aten.

cathyzhyi (changes requested):
Thanks for the change! The code looks good to me. Just need more comments and some adjustment to naming.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
Shukla-Gaurav: Done, Thanks :)

PR #569, opened 2022-02-09
[LINALG] Add E2E support for `aten.[Bool.Tensor|Float.Tensor]` op

Shukla-Gaurav:
- This commit adds lowering of `aten.Bool.Tensor` and
  `aten.Float.Tensor` op as a part of `convert-torch-to-linalg` pass.
- It also adds support for returning bool types.
- It also fixes lowering of the `aten.Int.Tensor` op for non-zero rank
  input tensors.
- If a scalar number is converted to a 0-d tensor and passed on to the
  `aten.Float.Tensor` op, it folds to the scalar number.

### DIFF
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -3804,29 +3804,40 @@ public:
 };
 } // namespace
 
-// Casts a 0d integer tensor to elemental type.
 namespace {
-class ConvertAtenIntTensorOp : public OpConversionPattern<AtenIntTensorOp> {
+// Casts a tensor of exactly one element to an elemental type.
+template <typename OpTy>
+class ConvertAtenTensorToScalarLikeOp : public OpConversionPattern<OpTy> {
 public:
-  using OpConversionPattern::OpConversionPattern;
+  using OpConversionPattern<OpTy>::OpConversionPattern;
   LogicalResult
-  matchAndRewrite(AtenIntTensorOp op, OpAdaptor adaptor,
+  matchAndRewrite(OpTy op,
+                  typename OpConversionPattern<OpTy>::OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
       return failure();
-    Value intTensor = adaptor.a();
-    auto tensorType = intTensor.getType().cast<RankedTensorType>();
+    Location loc = op.getLoc();
+    Value input = adaptor.a();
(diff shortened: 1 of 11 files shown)


lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3808,30 +3808,30 @@ class ConvertAtenSizeIntOp : public OpConversionPattern<AtenSizeIntOp> {
 };
 } // namespace
 
-// Casts a 0d integer tensor to elemental type.
 namespace {
-class ConvertAtenIntTensorOp : public OpConversionPattern<AtenIntTensorOp> {
+// Casts a 0d tensor to elemental type.
```
pashu123, line 3812: Nit: to an elemental

pashu123 (approved):
LGTM!

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -3808,30 +3808,30 @@ class ConvertAtenSizeIntOp : public OpConversionPattern<AtenSizeIntOp> {
...
       return failure();
-    Value intTensor = adaptor.a();
-    auto tensorType = intTensor.getType().cast<RankedTensorType>();
+    Value input = adaptor.a();
+    auto inputType = input.getType().cast<RankedTensorType>();
 
-    if (tensorType.getRank() != 0)
-      return rewriter.notifyMatchFailure(
-          op, "invalid rank: the rank of the input tensor must be 0");
+    if (inputType.getRank() != 0)
```
cathyzhyi, line 3825: The rank can be more than rank 0. It's ok as long as there is only one element.

cathyzhyi (changes requested):

pashu123:
@Shukla-Gaurav The [TORCH][MLIR] thing should change to [LINALG] since we are providing support for linalg backend in torch-mlir project. There is no sub-project [TORCH][MLIR] inside torch-mlir. @cathyzhyi Any opinions?

cathyzhyi:
> @Shukla-Gaurav The [TORCH][MLIR] thing should change to [LINALG] since we are providing support for linalg backend in torch-mlir project. There is no sub-project [TORCH][MLIR] inside torch-mlir. @cathyzhyi Any opinions?

yea, that makes sense.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
Shukla-Gaurav: Thanks for this point :)

Shukla-Gaurav:
@cathyzhyi Thanks for the review. It is now ready for re-review.

cathyzhyi (approved):
just one nit.Thanks!

PR #572, opened 2022-02-10
Add decomposition of _log_softmax op.

pashu123:
Decompose _log_softmax into log(softmax(x)).


e2e_testing/torchscript/basic.py
```diff
@@ -625,6 +625,24 @@ def SoftmaxIntArgTypeF64Module_basic(module, tu: TestUtils):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, tensor):
+        return torch.ops.aten._log_softmax(tensor, 0, False)
```
cathyzhyi, line 638: nit: comments for `0`, and `False`

cathyzhyi (approved):
Just one nit. 
Thanks for adding the unit test!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -468,6 +468,21 @@ class DecomposeAtenLogSoftmaxIntOp
 };
 } // namespace
 
+// Decompose aten.log_softmax op into: log(softmax(x))
```
ramiro050, line 471: Nit: `log_softmax` -> `_log_softmax`, and `softmax(x)` -> `_softmax(x)`, so that's it's clear why this is different to regular `log_softmax`

PR #573, opened 2022-02-10
[TORCH][MLIR] Add E2E support for `aten._unsafe_view` op.

gprateek93:
This commit adds decomposition of `aten._unsafe_view` op into
`aten.view` op.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -943,6 +943,31 @@ class DecomposeAtenNativeBatchNormOp
 };
 } // namespace
 
+// Decompse `Aten_UnsafeViewOp` into `AtenViewOp`. _unsafe_view() differs from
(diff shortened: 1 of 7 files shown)


pashu123:
@gprateek93 Make sure you add a test in the decompose_complex_ops.mlir file.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -943,6 +943,28 @@ class DecomposeAtenNativeBatchNormOp
 };
 } // namespace
 
+// Decompse `Aten_UnsafeViewOp` into `AtenViewOp`. _unsafe_view() differs from
```
pashu123, line 946: Please add the reference here: https://github.com/pytorch/pytorch/blob/364055b2771ecf9b54f1d67a8bf44bb5496476d4/aten/src/ATen/native/TensorShape.cpp#L2072

cathyzhyi (approved):
LGTM

PR #574, opened 2022-02-10
[MLIR][TORCH] Add TorchToTMTensor pass

vivekkhandelwal1:
This pass is added to lower ops, which can not be lowered
via the TorchToLinalg pass, such as `embedding_dense_backward`
op. This pass also uses torch-mlir's TMTensor Dialect to lower the
complex ops.

Also add torch.bincount op lowering with the help of TMTensor dialect

### DIFF
--- a/include/torch-mlir/Conversion/Passes.td
+++ b/include/torch-mlir/Conversion/Passes.td
@@ -114,4 +114,15 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "FuncOp"> {
   let constructor = "mlir::torch::createConvertTorchToTosaPass()";
 }
 
+def ConvertTorchToTMTensor : Pass<"convert-torch-to-tmtensor", "FuncOp"> {
+  let summary = "Convert recognized Torch ops to TMTensor/Linalg ops";
+  let description = [{
+    Convert ATen ops to tmtensor/linalg ops.
+
+    This pass is similar to the TorchToLinalg pass; the difference is that this
+    pass also makes use of TMTensor Dialect, which the former one doesn't.
+  }];
+  let constructor = "mlir::torch::createConvertTorchToTMTensorPass()";
+}
+
 #endif // TORCHMLIR_CONVERSION_PASSES
--- a/include/torch-mlir/Dialect/Torch/Utils/Utils.h
+++ b/include/torch-mlir/Dialect/Torch/Utils/Utils.h
@@ -9,10 +9,10 @@
 #ifndef TORCHMLIR_DIALECT_TORCH_UTILS_H
 #define TORCHMLIR_DIALECT_TORCH_UTILS_H
 
+#include "mlir/IR/PatternMatch.h"
 #include "mlir/IR/Value.h"
 #include "mlir/Support/LLVM.h"
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
-
 
 namespace mlir {
 namespace torch {
@@ -22,6 +22,9 @@ int64_t toPositiveDim(int64_t dim, int64_t inputRank);
 bool isValidDim(int64_t dim, int64_t inputRank);
 bool getListConstructElements(Value v, SmallVectorImpl<Value> &elems);
 torch_upstream::ScalarType getScalarTypeForType(Type type);
+// Helper to convert a tensor to a specific scalar type.
+Value convertTensorToDtype(PatternRewriter &rewriter, Location loc, Value input,
+                           Type dtype);
 
 } // namespace Torch
 } // namespace torch
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -25,7 +25,6 @@
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
-#include "torch-mlir/Conversion/Utils/Utils.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionDialect.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h"
 #include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
@@ -2254,12 +2253,8 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
 static Value createLinalgNeutralElementForReduceOp(OpBuilder &b, Location loc,
                                                    Operation *op,
                                                    Type elementType) {
-  if (isa<AtenSumOp, AtenSumDimIntListOp>(op)) {
-    if (elementType.isa<mlir::FloatType>())
-      return b.create<arith::ConstantOp>(loc, b.getZeroAttr(elementType));
-    else if (elementType.isa<mlir::IntegerType>())
-      return b.create<arith::ConstantOp>(loc, b.getZeroAttr(elementType));
-  }
+  if (isa<AtenSumOp, AtenSumDimIntListOp>(op))
+    return b.create<arith::ConstantOp>(loc, b.getZeroAttr(elementType));
 
   if (isa<AtenMaxOp>(op)) {
     if (elementType.isa<mlir::FloatType>())
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -21,12 +21,14 @@
 #include "mlir/IR/Matchers.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchDialect.h"
+#include "torch-mlir/Conversion/Utils/Utils.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionDialect.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h"
 #include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
+#include "llvm/ADT/APSInt.h"
 
 #include <numeric>
 
@@ -2510,16 +2512,28 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
 static Value createLinalgNeutralElementForReduceOp(OpBuilder &b, Location loc,
                                                    Operation *op,
                                                    Type elementType) {
-  if (isa<AtenSumOp, AtenSumDimIntListOp>(op) &&
-      elementType.isa<mlir::FloatType>())
-    return b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
-  if (isa<AtenMaxOp>(op) && elementType.isa<mlir::FloatType>())
-    return b.create<arith::ConstantOp>(
-        loc, b.getFloatAttr(
-                 elementType,
-                 APFloat::getLargest(
-                     elementType.cast<mlir::FloatType>().getFloatSemantics(),
-                     /*Negative=*/true)));
+  if (isa<AtenSumOp, AtenSumDimIntListOp>(op)) {
+    if (elementType.isa<mlir::FloatType>())
+      return b.create<arith::ConstantOp>(loc, b.getZeroAttr(elementType));
+    else if (elementType.isa<mlir::IntegerType>())
+      return b.create<arith::ConstantOp>(loc, b.getZeroAttr(elementType));
+  }
+
+  if (isa<AtenMaxOp>(op)) {
+    if (elementType.isa<mlir::FloatType>())
+      return b.create<arith::ConstantOp>(
+          loc, b.getFloatAttr(
+                   elementType,
+                   APFloat::getLargest(
+                       elementType.cast<mlir::FloatType>().getFloatSemantics(),
+                       /*Negative=*/true)));
+    else if (elementType.isa<mlir::IntegerType>() &&
+             elementType.getIntOrFloatBitWidth() != 8)
+      return b.create<arith::ConstantOp>(
+          loc, b.getIntegerAttr(elementType,
+                                APSInt::getSignedMinValue(
+                                    elementType.getIntOrFloatBitWidth())));
+  }
 
   op->emitError("unimplemented lowering in "
                 "createLinalgNeutralElementForReduceOp");
@@ -2529,17 +2543,31 @@ static Value createLinalgNeutralElementForReduceOp(OpBuilder &b, Location loc,
 static Value createLinalgPayloadCalculationForReduceOp(
     OpBuilder &b, Location loc, ValueRange payloadArgs, Operation *op,
     ArrayRef<Value> operands, Type resultElementType) {
-  if (isa<AtenSumOp, AtenSumDimIntListOp>(op) &&
-      resultElementType.isa<mlir::FloatType>()) {
+  if (isa<AtenSumOp, AtenSumDimIntListOp>(op)) {
     Value self =
         convertScalarToDtype(b, loc, payloadArgs[0], resultElementType);
     Value result = payloadArgs[1];
-    return b.create<arith::AddFOp>(loc, self, result);
-  } else if (isa<AtenMaxOp>(op) && resultElementType.isa<mlir::FloatType>()) {
+    if (resultElementType.isa<mlir::FloatType>())
+      return b.create<arith::AddFOp>(loc, self, result);
+    else if (resultElementType.isa<mlir::IntegerType>())
+      return b.create<arith::AddIOp>(loc, self, result);
+  } else if (auto max = dyn_cast<AtenMaxOp>(op)) {
     Value self =
         convertScalarToDtype(b, loc, payloadArgs[0], resultElementType);
     Value result = payloadArgs[1];
-    return b.create<arith::MaxFOp>(loc, self, result);
+    if (resultElementType.isa<mlir::FloatType>())
+      return b.create<arith::MaxFOp>(loc, self, result);
+    else if (resultElementType.isa<mlir::IntegerType>()) {
+      IntegerType intType = max.self()
+                                .getType()
+                                .cast<BaseTensorType>()
+                                .getDtype()
+                                .dyn_cast<mlir::IntegerType>();
+      if (intType.isUnsigned())
+        return b.create<arith::MaxUIOp>(loc, self, result);
+      if (intType.isSigned())
+        return b.create<arith::MaxSIOp>(loc, self, result);
+    }
   }
   op->emitError("unimplemented lowering in "
                 "createLinalgPayloadCalculationForReduceOp");
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -25,6 +25,7 @@
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+#include "torch-mlir/Conversion/Utils/Utils.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionDialect.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h"
 #include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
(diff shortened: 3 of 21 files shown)


cathyzhyi:
@vivekkhandelwal1 https://github.com/llvm/torch-mlir/pull/559 is merged. Could you change your PR to base on the TMTensor dialect instead of linalg_ext? Thanks!

vivekkhandelwal1:
> @vivekkhandelwal1 https://github.com/llvm/torch-mlir/pull/559 is merged. Could you change your PR to base on the TMTensor dialect instead of linalg_ext? Thanks!

Okay,  I will do that. Should we now add a scatter-like op in TMTensor instead of Linalg_ext?

cathyzhyi:
> > @vivekkhandelwal1 #559 is merged. Could you change your PR to base on the TMTensor dialect instead of linalg_ext? Thanks!
> 
> Okay, I will do that. Should we now add a scatter-like op in TMTensor instead of Linalg_ext?

We should still add the op in `linalg_ext` and copy over to TMTensor. We wouldn't want to diverge from linalg_ext.

vivekkhandelwal1:
> Could you change your PR to base on the TMTensor dialect instead of linalg_ext? 

By this, do you mean to say I should modify the pass so that instead of linalg_ext, it should be using TMTensor, or do you want me to just rebase this pass over the changes merged by you related to TMTensor?

cathyzhyi:
> modify the pass so that instead of linalg_ext, it should be using TMTensor

I meant the former `modify the pass so that instead of linalg_ext, it should be using TMTensor`

lib/Conversion/LowerAlgorithmicOps/LowerAlgorithmicOps.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
+#include "torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorDialect.h"
+#include "torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorOps.h"
+
+using namespace mlir;
+using namespace mlir::torch;
+using namespace mlir::torch::Torch;
+using namespace mlir::torch::TorchConversion;
+using namespace mlir::torch::torch_upstream; // For ScalarType and type
+using namespace TMTensor;
```
silvasean, line 36: As a matter of coding style, I'd prefer if we used explicit Torch::, TorchConversion::, torch_upstream::, TMTensor::.

lib/Conversion/LowerAlgorithmicOps/LowerAlgorithmicOps.cpp
```diff
@@ -0,0 +1,80 @@
...
+    target.addLegalDialect<linalg::LinalgDialect, StandardOpsDialect,
+                           math::MathDialect, tensor::TensorDialect,
+                           arith::ArithmeticDialect, TMTensorDialect>();
+    target.addLegalOp<GetNextSeedOp>();
+
+    TypeConverter typeConverter;
+    typeConverter.addConversion([](Type type) { return type; });
+    TorchConversion::setupBackendTypeConversion(target, typeConverter);
+
+    RewritePatternSet patterns(context);
```
silvasean, line 68: I would prefer to land this with at least one pattern.

silvasean (changes requested):

lib/Conversion/LowerAlgorithmicOps/LowerAlgorithmicOps.cpp
```diff
@@ -0,0 +1,80 @@
...
+    TorchConversion::getBackendTypeConversionDependentDialects(registry);
+  }
+
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+    ConversionTarget target(*context);
+    target.addLegalDialect<linalg::LinalgDialect, StandardOpsDialect,
+                           math::MathDialect, tensor::TensorDialect,
+                           arith::ArithmeticDialect, TMTensorDialect>();
+    target.addLegalOp<GetNextSeedOp>();
```
cathyzhyi, line 62: this could be removed and added as needed later.

cathyzhyi (changes requested):

lib/Conversion/LowerAlgorithmicOps/LowerAlgorithmicOps.cpp
vivekkhandelwal1: Let us merge this PR just as a pass since it will take some time to add an op lowering in this pass. Also, it will help others get unblocked who are waiting on this.

lib/Conversion/LowerAlgorithmicOps/LowerAlgorithmicOps.cpp
```diff
@@ -0,0 +1,76 @@
...
+public:
+  void getDependentDialects(DialectRegistry &registry) const override {
+    registry.insert<linalg::LinalgDialect>();
+    registry.insert<math::MathDialect>();
+    registry.insert<StandardOpsDialect>();
+    registry.insert<tensor::TensorDialect>();
+    registry.insert<arith::ArithmeticDialect>();
+    registry.insert<TMTensorDialect>();
+    TorchConversion::getBackendTypeConversionDependentDialects(registry);
+  }
```
joker-eph, line 54: This seems all unnecessary right now? Can you just add what you need when you'll need it?

vivekkhandelwal1: Should I remove the entire `getDependentDialects` method?

lib/Conversion/LowerAlgorithmicOps/LowerAlgorithmicOps.cpp
vivekkhandelwal1: Okay, I will add this op's support.

vivekkhandelwal1: Is it okay if I add the `torch.bincount op` (https://pytorch.org/docs/stable/generated/torch.bincount.html) support along with this pass in this PR?

include/torch-mlir/Conversion/LowerAlgorithmicOps/LowerAlgorithmicOps.h
```diff
@@ -0,0 +1,22 @@
...
+
+#ifndef TORCHMLIR_CONVERSION_LOWERALGORITHMICOPS_LOWERALGORITHMICOPS_H
+#define TORCHMLIR_CONVERSION_LOWERALGORITHMICOPS_LOWERALGORITHMICOPS_H
+
+#include "mlir/Pass/Pass.h"
+#include <memory>
+
+namespace mlir {
+namespace torch {
+std::unique_ptr<OperationPass<FuncOp>> createConvertLowerAlgorithmicOpsPass();
```
cathyzhyi, line 18: `Convert` can be removed.

include/torch-mlir/Conversion/Passes.td
```diff
@@ -114,4 +114,15 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "FuncOp"> {
   let constructor = "mlir::torch::createConvertTorchToTosaPass()";
 }
 
+def ConvertLowerAlgorithmicOps : Pass<"convert-lower-algorithmic-ops", "FuncOp"> {
```
cathyzhyi, line 117: Same here that `convert` need to be removed from the pass name. `lower` is a kind of `convert` already.

vivekkhandelwal1:
> This pass should be a Torch->Torch lowering utility that does not use TMTensor at all (nor the dialect conversion framework). It is really just a version of DecomposeComplexOps but specific to ops with a nontrivial choice of algorithm. We can then have a ConvertTorchToTMTensor pass for lowering Torch ops created by this pass to TMTensor.

@silvasean Could you please elaborate more on this? I am not able to understand your point.

silvasean:
There should be two passes:

lib/Dialect/Torch/Transforms/LowerAlgorithmicOps.cpp
 - responsibility is similar to DecomposeComplexOps -- it is a utility that converts Torch ops to other Torch ops to simplify backend implementation
 - the main difference is that this pass is specifically for ops where there is a quite nontrivial choice of algorithm / parallelization strategy (such as aten.nonzero)
 - This pass does NOT depend on TMTensor
 - This pass runs during torchscript-module-to-torch-backend-pipeline

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
 - Responsibility is to lower Torch ops to TMTensor
 - This pass might create other Torch ops, so it should run before TorchToLinalg / TorchToTosa
 - This pass depends on TMTensor
 - This pass runs during torch-backend-to-{linalg-on-tensors,tosa}-backend-pipeline

cathyzhyi:
There seems to be miscommunication about this. So the pass @vivekkhandelwal1 is trying to add in this PR should be called `TorchToTMTensor`.

vivekkhandelwal1:
I am getting the following error:
```
FAIL - "BincountModule_basic"

Unexpected outcome summary:

****** Failed tests - 1 tests
    FAIL - "BincountModule_basic"
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 303, in run_tests
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/linalg_on_tensors_backend.py", line 45, in compile
            return self.backend.compile(module)
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py", line 220, in compile
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/utils.py", line 49, in run_pipeline_with_repro_report
            raise Exception(f"""
        Exception:
        Lowering Linalg-on-Tensors IR to LLVM with RefBackend failed with the following diagnostics:
        error: failed to legalize operation 'arith.trunci'
        note: see current operation: %8 = "arith.trunci"(%7) : (tensor<?x1xi64>) -> tensor<?x1xi32>

        Error can be reproduced with:
        $ torch-mlir-opt -pass-pipeline='builtin.func(scf-bufferize),builtin.func(linalg-bufferize),builtin.func(refback-munge-memref-copy),func-bufferize,arith-bufferize,builtin.func(tensor-bufferize),builtin.func(finalizing-bufferize),refback-munge-calling-conventions,refback-insert-rng-globals,builtin.func(convert-linalg-to-loops),builtin.func(lower-affine),convert-scf-to-cf,builtin.func(refback-expand-ops-for-llvm),builtin.func(arith-expand),builtin.func(convert-math-to-llvm),convert-linalg-to-llvm,convert-memref-to-llvm,builtin.func(convert-arith-to-llvm),convert-std-to-llvm,convert-cf-to-llvm,reconcile-unrealized-casts' /tmp/BincountModule.mlir
        Add '-print-ir-after-all -mlir-disable-threading' to get the IR dump for debugging purpose.

Summary:
    Failed: 1

```
The detailed IR dump with debugging information is present at: https://gist.github.com/vivekkhandelwal1/023c38747270b996019a5d985030f426.

cathyzhyi:
> arith.trunci
It turns out we can't use `arith.trunci` on tensors so a bufferization pattern doesn't exist for this op. I think you might want to use aten ops to convert data type.

We will need a bufferization pass for the tmtensor dialect. I will add that for you.

vivekkhandelwal1:
@cathyzhyi, I have made the appropriate code changes to replace `arith.trunci`. Now only bufferization error has to be addressed by your pass.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,291 @@
...
+    auto &scatterOpBlock = scatterOpRegion.emplaceBlock();
+    scatterOpBlock.addArgument(resultElemType, scatterOpRegion.getLoc());
+    scatterOpBlock.addArgument(resultElemType, scatterOpRegion.getLoc());
+    auto blockArgs = scatterOpBlock.getArguments();
+
+    // Creating an add instruction inside the scatter op region to increment the
+    // frequency counter with one.
+    OpBuilder regionBuilder(scatterOpRegion);
+    Value add = regionBuilder.create<arith::AddIOp>(scatterOp->getLoc(),
+                                                    blockArgs[0], blockArgs[1]);
```
cathyzhyi, line 240: this can be replaced with `blockArgs[1] + 1` and the update tensor is not needed. also, please change the `blockArgs[0]` and `blockArgs[1]` to `update` and `bincount` for readability.

cathyzhyi (changes requested):
Added a few comments. Not done reviewing yet.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,243 @@
...
+        loc, rewriter.getI64IntegerAttr(1));
+    Value expandedInputTensor = rewriter.create<AtenUnsqueezeOp>(
+        loc, expandInputType, op.self(), torchCstOne);
+
+    // Converting the input element type from i64 to i32.
+    ValueTensorType truncatedTensorType = ValueTensorType::get(
+        op->getContext(), llvm::makeArrayRef(expandedInputSizes),
+        mlir::IntegerType::get(op->getContext(), 32,
+                               mlir::IntegerType::Signed));
+    int intType = (int)ScalarType::Int;
```
cathyzhyi, line 132: there is a helper for type conversion in https://github.com/llvm/torch-mlir/blob/486f95e84f587d020ba789b071b12f890510f1a1/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L130. Can you move the helper to Torch/utils/utils.cpp and use the helper here.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,243 @@
...
+    Value truncatedInputTensor = rewriter.create<AtenToDtypeOp>(
+        loc, truncatedTensorType, expandedInputTensor, /*dtype*/ convertIntVal,
+        /*non_blocking*/ falseVal,
+        /*copy*/ falseVal, /*memory_format*/ noneVal);
+    truncatedInputTensor = typeConverter->materializeTargetConversion(
+        rewriter, loc,
+        typeConverter->convertType(truncatedInputTensor.getType()),
+        truncatedInputTensor);
+
+    Type resultElemType = getTypeConverter()
```
cathyzhyi, line 146: getTypeConverter() => typeConverter

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,243 @@
...
+                              ->convertType(op->getResult(0).getType())
+                              .cast<RankedTensorType>()
+                              .getElementType();
+
+    Value updatesTensor;
+    if (inputType.hasStaticShape()) {
+      updatesTensor = rewriter.create<linalg::InitTensorOp>(
+          loc, inputType.getShape(), resultElemType);
+    } else {
+      Value inputSize = rewriter.create<tensor::DimOp>(loc, input, 0);
```
cathyzhyi, line 156: we can use `getDimOp` helper here and don't need to check for static shape.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,243 @@
...
+        rewriter.create<arith::MaxSIOp>(loc, maxInputPlusOne, minlength);
+    bincountSize = castIntToIndex(rewriter, loc, bincountSize);
+    Value bincountTensor = createInitTensor(rewriter, loc, {bincountSize},
+                                            resultElemType, constantZero);
+
+    auto scatterOp = rewriter.create<TMTensor::ScatterOp>(
+        loc, bincountTensor.getType(),
+        ValueRange{updatesTensor, truncatedInputTensor},
+        ValueRange{bincountTensor},
+        /*unique_indices*/ false);
```
cathyzhyi, line 179: `/*unique_indices*/  false`=> `/*unique_indices=*/false`

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,243 @@
...
+    int intType = (int)ScalarType::Int;
+    Value convertIntVal = rewriter.create<Torch::ConstantIntOp>(
+        loc, rewriter.getI64IntegerAttr(intType));
+    Value falseVal = rewriter.create<Torch::ConstantBoolOp>(loc, false);
+    Value noneVal = rewriter.create<Torch::ConstantNoneOp>(loc);
+    Value truncatedInputTensor = rewriter.create<AtenToDtypeOp>(
+        loc, truncatedTensorType, expandedInputTensor, /*dtype*/ convertIntVal,
+        /*non_blocking*/ falseVal,
+        /*copy*/ falseVal, /*memory_format*/ noneVal);
+    truncatedInputTensor = typeConverter->materializeTargetConversion(
```
cathyzhyi, line 141: nit: `Value indices` would be better

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,243 @@
...
+    auto &scatterOpBlock = scatterOpRegion.emplaceBlock();
+    Location scatterOpRegionLoc = scatterOpRegion.getLoc();
+    scatterOpBlock.addArguments(TypeRange{resultElemType, resultElemType},
+                                {scatterOpRegionLoc, scatterOpRegionLoc});
+    auto blockArgs = scatterOpBlock.getArguments();
+
+    // Creating an add instruction inside the scatter op region to increment the
+    // frequency counter with one.
+    OpBuilder regionBuilder(scatterOpRegion);
+    Value add = regionBuilder.create<arith::AddIOp>(scatterOp->getLoc(),
```
cathyzhyi, line 191: scatterOp->getLoc() => loc

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+    auto scatterOp = rewriter.create<TMTensor::ScatterOp>(
+        loc, bincountTensor.getType(),
+        ValueRange{updatesTensor, truncatedInputTensor},
+        ValueRange{bincountTensor},
+        /*unique_indices*/ false);
+
+    Region &scatterOpRegion = scatterOp.region();
+    auto &scatterOpBlock = scatterOpRegion.emplaceBlock();
+    Location scatterOpRegionLoc = scatterOpRegion.getLoc();
```
cathyzhyi, line 183: this is the same as `loc`

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,243 @@
...
+    if (!weights.getType().isa<mlir::torch::Torch::NoneType>())
+      return rewriter.notifyMatchFailure(
+          op, "Unimplemented, the weights operand is not incorporated.");
+
+    // Finding the maximum value in the input tensor.
+    SmallVector<int64_t> maxTensorSizes;
+    ValueTensorType maxTensorType = ValueTensorType::get(
+        op->getContext(), llvm::makeArrayRef(maxTensorSizes),
+        op.self().getType().cast<ValueTensorType>().getDtype());
+    Value maxTensor = rewriter.create<AtenMaxOp>(loc, maxTensorType, op.self());
```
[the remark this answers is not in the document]
vivekkhandelwal1: We can't do this because it expects a torch tensor.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,243 @@
...
+    // 3.) `bincount` a 1-d tensor maps to the original in scatter op
+    // with size equal to the max(max(input) + 1, minlength).
+    SmallVector<int64_t> expandedInputSizes{inputType.getShape()[0], 1};
+    ValueTensorType expandInputType = ValueTensorType::get(
+        op->getContext(), llvm::makeArrayRef(expandedInputSizes),
+        op.self().getType().cast<ValueTensorType>().getDtype());
+    Value torchCstOne = rewriter.create<Torch::ConstantIntOp>(
+        loc, rewriter.getI64IntegerAttr(1));
+    Value expandedInputTensor = rewriter.create<AtenUnsqueezeOp>(
+        loc, expandInputType, op.self(), torchCstOne);
```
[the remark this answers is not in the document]
vivekkhandelwal1: Same reason as above for this.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -2510,16 +2253,28 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
-    return b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
-  if (isa<AtenMaxOp>(op) && elementType.isa<mlir::FloatType>())
-    return b.create<arith::ConstantOp>(
-        loc, b.getFloatAttr(
-                 elementType,
-                 APFloat::getLargest(
-                     elementType.cast<mlir::FloatType>().getFloatSemantics(),
-                     /*Negative=*/true)));
+  if (isa<AtenSumOp, AtenSumDimIntListOp>(op)) {
+    if (elementType.isa<mlir::FloatType>())
```
cathyzhyi, line 2257: The if branch is not needed anymore

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,208 @@
...
+        mlir::IntegerType::get(context, 32, mlir::IntegerType::Signed));
+    indices = typeConverter->materializeTargetConversion(
+        rewriter, loc, typeConverter->convertType(indices.getType()), indices);
+
+    Type resultElemType = typeConverter->convertType(op->getResult(0).getType())
+                              .cast<RankedTensorType>()
+                              .getElementType();
+
+    SmallVector<Value, 1> inputSize;
+    if (!inputType.hasStaticShape())
```
cathyzhyi, line 123: we don't need to check static shape. It's ok to treat static and dynamic shape as the same. Refer to https://github.com/llvm/torch-mlir/blob/486f95e84f587d020ba789b071b12f890510f1a1/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp#L593

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,208 @@
...
+    auto &scatterOpBlock = scatterOpRegion.emplaceBlock();
+    scatterOpBlock.addArguments(TypeRange{resultElemType, resultElemType},
+                                {loc, loc});
+    auto blockArgs = scatterOpBlock.getArguments();
+
+    // Creating an add instruction inside the scatter op region to increment the
+    // frequency counter with one.
+    OpBuilder regionBuilder(scatterOpRegion);
+    Value add = regionBuilder.create<arith::AddIOp>(loc,
+                                                    /*bincount*/ blockArgs[1],
```
cathyzhyi, line 157: Please fix all similar places in this PR.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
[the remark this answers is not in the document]
cathyzhyi: I see. `op.self()` is used multiple times in this method. It probably worth a variable like `torchTypeInput`

vivekkhandelwal1:
@cathyzhyi, please re-review this PR when you find the time. I need this for index_put op.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,208 @@
...
+                              .cast<RankedTensorType>()
+                              .getElementType();
+
+    SmallVector<Value, 1> inputSize;
+    if (!inputType.hasStaticShape())
+      inputSize = getTensorSizesUntilDim(rewriter, loc, input, 0);
+    Value updatesTensor = rewriter.create<linalg::InitTensorOp>(
+        loc, inputSize, inputType.getShape(), resultElemType);
+
+    Value constantZero = rewriter.create<arith::ConstantOp>(
```
cathyzhyi, line 128: any specific reason why cst1 uses `inputType` while the cst0 uses `resultElemType`?

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,208 @@
...
+        loc, rewriter.getZeroAttr(inputType.getElementType()));
+    Value constantOne = rewriter.create<arith::ConstantIntOp>(
+        loc, 1, resultElemType.getIntOrFloatBitWidth());
+
+    // Bincount size = max(max(input) + 1, minlength)
+    Value maxInputPlusOne =
+        rewriter.create<arith::AddIOp>(loc, maxInput, constantOne);
```
cathyzhyi, line 135: Can you verify if this would work if the input is `torch.int32`? The `maxInput` and `cst1` would have different bit width in such case. It's probably easier to just convert the input to `int64` upfront.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,208 @@
...
+                                {loc, loc});
+    auto blockArgs = scatterOpBlock.getArguments();
+
+    // Creating an add instruction inside the scatter op region to increment the
+    // frequency counter with one.
+    OpBuilder regionBuilder(scatterOpRegion);
+    Value add = regionBuilder.create<arith::AddIOp>(loc,
+                                                    /*bincount*/ blockArgs[1],
+                                                    constantOne);
+    regionBuilder.create<TMTensor::YieldOp>(scatterOp->getLoc(), add);
```
cathyzhyi, line 159: scatterOp->getLoc() => loc

cathyzhyi (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,208 @@
...
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+    Location loc = op.getLoc();
+    MLIRContext *context = op->getContext();
+    Value input = adaptor.self();
+    Value minlength = adaptor.minlength();
+    Value weights = adaptor.weights();
+
+    // Check whether the input is a 1-d tensor or not.
```
cathyzhyi, line 68: Also need a guard saying the input tensor is integer type.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,208 @@
...
+    RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+    if (inputType.getRank() != 1)
+      return rewriter.notifyMatchFailure(
+          op, "Input tensor has to be one dimensional.");
+
+    // TODO: Incorporate the weight argument.
```
cathyzhyi, line 74: Can you add another `TODO` saying we also need to verify the input tensor elements are all positive.

include/torch-mlir/Dialect/Torch/Utils/Utils.h
```diff
@@ -9,11 +9,11 @@
 #ifndef TORCHMLIR_DIALECT_TORCH_UTILS_H
 #define TORCHMLIR_DIALECT_TORCH_UTILS_H
 
+#include "mlir/IR/PatternMatch.h"
```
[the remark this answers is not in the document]
vivekkhandelwal1: Because we have a `rewriter` variable of class `PatternRewriter` https://github.com/llvm/torch-mlir/blob/41bcdcdbe910ebcd7b0521ae64effaf42c948988/include/torch-mlir/Dialect/Torch/Utils/Utils.h#L26, if we remove this header file, it throws an error.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
vivekkhandelwal1: I don't think we need to add a guard for this. If we pass a tensor containing negative values as input, the python throws an error for this, saying:
```
RuntimeError: bincount only supports 1-d non-negative integral inputs.
```

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
vivekkhandelwal1: Same explanation as below for this. For this case as well, it throws an error, saying: 
```
RuntimeError: "bincount_cpu" not implemented for 'Float'
```

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
vivekkhandelwal1: I am not able to understand this point. Can you elaborate?

cathyzhyi: I mean you could remove `!inputType.hasStaticShape()` check and handle static and dynamic shapes with the same code.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cathyzhyi: yea, the error you are seeing above is from the TorchScript (baseline) running with the invalid input. As a compiler, we need to guard the invalid code in our code base.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,217 @@
...
+    if (inputType.getRank() != 1)
+      return rewriter.notifyMatchFailure(
+          op, "Input tensor has to be one dimensional.");
+
+    // Check whether the input tensor element type is i64 or not.
+    IntegerType inputIntegerType =
+        inputType.getElementType().cast<IntegerType>();
+    if (inputIntegerType.getWidth() != 64)
+      return rewriter.notifyMatchFailure(
+          op, "Input tensor element type has to be i64.");
```
cathyzhyi, line 80: Please change this to something like `Unimplemented: Integer width not equal to 64 are not supported.` to make it clear if it's an invalid input or unsupported case.

cathyzhyi (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -0,0 +1,219 @@
...
+    Value input = adaptor.self();
+    Value torchTypeInput = op.self();
+    Value minlength = adaptor.minlength();
+    Value weights = adaptor.weights();
+
+    // TODO: Add a check to verify that the input tensor elements are all
+    // non-negative.
+    // Check whether the input is a 1-d tensor of integer type or not.
+    RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+    if (inputType.getRank() != 1 && inputType.isa<mlir::IntegerType>())
```
cathyzhyi, line 73: `inputType.getRank() != 1 || !inputType.getElementType.isa<mlir::IntegerType>()`. Please update the failure message below to include integer tensor type.

cathyzhyi (changes requested):

cathyzhyi (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cathyzhyi: The condition is still incorrect. should fail either `inputType.getRank() != 1` or `!inputType.isa<mlir::IntegerType>()`.

cathyzhyi:
> Hi @cathyzhyi, I have addressed all the comments on this PR. Are we good to merge this PR?

The condition check is still incorrect. We can merge after you fix it.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
vivekkhandelwal1: Oh, I missed this. Sorry for that. Fixed now.

cathyzhyi (approved):

PR #579, opened 2022-02-10
Fix error in RefineTypes for constant alloc ops

ramiro050:
This commit fixes an error in the refine types pass of constant
allocation ops. The function used to set the dtype,
`fillInDtypeGivenDtypeAndDataType`, takes two torch types as arguments,
but a torch type and a standard MLIR type were being passed into it.

This commit also fixes the way the dtype was calculated in
`visitAtenToDtypeOp`. This op was also passing a standard MLIR type as
an argument to the `fillInDtypeGivenDtypeAndDataType`
function. Moreover, since the op `aten.to.dtype` has the dtype
argument as not optional, all that is needed is to match
against the int value to extract the dtype.

### DIFF
--- a/e2e_testing/torchscript/constant_alloc.py
+++ b/e2e_testing/torchscript/constant_alloc.py
@@ -11,6 +11,22 @@ from torch_mlir_e2e_test.torchscript.annotations import annotate_args, export
 
 # ==============================================================================
 
+class ZerosModuleDefaultDtype(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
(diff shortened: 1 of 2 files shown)


cathyzhyi (changes requested):
Thanks for the fix!
Could you also help to add an assert at the entry of 
`fillInDTypeGivenDTypeAndDataType` to check the passed in type is a Torch type. We can use `isa<TorchDialect>(type.getDialect())` to check that.

cathyzhyi (approved):

PR #580, opened 2022-02-11
Modify aten._log_softmax op decomposition for numerical stability.

pashu123:
`aten.log_softmax` is decomposed to be more numerically stable.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -449,7 +449,37 @@ public:
 };
 } // namespace
 
-// Decompose aten.log_softmax op into: log(softmax(x))
+// To avoid overflow we use the following decomposition rule:
+//  x_max = aten.max(x, dim, keepdim=True)[0]
+//  shifted = x - x_max
+//  shifted_logsumexp = aten.log(aten.sum(aten.exp(shifted), dim, keepdim=True))
+//  log_softmax = shifted - shifted_logsumexp
+template <typename OpTy>
+static Value getLogSoftmaxResult(OpTy op, PatternRewriter &rewriter) {
+  Location loc = op.getLoc();
+  Value dim = op.dim();
+  Value self = op.self();
+  BaseTensorType tensorType = self.getType().cast<BaseTensorType>();
+  Value xMax =
+      createMaxAlongDimension(rewriter, loc, op, self, dim, /*keepDim=*/true);
+  if (!xMax)
+    return nullptr;
+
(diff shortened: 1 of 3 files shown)


cathyzhyi (changes requested):
Thanks for being careful about numerical stability!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -468,25 +496,31 @@ class DecomposeAtenLogSoftmaxIntOp
     if (!tensorType.hasDtype() || !tensorType.getDtype().isa<mlir::FloatType>())
       return rewriter.notifyMatchFailure(op, "Only support floating type");
 
-    // softmax(x, dim)
-    Value softmax = rewriter.create<AtenSoftmaxIntOp>(loc, tensorType, self,
-                                                      dim, op.dtype());
-    rewriter.replaceOpWithNewOp<AtenLogOp>(op, op.getType(), softmax);
+    Value logSoftmax = getLogSoftmaxResult(op, rewriter);
```
cathyzhyi, line 499: nit: `logSoftmax` var is not needed.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -468,25 +496,31 @@ class DecomposeAtenLogSoftmaxIntOp
...
+    if (!matchPattern(op.half_to_float(), m_TorchConstantBool(&halfToFloat)))
+      return rewriter.notifyMatchFailure(
+          op, "Expected a boolean value for half_to_float");
+
+    // Currently, setting `halfToFloat` is not supported as the E2E testing for
+    // the same is not present on CPU.
+    if (halfToFloat)
+      return rewriter.notifyMatchFailure(
+          op, "halfToFloat is currently not supported.");
+    Value _logSoftmax = getLogSoftmaxResult(op, rewriter);
```
cathyzhyi, line 522: nit: `_logSoftmax` is not needed

cathyzhyi (approved):
just 2 nits.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ramiro050, line 499: A check is needed to make sure the return of `getLogSoftmaxResult` is not `nullptr`. The same applies for the `Aten_LogSoftmaxOp` version

pashu123: Done, thanks for the catch.

PR #581, opened 2022-02-11
Fix bug for aten_nll_loss op in the refine types pass

pashu123:
The check for `self.hasSizes` was missing before performing `.size()`
operation.

### DIFF
--- a/lib/Dialect/Torch/Transforms/RefineTypes.cpp
+++ b/lib/Dialect/Torch/Transforms/RefineTypes.cpp
@@ -1176,26 +1176,24 @@ ChangeResult TypeAnalyzer::visitAtenNllLossForwardOp(
   auto self = operands[0]->getValue();
   auto outputKnowledge =
       ValueKnowledge::getNotNonePessimisticValueState(op.getContext());
-
-  // Contains Knowledge of shape and dtype for the 1st result.
-  outputKnowledge.dtype = self.dtype;
-  int64_t reduction;
-  unsigned resultRank = self.sizes.size();
-
-  // Contains Knowledge of shape and dtype for the 2nd result.
   auto totalWeightKnowledge =
       ValueKnowledge::getNotNonePessimisticValueState(op.getContext());
-  totalWeightKnowledge.dtype = self.dtype;
-  totalWeightKnowledge.sizes.resize(0, kUnknownSize);
-  totalWeightKnowledge.hasSizes = true;
 
-  if (self.hasSizes &&
-      matchPattern(op.reduction(), m_TorchConstantInt(&reduction))) {
-    if (reduction != Reduction::None)
-      resultRank -= 1;
+  // `AtenNllLossForward` op returns two outputs, output and total_weight.
+  // The rank of the output depends on the reduction parameter and total_weight
+  // is a scalar value.
+  outputKnowledge.dtype = self.dtype;
+  totalWeightKnowledge.dtype = self.dtype;
+  totalWeightKnowledge.hasSizes = true;
+  if (self.hasSizes) {
+    int64_t reduction;
+    if (matchPattern(op.reduction(), m_TorchConstantInt(&reduction))) {
+      outputKnowledge.hasSizes = true;
+      unsigned resultRank = self.sizes.size();
+      if (reduction == Reduction::None)
+        outputKnowledge.sizes.resize(resultRank - 1, kUnknownSize);
+    }
   }
-  outputKnowledge.sizes.resize(resultRank - 1, kUnknownSize);
-  outputKnowledge.hasSizes = true;
   auto resultLattice = getLatticeElement(op.getResult(0)).join(outputKnowledge);
   resultLattice |=
       getLatticeElement(op.getResult(1)).join(totalWeightKnowledge);


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1175,26 +1175,28 @@ ChangeResult TypeAnalyzer::visitAtenNllLossForwardOp(
...
-  if (self.hasSizes &&
-      matchPattern(op.reduction(), m_TorchConstantInt(&reduction))) {
-    if (reduction != Reduction::None)
-      resultRank -= 1;
+    // Contains Knowledge of shape for the both results.
+    totalWeightKnowledge.sizes.resize(0, kUnknownSize);
+    totalWeightKnowledge.hasSizes = true;
+
+    if (self.hasSizes &&
+        matchPattern(op.reduction(), m_TorchConstantInt(&reduction))) {
```
cathyzhyi, line 1193: set `outputKnowledge.hasSizes` as true only when `reduction` is constant. When it's not constant hasSizes needs to be false.

cathyzhyi (changes requested):

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1176,26 +1176,27 @@ ChangeResult TypeAnalyzer::visitAtenNllLossForwardOp(
...
   auto totalWeightKnowledge =
       ValueKnowledge::getNotNonePessimisticValueState(op.getContext());
+
+  // Contains Knowledge of dtype for both results.
+  outputKnowledge.dtype = self.dtype;
   totalWeightKnowledge.dtype = self.dtype;
-  totalWeightKnowledge.sizes.resize(0, kUnknownSize);
-  totalWeightKnowledge.hasSizes = true;
+  if (self.hasSizes) {
+    int64_t reduction;
```
cathyzhyi, line 1186: Please move `reduction` and `resultRank` below as they are related to `outputKnowledge`. It's surprising to have `totalWeightKnowledge` in between.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1176,26 +1176,27 @@ ChangeResult TypeAnalyzer::visitAtenNllLossForwardOp(
...
+    unsigned resultRank = self.sizes.size();
 
-  if (self.hasSizes &&
-      matchPattern(op.reduction(), m_TorchConstantInt(&reduction))) {
-    if (reduction != Reduction::None)
-      resultRank -= 1;
+    // Contains Knowledge of shape for the both results.
```
cathyzhyi, line 1189: What does this comment mean?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1176,26 +1176,27 @@ ChangeResult TypeAnalyzer::visitAtenNllLossForwardOp(
...
       ValueKnowledge::getNotNonePessimisticValueState(op.getContext());
+
+  // Contains Knowledge of dtype for both results.
+  outputKnowledge.dtype = self.dtype;
   totalWeightKnowledge.dtype = self.dtype;
-  totalWeightKnowledge.sizes.resize(0, kUnknownSize);
-  totalWeightKnowledge.hasSizes = true;
+  if (self.hasSizes) {
+    int64_t reduction;
+    unsigned resultRank = self.sizes.size();
```
cathyzhyi, line 1187: `resultRank` is only needed in `matchPattern` if block. Please move to inside the block to make the scope more clear.

cathyzhyi (changes requested):

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1176,26 +1176,27 @@ ChangeResult TypeAnalyzer::visitAtenNllLossForwardOp(
...
-      matchPattern(op.reduction(), m_TorchConstantInt(&reduction))) {
-    if (reduction != Reduction::None)
-      resultRank -= 1;
+    // Contains Knowledge of shape for the both results.
+    // totalWeight is a 0d tensor.
+    totalWeightKnowledge.hasSizes = true;
+
+    if (matchPattern(op.reduction(), m_TorchConstantInt(&reduction))) {
+      if (reduction != Reduction::None)
+        resultRank -= 1;
```
ramiro050, line 1195: I think I had made a comment about this last time. I still don't understand why we subtract one here. From the documentation, (https://pytorch.org/docs/stable/generated/torch.nn.NLLLoss.html#torch.nn.NLLLoss), the input can be a rank-1 or rank-2 tensor. If the tensor is rank-1, and `reduction != None`, then `outputKnowledge.sizes` will be resized to size `-1`, which is not right. If the reduction is not `None` the result should be a scalar, so `resultRank = 0`. 

I would flip the condition in the `if` and do this instead

```cpp
outputKnowledge.hasSizes = true;
if (reduction == Reduction::None) {
  outputKnowledge.sizes.resize(resultRank - 1, kUnknownSize);
}
```

leaving `outputKnowledge.sizes` empty represents the scalar case.

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/RefineTypes.cpp
pashu123: I think I missed this last time. Thanks for pointing this out and Sorry for the re-review!

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1176,26 +1176,24 @@ ChangeResult TypeAnalyzer::visitAtenNllLossForwardOp(
...
+  totalWeightKnowledge.dtype = self.dtype;
+  if (self.hasSizes) {
+    int64_t reduction;
+    if (matchPattern(op.reduction(), m_TorchConstantInt(&reduction))) {
+      outputKnowledge.hasSizes = true;
+      unsigned resultRank = self.sizes.size();
+      if (reduction == Reduction::None)
+        outputKnowledge.sizes.resize(resultRank - 1, kUnknownSize);
+    }
+    totalWeightKnowledge.hasSizes = true;
```
ramiro050, line 1195: nit: I think this will always be true even if `self.hasSizes != true`, right? You could move this line up so to where the `dtype` is assigned to `totalWeightKnowledge` so that all the changes to `totalWeightKnowledge` are together.

ramiro050 (approved):
LGTM, just one nit

lib/Dialect/Torch/Transforms/RefineTypes.cpp
pashu123: Yes, thanks!

Issue #555, opened 2022-02-01
Missing support for importing node prim::Constant with list type


henrytwo:
I was trying out torch-mlir with LTC using the TorchScript backend but ran into an error when performing a basic transpose operation using a modified version of the `tanh` example script:

```
LAZY GRAPH
graph(%p0 : Tensor):
  %1 : int[] = prim::Constant[value=[1, 0]]()
  %2 : int[] = prim::Constant[value=[1, 0]]()
  %3 : Tensor = aten::permute(%p0, %2)
  return (%3)

error: unhandled prim::Constant node: %1 : int[] = prim::Constant[value=[1, 0]]()

Traceback (most recent call last):
  File "repro_prim_list_error.py", line 27, in <module>
    mb.import_function(script_function)
RuntimeError: see diagnostics
```

To reproduce:
```python
import lazy_tensor_core as ltc
import torch
from torch._C import CompilationUnit
from torch_mlir.dialects.torch.importer.jit_ir import ModuleBuilder

ltc._LAZYC._ltc_init_ts_backend()
device = torch.device('lazy')
dtype = torch.float32

x = torch.randn((4, 5), device=device, dtype=dtype)

# Capture lazy computation and convert to TorchScript IR
graph_str = ltc._LAZYC._get_ltc_tensors_backend([
    x.T
])

print("LAZY GRAPH")
print(graph_str)
graph = torch._C.parse_ir(graph_str)

# Create a torch.jit.ScriptFunction out of the graph
cu = CompilationUnit()
func_name = 'my_method'
script_function = cu.create_function(func_name, graph)

mb = ModuleBuilder()
mb.import_function(script_function)

print(mb.module)
```

It looks like the error stems from the `import_function` call of `ModuleBuilder`, which is part of `torch-mlir`. Currently my hacky solution around this problem is to make the SSA values corresponding to primitive lists a parameter, but I'm wondering if there is a _better_ way to do this.

cathyzhyi:
Hey Henry, thanks for your interest in the project! 
[...]

PR #584, opened 2022-02-11
Added support for importing node prim::Constant with list type

henrytwo:
Prior to this commit, importing a `prim::Constant` node with list type would result in an error since it was not supported. `ivalue_importer::importIValue` was modified to return the MlirValue corresponding to the root so its parent operation could be extracted.

Issue https://github.com/llvm/torch-mlir/issues/555

(cc: @cathyzhyi)

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/node_importer.cpp
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/node_importer.cpp
@@ -12,6 +12,8 @@
 
 #include <unordered_map>
 
+#include "class_annotator.h"
+#include "ivalue_importer.h"
 #include "mlir_utils.h"
 
 #include "mlir-c/BuiltinAttributes.h"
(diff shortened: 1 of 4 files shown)


python/torch_mlir/dialects/torch/importer/jit_ir/csrc/node_importer.cpp
```diff
@@ -180,6 +182,14 @@ void NodeImporter::importNode(Node *node, MlirBlock appendToBlock) {
...
               "value",
               mlirFlatSymbolRefAttrGet(context, toMlirStringRef(symName))));
+    } else if (output->type()->cast<c10::ListType>()) {
+      ClassAnnotator dummyAnnotator;
+      MlirValue listValue = importIValue(node->ival(c10::attr::value),
+                                         appendToBlock,
+                                         context,
+                                         dummyAnnotator);
+      mapResults(node, mlirOpResultGetOwner(listValue));
+      return; // Early return, since `importIValue` already added op to block
```
cathyzhyi, line 192: nit: The comment should be a full sentence with period at the end.

test/python/importer/jit_ir/node_import/prim.py
```diff
@@ -14,12 +15,16 @@
 mb = ModuleBuilder()
 
 
+# Import TorchScript IR string as ScriptFunction
+def import_ts_ir(ts_ir_str):
+    cu = CompilationUnit()
+    mb.import_function(cu.create_function('main', torch._C.parse_ir(ts_ir_str)))
```
cathyzhyi, line 21: nit: can we pass the function name as argument? And give a more descriptive name according to the specific test. In this case, it can be `__torch__.prim_Constant_list`.

cathyzhyi (changes requested):
Looks mostly good! Thanks!

cathyzhyi (approved):

PR #585, opened 2022-02-11
[LINALG] Add E2E support for `aten.eq.int` op

Shukla-Gaurav:
- This commit adds lowering of `aten.eq.int` op as a part of
  `convert-torch-to-std` pass.
- It also refactors the code for binary comparison ops lowering.
    
Note: The build will fail as it is dependent on bool return support  addressed in #569.

### DIFF
--- a/e2e_testing/torchscript/scalar_comparison.py
+++ b/e2e_testing/torchscript/scalar_comparison.py
@@ -1,0 +1,71 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+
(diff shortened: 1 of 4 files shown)


e2e_testing/torchscript/scalar_comparison.py
```diff
@@ -0,0 +1,70 @@
...
+        ([], torch.int64, True),
+        ([], torch.int64, True),
+    ])
+    def forward(self, lhs, rhs):
+        return int(lhs) > int(rhs)
+
+
+@register_test_case(module_factory=lambda: GtIntModule())
+def GtIntModule_basic(module, tu: TestUtils):
+    module.forward(torch.randint(-100, 100,()), torch.randint(-100, 100,()))
```
ramiro050, line 70: nit: the spacing between the `randint` arguments is not consistent in all the `randint`s in these changes

ramiro050 (approved):
LGTM! Just a small style comment

PR #587, opened 2022-02-11
Add folder for aten.gt/lt.float

cathyzhyi:


cathyzhyi:
> @cathyzhyi Should we add e2e tests?

This is folder not OpConverter so e2e tests are not needed IMO since the op would be folded away by canonicalizer. Unit test should be sufficient here.

ramiro050 (approved):
LGTM!

PR #588, opened 2022-02-12
[LINALG] Fix linalg generic result type argument in TorchToLinalg

ramiro050:
Some of the lowerings use the result type obtained from the op itself
to tell the `linalg::GenericOp` what the type of the result should be
rather than using the type of the result tensor given to the
`linalg::GenericOp`. This becomes a problem when the result type of
the op has static size information and the result tensor used in
`linalg::GenericOp` has dynamic dimensions, for `linalg::GenericOp`
expects the result type to be equal to the type of the output tensor.

This commit replaces the use of the result type from the op itself
with the type of the result tensor passed to `linalg::GenericOp`.

In order to not create too many dynamic/static versions of the same
e2e test, e2e tests have only been added to the ops that currently
fail when used with static sizes.

### DIFF
--- a/e2e_testing/torchscript/basic.py
+++ b/e2e_testing/torchscript/basic.py
@@ -480,6 +480,26 @@ def GatherModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
+class GatherStaticModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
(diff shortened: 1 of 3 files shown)


cathyzhyi (approved):
Thanks!

ramiro050:
The CI was failing because the lowering for the gather op did not cast the result of the entire lowering to the op result type. I've added the cast

PR #590, opened 2022-02-12
Bump llvm

nirvedhmeshram:
I am starting this patch so that I can share how my changes are failing using the CI and get some help on it.

### DIFF
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -11,8 +11,10 @@
 
 #include "../PassDetail.h"
 #include "mlir/Dialect/Arithmetic/IR/Arithmetic.h"
+#include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Math/IR/Math.h"
+#include "mlir/Dialect/StandardOps/IR/Ops.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Tensor/Utils/Utils.h"
 #include "mlir/Dialect/Traits.h"
@@ -78,7 +80,8 @@ static Value toPositiveDimDynamic(OpBuilder &b, Location loc, Value dim,
       b.create<arith::ConstantOp>(loc, b.getZeroAttr(inputRank.getType()));
   Value predDimGEZero =
       b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sge, dim, cst0);
-  Value dimInt = b.create<SelectOp>(loc, predDimGEZero, dim, dimAddInputRank);
+  Value dimInt =
+      b.create<arith::SelectOp>(loc, predDimGEZero, dim, dimAddInputRank);
   return dimInt;
 }
 
@@ -91,12 +94,12 @@ static void assertIsValidDim(OpBuilder &b, Location loc, Value dim,
       b.create<arith::ConstantOp>(loc, b.getZeroAttr(inputRank.getType()));
   Value predGEZero =
       b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sge, dim, cst0);
-  b.create<AssertOp>(loc, predGEZero,
-                     b.getStringAttr("dim must be greater or equal to zero"));
+  b.create<cf::AssertOp>(
+      loc, predGEZero, b.getStringAttr("dim must be greater or equal to zero"));
   Value predLTInputRank =
       b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::slt, dim, inputRank);
-  b.create<AssertOp>(loc, predLTInputRank,
-                     b.getStringAttr("dim must be smaller than inputRank"));
+  b.create<cf::AssertOp>(loc, predLTInputRank,
+                         b.getStringAttr("dim must be smaller than inputRank"));
 }
 
 // Hack to deal with the Torch list type arguments which is not supported end
@@ -147,8 +150,8 @@ static void checkDimEqualHelper(OpBuilder &b, Location loc, Value lhsDim,
   Value rhsDimInt = rhsType.isIndex() ? castIndexToInt(b, loc, rhsDim) : rhsDim;
   Value contractingDimEqual = b.create<arith::CmpIOp>(
       loc, arith::CmpIPredicate::eq, lhsDimInt, rhsDimInt);
-  b.create<AssertOp>(loc, contractingDimEqual,
-                     b.getStringAttr("mismatching contracting dimension"));
+  b.create<cf::AssertOp>(loc, contractingDimEqual,
+                         b.getStringAttr("mismatching contracting dimension"));
 }
 
 static SmallVector<Value> getTensorSizesUntilDim(OpBuilder &b, Location loc,
@@ -471,8 +474,8 @@ public:
         rewriter.create<arith::ConstantOp>(loc, IntegerAttr::get(intType, 1));
     Value groupEqual1 = rewriter.create<arith::CmpIOp>(
         loc, arith::CmpIPredicate::eq, groups, c1);
-    rewriter.create<AssertOp>(loc, groupEqual1,
-                              rewriter.getStringAttr("expect groups to be 1"));
+    rewriter.create<cf::AssertOp>(
+        loc, groupEqual1, rewriter.getStringAttr("expect groups to be 1"));
 
     // Pad the input tensor according to padding.
     SmallVector<int64_t, 4> paddingIncludingNC = {0, 0};
@@ -581,15 +584,16 @@ static void createLinalgPayloadCalculationForGatherOps(
   Value indexLTInputDim = b.create<arith::CmpIOp>(
       loc, arith::CmpIPredicate::slt, index,
       castIndexToInt(b, loc, getDimOp(b, loc, input, dim)));
-  b.create<AssertOp>(loc, indexLTInputDim,
-                     b.getStringAttr("index must be smaller than dim size"));
+  b.create<cf::AssertOp>(
+      loc, indexLTInputDim,
+      b.getStringAttr("index must be smaller than dim size"));
 
   // Assert index >= 0
   Value cst0 = b.create<arith::ConstantOp>(loc, b.getZeroAttr(index.getType()));
   Value indexGEThanZero =
       b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sge, index, cst0);
-  b.create<AssertOp>(loc, indexGEThanZero,
-                     b.getStringAttr("index must be larger or equal to 0"));
+  b.create<cf::AssertOp>(loc, indexGEThanZero,
+                         b.getStringAttr("index must be larger or equal to 0"));
 
   Value extract = b.create<tensor::ExtractOp>(loc, input, indices);
   b.create<linalg::YieldOp>(loc, extract);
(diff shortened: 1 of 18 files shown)


nirvedhmeshram:
This builds now, however fails in lowering from linalg to refbackend. This is because the names and organization of low level passes have changed. @cathyzhyi @silvasean could you please take a look at my very last commit and give some inputs on how to change it. Currently I see an error like this
```
JIT session error: Symbols not found: [ memrefCopy ]
Expected<T> must be checked before access or destruction.
Unchecked Expected<T> contained error:
Failed to materialize symbols: { (main, { _mlir_forward, forward, _mlir_ciface_forward, refbackend_consume_func_return_mrf32, _mlir__mlir_ciface_forward, _mlir_reeturn_mrf32 }) }Aborted
```
But only on some e2e tests while other e2e tests pass.

nirvedhmeshram:
I cant run all the tests because it just crashes with the error I mentioned above. When I try individual tests. `--filter Threshold` and `--filter TModule` passed but something like `--filter Conv2d` and `--filter HBC`  failed with the above error.
Please let me know if any specific tests you would like me to try.

nirvedhmeshram:
Thanks for looking into this @cathyzhyi . That instruction is valid in llvm IR and should get lowered right?
https://github.com/nirvedhmeshram/llvm-project/blob/main/mlir/test/Target/LLVMIR/llvmir-intrinsics.mlir#L346

cathyzhyi:
> https://github.com/nirvedhmeshram/llvm-project/blob/main/mlir/test/Target/LLVMIR/llvmir-intrinsics.mlir#L346

Ha, that's right. seems it should be lowered by `mlir-to-llvmir`?

jpienaar:
That change only added a new lowering, so all the unit tests are passing? (Just in case they show a simpler repro for the failure)

nirvedhmeshram:
I get ```
error: 'mlir-to-llvmir' does not refer to a registered pass or pass pipeline
```
Other thing to note is that in the file I shared its passed to `mlir-translate` not `mlir-opt`

**Comment by nirvedhmeshram** [comment]
@jpienaar So you mean end to end it is not guaranteed we have a flow? Since they are only showing unit tests?

**Comment by cathyzhyi** [comment]
> I get
> 
> ```
> error: 'mlir-to-llvmir' does not refer to a registered pass or pass pipeline
> ```
> 
> Other thing to note is that in the file I shared its passed to `mlir-translate` not `mlir-opt`

yea, `mlir-to-llvmir` happens in ExecutionEngine https://github.com/llvm/llvm-project/blob/84fe34a0b7fdd7bbf179981d1583693d5d5ec68b/mlir/lib/ExecutionEngine/ExecutionEngine.cpp#L245. Seems it's gonna be lowered but somehow ExecutionEngine can't resolve the intrinsic call symbol.

**Comment by jpienaar** [comment]
> @jpienaar So you mean end to end it is not guaranteed we have a flow? Since they are only showing unit tests?

I was just wondering about the unit tests as those would be smaller, so getting those all running first would highlight changes needed with a smaller input/pass surface. If those are working all, then one could tackle larger ones with hopefully most cases resorted. It just scopes the changes needed at each stage.

**Comment by nirvedhmeshram** [comment]
@jpienaar Yes unit tests are all running. Moreover the e2e tests that fail with the error I can run something like this without it giving any errors
```./build/bin/torch-mlir-opt -pass-pipeline='builtin.func(scf-bufferize),builtin.func(linalg-bufferize),func-bufferize,arith-bufferize,builtin.func(tensor-bufferize),builtin.func(finalizing-bufferize),refback-munge-calling-conventions,refback-insert-rng-globals,builtin.func(convert-linalg-to-loops),builtin.func(lower-affine),convert-scf-to-cf,builtin.func(refback-expand-ops-for-llvm),builtin.func(arith-expand),builtin.func(convert-math-to-llvm),convert-linalg-to-llvm,convert-memref-to-llvm,builtin.func(convert-arith-to-llvm),convert-std-to-llvm,reconcile-unrealized-casts' /tmp/Conv2dWithPaddingDilationStrideStaticModule.mlir```

**Comment by cathyzhyi** [comment]
> 

yea, the issue is only exposed when the IR output is processed by ExecutionEngine.  could you try if you revert https://reviews.llvm.org/D116099 would the test pass?

**Comment by nirvedhmeshram** [comment]
Actually that is not true, although it gives an IR, the IR looks like this
```module attributes {llvm.data_layout = "", torch.debug_module_name = "Conv2dWithPaddingDilationStrideStaticModule"} {
  llvm.func @abort()
  llvm.func @memrefCopy(i64, !llvm.ptr<struct<(i64, ptr<i8>)>>, !llvm.ptr<struct<(i64, ptr<i8>)>>)
more stuff
```
Is that abort function supposed to be there?

nirvedhmeshram:
> yea, the issue is only exposed when the IR output is processed by ExecutionEngine. could you try if you revert
I can try

cathyzhyi:
> memrefCopy

the `abort` is just a function definition. We might not call it during execution.

nirvedhmeshram:
@cathyzhyi I reverted the patch you mentioned and it did not change the behavior

cathyzhyi:
@nirvedhmeshram I am able to reproduce the failure with a unit test
```
torch-mlir-opt -convert-scf-to-cf -convert-arith-to-llvm -convert-memref-to-llvm -convert-std-to-llvm -reconcile-unrealized-casts $path_to_torch_mlir_src/torch-mlir/external/llvm-project/mlir/test/mlir-cpu-runner/copy.mlir | $path_to_torch_mlir_src/torch-mlir/build/bin/mlir-cpu-runner -e main -entry-point-result=void
```
JIT session error: Symbols not found: [ _mlir_ciface_print_memref_f32, memrefCopy ]
Error: Failed to materialize symbols: { (main, { main, _mlir_print_memref_f32, print_memref_f32, _mlir_main }) }

This is a unit test in mlir here https://github.com/llvm/llvm-project/blob/84fe34a0b7fdd7bbf179981d1583693d5d5ec68b/mlir/test/mlir-cpu-runner/copy.mlir#L1-L4. Notice the original test specified some path to shared libs. After removing those paths, the `memrefCopy` symbol can't be found. I think we are just missing the paths to shared libs in our e2e test when using the mlir execution engine.

nirvedhmeshram:
That's great that we have isolated the problem. I am not sure how to provide the shared libs to e2e tests though. Dont have much experience with something like this.

cathyzhyi:
> That's great that we have isolated the problem. I am not sure how to provide the shared libs to e2e tests though. Dont have much experience with something like this.

I don't either :D  @silvasean should be able to help us with this when he's back tmr.

silvasean:
It looks like it is expecting this function to be linked in: https://github.com/llvm/llvm-project/blob/93d1a623cecb6f732db7900baf230a13e6ac6c6a/mlir/lib/ExecutionEngine/CRunnerUtils.cpp#L48

@MaheshRavishankar suggested that this issue is coming from linalg.copy getting lowered through memref.copy which then calls into the CRunnerUtils.cpp:memrefCopy function. The workaround is to lower linalg.copy to linalg.generic ops. Mahesh, can you link to the code in IREE that is used to convert linalg.copy to linalg.generic so we can steal it?

MaheshRavishankar:
This should be the code. 

https://github.com/google/iree/blob/db0284d993840944cc12334ec300d1faf2db505b/iree/compiler/Codegen/Utils/Utils.cpp#L601

Do you mind if you just land this in core?

nirvedhmeshram:
Thanks @silvasean and @MaheshRavishankar 
I can land this in core. Should I first just try to fix it for torch-mlir and try to get this PR merged in before adding it into the core?

nirvedhmeshram:
Also, just went through one of the failing IR's in linalg and it does not have `linalg.copy`, could there be other ops that end up making `memref.copy` ?

nirvedhmeshram:
For example
 ```
%28 = arith.cmpi sge, %27, %19 : index
%29 = arith.select %28, %27, %19 : index
%30 = arith.subi %29, %19 : index
 %31 = tensor.extract_slice %arg1[%19] [%30] [1] : tensor<?xi32> to tensor<?xi32>
```
After `builtin.func(linalg-bufferize)'` pass becomes
```
%30 = arith.cmpi sge, %29, %21 : index
%31 = arith.select %30, %29, %21 : index
%32 = arith.subi %31, %21 : index   
 %33 = memref.alloc(%32) : memref<?xi32>
 %34 = memref.subview %1[%21] [%32] [1] : memref<?xi32> to memref<?xi32, #map2>
 memref.copy %34, %33 : memref<?xi32, #map2> to memref<?xi32>
 ```

MaheshRavishankar:
> For example
> 
> ```
> %28 = arith.cmpi sge, %27, %19 : index
> %29 = arith.select %28, %27, %19 : index
> %30 = arith.subi %29, %19 : index
> %31 = tensor.extract_slice %arg1[%19] [%30] [1] : tensor<?xi32> to tensor<?xi32>
> ```
> 
> After `builtin.func(linalg-bufferize)` pass becomes
> 
> ```
> %30 = arith.cmpi sge, %29, %21 : index
> %31 = arith.select %30, %29, %21 : index
> %32 = arith.subi %31, %21 : index   
> %33 = memref.alloc(%32) : memref<?xi32>
> %34 = memref.subview %1[%21] [%32] [1] : memref<?xi32> to memref<?xi32, #map2>
> memref.copy %34, %33 : memref<?xi32, #map2> to memref<?xi32>
> ```

Yeah, bufferization by default adds `memref.copy`. I dont know what `linalg-bufferize` calls, but if this is the upstream bufferization you need to update the callbacks to use the right allocation/deallocation/copy functions (like here https://github.com/google/iree/blob/db0284d993840944cc12334ec300d1faf2db505b/iree/compiler/Codegen/LLVMCPU/Passes.cpp#L284)

silvasean:
> For example
> 
> ```
> %28 = arith.cmpi sge, %27, %19 : index
> %29 = arith.select %28, %27, %19 : index
> %30 = arith.subi %29, %19 : index
> %31 = tensor.extract_slice %arg1[%19] [%30] [1] : tensor<?xi32> to tensor<?xi32>
> ```
> 
> After `builtin.func(linalg-bufferize)` pass becomes
> 
> ```
> %30 = arith.cmpi sge, %29, %21 : index
> %31 = arith.select %30, %29, %21 : index
> %32 = arith.subi %31, %21 : index   
> %33 = memref.alloc(%32) : memref<?xi32>
> %34 = memref.subview %1[%21] [%32] [1] : memref<?xi32> to memref<?xi32, #map2>
> memref.copy %34, %33 : memref<?xi32, #map2> to memref<?xi32>
> ```

Ugh, yeah, that's tricky. Could you try to write a pass to convert memref.copy to linalg.copy? You would insert the pass [here](https://github.com/llvm/torch-mlir/blob/869daf3c227d0e09604b93b56fbf45d30740af81/python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py#L181). linalg-to-loops should take care of lowering it then. I would call the pass `refback-munge-memref-copy`.

nirvedhmeshram:
Sounds good, yes I can do that.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -30,6 +32,7 @@ using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 using namespace mlir::torch::TorchConversion;
 using namespace mlir::torch::torch_upstream; // For ScalarType and type
+using namespace mlir::cf;
```
cathyzhyi, line 35: I would prefer to put the dialect prefix before the operation.

lib/RefBackend/RefBackend.cpp
```diff
@@ -354,3 +357,65 @@ std::unique_ptr<OperationPass<FuncOp>>
...
+
+class MungeMemrefCopy : public MungeMemrefCopyBase<MungeMemrefCopy> {
+  void getDependentDialects(DialectRegistry &registry) const override {
+    registry.insert<linalg::LinalgDialect>();
+  }
+
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+    RewritePatternSet patterns(&getContext());
+    // ConversionTarget target(*context);
```
cathyzhyi, line 408: nit: remove this line.

cathyzhyi (approved):
Great work!! Thanks a lot for tackle this complicated llvm bump. 
Just 2 nits.

PR #591, opened 2022-02-14
[LINALG] Fix name conflict of `self` keyword.

Shukla-Gaurav:
- The `self` name is being used as a keyword argument to the
  `torch.ops.aten.nll_loss_backward` function call, which produces
  name-conflict error with the python keyword `self` which is pointer to
  the current object.
- This commit fixes this issue by replacing the keyword argument by
  positional argument.


pashu123 (approved):
LGTM!

cathyzhyi (approved):

PR #594, opened 2022-02-14
[LINALG] Decompose `aten.batch_norm` into `aten.native_batch_norm`

Shukla-Gaurav:
- This commit decomposes the `aten.batch_norm` op into the
  `aten.native_batch_norm` op, instead of lowering it to the
  `linalg.generic` op.
- It also adds run-time asserts in the `aten.native_batch_norm` lowering
  to make sure that the shape of the weight, bias, running_mean, and
  running_var must match the num of features.
- Since the `aten.native_batch_norm` op is not supported at TOSA backend,
  all the modules that are dependent on the `aten.native_batch_norm` op
  will fail and therefore they should be removed from the TOSA `passing`
  set.
- It also moves `checkNotNone` to utility.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -849,6 +851,37 @@ class DecomposeAtenLayerNormOp : public OpRewritePattern<AtenLayerNormOp> {
...
+    Type emptyType = inputType.getWithSizesAndDtype({0}, inputType.getDtype());
+
+    // The first output tensor of the `AtenNativeBatchNormOp` is essentially
+    // `AtenBatchNormOp` result.
+    auto nativeBatchNorm = rewriter.create<AtenNativeBatchNormOp>(
+        op.getLoc(), inputType, /*meanType=*/emptyType,
+        /*invStdType=*/emptyType, op.input(), op.weight(), op.bias(),
+        op.running_mean(), op.running_var(), op.training(), op.momentum(),
+        op.eps());
+    rewriter.replaceOpWithNewOp<TensorStaticInfoCastOp>(
```
cathyzhyi, line 878: I think the cast is not needed

cathyzhyi (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1017,18 +1061,22 @@ class DecomposeAtenNativeBatchNormOp
     // 3. output = normalizedInput * weight + bias
     Value batchNormOutput = normalizedInput;
     if (!weight.getType().isa<Torch::NoneType>()) {
```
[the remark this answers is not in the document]
Shukla-Gaurav: Should we also make the prototype a bit simpler? like `static bool isNone(Value v)`? 
As it will give the caller an option whether you want to go ahead with none value or return failure in the case of none value.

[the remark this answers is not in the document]
Shukla-Gaurav: This can be handled later.

Shukla-Gaurav:
@cathyzhyi @ramiro050 All the comments has been addressed. Thanks :)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1096,19 +1143,23 @@ class DecomposeAtenNativeBatchNormOp
...
+      dim0EqualsNumFeatures(weight);
+
       weight = rewriter.create<AtenViewOp>(loc, reshapeType, weight,
                                            runningStatsSizeList);
       batchNormOutput = rewriter.create<AtenMulTensorOp>(
           loc, batchNormOutput.getType(), batchNormOutput, weight);
     }
-    if (!bias.getType().isa<Torch::NoneType>()) {
-      // Rank of `bias` must be exactly 1.
+    if (!failed(checkNotNone(rewriter, op, bias))) {
```
cathyzhyi, line 1157: I think just checking `bias.getType().isa<BaseTensorType>()` would be more straight forward.

cathyzhyi: We would need another guard check `bias.getType().isa<OptionalType>()` and failure return because OptionalType can't be handled at all while with the current code, OptionalType is handled the same way as NoneType.

cathyzhyi (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
Shukla-Gaurav: Anyway I have kept the `checkNotNone` as a utility.

cathyzhyi:
@sjarus @anupgangwar FYI, `batch_norm` now goes through `native_batch_norm` so TOSA backend needs to change as well.

anupgangwar:
@cathyzhyi Thanks for letting us know. What is the rationale for this change?

cathyzhyi:
> @cathyzhyi Thanks for letting us know. What is the rationale for this change?

So that we could just implement `native_batch_norm` and decompose `batch_norm` to `native_batch_norm`. Otherwise, both ops need to be implement separately.

anupgangwar:
ok, I will take a look and make the changes. It would have been good to have gotten a heads up on these changes though since it removes passing tests. May be that last bit is a comment for @Shukla-Gaurav.

Shukla-Gaurav:
> ok, I will take a look and make the changes. It would have been good to have gotten a heads up on these changes though since it removes passing tests. May be that last bit is a comment for @Shukla-Gaurav.

@anupgangwar Yeah, I should have mentioned that. Sorry for that.

e2e_testing/torchscript/xfail_sets.py
```diff
@@ -85,15 +85,11 @@
...
     "Conv2dWithPaddingDilationStrideStaticModule_basic",
-    "BatchNorm1DModule_basic",
-    "BatchNorm2DModule_basic",
-    "BatchNorm3DModule_basic",
     "FlattenStaticModule_basic",
     "FlattenRank0Module_basic",
     "ElementwiseFlattenBroadcastModule_basic",
     "SquareModule_basic",
     "MaxPool2dStaticModule_basic",
-    "ResNet18StaticModule_basic",
```
silvasean, line 96: Hi @Shukla-Gaurav, I'm not sure you realized that this list is a list of *passing* tests, so this line *removes* a passing test replacing it with one that's broken! It's not ok to unilaterally turn off passing tests! Please revert this PR!

cc @anupgangwar @sjarus

Shukla-Gaurav: @silvasean I understand that it was a mistake to push the changes without a proper discussion with TOSA team, really sorry for that! 

either I can revert this PR right away to make the resnet functional,
or I can wait till we fix it through assert based on the discussion at https://github.com/llvm/torch-mlir/issues/618 
What do you think @silvasean ?

silvasean: We should revert it.

Shukla-Gaurav: Sure.

PR #597, opened 2022-02-15
[LINALG] Add value tensor variant to `bernoulli_.float`

ramiro050:
This commit adds the op `PseudoAtenBernoulliFloatOp` that represents
`AtenBernoulli_FloatOp` without the underscore. This is needed to make
sure that the `ReduceOpVariants` pass turns the in-place op into an op
that takes value tensors as inputs, otherwise the
`MaximizeValueSemantics` pass will not be able to add value semantics
correctly.

### DIFF
--- a/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
+++ b/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
@@ -126,14 +126,21 @@ public:
       : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {}
   LogicalResult matchAndRewrite(Operation *op,
                                 PatternRewriter &rewriter) const override {
-    if (!isa<AtenUniform_Op>(op))
+    Location loc = op->getLoc();
+    Operation *newOp;
+    if (isa<AtenUniform_Op>(op)) {
+      newOp = rewriter.create<PseudoAtenUniformOp>(loc, op->getResultTypes(),
+                                                   op->getOperands());
+    } else if (isa<AtenBernoulli_FloatOp>(op)) {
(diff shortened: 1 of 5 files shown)


lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -126,11 +126,16 @@ class ReduceNonValueSemanticOps : public RewritePattern {
       : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {}
   LogicalResult matchAndRewrite(Operation *op,
                                 PatternRewriter &rewriter) const override {
-    if (!isa<AtenUniform_Op>(op))
+    Operation *newOp;
+    if (isa<AtenUniform_Op>(op))
```
cathyzhyi, line 130: nit: {} is needed for if blocks with multiple lines.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -126,11 +126,16 @@ class ReduceNonValueSemanticOps : public RewritePattern {
...
+      newOp = rewriter.create<PseudoAtenUniformOp>(
+          op->getLoc(), op->getResultTypes(), op->getOperands());
+    else if (isa<AtenBernoulli_FloatOp>(op))
+      newOp = rewriter.create<PseudoAtenBernoulliFloatOp>(
+          op->getLoc(), op->getResultTypes(), op->getOperands());
```
cathyzhyi, line 135: nit: this was in the existing code as well. Can you help to assign `op->getLoc()` to a `loc` variable.

cathyzhyi (approved):
LGTM. Just 2 nits.

PR #602, opened 2022-02-15
Add static type information support to `aten.mm`

ramiro050:
This commit adds static type information support to `aten.mm`. This is
needed for the forward pass of Bert training.


cathyzhyi:
@silvasean We think would be safe to infer the shape in case where the operands shape are statically known to be valid. Do you have concern about that? Just want to make sure our understanding about the long comments there is correct.

silvasean:
The scary comment is about doing something like `lhs.sizes[1]` but lhs.sizes is empty (such as for a rank-0 LHS), which will segfault the compiler. It looks like Ramiro is guarding that case correctly.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -822,32 +822,19 @@ ChangeResult TypeAnalyzer::visitAtenMmOp(
   auto &rhs = operands[1]->getValue();
   auto knowledge =
       ValueKnowledge::getNotNonePessimisticValueState(op->getContext());
+
+  // `aten.mm` expects both operands to be rank-2 tensors.
+  if (lhs.sizes.size() != 2 || rhs.sizes.size() != 2)
```
silvasean, line 827: I think that lhs.sizes.size() > 0 implies lhs.hasSizes in practice for our code, but you might want to mention in a comment that you are relying on that (or just check hasSizes explicitly)

ramiro050:
I've added the `hasSizes` check. @silvasean, @cathyzhyi PTAL.

cathyzhyi (approved):

silvasean (approved):

PR #603, opened 2022-02-16
Cmake build commands fix.

pashu123:
The external projects torch-mlir and torch-mlir-dialects should be
placed inside double quotes.


cathyzhyi (approved):
Thanks!

PR #605, opened 2022-02-16
[LINALG] Add decomposition of `aten.dropout` op

Shukla-Gaurav:
- This commit adds decomposition of `aten.dropout` op. It also covers the
  training mode of the same op.
- It also adds lowering of `aten.sub.float` op.

### DIFF
--- a/e2e_testing/torchscript/basic.py
+++ b/e2e_testing/torchscript/basic.py
@@ -939,7 +939,27 @@ def TensorFloatModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
-class DropoutModule(torch.nn.Module):
+class DropoutEvalIntModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.int64, True),
+    ])
+
+    def forward(self, x):
+        return torch.dropout(x, 0.2, train=False)
+
+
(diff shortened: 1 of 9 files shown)


pashu123:
@Shukla-Gaurav @cathyzhyi I think we have the implementation of aten.bernoulli so we can decompose dropout rather than identity-implementation.

Shukla-Gaurav:
> @Shukla-Gaurav @cathyzhyi I think we have the implementation of aten.bernoulli so we can decompose dropout rather than identity-implementation.

Sure, will look into it. Thanks :)

Shukla-Gaurav:
Now this also covers the training mode. Please review again @cathyzhyi @pashu123

cathyzhyi (approved):
looks good but let's wait for #654 to be merged first. Thanks!

e2e_testing/torchscript/basic.py
```diff
@@ -950,11 +990,11 @@ def __init__(self):
     ])
 
     def forward(self, x):
-        return torch.dropout(x, 0.0, False)
+        return torch.dropout(x, 0.3, train=True) * 0
```
cathyzhyi, line 993: Can the result be tested with mean, var?

cathyzhyi:
Please go ahead to merge when you are ready.

Shukla-Gaurav:
@cathyzhyi It looks like there is some issue with the build system? The latest push did not trigger the build.

cathyzhyi:
I think the build works now. Could you rebase and try again?

PR #606, opened 2022-02-17
Fix MaximizeValueSemantics handling of `torch.overwrite.tensor` op

ramiro050:
This commit fixes an issue with the way that the
`maximize-value-semantics` pass was handling the
`torch.overwrite.tensor` op. Before, the overwrite op would simply get
deleted and uses of the overwritten tensor would simply get replaced
with the overwriter tensor. This becomes a problem when one tensor has
a dynamic shape and the other has a static shape because ops that were
previously using the overwritten tensor will now get a tensor of a
different type.

To fix this, this commit adds a static cast to the overwriter tensor
so that it has the type of the overwritten tensor, and this casted
tensor is then used to replace the uses of the overwritten tensor.

Note: the reason I added the `aten.add.Tensor` op in the mlir test is to make sure that the insertion point for the cast is correct. If the insertion point code is removed from the maximize-value-semantics pass, that test will fail because the static cast will be inserted before the `aten.add.Tensor` op.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
@@ -884,8 +884,10 @@ def Torch_CopyToValueTensorOp : Torch_Op<"copy.to_vtensor", [
   let verifier = "return ::verify(*this);";
 }
 
-def Torch_OverwriteTensorOp : Torch_Op<"overwrite.tensor", [
-    AllowsTypeRefinement
+def Torch_OverwriteTensorContentsOp : Torch_Op<"overwrite.tensor.contents", [
+    TypesMatchWith<"overwritten tensor type is corresponding !torch.tensor of value tensor type",
+                   "value", "overwritten",
+                   "$_self.cast<ValueTensorType>().getWithoutValueSemantics()">
   ]> {
   let summary = "Ovewrite the contents of tensor with values from another.";
   let description = [{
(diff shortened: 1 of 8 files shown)


cathyzhyi (approved):
Thanks for the fix!

silvasean:
I would prefer we fix this (if possible) by requiring the two operands of torch.overwrite.tensor to have the same static information. I know that's a more invasive change, but I think it preserves the orthogonality of the system better.

silvasean:
To give a bit more insight here, the existing handling of `torch.copy.to_tensor`/`torch.copy.to_vtensor` ops already have this kind of behavior, and it has worked out really well. It makes a number of such transformations easier to reason about / have fewer special cases.

ramiro050:
> I would prefer we fix this (if possible) by requiring the two operands of torch.overwrite.tensor to have the same static information. I know that's a more invasive change, but I think it preserves the orthogonality of the system better.

Yeah, I like this a lot more. I've added the type invariant to the `torch.overwrite.tensor` op and modified the `RefineTypes` pass to maintain the invariant when adding type information. PTAL.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -897,8 +899,9 @@ def Torch_OverwriteTensorOp : Torch_Op<"overwrite.tensor", [
     might mutate `overwritten`, so this relationship need not hold for the
     entire program.
 
-    This op has undefined behavior if the two tensors have different
-    shapes or dtypes.
+    This op does not have the AllowsTypeRefinement trait because the types of the
+    two operands are coupled. Only places that know how to simultaneously update
```
silvasean, line 903: You might want to also mention / tighten the wording in this ODS description that this op only updates the tensor data (not metadata), and thus cannot change the (dynamic) shape of the overwritten tensor. There are cases in PyTorch that require modeling in-place updates to the metadata/shape, but we want to clearly isolate those with a new op when we start to model them (maybe we should rename this op torch.overwrite.tensor.contents).

This avoids complexities for a case like:

```
%0 : !torch.vtensor<[?],f32>
%1 : !torch.tensor<[2],f32>
torch.overwrite.tensor %0 overwrites %1
```

We want torch.overwrite.tensor.contents to have undefined behavior if %0  is not the same dynamic shape/dtype as %1 (which in this case means that %0 can be assumed to be [2],f32 at the program point where torch.overwrite.tensor executes). It's very similar to tensor_static_info_cast'ing `[?]` to `[2]` in terms of what we are allowed to assume.  torch.overwrite.tensor.contents is inserted internally by the compiler in a very small number of places, so maintaining these strong invariants isn't too big of a deal, I hope.

silvasean (approved):

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
ramiro050: Thanks for the explanation! I've updated the ODS description and renamed the op to `torch.overwrite.tensor.contents`.

PR #607, opened 2022-02-17
Add aten::max_pool2d_with_indices_backward

gpetters94:
Making a WIP here to get some feedback on how I'm approaching this. Should I be decomposing this as opposed to implementing it directly in linalg? The logic I'm currently working with is:

1. Flatten gradient output to a 1d tensor

2. Fill with zeroes

3. Loop through the indices, inserting 1 into the zeroed-out tensor at each point

4. Expand to the original size and return

That'd need TiledLoop, which I don't think is implemented. When I suggested doing something similar for Nonzero, Sean recommended I decompose the op for performance reasons - should I be doing the same here?

Also, an oddity that I can't place: the indices tensor is apparently has dtype f32 - shouldn't it be index? Am I missing something in ReefineTypes?

### DIFF
--- a/e2e_testing/torchscript/basic.py
+++ b/e2e_testing/torchscript/basic.py
@@ -1151,7 +1151,8 @@ class ConvolutionOverrideableModule(torch.nn.Module):
     @export
     @annotate_args([
         None,
-        ([], torch.float32, True),
+        ([-1, -1], torch.float32, True),
+        ([-1, -1], torch.float32, True),
     ])
     def forward(self, x, y):
         return torch.ops.aten.convolution_overrideable(x, y, None, [1], [0], [1], False, [0], 1)
--- a/e2e_testing/torchscript/basic.py
+++ b/e2e_testing/torchscript/basic.py
@@ -1153,7 +1153,7 @@ class MaxPool2dWithIndicesBackwardsModule(torch.nn.Module):
         None,
         ([-1, -1, -1], torch.float32, True),
         ([-1, -1, -1], torch.float32, True),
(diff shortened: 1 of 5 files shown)


cathyzhyi (commented):
> Loop through the indices, inserting 1 into the zeroed-out tensor at each point

is it some thing like output[indices[i]] = 1? That seems like a scatter op to me. You might need to lower to scatter op https://github.com/llvm/torch-mlir/blob/ed9bd556b3f59feb0fe1cca9776a3c2eb82fe826/external/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorOps.td#L99. However, I think I need more context. Could you point to the source of reference or describe what math calculation you are trying to do in c++ for easier code review? 

> Also, an oddity that I can't place: the indices tensor is apparently has dtype f32 - shouldn't it be index? Am I missing something in ReefineTypes?

Seems you annotated the 3 argument for indices as torch.f32

cathyzhyi:
> How far up the stack does the TMTensor dialect live

Vivek will add a pass https://github.com/llvm/torch-mlir/pull/574 so that torch ops can be converted to TMTtensor ops. You can add your code there. Also, there is https://pytorch.org/docs/stable/generated/torch.Tensor.index_put.html so you might want to consider if `max_pool2d_with_indices_backward` can be lowered to a combination of other torch ops.  That would make the implementation easier.

cathyzhyi:
closed as covered by https://github.com/llvm/torch-mlir/pull/739

PR #608, opened 2022-02-17
[TORCH] Fix the torch-op definition of aten.gelu

pashu123:
New str argument approximation is added.

### DIFF
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -1630,6 +1630,11 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
       gelu.emitError("unimplemented: non-floating point dtype");
       return nullptr;
     }
+    // TODO: Take approximation into account.
+    std::string approximate;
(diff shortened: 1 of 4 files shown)


pashu123:
@cathyzhyi Do you know how to check TorchStringTypes during runtime?

pashu123:
> > @cathyzhyi Do you know how to check TorchStringTypes during runtime?
> 
> You can add a `m_TorchConstantString` similar to `m_TorchConstantInt` and check at compile time.

Thanks.

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1630,6 +1630,12 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
       gelu.emitError("unimplemented: non-floating point dtype");
       return nullptr;
     }
+    // TODO: Take approximation into account.
+    std::string approximate;
+    if (!matchPattern(gelu.approximate(), m_TorchConstantStr(approximate)))
+      return nullptr;
+    if (approximate != "none")
+      return nullptr;
```
Shukla-Gaurav, line 1638: You can combine these two if statements.

cathyzhyi (approved):

PR #611, opened 2022-02-18
Fix handling of view-like ops in `maximize-value-semantics`

ramiro050:
This commit adds handling to the `maximize-value-semantics` pass for
the case where a view-like op depends on a tensor that has been
overwritten by a value tensor. The approach for removing the
dependency is to change the input to the view-like op to be a copy of
the value tensor that is being used to overwrite.

This commit also removes `AtenFill_ScalarOp` and
`AtenBernoulli_FloatOp` from the list of view-like ops, since these
ops now have a corresponding op with value semantics into which they
get converted in the `reduce-op-variants` pass.

### DIFF
--- a/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
+++ b/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
@@ -20,6 +20,18 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 
+static bool isViewLikeOp(Operation *op) {
+  // AtenContiguousOp might return a view, so this is conservatively
+  // correct. We could potentially be more precise and identify the cases
+  // that it does not return a view and treat those as having value
+  // semantics.
+  return isa<AtenBroadcastToOp, AtenContiguousOp, AtenExpandOp,
+             AtenFlattenUsingIntsOp, AtenPermuteOp, AtenReshapeOp,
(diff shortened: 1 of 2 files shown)


lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
```diff
@@ -52,6 +74,35 @@ class AbstractlyInterpretCopyToNonValueTensorOpUsersWithinABlock
...
         rewriter.eraseOp(overwriteTensor);
+      } else if (isViewLikeOp(user)) {
+        // This case currently only handles view-like ops that have one tensor
+        // input and one tensor output.
+        //
+        // The goal here is to transform view-like ops that depend on an
+        // overwritten tensor into ops that don't, so that the `overwrite` op
+        // can be removed. This is achieved as follows:
+        //
+        //   If the view-like op has as input a tensor `T` that has been
```
cathyzhyi, line 85: Personally, I think the two `If` paragraphs are not necessary as they describes the two specific possible cases view-like ops taking non-value tensor as input. The handling for those two cases are the same. The code is very self explanatory that. `In case view-like ops taking a non-value tensor, the operand can be replaced with a new non value tensor with the content of `currentlyHeldValueTensor`. This enable the abstraction interpretation propagate tensor value across view-like ops.`

cathyzhyi (approved):
Great work! This is very tricky! 
Just have some thoughts on the comments.

PR #612, opened 2022-02-18
Generate backward graph via functorch-aot module

pashu123:
Example to demonstrate the extraction of forward as well as
backward graph via Functorch's AOT module is added.

### DIFF
--- a/build_tools/torchscript_e2e_heavydep_tests/bert_functorch.py
+++ b/build_tools/torchscript_e2e_heavydep_tests/bert_functorch.py
@@ -1,0 +1,81 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+
+from transformers import AutoModelForMaskedLM, BertConfig
+import transformers
+import torch
+import torch.utils._pytree as pytree
+
+from torch_mlir_e2e_test.torchscript.framework import TestUtils
+from torch_mlir_e2e_test.torchscript.registry import register_test_case
+from torch_mlir_e2e_test.torchscript.annotations import annotate_args, export
+from functorch_utils import AOTModule, get_input_annotations
+
+torch.manual_seed(0)
+
+def getAnnotatedModule(ts_module, inputs):
+  export(ts_module.forward)
+  annotate_args_decorator = annotate_args(
+      get_input_annotations(inputs, dynamic=False))
+  annotate_args_decorator(ts_module.forward)
+  return ts_module
+
+pytree._register_pytree_node(
+    transformers.modeling_outputs.MaskedLMOutput,
+    lambda x: ([x.loss, x.logits], None),
+    lambda values, _: transformers.modeling_outputs.MaskedLMOutput(
+        loss=values[1], logits=values[1]
+    ),
+)
+
+torch.manual_seed(42)
+
+config = BertConfig(hidden_dropout_prob=0.0, attention_probs_dropout_prob=0.0)
+model_type = AutoModelForMaskedLM
+input_size = (1, 2)
+device = "cpu"
+dtype = torch.float
+
+model = model_type.from_config(config).to(device, dtype=dtype)
+input_ids = torch.randint(0, config.vocab_size, input_size).to(device)
+decoder_ids = torch.randint(0, config.vocab_size, input_size).to(device)
+train_inputs = {"input_ids": input_ids, "labels": decoder_ids}
+
+def inference_fn(model, input, labels):
+  return model(**input).loss.sum().backward()
+
+
+aot_module = AOTModule(model,
+                       train_inputs,
+                       labels=None,
+                       training_fn=inference_fn)
+aot_module.generate_graphs()
+
+# ==============================================================================
+# Forward test.
+forw_inputs = []
+for inp in aot_module.forward_inputs:
+  forw_inputs.append(inp.detach())
+
+ts_module_forw = getAnnotatedModule(aot_module.forward_graph,
+                                    aot_module.forward_inputs)
+
+
+@register_test_case(module_factory=lambda: ts_module_forw)
+def BERT_forward_basic(module, tu: TestUtils):
+  module.forward(*forw_inputs)
+
+# ==============================================================================
+# Backward test.
+back_inputs = []
+for inp in aot_module.backward_inputs:
+  back_inputs.append(inp.detach())
+
+ts_module_back = getAnnotatedModule(aot_module.backward_graph,
+                                    aot_module.backward_inputs)
+
+@register_test_case(module_factory=lambda: ts_module_back)
+def BERT_backward_basic(module, tu: TestUtils):
+  module.forward(*back_inputs)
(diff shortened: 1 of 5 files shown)


silvasean (changes requested):
I think you need to update https://github.com/llvm/torch-mlir/blob/main/build_tools/torchscript_e2e_heavydep_tests/generate_serialized_tests.sh to install functorch.

examples/aot_backward.py
```diff
@@ -0,0 +1,131 @@
+# Functorch is required.
+# Please refer to: https://github.com/pytorch/functorch/blob/main/README.md
```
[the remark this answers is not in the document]
pashu123: Done. Thanks.

pashu123:
@silvasean Addressed all your comments. PTAL.

build_tools/torchscript_e2e_heavydep_tests/aot_backward.py
```diff
@@ -0,0 +1,129 @@
...
+    min_cut_rematerialization_partition,
+)
+from torch import optim, fx
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+#### torch-mlir imports
+# To run the example, make sure the following are in your PYTHONPATH:
+# 1. /path/to/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir
```
silvasean, line 15: you can remove this comment

build_tools/torchscript_e2e_heavydep_tests/aot_backward.py
```diff
@@ -0,0 +1,129 @@
+from functorch.compile import (
+    memory_efficient_fusion,
```
silvasean, line 2: do we need all these imports?

build_tools/torchscript_e2e_heavydep_tests/aot_backward.py
```diff
@@ -0,0 +1,129 @@
...
+    f = torch.jit.freeze(f.eval())
+    torch.jit.save(f, "back.pt")
+    f = torch.jit.load("back.pt")
+    print("The backward torch-mlir graph is:")
+    generate_mlir_module(
+        f, generate_annotation(inps), "backward.mlir"
+    )
+    return f
+
+
```
silvasean, line 121: can you please check this in as an actual e2e test?

See the structure of https://github.com/llvm/torch-mlir/blob/main/build_tools/torchscript_e2e_heavydep_tests/minilm_sequence_classification.py

Notice:
```
@register_test_case(module_factory=lambda: getTracedRecursiveScriptModule())
def MiniLMSequenceClassification_basic(module, tu: TestUtils):
    module.forward(test_input)
```

Can you also please verify that you are able to run the resulting test (even if it fails) following the instructions at https://github.com/llvm/torch-mlir/tree/main/build_tools/torchscript_e2e_heavydep_tests/README.md

silvasean (changes requested):

build_tools/torchscript_e2e_heavydep_tests/aot_backward.py
pashu123: No, thanks for pointing this out.

build_tools/torchscript_e2e_heavydep_tests/aot_backward.py
pashu123: This test just demonstrates the generation of the backward graph via AOT autograd, the intention to add it as an e2e test is not there.

silvasean: I consider it a requirement to actually add it as an e2e test. Then we can all work together towards running it which is a major project goal.

silvasean:
Any reason we are closing this? It is definitely a project priority to run this, so I think it would be good to keep working on this.

pashu123:
> Any reason we are closing this? It is definitely a project priority to run this, so I think it would be good to keep working on this.

Sorry for closing. I will keep working on this.

pashu123:
@silvasean Any suggestions on where can I place the functorch_utils.py?

pashu123:
@silvasean This may return different return types not supported by refbackend. If you could re-iterate the pointers you mentioned in the meeting to support different return types it would be a help! I will start working on this.

pashu123:
> @silvasean @cathyzhyi The MiniLM test case is currently failing
> 
> ```
> FAIL - "MiniLMSequenceClassification_basic"                                                                                                                                                                        
>                                                                                                                                                                                                                    
> Unexpected outcome summary:                                                                              
>                                                                                                                                                                                                                    
>     FAIL - "MiniLMSequenceClassification_basic"                                                          
>         Compilation error: Traceback (most recent call last):                                                                                                                                                      
>           File "/home/prashant/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 248, in run_tests                                                  
>             compiled = config.compile(test.program_factory())                                            
>           File "/home/prashant/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/linalg_on_tensors_backend.py", line 37, in compile
>             module = convert_torchscript_module_to_torch_backend_contract_mlir(                          
>           File "/home/prashant/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/utils.py", line 75, in convert_torchscript_module_to_torch_backend_contract_mli
> r                                                                                                                                                                                                                  
>             raise Exception(f"""                                                                                                                                                                                   
>         Exception:                                                                                                                                                                                                 
>         PyTorch TorchScript module -> torch-mlir Object Graph IR import failed with:                                                                                                                               
>         Exception:                                                                                       
>         required keyword attribute 'unsafe' is undefined                                                                                                                                                           
>         Diagnostics:  
> ```
> 
> Any suggestions?

I did a fresh install and these problems were gone! Thanks to @cathyzhyi

silvasean:
> @silvasean This may return different return types not supported by refbackend. If you could re-iterate the pointers you mentioned in the meeting to support different return types it would be a help! I will start working on this.

You can take the torch fx graph and munge it to remove the None return types.

pashu123:
> > @silvasean This may return different return types not supported by refbackend. If you could re-iterate the pointers you mentioned in the meeting to support different return types it would be a help! I will start working on this.
> 
> You can take the torch fx graph and munge it to remove the None return types.

@silvasean Oh! Not for the None type, I have already done that in this PR. I was asking about returning multiple different tensors through ref-backend currently it supports only fixed return types.

silvasean:
> I was asking about returning multiple different tensors through ref-backend currently it supports only fixed return types.

Oh, the idea is that the refback-munge-calling-conventions pass would embed a string listing all the "consume_return" signatures it needs, and then in Python read that string, create all the consume_return functions, and then register them here: https://github.com/llvm/torch-mlir/blob/24f9de7120bcd95de3d341f325a46d19a6a2dc4c/python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py#L105

build_tools/torchscript_e2e_heavydep_tests/bert_functorch.py
```diff
@@ -0,0 +1,82 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+
```
silvasean, line 5: nit: one newline only

PR #613, opened 2022-02-18
[LINALG] Add E2E support for `aten.mean.dim` op

Shukla-Gaurav:
- This commit adds support for `aten.mean.dim` op.
- It also adds a new test script `stats.py` for statistics related ops.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -22,6 +22,18 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 
+// Helper function to check whether the `dtype` is None or Float type.
+static bool isNoneOrFloatDtype(MLIRContext *context, Value dtype) {
+  if (dtype.getType().isa<Torch::NoneType>())
+    return true;
+  int64_t dtypeInt;
+  if (!matchPattern(dtype, m_TorchConstantInt(&dtypeInt)))
+    return false;
+  Type resDtype =
+      getTypeForScalarType(context, (torch_upstream::ScalarType)dtypeInt);
+  return resDtype.isa<mlir::FloatType>();
+}
+
 // Helper function to compute the return type of the reduction function.
 // `dim` specifies the dimension to reduce and `keepDim` preserves the rank of
 // the input tensor.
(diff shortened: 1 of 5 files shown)


Shukla-Gaurav:
@cathyzhyi @ramiro050 ready for re-review. thanks :)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -34,11 +34,62 @@ static int getTensorRank(Value tensor) {
   return tensorRank;
 }
 
+// Helper function to compute the return type of the reduction function along
+// multiple dimensions. `dimList` specifies a list of dimensions to reduce and
+// `keepDim` preserves the rank of the input tensor.
+static LogicalResult
```
ramiro050, line 40: nit: You can use the return type `FailureOr<BaseTensorType>` to return a `rewriter.notifyMatchFailure` or a `BaseTensorType`, so that you don't have to pass the return value as an argument to the function. You can then check if the returned value failed by doing `if (failed(LogicalResult(returnedValue)))`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -34,11 +34,62 @@ static int getTensorRank(Value tensor) {
...
+          .cast<BaseTensorType>();
+  return success();
+}
+
 // Helper function to compute the return type of the reduction function.
 // `dim` specifies the dimension to reduce and `keepDim` preserves the rank of
 // the input tensor.
-static Type computeReductionType(PatternRewriter &rewriter, Operation *op,
-                                 Value input, Value dim, bool keepDim) {
+static Type computeReductionAlongDimType(PatternRewriter &rewriter,
```
ramiro050, line 90: This is just a special case of the `computeReductionAlongMultipleDimsType`, so you should rewrite it using that function.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -741,6 +797,131 @@ class DecomposeAtenMeanOp : public OpRewritePattern<AtenMeanOp> {
...
+
+    // Compute sum along dimensions specified in `dimList`.
+    Value none = rewriter.create<Torch::ConstantNoneOp>(loc);
+    Value sumAlongDims = rewriter.create<AtenSumDimIntListOp>(
+        loc, outputType, input, dimList, keepDim, /*dtype=*/none);
+
+    auto dimListConstruct = dimList.getDefiningOp<PrimListConstructOp>();
+    if (!dimListConstruct)
+      return rewriter.notifyMatchFailure(
+          op, "expect dimList to be constructed from list construct");
```
ramiro050, line 828: You should check this before creating any ops to avoid doing extra work. Also, the `if` needs curly braces

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -741,6 +797,131 @@ class DecomposeAtenMeanOp : public OpRewritePattern<AtenMeanOp> {
...
+    BaseTensorType inputType = input.getType().cast<BaseTensorType>();
+    if (!inputType.hasDtype() || !inputType.getDtype().isa<mlir::FloatType>()) {
+      return rewriter.notifyMatchFailure(
+          op, "only floating-point type is supported");
+    }
+
+    // Compute sum along dimensions specified in `dimList`.
+    Value none = rewriter.create<Torch::ConstantNoneOp>(loc);
+    Value sumAlongDims = rewriter.create<AtenSumDimIntListOp>(
+        loc, outputType, input, dimList, keepDim, /*dtype=*/none);
```
ramiro050, line 823: `AtenMeanDimOp` has a `dtype` argument that should be passed along into `AtenSumDimIntListOp`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -741,6 +797,131 @@ class DecomposeAtenMeanOp : public OpRewritePattern<AtenMeanOp> {
...
+  LogicalResult matchAndRewrite(AtenMeanDimOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value input = op.self();
+    Value dimList = op.dim();
+    Value keepDim = op.keepdim();
+    Type outputType = op.getType();
+
+    BaseTensorType inputType = input.getType().cast<BaseTensorType>();
+    if (!inputType.hasDtype() || !inputType.getDtype().isa<mlir::FloatType>()) {
```
ramiro050, line 815: I think you should also check that `op.dtype()` is float. I tried passing a type different from float, and PyTorch gives casting errors.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -741,6 +797,131 @@ class DecomposeAtenMeanOp : public OpRewritePattern<AtenMeanOp> {
...
+      return rewriter.notifyMatchFailure(op, "expect unbiased to be constant");
+    }
+
+    // Compute square(input - mean(input, dimList, keepdim=True)). Here,
+    // `keepdim` is set to True to ensure that the substraction is broadcast
+    // compatible.
+    Value none = rewriter.create<Torch::ConstantNoneOp>(loc);
+    Value constantTrue = rewriter.create<Torch::ConstantBoolOp>(loc, true);
+    BaseTensorType reduceType;
+    if (failed(computeReductionAlongMultipleDimsType(
```
ramiro050, line 879: `if` needs curly braces

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -839,6 +851,54 @@ class DecomposeAtenMeanOp : public OpRewritePattern<AtenMeanOp> {
...
+    Value keepDim = op.keepdim();
+    Value dtype = op.dtype();
+    Type outputType = op.getType();
+    MLIRContext *context = op.getContext();
+
+    BaseTensorType inputType = input.getType().cast<BaseTensorType>();
+    if (!inputType.hasDtype() || !inputType.getDtype().isa<mlir::FloatType>() ||
+        !isNoneOrFloatDtype(context, dtype)) {
+      return rewriter.notifyMatchFailure(
+          op, "only floating-point type is supported");
```
ashay, line 874: Just out of curiosity, what's our rationale behind choosing between `notifyMatchFailure()` versus `emitError()`?  Should we use `emitError()` here, so that the end user does not need to re-run the compilation with the `--debug` flag?

Shukla-Gaurav:
@cathyzhyi @ramiro050 Gentle reminder.

python/torch_mlir_e2e_test/test_suite/stats.py
```diff
@@ -0,0 +1,253 @@
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([3, 4], torch.float32, True),
+    ])
+    def forward(self, x):
+        return torch.mean(x)
```
cathyzhyi, line 24: prefer to use `torch.ops.aten.` operators directly.

cathyzhyi (approved):
Looks mostly good except we would prefer to use the aten operator directly in e2e tests.

ramiro050 (approved):
LGTM!

python/torch_mlir_e2e_test/test_suite/stats.py
Shukla-Gaurav: Thanks, done :)

PR #614, opened 2022-02-21
Extend tm_tensor.scatter op semantic to carry unique_indices attribute

vivekkhandelwal1:
There are cases where the op may update the same indices multiple
times. In this context, we can not parallelize updates. Instead,
we have to execute them sequentially. Adding a boolean attribute
to control the behavior.

Also adding test cases for invalid IR.


pashu123:
Please add the original IREE PR link to the commit body. This will make it easier to go back and forth. I think https://github.com/google/iree/pull/8365/files is the PR.

cathyzhyi (approved):
LGTM. Thanks!

Issue #618, opened 2022-02-23
NativeBatchNorm op is being decomposed before TorchToTosa lowering


anupgangwar:
From the e2e test framework, I see that the generated MLIR no longer contains the batch_norm op for BatchNorm*_basic tests. The op appears to have been decomposed further (as per it definition). This appears to be a side-effect of PR #594. Is there a way to not have the decomposition in the TOSA backend path?

Tagging @cathyzhyi @silvasean @sjarus.

silvasean:
Why do you want a batch norm op? It seems simpler to lower the individual decomposed ops.

We could design a way to avoid decomposing it, but I would like to understand the motivation.

anupgangwar:
For the near term, there are a few decomposed ops which are not supported in TorchToTosa yet. Then there are some opens around the .size op which are under discussion.

If this decomposition can be avoided in the TOSA backend path for now, it will again get the ResNet18 passing. Once the full set of ops are supported we can look at whether to enable the decomposition again.

silvasean:
For this specific case, I would prefer to fix this forward, as suppressing the decomposition temporarily without a failing upstream test to motivate it doesn't seem desirable. This type of thing will eventually be caught by CI and fixed on the spot once ResNet passes on TOSA upstream :)

sjarus:
Stepping in to assist @anupgangwar here after an internal conversation - there's some confusion here related to to previous advise @cathyzhyi had posted in #594  - the suggestion there was to lower aten.batch_norm to aten.native_batch_norm . This is doable according to Anup. 

However, the decomposition isn't actually to native_batch_norm as an op, but to the mechanical primitives of batch_norm. Maybe that's the intent of native_batch_norm, but then both aten.batch_norm and aten.native_batch_norm are listed in GeneratedAtenOps.td , so we expected to see a single op aten.native_batch_norm in the IR.

silvasean:
Is there some specific op in the decomposition that is hard to support? It should just be basic stuff like sqrt, add, mul, divide, sum, etc. which seem like we want to support anyway. It feels like the right long-term direction for the TOSA backend to support those ops, so since there is not immediate test breakage upsream, it's hard to justify an upstream change to suppress that behavior. If this is a problem for your locally, you can temporarily comment out that pattern in DecomposeComplexOps.

[...]

PR #619, opened 2022-02-24
Revert "[LINALG] Decompose `aten.batch_norm` into `aten.native_batch_norm`"

Shukla-Gaurav:
Reverts llvm/torch-mlir#594
The decomposition of `batch_norm` breaks its support at TOSA backend
due to the insertion of non-Tensor elements. 
For more details refer: https://github.com/llvm/torch-mlir/issues/618


cathyzhyi:
@Shukla-Gaurav I think it's ok to leave the code in place and just disable the patterns in `DecomposeComplexOps`. This would probably make things easier when you want to reenable the code.

silvasean:
> @Shukla-Gaurav I think it's ok to leave the code in place and just disable the patterns in `DecomposeComplexOps`. This would probably make things easier for you when you want to reenable the code.

As a matter of coding style, in LLVM we don't leave in dead code like this. The code is always saved in git and can be recovered from there.

cathyzhyi (approved):

PR #622, opened 2022-02-25
Don't override MLIR_TABLEGEN_EXE

stephenneuendorffer:
This should be set elsewhere depending on the build configuration.
In particular, we need to be careful when cross-compiling to pick
up the host mlir-tblgen.


silvasean:
LGTM in principle

PR #623, opened 2022-02-25
Ensure that torch-mlir-dialects is built when we're out of tree

stephenneuendorffer:
Its unclear to me what the right layering is here: Are you expecting
torch-mlir-dialects to always get built with LLVM?  This is pretty
breaking for us, if so.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,6 +80,7 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
 
   set(TORCH-MLIR_BUILT_STANDALONE 1)
   set(BACKEND_PACKAGE_STRING "LLVM ${LLVM_PACKAGE_VERSION}")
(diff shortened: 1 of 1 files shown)


stephenneuendorffer:
It seems like this builds fine as part of torch-mlir.. What is being gained here by forcing it to be built with LLVM?

cathyzhyi:
Sry, it should be optional to be built with LLVM as well.

stephenneuendorffer:
@cathyzhyi OK,  Do you think this patch should be merged, or do you want to have a separate configuration for it, so that it might also work both for in-tree an out-of-tree builds?

cathyzhyi (approved):

cathyzhyi:
> @cathyzhyi OK, Do you think this patch should be merged, or do you want to have a separate configuration for it, so that it might also work both for in-tree an out-of-tree builds?

actually, it will not be optional pretty soon with some some new PRs coming in. I think we can go ahead and merge this change.

PR #624, opened 2022-02-25
Add `reduction` support to `torch.nll_loss_forward`

ramiro050:
This commit does a couple of things. First, it fixes a bug in the
`linalg.generic` body of the `nll_loss_forward` lowering where the
`ignoreIndex` was being compared with the loop index rather than the
current element of the `target` tensor. This was not being caught by
the tests because they were not testing the case where `ingnoreIndex`
actually corresponds to a value in `target`. This has been fixed.

Second, this commit adds support for the `reduction` argument in
`torch.nll_loss_forward` as well as support for 1-D inputs. In order
to simplify the lowering code, I've refactored the code that creates
the `linalg.generic` ops for elementwise and reduction ops into static
functions, to avoid having boilerplate code for indexing maps, etc
that can be very error prone.

Note: The function `convertScalarToDtype` was moved to before all the
conversion patterns, but nothing in it was modified. The `createElementwiseLinalgGeneric` function and `createReductionLinalgGeneric` function also had very small modifications when refactored. It was mostly renaming things so that the code would work with the signature of the functions.

### DIFF
--- a/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
+++ b/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
@@ -181,6 +181,14 @@ static SmallVector<Value> getTensorSizes(OpBuilder &b, Location loc,
   return getTensorSizesUntilDim(b, loc, tensor, type.getRank() - 1);
 }
 
+static Value getTensorSize(OpBuilder &b, Location loc, Value tensor) {
+  SmallVector<Value> sizes(getTensorSizes(b, loc, tensor));
+  Value productResult = b.create<arith::ConstantOp>(loc, b.getIndexAttr(1));
+  for (Value size : sizes)
+    productResult = b.create<arith::MulIOp>(loc, productResult, size);
+  return castIndexToInt(b, loc, productResult);
+}
+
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -1273,52 +1502,67 @@ class ConvertAtenNllLossForwardOp
...
+          Value selectFinal =
+              rewriter.create<arith::SelectOp>(loc, cmpEq, zeroVal, negate);
+          b.create<linalg::YieldOp>(loc, selectFinal);
+        });
+
+    if (reduction == Reduction::Sum || reduction == Reduction::Mean) {
+      Value numOfElems = getTensorSize(rewriter, loc, finalRes);
+      numOfElems = convertScalarToDtype(rewriter, loc, numOfElems, elementType);
+      llvm::iota_range<int64_t> dimsToReduce(0, targetRank,
+                                             /*inclusive*/ false);
```
cathyzhyi, line 1551: nit: `/*inclusive=*/`

cathyzhyi (approved):
Great work!

PR #626, opened 2022-02-28
[MLIR][TORCH] Add E2E support for aten.new_zeros, aten.new_ones op

vivekkhandelwal1:
This commit adds lowering of `aten.new_zeros` and `aten.new_ones` op


e2e_testing/torchscript/constant_alloc.py
```diff
@@ -497,3 +497,209 @@ def forward(self, tensor):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, a):
+        return a.new_zeros(3, 4)
```
cathyzhyi, line 513: nit: prefer to use `torch.ops.aten.new_zeros`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1218,6 +1218,26 @@ class DecomposeAten_UnsafeViewOp : public OpRewritePattern<Aten_UnsafeViewOp> {
 };
 } // namespace
 
+namespace {
+// Decompose constant tensor like ops.
+template <typename OpTy>
```
cathyzhyi, line 1223: this can be replaced with `template <typename OpTy, typename NewOpTy>`. The if branches below can be removed and become
```
rewriter.replaceOpWithNewOp<NewOpTy>(op, op.getType(), op.size(),
                                               op.dtype(), op.layout(),
                                               op.device(), op.pin_memory());
```

cathyzhyi (changes requested):

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1535,6 +1542,18 @@ ChangeResult TypeAnalyzer::visitConstantTensorAllocLikeOp(
   return getLatticeElement(op.getResult()).join(knowledge);
 }
 
+template <typename OpTy>
+ChangeResult TypeAnalyzer::visitConstantTensorNewLikeOp(
+    OpTy op, ArrayRef<LatticeElement<ValueKnowledge> *> operands) {
+  auto input = operands[0]->getValue();
+  auto knowledge =
+      ValueKnowledge::getNotNonePessimisticValueState(op->getContext());
+  knowledge.hasSizes = true;
```
cathyzhyi, line 1551: this should be removed. `knowledge.hasSizes` is set in `  fillInSizesGivenSizesList`

cathyzhyi (approved):

PR #627, opened 2022-02-28
Add nonzero

gpetters94:
Definitely in need of cleanup still, and some of the ops that it decomposes into aren't supported, but here's a WIP patch for aten::nonzero. I'm going to go through output and add all the unsupported ops to the list shortly.


silvasean:
Hi @gpetters94 I would recommend working bottom up from the ops needed by the decomposition. That way you can add e2e tests as you go. LLVM coding conventions very strongly discourage adding code that is not functionally reachable/useful from an e2e flow.

PR #629, opened 2022-03-01
Eager mode

makslevental:


makslevental:
close in favor of https://github.com/llvm/torch-mlir/pull/635

PR #630, opened 2022-03-01
Add bufferization pass for TMTensor ops

cathyzhyi:
The pass is mostly borrowed from the BufferizeAnyLinalgOp pass in mlir
upstream with some minor changes. At a high level, it's a naive partial
bufferization pass which allocate new buffers for all the output
tensors. The initial value of an output buffer is copied from the
original buffer if there are uses of the original value.

One difference from linalg bufferization pass is the way to tell if
the loop body uses the init value of output operand. For TMTensor ops,
it differs from op to op because the payload region doesn't represent
the entire loop body.

### DIFF
--- a/python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
+++ b/python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
@@ -166,6 +166,7 @@ class RefBackendInvoker:
 LOWERING_PIPELINE = ",".join([
     # Bufferize.
     "builtin.func(scf-bufferize)",
+    "builtin.func(tm-tensor-bufferize)",
     "builtin.func(linalg-bufferize)",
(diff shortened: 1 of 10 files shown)


python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
```diff
@@ -166,6 +166,7 @@ def invoke(*args):
 LOWERING_PIPELINE = ",".join([
     # Bufferize.
     "builtin.func(scf-bufferize)",
+    "builtin.func(torch-mlir-tm-tensor-bufferize)",
```
silvasean, line 169: nit: technically "tm" in tm-tensor is already "torch-mlir", so it's a bit redundant to have torch-mlir prefix on the pass name.

silvasean (approved):

PR #632, opened 2022-03-01
[LINALG] Fix `aten.bernoulli` op lowering

Shukla-Gaurav:
- This commit adds E2E support for `aten.rand_like` and
  `aten.bernoulli_.Tensor` ops.
- Previously, the `aten.bernoulli(x)` was implemented as:
  `aten.bernoulli(x) = rand_like(x) < 0.5`, assuming 0.5 as default
  probability, whereas according to the pytorch documentation:
  https://pytorch.org/docs/stable/generated/torch.bernoulli.html#torch.bernoulli
  the input x in `aten.bernoulli(x)` is itself a tensor containing
  probabilities to be used for drawing the binary random number.
- So this commit fixes the `aten.bernoulli(x)` implementation as:
  `aten.bernoulli(x) = rand_like(x) < x`.
- It also fixes the case where the input to `aten.bernoulli_.float` is
  an integer tensor. In that case the input must be casted to float type
  before passing it as operand to `aten.rand_like` op.
  `aten.bernoulli_.float(x, p) = rand_like(float(x)) < p`.

### DIFF
--- a/e2e_testing/torchscript/rng.py
+++ b/e2e_testing/torchscript/rng.py
@@ -4,7 +4,6 @@ from torch_mlir_e2e_test.torchscript.framework import TestUtils
 from torch_mlir_e2e_test.torchscript.registry import register_test_case
 from torch_mlir_e2e_test.torchscript.annotations import annotate_args, export
 
-
 # ==============================================================================
 
 class UniformModule(torch.nn.Module):
@@ -90,17 +89,73 @@ class BernoulliModule(torch.nn.Module):
     @export
     @annotate_args([
         None,
-        ([-1, -1, -1], torch.float32, True),
+        ([-1, -1, -1], torch.float64, True),
+        ([-1, -1, -1], torch.float64, True),
+        ([-1, -1, -1], torch.float64, True),
     ])
-    def forward(self, a):
-        x = torch.bernoulli(a)
-        mean = torch.mean(x)
-        std = torch.std(x)
+    def forward(self, x, y, z):
+        a = torch.bernoulli(x)
+        b = torch.bernoulli(y)
+        c = torch.bernoulli(z)
+        mean = torch.cat([
+            torch.flatten(torch.mean(a)),
+            torch.flatten(torch.mean(b)),
+            torch.flatten(torch.mean(c))
+        ])
+        std = torch.cat([
+            torch.flatten(torch.std(a)),
+            torch.flatten(torch.std(b)),
+            torch.flatten(torch.std(c))
+        ])
         return  mean, std
+
 
 @register_test_case(module_factory=lambda: BernoulliModule())
 def BernoulliModule_basic(module, tu: TestUtils):
-    module.forward(tu.rand(256, 512, 64))
+    module.forward(
+        tu.rand(256, 512, 8).double(),
+        tu.rand(512, 1024, 4).double(),
+        tu.rand(512, 256, 4).double())
+
+# ==============================================================================
+
+class BernoulliZerosModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float64, True),
+    ])
+    def forward(self, x):
+        return torch.bernoulli(x)
+
+
+@register_test_case(module_factory=lambda: BernoulliZerosModule())
+def BernoulliZerosModule_basic(module, tu: TestUtils):
+    module.forward(torch.zeros(4, 8).double())
+
+# ==============================================================================
+
+class BernoulliOnesModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float64, True),
+    ])
+    def forward(self, x):
+        return torch.bernoulli(x)
+
+
+@register_test_case(module_factory=lambda: BernoulliOnesModule())
+def BernoulliOnesModule_basic(module, tu: TestUtils):
+    module.forward(torch.ones(4, 8).double())
+
+# ==============================================================================
 
 class BernoulliFloatModule(torch.nn.Module):
     def __init__(self):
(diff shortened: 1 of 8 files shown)


e2e_testing/torchscript/rng.py
```diff
@@ -90,17 +89,34 @@ def __init__(self):
...
+        ])
         return  mean, std
 
 @register_test_case(module_factory=lambda: BernoulliModule())
 def BernoulliModule_basic(module, tu: TestUtils):
-    module.forward(tu.rand(256, 512, 64))
+    module.forward(
+        tu.rand(256, 512, 8).double(),
+        tu.rand(512, 1024, 4).double(),
+        tu.rand(512, 256, 4).double())
```
pashu123, line 117: Please also test extreme cases like `torch.zeros` and `torch.ones` or `torch.ones() * 0.7` to test different distributions.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -953,6 +953,24 @@ def Torch_PseudoAtenBernoulliFloatOp: Torch_Op<"pseudo.aten.bernoulli.float", [
   let assemblyFormat = "$self `,` $p `,` $generator attr-dict `:` type($self) `,` type($p) `,` type($generator) `->` type($result)";
 }
 
+// The corresponding without underscore variant for `torch.aten.bernoulli_.Tensor`
```
pashu123, line 956: @cathyzhyi Shouldn't we have another `.td` file for `pseudoAtenOps`. This will also act as a clear separation from core torch ops and pseudo ops.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -872,87 +872,179 @@ class DecomposeAtenHardsigmoidOp : public OpRewritePattern<AtenHardsigmoidOp> {
...
+  // `constDtypeF64` is an integer constant corresponding to the Float64 dtype.
+  int dtypeF64 = (int)getScalarTypeForType(rewriter.getF64Type());
+  Value constDtypeF64 =
+      rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(dtypeF64));
+  auto floatResType = inputType.getWithSizesAndDtype(inputType.getSizes(),
+                                                     rewriter.getF64Type());
+  Value floatTensor = rewriter.create<AtenToDtypeOp>(
+      loc, floatResType, input, constDtypeF64, /*non_blocking=*/constFalse,
+      /*copy=*/constFalse, /*memory_format=*/none);
+  Value random = rewriter.create<AtenRandLikeOp>(
```
pashu123, line 953: Nit: `randomVal` instead of random.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -872,87 +872,179 @@ class DecomposeAtenHardsigmoidOp : public OpRewritePattern<AtenHardsigmoidOp> {
...
+  Value lessThanP =
+      rewriter.create<AtenLtTensorOp>(loc, boolResType, random, prob);
+
+  // `constDtypeInput` is an integer constant corresponding to `input` dtype.
+  int dtypeInput = (int)getScalarTypeForType(inputType.getDtype());
+  Value constDtypeInput = rewriter.create<ConstantIntOp>(
+      loc, rewriter.getI64IntegerAttr(dtypeInput));
+  // As the output is expected to be of the `input` type, convert the boolean
+  // tensor `lessThanP` to a `input` type tensor.
+  Value output = rewriter.create<AtenToDtypeOp>(
```
pashu123, line 969: I think we can create a new function for type conversion, `static Value convertDtype(Value, from ,to)`.

pashu123 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -872,87 +872,179 @@ class DecomposeAtenHardsigmoidOp : public OpRewritePattern<AtenHardsigmoidOp> {
...
+  // The `prob` is expected to be a float type tensor.
+  if (!probType.hasDtype() || !probType.getDtype().isa<mlir::FloatType>()) {
+    (void)rewriter.notifyMatchFailure(
+        op, "probabilities must be a float type tensor");
+    return nullptr;
+  }
+
+  // Since the `aten.rand_like` op expects float-type operand, create a
+  // float-type tensor with the same shape as that of the `input`.
+  Value none = rewriter.create<ConstantNoneOp>(loc);
```
cathyzhyi, line 942: There is a `convertTensorToDtype` helper that can be used for this now.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -750,7 +753,8 @@ static void fillInDTypeGivenDTypeIntAndInputDType(ValueKnowledge &knowledge,
                                                   Value dtype,
                                                   Type inputDType) {
   int64_t dtypeInt;
-  if (dtype.getType().isa<Torch::NoneType>())
+  if (dtype.getType().isa<Torch::NoneType>() ||
+      dtype.getType().isa<OptionalType>())
```
cathyzhyi, line 757: actually for `OptionalType` we can't deduce the `dtype` at compile time because the value can be either None or not None at runtime. It's correct to return without filing in `knowledge.dtype` for `OptionalType`

cathyzhyi (changes requested):
Thanks for the fixes!

e2e_testing/torchscript/rng.py
Shukla-Gaurav: Done. Thanks :)

lib/Dialect/Torch/Transforms/RefineTypes.cpp
Shukla-Gaurav: Thanks!

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
Shukla-Gaurav: I can handle that in a subsequent PR. Thanks!

Shukla-Gaurav:
@cathyzhyi @pashu123 I have addressed all the comments. Can you please have a look?

cathyzhyi (approved):

PR #633, opened 2022-03-02
[LINALG] Add handling of inferred dimension in size list of `view` op

ramiro050:
The view op allows for the new shape argument to have a -1 value for
one of the dimensions, and the op is expected to deduce the size of
that dimension by looking at the sizes of the other dimensions and
comparing it to the total number of elements in the original
tensor. This commit adds this functionality.

Note: this only works when there is enough static information to deduce the dimension


cathyzhyi (approved):
LGTM!

PR #635, opened 2022-03-02
[TORCH][MLIR] Eager mode

makslevental:
Eager mode for pytorch with torch-mlir backend. Uses `python` dispatc…h key and `__torch_dispatch__` mechanism.

TODOs before landing:

1. backwards pass
2. move to a better place (including methods reused from examples)
2. tests (especially numerical correctness)
3. better doc/comment strings/types/exceptions

short term future work:

1. remove `TorchMLIRTensor`
2. compile cache
3. tracing jit


silvasean:
Can you add a test config for the e2e test framework?

https://github.com/llvm/torch-mlir/tree/main/python/torch_mlir_e2e_test/torchscript/configs

python/torch_mlir/eager/torch_dispatch.py
```diff
@@ -0,0 +1,353 @@
...
+    2. for each arg to the op:
+        a. if arg is a tensor: add input to the graph corresponding to arg
+           if arg is a constant (bool, int, list): inline a constant (at the top of the graph)
+        b. wire value from previous step to the op (ie SSA value corresponding to graph input is "passed" to op)
+        c. if arg is a tensor: fill in shape
+
+    The (poorly handled/poorly understood) gotcha here is args with type Optional[Tensors] with values None.
+    These pytorch types correspond to !torch.optional<!torch.vtensor>, which torch-mlir complains about having a torch.type_bound
+    (something something torch.type_bound can only be a func arg_attr for torch.vtensor). The hacky solution here works but
+    there's probably some straightforward way to handle this that I didn't stumble upon.
```
makslevental, line 169: This is a place where I expect someone to tell me that there's a more appropriate way to handle it.

python/torch_mlir/eager/torch_dispatch.py
```diff
@@ -0,0 +1,353 @@
...
+    graph = torch._C.Graph()
+    node = graph.insertNode(graph.create(schema.name, len(schema.returns)))
+    for i, arg in enumerate(schema.arguments):
+        if arg.name in kwargs:
+            val = kwargs[arg.name]
+        else:
+            val = args[i]
+
+        if arg.type.isSubtypeOf(torch.TensorType.get()) or (
+            isinstance(arg.type, torch.OptionalType) and val is not None
```
makslevental, line 187: I would prefer to write something like `arg.type.isSubtypeOf(torch.OptionalType(torch.TensorType.get()))` and below to check for `Optional[Tensor]` but the problem is that 

```
arg: torch.Tensor = ...
arg.type.isSubtypeOf(torch.OptionalType(torch.TensorType.get())) == True
```

python/torch_mlir/eager/torch_dispatch.py
```diff
@@ -0,0 +1,353 @@
...
+def build_script_function(
+    schema: torch._C.FunctionSchema,
+    args: List[torch._C.Argument],
+    kwargs: Dict[str, Any],
+) -> torch.jit.ScriptFunction:
+    """
+    Build a torch.jit.ScriptFunction that corresponds to the schema.
+    This is necessary so that we can pass a "hydrated" module to ModuleBuilder::importModule.
+    Hydrated meaning with inlined constants with correct values; a gotcha here is that if you don't inline the constants
+    then importModule assume default values (i think?) thereby (potentially) failing to support the op (not sure how this makes sense but native_batch_norm for example doesn't work without train=False)
```
silvasean, line 155: I would describe this process better as that the constant values are essentially similar to the shapes of the input tensors. Just as by default we need to naively compile for every shape, we also need to compile for every set of non-tensor inputs. Only very carefully can we erase their values.

python/torch_mlir/eager/torch_dispatch.py
```diff
@@ -0,0 +1,353 @@
...
+        node.addInput(inp)
+
+    if node.hasMultipleOutputs():
+        for outp in node.outputs():
+            graph.registerOutput(outp)
+    else:
+        graph.registerOutput(node.output())
+
+    # this name will ultimately be the name of the func op in the mlir module and thus needs to be unique
+    # per op (since we keep the module around), hence smash together all of the things
```
makslevental, line 207: I guess this is necessarily valuable given that we're currently going with the "one module per op" approach...

silvasean (commented):
drive-by comment. will need a few passes here on the review.

python/torch_mlir/eager/torch_dispatch.py
```diff
@@ -0,0 +1,353 @@
...
+    fn = torch._C._create_function_from_graph(fn_name, graph)
+    return fn
+
+
+def build_module(mb, jit_function: ScriptFunction, annotation: Annotation) -> ir.Module:
+    func_op = _get_func_op_with_name(mb.module, jit_function.name)
+    # print("mlir op ir: \n", func_op)
+    assert (
+        func_op is not None
+    ), "Unable to find FuncOp in new module. Make sure function was imported correctly into ModuleBuilder"
```
makslevental, line 218: This should be a `return None` or something like that (alternatively more robust Exception raising and checking)

python/torch_mlir/eager/torch_dispatch.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+class TorchMLIRTensor(torch.Tensor):
+    elem: torch.Tensor
+    __backend = RefBackendLinalgOnTensorsBackend()
+    __slots__ = ["elem"]
+
+    @classmethod
+    def backend(cls):
+        return cls.__backend
```
makslevental, line 278: hacky attempt at static member

python/torch_mlir/eager/torch_dispatch.py
silvasean: Right now we don't have a calling convention that can pass optional types dynamically. So the optionality should be part of the compiled cache key (that is, we specialize for either the None or not-None case).

python/torch_mlir/eager/torch_dispatch.py
silvasean: Yeah, we definitely want this pretty hardened. I would like us to fallback pretty cleanly to the regular aten op, so that this path is a pure value-add and doesn't compromise functionality.

CMakeLists.txt
```diff
@@ -77,6 +77,7 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   # to import lots of dependencies from AddMLIRPython to make this work.
   set(MLIR_ENABLE_BINDINGS_PYTHON 1)
   option(TORCH_MLIR_ENABLE_JIT_IR_IMPORTER "Enables JIT IR Importer" ON)
+  option(TORCH_MLIR_ENABLE_EAGER_MODE "Enables Eager Mode" ON)
```
silvasean, line 80: I think we should have this always on. It doesn't have any new dependency or anything, right? (it doesn't even seem to have C++ code)

e2e_testing/torchscript/xfail_sets.py
```diff
@@ -21,6 +21,21 @@
 }
 REFBACKEND_XFAIL_SET = COMMON_TORCH_MLIR_LOWERING_XFAILS
 
+EAGERMODEBACKED_XFAIL_SET = REFBACKEND_XFAIL_SET.union({
```
silvasean, line 24: as mentioned on chat, we should have clean fallback, so that eager mode "always Just Works"

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,406 @@
...
+    )
+    func_annotation = Annotation(annotations)
+    mb = ModuleBuilder()
+    mb.import_function(script_fun)
+    eager_module = build_module(mb, script_fun, func_annotation)
+
+    qualified_op_name = f"{schema.name}" + (
+        f".{schema.overload_name}" if schema.overload_name else ""
+    )
+    # TODO: replace with no repro report passman
```
silvasean, line 365: Can you address this TODO? If this is still useful for debugging adding a global "enable_debug_mode" could be useful.

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,406 @@
...
+    run_pipeline_with_repro_report(
+        eager_module,
+        "torch-function-to-torch-backend-pipeline,torch-backend-to-linalg-on-tensors-backend-pipeline",
+        qualified_op_name,
+    )
+    compiled = backend.compile(eager_module)
+    jit_module = backend.load(compiled)
+
+    mlir_op = getattr(jit_module, script_fun.name)
+    assert mlir_op is not None
```
silvasean, line 375: So the idea is that if this fails, we catch the exception?

examples/lazytensor_tanh.py
```diff
@@ -26,9 +26,9 @@
     import RefBackendLinalgOnTensorsBackend
 from torch_mlir.passmanager import PassManager
 
-from utils.annotator import Annotation
-from utils.torch_mlir_types import TorchTensorType
-from lazytensor.builder import build_module
+from torch_mlir.eager_mode.annotator import Annotation
```
silvasean, line 29: Can you please copy this code and have it be an internal detail of eager_mode instead of moving it? I really don't think this is a public API that we want people to be seeing. Most of these examples are out of date anyway.

Also, can you add a TODO on Annotation that this is really something that is coupled to and belongs next to `ModuleBuilder.import_function`

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,406 @@
...
+
+def check_supported_op(schema: torch._C.FunctionSchema) -> bool:
+    return (
+        "torch."
+        + schema.name.replace("::", ".")
+        + ("." + schema.overload_name if schema.overload_name else "")
+    ) in SUPPORTED_OPS
+
+
+boolean_dispatched: "weakref.WeakKeyDictionary[Callable, Dict[str, Callable]]" = weakref.WeakKeyDictionary()  # noqa: T484
```
silvasean, line 50: what is this noqa thing? Is any of this code copied? If so, there are licensing implications we need to sort out, such as by isolating the copied code into a separate file like I did for upstream_shape_functions.py in my PR.

silvasean (changes requested):

e2e_testing/torchscript/xfail_sets.py
makslevental: One problem here is the first block of tests that fail. In fact they do not fail (the outputs compare as equal but because [reporting.py](https://github.com/llvm/torch-mlir/blob/5d9222383ce80567911794e4104f336279911a87/python/torch_mlir_e2e_test/torchscript/reporting.py#L156-L159) checks dtypes you get inconsistent reporting and an exception [here](https://github.com/llvm/torch-mlir/blob/5d9222383ce80567911794e4104f336279911a87/python/torch_mlir_e2e_test/torchscript/reporting.py#L250)). As the comment says: this is because, for `.Scalar` variants, what actually happens is that a zero dim tensor `float64` is funneled into these ops at the `torch_dispatch` level.

python/torch_mlir/eager_mode/annotator.py
```diff
@@ -0,0 +1,58 @@
+# -*- Python -*-
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+#
+# pylint: disable=no-member, no-name-in-module, invalid-name, missing-function-docstring, fixme
+
+from typing import Iterable, Union
+from torch.fx import GraphModule
```
silvasean, line 10: remove torch.fx dep

python/torch_mlir/eager_mode/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
+#-------------------------------------------------------------------------------
+# Setup PyTorch
+#-------------------------------------------------------------------------------
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
+TorchMLIRProbeForPyTorchInstall()
```
silvasean, line 6: do we still need to probe this even though it is pure Python?

makslevental: yea i guess not

e2e_testing/torchscript/xfail_sets.py
makslevental: as discussed this is unfortunately an upstream issue; more documentation here https://github.com/pytorch/pytorch/issues/74400

python/torch_mlir/eager_mode/torch_mlir_types.py
```diff
@@ -0,0 +1,133 @@
...
+    to define a `to_mlir(self, ir.Context) -> ir.Type`.
+    Each class is allowed to have different types of
+    __init__ methods depending on the information they
+    require to produce the given MLIR representation.
+    """
+    @abc.abstractmethod
+    def to_mlir(self, context: ir.Context) -> ir.Type:
+        pass
+
+class TorchTensorTypeError(Exception):
```
silvasean, line 41: This is really just part of the API of ir_building.py, right? Can you move it there?

python/torch_mlir_e2e_test/torchscript/configs/eager_mode.py
```diff
@@ -0,0 +1,45 @@
...
+class EagerModeTestConfig(TestConfig):
+    """Trivial test config that exercises eager mode plumbing"""
+
+    def __init__(self):
+        super().__init__()
+
+    def compile(self, program: torch.nn.Module) -> torch.nn.Module:
+        return program
+
+    def run(self, artifact: torch.nn.Module, trace: Trace) -> Trace:
```
silvasean, line 30: how long does it take to run the full test suite with this config? Getting this fast enough that it can be running under CI should be our highest-priority goal (in follow-up commits, if necessary).

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,258 @@
...
+
+def check_supported_op(schema: torch._C.FunctionSchema) -> bool:
+    return (
+        "torch."
+        + schema.name.replace("::", ".")
+        + ("." + schema.overload_name if schema.overload_name else "")
+    ) in SUPPORTED_OPS
+
+
+# shameless SO copy pasta
```
silvasean, line 40: Please, no copypasta! That is bad for licensing reasons.

Note that MLIR allows arbitrary strings as function names, so you shouldn't need to munge this at all. (can just remove this code)

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,258 @@
...
+
+        node.addInput(inp)
+
+    if node.hasMultipleOutputs():
+        for outp in node.outputs():
+            graph.registerOutput(outp)
+    else:
+        graph.registerOutput(node.output())
+
+    # this name will ultimately be the name of the func op in the mlir module and thus needs to be unique
```
silvasean, line 121: Please use proper capitalization and punctutation in the comments throughout this PR: https://llvm.org/docs/CodingStandards.html#commenting

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,258 @@
...
+    if hasattr(op, "op_name"):
+        op_name = op.op_name
+    elif hasattr(op, "__name__"):
+        # handle builtin_function_or_method
+        op_name = op.__name__
+    else:
+        raise RuntimeError(f"op {op} has no name")
+
+    if op_name == "detach":
+        raise RuntimeError("detaching")
```
silvasean, line 222: comment

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,258 @@
...
+    if "layout" in kwargs and kwargs["layout"] is not None:
```
silvasean, line 223: can you explain at a high level "why it is ok" that we change the layout and memory format from what the user requested? Or in other words, why not just raise an error in these cases?

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,258 @@
...
+
+        for i in range(len(tensor_kwargs)):
+            arg, arg_dtype = tensor_kwargs[i]
+            annotations.append(TorchTensorType(shape=tuple(arg.shape), dtype=arg_dtype))
+            tensor_kwargs_flat.append(arg)
+
+    return annotations, tensor_args, tensor_kwargs_flat
+
+
+def fix_inplace(schema, out, all_tensor_args):
```
silvasean, line 183: Comment on what this is doing.

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,258 @@
...
+    )
+
+    eager_module = build_module(script_fun, annotations)
+    with eager_module.context:
+        pm = PassManager.parse(
+            "torch-function-to-torch-backend-pipeline,torch-backend-to-linalg-on-tensors-backend-pipeline"
+        )
+        pm.run(eager_module)
+    compiled = backend.compile(eager_module)
+    jit_module = backend.load(compiled)
```
silvasean, line 251: jit_module isn't a great name. Maybe loaded_module (and call the above "compiled" as "compiled_module" instead).

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,258 @@
...
+    mlir_op = getattr(jit_module, script_fun.name)
```
silvasean, line 252: mlir_op isn't the best name here. Perhaps "op_callable".

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,258 @@
...
+from typing import Any
+from typing import List, Dict
+
+import numpy as np
+import torch
+import torch._C
+from torch.utils._pytree import tree_map
+from torch_mlir._mlir_libs._mlir.passmanager import PassManager
+
+from torch_mlir.dialects.torch.importer.jit_ir.build_tools.registry import Registry
```
silvasean, line 17: This Registry class should only be used by build tools, not production code. Can you write your code to not use it? Doesn't PyTorch have a way to introspect the schema directly?

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,258 @@
...
+        # TODO: what's the cost here?
+        lambda x: x.detach().contiguous().numpy() if isinstance(x, torch.Tensor) else x,
+        normalized_kwargs,
+    )
+
+    annotations = []
+    tensor_args = []
+    for i, arg in enumerate(unwrapped_normalized_args):
+        if isinstance(arg, np.ndarray):
+            # if we get a 0 size dim we need to replace it with symbolic dim
```
silvasean, line 148: IIUC, the issue here is that our backends don't support `tensor<0x...` well, right? Can you be more specific about what goes wrong, or why we need this hack?

silvasean (changes requested):

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
makslevental: changed to raise exception.

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
makslevental: response below doesn't apply here; we don't handle `detach`ing because we don't handle anything related to wengert list (ie grad flow graph) construction.

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
makslevental: `op_mlir_backend_callable` to emphasize that it's not a pytorch callable

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
makslevental: you can get `is_mutable` from the schema object but the alias info doesn't have python bindings:

```
>>> schema.arguments[0].alias_info

  File "/home/mlevental/miniconda3/envs/torch-mlir/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3251, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-15-b9d9b5a0259d>", line 1, in <module>
    schema.arguments[0].alias_info
AttributeError: 'torch._C.Argument' object has no attribute 'alias_info'
```

python/torch_mlir_e2e_test/torchscript/configs/eager_mode.py
silvasean: I would say 1-3 minutes would be acceptable.

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,272 @@
...
+    new_args, new_kwargs = new_args_and_kwargs
+    return new_args, new_kwargs
+
+
+def build_script_function(
+    schema: torch._C.FunctionSchema,
+    args: List[torch._C.Argument],
+    kwargs: Dict[str, Any],
+) -> torch.jit.ScriptFunction:
+    """
```
silvasean, line 80: Nit: elsewhere in Torch-MLIR, we always have a one-line summary on the same line as the opening `"""`. Can you update the docstrings to match that convention?

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,272 @@
...
+    Build a torch.jit.ScriptFunction that corresponds to the schema.
+    Constants are inlined for the purposes of invalidating the compile cache when they change.
+
+    Flow:
```
silvasean, line 84: This is an implementation detail, right? It should be a regular comment in the body, rather than part of the docstring.

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,272 @@
...
+    return annotations, tensor_args, tensor_kwargs_flat
+
+
+def write_back_to_mutable(
+    op_ods: JitOperator,
+    out: Union[np.ndarray, List[np.ndarray]],
+    all_tensor_args: List[np.ndarray],
+):
+    """
+    Because of how we pass values to the backend (by copying the tensor to a numpy array)
```
silvasean, line 183: There's actually two parts here. One is that we copy Torch to numpy, so we need a write-back. Secondly, Torch-MLIR right now only handles arguments with value semantics, so we need to manually fake it, which we can only do right now for these special cases.

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -0,0 +1,272 @@
...
+        np.copyto(all_tensor_args[0], out)
+
+    return out
+
+
+def try_torch_mlir_eager(op, args, kwargs, backend):
+    if hasattr(op, "op_name"):
+        op_name = op.op_name
+    elif hasattr(op, "__name__"):
+        # Handle builtin_function_or_method
```
silvasean, line 223: punctuation at end.

e2e_testing/torchscript/xfail_sets.py
silvasean: nit: Call this EAGER_MODE_XFAIL_SET to make it consistent with the `config_choices`.

silvasean (changes requested):
mostly nits.

python/torch_mlir/eager_mode/torch_mlir_tensor.py
```diff
@@ -0,0 +1,95 @@
...
+from torch.utils._pytree import tree_map
+
+from torch_mlir.eager_mode.torch_mlir_dispatch import (
+    try_torch_mlir_eager,
+    UnsupportedByTorchMlirEagerMode,
+)
+from torch_mlir_e2e_test.linalg_on_tensors_backends import refbackend
+
+
+class TorchMLIRTensor(torch.Tensor):
```
silvasean, line 17: Is there any upstream documentation describing the extension point you are using here? (_make_wrapper_subclass/`__torch_dispatch_``, etc.) It would be good to link it in if it exists.

silvasean (approved):

python/torch_mlir/eager_mode/torch_mlir_tensor.py
makslevental: I spoke with brian hirsch and he said the best documentation for how to use wrapper_subclass is https://github.com/albanD/subclass_zoo. I can add as a comment.

PR #641, opened 2022-03-03
Add hugging face `albert-base-v2` in torchscript_e2e_heavydep_tests

pashu123:
`albert-base-v2` for sequence classification is added in e2e_heavy_test.


pashu123:
@cathyzhyi There are many hugging face models that can be placed under the same class `AutoModelForSequenceClassification` and hence I have changed the minilm_sequence_classification.py to hf_sequence_classification to take care of all hugging face sequence classification models.

silvasean (approved):

PR #642, opened 2022-03-04
Fix handling of overwrite of view in maximize-value-semantics

ramiro050:
In order to replace a tensor overwrite with copies of the tensor being
used to overwrite it, the rewrite pattern
`AbstractlyInterpretCopyToNonValueTensorOpUsersWithinABlock` needs the
overwritten tensor to come from a `CopyToNonValueTensorOp`. This
commit adds a pattern that ensures that each overwrite of a tensor is
overwriting the result of a `CopyToNonValueTensorOp`.


test/Dialect/Torch/maximize-value-semantics.mlir
```diff
@@ -59,6 +59,17 @@ func @mutation_followed_by_view_like_ops(%value_t: !torch.vtensor, %overwriter:
...
 }
 
+// CHECK-LABEL:   func @mutation_of_view_like_op_result(
+// CHECK-SAME:                                             %[[VALUE_T:.*]]: !torch.vtensor, %[[OVERWRITER:.*]]: !torch.vtensor, %[[INT_LIST:.*]]: !torch.list<!torch.int>) -> !torch.vtensor {
+// CHECK:           return %[[OVERWRITER]] : !torch.vtensor
+func @mutation_of_view_like_op_result(%value_t: !torch.vtensor, %overwriter: !torch.vtensor, %int_list: !torch.list<!torch.int>) -> !torch.vtensor {
+  %t = torch.copy.to_tensor %value_t : !torch.tensor
+  %view = torch.aten.view %t, %int_list : !torch.tensor, !torch.list<!torch.int> -> !torch.tensor
+  torch.overwrite.tensor.contents %overwriter overwrites %view : !torch.vtensor, !torch.tensor
+  %result = torch.copy.to_vtensor %view : !torch.vtensor
```
cathyzhyi, line 69: I think we also need to make sure there is not any use of `%t` after the `torch.overwrite.tensor.contents`.  The problem with overwriting the result of a view like op is that the result is aliased with the input operands of view like ops ((or other ops with aliasing). 

I am not sure how can we model the case where there is indeed use of `%t` after the overwrite but maybe that's not common in real models? @silvasean Do you have any comments?

ramiro050: That's a really good point. Yeah if there's a use of `%t` after the overwrite, that would complicate the analysis. At the moment I can't think of a way of doing it. However, verifying that the intermediate values of view-like ops between the `torch.copy.to_tensor` and the `torch.overwrite.tensor.contents` are not used after the overwrite should be straight forward.

cathyzhyi: I am thinking we can handle both view like ops and overwrites in the same pattern. What I have in mind is combining `AbstractlyInterpretCopyToNonValueTensorOpUsersWithinABlock` and `RewriteViewLikeSubgraph` into one conversion pattern. 

First step is collecting all the aliasing non-value tensors similar to this part https://github.com/llvm/torch-mlir/blob/486f95e84f587d020ba789b071b12f890510f1a1/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp#L130-L141. In addition to collecting the users as the existing code, all the non-value tensors need to be collected as well. Let's call this aliasing set for now.

The second step is doing abstract interpretation similarly to `AbstractlyInterpretCopyToNonValueTensorOpUsersWithinABlock`.  All the collected users in the first step needs to be sorted. At each operation, only one value from the aliasing set can be available (meaning its current value can be replaced with a value tensor). At the `CopyToNonValueTensorOp`, its result non value tensor is available with value of its operand. Whenever, a `overwrite` is seen, the overwritten non value tensor become available and others in the aliasing set are unavailable. If any uses of an unavailable non value tensor is seen the pattern fails.  If the interpretation finished successfully, all the overwrites can be removed. All uses of the aliasing set should be replaced with corresponding available value tensor.

silvasean: Yi, that seems like a good generalization of the current logic. One thing to keep in mind is the handling of "read-only" ops in nested subregions (like in If/Loop). From the perspective of this algorithm, they can be treated as if they were in the parent block. If there are mutating ops in subregions, then we cannot abstractly interpret past them (without doing something a lot more sophisticated AFAICT).

ramiro050:
I have updated this patch with @cathyzhyi's algorithm. The only difference made to the algorithm is that I allow multiple aliases to be available at any given op. This is needed for cases like this one

https://github.com/llvm/torch-mlir/blob/3d9ba5e525a64b1de0081bd4349dac1266be9cb1/test/Dialect/Torch/maximize-value-semantics.mlir#L138-L151

where the last `unsqueeze` needs to know that alias `%1` is available even though a new alias `%2` was created after alias `%1`.

The only time that the number of aliases available is restricted to 1 is right after an overwrite takes place.

I also separated the interpretation/validation from the rewriting because it was very hard to not leave the graph in an invalid state when an invalid case was encountered when doing the validation and rewriting together.

The case @silvasean mentioned for `if/loop` is not yet implemented. I will add it in a follow up patch. I don't think it will be too complicated, but it requires one to be a bit more careful in the interpretation. Namely, the sorting function has to take into account ops in sub-blocks.

Please re-review.

test/Dialect/Torch/maximize-value-semantics.mlir
```diff
@@ -59,6 +59,28 @@ func @mutation_followed_by_view_like_ops(%value_t: !torch.vtensor, %overwriter:
...
+  return %result : !torch.vtensor
+}
+
+// CHECK-LABEL:   func @value_tensor_used_after_copy_was_mutated(
+// CHECK-SAME:                                                       %[[VALUE_T:.*]]: !torch.vtensor,
+// CHECK-SAME:                                                       %[[OVERWRITER:.*]]: !torch.vtensor) -> (!torch.vtensor, !torch.vtensor) {
+// CHECK:           return %[[VALUE_T]], %[[OVERWRITER]] : !torch.vtensor, !torch.vtensor
+func @value_tensor_used_after_copy_was_mutated(%value_t: !torch.vtensor, %overwriter: !torch.vtensor) -> (!torch.vtensor, !torch.vtensor) {
```
ramiro050, line 77: I added this test to ensure that the `overwrite` op is always rewritten before the `copy.to_tensor` op. If the copy op is rewritten first, then the overwritten tensor will be replaced with `%value_t`, which means that when the overwrite is rewritten, the return of `%value_t` will be replaced with `%overwriter`, which is not correct.

cathyzhyi (approved):
Great work!

lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
```diff
@@ -37,136 +37,169 @@ class AbstractlyInterpretCopyToNonValueTensorOpUsersWithinABlock
...
   }
-};
-} // namespace
 
-namespace {
-// Calculate a forward slice starting from a CopyToNonValueTensorOp
-// and ending at CopyToValueTensorOp's. If all intervening ops
-// are just view-like operations (i.e. no mutation), then we can trivially
-// convert them all to value semantics.
-class RewriteViewLikeSubgraph
```
cathyzhyi, line 112: The `RewriteViewLikeSubgraph` still has value in cases where the use-def tree spanning multiple blocks with branches but with no writes. We can keep it until a more sophisticated approach is implemented.

silvasean: I think that the improved version that Ramiro is working on should handle this case too. Looking forward to having these two fully unified!

ramiro050: I've added back the `RewriteViewLikeSubgraph` in the meantime while I work on improving the abstract interpretation pas

PR #643, opened 2022-03-04
Add optional int list ODS type

gpetters94:


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -226,6 +226,7 @@ def get_by_triple(self, key: Tuple[str, str, str]):
     "int": "Torch_IntType",
     "int[]": "TorchIntListType",
     "int?": "TorchOptionalIntType",
+    "int[]?": "TorchOptionalIntListType",
```
cathyzhyi, line 229: There is some confusion in naming. "Tensor?[]": "AnyTorchOptionalTensorListType" means the tensor element is optional but "int[]?": "TorchOptionalIntListType" means the list is optional.

Can you help to change all list type to ListOfXXXType. Optional list type becomes OptionalListOfXXXType and when list element is optional ListOfOptionalXXXType.

cathyzhyi (changes requested):

gpetters94:
Are these names acceptable? Some of them are kind of unwieldy.

cathyzhyi:
@silvasean Hey Sean, do you have any comments on the names? This is for addressing https://github.com/llvm/torch-mlir/pull/643#discussion_r819681492 to make the difference between `Tensor[]?` and `Tensor?[]`

silvasean:
Following ODS conventions I think we should call these `AnyTorch*` (except the ones that are `Torch_X` naturally). So I would go with `Tensor[]?` as `AnyTorchOptionalListOfTensorType` and `Tensor?[]` as `AnyTorchListOfOptionalTensorType` (and `Tensor?[]?` would be `AnyTorchOptionalListOfOptionalTensorType`). I think the important thing is that they read left-to-right directly and consistently (so you just reverse the reading of the torch type string: `Tensor[]?` -> `Tensor, [], ?` --(reverse)--> `?, [], Tensor` -> `AnyTorchOptionalListOfTensorType`)

cathyzhyi (approved):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -20,15 +20,16 @@
...
     "Tensor?": "AnyTorchOptionalTensorType",
-    "Tensor?[]": "AnyTorchOptionalTensorListType",
-    "Tensor[]": "AnyTorchTensorListType",
+    "Tensor?[]": "AnyTorchListOfOptionalTensorType",
+    "Tensor[]": "AnyTorchListOfTensorType",
     "Scalar": "AnyTorchScalarType",
     "Scalar?": "AnyTorchOptionalScalarType",
     "int": "Torch_IntType",
-    "int[]": "TorchIntListType",
+    "int[]": "ListOfTorchIntType",
```
silvasean, line 28: AnyTorchListOfIntType

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -20,15 +20,16 @@
...
     "int?": "TorchOptionalIntType",
```
silvasean, line 29: AnyTorchOptionalIntType

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -20,15 +20,16 @@
...
+    "int[]?": "OptionalListOfTorchIntType",
```
silvasean, line 30: AnyTorchOptionalListOfIntType

cathyzhyi:
@gpetters94 the change looks good. You can go ahead and merge when you are ready.

PR #646, opened 2022-03-08
add one mnist example, please make sure if it's okay,  -s

JasonMaojinsong:


pashu123:
@JasonMaojinsong I think an e2e inference test is added here: https://github.com/llvm/torch-mlir/blob/main/e2e_testing/torchscript/mlp.py If you can add your test here.

silvasean:
We already have a ResNet18 example for ImageNet. This example seems very similar to that one, so I'm not sure we should add it.

JasonMaojinsong:
> We already have a ResNet18 example for ImageNet. This example seems very similar to that one, so I'm not sure we should add it. 

@silvasean,  thnaks, it's my local example for small case debugging, looks it can't provide too much info compared to resnet18 e2e example,   I will close this pull request.

PR #648, opened 2022-03-09
Aten range op

erman-gurses:
I do this pull request for debugging purposes. The RangeIntModule is still failing.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1189,6 +1189,65 @@ class DecomposeAtenArangeStartOp : public OpRewritePattern<AtenArangeStartOp> {
...
+// The `aten.range.step` op is converted to `aten.arange.start_step` op.
+class DecomposeAtenRangeStepOp : public OpRewritePattern<AtenRangeStepOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenRangeStepOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+
+    double_t end;
+
+    if (!matchPattern(op.end(), m_TorchConstantFloat(&end))) {
```
ramiro050, line 1236: This is looking good! I just have a suggestion for this part. I think it might be good to add lowerings for the ops

- `aten::add.int : (int, int) -> (int)`
- `aten::add.int_float : (int, float) -> (float)`

This allows you to do something like

```cpp
Value one = rewriter.create<Torch::ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
Value end;
Type endType = op.end().getType();
if (endType.isa<Torch::FloatType>())
  end = rewriter.create<Torch::AddIntFloatOp>(loc, one, op.end());
else if (endType.isa<Torch::IntType>())
  end = rewriter.create<Torch::AddIntOp>(loc, one, op.end());
else
```

This removes the dependency of having `op.end()` be `torch.constant.float` or `torch.constant.int`.

ramiro050, line 1236: This is looking good! I just have a suggestion for this part. I think it might be good to add lowerings for the ops

- `aten::add.int : (int, int) -> (int)`
- `aten::add.int_float : (int, float) -> (float)`

This allows you to do something like

```cpp
Value one = rewriter.create<Torch::ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
Value end;
Type endType = op.end().getType();
if (endType.isa<Torch::FloatType>())
  end = rewriter.create<Torch::AddIntFloatOp>(loc, one, op.end());
else if (endType.isa<Torch::IntType>())
  end = rewriter.create<Torch::AddIntOp>(loc, one, op.end());
else
```

This removes the dependency of having `op.end()` be `torch.constant.float` or `torch.constant.int`.

erman-gurses: Thanks! Actually, when I was implementing that part, I was looking for something like you mentioned above. Then figured it out there no op like that. These might be second practice for me after I have done this one.

ramiro050 (changes requested):
Good job on your first op! I have a few comments below

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1191,6 +1191,189 @@ class DecomposeAtenArangeStartOp : public OpRewritePattern<AtenArangeStartOp> {
...
+      }
+      if (op.step().getType().isa<Torch::IntType>()) {
+        int64_t stepInt;
+        if(!matchPattern(op.step(), m_TorchConstantInt(&stepInt)))
+          return rewriter.notifyMatchFailure(
+              op, "step should be torch constant of integer type.");
+        step = (double_t) stepInt;
+      }
+      if((start >= 0 && end >= 0) || (start <= 0 && end <= 0)){
+        int diff = std::abs(end) - std::abs(start);
```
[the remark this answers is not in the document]
erman-gurses: Thanks for the good catch, I handle that gap.

PR #650, opened 2022-03-09
[MLIR][TORCH] Add value tensor variant to aten::_index_put_impl_

vivekkhandelwal1:
This commit adds the op `PseudoAtenIndexPutImplOp` that represents
`Aten_IndexPutImpl_Op` without the underscore. This is needed to
make sure that the `ReduceOpVariants` pass turns the in-place op
into an op that takes value tensors as inputs, otherwise the
semantics correctly.

This commit also adds the lowering of `PseudoAtenIndexPutImplOp` op.

### DIFF
--- a/e2e_testing/torchscript/index_put.py
+++ b/e2e_testing/torchscript/index_put.py
@@ -1,0 +1,112 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+
+import torch
+
+from torch_mlir_e2e_test.torchscript.framework import TestUtils
+from torch_mlir_e2e_test.torchscript.registry import register_test_case
+from torch_mlir_e2e_test.torchscript.annotations import annotate_args, export
+
+# ==============================================================================
+
+
+class IndexPutImplOneDimFloatNonAccumulateModule(torch.nn.Module):
+
+  def __init__(self):
+    super().__init__()
+
+  @export
+  @annotate_args([
+      None,
+      ([-1], torch.float32, True),
+      ([-1], torch.int64, True),
+      ([-1], torch.float32, True),
+  ])
+  def forward(self, input, index, value):
+    return torch.ops.aten._index_put_impl_(input, (index,), value,
+                                          accumulate=False,
+                                          unsafe=False)
+
+
+@register_test_case(module_factory=lambda: IndexPutImplOneDimFloatNonAccumulateModule())
+def IndexPutImplOneDimFloatNonAccumulateModule_basic(module, tu: TestUtils):
+  module.forward(tu.rand(100), torch.randint(100, (250,)),
+                 tu.rand(250))
+
+
+class IndexPutImplOneDimIntNonAccumulateModule(torch.nn.Module):
+
+  def __init__(self):
+    super().__init__()
+
+  @export
+  @annotate_args([
+      None,
+      ([-1], torch.int64, True),
+      ([-1], torch.int64, True),
+      ([-1], torch.int64, True),
+  ])
+  def forward(self, input, index, value):
+    return torch.ops.aten._index_put_impl_(input, (index,), value,
+                                          accumulate=False,
+                                          unsafe=False)
+
+
+@register_test_case(module_factory=lambda: IndexPutImplOneDimIntNonAccumulateModule())
+def IndexPutImplOneDimIntNonAccumulateModule_basic(module, tu: TestUtils):
+  module.forward(torch.randint(1000, (200,)), torch.randint(100, (300,)),
+                 torch.randint(10000, (300,)))
+
+
+class IndexPutImplOneDimFloatAccumulateModule(torch.nn.Module):
+
+  def __init__(self):
+    super().__init__()
+
+  @export
+  @annotate_args([
+      None,
+      ([-1], torch.float32, True),
+      ([-1], torch.int64, True),
+      ([-1], torch.float32, True),
+  ])
+  def forward(self, input, index, value):
+    # Since the input is updated in-place, we pass input.clone() in place
+    # of input to avoid wrong results.
+    return torch.ops.aten._index_put_impl_(input.clone(), (index,), value,
+                                          accumulate=True,
+                                          unsafe=False)
+
+
+@register_test_case(module_factory=lambda: IndexPutImplOneDimFloatAccumulateModule())
+def IndexPutImplOneDimFloatAccumulateModule_basic(module, tu: TestUtils):
+  module.forward(tu.rand(1000), torch.randint(10, (500,)),
+                 tu.rand(500))
+
+
+class IndexPutImplOneDimIntAccumulateModule(torch.nn.Module):
+
+  def __init__(self):
+    super().__init__()
+
+  @export
+  @annotate_args([
+      None,
+      ([-1], torch.int64, True),
+      ([-1], torch.int64, True),
+      ([-1], torch.int64, True),
+  ])
+  def forward(self, input, index, value):
+    # Since the input is updated in-place, we pass input.clone() in place
+    # of input to avoid wrong results.
+    return torch.ops.aten._index_put_impl_(input.clone(), (index,), value,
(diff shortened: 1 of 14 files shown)


cathyzhyi (changes requested):

ramiro050 (changes requested):

e2e_testing/torchscript/index_put.py
```diff
@@ -0,0 +1,56 @@
...
+
+  def __init__(self):
+    super().__init__()
+
+  @export
+  @annotate_args([
+      None,
+      ([10], torch.float32, True),
+      ([6], torch.int64, True),
+      ([6], torch.float32, True),
```
ramiro050, line 25: Not sure why the tests are failing, but I did notice some strange behavior. If you set the sizes for the `index` and `value` in the `annotate_args` to zero, the tests pass! I also noticed the following:

Assume that `input = torch.zeros(10, dtype=torch.float)`, `index = torch.tensor([0,0,0], dtype=torch.int64)`, and `value = torch.tensor([3,4,5], dtype=torch.float)`. 

If you set the sizes for the `index` and `value` in `annotate_args` to:

- `0` then `output[0] = 3 + 4 + 5`
- `1` then `output[0] = (3 + 3) + 4 + 5`
- `2` then `output[0] = (3 + 3) + (4 + 4) + 5`
- `3` then `output[0] = (3 + 3) + (4 + 4) + (5 + 5)`

This is why your test is failing. For some reason every value gets doubled before getting added to the tensor. I can't really think of what could be causing this behavior. @cathyzhyi, have you seen anything like this?

cathyzhyi: I am still investigating this.  Your experiment result looks interesting! It almost look like the testing run is reusing the tensor from the baseline run.

cathyzhyi: yea, if you print the `input` here https://github.com/llvm/torch-mlir/blob/1a2a9e066f320e725927a7094b5a668eae358453/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/linalg_on_tensors_backend.py#L53 It shows the `input` of testing run contains the result from the baseline run already because the torchscript run would update the `input` in place.

ramiro050: Cool! That also explains why setting the `yieldValue` to the original value was working in the case of no accumulation.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -176,6 +176,123 @@ class ConvertAtenBincountOp : public OpConversionPattern<AtenBincountOp> {
...
+          op, "Input element type should be same as the values element type.");
+
+    SmallVector<Value> indicesList;
+    getListConstructElements(adaptor.indices(), indicesList);
+    // The size of the list of the index tensors should not be greater than the
+    // input rank.
+    if ((int64_t)indicesList.size() > inputType.getRank())
+      return rewriter.notifyMatchFailure(
+          op, "Indices list size cannot be greater than the input rank.");
+
```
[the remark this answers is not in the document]
vivekkhandelwal1: Similar code has been added by you here: https://github.com/llvm/torch-mlir/blob/1a2a9e066f320e725927a7094b5a668eae358453/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp#L2869-L2871

e2e_testing/torchscript/index_put.py
vivekkhandelwal1: What could be the solution to this? @ramiro050 @cathyzhyi

e2e_testing/torchscript/index_put.py
```diff
@@ -0,0 +1,56 @@
...
+  ])
+  def forward(self, input, index, value):
+    return torch.ops.aten._index_put_impl_(input, (index,), value, True, False)
+
+
+@register_test_case(module_factory=lambda: IndexPutImplOneDimensionalFloatModule())
+def IndexPutImplOneDimensionalFloatModule_basic(module, tu: TestUtils):
+  module.forward(tu.rand(10), torch.tensor([0, 4, 1, 3, 6, 4]),
```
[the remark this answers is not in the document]
vivekkhandelwal1: I will do this once the test cases pass.

e2e_testing/torchscript/index_put.py
vivekkhandelwal1: As of now, can we go ahead with this temporary fix?

ramiro050: Yeah, I think they would be fine

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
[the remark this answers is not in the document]
vivekkhandelwal1: > Similar code has been added by you here:
> 
> https://github.com/llvm/torch-mlir/blob/1a2a9e066f320e725927a7094b5a668eae358453/lib/Conversion/TorchToLinalg/TorchToLinalg.cpp#L2869-L2871

I thought you were asking something related to the condition in the original code. That's why I referred to this code. I guess I misunderstood your comment.

e2e_testing/torchscript/index_put.py
silvasean: We will eventually need to extend the testing framework to correctly check that inputs are updated in place. However, in Torch-MLIR we don't really have runtime support for that (note that we set `True` on the annotations saying that the input should be treated with value semantics, i.e. in-place updates are not externally visible from the compiled code). So I think the right solution for now is for the e2e test framework to clone inputs before passing them to torchscript or native_torch functions. I.e., I think the right fix at the moment is to clone tensor inputs in these two lines:

https://github.com/llvm/torch-mlir/blob/1a2a9e066f320e725927a7094b5a668eae358453/python/torch_mlir_e2e_test/torchscript/configs/torchscript.py#L31

https://github.com/llvm/torch-mlir/blob/1a2a9e066f320e725927a7094b5a668eae358453/python/torch_mlir_e2e_test/torchscript/configs/native_torch.py#L27

We could make Trace.inputs a private member Trace._inputs and then have a method `Trace.get_inputs()` which clones appropriately. Then we can remove this manual input.clone()

Can you do that in a follow-up patch?

vivekkhandelwal1: Okay, I will do this in a follow-up patch.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -176,6 +176,135 @@ class ConvertAtenBincountOp : public OpConversionPattern<AtenBincountOp> {
...
+    if (inputType.getElementType() != valuesType.getElementType())
+      return rewriter.notifyMatchFailure(
+          op, "Input element type should be same as the values element type.");
+
+    SmallVector<Value> indicesList;
+    getListConstructElements(adaptor.indices(), indicesList);
+    // The size of the list of the index tensors should not be greater than the
+    // input rank.
+    if ((int64_t)indicesList.size() > inputType.getRank() &&
+        indicesList.size() > 0)
```
cathyzhyi, line 231: `indicesList.size() > 0` can be removed

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -176,6 +176,135 @@ class ConvertAtenBincountOp : public OpConversionPattern<AtenBincountOp> {
...
+      return rewriter.notifyMatchFailure(
+          op, "Indices list size should not be greater than the input rank.");
+
+    // TODO: Add support for cases with indices list size smaller than the input
+    // rank and indices list of size zero.
+    if ((int64_t)indicesList.size() < inputType.getRank() ||
+        indicesList.size() == 0)
```
cathyzhyi, line 238: `indicesList.size() == 0` can be removed as it's included in the previous condition.

cathyzhyi (changes requested):

vivekkhandelwal1:
Hi @ramiro050 @cathyzhyi, could you please re-review this PR?

cathyzhyi (changes requested):
Looks mostly good except one place.

cathyzhyi (approved):

PR #655, opened 2022-03-10
Add convolution_backward and convolution_backward_overrideable

gpetters94:
I've been working on an implementation for `convolution_backward` and I'm running into a roadblock that I hope some fresh eyes will help me figure out. I've got the 1D case almost finished (except the bias gradient) but I'm not getting the expected results for the input and weight gradients. The weight gradient in particular is orders of magnitude off.


silvasean:
I would suggest that you run your test case and inspect the results with highly structured inputs, rather than random inputs. For example, see how the code behaves with a kernel that is [1, 0, 0] instead of [rand(), rand(), rand()]. Then it should be easier to pinpoint the issues.

gpetters94:
Going to reopen this so it's a bit cleaner since I'm almost done.

PR #659, opened 2022-03-11
Add maximize-value-semantics support for ops with two tensor inputs

ramiro050:
This commit adds value semantics support for ops such as
`aten.view_as` and `aten.expand_as` that take two tensors as input.

This is needed for the implementation of `aten.expand_as` that 
@vivekkhandelwal1 is currently working on.

I also simplified a few things in the implementation.

### DIFF
--- a/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
+++ b/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
@@ -21,12 +21,18 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 
+static Value assertNonValueTensor(Value tensor) {
+  assert(tensor.getType().isa<NonValueTensorType>() &&
+         "tensor is expected to be a non-value tensor");
+  return tensor;
+}
+
 static bool isViewLikeOp(Operation *op) {
   // AtenContiguousOp might return a view, so this is conservatively
   // correct. We could potentially be more precise and identify the cases
   // that it does not return a view and treat those as having value
   // semantics.
-  return isa<AtenBroadcastToOp, AtenContiguousOp, AtenExpandOp,
+  return isa<AtenBroadcastToOp, AtenContiguousOp, AtenExpandAsOp, AtenExpandOp,
              AtenFlattenUsingIntsOp, AtenPermuteOp, AtenReshapeOp,
              AtenSelectIntOp, AtenSliceTensorOp, AtenSqueezeDimOp,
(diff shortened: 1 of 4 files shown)


lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
```diff
@@ -148,40 +150,41 @@ class AbstractlyInterpretCopyToNonValueTensorOpUsersWithinABlock
...
-
-        // We currently only support view-like ops with one tensor input and one
-        // tensor output, meaning that the tensor use-def chains form a tree.
-        // This will not be the case for an op like `torch.aten.view_as`, so
-        // we will need to add a set to prune duplicate visitation.
-        if (llvm::count_if(op->getOperands(), isTensor) != 1 ||
-            llvm::count_if(op->getResults(), isTensor) != 1 ||
-            !isTensor(op->getOperand(0)) || !isTensor(op->getResult(0))) {
+        // We currently only support view-like ops with one tensor output.
+        if (op->getNumResults() == 0 ||
```
cathyzhyi, line 168: We might want to check `op->getNumResults() != 1` here just in case.

lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
```diff
@@ -148,40 +150,41 @@ class AbstractlyInterpretCopyToNonValueTensorOpUsersWithinABlock
...
         }
 
-        llvm::append_range(workList, op->getResult(0).getUsers());
+        Value opResult = op->getResult(0);
+        // There are cases where a view-like op will be partially converted to
+        // value semantics, resulting in at least one of the inputs being a
+        // non-value tensor and the output being a value tensor. If this is the
+        // case then there is no need to look at the users of the result of the
+        // op.
+        if (opResult.getType().isa<NonValueTensorType>())
```
cathyzhyi, line 181: There are two assumptions here:
1. the first operand is aliased with the single result
2. the `copyToNonValueTensor` of the first operand is processed before the second tensor operand.

For 1, I think there is currently no accurate way to tell which operand aliasing with which result from a torch op at this moment. The jit operator schema does provide this information so we can add that in the future. For now, we might want to add a comment about this.

For 2, We could add some guard.

cathyzhyi (changes requested):

lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
ramiro050: Done!

cathyzhyi (approved):

PR #669, opened 2022-03-16
Improve "pseudo" op terminology.

silvasean:
The term "pseudo" is very vague and was getting confusing (I felt I had
to explain it in every comment referencing it). Instead, rework the
"pseudo" ops to instead be named:

- MLIR Syntax: `torch.valsem.*`
- C++ / ODS: `ValsemVariant*Op`

This makes it clear what the concept is, and avoids confusion with other
things that might be called "pseudo", since these are very specific and
should be 100% consistently named w.r.t. the non-valsem-variant ops that
they correspond to.


silvasean:
Ignore the "elide !torch." commit. These two commits don't commute, and GitHub doesn't allow basing one PR on top of another.

silvasean:
Landed that PR. Ignore previous comment.

cathyzhyi (approved):

Issue #665, opened 2022-03-15
`aten::native_layer_norm` 2nd and 3rd results not implemented correctly


silvasean:
When revamping the shape library, I noticed that the shape function for `torch.aten.native_layer_norm` was not implemented correctly for the 2nd and 3rd results. I tried fixing the buggy shape function, but it seems that the TorchToLinalg lowering broke if I did that (the shape function actually calculates the wrong rank).

I left some comments (including a correct version of the code) in shape_lib_gen.py about the issue and expected behavior. 
https://github.com/llvm/torch-mlir/blob/a5fe0cf06308af3a372e40ae927995f7920fb55d/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py#L766

Our testing didn't catch this because for some reason the test was only checking the first result, rather than all 3:
https://github.com/llvm/torch-mlir/blob/a5fe0cf06308af3a372e40ae927995f7920fb55d/e2e_testing/torchscript/norm_like.py#L194

To fix this bug, you should be able to uncomment the correct code linked above in shape_lib_gen.py, and then debug what TorchToLinalg is doing wrong.

cathyzhyi:
I think this is covered in https://github.com/llvm/torch-mlir/pull/570 by @gprateek93.  Prateek, could you split  https://github.com/llvm/torch-mlir/pull/570 and check in the part that fixes this issue?

gprateek93:
> I think this is covered in #570 by @gprateek93. Prateek, could you split #570 and check in the part that fixes this issue?

Sure.

gprateek93:
Refer to this PR: [#674](https://github.com/llvm/torch-mlir/pull/674)

PR #674, opened 2022-03-16
[TORCH][MLIR] Fix the return types of `aten.native_layer_norm`.

gprateek93:
This commit fixes the 2nd and 3rd return types of the `aten.native_layer_norm`.
Previously the mean and rSTD were returned with reduction dims removed.
This commit fixes this and keeps the reduction dims of the results.

Signed-Off-By: Prateek Gupta <prateek@nord-labs.com>

fixes: #665

### DIFF
--- a/e2e_testing/torchscript/norm_like.py
+++ b/e2e_testing/torchscript/norm_like.py
@@ -204,13 +204,33 @@ class NativeLayerNormModule(torch.nn.Module):
     ])
     def forward(self, x, weight, bias):
         list = [2, 2, 3]
-        # TODO: Fix the case of the other return values.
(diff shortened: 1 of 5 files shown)


cathyzhyi (approved):
Thanks for the fix!

e2e_testing/torchscript/norm_like.py
```diff
@@ -206,13 +206,34 @@ def forward(self, x, weight, bias):
         list = [2, 2, 3]
         # TODO: Fix the case of the other return values.
```
ramiro050, line 207: This TODO is no longer needed, right?

silvasean:
Thanks for the quick follow-up!

PR #676, opened 2022-03-16
[MLIR][TORCH] Add value tensor variant to aten::copy_ op

vivekkhandelwal1:
This commit adds the op `ValsemVariantAtenCopyOp` that represents
`AtenCopy_Op` without the underscore. This is needed to make sure
that the `ReduceOpVariants` pass turns the in-place op into an op
that takes value tensors as inputs, otherwise the
semantics correctly.

This commit also adds the lowering of `ValsemVariantAtenCopyOp`.

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -878,6 +878,90 @@ public:
 };
 } // namespace
 
+// Broadcasts input tensor based on the broadcastToShape.
+static LogicalResult broadcastToGivenShape(Operation *op,
+                                           ConversionPatternRewriter &rewriter,
+                                           Value input,
+                                           SmallVector<Value> broadcastToShape,
+                                           Value &result) {
+  RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+  ArrayRef<int64_t> inputShape = inputType.getShape();
+  if (broadcastToShape.size() < inputShape.size()) {
+    return rewriter.notifyMatchFailure(
+        op, "invalid shape: broadcastToShape size must not be smaller than the "
+            "size of the input shape");
+  }
+
+  Type elementType = inputType.getElementType();
+  Location loc = op->getLoc();
+  MLIRContext *context = op->getContext();
+  SmallVector<Value> outShape;
+
+  // Create affine map and shapes for tensor initialization.
+  SmallVector<AffineExpr> outExpr;
+  Value zero =
+      rewriter.create<arith::ConstantOp>(loc, rewriter.getI64IntegerAttr(0));
+  size_t diff = broadcastToShape.size() - inputShape.size();
+  for (size_t i = 0; i < broadcastToShape.size(); i++) {
+    Value shapeValue = broadcastToShape[i];
+    size_t j = i - diff;
+    if (i < diff) {
+      Value isValid = rewriter.create<arith::CmpIOp>(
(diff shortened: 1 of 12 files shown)


lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -878,6 +878,88 @@ class ConvertAtenCatOp : public OpConversionPattern<AtenCatOp> {
...
 } // namespace
 
+// Broadcasts input tensor based on the broadcastToShape.
+template <typename OpTy>
+static LogicalResult
+broadcastToGivenShape(OpTy op, ConversionPatternRewriter &rewriter, Value input,
+                      SmallVector<Value> broadcastToShape, Value &result) {
+  RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+  ArrayRef<int64_t> inputShape = inputType.getShape();
+  if (broadcastToShape.size() < inputShape.size())
```
ramiro050, line 888: `if` needs curly braces

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -878,6 +878,88 @@ class ConvertAtenCatOp : public OpConversionPattern<AtenCatOp> {
 };
 } // namespace
 
+// Broadcasts input tensor based on the broadcastToShape.
+template <typename OpTy>
+static LogicalResult
+broadcastToGivenShape(OpTy op, ConversionPatternRewriter &rewriter, Value input,
```
ramiro050, line 884: I don't think this needs to be a template. I also noticed that there is nothing in the body of the function that requires information about the `op`. The `op` is only used to get a location and a context, however you can get a context from the rewriter itself. Here, I would use a similar signature to other helper functions

```cpp
broadcastToGivenShape(PatternRewriter &rewriter, Location loc, ...)
```

This makes it clear to the user that what this function does is `op`-independent.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -992,6 +1009,73 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
...
+    }
+
+    // The size of the src tensor can be different from the self but should be
+    // broadcastable. Therefore, broadcasting the src tensor to match the size
+    // of the self tensor.
+    SmallVector<Value> selfSizes = getTensorSizes(rewriter, loc, self);
+    for (unsigned i = 0; i < selfSizes.size(); i++)
+      selfSizes[i] = castIndexToInt(rewriter, loc, selfSizes[i]);
+    Value broadcastedSrc;
+    if (failed(broadcastToGivenShape<ValsemVariantAtenCopyOp>(
```
ramiro050, line 1047: `if` needs curly braces

ramiro050 (changes requested):
Thanks for refactoring the broadcast code!

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: Never mind, you do need the `op` as an argument to create the `notifyMatchFailure`. So you should keep the signature unchanged. However, I would remove the template code and pass the `op` as `Operation *op`.

ramiro050 (approved):
LGTM! Just one small comment

pashu123 (approved):
LGTM!

Issue #673, opened 2022-03-16
Bump LLVM to support upstream breaking change


Vigilans:
In https://github.com/google/iree/pull/8542, `IREE` has integrated LLVM to commit https://github.com/llvm/llvm-project/commit/8361c5da30588d3d4a48eae648f53be1feb5cfad,  among which a breaking change commit https://github.com/llvm/llvm-project/commit/7294be2b8e9ada57e051bac6a20ca8b4dbf46475 in pretty-print is introduced:

> A side-effect of the change is that the pretty printed form changes from:
> ```
> %1 = linalg.fill(%cst, %0) : f32, tensor<?x?xf32> -> tensor<?x?xf32>
> ```
> changes to
> ```
> %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<?x?xf32>) -> tensor<?x?xf32>
> ```

As a result, `.mlir` model generated by `torch-mlir` cannot be compiled with `IREE` in latest main branch, which will raise compilation error on `linalg.fill`.

Therefore, it may be suitable for `torch-mlir` to bump LLVM version to support this upstream breaking change.

---

I bumped LLVM version to the same commit as in `google/iree`, and had managed to apply patches myself to produce a working `torch-mlir` build (also produce `.mlir` model that `IREE` can compile). The commits are published in [my fork of torch-mlir](https://github.com/Vigilans/torch-mlir/tree/vigilans/llvm-bump), including:

* [Adaptation to LLVM's commit 23aa5a744666b281af807b1f598f517bf0d597cb](https://github.com/llvm/torch-mlir/commit/d8ebfe4110a0354b085a9637ce389e883cd753ca): **Rename the Standard dialect to the Func dialect**
* [Adaptation to LLVM's commit e7c7b16a849fb40169a708d27cb59747139ed8c7](https://github.com/llvm/torch-mlir/commit/ca96c31acc0d4639c367db028e78b3e19fccadde): **Remove the use of RegionBranchOpInterface::verifyTypes**
* [Adaptation to LLVM's commit 9eaff42360f4430e2baba28dd8d119137caae486](https://github.com/llvm/torch-mlir/commit/d441889ebe448e69aa42398c0eba6294c1d716f6): **Move Parser.h to Parser/**
[...]

PR #677, opened 2022-03-17
Bump LLVM at 8361c5da30588d3d4a48eae648f53be1feb5cfad

cathyzhyi:
thanks @Vigilans for the initial patches! 

Other than the changes listed in #673:
- the [linalg::GeneralizePadOpPattern](https://github.com/llvm/llvm-project/blob/8361c5da30588d3d4a48eae648f53be1feb5cfad/mlir/lib/Dialect/Linalg/Transforms/Transforms.cpp#L833) is added to refbackend as it's removed from linalg-bufferize.

- Added `NoSideEffect` traits to TorchConversion ops so that they can be removed if no user exists.  [passing log before bump](https://gist.github.com/cathyzhyi/6b6e70a1da192d5a1e8bce7877faac17), [failing log after bump](https://gist.github.com/b32923fa7252a20b96fe840f917895a2), [passing log with fix](https://gist.github.com/520d67af6b9773a419e31362b4065ff6)

Fixes: #673

### DIFF
--- a/examples/torchfx/builder.py
+++ b/examples/torchfx/builder.py
@@ -23,7 +23,7 @@ from itertools import chain
 
 from torch_mlir import ir
 import torch_mlir.dialects.torch as torch_d
-from torch_mlir.dialects import builtin, std
+from torch_mlir.dialects import builtin, std, func
 
 import torch.fx
 from torch.fx.experimental.fx_acc import acc_ops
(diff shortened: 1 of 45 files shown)


cathyzhyi:
> @sjarus Hey Suraj, there are TOSA failures after bumping llvm. Seems it's because `torch.constant.int` not converted correctly. Are you aware any changes from the TOSA side upstream that could cause this?

nvm, it's not specific to TOSA.

sjarus:
> > @sjarus Hey Suraj, there are TOSA failures after bumping llvm. Seems it's because `torch.constant.int` not converted correctly. Are you aware any changes from the TOSA side upstream that could cause this?
> 
> nvm, it's not specific to TOSA.

Sorry you caught me at lunchtime, but that's the kind of ticket we all like - the one that's resolved by the time we see it :)

silvasean (approved):

Issue #651, opened 2022-03-09
Fix RefineType for ops like aten.log


cathyzhyi:
The existing code assumes the result tensor type is the same as input type for  a few aten ops like `log`, `exp`, `erf`. See https://github.com/llvm/torch-mlir/blob/486f95e84f587d020ba789b071b12f890510f1a1/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L221-L235
This incorrect. The result  tensor of  these ops should always have the default dtype rather than the same as the input type. E2E tests for none default dtype input need to be added.

binfoo1993:
Hi @cathyzhyi,

I am Ph.D. student at George Washington University, who is interested in ML compiler.

I want to work on this issue.

cathyzhyi:
@binfoo1993 Hey Qiang, thanks for your interest in the project! You could refer to https://github.com/llvm/torch-mlir/wiki/Torch-ops-E2E-implementation for some basic information. Please let me know if you have further questions.

binfoo1993:
> @binfoo1993 Hey Qiang, thanks for your interest in the project! You could refer to https://github.com/llvm/torch-mlir/wiki/Torch-ops-E2E-implementation for some basic information. Please let me know if you have further questions.

Thanks for assigning this issue to me!

binfoo1993:
Hi @cathyzhyi,

I did some source code reading, trying to understand this right:

- For torch ops such as `exp`, `log`, `cos` etc, the output tensor's dtype could be different with input's. For example 
```python
import torch as th
a = th.IntTensor([1, 2, 3, 4])
print(th.exp(a).dtype) # torch.float32
```
- But existing codes assume they are of the same dtype.

So we should:

-  Find all ops satisfying this property, use another `if` branch to handle the type analysis.
- No matter what the input dtype is, always use `float32` for the result tensor. But the size should still be the same.
- Add e2e test for input tensor with dtypes not `float32`.

Besides, I have one concern:

- I notice our codes also handle `float64` dtype, for this input dtype, the output should also be of `float64`.
```python
import torch as th
a = th.DoubleTensor([1, 2, 3, 4])
print(th.exp(a).dtype) # torch.float64
```

Please let me know if I understand this wrong!

cathyzhyi:
[...]

PR #679, opened 2022-03-17
Fix refineTypes

binfoo1993:
Hi,

In this PR, I fix the RefineType issue in Issue #651.

For Ops:

- AtenTanhOp
- AtenExpOp 
- AtenSinOp
- AtenCosOp
- AtenSigmoidOp
- AtenReciprocalOp 
- AtenLogOp
- AtenSqrtOp
- AtenLog2Op 
- AtenRsqrtOp 
- AtenErfOp

The resulting dtype is always `float32`, except for `float64` input, the output is also `float64`.

The test input cases are also added.

@cathyzhyi

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -1009,20 +1009,35 @@ public:
 };
 } // namespace
 
-// Normalization formula:
-//   ((input - mean) / sqrt(var + eps)) * weight + bias
-static Value createLinalgPayloadCalculationForNormOps(
-    OpBuilder &b, Location loc, Type elemTy, Value input, Value mean, Value var,
-    Value eps, Value weight, Value bias) {
-  Value inputSubMean = b.create<arith::SubFOp>(loc, input, mean);
+/// Inverted STD: rSTD = 1 / sqrt(var + eps).
+static Value calculateRSTD(OpBuilder &b, Location loc, Type elemTy, Value eps,
+                           Value var) {
   // The eps is always f64.
   Value truncatedEps = b.create<arith::TruncFOp>(loc, elemTy, eps);
   Value varPlusEps = b.create<arith::AddFOp>(loc, var, truncatedEps);
   Value rSTD = b.create<math::RsqrtOp>(loc, varPlusEps);
+  return rSTD;
+}
+
+// Normalization formula:
+//   ((input - mean) * rSTD * weight + bias
+static Value createLinalgPayloadCalculationForNormOpsWithRSTD(
+    OpBuilder &b, Location loc, Type elemTy, Value input, Value mean,
+    Value rSTD, Value eps, Value weight, Value bias) {
+  Value inputSubMean = b.create<arith::SubFOp>(loc, input, mean);
   Value temp = b.create<arith::MulFOp>(loc, inputSubMean, rSTD);
   Value timesWeight = b.create<arith::MulFOp>(loc, temp, weight);
   Value plusBias = b.create<arith::AddFOp>(loc, timesWeight, bias);
   return plusBias;
+}
+
+static Value createLinalgPayloadCalculationForNormOpsWithVar(
+    OpBuilder &b, Location loc, Type elemTy, Value input, Value mean, Value var,
+    Value eps, Value weight, Value bias) {
+  Value rSTD = calculateRSTD(b, loc, elemTy, eps, var);
+  Value result = createLinalgPayloadCalculationForNormOpsWithRSTD(
+      b, loc, elemTy, input, mean, rSTD, eps, weight, bias);
+  return result;
 }
 
 namespace {
@@ -1117,9 +1132,10 @@ public:
                 [&](OpBuilder &b, Location loc, ValueRange args) {
                   Value input = args[0], weight = args[1], bias = args[2],
                         mean = args[3], var = args[4];
-                  Value result = createLinalgPayloadCalculationForNormOps(
-                      b, loc, var.getType(), input, mean, var, eps, weight,
-                      bias);
+                  Value result =
+                      createLinalgPayloadCalculationForNormOpsWithVar(
+                          b, loc, var.getType(), input, mean, var, eps, weight,
+                          bias);
                   b.create<linalg::YieldOp>(loc, result);
                 })
             .getResult(0);
@@ -1139,13 +1155,12 @@ public:
 // |  meanAndVarShape  |   normalizedShape  |
 // +-------------------+---------------------
 // <------------+ inputShape +-------------->
-
 // There are the following steps:
 // Step 1. Check if all the arguments meet the requirements.
 // Step 2. Common parts to be used for getting mean and var.
 //         This includes elements count, affineMap and iteratorTypes.
 // Step 3. Get mean.
-// Step 4. Get var.
+// Step 4. Get rSTD.
 // Step 5. Get layernorm.
 namespace {
 class ConvertAtenNativeLayerNormOp
@@ -1283,7 +1298,7 @@ public:
                     .getResult(0);
     Value mean = genMeanOrVarCalculation(sum);
 
-    // Step 4. Get var.
+    // Step 4. Get rSTD.
 
     // Calculate squareSum for the layer.
     SmallVector<AffineMap> squareSumIndexingMaps{
@@ -1310,6 +1325,21 @@ public:
                 })
             .getResult(0);
     Value var = genMeanOrVarCalculation(squareSum);
+    Value rSTDTensor = rewriter.create<linalg::InitTensorOp>(
+        loc, meanAndVarShapeSizes, elemTy);
+    SmallVector<AffineMap> rSTDIndexingMap(
+        2, rewriter.getMultiDimIdentityMap(meanAndVarShapeRank));
+
+    Value rSTD = rewriter
+                     .create<linalg::GenericOp>(
+                         loc, rSTDTensor.getType(), var, rSTDTensor,
+                         rSTDIndexingMap, meanAndVarIterationTypes,
+                         [&](OpBuilder &b, Location loc, ValueRange args) {
+                           Value result =
+                               calculateRSTD(b, loc, elemTy, eps, args[0]);
+                           b.create<linalg::YieldOp>(loc, result);
+                         })
+                     .getResult(0);
 
     // Step 5. Get layernorm.
 
@@ -1320,7 +1350,6 @@ public:
     auto normalizedShapeAffineMap = AffineMap::get(
         /*dimCount=*/inputRank,
         /*symbolCount=*/0, normalizedShapeExprs, context);
-
     auto inputSizes = getTensorSizes(rewriter, loc, input);
     Value initLayerNormTensor =
         rewriter.create<linalg::InitTensorOp>(loc, inputSizes, elemTy);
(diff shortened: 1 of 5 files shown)


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -524,6 +524,19 @@ ChangeResult TypeAnalyzer::visitOperation(
...
   }
 
+  // Dtype is always float32, except for float64.
+  if (isa<AtenTanhOp, AtenExpOp, AtenSinOp, AtenCosOp,
+          AtenSigmoidOp, AtenReciprocalOp, AtenLogOp, AtenSqrtOp,
+          AtenLog2Op, AtenRsqrtOp, AtenErfOp>(op)) {
+    ValueKnowledge knowledge =
+        ValueKnowledge::getNotNonePessimisticValueState(op->getContext());
+    knowledge.dtype = Float32Type::get(op->getContext());
+    if (operands[0]->getValue().dtype == Float64Type::get(op->getContext())) {
```
ramiro050, line 534: nit: you can do `operands[0]->getValue().dtype.isa<Float64Type>()`

ramiro050 (changes requested):
Thanks for working on this! I just have a couple of comments

lib/Dialect/Torch/Transforms/RefineTypes.cpp
binfoo1993: Got it. Thank you!

e2e_testing/torchscript/elementwise.py
```diff
@@ -25,7 +25,9 @@ def __init__(self):
     @export
     @annotate_args([
         None,
+        ([-1, -1], torch.int32, True),
         ([-1, -1], torch.float32, True),
+        ([-1, -1], torch.float64, True),
```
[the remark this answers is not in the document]
binfoo1993: Thank you! Will change accordingly.

cathyzhyi:
@binfoo1993 could you also change the commit title to be more specific like `Fix dtype inference for a few arithmetic ops in RefineTypes `

e2e_testing/torchscript/elementwise.py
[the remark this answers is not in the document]
ramiro050: Also, you can run these tests locally by doing:

```
python -m e2e_testing.torchscript.main --filter Elem -v
```

If you have not set your Python path, you need to first run

```
export PYTHONPATH=`pwd`/build/tools/torch-mlir/python_packages/torch_mlir:`pwd`/examples
```

binfoo1993:
@cathyzhyi @ramiro050 I just added test cases for these Ops with non-default dtype input (int64), run them locally and got this error:
```MLIR
// -----// IR Dump After VerifyInvariantsBeforeBackendLowering //----- //
module attributes {torch.debug_module_name = "ElementwiseLog2IntModule"} {
  func @forward(%arg0: !torch.vtensor<[?,?],si64>) -> !torch.vtensor<[?,?],f32> {
    %0 = torch.aten.log2 %arg0 : !torch.vtensor<[?,?],si64> -> !torch.vtensor<[?,?],f32>
  }
}


// -----// IR Dump After ConvertTorchToTMTensor //----- //
func @forward(%arg0: !torch.vtensor<[?,?],si64>) -> !torch.vtensor<[?,?],f32> {
  %0 = torch.aten.log2 %arg0 : !torch.vtensor<[?,?],si64> -> !torch.vtensor<[?,?],f32>
}

/home/qiang/workspace/torch-mlir/e2e_testing/torchscript/elementwise.py:807:15: error: 'math.log2' op operand #0 must be floating-point-like, but got 'i64'
        return torch.log2(a)
              ^
/home/qiang/workspace/torch-mlir/e2e_testing/torchscript/elementwise.py:807:15: note: see current operation: %10 = "math.log2"(%arg1) : (i64) -> i64
// -----// IR Dump After ConvertTorchToLinalg Failed //----- //
"builtin.func"() ({
^bb0(%arg0: !torch.vtensor<[?,?],si64>):
  %0 = "torch_c.to_builtin_tensor"(%arg0) : (!torch.vtensor<[?,?],si64>) -> tensor<?x?xi64>
  %1 = "arith.constant"() {value = 1 : index} : () -> index
  %2 = "arith.constant"() {value = 0 : index} : () -> index
  %3 = "tensor.dim"(%0, %2) : (tensor<?x?xi64>, index) -> index
  %4 = "arith.constant"() {value = 1 : index} : () -> index
  %5 = "tensor.dim"(%0, %4) : (tensor<?x?xi64>, index) -> index
  %6 = "linalg.init_tensor"(%3, %5) {static_sizes = [-1, -1]} : (index, index) -> tensor<?x?xf32>
  %7 = "linalg.generic"(%0, %6) ({
  ^bb0(%arg1: i64, %arg2: f32):
    %10 = "math.log2"(%arg1) : (i64) -> i64
    "linalg.yield"(%10) : (i64) -> ()
  }) {indexing_maps = [affine_map<(d0, d1) -> (d0, d1)>, affine_map<(d0, d1) -> (d0, d1)>], iterator_types = ["parallel", "parallel"], operand_segment_sizes = dense<1> : vector<2xi32>} : (tensor<?x?xi64>, tensor<?x?xf32>) -> tensor<?x?xf32>
  %8 = "tensor.cast"(%7) : (tensor<?x?xf32>) -> tensor<?x?xf32>
  %9 = "torch_c.from_builtin_tensor"(%8) : (tensor<?x?xf32>) -> !torch.vtensor<[?,?],f32>
  "std.return"(%9) : (!torch.vtensor<[?,?],f32>) -> ()
}) {sym_name = "forward", type = (!torch.vtensor<[?,?],si64>) -> !torch.vtensor<[?,?],f32>} : () -> ()
```

As we can see, although the resulting dtype of the torch Op is `f32`, the lowered IR doesn't convert the input type to `f32`, so the `math.log2` op operates on a `i64` input, which is not supported.
```
error: 'math.log2' op operand #0 must be floating-point-like, but got 'i64'
```
So I think we need to add type conversion op (`i64`->`f32`) before we use `linalg` ops on the input tensor.

I think I need to add some codes into the lowering rule of these torch Ops, can you tell me which files I need to look at?

cathyzhyi:
you can convert the element type to float type with `convertScalarToDtype` helper before passing to the the `math::log2` here https://github.com/llvm/torch-mlir/blob/218b4875d578419c1e3484fd77c47f5790bcf17d/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L239. You could refer to https://github.com/llvm/torch-mlir/blob/218b4875d578419c1e3484fd77c47f5790bcf17d/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L345 as an example of using this helper.

binfoo1993:
@cathyzhyi 

I added dtype conversion accordingly and passed the e2e test case for int32 input.

But I can't find the codes for `AtenSinOp` and `AtenCosOp`.

And there is no test cases for torch.sin, torch.cos, and torch.exp.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -192,22 +192,52 @@ static Value buildUnitNormalCdf(OpBuilder &b, Location &loc, Value x) {
...
-  if (isa<AtenLogOp>(op))
-    return b.create<math::LogOp>(loc, payloadArgs[0]);
-  if (isa<AtenErfOp>(op))
-    return b.create<math::ErfOp>(loc, payloadArgs[0]);
-  if (isa<AtenSqrtOp>(op))
-    return b.create<math::SqrtOp>(loc, payloadArgs[0]);
-  if (isa<AtenRsqrtOp>(op))
-    return b.create<math::RsqrtOp>(loc, payloadArgs[0]);
+  if (auto tanh = dyn_cast<AtenTanhOp>(op)) {
+    Type dtype = converter->convertType(tanh.getType())
```
cathyzhyi, line 200: the if body are identical for all the ops below. Can you use template function to reduce the duplicated code?
something like 
```
static template<typename TorchOpTy, typename MathOpTy>
createLinalgPayloadCalculationForElementwiseMathOp(OpBuilder &b, Location loc, TypeConverter *converter,
    ValueRange payloadArgs, Operation *op)
```

cathyzhyi:
ha, right, you need to add support in this file https://github.com/llvm/torch-mlir/blob/5b9bdfaf3fb56c5aed12d689e21c9884b1c5a789/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py#L315. Please refer to https://github.com/llvm/torch-mlir/blob/main/docs/adding_a_shape_function.md.

binfoo1993:
@cathyzhyi 
All test cases (with int32 input) for all the 11 Ops passed.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -189,25 +189,52 @@ static Value buildUnitNormalCdf(OpBuilder &b, Location &loc, Value x) {
...
 }
 
+template<typename TorchOpTy, typename MathOpTy>
+static Value createCalculationForMathOpWithDtypeConversion(
+              OpBuilder &b, Location loc, TypeConverter *converter,
+              ValueRange payloadArgs, Operation *op) {
+  auto opCasted = dyn_cast<TorchOpTy>(op);
+  Type dtype = converter->convertType(opCasted.getType())
+                    .template cast<RankedTensorType>()
+                    .getElementType();
```
ramiro050, line 199: You can do

https://github.com/llvm/torch-mlir/blob/ae4714aced1c9416f3854530333601d42c309140/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L915-L917

This allows you to get rid of the first template argument `TorchOpTy`

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -189,25 +189,52 @@ static Value buildUnitNormalCdf(OpBuilder &b, Location &loc, Value x) {
...
     return b.create<math::FloorOp>(loc, payloadArgs[0]);
   if (isa<AtenCeilOp>(op))
     return b.create<math::CeilOp>(loc, payloadArgs[0]);
-  if (isa<AtenLogOp>(op))
-    return b.create<math::LogOp>(loc, payloadArgs[0]);
-  if (isa<AtenErfOp>(op))
-    return b.create<math::ErfOp>(loc, payloadArgs[0]);
-  if (isa<AtenSqrtOp>(op))
-    return b.create<math::SqrtOp>(loc, payloadArgs[0]);
+  if (isa<AtenTanhOp>(op))
```
ramiro050, line 211: nit: Here we follow the style convention of adding curly braces to any `if` with a body more than one line long

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -235,14 +270,16 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
-    Type elementType = payloadArgs[0].getType();
+  if (auto sigmoid = dyn_cast<AtenSigmoidOp>(op)) {
+    Type dtype = converter->convertType(sigmoid.getType())
+                     .cast<RankedTensorType>()
+                     .getElementType(); 
+    Value arg = convertScalarToDtype(b, loc, payloadArgs[0], dtype);
+    Type elementType = arg.getType();
     auto one = b.create<arith::ConstantOp>(loc, FloatAttr::get(elementType, 1));
-    auto negate = b.create<arith::NegFOp>(loc, payloadArgs[0]);
+    auto negate = b.create<arith::NegFOp>(loc, arg);
```
ramiro050, line 282: nit: You can use the helper function you wrote for the `NegFOp` to avoid doing the `dtype` code above

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -189,25 +189,60 @@ static Value buildUnitNormalCdf(OpBuilder &b, Location &loc, Value x) {
   return buildNormalCdf(b, loc, x, zero, one);
 }
 
+template<typename MathOpTy>
+static Value createCalculationForMathOpWithDtypeConversion(
+              OpBuilder &b, Location loc, TypeConverter *converter,
+              ValueRange payloadArgs, Operation *op) {
```
ramiro050, line 195: I think it's a bit safer to only allow one `Value` to be passed rather than the `ValueRange payloadArgs`, since the function only uses the first `payloadArg`. Also, the `Location loc` can be accessed through the `Operation *op` by doing `Location loc = op->getLoc();`, so that argument is also not needed in this function.

ramiro050 (approved):
LGTM! I just have a couple of small comments. Also, once you're done addressing the comments, please squash your commits into a single commit and then run `git clang-format HEAD~1`. This will make some formatting changes that you can then commit again.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -524,6 +524,19 @@ ChangeResult TypeAnalyzer::visitOperation(
...
   }
 
+  // Dtype is always float32, except for float64.
+  if (isa<AtenTanhOp, AtenExpOp, AtenSinOp, AtenCosOp,
+          AtenSigmoidOp, AtenReciprocalOp, AtenLogOp, AtenSqrtOp,
+          AtenLog2Op, AtenRsqrtOp, AtenErfOp>(op)) {
+    ValueKnowledge knowledge =
+        ValueKnowledge::getNotNonePessimisticValueState(op->getContext());
+    knowledge.dtype = Float32Type::get(op->getContext());
+    if (operands[0]->getValue().dtype.isa<Float64Type>()) {
```
ramiro050, line 534: I think the tests are failing because of this line. I forgot that `operands[0]->getValue().dtype` can be a `nullptr`. You can either check that it is not null before doing `isa<...>()`, or do what you were doing before where you compare to a new `Float64Type`.

ramiro050, line 534: I think the tests are failing because of this line. I forgot that `operands[0]->getValue().dtype` can be a `nullptr`. You can either check that it is not null before doing `isa<...>()`, or do what you were doing before where you compare to a new `Float64Type`.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -189,25 +189,61 @@ static Value buildUnitNormalCdf(OpBuilder &b, Location &loc, Value x) {
...
-    return b.create<math::LogOp>(loc, payloadArgs[0]);
-  if (isa<AtenErfOp>(op))
-    return b.create<math::ErfOp>(loc, payloadArgs[0]);
-  if (isa<AtenSqrtOp>(op))
-    return b.create<math::SqrtOp>(loc, payloadArgs[0]);
-  if (isa<AtenRsqrtOp>(op))
-    return b.create<math::RsqrtOp>(loc, payloadArgs[0]);
+  if (isa<AtenTanhOp>(op)) {
+    return createCalculationForMathOpWithDtypeConversion<math::TanhOp>
+                                  (b, converter, payloadArgs[0], op);
```
ramiro050, line 213: Also, I don't think `git clang-format HEAD~1` formatted successfully. This indentation does not look right to me. (note, you do have to re-commit the modified changes made by `clang-format)

lib/Dialect/Torch/Transforms/RefineTypes.cpp
binfoo1993: No problem!

lib/Conversion/TorchToLinalg/Uncategorized.cpp
binfoo1993: Oh, sorry. I forgot. Will do that!

lib/Dialect/Torch/Transforms/RefineTypes.cpp
binfoo1993: I have one more question.
What should the output dtype be if the input dtype is `nullptr` for these ops?
If a `null` input dtype means the input could be any dtype, then I think the output dtype should also be `null` .

ramiro050: That's a really good point! Yeah, I agree that `dtype` should indeed be `null` if input is `null`. I think what you can do is to first check if the input `dtype` is null. If so, you can just return the `getNotNonePessimisticValueState` knowledge. Otherwise you can determine which float type to return.

PR #682, opened 2022-03-18
Reduce compilation time for TorchOps.cpp.inc

silvasean:
The `assemblyFormat` stuff (which generates unrolled, per-op C++ code)
was taking up a lot of compile time, and all the ops are essentially
printed with the same logic. So this PR makes them all call the same
helper function. This is done by using
`let hasCustomAssemblyFormat = 1` and then implementing `FooOp::parse`
and `FooOp::print`.

Additionally, the `Generated*Ops.td` files are all collapsed into just
`GeneratedTorchOps.td` (there is no reason to have the files separate,
since the files are very large anyway so one is always having to search
within them -- editors don't care that the file to search is now a bit
bigger :) ).

This reduces TorchOpsODSGenerated.cpp compile time (which is now
GeneratedTorchOps.cpp) from 39 to 31 seconds on my machine. This is
actually less than I expected, but this PR is an overall cleanup to the
code anyway. The next step will be to introduce (better) functionality
upstream for sharding the TorchOps.cpp.inc file, so that we can truly
parallelize the O(#ops) costs. This is also necessary, because after
this PR, TorchDialect.cpp is now the slowest file to compile, due to the
`addOperations<... all the ops ...>` call, which needs to be shareded
too.


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -98,73 +101,87 @@ def raw_emit_op(operator: JitOperator, f: TextIO, *, traits: List[str],
...
-    generic_result_name = lambda i: "result" + (str(i) if multiple_results else "")
-
-    p(f"def {td_def_name} : Torch_Op<{emitter.quote(op_name)}, [")
-    with emitter.indent():
-        with emitter.indent():
-            p(",\n".join(traits))
-        p("]> {")
-    with emitter.indent():
+
+    def generic_result_name(i): return "result" + \
```
ramiro050, line 110: nit: I think the formatting here would be better if the `return` starts in the next line

ramiro050 (approved):
LGTM! Just a small style comment

PR #687, opened 2022-03-23
Add non-default dtype support for a few elementwise math ops.

binfoo1993:
This PR adds non-default dtype support for elementwise math ops:

- AtenTanhOp
- AtenExpOp 
- AtenSinOp
- AtenCosOp
- AtenSigmoidOp
- AtenReciprocalOp 
- AtenLogOp
- AtenSqrtOp
- AtenLog2Op 
- AtenRsqrtOp 
- AtenErfOp

Changes:

- Fix type inference: set the output dtype to `float32`, except for `float64` and `null` input. 
- Fix Torch2Linalg conversion: add dtype conversion over input value before math op.
- Add test cases with `int32` input.

@ramiro050 @cathyzhyi

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -189,25 +189,60 @@ static Value buildUnitNormalCdf(OpBuilder &b, Location &loc, Value x) {
   return buildNormalCdf(b, loc, x, zero, one);
 }
 
+template <typename MathOpTy>
+static Value createCalculationForMathOpWithDtypeConversion(
+    OpBuilder &b, TypeConverter *converter, Value payloadArg, Operation *op) {
+  Type dtype = converter->convertType(op->getResult(0).getType())
+                   .template cast<RankedTensorType>()
+                   .getElementType();
+  Location loc = op->getLoc();
+  Value arg = convertScalarToDtype(b, loc, payloadArg, dtype);
+  return b.create<MathOpTy>(loc, arg);
+}
+
 static Value createLinalgPayloadCalculationForElementwiseOp(
     OpBuilder &b, Location loc, TypeConverter *converter,
     ValueRange payloadArgs, Operation *op, ArrayRef<Value> operands) {
-  if (isa<AtenTanhOp>(op))
-    return b.create<math::TanhOp>(loc, payloadArgs[0]);
-  if (isa<AtenExpOp>(op))
-    return b.create<math::ExpOp>(loc, payloadArgs[0]);
   if (isa<AtenFloorOp>(op))
     return b.create<math::FloorOp>(loc, payloadArgs[0]);
   if (isa<AtenCeilOp>(op))
     return b.create<math::CeilOp>(loc, payloadArgs[0]);
-  if (isa<AtenLogOp>(op))
-    return b.create<math::LogOp>(loc, payloadArgs[0]);
-  if (isa<AtenErfOp>(op))
-    return b.create<math::ErfOp>(loc, payloadArgs[0]);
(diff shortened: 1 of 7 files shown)


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -770,19 +804,22 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
       reciprocal.emitError("unimplemented: non-floating point dtype");
       return nullptr;
     }
-
-    Type elementType = payloadArgs[0].getType();
```
cathyzhyi, line 774: the code above requiring input to be float can be deleted. The e2e test added for `reciprocal` should have failed because of this code. Could you double check if it's the case?

cathyzhyi (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
ramiro050: I think the check above is for the return type, which will be a float because of the `RefineTypes` code, so the test should be passing

binfoo1993: Actually, the e2e test for `reciprocal` works for `int32` input. Because op's `getType()` returns the type of the first result of the Operation:
```C++
    ResultType getType() {
      auto resultTy = this->getOperation()->getResult(0).getType();
      return resultTy.template cast<ResultType>();
    }
```
And the test passed locally:
```
PASS - "ElementwiseReciprocalIntModule_basic"
```

Since we set the resulting dtype to float anyway, so I think the code should work for any input dtype which is not `null`. But I do worry about the case of 'null' input dtype.

cathyzhyi: ha, you are right! The code just need to be deleted then.

lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
```diff
@@ -39,11 +39,11 @@ module {
   func @__torch__.torch_mlir.dialects.torch.importer.jit_ir.build_tools.upstream_shape_helpers._copy(%arg0: !torch.list<int>) -> !torch.list<int> {
     %true = torch.constant.bool true
     %0 = torch.prim.ListConstruct  : () -> !torch.list<int>
-    %1 = torch.aten.len.t %arg0 : !torch.list<int> -> !torch.int
+    %1 = torch.aten.len.t %arg0 :  !torch.list<int> ->  !torch.int
     torch.prim.Loop %1, %true, init() {
     ^bb0(%arg1: !torch.int):
-      %2 = torch.aten.__getitem__.t %arg0, %arg1 : !torch.list<int>, !torch.int -> !torch.int
-      %3 = torch.aten.append.t %0, %2 : !torch.list<int>, !torch.int -> !torch.list<int>
+      %2 = torch.aten.__getitem__.t %arg0, %arg1 :  !torch.list<int>, !torch.int ->  !torch.int
```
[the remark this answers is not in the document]
binfoo1993: Yeah, I don't know why.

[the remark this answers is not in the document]
binfoo1993: I can ignore this file in the commit. I think it should have something to do with my local environment.

ramiro050:
Can you add `ElementwiseExpModule_basic` here:

https://github.com/llvm/torch-mlir/blob/fe8ac57e6d15b0cbfd3a4e3f675d4182c1e44b3c/e2e_testing/torchscript/xfail_sets.py#L33

ramiro050:
Everything looks good! I will go ahead and merge this. Thanks for working on this!

PR #689, opened 2022-03-23
Got LTC working until compile

antoniojkim:
The changes in this PR get the Torch-MLIR LTC Backend to a point where an example run can successfully run up until the actual compile. This means that the graph is properly traced with the Torch-MLIR LTC backend and is ready to be compiled.

Next steps is to implementing the lowering itself down to MLIR from the LTC Node IR representation. The idea is to make use of a lot of the existing TorchScript IR -> MLIR infrastructure that already exists to make it easier to integrate.

CC: @ke1337 @henrytwo


silvasean:
Looks good to me. I assume that @wconstab has taken a look at this? Adding them as a reviewer just in case.

silvasean (approved):

python/torch_mlir/csrc/backend/LazyShapeInference.cpp
```diff
@@ -0,0 +1,21 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#include "LazyShapeInference.h"
+#include "../utils/exception.h"
+
+namespace torch {
+namespace lazy {
+
+std::vector<Shape> compute_shape_detach(const at::Tensor& self) {
```
wconstab, line 16: I'd generally encourage you to upstream these to torch to be shared by all backends; I'm not sure if we want to handle the detach op in other backends though

PR #693, opened 2022-03-24
Add 2D case for at::convolution

gpetters94:
Here's the implementation of `at::convolution` and a decomposition of `at::convolution_overrideable`. The 2D case is all that's implemented for `at::convolution`, since the 1D and 3D cases require changes to linalg. I don't think we need them right now, looking at our op list, but I need to look further into the list to be sure. `convolution_overrideable` just decomposes directly into `convolution`.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -449,58 +449,65 @@ public:
 } // namespace
 
 namespace {
-class ConvertAtenConv2dOp : public OpConversionPattern<AtenConv2dOp> {
+class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
 public:
   using OpConversionPattern::OpConversionPattern;
   LogicalResult
-  matchAndRewrite(AtenConv2dOp op, OpAdaptor adaptor,
+  matchAndRewrite(AtenConvolutionOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     Location loc = op->getLoc();
     MLIRContext *context = op->getContext();
     Value input = adaptor.input();   /* in form of N*C*H*W */
     Value weight = adaptor.weight(); /* in form of F*C*H*W */
-    Value groups = adaptor.groups();
 
     Type elementType =
         input.getType().cast<RankedTensorType>().getElementType();
     if (!elementType.isa<mlir::FloatType>())
       return op.emitError("unimplemented: non-floating point type");
+    size_t inRank = input.getType().cast<RankedTensorType>().getRank();
+    if (inRank != 4)
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only 2D convolution currently supported");
 
     Type intType = IntegerType::get(context, 64);
     auto castIndexToInt = [&](Value v) {
       return rewriter.create<arith::IndexCastOp>(loc, intType, v);
     };
 
-    Value N = getDimOp(rewriter, loc, input, 0);
-    Value Hin = getDimOp(rewriter, loc, input, 2);
-    Value Win = getDimOp(rewriter, loc, input, 3);
-    Value F = getDimOp(rewriter, loc, weight, 0);
-    Value weightH = getDimOp(rewriter, loc, weight, 2);
-    Value weightW = getDimOp(rewriter, loc, weight, 3);
-
-    // Pattern match against the op's original operands, because otherwise we
-    // will get the lowered version of the operands which is harder to pattern
-    // match.
     SmallVector<int64_t> paddingInts;
     if (!matchPattern(op.padding(), m_TorchConstantIntList(paddingInts))) {
       return rewriter.notifyMatchFailure(
           op, "only support constant padding values");
     }
-
-    SmallVector<int64_t, 2> strideInts;
+    SmallVector<int64_t> strideInts;
     if (!matchPattern(op.stride(), m_TorchConstantIntList(strideInts)))
       return rewriter.notifyMatchFailure(op,
                                          "only support constant int strides");
-    SmallVector<int64_t, 2> dilationInts;
+    SmallVector<int64_t> dilationInts;
     if (!matchPattern(op.dilation(), m_TorchConstantIntList(dilationInts)))
       return rewriter.notifyMatchFailure(op,
                                          "only support constant int dilations");
-    Value c1 =
-        rewriter.create<arith::ConstantOp>(loc, IntegerAttr::get(intType, 1));
-    Value groupEqual1 = rewriter.create<arith::CmpIOp>(
-        loc, arith::CmpIPredicate::eq, groups, c1);
-    rewriter.create<cf::AssertOp>(
-        loc, groupEqual1, rewriter.getStringAttr("expect groups to be 1"));
+
+    Value N = getDimOp(rewriter, loc, input, 0);
+    SmallVector<Value> inDims;
+    for (size_t i = 2; i < inRank; i++)
+      inDims.push_back(getDimOp(rewriter, loc, input, i));
+    Value F = getDimOp(rewriter, loc, weight, 0);
+    SmallVector<Value> weightDims;
+    for (size_t i = 2; i < inRank; i++)
+      weightDims.push_back(getDimOp(rewriter, loc, weight, i));
+
+    // Guard unused values (transposed, groups)
+    int64_t group_size;
+    if (!matchPattern(op.groups(), m_TorchConstantInt(&group_size)) ||
+        group_size != 1)
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only group size of 1 supported");
+    bool transposed = true;
+    if (!matchPattern(op.transposed(), m_TorchConstantBool(&transposed)) ||
+        transposed)
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only non-transposed convolution supported");
 
     // Pad the input tensor according to padding.
     SmallVector<int64_t, 4> paddingIncludingNC = {0, 0};
@@ -516,15 +523,14 @@ public:
     SmallVector<Value> strideIntValues =
         getAsConstantIntValues(rewriter, loc, strideInts);
 
-    Value Hout = torch_to_linalg::getOutputDimForConvOps(
-        rewriter, loc, Hin, paddingIntValues[0], dilationIntValues[0],
-        castIndexToInt(weightH), strideIntValues[0]);
-    Value Wout = torch_to_linalg::getOutputDimForConvOps(
-        rewriter, loc, Win, paddingIntValues[1], dilationIntValues[1],
-        castIndexToInt(weightW), strideIntValues[1]);
+    SmallVector<Value> outDims{N, F};
+    for (size_t i = 0; i < inRank - 2; i++)
+      outDims.push_back(torch_to_linalg::getOutputDimForConvOps(
+          rewriter, loc, inDims[i], paddingIntValues[i], dilationIntValues[i],
+          castIndexToInt(weightDims[i]), strideIntValues[i]));
 
-    Value initTensor = rewriter.create<linalg::InitTensorOp>(
-        loc, ValueRange{N, F, Hout, Wout}, elementType);
+    Value initTensor =
+        rewriter.create<linalg::InitTensorOp>(loc, outDims, elementType);
 
     Value bias = adaptor.bias();
     Value biasInitTensor;
@@ -559,14 +565,17 @@ public:
 
     auto stridesAttr = rewriter.getI64VectorAttr(strideInts);
     auto dilationAttr = rewriter.getI64VectorAttr(dilationInts);
-    Value conv2d =
+
+    // TODO: add 1D and 3D case
+    Value conv =
         rewriter
             .create<linalg::Conv2DNchwFchwOp>(
                 loc, biasInitTensor.getType(), ValueRange{paddedInput, weight},
                 biasInitTensor, stridesAttr, dilationAttr)
             .getResult(0);
+
     Type newResultType = getTypeConverter()->convertType(op.getType());
-    rewriter.replaceOpWithNewOp<tensor::CastOp>(op, newResultType, conv2d);
+    rewriter.replaceOpWithNewOp<tensor::CastOp>(op, newResultType, conv);
     return success();
   }
 };
@@ -584,6 +593,6 @@ void mlir::torch::torch_to_linalg::populateLinearPatternsAndLegality(
   patterns.add<ConvertAtenBmmOp>(typeConverter, context);
   target.addIllegalOp<AtenLinearOp>();
   patterns.add<ConvertAtenLinearOp>(typeConverter, context);
-  target.addIllegalOp<AtenConv2dOp>();
-  patterns.add<ConvertAtenConv2dOp>(typeConverter, context);
+  target.addIllegalOp<AtenConvolutionOp>();
+  patterns.add<ConvertAtenConvolutionOp>(typeConverter, context);
 }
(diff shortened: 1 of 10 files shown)


cathyzhyi:
The code seems to be very similar to conv2d. Can we just decompose `aten::convolution ` to `aten::conv2d` when it's 2D?

gpetters94:
We could do that, yeah. Though if we do end up adding 1D and 3D cases it might make more sense to decompose conv1d/conv2d/conv3d to `aten::convolution`.

cathyzhyi:
> We could do that, yeah. Though if we do end up adding 1D and 3D cases it might make more sense to decompose conv1d/conv2d/conv3d to `aten::convolution`.

yea, that works too. Can you decompose conv2d to `aten::convolution` to reduce duplication?

e2e_testing/torchscript/conv.py
```diff
@@ -133,3 +133,59 @@ def forward(self, x):
...
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1, -1], torch.float32, True),
+        ([-1, -1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, inputVec, weight):
+        return torch.ops.aten.convolution(inputVec, weight, bias=None, stride=[1, 1], padding=[0, 0], dilation=[1, 1], transposed=False, output_padding=[0, 0], groups=1)
```
cathyzhyi, line 168: This line is too long. can you change the argument format to be the same as the other tests?

e2e_testing/torchscript/conv.py
```diff
@@ -133,3 +133,59 @@ def forward(self, x):
...
+
+@register_test_case(module_factory=lambda: ConvolutionModule2D())
+def ConvolutionModule2D_basic(module, tu: TestUtils):
+    with torch.backends.mkldnn.flags(enabled=False):
```
cathyzhyi, line 172: why is this line needed?

e2e_testing/torchscript/conv.py
```diff
@@ -133,3 +133,59 @@ def forward(self, x):
...
+        module.forward(torch.ones(3, 3, 10, 10), torch.ones(3, 3, 2, 2))
```
cathyzhyi, line 173: can we use more random data rather than all ones?

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -465,36 +465,40 @@ class ConvertAtenConv2dOp : public OpConversionPattern<AtenConv2dOp> {
         input.getType().cast<RankedTensorType>().getElementType();
     if (!elementType.isa<mlir::FloatType>())
       return op.emitError("unimplemented: non-floating point type");
+    SmallVector<Value> inSizes = getTensorSizes(rewriter, loc, input);
```
cathyzhyi, line 468: We can get the rank directly from input ranked tensor type.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -465,36 +465,40 @@ class ConvertAtenConv2dOp : public OpConversionPattern<AtenConv2dOp> {
...
+    size_t inRank = inSizes.size();
+    if (inRank != 4)
+      return rewriter.notifyMatchFailure(
+          op, "only 2D convolution currently supported");
```
cathyzhyi, line 472: nit: add `unimplemented:` to be consistent with other error msg

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -559,14 +562,32 @@ class ConvertAtenConv2dOp : public OpConversionPattern<AtenConv2dOp> {
 
     auto stridesAttr = rewriter.getI64VectorAttr(strideInts);
     auto dilationAttr = rewriter.getI64VectorAttr(dilationInts);
-    Value conv2d =
+    Value conv;
```
cathyzhyi, line 565: can you remove the commented out code and add a // TODO:  add support for 2d and 2d case

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -27,6 +27,7 @@
     "int": "Torch_IntType",
     "int[]": "TorchIntListType",
     "int?": "TorchOptionalIntType",
+    "int[]?": "TorchOptionalIntListType",
```
cathyzhyi, line 30: Can you fix this in the other PR I mentioned before? https://github.com/llvm/torch-mlir/pull/643

cathyzhyi (changes requested):

cathyzhyi:
Can you help to fix TOSA as well here? https://github.com/llvm/torch-mlir/blob/5b9bdfaf3fb56c5aed12d689e21c9884b1c5a789/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L1681. Just need to change the conversion pattern to apply for `aten::convolution` and guard `inputRank == 4` as we do in Torch to Linalg conversion.

lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
```diff
@@ -39,11 +39,11 @@ module {
   func @__torch__.torch_mlir.dialects.torch.importer.jit_ir.build_tools.upstream_shape_helpers._copy(%arg0: !torch.list<int>) -> !torch.list<int> {
     %true = torch.constant.bool true
     %0 = torch.prim.ListConstruct  : () -> !torch.list<int>
-    %1 = torch.aten.len.t %arg0 : !torch.list<int> -> !torch.int
+    %1 = torch.aten.len.t %arg0 :  !torch.list<int> ->  !torch.int
```
cathyzhyi, line 42: Can you rebase to the latest and rerun `update_shape_lib.sh`. A recent fix from https://github.com/llvm/torch-mlir/pull/687 should avoid adding the extra spaces.

e2e_testing/torchscript/conv.py
gpetters94: We've had some issues with running certain e2e tests because the mkldnn backend doesn't support some ops. I think convolution was one of those, but it may have been convolution_backward.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
gpetters94: Do you mind if I change it in this PR?

gpetters94:
Re:TOSA, I've added the requested changes and it is still failing. TOSA is out of my wheelhouse, so that's the best I can do.

cathyzhyi:
> Re:TOSA, I've added the requested changes and it is still failing. TOSA is out of my wheelhouse, so that's the best I can do.

The failure is at https://github.com/llvm/torch-mlir/blob/5b9bdfaf3fb56c5aed12d689e21c9884b1c5a789/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L1715. This is because TOSA doesn't handle static weight. Can you add a test with static weight?

sjarus:
I think you mean doesn't handle dynamic weight, @cathyzhyi  ? Yes the way around is to ensure the weight is static . Please see examples of static test versions that work for TOSA, e.f. vision_models.py has a ResNet18Static that works , while the original dynamic one doesn't yet.

cathyzhyi:
> I think you mean doesn't handle dynamic weight, @cathyzhyi ? Yes the way around is to ensure the weight is static . Please see examples of static test versions that work for TOSA, e.f. vision_models.py has a ResNet18Static that works , while the original dynamic one doesn't yet.

yea.. sry, I meant dynamic weight. Thanks for confirming!

gpetters94:
I think the TOSA fix is working, but I'm not sure how to mark the test as an expected pass for the TOSA pass.

cathyzhyi:
you can the test to this list https://github.com/llvm/torch-mlir/blob/1b505cbac506d80a466f88afc4d0c6a6aa3d347c/e2e_testing/torchscript/xfail_sets.py#L25

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -1701,6 +1701,12 @@ LogicalResult ConvertAtenOp<AtenConvolutionOp>::matchAndRewrite(
 
   if (inputTy.getRank() != 4)
     return op.emitError("Unimplemented: only 2D convolutions supported");
+  
+  bool staticWeight = true;
```
cathyzhyi, line 1705: nit: you can use `hasStaticShape` on RankedTensorType to check for static shape

python/torch_mlir_e2e_test/test_suite/conv.py
```diff
@@ -213,3 +213,29 @@ def ConvolutionModule2D_basic(module, tu: TestUtils):
...
+                                          padding=[0, 0],
+                                          dilation=[1, 1],
+                                          transposed=False,
+                                          output_padding=[0, 0],
+                                          groups=1)
+
+@register_test_case(module_factory=lambda: ConvolutionModule2DStatic())
+def ConvolutionModule2DStatic_basic(module, tu: TestUtils):
+    with torch.backends.mkldnn.flags(enabled=False):
+        module.forward(torch.randn(3, 3, 10, 10), torch.randn(3, 3, 2, 2))
```
cathyzhyi, line 241: nit: please add an extra empty line at the end of the file.

cathyzhyi (changes requested):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -1692,13 +1692,19 @@ LogicalResult ConvertAtenOp<AtenConv2dOp>::matchAndRewrite(
...
   auto inputElemTy = inputTy.getElementType();
   auto weightElemTy = weightTy.getElementType();
   auto inputShape = inputTy.getShape();
   auto weightShape = weightTy.getShape();
 
+  if (inputTy.getRank() != 4)
+    return op.emitError("Unimplemented: only 2D convolutions supported");
+
+  if (!weightTy.hasStaticShape())
+    return op->emitError("Unimplemented: TOSA only supports static weight");
```
cathyzhyi, line 1706: nit: use `op.` for consistency.

python/torch_mlir_e2e_test/test_suite/conv.py
```diff
@@ -133,3 +133,109 @@ def forward(self, x):
 def Conv2dWithPaddingDilationStrideStaticModule_basic(module, tu: TestUtils):
     t = tu.rand(5, 2, 10, 20)
     module.forward(t)
+
+# ==============================================================================
+
+# class ConvolutionModule1D(torch.nn.Module):
```
cathyzhyi, line 139: those tests could be uncommented and added into xfail_set.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -750,6 +750,47 @@ class DecomposeAtenWhereScalarSelfOp
...
+class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenConv2dOp op,
+                                PatternRewriter &rewriter) const override {
+
+    Value emptyList = rewriter.create<PrimListConstructOp>(
+        op.getLoc(), Torch::ListType::get(Torch::IntType::get(op.getContext())),
+        SmallVector<Value>());
+    Value False = rewriter.create<Torch::ConstantBoolOp>(op.getLoc(), false);
```
cathyzhyi, line 783: nit: use `cstFalse` to follow the naming convention.

python/torch_mlir_e2e_test/test_suite/conv.py
```diff
@@ -133,3 +133,109 @@ def forward(self, x):
...
+    @annotate_args([
+        None,
+        ([-1, -1, -1, -1], torch.float32, True),
+        ([-1, -1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, inputVec, weight):
+        return torch.ops.aten.convolution(inputVec,
+                                          weight,
+                                          bias=None,
+                                          stride=[1, 1],
```
cathyzhyi, line 179: shall we change the strides to values other than 1 and padding to non zero?

cathyzhyi (changes requested):
`'aten::convolution` has more arguments than `aten::conv2d` like `transposed` and `output_padding`. If we don't support those arguments we need to guard against the values that are not supported.

gpetters94:
I've added a guard for transposed. As I understand it output_padding is just a compiler hint, so we shouldn't need to worry about it.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -465,42 +465,53 @@ class ConvertAtenConv2dOp : public OpConversionPattern<AtenConv2dOp> {
...
+    Value N = getDimOp(rewriter, loc, input, 0);
+    SmallVector<Value> inDims;
+    for (size_t i = 2; i < inRank; i++)
+      inDims.push_back(getDimOp(rewriter, loc, input, i));
+    Value F = getDimOp(rewriter, loc, weight, 0);
+    SmallVector<Value> weightDims;
+    for (size_t i = 2; i < inRank; i++)
+      weightDims.push_back(getDimOp(rewriter, loc, weight, i));
+
+    // Guard unused values (transposed, groups)
```
cathyzhyi, line 501: Can you change both guards to be consistent with other guards in the codebase using `!matchPattern`?

cathyzhyi (changes requested):

gpetters94:
Okay, the guards are both `matchPattern`s now.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -465,42 +465,48 @@ class ConvertAtenConv2dOp : public OpConversionPattern<AtenConv2dOp> {
...
+    int64_t group_size;
+    if(!matchPattern(op.groups(), m_TorchConstantInt(&group_size)) || group_size != 1)
+      return rewriter.notifyMatchFailure(op,
+                                         "unimplemented: only group size of 1 supported");
+    bool transposed;
+    if(!matchPattern(op.transposed(), m_TorchConstantBool(&transposed)))
```
cathyzhyi, line 507: also need to check `transposed` is false

python/torch_mlir_e2e_test/test_suite/conv.py
cathyzhyi: this is not addressed yet.

cathyzhyi (changes requested):

lib/Conversion/TorchToLinalg/Linear.cpp
gpetters94: Good point, I've added that check.

cathyzhyi: `bool transposed`  need to be initialized with `true`. Otherwise, when `op.transposed` is variable and `transposed` is initialized with `false`, the guard won't catch the case correctly.

e2e_testing/torchscript/conv.py
cathyzhyi: I missed this the last time. This code shouldn't be part of our e2e test right? because we don't use mkldnn backend in our e2e testing.

cathyzhyi (changes requested):

python/torch_mlir_e2e_test/test_suite/conv.py
```diff
@@ -133,3 +133,135 @@ def forward(self, x):
...
+                                          stride=[3, 3],
+                                          padding=[2, 2],
+                                          dilation=[1, 1],
+                                          transposed=False,
+                                          output_padding=[0, 0],
+                                          groups=1)
+
+@register_test_case(module_factory=lambda: ConvolutionModule2DStrided())
+def ConvolutionModule2DStrided_basic(module, tu: TestUtils):
+    with torch.backends.mkldnn.flags(enabled=False):
```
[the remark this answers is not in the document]
gpetters94: Yep, it was convolution_backward. I've updated accordingly.

cathyzhyi (approved):

PR #695, opened 2022-03-24
[MLIR][TORCH] Add support for same input and output shapes for view op

vivekkhandelwal1:
This commit adds support for the cases of view op where the rank and
the shapes of the input and result are equal.

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -107,19 +107,6 @@ public:
     auto resultType =
         typeConverter->convertType(op.getType()).cast<RankedTensorType>();
     int64_t resultRank = resultType.getRank();
-    // Currently, we only handle the expanding OR collapsing cases, we do not
-    // handle expanding And collapsing happening at the same time or cases where
-    // it's neither collapsing nor expanding like view of [2,3] for 3x2 tensor.
-    // TODO: For the expanding And collapsing case, we will need to identify
-    // which dimensions are collapsing and which are expanding and do it in two
-    // steps.
-    // TODO: For neither collapsing nor expanding, we could find a intermediate
-    // shape to collapse and then expanded to the target shape. Like [2,3] =>
-    // [6] => [3, 2].
-    if (inputRank == resultRank)
-      return rewriter.notifyMatchFailure(
-          op, "unimplemented: the view op is neither expanding nor collapsing");
-
     if (resultRank == 0)
       return rewriter.notifyMatchFailure(op,
                                          "result shape of rank 0 is invalid");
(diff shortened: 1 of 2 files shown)


cathyzhyi (changes requested):

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -147,212 +134,236 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
-              op, "at most one element in size list is allowed to be -1");
-        }
-        inferredDimension = outputDim;
+    Value result;
+    // For the case when inputRank == resultRank, we only have support for the
+    // cases where not only the ranks of the input and result are equal but also
+    // the shapes. If the shape of the input tensor is static, then this case is
+    // automatically handled by PyTorch, but we need this code when the shape of
+    // the input is dynamic. The result here is the input with the shape
+    // transformed to static from dynamic.
```
ramiro050, line 145: nit: The result is not being transformed from dynamic to static. All that is happening is that checks are made during runtime.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -147,212 +134,236 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
-        if (inferredDimension.hasValue()) {
-          return rewriter.notifyMatchFailure(
-              op, "at most one element in size list is allowed to be -1");
-        }
-        inferredDimension = outputDim;
+    Value result;
+    // For the case when inputRank == resultRank, we only have support for the
+    // cases where not only the ranks of the input and result are equal but also
+    // the shapes. If the shape of the input tensor is static, then this case is
+    // automatically handled by PyTorch, but we need this code when the shape of
```
Shukla-Gaurav, line 143: `automatically handled by PyTorch?` -> We handle such cases as a part of canonicalization and `aten.view` op folds only in the case of unity tensors.

Shukla-Gaurav (changes requested):

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -147,214 +134,229 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+    // cases where not only the ranks of the input and result are equal but also
+    // the shapes. We need this code when the shape of the input tensor is
+    // dynamic. The result in this case is the input tensor itself.
+    if (inputRank == resultRank) {
+      for (unsigned i = 0; i < inputRank; i++)
+        checkDimEqualHelper(rewriter, loc, inputSize[i], outputSizeInt[i]);
+      rewriter.replaceOpWithNewOp<tensor::CastOp>(op, resultType, input);
+      return success();
+    } else {
+      ArrayRef<Value> expandedShapeInt =
```
Shukla-Gaurav, line 149: I guess there is no need of `else` statement here. This will help in reducing unnecessary indentation.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -107,19 +107,6 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
     int64_t resultRank = resultType.getRank();
-    // Currently, we only handle the expanding OR collapsing cases, we do not
-    // handle expanding And collapsing happening at the same time or cases where
-    // it's neither collapsing nor expanding like view of [2,3] for 3x2 tensor.
-    // TODO: For the expanding And collapsing case, we will need to identify
-    // which dimensions are collapsing and which are expanding and do it in two
-    // steps.
-    // TODO: For neither collapsing nor expanding, we could find a intermediate
-    // shape to collapse and then expanded to the target shape. Like [2,3] =>
-    // [6] => [3, 2].
```
Shukla-Gaurav, line 118: This comment still stands true. Please do not remove it.

Shukla-Gaurav (changes requested):

lib/Conversion/TorchToLinalg/DataMovement.cpp
vivekkhandelwal1:
```
we do not handle expanding And collapsing happening at the same time or cases where it's neither collapsing nor expanding 
```
The original comment might be partially true. Because now we support cases where it's neither collapsing nor expanding. Or do you not count the `no change` cases in this category?

Shukla-Gaurav: Your change only handles `identity` cases but not the cases like [2,3] for 3x2 tensor(neither collapsing nor expanding) as mentioned in the comment. You can add few extra lines in the same comment mentioning that identity cases are covered.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -207,17 +213,17 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+            op, "unimplemented: an inferred dimension is only supported when "
+                "there is enough static shape information to determine its "
+                "size, or when the input tensor is being flattened to a "
+                "single dimension");
       }
 
       auto productReduceKnownSizes = [](const ArrayRef<int64_t> sizes) {
         auto knownSizes = llvm::make_filter_range(
             sizes, [](int64_t val) { return val != kUnknownSize; });
-        return std::accumulate(knownSizes.begin(), knownSizes.end(), /*init=*/1,
```
cathyzhyi, line 219: The formatting with wrapping line changes are mostly not needed. Can you revert those back so that actual changes need are more obvious.

cathyzhyi (changes requested):

Shukla-Gaurav (approved):
LGTM! Thanks!

PR #700, opened 2022-03-25
[tosa] Support for Aten[Unsqueeze|Contiguous|Dropout|Reshape|View] ops

anupgangwar:

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -2329,6 +2329,116 @@ LogicalResult ConvertAtenOp<AtenThresholdOp>::matchAndRewrite(
   return success();
 }
 
+template <>
+LogicalResult ConvertAtenOp<AtenUnsqueezeOp>::matchAndRewrite(
+    AtenUnsqueezeOp op, OpAdaptor adaptor,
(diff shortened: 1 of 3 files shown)


anupgangwar:
Tagging @cathyzhyi for review.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -2329,6 +2329,118 @@ LogicalResult ConvertAtenOp<AtenThresholdOp>::matchAndRewrite(
...
+    ConversionPatternRewriter &rewriter) const {
+
+  // Not a tensor type.
+  auto selfType = adaptor.input().getType().dyn_cast<TensorType>();
+  if (!selfType)
+    return op.emitError("Only tensor types are currently supported");
+
+  // FIXME: train and p are not handled.
+
+  if (!op.train().getType().isa<Torch::NoneType>()) {
```
cathyzhyi, line 2402: the `if` is not needed because `train` can only be bool type.

cathyzhyi (approved):
looks mostly good except one if.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
anupgangwar: Incorporated in the latest push.

PR #701, opened 2022-03-25
Move e2e test definitions into the `torch_mlir_e2e_test` package

silvasean:
This is the first step to making the e2e framework convenient to use
by downstream backends.


cathyzhyi (approved):

silvasean:
A note for folks that run into this PR: when rebasing past this PR, you will need to change your e2e tests that are added in e2e_testing/torchscript/* to python/torch_mlir_e2e_test/test_suite/*.

Issue #660, opened 2022-03-11
bazel build support


garymm:
It'd be nice to be able to depend on this code in a project that uses Bazel.
[LLVM seems to have some support for bazel](https://github.com/llvm/llvm-project/tree/main/utils/bazel), so maybe this wouldn't be too hard.

stellaraccident:
I don't think any of the current contributors are interested in supporting such an effort -- which tends to be expensive over time. But the precedent in llvm may apply here: allowing for an unsupported place to put contributed build files.

What do you mean by "depend on"? Do you just need a c++ dependency on the dialects? Trying to get the whole python plus pytorch dependency? (I can't see how this is useful unless if you area shop that must have all software hermetically built by bazel exclusively: just build the wheel and install it like any other python dep)

garymm:
I'm still exploring but I at least want a C++ dependency on the dialects. Is there some way to get that in my bazel project?

stellaraccident:
People who need this would need to implement it. It seems unlikely that it is something that contributors would do without having a need for it themselves.

I suspect there is a path for landing it on an unsupported way, given the precedent in llvm.

silvasean:
@asaadaldien was also interested in bazel, though I don't know if the exact use case is the same.

I am okay with following LLVM precedent here of having it unsupported on the side and let those interested maintain their own build files.

asaadaldien:
@garymm yes, you can write bazel rules for each cmake library easily with very little manual effort one caveat  there is a small header circular dependency between `Troch\Transforms\* && TorchConversion\` that I had to resolve.

I got everything c++ building with bazel currently wrestling with python side.

As @silvasean mentioned maybe we can add/maintain unsupported `\WORKSPACE \BUILD` for downstream bazel users ?

silvasean:
> @garymm yes, you can write bazel rules for each cmake library easily with very little manual effort one caveat there is a small header circular dependency between `Troch\Transforms\* && TorchConversion\` that I had to resolve.

What was the issue? I would love to have the fix upstream!

garymm:
@asaadaldien sounds like a PR adding Bazel files might be welcome, though it's not in my power to approve.

asaadaldien:
> > @garymm yes, you can write bazel rules for each cmake library easily with very little manual effort one caveat there is a small header circular dependency between `Troch\Transforms\* && TorchConversion\` that I had to resolve.
> 
> What was the issue? I would love to have the fix upstream!

[...]

PR #706, opened 2022-03-28
Add bazel build support (1/N)

asaadaldien:
This PR adds rules for building the compiler part with bazel, 
a followup PRs will include rules for building the python bindings.

Towards fixing #660

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ __pycache__
 *.egg-info
 *.whl
 /wheelhouse
+
+# Bazel
+bazel-*
--- a/utils/bazel/.bazelrc
+++ b/utils/bazel/.bazelrc
@@ -1,0 +1,13 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+build --action_env=CC=clang
+build --action_env=CXX=clang++
+build --cxxopt=-std=c++14 
+build --host_cxxopt=-std=c++14 
+build --cxxopt=-D_GLIBCXX_USE_CXX11_ABI=0
+build --cxxopt=-U__GXX_ABI_VERSION
+build --cxxopt=-D__GXX_ABI_VERSION=1011
+build --cxxopt=-DPYBIND11_COMPILER_TYPE=\"_gcc\"
+build --cxxopt=-DMLIR_PYTHON_PACKAGE_PREFIX=torch_mlir.
--- a/utils/bazel/BUILD.bazel
+++ b/utils/bazel/BUILD.bazel
@@ -1,0 +1,5 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Required to reference .bzl files in this package
--- a/utils/bazel/WORKSPACE.bazel
+++ b/utils/bazel/WORKSPACE.bazel
@@ -1,0 +1,46 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+http_archive(
+    name = "bazel_skylib",
+    urls = [
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-skylib/releases/download/1.0.3/bazel-skylib-1.0.3.tar.gz",
+        "https://github.com/bazelbuild/bazel-skylib/releases/download/1.0.3/bazel-skylib-1.0.3.tar.gz",
+    ],
+    sha256 = "1c531376ac7e5a180e0237938a2536de0c54d93f5c278634818e0efc952dd56c",
+)
+load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
+
+load("@bazel_skylib//:workspace.bzl", "bazel_skylib_workspace")
+
+bazel_skylib_workspace()
+
+new_local_repository(
+    name = "llvm-raw",
+    build_file_content = "# empty",
+    path = "../../externals/llvm-project",
+)
+
+load("@llvm-raw//utils/bazel:configure.bzl", "llvm_configure", "llvm_disable_optional_support_deps")
+
+llvm_configure(
+  name = "llvm-project",
+  repo_mapping = {
+    "@python_runtime": "@local_config_python",
+  },
+)
+llvm_disable_optional_support_deps()
+
+new_local_repository(
+    name = "torch-mlir-raw",
+    build_file_content = "# empty",
+    path = "../../"
+)
+
+load("@torch-mlir-raw//utils/bazel:configure.bzl", "torch_mlir_configure")
+
+torch_mlir_configure(
+  name = "torch-mlir"
+)
--- a/utils/bazel/configure.bzl
+++ b/utils/bazel/configure.bzl
@@ -1,0 +1,56 @@
+# This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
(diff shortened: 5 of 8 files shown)


silvasean:
Let's put all this under utils/bazel and try to keep this as consistent with the way that LLVM does it in https://github.com/llvm/llvm-project/tree/main/utils/bazel

Also, we probably want to comply with the "peripheral tier" requirements in https://llvm.org/docs/SupportPolicy.html#id2 in particular the part "Code in this tier should". @GMNGeoffrey -- what would you recommend we do for testing our bazel build in a way that doesn't alert the main developers? We are a vanilla GH project and use GH actions for our CI. Maybe a nightly bazel build in GH actions that is somehow set to only notify bazel folks (the notifications are customizable like that, right?).

asaadaldien:
> Let's put all this under utils/bazel and try to keep this as consistent with the way that LLVM does it in https://github.com/llvm/llvm-project/tree/main/utils/bazel
> 
> Also, we probably want to comply with the "peripheral tier" requirements in https://llvm.org/docs/SupportPolicy.html#id2 in particular the part "Code in this tier should". @GMNGeoffrey -- what would you recommend we do for testing our bazel build in a way that doesn't alert the main developers? We are a vanilla GH project and use GH actions for our CI. Maybe a nightly bazel build in GH actions that is somehow set to only notify bazel folks (the notifications are customizable like that, right?).

Thanks @silvasean, that will be great to have a setup where breaking bazel won't be merge blocking and downstream bazel will maintain fixing it moving forward.

.bazelrc
```diff
@@ -0,0 +1,9 @@
+build --action_env CC=clang
```
garymm, line 1: Would using an LLVM toolchain from here be a more hermetic alternative (i.e., not require the user to install clang manually)?
https://github.com/grailbio/bazel-toolchain

I used it locally and it seemed to work.

.bazelrc
```diff
@@ -0,0 +1,9 @@
...
+build --action_env CXX=clang++
+build --cxxopt=-std=c++14 
+build --host_cxxopt=-std=c++14 
+build --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=0"
```
garymm, line 5: Nit: be consistent about quoting flag values. Most of the above are not quoted but this and most below are.

.bazelrc
```diff
@@ -0,0 +1,9 @@
+build --action_env CC=clang
+build --action_env CXX=clang++
```
garymm, line 2: Nit: be consistent about equals sign between flag name and flag value. Below you have it, here you don't.

BUILD
```diff
@@ -0,0 +1,818 @@
+load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library", "gentbl_filegroup", "td_library")
```
garymm, line 1: can we name this BUILD.bazel instead of BUILD? It makes it totally obvious what the file is for, syntax highlighting etc.

Same for WORKSPACE.bazel.

garymm (commented):
Thanks for this!

BUILD
asaadaldien: Done!

.bazelrc
asaadaldien: Done!

.bazelrc
asaadaldien: Done!

.bazelrc
asaadaldien: IMO using host clang is preferred in cases where we build a cmake version with the same clang binary.

asaadaldien:
> > Let's put all this under utils/bazel and try to keep this as consistent with the way that LLVM does it in https://github.com/llvm/llvm-project/tree/main/utils/bazel
> > Also, we probably want to comply with the "peripheral tier" requirements in https://llvm.org/docs/SupportPolicy.html#id2 in particular the part "Code in this tier should". @GMNGeoffrey -- what would you recommend we do for testing our bazel build in a way that doesn't alert the main developers? We are a vanilla GH project and use GH actions for our CI. Maybe a nightly bazel build in GH actions that is somehow set to only notify bazel folks (the notifications are customizable like that, right?).
> 
> Thanks @silvasean, that will be great to have a setup where breaking bazel won't be merge blocking and downstream bazel will maintain fixing it moving forward.

Regarding moving this to `utils/bazel`, do you see this as a hard requirement ? because its is going to add one more layer of complexity to the bazel build by doing what [llvm](https://sourcegraph.com/github.com/llvm/llvm-project/-/blob/utils/bazel/configure.bzl?L34) did and add configuration for overlaying the build files into the root directory.

BUILD
stellaraccident: Thanks. Really important to avoid naked BUILD and WORKSPACE files (i wish the bazel folks would deprecate them): on case insensitive file systems, which includes Macos, both bare names tend to conflict with directories that sometimes need to exist (ie. Blocks a sibling build/ directory, which can't even be changed for pip builds).

silvasean:
> > > Let's put all this under utils/bazel and try to keep this as consistent with the way that LLVM does it in https://github.com/llvm/llvm-project/tree/main/utils/bazel
> > > Also, we probably want to comply with the "peripheral tier" requirements in https://llvm.org/docs/SupportPolicy.html#id2 in particular the part "Code in this tier should". @GMNGeoffrey -- what would you recommend we do for testing our bazel build in a way that doesn't alert the main developers? We are a vanilla GH project and use GH actions for our CI. Maybe a nightly bazel build in GH actions that is somehow set to only notify bazel folks (the notifications are customizable like that, right?).
> > 
> > 
> > Thanks @silvasean, that will be great to have a setup where breaking bazel won't be merge blocking and downstream bazel will maintain fixing it moving forward.
> 
> Regarding moving this to `utils/bazel`, do you see this as a hard requirement ? because its is going to add one more layer of complexity to the bazel build by doing what [llvm](https://sourcegraph.com/github.com/llvm/llvm-project/-/blob/utils/bazel/configure.bzl?L34) did and add configuration for overlaying the build files into the root directory.

Yes, I think that being consistent with what LLVM does here is a hard requirement. Hopefully this isn't too burdensome -- I don't think we have to "innovate" here, just follow what they are doing.

asaadaldien:
> > > > Let's put all this under utils/bazel and try to keep this as consistent with the way that LLVM does it in https://github.com/llvm/llvm-project/tree/main/utils/bazel
> > > > Also, we probably want to comply with the "peripheral tier" requirements in https://llvm.org/docs/SupportPolicy.html#id2 in particular the part "Code in this tier should". @GMNGeoffrey -- what would you recommend we do for testing our bazel build in a way that doesn't alert the main developers? We are a vanilla GH project and use GH actions for our CI. Maybe a nightly bazel build in GH actions that is somehow set to only notify bazel folks (the notifications are customizable like that, right?).
> > > 
> > > 
> > > Thanks @silvasean, that will be great to have a setup where breaking bazel won't be merge blocking and downstream bazel will maintain fixing it moving forward.
> > 
> > 
> > Regarding moving this to `utils/bazel`, do you see this as a hard requirement ? because its is going to add one more layer of complexity to the bazel build by doing what [llvm](https://sourcegraph.com/github.com/llvm/llvm-project/-/blob/utils/bazel/configure.bzl?L34) did and add configuration for overlaying the build files into the root directory.
> 
> Yes, I think that being consistent with what LLVM does here is a hard requirement. Hopefully this isn't too burdensome -- I don't think we have to "innovate" here, just follow what they are doing.

SGTM, I will follow what llvm did. and update the PR

GMNGeoffrey:
> what would you recommend we do for testing our bazel build in a way that doesn't alert the main developers? We are a vanilla GH project and use GH actions for our CI. Maybe a nightly bazel build in GH actions that is somehow set to only notify bazel folks (the notifications are customizable like that, right?).

The GitHub actions notification support seemed quite poor when I looked at it. I think you need like a custom job or something, but I don't know much about it.

asaadaldien:
I updated the PR (copy-and-modify what @GMNGeoffrey  did for llvm-project =) ), now torch-mlir bazel builds from `utils/bazel`

asaadaldien:
What do you think @silvasean should I add a github action in this PR, or in a  followup ?

asaadaldien:
What do you think @silvasean should I add a github action in this PR, or in a  followup ?

silvasean:
Let's definitely leave the GH Actions stuff to a follow-up -- those are always a headache to add/debug.

@GMNGeoffrey - could you look over this PR before we merge it and make sure it passes your sniff test?

asaadaldien:
@silvasean any advice regarding merging this ? After this is merged I have follows up PRs to add:
- GH actions for bazel.
- Python bindings.
- Support and run lit tests.

silvasean:
I think it is good to merge!

PR #707, opened 2022-03-28
Eager mode examples and README

makslevental:
Add a simple and ResNet18 example for Eager mode. Also add a bit of description in the README

### DIFF
--- a/examples/eager_mode.py
+++ b/examples/eager_mode.py
@@ -1,0 +1,34 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+
+import torch
+
+from torch_mlir.eager_mode.torch_mlir_tensor import TorchMLIRTensor
+
+torch_a = torch.randn(5, requires_grad=True)
+torch_b = torch.randn(5, requires_grad=True)
+
+torch_c = torch_a + torch_b
+torch_d = torch_a * torch_b
+torch_e = torch_c / torch_d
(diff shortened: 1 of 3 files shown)


examples/eager_mode.py
```diff
@@ -0,0 +1,17 @@
...
+from torch_mlir.eager_mode.torch_mlir_tensor import TorchMLIRTensor
+
+a = TorchMLIRTensor(torch.randn(5, requires_grad=True))
+b = TorchMLIRTensor(torch.randn(5, requires_grad=True))
+
+c = a + b
+d = a * b
+e = c / d
+
+e.sum().backward()
```
silvasean, line 17: print the sum.
Also print out some gradients after the backward call too to show off how seamless this all is :)

(also, random question: does TorchMLIRTensor also cause the backward graph to execute through Torch-MLIR?)

silvasean (changes requested):

examples/eager_mode.py
makslevental: >also, random question: does TorchMLIRTensor also cause the backward graph to execute through Torch-MLIR

yup - it's like magic (the `grad_fn` gets attached to the `TorchMLIRTensor` since it's a subclass of `torch.Tensor` and then during backwards `__torch_dispatch__` fires just like during the forward pass).

examples/eager_mode.py
```diff
@@ -0,0 +1,34 @@
...
+print("PyTorch grad b: ", torch_b.grad)
+
+a = TorchMLIRTensor(torch_a)
+b = TorchMLIRTensor(torch_b)
+
+c = a + b
+d = a * b
+e = c / d
+loss = e.sum()
+print("TorchMLIR loss: ", loss)
```
silvasean, line 30: nit: Torch-MLIR with a `-`.

silvasean (approved):

PR #709, opened 2022-03-28
[SimplifyShapeCalculations] Fix AbstractlyInterpretListOpsWithinABlock

### LINKED ISSUE #704: torch-simplify-shape-calculations is incorrectly changing evaluation ops
I tried adding this test to `test/Dialect/Torch/simplify-shape-calculations.mlir`:
```
// Test the different supported mutation ops.
// CHECK-LABEL:   func @abstractly_interpret_list_ops$mix_mutation_and_evaluation_ops(
// CHECK:             %[[SHAPE:.*]] = torch.prim.ListConstruct %int1, %int2, %arg1, %arg2 : (!torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
// CHECK:             torch.shape.calculate.yield.shapes %[[SHAPE]] : !torch.list<int>
func @abstractly_interpret_list_ops$mix_mutation_and_evaluation_ops(%arg0: !torch.vtensor, %arg1: !torch.int, %arg2: !torch.int) -> !torch.vtensor {
  %int0 = torch.constant.int 0
  %int1 = torch.constant.int 1
  %int2 = torch.constant.int 2
  %int3 = torch.constant.int 3
  %0 = torch.shape.calculate {
    torch.shape.calculate.yield %arg0 : !torch.vtensor
  } shapes {
    %1 = torch.prim.ListConstruct %int1, %int1 : (!torch.int, !torch.int) -> !torch.list<int>
    %2 = torch.aten.len.t %1 : !torch.list<int> -> !torch.int
    %3 = torch.aten._set_item.t %1, %int1, %2 : !torch.list<int>, !torch.int, !torch.int -> !torch.list<int>
    %4 = torch.aten.append.t %1, %arg1 : !torch.list<int>, !torch.int -> !torch.list<int>
    torch.aten.insert.t %1, %int3, %arg2 : !torch.list<int>, !torch.int, !torch.int
    torch.shape.calculate.yield.shapes %1 : !torch.list<int>
  } : !torch.vtensor
  return %0 : !torch.vtensor
}
```

The current result is `torch.prim.ListConstruct %int1, %int4, %arg1, %arg2`, where ` %int4 = torch.constant.int 4`.

It looks to me like `lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp`  `AbstractlyInterpretListOpsWithinABlock` is not taking steps to avoid propagating into uses of the list before the modification happens.

Note: I saw this when investigating why shapes were no longer successfully being computed for `torch.aten.constant_pad_nd`. It was because the negative offset was not being normalized at all places in `SimplifyShapeCalculations.cpp`, and when I added normalization then I got failures like the above. By strengthening the canonicalization of `torch.aten.len.t` and `torch.aten.__getitem__.t` I can avoid pushing `SimplifyShapeCalculations.cpp`, and I wish to open a PR with this approach to unblock `torch.aten.constant_pad_nd`.

silvasean:
The logic in the rewriting phase had a bug in case of a read-only op
coming before mutation ops. The logic would use the op itself as the
"latest literal", but that is not correct, because later on we replace
the op itself with the *final* "latest literal", assuming that all uses
of the op have been rewritten -- that was working in general, except for
any read-only ops at the beginning.

Big thanks to @ljfitz for the tiny reproducer!

Fixes #704

### DIFF
--- a/lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp
+++ b/lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp
@@ -199,7 +199,8 @@ public:
       return failure();
 
     // Rewrite all users to use the appropriate list literals.
-    Value latestLiteral = op;
+    Value latestLiteral = rewriter.create<PrimListConstructOp>(
+        op->getLoc(), op.getType(), op->getOperands());
     int nextLiteral = 0;
     for (Operation *user : usersToInterpret) {
       if (auto append = dyn_cast<AtenAppendTOp>(user)) {
--- a/test/Dialect/Torch/simplify-shape-calculations.mlir
+++ b/test/Dialect/Torch/simplify-shape-calculations.mlir
@@ -191,6 +191,25 @@ func @abstractly_interpret_list_ops$mutation_ops(%arg0: !torch.vtensor, %arg1: !
   return %0 : !torch.vtensor
 }
 
+// Test interspersed mutation and evaluation ops.
+// CHECK-LABEL:   func @abstractly_interpret_list_ops$mix_mutation_and_evaluation_ops(
+// CHECK:             %[[SHAPE:.*]] = torch.prim.ListConstruct %int0, %int1, %int2 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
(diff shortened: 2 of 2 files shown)


ljfitz (approved):
Nice fix. I can also confirm that this also addresses the more complex example I encountered.

PR #710, opened 2022-03-28
Improve list index normalization SimplifyShapeCalculations

ljfitz:
The reified code to compute the shape of `torch.aten.constant_pad_nd` uses negative indices when setting list elements. This was not converted to a positive offset in one place in `SimplifyShapeCalculations` which prevented computation of the static shape.

Additionally I added a utility function `getMatchedListDim()` to replace a number of equivalent pieces of code, as the same code was needed for the fix.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/Utils/Utils.h
+++ b/include/torch-mlir/Dialect/Torch/Utils/Utils.h
@@ -21,6 +21,11 @@ namespace Torch {
 int64_t toPositiveDim(int64_t dim, int64_t inputRank);
 bool isValidDim(int64_t dim, int64_t inputRank);
 bool getListConstructElements(Value v, SmallVectorImpl<Value> &elems);
+/// Returns the index indicated by `v` for a list of given `length`.
+/// If the index is negative, it is adjusted to `length` + `v`.
+/// `None` is returned the index is not an integer in the range [0,`length).
+llvm::Optional<int64_t> matchLegalConstantIndexIntoListOfSize(Value v,
+                                                              int64_t length);
 torch_upstream::ScalarType getScalarTypeForType(Type type);
 // Helper to convert a tensor to a specific scalar type.
 Value convertTensorToDtype(PatternRewriter &rewriter, Location loc, Value input,
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -725,18 +725,14 @@ OpFoldResult AtenSizeIntOp::fold(ArrayRef<Attribute> operands) {
   if (!type || !type.hasSizes())
     return nullptr;
 
-  int64_t inputRank = type.getSizes().size();
-  int64_t dim;
-  if (!matchPattern(this->dim(), m_TorchConstantInt(&dim)))
+  llvm::Optional<int64_t> dimOpt = matchLegalConstantIndexIntoListOfSize(
+      this->dim(), type.getSizes().size());
+  if (!dimOpt)
     return nullptr;
-  dim = toPositiveDim(dim, inputRank);
-  if (!isValidDim(dim, inputRank))
-    return nullptr;
-
-  if (type.getSizes()[dim] == kUnknownSize)
+  if (type.getSizes()[*dimOpt] == kUnknownSize)
     return nullptr;
   return IntegerAttr::get(IntegerType::get(getContext(), 64),
-                          type.getSizes()[dim]);
+                          type.getSizes()[*dimOpt]);
 }
 
 //===----------------------------------------------------------------------===//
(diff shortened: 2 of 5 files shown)


include/torch-mlir/Dialect/Torch/Utils/Utils.h
```diff
@@ -21,6 +21,10 @@ namespace Torch {
 int64_t toPositiveDim(int64_t dim, int64_t inputRank);
 bool isValidDim(int64_t dim, int64_t inputRank);
 bool getListConstructElements(Value v, SmallVectorImpl<Value> &elems);
+/// Returns the dimension indicated by `v` for a list of given `rank`.
```
silvasean, line 24: For consistency (and I also think it reads a little better), I think it would be nice to make this a matcher, so that instead of

```
        if (!matchPattern(setItem.idx(), m_TorchConstantInt(&index)))
          return failure();
```

We instead do
```
        if (!matchPattern(setItem.idx(), m_LegalConstantIndexIntoList(&index, runningList.size())))
          return failure();
```

silvasean (changes requested):

include/torch-mlir/Dialect/Torch/Utils/Utils.h
silvasean: I think there is a separate discussion if we want to migrate some of these

```
        if (!matchPattern(value, m_Something(&thing)))
          return failure();
```

to

```
auto valueOpt = matchSomething(value);
if (!valueOpt)
  return failure();
```

or

```
int64_t thing;
if (matchSomething(&thing))
  return failure()
// (kind of like getListConstructElements)
```

as a matter of style. Right now I want to avoid having too many ways of matching in the codebase.

silvasean: Actually, on thinking about it more, I'll need to do a pass through the codebase to make this consistant anyways because of getListConstructElements, so for now, I'm fine moving forward with this patch, calling the function matchLegalConstantIndexIntoListOfSize. Sorry for all the back and forth.

ljfitz: I've renamed the function as you suggested.

I understand the discussion on consistency. I'm happy to contribute to this. For `matchLegalConstantIndexIntoListOfSize` specifically there were more places in the codebase with the same tests, but with different messages produced for the different reasons it could fail. I think it would be useful to think if we'd want to keep the ability to generate different failure messages. I'd appreciate if this can indeed be done in a new PR.

silvasean: You might want to update the comment too -- it seems to be written from the perspective of a shape, but the utility is really about a general list.

silvasean (approved):

ljfitz:
The PR is failing for an unrelated reason, presumably an update to `torch`. Any tips on where this should be resolved?

```
python -m e2e_testing.torchscript.main --filter IndexPutOneDimIntAccumulateModule_basic --verbose
FAIL - "IndexPutOneDimIntAccumulateModule_basic"

Unexpected outcome summary:

****** Failed tests - 1 tests
    FAIL - "IndexPutOneDimIntAccumulateModule_basic"
          File "/scratch/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 248, in run_tests
          File "/scratch/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/linalg_on_tensors_backend.py", line 37, in compile
            module = convert_torchscript_module_to_torch_backend_contract_mlir(
          File "/scratch/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/utils.py", line 74, in convert_torchscript_module_to_torch_backend_contract_mlir
            raise Exception(f"""
        Exception: 
        PyTorch TorchScript module -> torch-mlir Object Graph IR import failed with:
        Exception:
        required keyword attribute 'unroll_dim' is undefined
        Diagnostics:
```

silvasean:
> The PR is failing for an unrelated reason, presumably an update to `torch`. Any tips on where this should be resolved?
> 
> ```
> python -m e2e_testing.torchscript.main --filter IndexPutOneDimIntAccumulateModule_basic --verbose
> FAIL - "IndexPutOneDimIntAccumulateModule_basic"
> 
> Unexpected outcome summary:
> 
>     FAIL - "IndexPutOneDimIntAccumulateModule_basic"
>         Compilation error: Traceback (most recent call last):
>           File "/scratch/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 248, in run_tests
>             compiled = config.compile(test.program_factory())
>           File "/scratch/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/linalg_on_tensors_backend.py", line 37, in compile
>             module = convert_torchscript_module_to_torch_backend_contract_mlir(
>           File "/scratch/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/utils.py", line 74, in convert_torchscript_module_to_torch_backend_contract_mlir
>             raise Exception(f"""
>         Exception: 
>         PyTorch TorchScript module -> torch-mlir Object Graph IR import failed with:
>         Exception:
>         required keyword attribute 'unroll_dim' is undefined
>         Diagnostics:
> ```

You can go ahead and push. It is unrelated to your PR.

PR #711, opened 2022-03-29
[example/ts_resnet18] removed class wrapper

burntfalafel:
You don't need the extra class wrapper `TestModule()` wrapper


pashu123 (approved):
LGTM! Wait for @cathyzhyi.

cathyzhyi (approved):

PR #717, opened 2022-03-30
[MLIR][TORCH] Add E2E support for aten.new_empty op

vivekkhandelwal1:
This commit decomposes `aten.new_empty` op into `aten.empty.memory_format` op.

This commit also made a dtype fix to the constant tensor allocation like ops.
Earlier the dtype for the result was inferred from the result type; now, it's
being evaluated as per the original definition of the op.


vivekkhandelwal1:
There were some issues with the original branch related to merging. I have created this new PR with the code same as https://github.com/llvm/torch-mlir/pull/697.

PR #719, opened 2022-03-30
Fix typo in release assets workflow name

GMNGeoffrey:


silvasean:
It is actually correct as-is. It should probably be renamed "Release+Asserts"
https://github.com/llvm/torch-mlir/blob/969785d1b64e0ed0c64be98d83276678d9a75554/.github/workflows/buildAndTest.yml#L16

GMNGeoffrey:
Ah because it's publishing a release and says assets below, I assumed :grin: "Release+Asserts" doesn't really help IMO because "release" is being used as both a build configuration and a publishing of artifacts. Personally, I think it's weird that the workflow that publishes releases is called "build and test" though

PR #725, opened 2022-03-31
Added example Torch MLIR LTC Backend

henrytwo:
Added example vendor implementation of MLIR LTC backend, which shows what information is exposed at the compile stage. Also included is a script which uses this backend implementation to demonstrate the full end-to-end flow from PyTorch model to MLIR.

**Note: This PR is based on https://github.com/llvm/torch-mlir/pull/724, but GitHub does not allow for setting the base to a fork while keeping this PR in the upstream repo, so there are several shared commits. Only the last three are specifically part of this PR.**

Marked as draft while waiting for a dependency PR to land: https://github.com/llvm/torch-mlir/pull/724

cc: @antoniojkim @ke1337

### DIFF
--- a/examples/ltc_backend/CMakeLists.txt
+++ b/examples/ltc_backend/CMakeLists.txt
@@ -1,0 +1,63 @@
+###########################################################################
+# Setup PyTorch
+###########################################################################
+
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/python/torch_mlir/cmake/modules")
+include(TorchMLIRPyTorch)
+TorchMLIRProbeForPyTorchInstall()
+find_package(Torch 1.11 REQUIRED)
+
+TorchMLIRConfigurePyTorch()
+
+###########################################################################
+# Setup Python development
+###########################################################################
+
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/external/llvm-project/mlir/cmake/modules")
+include(MLIRDetectPythonEnv)
+mlir_configure_python_dev_packages()
+
+###########################################################################
+# Library definition
+###########################################################################
+
+include_directories(BEFORE
+        ${TORCH_INCLUDE_DIRS}
+        ${CMAKE_CURRENT_SOURCE_DIR}
+        ${CMAKE_CURRENT_BINARY_DIR}
+        ${Python3_INCLUDE_DIRS}
+        ${PYTHON_H_DIR}
+        ${PROJECT_SOURCE_DIR}/python
+        )
+link_directories("${TORCH_INSTALL_PREFIX}/lib")
+link_directories(${CMAKE_CURRENT_SOURCE_DIR}/ltc_backend/lib)
+add_link_options(-Wl,-rpath,$ORIGIN/ltc_backend/lib)
+
+file(GLOB LTC_BACKEND_CSRC CONFIGURE_DEPENDS
+        "ltc_backend/csrc/*.h"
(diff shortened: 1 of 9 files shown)


examples/ltc_backend/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
+if ("${TORCH_DIR}" STREQUAL "")
```
silvasean, line 1: shouldn't this be set below in TorchMLIRProbeForPyTorchInstall?

examples/ltc_backend/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
...
+    get_filename_component(TORCH_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../../pytorch/" ABSOLUTE)
+endif()
+message(STATUS "TORCH_DIR=${TORCH_DIR}")
+
+if ("${TORCH_MLIR_DIR}" STREQUAL "")
+    get_filename_component(TORCH_MLIR_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../" ABSOLUTE)
```
silvasean, line 7: can we use https://cmake.org/cmake/help/latest/variable/PROJECT_SOURCE_DIR.html ?

examples/ltc_backend/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
...
+        SUFFIX "${PYTHON_MODULE_EXTENSION}"
+        CXX_VISIBILITY_PRESET "hidden"
+        COMPILE_FLAGS "${TORCH_CXXFLAGS} -Wno-pedantic"
+        )
+
+###########################################################################
+# Compile Options
+###########################################################################
+
+add_compile_options(-Wno-sign-compare)
```
silvasean, line 79: Is this to deal with upstream warnings? Can you file an upstream bug for these? Some seem pretty serious.

examples/ltc_backend/CMakeLists.txt
```diff
@@ -0,0 +1,94 @@
...
+add_compile_options(-Wno-unknown-pragmas)
+add_compile_options(-Wno-deprecated-declarations)
+add_compile_options(-Wno-return-type)
+
+if ("${CMAKE_BUILD_TYPE}" STREQUAL "Debug")
```
silvasean, line 84: These seem pretty close to the defaults that CMake already chooses. Can we remove these in the interest of simplicity?

silvasean (approved):

examples/ltc_backend/CMakeLists.txt
henrytwo: I took a look and it seems like these compiler options were left over from the CMake file I referenced, so I went ahead and removed them. Doesn't seem like these particular warnings appear when building this part of the project

examples/ltc_backend/ltc_backend/csrc/backend/backend_impl.h
```diff
@@ -0,0 +1,27 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#pragma once
+
+#include <torch/csrc/lazy/backend/backend_interface.h>
+
+namespace at {
+// This function is defined in the codegenerated RegisterLazy.cpp file.
+extern TORCH_API void RegisterTorchMlirLazyNativeFunctions();
```
antoniojkim, line 16: nit: I don't think the `extern` is necessary

antoniojkim (approved):

PR #726, opened 2022-04-01
Fix out-of-tree cmake build of `torch-mlir-dialects`

oowekyala:
This follows up on #623 for out-of-tree builds of torch-mlir, which added building `torch-mir-dialects` as a subdirectory.

We found that we needed some more `cmake` adaptions to make this work. We're initially opening this PR as a draft to gather opinions as to whether this approach suits at least the short-term `torch-mlir-dialects` goals.

Our goal is to support both in-tree and out-of-tree builds of `torch-mlir` with minimum hassle, for instance by using the same variable names in both setups. Specifically in `externals/llvm-external-projects/torch-mlir-dialects/CMakeLists.txt` :
- We use `MLIR_FOUND` to detect that it is being build as a subdirectory and the llvm+mlir cmake infrastructure is already set up (via find_package in the parent build) as opposed to an in-tree build.
- For in-tree,  the setting of variables and loading of llvm+mlir cmake infrastructure is now conditionally performed.
- For in-tree, the names of cmake variables being defined for are adjusted to match those `llvm-project` makes available through `find_package(MLIR REQUIRED CONFIG)`, under the assumption that those are the more "standardized" names.

Further optional items we could contribute if you're interested:
- build instructions for an out-of-tree setup (eg in README.md)
- a CI flow to test an out-of-tree setup, so that all code paths of the cmake scripts are tested (even though we expect most llvm developers to use in-tree builds)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -108,8 +108,7 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/include)
 
 function(torch_mlir_target_includes target)
   set(_dirs
-    $<BUILD_INTERFACE:${MLIR_INCLUDE_DIR}>
-    $<BUILD_INTERFACE:${MLIR_GENERATED_INCLUDE_DIR}>
+    $<BUILD_INTERFACE:${MLIR_INCLUDE_DIRS}>
     $<BUILD_INTERFACE:${TORCH_MLIR_SOURCE_DIR}/include>
     $<BUILD_INTERFACE:${TORCH_MLIR_BINARY_DIR}/include>
   )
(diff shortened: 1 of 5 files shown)


silvasean:
Thanks for working on this. Out-of-tree build instructions and CI would be very much appreciated. I'll let @cathyzhyi review the PR itself since she is most familiar with torch-mlir-dialects.

cathyzhyi (approved):
Thanks for the change! I am not very familiar with llvm cmakes but the change looks reasonable to the best of my knowledge. 

> Further optional items we could contribute if you're interested:
>    build instructions for an out-of-tree setup (eg in README.md)
>    a CI flow to test an out-of-tree setup, so that all code paths of the cmake scripts are tested (even though we expect most llvm developers to use in-tree builds)

Those are great items to add. Thanks!

oowekyala:
@silvasean @cathyzhyi Thanks for the reviews! I'll put together a followup PR for those additional items.

PR #729, opened 2022-04-02
Add 1D, weight, and reduction support to nll_loss_backward

ramiro050:
This commit adds the following support to the op `nll_loss_backward`:
- `input` tensor can be rank-1
- `weight` parameter
- `reduction` parameter
- `target`, `grad_output`, `total_weight` can be rank-0
- Checks that input tensors are of the expected type

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -28,6 +28,13 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 
+// Check if a ranked-tensor has the specified element type.
+template <typename elementType> static bool hasElementType(Value tensor) {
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1456,89 +1458,136 @@ class ConvertAtenNllLossBackwardOp
...
+
+                  if (reduction == torch_upstream::Reduction::Mean) {
+                    gradOutElem = b.create<arith::DivFOp>(loc, gradOutElem,
+                                                          totalWeightElem);
+                  }
+
+                  Value negGradOutElem =
+                      b.create<arith::NegFOp>(loc, gradOutElem);
+                  Value weightElem = getConstant(b, loc, 1, resultElementType);
+                  if (!weightIsNone)
```
cathyzhyi, line 1567: nit: {} is needed

cathyzhyi (approved):
LGTM except one nit

PR #731, opened 2022-04-03
[LINALG] Add the lowering of `aten.ne.Scalar` op

pashu123:
The lowering of `aten.ne.Scalar` op has been added to
the linalg backend.


pashu123:
@ramiro050 I think it can be decomposed into `torch.logical_not(torch.eq)` but I don't find any suitable operator for `logical_not`. Do you have any suggestions?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -617,6 +617,28 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
 
+  if (auto neScalar = dyn_cast<AtenNeScalarOp>(op)) {
+    Type dtype = neScalar.self().getType().cast<BaseTensorType>().getDtype();
+    Value otherPromoted =
+        convertScalarToDtype(b, loc, operands[1], payloadArgs[0].getType());
+
+    if (dtype.isa<mlir::FloatType>())
+      return b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UNE,
+                                     payloadArgs[0], otherPromoted);
+    if (dtype.isa<mlir::IntegerType>()) {
```
cathyzhyi, line 628: Can you refer to https://github.com/llvm/torch-mlir/blob/5b9bdfaf3fb56c5aed12d689e21c9884b1c5a789/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L153 and add something similar like `createNe`?

cathyzhyi (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
pashu123: Added PTAL.

cathyzhyi (approved):

Issue #678, opened 2022-03-17
[RFC] Adding support for lowering control flow.


gprateek93:
Hi,
My team is working in integrating [functorch](https://github.com/pytorch/functorch), so that we can get complex ops decomposed before applying the lowering pass. Some of these decompositions are designed such that they incorporate for/while loops in them. These loops can be mapped to the `torch::PrimLoopOp` in torch-mlir. For end to end execution lowering of this operation is required.
Some examples from the functorch repo where loops are used:

1. [torch.split_with_sizes](https://github.com/pytorch/functorch/blob/6f3e1371cd2ace8b90893288d47281091e420a1d/functorch/_src/decompositions.py#L385)
```
def split_with_sizes(self: Tensor, split_sizes: List[int], dim: int = 0) -> List[Tensor]:
    num_splits = len(split_sizes)
    splits = []
    start_idx = 0
    for i in range(num_splits):
        length = split_sizes[i]
        splits.append(self.narrow(dim, start_idx, length))
        start_idx += length
    return splits
```

2. [torch.split](https://github.com/pytorch/functorch/blob/6f3e1371cd2ace8b90893288d47281091e420a1d/functorch/_src/decompositions.py#L405)
```
def split(self: Tensor, split_size: int, dim: int = 0) -> List[Tensor]:
    input_sizes = self.shape
    dim_size = input_sizes[dim]
    if split_size == 0:
        assert(dim_size == 0)
        return [self]
    chunks = (dim_size + split_size - 1) // split_size
    split_sizes = [split_size for i in range(chunks)]
    split_sizes[chunks - 1] = split_size - (split_size * chunks - dim_size)
    return aten.split_with_sizes(self, split_sizes, dim)
```
I am working on adding lowering for `Torch::PrimLoopOp` into `scf.while` operation.
This would require two cases:

1. **Lowering while control flow**:
This can be carried on the same lines as the lowering of mhlo.WhileOp to scf.WhileOp. [Code](https://github.com/tensorflow/mlir-hlo/blob/078a98ec5b7127fd46c050c8bcb81e6e97ad0ecb/lib/Dialect/mhlo/transforms/legalize_control_flow.cc#L78)
2. **Lowering for control flow**:
This can be carried by changing the adding a condition of `i < maxTripCount` as `scf.ConditionOp` and follow same lowering as for the while loop case.

One more alternative is to create ``LinalgExt::WhileOp`` if IREE does not support scf dialect for now.

[...]

PR #734, opened 2022-04-04
[TORCH][MLIR]Add lowering for control flow.

gprateek93:
1. This commit adds lowering of while like prim loop to scf.while
operation.
2. Convert for-like prim loops into scf.for op.

Issue:  [#678](https://github.com/llvm/torch-mlir/issues/678)

### DIFF
--- a/lib/Conversion/TorchToSCF/TorchToSCF.cpp
+++ b/lib/Conversion/TorchToSCF/TorchToSCF.cpp
@@ -10,10 +10,13 @@
 #include "torch-mlir/Conversion/TorchToSCF/TorchToSCF.h"
 
 #include "../PassDetail.h"
+#include "mlir/Dialect/Arithmetic/IR/Arithmetic.h"
 #include "mlir/Dialect/SCF/SCF.h"
+#include "mlir/IR/BuiltinTypes.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchDialect.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchTypes.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionDialect.h"
 #include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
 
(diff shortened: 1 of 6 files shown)


silvasean:
> 2\. TODO: Convert for-like prim loops into while like prim loops.

I would recommend lowering for-like loops to scf.for directly to preserve information. It is probably less code than converting the for-like to while-like as well.

test/Conversion/TorchToSCF/basic.mlir
```diff
@@ -64,3 +64,37 @@ func @aten.prim.if$nested(%arg0: !torch.bool, %arg1: !torch.bool) -> !torch.int
...
+  // CHECK-SAME:    (%[[LOOP_ARG:.*]] = %[[FLOAT_VAL]]) : (f64) -> f64 {
+  // CHECK-NEXT:      %[[TORCH_LOOP_ARG:.*]] = torch_c.from_f64 %[[LOOP_ARG]]
+  // CHECK-NEXT:      %[[TORCH_CONDITION:.*]] = torch.aten.lt.float_int %[[TORCH_LOOP_ARG]], %[[ARG0]]
+  // CHECK-NEXT:      %[[CONDITION:.*]] = torch_c.to_i1 %[[TORCH_CONDITION]]
+  // CHECK-NEXT:      %[[CONDITION_:.*]] = torch_c.to_i1 %[[TORCH_CONDITION]]
+  // CHECK-NEXT:      scf.condition(%[[CONDITION_]]) %[[LOOP_ARG]]
+  // CHECK-NEXT:    } do {
+  // CHECK-NEXT:    ^bb0(%[[BLOCK_ARG:.*]]: f64):
+  // CHECK-NEXT:      %[[TORCH_BLOCK_ARG:.*]] = torch_c.from_f64 %[[BLOCK_ARG]]
+  // CHECK-NEXT:      %[[TORCH_VAL:.*]] = torch.aten.mul.float %[[TORCH_BLOCK_ARG]], %[[TORCH_BLOCK_ARG]]
```
silvasean, line 83: why is this op in the "after" region? I think you should be able to do the lowering by just putting everything in the "before" region and leaving the "after" region empty.

gprateek93: The before region of the `scf.while` contains the condition of the while loop while the after region has the body of the loop. That is why I coded it like that.

gprateek93: Even the lowering for `mhlo.while` to `scf.while` followed the same, which I took as a reference. https://github.com/tensorflow/mlir-hlo/blob/078a98ec5b7127fd46c050c8bcb81e6e97ad0ecb/lib/Dialect/mhlo/transforms/legalize_control_flow.cc#L78

silvasean: I don't think there is any such restriction for scf.while. You should be able to put the whole computation in the "before" region.

silvasean: Or another alternative is to calculate the condition in the after region and the pass it through the iterargs into the before region -- I think you might need to do that to handle the initial condition cleanly.

silvasean: mhlo.while structurally different from a while-like PrimLoopOp though. A while-like PrimLoopOp is actually do-while-like with a peeled initial check for the condition.

silvasean: Yeah, I think the natural lowering is to put the body of the PrimLoopOp into the "after" region of scf.while, and then pass the condition as an iterArg back to the "before" region. Then the PrimLoopOp initialCondition is handled by passing it as an iterArg to scf.while.

silvasean: It's unfortunate that PrimLoopOp is defined this way, because it forces the frontend to peel the condition, which make it hard to inline large conditions. Here is an example:

```
def f(x):
    while x.sum() > 0:
        x = x * x
    return x
```

```
graph(%x.1 : Tensor):
  %22 : NoneType = prim::Constant()
  %1 : int = prim::Constant[value=9223372036854775807]() # /tmp/repro.py:4:4
  %5 : int = prim::Constant[value=0]() # /tmp/repro.py:4:20
   // Notice the aten::sum/gt/Bool is replicated here...
  %23 : Tensor = aten::sum(%x.1, %22) # /tmp/repro.py:4:10
  %24 : Tensor = aten::gt(%23, %5) # /tmp/repro.py:4:10
  %26 : bool = aten::Bool(%24) # /tmp/repro.py:4:10
  %x : Tensor = prim::Loop(%1, %26, %x.1) # /tmp/repro.py:4:4
    block0(%9 : int, %x.15 : Tensor):
      %x.9 : Tensor = aten::mul(%x.15, %x.15) # /tmp/repro.py:5:12
     // ... and replicated here
      %4 : Tensor = aten::sum(%x.9, %22) # /tmp/repro.py:4:10
      %6 : Tensor = aten::gt(%4, %5) # /tmp/repro.py:4:10
      %8 : bool = aten::Bool(%6) # /tmp/repro.py:4:10
      -> (%8, %x.9)
  return (%x)
```

gprateek93: Thanks @silvasean for your insights. Will make necessary changes in the code.

lib/Conversion/TorchToSCF/TorchToSCF.cpp
```diff
@@ -61,18 +69,281 @@ class ConvertTorchPrimIfOp : public OpConversionPattern<PrimIfOp> {
 };
 } // namespace
 
+namespace {
+
+// `getTypeConvertedValue` creates a TorchConversion operation to convert the
+// `type` of the `operand` from a torch type to mlir type and vice versa.
+static Value getTypeConvertedValue(ConversionPatternRewriter &rewriter,
```
silvasean, line 76: You shouldn't need this -- use the TypeConvert itself -- it has methods like materializeSourceConversion materializeTargetConversion for this.

gprateek93: Done.

lib/Conversion/TorchToSCF/TorchToSCF.cpp
```diff
@@ -10,12 +10,21 @@
...
+#include "mlir/IR/Region.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchDialect.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchTypes.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionDialect.h"
+#include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h"
 #include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
+#include "llvm/ADT/None.h"
+#include "llvm/Frontend/OpenMP/OMP.h.inc"
```
pashu123: Any reasons why we are including OMP?

lib/Conversion/TorchToSCF/TorchToSCF.cpp
```diff
@@ -61,18 +63,248 @@ class ConvertTorchPrimIfOp : public OpConversionPattern<PrimIfOp> {
...
+            typeConverter->convertTypes(op.getResultTypes(), newResultTypes)))
+      return rewriter.notifyMatchFailure(
+          op, "could not convert PrimLoopOp outputs");
+
+    // Create scf.while operation using the operands of torch::primloop. The
+    // first argument of the primloop correspond to `maxTripCount`  which
+    // can be omitted in the `scf.while` operation.
+    Value condition = adaptor.initialCondition();
+    ValueRange iterArgsInit = adaptor.iterArgsInit();
+    SmallVector<Value> newWhileOpOperands{condition};
```
pashu123, line 91: Can be renamed to `scfWhileOperands`.

pashu123 (changes requested):
Please add an e2e test.

silvasean:
Thanks, this is what I had in mind :+1:

gprateek93:
> Please add an e2e test.

added e2e test.

python/torch_mlir_e2e_test/test_suite/control_flow.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.int64, True)
+    ])
+    def forward(self, x):
+        x_val = x.size(0)
+        sum = 0
+        for i in range(x_val):
```
silvasean, line 28: Can you test a while-like loop too?

python/torch_mlir_e2e_test/test_suite/control_flow.py
```diff
@@ -0,0 +1,35 @@
...
+            sum += i
+        return sum
+        
+
+@register_test_case(module_factory=lambda: TorchPrimLoopForLikeModule())
+def TorchPrimLoopForLikeModule_basic(module, tu: TestUtils):
+    module.forward(torch.randint(0,10,(6,8)))
```
silvasean, line 35: nit: please format this (spacing around the `,`'s)

silvasean (approved):

pashu123 (approved):
LGTM!

PR #736, opened 2022-04-04
Make e2e testing parallel

cathyzhyi:
This change makes the e2e testing parallel using the multiprocessing
python module.

This also fixes the inplace update tensor issue we had
where the torchscript execution would update the input value inplace
resulting the actual test not being able to see the original input
value.

### DIFF
--- a/python/torch_mlir_e2e_test/torchscript/framework.py
+++ b/python/torch_mlir_e2e_test/torchscript/framework.py
@@ -26,7 +26,7 @@ from typing import Any, Callable, List, NamedTuple, Optional, TypeVar, Union, Di
 import traceback
 
 import torch
-
+import torch.multiprocessing as mp
 
 TorchScriptValue = Union[int, float, List['TorchScriptValue'],
                          Dict['TorchScriptValue',
@@ -53,10 +53,6 @@ class TraceItem(NamedTuple):
 # public boundary.
 Trace = List[TraceItem]
 
-# A type shared between the result of `TestConfig.compile` and the input
-# to `TestConfig.run`. Each backend will likely have a different definition of
-# this type.
-CompiledArtifact = TypeVar('CompiledArtifact')
 
 # Clone all the tensor values.
 def clone_torch_script_value(v: TorchScriptValue):
@@ -76,6 +72,26 @@ def clone_torch_script_value(v: TorchScriptValue):
     assert False, "unhandled cloning of TorchScriptValue value type"
 
 
+# This clone helper is used to work around issues with output tensors when
+# using multiprocessing module to run tests. The error happens for tests like
+# ContiguousModule_basic where the output tensor aliases with an input tensor.
+# When the output tensor is not cloned, the testing trace would be modified for
+# unknown reason when passed through the shared memory through synchronized
+# queue for example.
+# TODO: Figure out the root cause of the failure and fix properly.
+def clone_trace(trace: Trace) -> Trace:
+    return [
+        TraceItem(symbol=item.symbol,
+                  inputs=item.inputs,
+                  output=clone_torch_script_value(item.output))
+        for item in trace
+    ]
+
+
+# A type shared between the result of `TestConfig.compile` and the input
+# to `TestConfig.run`. Each backend will likely have a different definition of
+# this type.
+CompiledArtifact = TypeVar('CompiledArtifact')
 
 class TestConfig(abc.ABC):
     """The interface implemented by backends to run tests.
@@ -158,6 +174,7 @@ class TestUtils:
     TODO: Figure out how to seed reset properly scoped to just a test case
     (such as when running tests in parallel)
     """
+
     def __init__(self):
         torch.manual_seed(0)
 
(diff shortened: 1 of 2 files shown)


python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -53,6 +53,32 @@ class TraceItem(NamedTuple):
...
+    if isinstance(v, float) or isinstance(v, int) or isinstance(v, str):
+        return v
+    assert("unhandled cloning of TorchScriptValue value type")
+
+
+# This clone helper is used to work around issues with output tensors when
+# using multiprocessing module to run tests. The error happens for tests like
+# ContiguousModule_basic where the output tensor aliases with an input tensor.
+# When the output tensor is not cloned, the testing trace would be modified for
+# unknown reason when passed through the shared memory through synchornized
```
silvasean, line 71: typo synchronized

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -53,6 +53,32 @@ class TraceItem(NamedTuple):
 # public boundary.
 Trace = List[TraceItem]
 
+# Clone all the tensor values.
+def clone_torchscriptvalue(v: TorchScriptValue):
```
silvasean, line 57: nit: clone_torch_script_value for consistency

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
-        except Exception as e:
-            results.append(
-                TestResult(unique_name=test.unique_name,
-                           compilation_error=None,
-                           runtime_error="".join(traceback.format_exception(
-                               type(e), e, e.__traceback__)),
-                           trace=None,
-                           golden_trace=None))
-            continue
+def compile_and_run_test(test_name: str, results: mp.Queue, tests: Dict[str, Test],
```
silvasean, line 272: nit: can you change this function to have a signature `compile_and_run_test(test: Test, config: TestConfig) -> TestResult` (that knows nothing about parallelism/etc.? And then inline the other parts of this function into `def worker` below.

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
+
+    results.append(TestResult(unique_name=test.unique_name,
+                        compilation_error=None,
+                        runtime_error=None,
+                        trace=clone_trace(trace),
+                        golden_trace=golden_trace))
+
+
+def run_tasks_in_parrallel(task_queue: mp.Queue, task):
+    NUMBER_OF_PROCESSES = min(mp.cpu_count(), task_queue.qsize())
```
silvasean, line 310: Usually it is good to have slightly more processes than CPU's. So maybe something like `int(mp.cpu_count() * 1.1)` or something.

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
+    for i in range(NUMBER_OF_PROCESSES):
+        task_queue.put('STOP')
+    for p in processes:
+        p.join()
+
+
+def run_tests(tests: List[Test], config: TestConfig) -> List[TestResult]:
+    """Invoke the given `Test`'s with the provided `TestConfig`."""
+
+    manager = mp.Manager()
```
silvasean, line 325: Can you add a comment here describing the overall strategy.

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
+    for test in tests:
+        tests_queue.put(test.unique_name)
+
+    test_dict = {test.unique_name: test for test in tests}
+
+    def worker(q: mp.Queue):
+        for test_name in iter(q.get, 'STOP'):
+            compile_and_run_test(test_name, sync_results, test_dict, config)
+
+    run_tasks_in_parrallel(tests_queue, worker)
```
silvasean, line 341: typo: parallel

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
+    # This is needed because autograd does not support crossing process
+    # boundaries.
+    torch.autograd.set_grad_enabled(False)
+    for test in tests:
+        tests_queue.put(test.unique_name)
+
+    test_dict = {test.unique_name: test for test in tests}
+
+    def worker(q: mp.Queue):
+        for test_name in iter(q.get, 'STOP'):
```
silvasean, line 338: nit: `"QUEUE_SENTINEL"` to make it more obvious (also, for consistency, use `"`)

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
+        p.join()
+
+
+def run_tests(tests: List[Test], config: TestConfig) -> List[TestResult]:
+    """Invoke the given `Test`'s with the provided `TestConfig`."""
+
+    manager = mp.Manager()
+    tests_queue = manager.Queue()
+    task_queue = manager.Queue()
+    sync_results = manager.list()
```
silvasean, line 328: why is this list instead of queue? The signature of compile_and_run_test says `mp.Queue`.

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
+
+    test_dict = {test.unique_name: test for test in tests}
+
+    def worker(q: mp.Queue):
+        for test_name in iter(q.get, 'STOP'):
+            compile_and_run_test(test_name, sync_results, test_dict, config)
+
+    run_tasks_in_parrallel(tests_queue, worker)
+    tests_with_results = {result.unique_name for result in sync_results}
+    all_tests = { test.unique_name for test in tests }
```
silvasean, line 343: nit: no spaces in `{...}`"  `{test.unique_name for test in tests}`

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
+    aborted_tests = list(all_tests - tests_with_results)
+    aborted_tests = [
```
silvasean, line 345: nit: create a new variable here `aborted_test_results`

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
+        TestResult(
+            unique_name=aborted_test_name,
+            compilation_error=None,
+            runtime_error=
+            "Testing process terminated. This is mostly likely because of hitting runtime assertion.\n",
```
silvasean, line 350: Is there a way to get the stdout of the failing process? I'm thinking what the debugging flow will be if there is a crash -- how do we see the error / repro / debug it?

Also, most crashes I have seen happen inside the compiler (failed `cast<>`, missing case in the importer, failed `assert`, ...). Maybe say "Either the compiler crashed or the compiled code crashed at runtime."

silvasean (changes requested):

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +269,89 @@ def generate_golden_trace(test: Test) -> Trace:
...
+                       compilation_error=None,
+                       runtime_error="".join(
+                           traceback.format_exception(type(e), e,
+                                                      e.__traceback__)),
+                       trace=None,
+                       golden_trace=None))
+        return
+
+    results.append(TestResult(unique_name=test.unique_name,
+                        compilation_error=None,
```
silvasean, line 303: nit: is this formatted?

python/torch_mlir_e2e_test/torchscript/framework.py
silvasean: Okay, that seems acceptable for now.

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -238,40 +278,95 @@ def generate_golden_trace(test: Test) -> Trace:
...
+            sync_results.append(
+                compile_and_run_test(tests_dict[test_name], config))
+
+    run_workers_in_parallel(tests_queue, worker)
+    tests_with_results = {result.unique_name for result in sync_results}
+    all_tests = {test.unique_name for test in tests}
+    # For processes that are crashed due to compile time or runtime error,
+    # the error outputs are printed out all together but no TestResult is
+    # produced when the process crashed.
+    # TODO: find a clean way to capture the output from crashed process and
```
silvasean, line 357: nit: capitalize Find.

silvasean (approved):

cathyzhyi:
hmm, it seems with multiprocess library the CI build takes longer to finish 30m vs 2m for refbackend tests.

cathyzhyi:
Anyway I can investigate the issue on CI machines directly? On my local machine, it's 4m(after change) vs 45mins(before change) as expected.

silvasean:
I think the CI machines have only 2 cores. It might be possible to reproduce by limiting to 2 cores locally.

But no, I haven't found a way to easily reproduce the CI runs except by pushing new commits to PR's.

> On my local machine, it's 4m(after change) vs 45mins(before change) as expected.

What was taking 45 minutes before this change?

cathyzhyi:
> What was taking 45 minutes before this change?

running all the e2e tests with refbackend on my local machine takes 45mins before this change.

silvasean:
> > What was taking 45 minutes before this change?
> 
> running all the e2e tests with refbackend on my local machine takes 45mins before this change.

Interesting... does this include heavydep tests as well? I think it takes <5 minutes on my local machine (without heavydep). And the CI machines take ~2 min (w/o heavydep), and they are very slow.

PR #737, opened 2022-04-06
[LINALG] Add torch.layout information

pashu123:
torch.layout information has been added.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/Utils/TorchUpstream.h
+++ b/include/torch-mlir/Dialect/Torch/Utils/TorchUpstream.h
@@ -97,6 +97,13 @@ enum MemoryFormat {
   ChannelsLast3d
 };
 
+//===----------------------------------------------------------------------===//
+// Possible values for `layout` argument in PyTorch ops that support it.
+// Source:
+// https://github.com/pytorch/pytorch/blob/master/c10/core/Layout.h
+//===----------------------------------------------------------------------===//
+enum Layout { Strided, Sparse, SparseCsr, Mkldnn, NumOptions };
+
 } // namespace torch_upstream
 } // namespace torch
 } // namespace mlir
--- a/lib/Conversion/TorchToLinalg/TensorConstructors.cpp
+++ b/lib/Conversion/TorchToLinalg/TensorConstructors.cpp
@@ -121,9 +121,8 @@ public:
 
     // TODO: Add support for layout, pin_memory features.
     // Only `none` layout is supported.
-    if (!op.layout().getType().template isa<Torch::NoneType>())
-      return rewriter.notifyMatchFailure(
-          op, "unimplemented: only default layout is supported");
+    // At this point all tensors should have value semantics, and hence the
+    // `layout` check can be ignored.
 
     // The pin_memory should be either `False` or `none`.
     bool pinMemory;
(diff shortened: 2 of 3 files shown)


lib/Conversion/TorchToLinalg/TensorConstructors.cpp
```diff
@@ -181,12 +181,17 @@ class ConvertAtenEmptyMemoryFormatOp
...
 
     // TODO: Add support for layout, pin_memory and memory_format features.
-    // Only `none` layout is supported.
-    if (!op.layout().getType().template isa<Torch::NoneType>())
+    // Only 'strided' and `none` layout is supported. The memory layout should
+    // be `none` or `contiguous` to make sure that default strides are used.
+    int64_t layout;
+    if (!op.layout().getType().template isa<Torch::NoneType>() &&
+        (!matchPattern(op.layout(), m_TorchConstantInt(&layout)) ||
+         layout != torch_upstream::Layout::Strided))
```
cathyzhyi, line 194: the last check `layout !=torch_upstream::Layout::Strided` should only be executed when `matchPattern(op.layout(), m_TorchConstantInt(&layout)` succeed. You can refer to the code here:
https://github.com/llvm/torch-mlir/blob/5b9bdfaf3fb56c5aed12d689e21c9884b1c5a789/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp#L205-L213. There are multiple other places in this file with similar issue. Can you help to fix those as well? Maybe add helps like `checkIfNoneOrConstantbool`, `checkIfNoneOrConstantInt` etc

include/torch-mlir/Dialect/Torch/Utils/TorchUpstream.h
```diff
@@ -97,6 +97,16 @@ enum MemoryFormat {
   ChannelsLast3d
 };
 
+//===----------------------------------------------------------------------===//
+// Possible values for `layout` argument in PyTorch ops that support it.
+// Source:
+// https://pytorch.org/docs/stable/tensor_attributes.html#torch.torch.layout
```
cathyzhyi, line 103: I think this is not the right source. Looking at pytorch's implementation of empty here:
https://github.com/pytorch/pytorch/blob/3c1c7fd26e94d00cfa50b7e6dab79f7c7988ae40/aten/src/ATen/EmptyTensor.cpp#L98-L106

I think it's this enum that needs to be copied https://github.com/pytorch/pytorch/blob/3c1c7fd26e94d00cfa50b7e6dab79f7c7988ae40/c10/core/Layout.h#L9

cathyzhyi (changes requested):

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
pashu123: Okay sure.

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
```diff
@@ -181,12 +181,17 @@ class ConvertAtenEmptyMemoryFormatOp
...
                   ConversionPatternRewriter &rewriter) const override {
+
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
       return failure();
 
     // TODO: Add support for layout, pin_memory and memory_format features.
-    // Only `none` layout is supported.
-    if (!op.layout().getType().template isa<Torch::NoneType>())
+    // Only 'strided' and `none` layout is supported. The memory layout should
+    // be `none` or `contiguous` to make sure that default strides are used.
```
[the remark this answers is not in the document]
pashu123: I see, I will remove the layout check.

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
pashu123: As suggested by sean, I am removing the layout check.

cathyzhyi (approved):

PR #733, opened 2022-04-04
Add README instructions and CI flow for OOT build

oowekyala:
(this is not reviewable yet, I'm using the PR to test the CI)

(OOT = out-of-tree)

Followup on #726

todo description

PR #738, opened 2022-04-06
Add README instructions for OOT build

oowekyala:
(OOT = out-of-tree)

Followup on #726 (#733 was the draft)

* Add instructions in the README to show how to build the project OOT
* Add a ci job that performs an OOT build with tests
  * Part of the existing job (setup logic) has been extracted into a custom [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) in order to share it between the two jobs.

Both jobs run in parallel, on separate git clones of the project. I found no way to make them work on the same clone without making them entirely sequential.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -2,7 +2,7 @@ name: "Setup build environment"
 description: "Setup the build environment. An action so that it can be shared between in-tree/out-of-tree jobs"
 
 inputs:
-  cache-key:
+  cache-suffix:
     description: |
       Additional string that is used to compute the ccache hash.
       Different jobs running the action need distinct values for this key,
@@ -33,4 +33,4 @@ runs:
   - name: Ccache for C++ compilation
     uses: hendrikmuhs/ccache-action@4687d037e4d7cf725512d9b819137a3af34d39b3
     with:
-      key: ${{ runner.os }}-clangreleaseasserts-${{ steps.get-submodule-hash.outputs.hash }}-${{ inputs.cache-key }}
+      key: ${{ runner.os }}-clangreleaseasserts-${{ steps.get-submodule-hash.outputs.hash }}${{ inputs.cache-suffix }}
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -1,0 +1,28 @@
+name: "Setup build environment"
+description: "Setup the build environment. An action so that it can be shared between in-tree/out-of-tree jobs"
+
+runs:
+  using: "composite"
+  steps:
+  - name: Set up Python
+    uses: actions/setup-python@v2
+    with:
+      python-version: 3.9
+  - name: Install MLIR Python depends
+    run: |
+      python -m pip install -r $GITHUB_WORKSPACE/externals/llvm-project/mlir/python/requirements.txt
+    shell: bash
+  - name: Install PyTorch nightly depends
+    run: |
+      python -m pip install -r requirements.txt
+    shell: bash
+  - name: Install Ninja
+    uses: llvm/actions/install-ninja@55d844821959226fab4911f96f37071c1d4c3268
+  - name: Get Submodule Hash
+    id: get-submodule-hash
+    run: echo "::set-output name=hash::$(md5sum $(git submodule status))"
+    shell: bash
+  - name: Ccache for C++ compilation
+    uses: hendrikmuhs/ccache-action@4687d037e4d7cf725512d9b819137a3af34d39b3
+    with:
+      key: ${{ runner.os }}-clangreleaseasserts-${{ steps.get-submodule-hash.outputs.hash }}
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -1,6 +1,14 @@
 name: "Setup build environment"
 description: "Setup the build environment. An action so that it can be shared between in-tree/out-of-tree jobs"
 
+inputs:
+  cache-key:
+    description: |
+      Additional string that is used to compute the ccache hash.
+      Different jobs running the action need distinct values for this key,
+      but the content is irrelevant.
+    required: true
+
 runs:
   using: "composite"
   steps:
@@ -25,4 +33,4 @@ runs:
   - name: Ccache for C++ compilation
     uses: hendrikmuhs/ccache-action@4687d037e4d7cf725512d9b819137a3af34d39b3
     with:
-      key: ${{ runner.os }}-clangreleaseasserts-${{ steps.get-submodule-hash.outputs.hash }}
+      key: ${{ runner.os }}-clangreleaseasserts-${{ steps.get-submodule-hash.outputs.hash }}-${{ inputs.cache-key }}
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -16,7 +16,8 @@ on:
 
 jobs:
   build:
-    name: Build and Test and maybe Publish (Release Asserts)
+    name: Build and Test (Release Asserts)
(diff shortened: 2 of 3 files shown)


.github/actions/setup-build/action.yml
```diff
@@ -0,0 +1,28 @@
+name: "Setup build environment"
```
silvasean, line 1: nit: why setup-build/action.yml instead of setupBuild.yml?

.github/workflows/buildAndTest.yml
```diff
@@ -85,11 +66,10 @@ jobs:
           TORCH_MLIR_PYTHON_PACKAGE_VERSION="${{ github.event.inputs.python_package_version }}" \
           ./build_tools/build_python_wheels.sh
 
-    # If we were given a release_id, then upload the package we just built
-    # to the github releases page.
+    # upload the package we just built to the github releases page.
```
silvasean, line 69: Nit: use proper capitalization.

.github/workflows/buildAndTest.yml
```diff
@@ -99,10 +79,60 @@ jobs:
...
-      if: github.event.inputs.release_id != ''
       id: publish_release
+      if: github.event.inputs.release_id != ''
       uses: eregon/publish-release@v1
       env:
         GITHUB_TOKEN: ${{ secrets.WORKFLOW_INVOCATION_TOKEN }}
       with:
         release_id: ${{ github.event.inputs.release_id }}
+
+  build-out-of-tree:
```
silvasean, line 90: Do we really want to re-build everything twice on every commit and re-run all the tests?

I think that for now it would be good to do this as a scheduled job that runs post-commit or nightly/N-times-per-day.

README.md
```diff
@@ -65,23 +68,44 @@ cmake -GNinja -Bbuild \
...
+```shell
+  -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+```
+* Enabling LLD (links in seconds compared to minutes)
+```shell
+  -DCMAKE_EXE_LINKER_FLAGS_INIT="-fuse-ld=lld" -DCMAKE_MODULE_LINKER_FLAGS_INIT="-fuse-ld=lld" -DCMAKE_SHARED_LINKER_FLAGS_INIT="-fuse-ld=lld"
 # Use --ld-path= instead of -fuse-ld=lld for clang > 13
+```
 
+If you have built llvm-project separately in the directory `$LLVM_INSTALL_DIR`, you can also build the project *out-of-tree* using the following command as template:
```
silvasean, line 83: can you add a new heading here `## Building against a pre-built LLVM`

silvasean (changes requested):

.github/actions/setup-build/action.yml
oowekyala: I'm pretty sure I tried that and it didn't work. Actions need a file called [action.yml](https://docs.github.com/en/actions/creating-actions/about-custom-actions#types-of-actions). If you're referring to why the action isn't named in camel case, I think this is just the conventional style for github actions

.github/workflows/buildAndTest.yml
oowekyala: Thanks for your feedback, you're right that this could become a problem. I think I can change that to run on push (ie on merge). Another option would be to disable tests in this second build, since they're probably identical, and we're just testing whether cmake builds everything correctly. IMHO, keeping the build running in PRs provides for better feedback, as otherwise you could find out a change breaks main only after it was merged. Wdyt?

silvasean: I think if we can reduce the time taken to just building Torch-MLIR (with the LLVM build cached almost always) and running the check-torch-mlir-all, then it would be okay to run on every PR.

oowekyala: There was a caching problem with the PR at the time you reviewed, which is fixed in [c579ff0](https://github.com/llvm/torch-mlir/pull/738/commits/c579ff0ec42c6df9946aa955a88d1de5bb9fff9d). See eg https://github.com/ljfitz/torch-mlir/actions/runs/2162118862 for a trial run, which shows that the total build time is roughly equivalent to the current build time. For the in-tree job, it should be identical. The out-of-tree job takes a couple of minutes longer but runs in parallel.

oowekyala: Note that the first out-of-tree build on these commits will probably take > an hour because it's running without a cache

silvasean:
Hi @oowekyala -- is this ready for me to review again?

powderluv:
Just want to say thank you for this PR. It is a great time saver.

oowekyala:
> Hi @oowekyala -- is this ready for me to review again?

Yes I think so

powderluv:
I am trying to add OSX support once this lands, should we expand what is in `build_python_wheels.sh`  into the GH actions themselves ?  

We also may want to leave ccache optional.

.github/workflows/buildAndTest.yml
```diff
@@ -16,33 +16,16 @@ on:
 
 jobs:
   build:
-    name: Build and Test (Release Asserts)
+    name: Build and Test and maybe Publish (Release Asserts)
```
silvasean, line 19: I believe this name is load bearing and needs to be updated in https://github.com/llvm/torch-mlir/blob/main/.github/workflows/releaseSnapshotPackage.yml

.github/workflows/buildAndTest.yml
```diff
@@ -106,3 +89,55 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+    - name: Get torch-mlir
+      uses: actions/checkout@v2
+      with:
+        submodules: 'true'
+    - uses: ./.github/actions/setup-build
+      with:
+        cache-suffix: '-out-of-tree'
+    - name: Build LLVM (standalone)
```
silvasean, line 104: Maybe comment here that this is expected to almost always be cached.

.github/workflows/buildAndTest.yml
```diff
@@ -106,3 +89,55 @@ jobs:
...
+          -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+          -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ \
+          -DMLIR_DIR="$(pwd)/llvm-build/lib/cmake/mlir/" \
+          -DLLVM_DIR="$(pwd)/llvm-build/lib/cmake/llvm/" \
+          -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
+          -DPython3_EXECUTABLE=$(which python) \
+          .
+        ninja -Cbuild check-torch-mlir-all
+
+    - name: RefBackend - TorchScript end-to-end tests
```
silvasean, line 134: I don't think we need to run the e2e tests. check-torch-mlir-all should cover all the degrees of freedom that could vary between in-tree and out-of-tree.

silvasean (changes requested):

.github/workflows/buildAndTest.yml
oowekyala: I renamed it back to the original name. I don't think the new name is a significant improvement, and I don't want to stir things too much

silvasean:
Is this ready to land from your end @oowekyala ?

powderluv:
Lets push and iterate on this. I will kick off a build

silvasean:
In the future please squash to avoid polluting the commit history with incremental local changes.

oowekyala:
> In the future please squash to avoid polluting the commit history with incremental local changes.

@silvasean sorry about that, will do.

Happy to see this merged!

powderluv:
Ah my bad. Will squash in the future. 

Thanks for the contribution

PR #739, opened 2022-04-06
[MLIR][TORCH] Add E2E support for aten::max_pool2d_with_indices_backward op

vivekkhandelwal1:
This commit adds lowering of `aten::max_pool2d_with_indices_backward` op.

### DIFF
--- a/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
+++ b/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
@@ -305,6 +305,222 @@ public:
 };
 } // namespace
 
+namespace {
+// The original implementation of the op is as follows:
+//
+// Indices and GradOutput Layout: [N, C, H, W] or [C, H, W]
+// Input Layout: [N, C, Hin, Win] or [C, Hin, Win]
+//
+// for i in range(N):
+//   for j in range(C):
+//       for k in range(H):
+//           for l in range(W):
+//               index = indices[i, j, k, l]
+//               result[i, j, index/Win, index%Win] += gradOutput[i, j, k, l]
+//
+//                    OR
+//
+// for i in range(C):
+//   for j in range(H):
+//       for k in range(W):
+//           index = indices[i, j, k]
+//           result[i, index/Win, index%Win] += gradOutput[i, j, k]
+//
+class ConvertAtenMaxPool2dWithIndicesBackwardOp
+    : public OpConversionPattern<AtenMaxPool2dWithIndicesBackwardOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenMaxPool2dWithIndicesBackwardOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    Location loc = op.getLoc();
+    MLIRContext *context = op->getContext();
+    Value gradOutput = adaptor.grad_output();
+    Value input = adaptor.self();
+    RankedTensorType gradOutputType =
+        gradOutput.getType().cast<RankedTensorType>();
+    Type gradOutputElemType = gradOutputType.getElementType();
+    RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+    Type inputElemType = inputType.getElementType();
+    int64_t tensorOperandRank = inputType.getRank();
+
+    // `TMTensor::ScatterOp` expects indices of element type i32.
+    Value indices = convertTensorToDtype(
+        rewriter, loc, op.indices(),
+        mlir::IntegerType::get(context, 32, mlir::IntegerType::Signed));
+    indices = typeConverter->materializeTargetConversion(
+        rewriter, loc, typeConverter->convertType(indices.getType()), indices);
+    RankedTensorType indicesType = indices.getType().cast<RankedTensorType>();
+    Type indicesElemType = indicesType.getElementType();
+
+    // The element type of the `input` and `grad_output` should be same.
+    if (inputElemType != gradOutputElemType)
+      return rewriter.notifyMatchFailure(
+          op,
+          "Input element type should be same as the grad_output element type.");
+
+    // Since the scatter op requires indices to be a 2-d tensor, we create a new
+    // 5-d/4-d tensor (depending on the original indices layout) comprising the
+    // index values. We will collapse this tensor into a 2-d tensor. The
+    // algorithm for the creation of updated indices tensor is as follows:
+    //
+    // for i in range(N):
+    //   for j in range(C):
+    //     for k in range(H):
+    //       for l in range(W):
+    //         for m in range(4):
+    //           if m == 0:
+    //             updatedIndices[N][C][H][W][0] = i
+    //           if m == 1:
+    //             updatedIndices[N][C][H][W][1] = j
+    //           if m == 2:
+    //             updatedIndices[N][C][H][W][2] =
+    //                                      originalIndices[i, j, k, l] / Win
+    //           if m == 3:
+    //             updatedIndices[N][C][H][W][3] =
+    //                                      originalIndices[i, j, k, l] % Win
+    //
+    //                  OR
+    //
+    //  for j in range(C):
+    //   for k in range(H):
+    //     for l in range(W):
+    //       for m in range(3):
+    //         if m == 0:
+    //           updatedIndices[C][H][W][0] = i
+    //         if m == 1:
+    //          updatedIndices[C][H][W][1] = originalIndices[i, j, k, l] / Win
+    //         if m == 2:
+    //           updatedIndices[C][H][W][2] = originalIndices[i, j, k, l] % Win
+
+    SmallVector<Value> inputShape = getTensorSizes(rewriter, loc, input);
+
+    SmallVector<AffineExpr> originalIndicesDimExprs, updatedIndicesDimExprs;
+    for (int64_t i = 0; i < tensorOperandRank; i++) {
+      originalIndicesDimExprs.push_back(rewriter.getAffineDimExpr(i));
+      updatedIndicesDimExprs.push_back(rewriter.getAffineDimExpr(i));
+    }
+    updatedIndicesDimExprs.push_back(
+        rewriter.getAffineDimExpr(tensorOperandRank));
+
+    SmallVector<AffineMap> indexingMaps = AffineMap::inferFromExprList(
+        {originalIndicesDimExprs, updatedIndicesDimExprs});
+    SmallVector<StringRef> iteratorTypes(tensorOperandRank + 1, "parallel");
+
+    SmallVector<OpFoldResult> updatedIndicesShape =
+        getAsOpFoldResult(getTensorSizes(rewriter, loc, indices));
+    updatedIndicesShape.push_back(rewriter.getIndexAttr(tensorOperandRank));
+
+    Value initTensor = rewriter.create<linalg::InitTensorOp>(
+        loc, updatedIndicesShape, indicesElemType);
+
+    Value wIn = inputShape[tensorOperandRank - 1];
+    SmallVector<Value> cstValues;
+    for (int64_t i = 0; i < tensorOperandRank; i++)
+      cstValues.push_back(rewriter.create<arith::ConstantIndexOp>(loc, i));
+
+    Value updatedIndices =
+        rewriter
+            .create<linalg::GenericOp>(
+                loc, initTensor.getType(), indices, initTensor, indexingMaps,
+                iteratorTypes,
+                [tensorOperandRank, wIn, cstValues,
+                 indicesElemType](OpBuilder &b, Location loc, ValueRange args) {
+                  Value index = castIntToIndex(b, loc, args[0]);
+                  Value updatedIndex = cstValues[0];
+                  Value lastDim =
+                      b.create<linalg::IndexOp>(loc, tensorOperandRank);
+
+                  for (int64_t i = tensorOperandRank - 1; i >= 0; i--) {
+                    Value result;
+                    if (i == tensorOperandRank - 1)
+                      result = b.create<arith::RemSIOp>(loc, index, wIn);
+                    if (i == tensorOperandRank - 2)
+                      result = b.create<arith::FloorDivSIOp>(loc, index, wIn);
+                    if (i == tensorOperandRank - 3 ||
+                        i == tensorOperandRank - 4)
+                      result = b.create<linalg::IndexOp>(loc, i);
+
+                    Value pred = b.create<arith::CmpIOp>(
+                        loc, arith::CmpIPredicate::eq, lastDim, cstValues[i]);
+                    Value addAmount = b.create<arith::SelectOp>(
+                        loc, pred, result, cstValues[0]);
+                    updatedIndex =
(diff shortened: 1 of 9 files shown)


vivekkhandelwal1:
The result is turning out to be all zeros. I can't find an issue with the algorithm. If someone else finds a problem, please do comment. 

Error message received:
```
****** Failed tests - 6 tests
    FAIL - "MaxPool2dWithIndicesBackwardModule_basic"
        @ output of call to "forward"
        ERROR: value (Tensor with shape=[1, 4, 4] min=+0.0, max=+0.0, mean=+0.0) is not close to golden value (Tensor with shape=[1, 4, 4] min=+0.0, max=+1.485, mean=+0.09281)

```

cathyzhyi:
Can you fix the commit and PR title to be consistent with the PR?

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -228,6 +228,90 @@ class ConvertAtenAdaptiveAvgPool2dOp
...
+          resultIndices.push_back(
+              b.create<arith::FloorDivSIOp>(loc, index, wIn));
+          resultIndices.push_back(b.create<arith::RemUIOp>(loc, index, wIn));
+          Value temp = b.create<tensor::ExtractOp>(loc, result, resultIndices);
+          if (gradOutputType.getElementType().isa<mlir::FloatType>())
+            out = b.create<arith::AddFOp>(loc, temp, out);
+          else
+            out = b.create<arith::AddIOp>(loc, temp, out);
+
+          b.create<tensor::InsertOp>(loc, out, result, resultIndices);
```
cathyzhyi, line 304: `tensor::InsertOp` can't be used here because this wont update the result tensor inplace but return a new tensor. See https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorinsert-mlirtensorinsertop. `It returns a copy of dest with the proper slice updated with the value of scalar.` I think Scatter op should be used instead.

cathyzhyi (changes requested):

vivekkhandelwal1:
I have to add comments to the code. I will do that and push the updated code.

lib/Conversion/TorchToLinalg/Pooling.cpp
vivekkhandelwal1: I have made the changes to the code accordingly to use the scatter op.

vivekkhandelwal1:
> The result is turning out to be all zeros. I can't find an issue with the algorithm. If someone else finds a problem, please do comment.
> 
> Error message received:
> 
> ```
> ****** Failed tests - 6 tests
>     FAIL - "MaxPool2dWithIndicesBackwardModule_basic"
>         ERROR: value (Tensor with shape=[1, 4, 4] min=+0.0, max=+0.0, mean=+0.0) is not close to golden value (Tensor with shape=[1, 4, 4] min=+0.0, max=+1.485, mean=+0.09281)
> ```

I have changed the algorithm. Not getting this issue anymore.

vivekkhandelwal1:
> I have to add comments to the code. I will do that and push the updated code.

I have added the comments and updated the code.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -305,6 +305,239 @@ class ConvertValsemVariantAtenIndexPutImplOp
...
+    // factor in the collapsed dimensions of the tensor. The algorithm for the
+    // updation of indices is as follows:
+    //
+    // for i in range(N):
+    //   for j in range(C):
+    //     for k in range(H):
+    //         for l in range(W):
+    //             updatedIndices[i * (C*H*W) + j * (H*W) + k * (W) + l] =
+    //             originalIndices[i, j, k, l] + j * (Hin*Win) + i * (C*Hin*Win)
+    //
```
cathyzhyi, line 394: it might be easier just convert the indices to the form expected by the scatter op:
```
for i in range(N):
   for j in range(C):
      for k in range(H):
         for l in range(W):
            for m in range(4):
               m == 0:
                  updateIndices[i * (C*H*W) + j * (H*W) + k * (W) + l][0] = i
               m == 1:
                  updateIndices[i * (C*H*W) + j * (H*W) + k * (W) + l][1] = j
               m == 2:
                  updateIndices[i * (C*H*W) + j * (H*W) + k * (W) + l][2] = originalIndices[i, j, k, l] / Hin
               m == 3:
                  updateIndices[i * (C*H*W) + j * (H*W) + k * (W) + l][3] = originalIndices[i, j, k, l] % Hin
```
In this way the scatter op can be used directly without expanding the result in the end.

cathyzhyi (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -305,6 +305,266 @@ class ConvertValsemVariantAtenIndexPutImplOp
...
+                  Value predTwo = b.create<arith::CmpIOp>(
+                      loc, arith::CmpIPredicate::eq, lastDim, cstIndexTwo);
+                  Value predThree = b.create<arith::CmpIOp>(
+                      loc, arith::CmpIPredicate::eq, lastDim, cstIndexThree);
+
+                  Value outIndexDimZero = b.create<arith::SelectOp>(
+                      loc, predZero, b.create<linalg::IndexOp>(loc, 0),
+                      cstIndexZero);
+
+                  if (tensorOperandRank == 4) {
```
cathyzhyi, line 476: the 2 if branches are very similar, can you try to refactor them to deduplicate the code? Something like:
```
sum = 0
for i in Rank-1..0:
  pred = pred(cst(i))
  if (i == Rank-1)
      result = index%Win
  if (i == Rank-2)
      result = index/Win
  if (i == Rank-3 || i == Rank-4)
      result = linalg.index(i)
  addAmount = select(pred, amount, 0)
  sum += addAmount
```

cathyzhyi (changes requested):

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -1664,3 +1664,124 @@ def forward(self, x):
 @register_test_case(module_factory=lambda: ToCopyWithDTypeFalsePinMemoryModule())
 def ToCopyWithDTypeFalsePinMemoryModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 2, 4))
+
+# ==============================================================================
+
+class MaxPool2dWithIndicesBackwardModule(torch.nn.Module):
```
cathyzhyi, line 1670: nit: `MaxPool2dWithIndicesBackwardStatic4DModule` to match `MaxPool2dWithIndicesBackwardDynamic4DModule` below.

cathyzhyi, line 1670: nit: `MaxPool2dWithIndicesBackwardStatic4DModule` to match `MaxPool2dWithIndicesBackwardDynamic4DModule` below.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -1664,3 +1664,124 @@ def forward(self, x):
...
+       dilation=[1, 1]
+       ceil_mode=False
+       return torch.ops.aten.max_pool2d_with_indices_backward(output, input, kernel_size, stride, padding, dilation, ceil_mode, indices)
+
+@register_test_case(module_factory=lambda: MaxPool2dWithIndicesBackwardModule())
+def MaxPool2dWithIndicesBackwardModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(2, 4, 7, 6), tu.rand(2, 4, 6, 5), torch.randint(16, (2, 4, 7, 6)))
+
+
+class MaxPool2dWithIndicesBackwardSingleChannelModule(torch.nn.Module):
```
cathyzhyi, line 1694: I am not sure we need to test this particular case. Is there any code path that's covered by this test but not covered by the Multiple channel below?

cathyzhyi, line 1694: I am sure we need to test this particular case.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -1664,3 +1664,124 @@ def forward(self, x):
...
+       dilation=[1, 1]
+       ceil_mode=False
+       return torch.ops.aten.max_pool2d_with_indices_backward(output, input, kernel_size, stride, padding, dilation, ceil_mode, indices)
+
+@register_test_case(module_factory=lambda: MaxPool2dWithIndicesBackwardSingleChannelModule())
+def MaxPool2dWithIndicesBackwardSingleChannelModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(1, 2, 2), tu.rand(1, 4, 4), torch.randint(16, (1, 2, 2)))
+
+
+class MaxPool2dWithIndicesBackwardMultipleChannelModule(torch.nn.Module):
```
cathyzhyi, line 1718: nit: `MaxPool2dWithIndicesBackwardStatic3DModule`

cathyzhyi, line 1718: nit: `MaxPool2dWithIndicesBackwardStatic3DModule`

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -1664,3 +1664,124 @@ def forward(self, x):
...
+
+    @export
+    @annotate_args([
+        None,
+        ([2, 4, 7, 6], torch.float32, True),
+        ([2, 4, 6, 5], torch.float32, True),
+        ([2, 4, 7, 6], torch.int64, True),
+    ])
+    def forward(self, output, input, indices):
+       kernel_size=[2, 2]
```
cathyzhyi, line 1682: the indentations for all the `def forward` are off.

cathyzhyi, line 1682: the indentations for all the `def forward` are off.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -1664,3 +1664,124 @@ def forward(self, x):
...
+       stride=[1, 1]
+       padding=[1, 1]
+       dilation=[1, 1]
+       ceil_mode=False
+       return torch.ops.aten.max_pool2d_with_indices_backward(output, input, kernel_size, stride, padding, dilation, ceil_mode, indices)
```
cathyzhyi, line 1687: nit:  you could use `yapf --in-place --recursive --style="{indent_width: 4}"` to check the format.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -305,6 +305,232 @@ class ConvertValsemVariantAtenIndexPutImplOp
...
+    SmallVector<AffineMap> indexingMaps = AffineMap::inferFromExprList(
+        {originalIndicesDimExprs, updatedIndicesDimExprs});
+    SmallVector<StringRef> iteratorTypes(tensorOperandRank + 1, "parallel");
+
+    SmallVector<int64_t> updatedIndicesStaticShape(tensorOperandRank, -1);
+    updatedIndicesStaticShape.push_back(tensorOperandRank);
+    Value initTensor = rewriter.create<linalg::InitTensorOp>(
+        loc, indicesShape, updatedIndicesStaticShape, indicesElemType);
+
+    Value cstWin = inputShape[tensorOperandRank - 1];
```
cathyzhyi, line 419: nit: `Win` since it's not constant.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -305,6 +305,232 @@ class ConvertValsemVariantAtenIndexPutImplOp
...
+
+                  updatedIndex = b.create<arith::IndexCastOp>(
+                      loc, indicesElemType, updatedIndex);
+                  b.create<linalg::YieldOp>(loc, updatedIndex);
+                })
+            .getResult(0);
+
+    // Creating a new tensor initialized with zeros and size same as the input
+    // tensor.
+    Value output =
```
cathyzhyi, line 463: there is a `createZeroInitTensor` helper for this.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -305,6 +305,232 @@ class ConvertValsemVariantAtenIndexPutImplOp
...
+                [tensorOperandRank, cstWin, cstValues,
+                 indicesElemType](OpBuilder &b, Location loc, ValueRange args) {
+                  Value index = castIntToIndex(b, loc, args[0]);
+                  Value updatedIndex = cstValues[0];
+                  Value lastDim =
+                      b.create<linalg::IndexOp>(loc, tensorOperandRank);
+
+                  for (int64_t i = tensorOperandRank - 1; i >= 0; i--) {
+                    Value result;
+                    if (i == tensorOperandRank - 1)
```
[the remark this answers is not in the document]
vivekkhandelwal1: Switch-case can't be used here because it expects the condition to be a constant expression. While the condition here is a non-const expression for ex: `tensorOperandRank - 1`, `tensorOperandRank - 2`, etc.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -305,6 +305,232 @@ class ConvertValsemVariantAtenIndexPutImplOp
...
+    }
+    updatedIndicesDimExprs.push_back(
+        rewriter.getAffineDimExpr(tensorOperandRank));
+
+    SmallVector<AffineMap> indexingMaps = AffineMap::inferFromExprList(
+        {originalIndicesDimExprs, updatedIndicesDimExprs});
+    SmallVector<StringRef> iteratorTypes(tensorOperandRank + 1, "parallel");
+
+    SmallVector<int64_t> updatedIndicesStaticShape(tensorOperandRank, -1);
+    updatedIndicesStaticShape.push_back(tensorOperandRank);
```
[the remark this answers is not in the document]
vivekkhandelwal1: I think it's needed because:
a.) The `updatedIndices` tensor is of Rank = rank(indices) + 1
b.) We need to create the tensor with the static sizes when the original indices are also statically sized; otherwise, it results in an issue with the collapsing at the later point.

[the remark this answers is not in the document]
cathyzhyi: I see. Then you could use:
```
SmallVector<OpFoldResult> updatedIndicesShape =
       getAsOpFoldResult(getTensorSizes(rewriter, loc, indices));
auto cstTensorOperandRank = rewriter.getIndexAttr(tensorOperandRank);
updatedIndicesShape.push_back(cstTensorOperandRank);
Value initTensor = rewriter.create<linalg::InitTensorOp>(
       loc, updatedIndicesDynamicShape, indicesElemType);
```
Putting the dynamic dims and static dims into one array makes the code cleaner.

python/torch_mlir_e2e_test/test_suite/basic.py
cathyzhyi: have you tried with yapf? I think the formatter would format this part as:
```
+        kernel_size = [2, 2]
+        stride = [1, 1]
+        padding = [1, 1]
+        dilation = [1, 1]
+        ceil_mode = False
+        return torch.ops.aten.max_pool2d_with_indices_backward(
+            output, input, kernel_size, stride, padding, dilation, ceil_mode,
+            indices)
```
Can you update the format accordingly?

cathyzhyi (approved):

PR #740, opened 2022-04-06
Fix misuse of `normalize_function` api

makslevental:
Fixes a bug in use of upstream `normalize_function` in our `normalize_args_kwargs` (in eager mode) and introduces unit tests.

NB: `shouldnt_normalize` currently fails i.e., args *will* successfully normalize despite being incorrect due to an [upstream bug](https://github.com/pytorch/pytorch/issues/75342).

### DIFF
--- a/python/test/eager_mode/normalize_args_kwargs.py
+++ b/python/test/eager_mode/normalize_args_kwargs.py
@@ -1,0 +1,84 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+
+# RUN: %PYTHON %s | FileCheck %s
+
+
+import torch
+
+from torch_mlir.eager_mode.torch_mlir_dispatch import normalize_args_kwargs
+
+
+def run_test(test, XFAIL=False, XPASS=False):
+    try:
+        test()
+        print(("X" if XPASS else "") + f"PASS - {test.__name__}")
+    except Exception as e:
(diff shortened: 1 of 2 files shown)


python/test/eager_mode/normalize_args_kwargs.py
```diff
@@ -0,0 +1,81 @@
...
+@register_test_case(module_factory=lambda: None)
+def shouldnt_normalize3(_, tu: TestUtils):
+    target = torch.ops.aten.max_pool2d_with_indices.default.overloadpacket
+    args = (tu.rand(1, 3, 32, 32),)
+    kwargs = {"padding": None}
+    normalize_args_kwargs(target, args, kwargs)
+
+
+def main():
+    config = EagerModeTestConfig()
```
silvasean, line 75: why are you using the e2e framework? You should be able to just run the code you want and print out any exceptions/etc that are of interest to check.

silvasean (changes requested):

python/test/eager_mode/normalize_args_kwargs.py
makslevental: because the e2e framework has niceties like `register_test_case`, which will capture the name of the method and print it when reporting, as well as TestUtils which makes rng deterministic. 

both of these things can be rebuilt but isn't it better to reuse? 

as well as the fact that the adjacent FileCheck tests also [use the e2e framework](https://github.com/llvm/torch-mlir/blob/1d5b5a89e8e9ac75e4837b158fc4230eb7388b28/python/test/torchscript_e2e_test/basic.py#L17).

silvasean: Yes, those are the tests of the e2e framework *itself* :)

You can just do

# CHECK-LABEL: Description of Test
# CHECK: AssertionError: ... or whatever...
print("Description of Test")
# Do the thing

PR #741, opened 2022-04-07
Add convolution_backward

gpetters94:
I'm still working on the nonzero padding case, and it's not very efficient, but here's the implementation for convolution_backward.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
@@ -151,6 +151,36 @@ m_TorchConstantIntList(SmallVectorImpl<int64_t> &bind_values) {
 }
 
 namespace detail {
+/// Matches the constant bools stored in a `torch.ListConstruct`.
+struct torch_bool_list_construct_op_binder {
+  SmallVectorImpl<bool> &bind_values;
+
+  /// Creates a matcher instance that binds the value to bvs if match succeeds.
+  torch_bool_list_construct_op_binder(SmallVectorImpl<bool> &bvs)
+      : bind_values(bvs) {}
+
+  bool match(Operation *op) {
+    auto listConstruct = dyn_cast<Torch::PrimListConstructOp>(op);
+    if (!listConstruct)
+      return false;
+    for (Value value : listConstruct.elements()) {
+      bool num;
+      if (matchPattern(value, m_TorchConstantBool(&num)))
+        bind_values.push_back(num);
+      else
+        return false;
+    }
+    return true;
+  }
+};
+} // namespace detail
+
+/// Matches the constant bools stored in a `torch.prim.ListConstruct`.
+inline detail::torch_bool_list_construct_op_binder
+m_TorchConstantBoolList(SmallVectorImpl<bool> &bind_values) {
+  return detail::torch_bool_list_construct_op_binder(bind_values);
+}
+namespace detail {
 /// Matches the expected tensor and dim from `torch.aten.size.int`.
 struct torch_tensor_size_int_op_binder {
   int64_t *dim;
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -646,8 +646,6 @@ void mlir::torch::torch_to_linalg::populateLinearPatternsAndLegality(
   patterns.add<ConvertAtenFlipOp>(typeConverter, context);
   target.addIllegalOp<AtenMatmulOp>();
   patterns.add<ConvertAtenMatmulOp>(typeConverter, context);
-  target.addIllegalOp<AtenFlipOp>();
-  patterns.add<ConvertAtenFlipOp>(typeConverter, context);
   target.addIllegalOp<AtenBmmOp>();
   patterns.add<ConvertAtenBmmOp>(typeConverter, context);
   target.addIllegalOp<AtenLinearOp>();
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -93,7 +93,6 @@ public:
 
 namespace {
 class ConvertAtenFlipOp : public OpConversionPattern<AtenFlipOp> {
-
 public:
   using OpConversionPattern::OpConversionPattern;
   LogicalResult
@@ -534,11 +533,6 @@ public:
       return rewriter.create<arith::IndexCastOp>(loc, intType, v);
     };
 
-    SmallVector<int64_t> paddingInts;
-    if (!matchPattern(op.padding(), m_TorchConstantIntList(paddingInts))) {
-      return rewriter.notifyMatchFailure(
-          op, "only support constant padding values");
-    }
     SmallVector<int64_t> strideInts;
     if (!matchPattern(op.stride(), m_TorchConstantIntList(strideInts)))
       return rewriter.notifyMatchFailure(op,
@@ -557,7 +551,7 @@ public:
     for (size_t i = 2; i < inRank; i++)
       weightDims.push_back(getDimOp(rewriter, loc, weight, i));
 
-    // Guard unused values (transposed, groups)
+    // Guard unused values (transposed, groups).
     int64_t group_size;
     if (!matchPattern(op.groups(), m_TorchConstantInt(&group_size)) ||
         group_size != 1)
@@ -570,14 +564,13 @@ public:
           op, "unimplemented: only non-transposed convolution supported");
 
     // Pad the input tensor according to padding.
-    SmallVector<int64_t, 4> paddingIncludingNC = {0, 0};
-    paddingIncludingNC.insert(paddingIncludingNC.end(), paddingInts.begin(),
-                              paddingInts.end());
-    Value paddedInput = torch_to_linalg::getZeroPaddedTensor(
-        op, rewriter, input, paddingIncludingNC);
+    SmallVector<Value> paddingIntValues;
+    getListConstructElements(adaptor.padding(), paddingIntValues);
+    paddingIntValues = getTypeConvertedValues(rewriter, loc, getTypeConverter(),
+                                              paddingIntValues);
+    Value paddedInput = torch_to_linalg::getDynamicZeroPaddedTensor(
+        op, rewriter, input, paddingIntValues, 2);
 
-    SmallVector<Value> paddingIntValues =
-        getAsConstantIntValues(rewriter, loc, paddingInts);
     SmallVector<Value> dilationIntValues =
         getAsConstantIntValues(rewriter, loc, dilationInts);
     SmallVector<Value> strideIntValues =
@@ -626,7 +619,7 @@ public:
     auto stridesAttr = rewriter.getI64VectorAttr(strideInts);
     auto dilationAttr = rewriter.getI64VectorAttr(dilationInts);
 
-    // TODO: add 1D and 3D case
+    // TODO: add 1D and 3D case.
     Value conv =
         rewriter
             .create<linalg::Conv2DNchwFchwOp>(
@@ -651,6 +644,8 @@ void mlir::torch::torch_to_linalg::populateLinearPatternsAndLegality(
   patterns.add<ConvertAtenFlipOp>(typeConverter, context);
   target.addIllegalOp<AtenMatmulOp>();
   patterns.add<ConvertAtenMatmulOp>(typeConverter, context);
+  target.addIllegalOp<AtenFlipOp>();
+  patterns.add<ConvertAtenFlipOp>(typeConverter, context);
   target.addIllegalOp<AtenBmmOp>();
   patterns.add<ConvertAtenBmmOp>(typeConverter, context);
   target.addIllegalOp<AtenLinearOp>();
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -533,6 +533,10 @@ public:
       return rewriter.create<arith::IndexCastOp>(loc, intType, v);
     };
 
+    SmallVector<Value> paddingIntValues;
+    if (!getListConstructElements(adaptor.padding(), paddingIntValues))
+      return rewriter.notifyMatchFailure(
+          op, "only support padding from a list construct");
     SmallVector<int64_t> strideInts;
     if (!matchPattern(op.stride(), m_TorchConstantIntList(strideInts)))
       return rewriter.notifyMatchFailure(op,
@@ -564,12 +568,10 @@ public:
           op, "unimplemented: only non-transposed convolution supported");
 
     // Pad the input tensor according to padding.
-    SmallVector<Value> paddingIntValues;
-    getListConstructElements(adaptor.padding(), paddingIntValues);
     paddingIntValues = getTypeConvertedValues(rewriter, loc, getTypeConverter(),
                                               paddingIntValues);
     Value paddedInput = torch_to_linalg::getDynamicZeroPaddedTensor(
-        op, rewriter, input, paddingIntValues, 2);
+        op, rewriter, input, paddingIntValues, /*unpaddedDims=*/2);
 
     SmallVector<Value> dilationIntValues =
         getAsConstantIntValues(rewriter, loc, dilationInts);
--- a/lib/Conversion/TorchToLinalg/Utils.cpp
+++ b/lib/Conversion/TorchToLinalg/Utils.cpp
@@ -77,14 +77,16 @@ Value torch_to_linalg::getDynamicZeroPaddedTensor(
     int unpaddedDims) {
   assert(input.getType().isa<RankedTensorType>() &&
          "input must be RankedTensorType");
+  unsigned int inRank = getTensorRank(input);
   Location loc = op->getLoc();
 
-  Value c1 = b.create<arith::ConstantOp>(loc, b.getIndexAttr(1));
   SmallVector<Value> inputDims = getTensorSizes(b, loc, input);
 
   Value c0 = b.create<arith::ConstantOp>(loc, b.getI64IntegerAttr(0));
-  SmallVector<Value> paddingIncludingUnchanged{c0, c0};
+  SmallVector<Value> paddingIncludingUnchanged(unpaddedDims, c0);
   paddingIncludingUnchanged.append(padding);
+  assert(unpaddedDims + padding.size() == inRank &&
+         "sum of unpaddedDims and padding.size() must equal to inputRank");
   for (auto pad = paddingIncludingUnchanged.begin();
        pad < paddingIncludingUnchanged.end(); pad++)
     *pad = castIntToIndex(b, loc, *pad);
@@ -94,15 +96,14 @@ Value torch_to_linalg::getDynamicZeroPaddedTensor(
   Type inputType = RankedTensorType::get(
       llvm::ArrayRef<int64_t>(SmallVector<int64_t>(inputRank, kUnknownSize)),
       elementType);
-  input = b.create<tensor::CastOp>(loc, inputType, input);
 
-  SmallVector<Value> strides(inputRank, c1);
   Value cf0 =
       b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
   SmallVector<OpFoldResult> paddingValues =
       getAsOpFoldResult(paddingIncludingUnchanged);
-  return tensor::createPadScalarOp(inputType, input, cf0, paddingValues,
-                                   paddingValues, false, loc, b);
+  return tensor::createPadScalarOp(inputType, input, cf0, /*low=*/paddingValues,
+                                   /*high=*/paddingValues, /*packing=*/false,
+                                   loc, b);
 }
 
 Value torch_to_linalg::getOutputDimForConvOps(OpBuilder &b, Location loc,
--- a/lib/Conversion/TorchToLinalg/Utils.cpp
+++ b/lib/Conversion/TorchToLinalg/Utils.cpp
@@ -16,6 +16,7 @@
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Tensor/Utils/Utils.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/IR/Matchers.h"
 #include "torch-mlir/Conversion/Utils/Utils.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchDialect.h"
@@ -98,8 +99,10 @@ Value torch_to_linalg::getDynamicZeroPaddedTensor(
   SmallVector<Value> strides(inputRank, c1);
   Value cf0 =
       b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
-  return createPadScalarOp(input, cf0, paddingIncludingUnchanged,
-                           paddingIncludingUnchanged, false, loc, b);
+  SmallVector<OpFoldResult> paddingValues =
+      getAsOpFoldResult(paddingIncludingUnchanged);
+  return tensor::createPadScalarOp(inputType, input, cf0, paddingValues,
+                                   paddingValues, false, loc, b);
 }
 
 Value torch_to_linalg::getOutputDimForConvOps(OpBuilder &b, Location loc,
@@ -290,22 +293,4 @@ Value torch_to_linalg::createElementwiseLinalgGeneric(
                                  /*outputs=*/initTensor, indexingMaps,
                                  iteratorTypes, bodyBuild)
       .getResult(0);
-}
-
-Value torch_to_linalg::createPadScalarOp(Value source, Value pad,
-                                         SmallVector<Value> low,
-                                         SmallVector<Value> high, bool nofold,
-                                         Location loc, OpBuilder &builder) {
-  auto padTensorOp =
-      builder.create<tensor::PadOp>(loc, source, low, high, nofold);
-  int rank = padTensorOp.getResultType().getRank();
-  SmallVector<Type> blockArgTypes(rank, builder.getIndexType());
-  SmallVector<Location> blockArgLocs(rank, loc);
-  auto &region = padTensorOp.region();
-  // `builder.createBlock` changes the insertion point within the block. Create
-  // a guard to reset the insertion point of the builder after it is destroyed.
-  OpBuilder::InsertionGuard guard(builder);
-  builder.createBlock(&region, region.end(), blockArgTypes, blockArgLocs);
-  builder.create<tensor::YieldOp>(loc, pad);
-  return padTensorOp;
 }
--- a/lib/Conversion/TorchToLinalg/Utils.cpp
+++ b/lib/Conversion/TorchToLinalg/Utils.cpp
@@ -22,6 +22,7 @@
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+#include <cstdint>
 
 using namespace mlir;
 using namespace mlir::torch;
@@ -65,6 +66,51 @@ Value torch_to_linalg::getZeroPaddedTensor(
       loc,
       b.getZeroAttr(input.getType().cast<RankedTensorType>().getElementType()));
   return getPaddedTensor(op, b, input, paddingInts, paddingInts, c0);
+}
+
+// Helper function that adds dynamic padding to a tensor, ignoring unpaddedDims
+// dimensions at the beginning. The high and low padding are the same, and the
+// padding value is zero.
+Value torch_to_linalg::getDynamicZeroPaddedTensor(
+    Operation *op, OpBuilder &b, Value &input, SmallVectorImpl<Value> &padding,
+    int unpaddedDims) {
+  assert(input.getType().isa<RankedTensorType>() &&
+         "input must be RankedTensorType");
+  Location loc = op->getLoc();
+
+  Value c1 = b.create<arith::ConstantOp>(loc, b.getIndexAttr(1));
+  SmallVector<Value> inputDims = getTensorSizes(b, loc, input);
+
+  Value c0 = b.create<arith::ConstantOp>(loc, b.getI64IntegerAttr(0));
+  SmallVector<Value> paddingIncludingUnchanged{c0, c0};
+  paddingIncludingUnchanged.append(padding);
+  for (auto pad = paddingIncludingUnchanged.begin();
+       pad < paddingIncludingUnchanged.end(); pad++)
+    *pad = castIntToIndex(b, loc, *pad);
+
+  Value c2 = b.create<arith::ConstantOp>(loc, b.getI64IntegerAttr(2));
+  auto inputRank = input.getType().cast<RankedTensorType>().getRank();
+  SmallVector<Value> outputDims{inputDims[0], inputDims[1]};
+  for (auto i = 0; i < inputRank; i++) {
+    if (i >= unpaddedDims) {
+      Value padDouble = b.create<arith::MulIOp>(loc, padding[i - 2], c2);
+      Value dim = b.create<arith::AddIOp>(loc, inputDims[i],
+                                          castIntToIndex(b, loc, padDouble));
+      outputDims.push_back(dim);
+    }
+  }
+
+  Type elementType = input.getType().cast<RankedTensorType>().getElementType();
+  Type inputType = RankedTensorType::get(
+      llvm::ArrayRef<int64_t>(SmallVector<int64_t>(inputRank, kUnknownSize)),
+      elementType);
+  input = b.create<tensor::CastOp>(loc, inputType, input);
+
+  SmallVector<Value> strides(inputRank, c1);
+  Value cf0 =
+      b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
+  return createPadScalarOp(input, cf0, paddingIncludingUnchanged,
+                           paddingIncludingUnchanged, false, loc, b);
 }
 
 Value torch_to_linalg::getOutputDimForConvOps(OpBuilder &b, Location loc,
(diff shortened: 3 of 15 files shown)


gpetters94:
I think I've figured out why nonzero padding isn't working. It looks like if the padding value is > 0, then it indicates how much padding grad_output already has. So I shouldn't need to fiddle with adding more padding to it. The question is whether the linalg convolution ops will handle arbitrarily padded tensors, or if I should be shrinking the grad_output to fit linalg's needs.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -448,6 +449,242 @@ class ConvertAtenLinearOp : public OpConversionPattern<AtenLinearOp> {
...
+    Value ci0 =
+        rewriter.create<arith::ConstantOp>(loc, rewriter.getI64IntegerAttr(0));
+    Value groupEqual1 = rewriter.create<arith::CmpIOp>(
+        loc, arith::CmpIPredicate::eq, groups, c1);
+    rewriter.create<cf::AssertOp>(
+        loc, groupEqual1, rewriter.getStringAttr("expect groups to be 1"));
+
+    // Pad the input tensor according to padding.
+    SmallVector<int64_t> paddingIncludingNC = {0, 0};
+    paddingIncludingNC.insert(paddingIncludingNC.end(), paddingInts.begin(),
```
ramiro050, line 523: nit: you can also use `append`:

https://llvm.org/doxygen/classllvm_1_1SmallVectorImpl.html#a0809c4018a9fd60e3149986d9f93c791

ramiro050 (commented):
Haven't looked at the whole thing, but here are a few comments for now

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -448,6 +449,242 @@ class ConvertAtenLinearOp : public OpConversionPattern<AtenLinearOp> {
...
+    SmallVector<Value> dilationIntValues =
+        getAsConstantIntValues(rewriter, loc, dilationInts);
+    SmallVector<Value> strideIntValues =
+        getAsConstantIntValues(rewriter, loc, strideInts);
+
+    auto stridesAttr = rewriter.getI64VectorAttr(strideInts);
+    auto dilationAttr = rewriter.getI64VectorAttr(dilationInts);
+
+    Value c1 =
+        rewriter.create<arith::ConstantOp>(loc, IntegerAttr::get(intType, 1));
```
ramiro050, line 511: nit: Here i would do the same thing as in `ci0`, use `getI64IntegerAttr` to make it clear that this value will always be `I64`

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -448,6 +449,242 @@ class ConvertAtenLinearOp : public OpConversionPattern<AtenLinearOp> {
...
+            rewriter, loc, lhDim, ci0, dilation, castIndexToInt(rhDim), stride);
+      }
+      // TODO: dilation, stride
+      Value c1 =
+          rewriter.create<arith::ConstantOp>(loc, rewriter.getIndexAttr(1));
+      Value res = rewriter.create<arith::AddIOp>(loc, lhDim, rhDim);
+      return rewriter.create<arith::SubIOp>(loc, res, c1);
+    };
+    auto convolve = [&](Value lhs, Value rhs, SmallVector<Value> &lhDims,
+                        SmallVector<Value> &rhDims, bool valid) -> Value {
```
ramiro050, line 547: what does `valid` represent here?

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -448,6 +449,242 @@ class ConvertAtenLinearOp : public OpConversionPattern<AtenLinearOp> {
...
+      SmallVector<Value> outDims{lhDims[0], rhDims[0]};
+      for (size_t i = 0; i < inRank - 2; i++)
+        outDims.push_back(
+            getConvOutputDim(lhDims[i + 2], rhDims[i + 2], paddingIntValues[i],
+                             strideIntValues[i], dilationIntValues[i], valid));
+
+      Value initTensor =
+          rewriter.create<linalg::InitTensorOp>(loc, outDims, elementType);
+
+      Value biasInitTensor;
```
ramiro050, line 562: nit: unnecessary line

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -448,6 +449,242 @@ class ConvertAtenLinearOp : public OpConversionPattern<AtenLinearOp> {
...
+      SmallVector<AffineExpr> exprs;
+      for (size_t i = 0; i < tSizes.size(); i++)
+        exprs.push_back(rewriter.getAffineDimExpr(i));
+      SmallVector<AffineMap> indexingMaps =
+          AffineMap::inferFromExprList({exprs, exprs});
+      Value outTensor = rewriter.create<linalg::InitTensorOp>(
+          loc, tSizes, tType.getElementType());
+
+      return rewriter
+          .create<linalg::GenericOp>(
```
ramiro050, line 599: nit: if you move this function

https://github.com/llvm/torch-mlir/blob/9ec0683e928bdd1acea9aeb025ed542332fab902/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L38

to one of the utils files, you can use it to avoid writing all the `iteratorTypes`, `indexingMaps`, ... boilerplate code

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -448,6 +449,242 @@ class ConvertAtenLinearOp : public OpConversionPattern<AtenLinearOp> {
...
+      arrangement.push_back(i);
+    std::iter_swap(arrangement.begin(), arrangement.begin() + 1);
+
+    weight = rearrange(weight, arrangement);
+    std::iter_swap(weightDims.begin(), weightDims.begin()+1);
+    Value resInput =
+        convolve(grad, rotate(weight, 2, inRank), gradDims, weightDims, false);
+
+    // Convolution between input and grad_output (weight gradient)
+    std::iter_swap(inputDims.begin(), inputDims.begin()+1);
```
ramiro050, line 656: nit: spacing in `begin()+1` is not consistent with the rest of the code

lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -67,6 +67,21 @@ Value torch_to_linalg::getPaddedTensor(Operation *op, OpBuilder &b,
...
+  Location loc = op->getLoc();
+  SmallVector<OpFoldResult> lowPaddings =
+      getIndexIntsAsOpFoldResult(b, lowPaddingInts);
+  SmallVector<OpFoldResult> highPaddings =
+      getIndexIntsAsOpFoldResult(b, highPaddingInts);
+  Value paddedInput = tensor::createPadScalarOp(
+      outType, input, pad, /*low=*/lowPaddings, /*high=*/highPaddings,
+      /*packing=*/false, loc, b);
+  return paddedInput;
+}
```
ramiro050, line 83: This is almost the exact same code as `getPaddedTensor` above. Can you write one in terms of the other?

ramiro050 (commented):
A few more comments

lib/Conversion/TorchToLinalg/Linear.cpp
gpetters94: Whether the convolution is valid (i.e. unpadded)

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -448,6 +449,242 @@ class ConvertAtenLinearOp : public OpConversionPattern<AtenLinearOp> {
...
+                                         "only support constant int strides");
+    SmallVector<int64_t> dilationInts;
+    if (!matchPattern(op.dilation(), m_TorchConstantIntList(dilationInts)))
+      return rewriter.notifyMatchFailure(op,
+                                         "only support constant int dilations");
+
+    for (size_t i = 0; i < paddingInts.size(); i++) {
+      if (paddingInts[i] != 0)
+        return rewriter.notifyMatchFailure(op, "only support 0 padding");
+      paddingInts[i] += 1;
```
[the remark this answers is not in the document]
gpetters94: It's unnecessary now, since I'm going to need to rewrite some of this implementation.

gpetters94:
Okay, here's the new and improved implementation. It looks like it's working with both padded and unpadded input. It could probably do with some more checks, though.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,126 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+        loc, ListType::get(IntType::get(context)),
+        SmallVector<Value>(gradRank - 2, ci1));
+    Value gradInput = rewriter.create<Torch::AtenConvolutionOp>(
+        loc, op.getResultTypes()[0], gradOutput, weightTransposed, cNone,
+        op.stride(), gradInputPadding, op.dilation(), op.transposed(),
+        op.output_padding(), op.groups());
+
+    // Shrink input_grad
+    for (auto i = 2; i < gradRank; i++) {
+      if (paddingInts[i - 2] != 0) {
```
cathyzhyi, line 865: nit: this can be easily checked at runtime with cmp and select. Then we can remove the restriction for int constant.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,126 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+
+    // Transpose inputs
+    Value gradOutputTransposed = rewriter.create<Torch::AtenTransposeIntOp>(
+        loc, gradOutput.getType(), gradOutput, ci0, ci1);
+    Value inputTransposed = rewriter.create<Torch::AtenTransposeIntOp>(
+        loc, input.getType(), input, ci0, ci1);
+    Value weightTransposed = rewriter.create<Torch::AtenTransposeIntOp>(
+        loc, weight.getType(), weight, ci0, ci1);
+
+    // Convolve grad_output with weight (padding = [1, 1])
```
cathyzhyi, line 854: I am not familiar with the algorithm. Is the padding always `1,1` for all cases here?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,126 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+    MLIRContext *context = op.getContext();
+    Value ci0 = rewriter.create<Torch::ConstantIntOp>(
+        loc, rewriter.getI64IntegerAttr(0));
+    Value ci1 = rewriter.create<Torch::ConstantIntOp>(
+        loc, rewriter.getI64IntegerAttr(1));
+    Value cNone = rewriter.create<Torch::ConstantNoneOp>(loc);
+    Value cFalse = rewriter.create<Torch::ConstantBoolOp>(
+        loc, rewriter.getBoolAttr(false));
+
+    Value gradOutput = op.grad_output();
```
cathyzhyi, line 821: how is `output_mask` used in the calculation? Do we need check against all true?

cathyzhyi (changes requested):
Thanks for the change. Much easier to follow than the previous version!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,126 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+    if (!matchPattern(op.padding(), m_TorchConstantIntList(paddingInts)))
+      return rewriter.notifyMatchFailure(op,
+                                         "only constant padding is supported");
+
+    if (gradRank != 4)
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only 2D convolutions supported");
+
+    // Rotate weight
+    SmallVector<Value> axis;
```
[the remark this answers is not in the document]
gpetters94: I think the plural of "axis" is "axes" - is that name okay?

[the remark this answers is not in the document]
cathyzhyi: ha, I see. yea, that's fine.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gpetters94: Output mask determines what we return. If any value is `False`, then that part of the return tuple is `None`. Is returning `None` supported in torch-mlir? If so, I can just do a runtime check and return the appropriate value.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gpetters94: I'm familiar with cmp and select in arith, but not in Torch. What ops do we use for that?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
cathyzhyi: ha, I see. I think `None` is currently not supported. We could add a `m_TorchConstantBoolList` pattern and check against it for now.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gpetters94: Doing some further testing, it looks like it isn't. For stride=1 and dilation=1 it looks like it's the size of the weight - 1. For other values of stride and dilation I'm not sure what it's supposed to be. I'll research further in depth but for now I'm going to add an unsupported check for dilation>1 and stride>1 and put in weight size -1.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,144 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+      return rewriter.notifyMatchFailure(op,
+                                         "only constant padding is supported.");
+    if (gradRank != 4)
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only 2D convolutions supported.");
+    SmallVector<bool> outMask;
+    if (!matchPattern(op.output_mask(), m_TorchConstantBoolList(outMask)))
+      return rewriter.notifyMatchFailure(
+          op, "only constant output_mask is supported.");
+    bool outputAll = true;
```
cathyzhyi, line 842: you can use `if(llvm::all_of(outMask, [](bool mask){return mask;}))` to check for all true value.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
cathyzhyi: And a comment and a // TODO here would be good.

python/torch_mlir_e2e_test/test_suite/backprop.py
```diff
@@ -56,6 +56,84 @@ def TanhBackward_basic(module, tu: TestUtils):
...
+
+   @export
+   @annotate_args([
+       None,
+       ([-1, -1, -1], torch.float32, True),
+       ([-1, -1, -1], torch.float32, True),
+       ([-1, -1, -1], torch.float32, True),
+   ])
+   def forward(self, grad_out, input_vec, weight):
+       return torch.ops.aten.convolution_backward(grad_out, input_vec, weight, bias_sizes=None, stride=[1], padding=[0], dilation=[1], transposed=False, output_padding=[0], groups=1, output_mask=[True, True, True])
```
cathyzhyi, line 71: can you use ` yapf --in-place --recursive --style="{indent_width: 4}"` to format the added test cases.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,144 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+        loc, op.getResultTypes()[0], gradOutput, weightTransposed, cNone,
+        op.stride(), gradInputPadding, op.dilation(), op.transposed(),
+        op.output_padding(), op.groups());
+
+    // Shrink input_grad
+    for (auto i = 2; i < gradRank; i++) {
+      if (paddingInts[i - 2] != 0) {
+        Value dim = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(i));
+        // (input_grad.size()[x] - input_vec.size()[x])//2
```
cathyzhyi, line 882: is this part for calculating the padding? Is that just padding[i]?

gpetters94: Nope, I realized that I don't have to fiddle with the padding. I can just resize it to the proper size and it should work.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gpetters94: As per my comment below, I'm adding an explanation and keeping it as-is.

python/torch_mlir_e2e_test/test_suite/backprop.py
cathyzhyi: Sry, I meant all the tests with in the current file.

gpetters94:
Okay, I've addressed the changes and rebased to main. Let me know if I'm good to merge.

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
```diff
@@ -150,6 +150,36 @@ m_TorchConstantIntList(SmallVectorImpl<int64_t> &bind_values) {
...
+        bind_values.push_back(num);
+      else
+        return false;
+    }
+    return true;
+  }
+};
+} // namespace detail
+
+/// Matches the constant integers stored in a `torch.prim.ListConstruct`.
```
cathyzhyi, line 178: `integers` => `bools`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,142 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+        loc, op.getResultTypes()[0], gradOutput, weightTransposed, cNone,
+        op.stride(), gradInputPadding, op.dilation(), op.transposed(),
+        op.output_padding(), op.groups());
+
+    // Shrink input_grad
+    for (auto i = 2; i < gradRank; i++) {
+      if (paddingInts[i - 2] > 0) {
+        Value dim = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(i));
+        Value start =
```
cathyzhyi, line 881: nit: use different variable names rather just `start` to improve readability. `Value gradInputSize, Value inputSize, Value sizeDiff, Value start`. Comments with the formula for calculation would be helpful as well.

cathyzhyi (changes requested):
IIUC, this implementation only support cases where `dilation [1,1]`, `stride [1,1]`, `transpose=False`, `bias_sizes=None`. Can we add guard to all those cases?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gpetters94: You're right, I've double checked the math and I have what should be the correct implementation. I'm working on implementing it, but I'm running into issues with `tensor::PadOp`. I'm going to try to write it with a different op (`tensor::InsertSlice`) and push that up tomorrow.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gpetters94: Does the current commit look good in this regard? I can't think of a good name for the intermediate values for some of these, but I think the comments make it readable enough.

cathyzhyi (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,177 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+        loc, weight.getType(), weight, ci0, ci1);
+
+    // Calculate padding for first convolution.
+    SmallVector<Value> gradInputPaddingValues;
+    for (auto i = 2; i < gradRank; i++) {
+      Value dim = rewriter.create<Torch::ConstantIntOp>(
+          loc, rewriter.getI64IntegerAttr(i));
+      Value outDim = rewriter.create<Torch::AtenSizeIntOp>(loc, input, dim);
+
+      // Calculate 1 + (weightDim // 2) * 2, which fixes issues with even-sized
```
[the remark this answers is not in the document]
gpetters94: Just to indicate it's floordiv, not div.

lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -67,6 +68,43 @@ Value torch_to_linalg::getZeroPaddedTensor(
   return getPaddedTensor(op, b, input, paddingInts, paddingInts, c0);
 }
 
+Value torch_to_linalg::getZeroPaddedTensor(
```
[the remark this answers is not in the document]
gpetters94: Thanks, that was a relic of a previous implementation.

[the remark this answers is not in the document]
gpetters94: Yep, This adds dynamic padding to a tensor, ignoring the first two channels. Using `tensor::PadOp` was causing errors with only dynamic values, so I rewrote it with other ops. I'll rename it to something like `getDynamicZeroPaddedTensor`.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -91,6 +91,65 @@ class ConvertAtenMmOp : public OpConversionPattern<AtenMmOp> {
 };
 } // namespace
 
+namespace {
+class ConvertAtenFlipOp : public OpConversionPattern<AtenFlipOp> {
```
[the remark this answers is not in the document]
gpetters94: Yep, I'll put a PR up for them.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,177 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+    }
+    for (Value dilation : dilations) {
+      Value cmp = rewriter.create<Torch::AtenEqIntOp>(loc, dilation, ci1);
+      rewriter.create<Torch::RuntimeAssertOp>(
+          loc, cmp, "unimplemented: only dilations of 1 supported.");
+    }
+
+    // Rotate weight.
+    SmallVector<Value> axes;
+    for (auto i = 2; i < gradRank; i++) {
```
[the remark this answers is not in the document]
gpetters94: Does that work? I don't think `i` is defined outside the loop.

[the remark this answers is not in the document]
cathyzhyi: oh, nvm, I missed the `i`.

lib/Conversion/TorchToLinalg/Utils.cpp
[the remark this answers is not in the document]
cathyzhyi: Can you add this as the comment of the API? Otherwise, other people need to look at the implementation to know how the API is supposed to work.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -778,6 +779,177 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+      // weight.
+      Value weightDim = rewriter.create<Torch::AtenSizeIntOp>(loc, weight, dim);
+      weightDim =
+          rewriter.create<Torch::AtenFloordivIntOp>(loc, weightDim, ci2);
+      weightDim = rewriter.create<Torch::AtenMulIntOp>(loc, weightDim, ci2);
+      weightDim = rewriter.create<Torch::AtenAddIntOp>(loc, weightDim, ci1);
+      Value gradOutDim =
+          rewriter.create<Torch::AtenSizeIntOp>(loc, gradOutput, dim);
+
+      // Calculate (((outDim - 1) * stride) + weightDim - gradOutDim) // 2, the
```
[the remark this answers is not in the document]
gpetters94: I'm double checking, but I believe this only works for stride 1.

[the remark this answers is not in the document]
gpetters94: It only works for stride 1 as-is. I don't think that has to do with the padding.

lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -67,6 +68,45 @@ Value torch_to_linalg::getZeroPaddedTensor(
...
+  assert(input.getType().isa<RankedTensorType>() &&
+         "input must be RankedTensorType");
+  Location loc = op->getLoc();
+  Type elementType = input.getType().cast<RankedTensorType>().getElementType();
+  auto inputRank = input.getType().cast<RankedTensorType>().getRank();
+  Value c0 = b.create<arith::ConstantOp>(loc, b.getI64IntegerAttr(0));
+  Value c1 = b.create<arith::ConstantOp>(loc, b.getIndexAttr(1));
+  Value c2 = b.create<arith::ConstantOp>(loc, b.getI64IntegerAttr(2));
+  SmallVector<Value> inputDims = getTensorSizes(b, loc, input);
+  SmallVector<Value> outputDims{inputDims[0], inputDims[1]};
```
cathyzhyi, line 85: nit: please group lines for `outputDims` together. It makes the code hard to follow when starting with `outputDims` then switch to `paddingIncludingUnchanged` then coming back to `outputDims` again.

lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -67,6 +68,45 @@ Value torch_to_linalg::getZeroPaddedTensor(
...
+  SmallVector<Value> strides(inputRank, c1);
+  SmallVector<Value> paddingIncludingUnchanged{c0, c0};
+  paddingIncludingUnchanged.append(padding);
+  for (auto pad = paddingIncludingUnchanged.begin();
+       pad < paddingIncludingUnchanged.end(); pad++)
+    *pad = castIntToIndex(b, loc, *pad);
+
+  for (auto i = 0; i < inputRank - 2; i++) {
```
cathyzhyi, line 93: Please refer to the implementation of `getPaddedTensor`  https://github.com/llvm/torch-mlir/blob/52c9987115b39c2f6673315a21f6cdff0d1f51ca/lib/Conversion/TorchToLinalg/Utils.cpp#L42-L52. We should be able to use `tensor::createPadScalarOp` here as well.

cathyzhyi (changes requested):
Not done reviewing yet. Left some comments regarding the `getDynamicZeroPaddedTensor ` API.

cathyzhyi:
Actually, don't we only accept the `padding` as constant int list? Why do we need the `getDynamicZeroPaddedTensor` API?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -779,6 +780,177 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+          op, "only constant transposed is supported.");
+    if (transposed)
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: transposed convolutions are not supported.");
+
+    // Add guards for unsupported values.
+    if (!llvm::all_of(outMask, [](bool mask) { return mask; }))
+      return rewriter.notifyMatchFailure(
+          op, "false values in output_mask not supported.");
+    for (Value stride : strides) {
```
cathyzhyi, line 861: I would suggest for simplicity let's stick to the existing limitation of `aten::convolution` here https://github.com/llvm/torch-mlir/blob/52c9987115b39c2f6673315a21f6cdff0d1f51ca/lib/Conversion/TorchToLinalg/Linear.cpp#L477-L489. Limit to constant int list and use compilation guard rather than `Torch::RuntimeAssertOp`

gpetters94:
> Actually, don't we only accept the `padding` as constant int list? Why do we need the `getDynamicZeroPaddedTensor` API?

The calculation for the correct padding, `gradInputPaddingValues`, requires a bunch of dynamic values (sizes of some of the tensors, etc) so I couldn't just implement it at compile time.

cathyzhyi:
> > Actually, don't we only accept the `padding` as constant int list? Why do we need the `getDynamicZeroPaddedTensor` API?
> 
> The calculation for the correct padding, `gradInputPaddingValues`, requires a bunch of dynamic values (sizes of some of the tensors, etc) so I couldn't just implement it at compile time.

when you need dynamic value you can alway got from `op.padding()` right?

gpetters94:
> > > Actually, don't we only accept the `padding` as constant int list? Why do we need the `getDynamicZeroPaddedTensor` API?
> > 
> > 
> > The calculation for the correct padding, `gradInputPaddingValues`, requires a bunch of dynamic values (sizes of some of the tensors, etc) so I couldn't just implement it at compile time.
> 
> when you need dynamic value you can always get from `op.padding()` right?

I mean the padding for the grad_input calculation:

```
      Value padVal = rewriter.create<Torch::AtenSubIntOp>(loc, outDim, ci1);
      padVal =
          rewriter.create<Torch::AtenMulIntOp>(loc, padVal, strides[i - 2]);
      padVal = rewriter.create<Torch::AtenAddIntOp>(loc, padVal, weightDim);
      padVal = rewriter.create<Torch::AtenSubIntOp>(loc, padVal, gradOutDim);
      padVal = rewriter.create<Torch::AtenFloordivIntOp>(loc, padVal, ci2);

      gradInputPaddingValues.push_back(padVal);
```

When I tried to use `tensor::PadOp` to pad using that, I got errors so I wrote the dynamic padding helper to do so. I could remove the guard for constant padding, since it should be handled with the helper.

lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -67,6 +68,51 @@ Value torch_to_linalg::getZeroPaddedTensor(
...
+  Type elementType = input.getType().cast<RankedTensorType>().getElementType();
+  Type inputType = RankedTensorType::get(
+      llvm::ArrayRef<int64_t>(SmallVector<int64_t>(inputRank, kUnknownSize)),
+      elementType);
+  input = b.create<tensor::CastOp>(loc, inputType, input);
+
+  SmallVector<Value> strides(inputRank, c1);
+  Value cf0 =
+      b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
+  return createPadScalarOp(input, cf0, paddingIncludingUnchanged,
```
cathyzhyi, line 112: Any reason why we can't use the existing helper `tensor::createPadScalarOp` ?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -779,6 +780,173 @@ class DecomposeAtenConv2dOp : public OpRewritePattern<AtenConv2dOp> {
...
+      Value weightDim = rewriter.create<Torch::AtenSizeIntOp>(loc, weight, dim);
+      weightDim =
+          rewriter.create<Torch::AtenFloordivIntOp>(loc, weightDim, ci2);
+      weightDim = rewriter.create<Torch::AtenMulIntOp>(loc, weightDim, ci2);
+      weightDim = rewriter.create<Torch::AtenAddIntOp>(loc, weightDim, ci1);
+      Value gradOutDim =
+          rewriter.create<Torch::AtenSizeIntOp>(loc, gradOutput, dim);
+
+      // Calculate (((outDim - 1) * stride) + weightDim - gradOutDim) // 2, the
+      // padding value for this dimension.
```
cathyzhyi, line 905: Can you add more comments about how you got this formula so that it's easier to review? Like from which original pytorch source you deduced this one?

lib/Conversion/TorchToLinalg/Utils.cpp
gpetters94: Yep, the existing help requires `OpFoldResult`s, which as far as I'm aware have to be static values. This should probably get upstreamed, but until it does putting it here is the only solution I could think of.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gpetters94: I got it from the docs online, is putting the URL in the comment okay?

lib/Conversion/TorchToLinalg/Utils.cpp
cathyzhyi: You can convert dynamic values to OpFoldResult with this helper https://github.com/llvm/llvm-project/blob/e1318078a4e160eb723bcbcfcdcc9a1b618f7067/mlir/include/mlir/Dialect/Utils/StaticValueUtils.h#L51

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
cathyzhyi: Yea, that would be helpful.

lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -67,6 +69,42 @@ Value torch_to_linalg::getZeroPaddedTensor(
...
+  for (auto pad = paddingIncludingUnchanged.begin();
+       pad < paddingIncludingUnchanged.end(); pad++)
+    *pad = castIntToIndex(b, loc, *pad);
+
+  auto inputRank = input.getType().cast<RankedTensorType>().getRank();
+  Type elementType = input.getType().cast<RankedTensorType>().getElementType();
+  Type inputType = RankedTensorType::get(
+      llvm::ArrayRef<int64_t>(SmallVector<int64_t>(inputRank, kUnknownSize)),
+      elementType);
+  input = b.create<tensor::CastOp>(loc, inputType, input);
```
cathyzhyi, line 97: The first argument of `createPadScalarOp` is the result type so I don't think you need to cast the input to result type.  You could follow the example here https://github.com/llvm/torch-mlir/blob/9208bf0eb63e3589b54b23324b5202916436a220/lib/Conversion/TorchToLinalg/Utils.cpp#L42-L44 to deduce the result type.

lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -67,6 +69,42 @@ Value torch_to_linalg::getZeroPaddedTensor(
...
+
+  SmallVector<Value> strides(inputRank, c1);
+  Value cf0 =
+      b.create<arith::ConstantOp>(loc, b.getFloatAttr(elementType, 0.0));
+  SmallVector<OpFoldResult> paddingValues =
+      getAsOpFoldResult(paddingIncludingUnchanged);
+  return tensor::createPadScalarOp(inputType, input, cf0, paddingValues,
+                                   paddingValues, false, loc, b);
```
cathyzhyi, line 105: nit: add comments for arguments so that it's easier to follow. refer to https://github.com/llvm/torch-mlir/blob/9208bf0eb63e3589b54b23324b5202916436a220/lib/Conversion/TorchToLinalg/Utils.cpp#L49-L51

cathyzhyi (changes requested):
Added some comments. Will review the conversion pattern later.

lib/Conversion/TorchToLinalg/Utils.cpp
gpetters94: inferResultType uses ints, so I don't think that's possible here.

gpetters94:
This is stale, closing.

Issue #727, opened 2022-04-01
Adding shape information for Torch MLIR LTC Backend


henrytwo:
After https://github.com/llvm/torch-mlir/pull/725 lands, the next goal is to try to add shape information to the output MLIR.

`lazy::Node` contains tensor shape information (retrieved by calling `shapes()`); however, this information is not retained when it's converted to a `jit::Node`. ([This](https://github.com/henrytwo/torch-mlir/blob/2de2f1182452b6d070d2ed484804c056436215b9/python/torch_mlir/csrc/base_lazy_backend/mlir_node_lowering.cpp#L193) shows that the only thing extracted from the `lazy::Node` during lowering is its symbol)

It also looks like there is nothing on the MLIR generation side for adding shape information (as far as I could see).

I don't think there is anything on the JIT side that enables storing tensor shape information, which might make this more difficult. My immediate thought is to try to map `lazy::Node -> MlirOperation` to insert shape information, but that doesn't feel like the most proper way to do this.

@silvasean, do you have any thoughts or ideas on how to proceed with this?

cc: @antoniojkim @ke1337

silvasean:
`jit::Value`'s of `c10::TensorType` should be able to hold a shape.

henrytwo:
Ah that's good to hear -- I'll investigate what we can do upstream

henrytwo:
I opened an issue on the PyTorch side here: https://github.com/pytorch/pytorch/issues/75217

henrytwo:
Closing this issue now that we have a PR open: https://github.com/llvm/torch-mlir/issues/727

PR #742, opened 2022-04-07
Generate MLIR with shape information via LTC frontend

henrytwo:
Previously, the intermediate JIT graph was generated without shape information, which caused generic `!torch.tensor` types to be scattered throughout the final MLIR. This PR improves the lowering by injecting each `jit::Value` of tensor type with its corresponding shape from the origin `lazy::Node`. Additionally, the MLIR parameter and return values are also generated with their respective shapes.

Example output for an FC MNIST model:
```
func @graph(%arg0: !torch.vtensor<[1,5],f32>, %arg1: !torch.vtensor<[1],si64>, %arg2: !torch.float, %arg3: !torch.int, %arg4: !torch.int, %arg5: !torch.int, %arg6: !torch.vtensor<[10,5],f32>, %arg7: !torch.vtensor<[10],f32>, %arg8: !torch.vtensor<*,f32>, %arg9: !torch.float) -> (!torch.vtensor<[1,5],f32>, !torch.vtensor<[1],si64>, !torch.vtensor<[10,5],f32>, !torch.vtensor<[10],f32>, !torch.vtensor<[1,10],f32>, !torch.vtensor<*,f32>, !torch.vtensor<[10],f32>, !torch.vtensor<[10,5],f32>) {
  %0 = torch.aten.detach %arg6 :  !torch.vtensor<[10,5],f32> ->  !torch.vtensor<[10,5],f32>
  %int1 = torch.constant.int 1
  %int0 = torch.constant.int 0
  %1 = torch.prim.ListConstruct %int1, %int0 : (!torch.int, !torch.int) -> !torch.list<int>
  %int1_0 = torch.constant.int 1
  %int0_1 = torch.constant.int 0
  %2 = torch.prim.ListConstruct %int1_0, %int0_1 : (!torch.int, !torch.int) -> !torch.list<int>
  %3 = torch.aten.permute %0, %2 :  !torch.vtensor<[10,5],f32>, !torch.list<int> ->  !torch.vtensor<[5,10],f32>
  %4 = torch.aten.detach %arg7 :  !torch.vtensor<[10],f32> ->  !torch.vtensor<[10],f32>
  %5 = torch.aten.addmm %4, %arg0, %3, %arg5, %arg4 :  !torch.vtensor<[10],f32>, !torch.vtensor<[1,5],f32>, !torch.vtensor<[5,10],f32>, !torch.int, !torch.int ->  !torch.vtensor<[1,10],f32>
  %6 = torch.aten.relu %5 :  !torch.vtensor<[1,10],f32> ->  !torch.vtensor<[1,10],f32>
  %int1_2 = torch.constant.int 1
  %7 = torch.aten._log_softmax %6, %int1_2, %false :  !torch.vtensor<[1,10],f32>, !torch.int, !torch.bool ->  !torch.vtensor<[1,10],f32>
  %none = torch.constant.none
  %int1_3 = torch.constant.int 1
  %int-100 = torch.constant.int -100
  %output, %total_weight = torch.aten.nll_loss_forward %7, %arg1, %none, %int1_3, %int-100 :  !torch.vtensor<[1,10],f32>, !torch.vtensor<[1],si64>, !torch.none, !torch.int, !torch.int ->  !torch.vtensor<*,f32>, !torch.vtensor<*,f32>
  %8 = torch.prim.TupleConstruct %output, %total_weight : !torch.vtensor<*,f32>, !torch.vtensor<*,f32> -> !torch.tuple<vtensor, vtensor>
  %none_4 = torch.constant.none
  %int1_5 = torch.constant.int 1
  %int-100_6 = torch.constant.int -100
  %9 = torch.aten.nll_loss_backward %arg8, %7, %arg1, %none_4, %int1_5, %int-100_6, %total_weight :  !torch.vtensor<*,f32>, !torch.vtensor<[1,10],f32>, !torch.vtensor<[1],si64>, !torch.none, !torch.int, !torch.int, !torch.vtensor<*,f32> ->  !torch.vtensor<[1,10],f32>
  %int1_7 = torch.constant.int 1
  %int6 = torch.constant.int 6
  %10 = torch.aten._log_softmax_backward_data %9, %7, %int1_7, %int6 :  !torch.vtensor<[1,10],f32>, !torch.vtensor<[1,10],f32>, !torch.int, !torch.int ->  !torch.vtensor<[1,10],f32>
  %11 = torch.aten.threshold_backward %10, %6, %arg3 :  !torch.vtensor<[1,10],f32>, !torch.vtensor<[1,10],f32>, !torch.int ->  !torch.vtensor<[1,10],f32>
  %int1_8 = torch.constant.int 1
  %int0_9 = torch.constant.int 0
  %12 = torch.prim.ListConstruct %int1_8, %int0_9 : (!torch.int, !torch.int) -> !torch.list<int>
  %int1_10 = torch.constant.int 1
  %int0_11 = torch.constant.int 0
  %13 = torch.prim.ListConstruct %int1_10, %int0_11 : (!torch.int, !torch.int) -> !torch.list<int>
  %14 = torch.aten.permute %arg0, %13 :  !torch.vtensor<[1,5],f32>, !torch.list<int> ->  !torch.vtensor<[5,1],f32>
  %15 = torch.aten.mm %14, %11 :  !torch.vtensor<[5,1],f32>, !torch.vtensor<[1,10],f32> ->  !torch.vtensor<[5,10],f32>
  %int1_12 = torch.constant.int 1
  %int0_13 = torch.constant.int 0
  %16 = torch.prim.ListConstruct %int1_12, %int0_13 : (!torch.int, !torch.int) -> !torch.list<int>
  %int1_14 = torch.constant.int 1
  %int0_15 = torch.constant.int 0
  %17 = torch.prim.ListConstruct %int1_14, %int0_15 : (!torch.int, !torch.int) -> !torch.list<int>
  %18 = torch.aten.permute %15, %17 :  !torch.vtensor<[5,10],f32>, !torch.list<int> ->  !torch.vtensor<[10,5],f32>
  %19 = torch.aten.detach %18 :  !torch.vtensor<[10,5],f32> ->  !torch.vtensor<[10,5],f32>
  %20 = torch.aten.add.Tensor %0, %19, %arg2 :  !torch.vtensor<[10,5],f32>, !torch.vtensor<[10,5],f32>, !torch.float ->  !torch.vtensor<[10,5],f32>
  %int0_16 = torch.constant.int 0
  %21 = torch.prim.ListConstruct %int0_16 : (!torch.int) -> !torch.list<int>
  %none_17 = torch.constant.none
  %22 = torch.aten.sum.dim_IntList %11, %21, %true, %none_17 :  !torch.vtensor<[1,10],f32>, !torch.list<int>, !torch.bool, !torch.none ->  !torch.vtensor<[1,10],f32>
  %int10 = torch.constant.int 10
  %23 = torch.prim.ListConstruct %int10 : (!torch.int) -> !torch.list<int>
  %int10_18 = torch.constant.int 10
  %24 = torch.prim.ListConstruct %int10_18 : (!torch.int) -> !torch.list<int>
  %25 = torch.aten.reshape %22, %24 :  !torch.vtensor<[1,10],f32>, !torch.list<int> ->  !torch.vtensor<[10],f32>
  %26 = torch.aten.detach %25 :  !torch.vtensor<[10],f32> ->  !torch.vtensor<[10],f32>
  %27 = torch.aten.add.Tensor %4, %26, %arg9 :  !torch.vtensor<[10],f32>, !torch.vtensor<[10],f32>, !torch.float ->  !torch.vtensor<[10],f32>
  return %arg0, %arg1, %20, %27, %6, %output, %26, %19 : !torch.vtensor<[1,5],f32>, !torch.vtensor<[1],si64>, !torch.vtensor<[10,5],f32>, !torch.vtensor<[10],f32>, !torch.vtensor<[1,10],f32>, !torch.vtensor<*,f32>, !torch.vtensor<[10],f32>, !torch.vtensor<[10,5],f32>
}
```

Note: This PR is based on https://github.com/llvm/torch-mlir/pull/725, but GitHub does not allow for setting the base to a fork while keeping this PR in the upstream repo, so there are several shared commits. Only the last one is specifically part of this PR.

Marked as draft while waiting for a dependency PR to land: https://github.com/llvm/torch-mlir/pull/725

Resolves: https://github.com/llvm/torch-mlir/issues/727

cc: @antoniojkim @ke1337

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/import_options.h
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/import_options.h
@@ -1,0 +1,29 @@
+//===- import_options.h -----------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef TORCHMLIRJITIRIMPORTER_CSRC_IMPORT_OPTIONS_H
+#define TORCHMLIRJITIRIMPORTER_CSRC_IMPORT_OPTIONS_H
+
+namespace torch_mlir {
+// Common import options across importers. We define this as a struct to avoid
+// an unstructured proliferation of different kinds of ways to control different
+// parts of the import process.
+struct ImportOptions {
+  // If this is set to true, then all tensors in the program can be assumed to
+  // have value semantics. This can happen, for example, when coming from
+  // LazyTensorCore since conversion to value semantics has already happened at
+  // a higher level there before we see the program. For
+  // calling-convention-impacting decisions, this flag should be interpreted as
+  // a requirement to use a value-semantic tensor type (!torch.vtensor) in
+  // signatures.
+  bool assumeTensorsHaveValueSemantics = false;
+};
+} // namespace torch_mlir
+
+#endif // TORCHMLIRJITIRIMPORTER_CSRC_IMPORT_OPTIONS_H
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.h
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.h
@@ -10,6 +10,8 @@
 #ifndef TORCHMLIRJITIRIMPORTER_CSRC_TORCH_TO_MLIR_UTILS_H
 #define TORCHMLIRJITIRIMPORTER_CSRC_TORCH_TO_MLIR_UTILS_H
 
+#include "import_options.h"
+
 #include <memory>
 
 #include "pybind.h"
@@ -37,14 +39,16 @@ MlirType getMlirTypeForTorchScalarType(MlirLocation loc,
 /// Maps a torch type to a corresponding MlirType. Returns a null type
 /// on failure and emits a diagnostic.
 MlirType getMlirTypeFromTorchType(MlirLocation loc,
-                                  const c10::TypePtr &torchType);
+                                  const c10::TypePtr &torchType,
+                                  const ImportOptions &importOptions = {});
 
 /// Creates a FunctionType suitable for expressing the signature of `schema`.
 ///
 /// This can differ from the type inferred from the block of a
 /// torch::jit::Function due to derefinement.
 MlirType getFunctionTypeFromSchema(MlirContext context,
-                                   const c10::FunctionSchema &schema);
+                                   const c10::FunctionSchema &schema,
+                                   const ImportOptions &importOptions = {});
 
 /// Creates an appropriate MlirAttribute that holds the same values as `tensor`.
 MlirAttribute convertTensorToMlirElementsAttr(at::Tensor tensor,
@@ -58,7 +62,8 @@ MlirLocation getMlirLocationFromNode(MlirContext context,
 
 std::vector<MlirType>
 getMlirTypesFromValues(MlirLocation loc,
-                       c10::ArrayRef<torch::jit::Value *> values);
+                       c10::ArrayRef<torch::jit::Value *> values,
+                       const ImportOptions &importOptions = {});
 
 std::vector<MlirValue> derefineValues(c10::ArrayRef<MlirValue> values,
                                       c10::ArrayRef<MlirType> expectedTypes,
--- a/build_tools/autogen_ltc_backend.py
+++ b/build_tools/autogen_ltc_backend.py
@@ -158,7 +158,7 @@ class MlirLazyIr(codegen.gen_lazy_tensor.dest.LazyIR):
     size_t i = 0;
     {emplace_arguments_str}
     {emplace_kwarguments}
-    torch::lazy::TorchMlirOpVector {schema.aten_name}_out = torch::lazy::LowerTorchMlirBuiltin(function, op().op, arguments, kwarguments);
+    torch::lazy::TorchMlirOpVector {schema.aten_name}_out = torch::lazy::LowerTorchMlirBuiltin(function, op().op, shapes(), arguments, kwarguments);
     CHECK_EQ({schema.aten_name}_out.size(), {len(func.returns)});
   
     return {schema.aten_name}_out;
--- a/examples/ltc_backend_mnist.py
+++ b/examples/ltc_backend_mnist.py
@@ -42,7 +42,7 @@ def main(device):
     class Model(torch.nn.Module):
         def __init__(self):
             super().__init__()
-            self.fc1 = torch.nn.Linear(5, 5)
+            self.fc1 = torch.nn.Linear(5, 10)
 
         def forward(self, x):
             out = self.fc1(x)
--- a/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
+++ b/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
@@ -36,9 +36,6 @@ TorchMlirComputation::TorchMlirComputation(
     const std::shared_ptr<torch::jit::Graph>& graph)
     : func_op_(std::move(func_op)), mlir_context_(std::move(mlir_context)),
       graph_(graph), num_results_(graph_->outputs().size()) {
-
-  // TODO(henrytu): Save parameter shape information.
-
   for (torch::jit::Value* input : graph_->inputs()) {
     parameter_names_.push_back(input->debugName());
   }
@@ -144,22 +141,15 @@ void TorchMlirLoweringContext::AddParameter(
 ComputationPtr TorchMlirLoweringContext::Build() {
   PRINT_FUNCTION();
 
+  // Insert return values into graph.
   for (torch::jit::Value* output : root_tuple_) {
     graph_->block()->registerOutput(output);
   }
 
-  // Create jit::Function from jit::Graph.
-  c10::QualifiedName name("graph");
-  auto cu = std::make_shared<torch::jit::CompilationUnit>();
-  // IMPORTANT: We pass in a COPY of the graph into create_function, since it
-  //            may get mutated in the process.
-  auto jit_fn = cu->create_function(std::move(name), std::move(graph_->copy()));
-
   // Generate MLIR.
-  MlirOperation func_op =
-      torch_mlir::importJitFunctionAsFuncOp(mlir_context_, jit_fn);
+  MlirOperation func_op = torch_mlir::importJitFunctionAsFuncOp(
+      mlir_context_, generate_jit_fn().get());
 
-  // TODO(henrytu): Inject tensor shapes into func_op
   return std::make_shared<TorchMlirComputation>(func_op, mlir_context_, graph_);
 }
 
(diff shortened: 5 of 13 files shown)


silvasean:
Thanks! Looks good.

Can you also see if ` !torch.tensor` can be replaced with `!torch.vtensor` here? I think the Lazy IR should have value semantics, e.g. the code in lazy_tensor_core/lazy_tensor_core/csrc/view.h

henrytwo:
Hm I took a look at the code and it looks like the only place where `torchMlirTorchValueTensorTypeGet` is called is [here](https://github.com/henrytwo/torch-mlir/blob/b6d13301fc4179a76f4e66b6d58d5d344bc663a7/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/ivalue_importer.cpp#L503-L539) when creating `type_bound` attributes. The [current location for type generation](https://github.com/henrytwo/torch-mlir/blob/0cb216a1adcdcc8c948503c0f856604997d13c1d/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp#L137-L154) seems to always produce non-value semantic tensor type. 

Given that this is how this is setup, what's the expected way to generate value semantic tensors?

I also looked through some of the Lazy IR code (including a file similar to the one you listed) and [didn't see anything in there](https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/view_ops/view.h) explicitly about value semantics. If I understand correctly though, operations with an underscore are done in place, and therefore _shouldn't_ have value semantics?

silvasean:
> Hm I took a look at the code and it looks like the only place where `torchMlirTorchValueTensorTypeGet` is called is [here](https://github.com/henrytwo/torch-mlir/blob/b6d13301fc4179a76f4e66b6d58d5d344bc663a7/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/ivalue_importer.cpp#L503-L539) when creating `type_bound` attributes. The [current location for type generation](https://github.com/henrytwo/torch-mlir/blob/0cb216a1adcdcc8c948503c0f856604997d13c1d/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp#L137-L154) seems to always produce non-value semantic tensor type.
> 
> Given that this is how this is setup, what's the expected way to generate value semantic tensors?

I think we would basically need to thread a bool through the relevant code.

We could create an `import_options.h` with a struct like:

```
struct ImportOptions {
  bool assumeTensorHaveValueSemantics = false;
};
```

To make things a bit more organized vs a loose bool.


> 
> I also looked through some of the Lazy IR code (including a file similar to the one you listed) and [didn't see anything in there](https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/view_ops/view.h) explicitly about value semantics. If I understand correctly though, operations with an underscore are done in place, and therefore _shouldn't_ have value semantics?

I think this is a question for the LTC devs about the semantics of the lazy IR. They might prefer the term "purely functional" over "value semantics" but it means the same thing in this context.

henrytwo:
> `assumeTensorHaveValueSemantics`

Does this mean you want us to assume tensors _do_ have value semantics when generating MLIR, unless if we run into a case where we know that it shouldn't?

henrytwo:
> There are a few other cases besides the trailing underscore variants, such as batch_norm updating the running_mean/var in place.

Is there an authoritative list to lookup what should and should not have value semantics? I can see in the [tablegen](https://github.com/henrytwo/torch-mlir/blob/7c3ba25238ac73850fcdd698be1fb084f8a58e49/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td) that certain ops are listed with `HasValueSemantics`. Would this be a sufficient check?

silvasean:
In general the alias annotations (which determine HasValueSemantics) are the source of truth. There are a few exceptions: https://github.com/pytorch/pytorch/issues/73050#issuecomment-1051382044

henrytwo:
@silvasean Just to clarify before I proceed with implementing these changes, do you want *all* the tensors to be of type `!torch.vtensor` when generating MLIR using the LTC frontend, or should this decision be based based on alias annotations + list of exceptions?

silvasean:
> @silvasean Just to clarify before I proceed with implementing these changes, do you want _all_ the tensors to be of type `!torch.vtensor` when generating MLIR using the LTC frontend, or should this decision be based based on alias annotations + list of exceptions?

I would phrase my request as "where it is trivial to know by construction that the tensor has value semantics, use !torch.vtensor", which I think practically speaking means that if the Lazy IR provides a guarantee about all tensors having value semantics, then use !torch.vtensor, otherwise, just don't worry about it and MaximizeValueSemantics will do the best it can later.

henrytwo:
@silvasean Please have a look at the latest commit. I took a look at [this](https://discord.com/channels/636084430946959380/938463238985760860/961046571728896032) discussion you had with @antoniojkim on Discord it looks like LTC makes all inline operations functional, so I've added a flag to use `!torch.vtensor`

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/import_options.h
```diff
@@ -0,0 +1,19 @@
...
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef TORCHMLIRJITIRIMPORTER_CSRC_IMPORT_OPTIONS_H
+#define TORCHMLIRJITIRIMPORTER_CSRC_IMPORT_OPTIONS_H
+
+namespace torch_mlir {
+struct ImportOptions {
+  bool assumeTensorHaveValueSemantics = false;
```
silvasean, line 15: Can you add a comment here something like:

"""
If this is set to true, then all tensors in the program can be assumed to have value semantics. This can happen, for example, when coming from LazyTensorCore since conversion to value semantics has already happened at a higher level there before we see the program. For calling-convention-impacting decisions, this flag should be interpreted as a requirement to use a value-semantic tensor type (`!torch.vtensor`) in signatures.
"""

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/import_options.h
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef TORCHMLIRJITIRIMPORTER_CSRC_IMPORT_OPTIONS_H
+#define TORCHMLIRJITIRIMPORTER_CSRC_IMPORT_OPTIONS_H
+
+namespace torch_mlir {
+struct ImportOptions {
```
silvasean, line 14: Can you add a comment here:

"""
Common import options across importers. We define this as a struct to avoid an unstructured proliferation of different kinds of ways to control different parts of the import process.
"""

silvasean (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.h
```diff
@@ -37,14 +39,16 @@ MlirType getMlirTypeForTorchScalarType(MlirLocation loc,
 /// Maps a torch type to a corresponding MlirType. Returns a null type
 /// on failure and emits a diagnostic.
 MlirType getMlirTypeFromTorchType(MlirLocation loc,
-                                  const c10::TypePtr &torchType);
+                                  const c10::TypePtr &torchType,
+                                  ImportOptions importOptions);
```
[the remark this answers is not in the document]
henrytwo: Would you like `importOptions` to have a default value for all functions where it's a parameter, or only instances where it is currently used with a dummy value?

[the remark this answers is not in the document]
silvasean: I think it should always have a default for the functions that you are adding it to in this PR.

silvasean (approved):

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/import_options.h
```diff
@@ -0,0 +1,29 @@
...
+// parts of the import process.
+struct ImportOptions {
+  // If this is set to true, then all tensors in the program can be assumed to
+  // have value semantics. This can happen, for example, when coming from
+  // LazyTensorCore since conversion to value semantics has already happened at
+  // a higher level there before we see the program. For
+  // calling-convention-impacting decisions, this flag should be interpreted as
+  // a requirement to use a value-semantic tensor type (!torch.vtensor) in
+  // signatures.
+  bool assumeTensorHaveValueSemantics = false;
```
silvasean, line 25: nit: assumeTensor**s**HaveValueSemantics

PR #744, opened 2022-04-08
Add bazel build support (2/N) 

asaadaldien:
This PR adds very basic GH action workflow to build `@torch-mlir//...`


.github/workflows/bazelBuildAndTest.yml
```diff
@@ -0,0 +1,30 @@
+name: Bazel Build and Test
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      release_id:
```
silvasean, line 10: I don't think you need this release_id thing.

.github/workflows/bazelBuildAndTest.yml
```diff
@@ -0,0 +1,30 @@
...
+      with:
+        python-version: 3.9
+    - name: Get torch-mlir
+      uses: actions/checkout@v2
+      with:
+        submodules: 'true'
+    - name: Build with bazel
+      run: |
+        cd $GITHUB_WORKSPACE/utils/bazel
+        bazel build @torch-mlir//...
```
silvasean, line 30: nit: newline here to avoid this red thing in the github ui

.github/workflows/bazelBuildAndTest.yml
```diff
@@ -0,0 +1,30 @@
+name: Bazel Build and Test
+
+on:
+  push:
+    branches:
+      - main
```
silvasean, line 6: who does this notify when it fails?

silvasean (changes requested):

.github/workflows/bazelBuildAndTest.yml
asaadaldien: My understanding this is for running this workflow on new commits on main branch.
Setting notifications and if this workflow is merge blocking or not isn't specified here If I understand GH actions documentations correctly.

.github/workflows/bazelBuildAndTest.yml
asaadaldien: You are right we don't want to push  artifacts here, will remove this. thansk!

asaadaldien:
I changed the workflow as discussed on discord
- Only runs on new commits merged to main nor PRs
- When it fails it will notify google group email where downstream bazel users are subscribed to
    Note: this section of code for sending a notification email is commented out because we need to add the mail/credentials information to the repo encrypted env variables https://docs.github.com/en/actions/security-guides/encrypted-secrets which I don't have access to do.

silvasean:
> I changed the workflow as discussed on discord
> 
> * Only runs on new commits merged to main nor PRs
> * When it fails it will notify google group email where downstream bazel users are subscribed to
>   Note: this section of code for sending a notification email is commented out because we need to add the mail/credentials information to the repo encrypted env variables https://docs.github.com/en/actions/security-guides/encrypted-secrets which I don't have access to do.

I can add the repo secrets. PM me the values to set them to.

asaadaldien:
Hey @silvasean I tried the notifications it failed with server_address = "". did you push the `secrets.SMTP_SERVER` value ?

silvasean:
I just updated secrets.SMTP_SERVER -- let me know if the issue persists -- there might be a deeper issue.

asaadaldien:
@silvasean, I tried running the bazel action on this PR and it can't fetch any `secrets` which might be a security thing otherwise people can send PRs and exploit these values `echo secrets.VAL`. I think we can merge this and the when the bazel build fails see if we got an email or not :), wdyt ?

silvasean:
> @silvasean, I tried running the bazel action on this PR and it can't fetch any `secrets` which might be a security thing otherwise people can send PRs and exploit these values `e.g echo secrets.VAL`. I think we can merge this and the when the bazel build fails see if we got an email or not :), wdyt ?

Let's give it a shot!

asaadaldien:
Thanks @silvasean, I am merging this!

PR #745, opened 2022-04-08
Add unit tests for `build_script_function`

makslevental:
Add unit tests for `build_script_function`.

NB: this PR is stacked on top of https://github.com/llvm/torch-mlir/pull/740, so only the most recent commit is the actual diff.

### DIFF
--- a/python/test/eager_mode/normalize_args_kwargs.py
+++ b/python/test/eager_mode/normalize_args_kwargs.py
@@ -8,19 +8,12 @@
 
 import torch
 
+from framework import run_test
 from torch_mlir.eager_mode.torch_mlir_dispatch import normalize_args_kwargs
 
 
-def run_test(test, XFAIL=False, XPASS=False):
-    try:
-        test()
-        print(("X" if XPASS else "") + f"PASS - {test.__name__}")
-    except Exception as e:
-        print(("X" if XFAIL else "") + f"FAIL - {test.__name__}")
-        print(e)
-
-
 # CHECK: PASS - should_normalize
+@run_test
 def should_normalize():
     target = torch.ops.aten.max_pool2d_with_indices.default.overloadpacket
     args = (torch.randn((1, 3, 32, 32)),)
@@ -44,6 +37,7 @@ def should_normalize():
 
 # CHECK: FAIL - shouldnt_normalize1
 # CHECK: Couldn't normalize args and kwargs
+@run_test
 def shouldnt_normalize1():
     target = torch.ops.aten.max_pool2d_with_indices.default.overloadpacket
     args = (torch.randn((1, 3, 32, 32)),)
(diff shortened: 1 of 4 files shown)


python/test/eager_mode/normalize_args_kwargs.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+import torch
+
+from framework import register_test, run_tests
+from torch_mlir.eager_mode.torch_mlir_dispatch import normalize_args_kwargs
+
+
+# CHECK: PASS - should_normalize
+@register_test
```
silvasean, line 16: I don't think you need to "register" anything. Just running it immediately is fine. The requrements of the e2e testing framework are a bit different since we need to drive it from the command line, filter tests, etc. But in this case we already have lit which can do that at .py-file granularity which seems to be enough.

silvasean (changes requested):

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -125,7 +126,11 @@ def build_script_function(
     else:
         graph.registerOutput(node.output())
 
-    fn_name = str(node).strip()
+    # The first alternation handles non-alpha characters at the beginning.
+    # MLIR doesn't support such symbol names unless quoted, hence safer to remove.
+    # The second alternation handles non-word characters in the rest of the name (eg =, :, etc).
+    # LLVM doesn't support such symbol names.
+    fn_name = re.sub(r'^[^a-zA-Z]+|[\W_]+', '', str(node))
```
silvasean, line 133: I'm not sure I understand the requirements here. In what way is this name load-bearing? Can we just call it "f"? If not, then I think we also need to mangle the tensor shapes we specialized for as well.

makslevental: it's load-bearing because we use this name to pull the exported function out of the module here

https://github.com/llvm/torch-mlir/blob/9ec0683e928bdd1acea9aeb025ed542332fab902/python/torch_mlir/eager_mode/torch_mlir_dispatch.py#L260

>op_mlir_backend_callable = getattr(loaded_module, script_fun.name)

right now there's only ever one exported function in that module but in the future there could be others (in the case that we decide to incrementally compile and inline or not).

>If not, then I think we also need to mangle the tensor shapes we specialized for as well.

good idea!

silvasean: Looking at the code, it seems like it would still continue to work if we just called the function "f" every time. Right? That seems simplest to do so I think we should do it.

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -6,6 +6,7 @@
 from typing import Any, Callable, Tuple, Union
 from typing import List, Dict
 
+import re
```
silvasean, line 9: nit: can remoe the import

silvasean (approved):

PR #747, opened 2022-04-11
Fix small bug in annotate args in eager

makslevental:
Tensor kwargs are offset indexed (e.g., starting position `i=5`), and just `range`ing over wasn't correct.

Also add tests.

NB: this PR is stacked on top of https://github.com/llvm/torch-mlir/pull/745, so only the most recent commit is the actual diff.

### DIFF
--- a/python/torch_mlir/eager_mode/torch_mlir_dispatch.py
+++ b/python/torch_mlir/eager_mode/torch_mlir_dispatch.py
@@ -168,8 +168,7 @@ def annotate_args_kwargs(
             if isinstance(arg, np.ndarray):
                 tensor_kwargs[arg_idxs[kw]] = (arg, normalized_kwargs[kw].dtype)
 
-        for i in range(len(tensor_kwargs)):
-            arg, arg_dtype = tensor_kwargs[i]
+        for _i, (arg, arg_dtype) in sorted(tensor_kwargs.items()):
             annotations.append(TorchTensorType(shape=tuple(arg.shape), dtype=arg_dtype))
             tensor_kwargs_flat.append(arg)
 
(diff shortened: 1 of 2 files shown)


python/test/eager_mode/build_script_function.py
```diff
@@ -0,0 +1,316 @@
...
+        save_mean=save_mean,
+        save_invstd=save_invstd,
+    )
+
+    script_fun = build_script_function(target._schema, args, kwargs)
+    print(script_fun.graph)
+
+
+def main():
+    run_test(simple)
```
silvasean, line 302: same comment about the decorator

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -123,7 +126,11 @@ def build_script_function(
     else:
         graph.registerOutput(node.output())
 
-    fn_name = str(node).strip()
+    # The first alternation handles non-alpha characters at the beginning.
+    # MLIR doesn't support such symbol names unless quoted, hence safer to remove.
```
silvasean, line 130: The quoting is just the textual format. Arbitrary strings should be allowed. So I don't think this should be a factor. Downstream projects might have bugs.

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
```diff
@@ -123,7 +126,11 @@ def build_script_function(
...
+    # The second alternation handles non-word characters in the rest of the name (eg =, :, etc).
+    # LLVM (and therefore IREE) doesn't support such symbol names.
```
silvasean, line 132: My understanding is that LLVM allows arbitrary symbol names too. Are you sure you need this? Also, can you remove the mention of IREE? IREE is not a part of Torch-MLIR.

silvasean (changes requested):

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
makslevental: addressed in the PR that this change associated with

python/torch_mlir/eager_mode/torch_mlir_dispatch.py
makslevental: addressed in the PR that this change is associated with (IREE removed there)

silvasean (approved):

Issue #751, opened 2022-04-13
`torch.bool` presents as unhandled signless `i1`


makslevental:
Minimal repro:

```python
import torch

from torch_mlir.dialects.torch.importer.jit_ir import ModuleBuilder, ClassAnnotator
from torch_mlir.passmanager import PassManager
from torch_mlir_e2e_test.linalg_on_tensors_backends import refbackend

class BoolMod(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, a, b):
        return a * b


mod = BoolMod()
class_annotator = ClassAnnotator()
recursivescriptmodule = torch.jit.script(mod)

class_annotator.exportNone(recursivescriptmodule._c._type())
class_annotator.exportPath(recursivescriptmodule._c._type(), ["forward"])
class_annotator.annotateArgs(
    recursivescriptmodule._c._type(),
    ["forward"],
    [None, ([-1, -1], torch.bool, True), ([-1, -1], torch.bool, True)],
)

mb = ModuleBuilder()
mb.import_module(recursivescriptmodule._c, class_annotator)

backend = refbackend.RefBackendLinalgOnTensorsBackend()
with mb.module.context:
    pm = PassManager.parse(
        "torchscript-module-to-torch-backend-pipeline,torch-backend-to-linalg-on-tensors-backend-pipeline"
    )
    pm.run(mb.module)

compiled = backend.compile(mb.module)
jit_module = backend.load(compiled)

a = torch.ones((10, 10), dtype=torch.bool).numpy()
b = torch.zeros((10, 10), dtype=torch.bool).numpy()
jit_module.forward(a, b)
```

Exception:

```
LLVM ERROR: unhandled type for getScalarTypeForType
```

Source of exception is https://github.com/llvm/torch-mlir/blob/d46f169c1a8443caa62b23b331779399f0d0c563/lib/Dialect/Torch/Utils/Utils.cpp#L55

Solution is to add `type.isSignlessInteger(1)` in the conditional that returns `torch_upstream::ScalarType::Bool`.

PR #752, opened 2022-04-13
Fix `torch.bool` parsed as signless `i1`

makslevental:
Fixes https://github.com/llvm/torch-mlir/issues/751 where `torch.bool` is parsed as signless `i1`.

### DIFF
--- a/lib/Dialect/Torch/Utils/Utils.cpp
+++ b/lib/Dialect/Torch/Utils/Utils.cpp
@@ -50,7 +50,7 @@ torch_upstream::ScalarType Torch::getScalarTypeForType(Type type) {
     return torch_upstream::ScalarType::Long;
   if (type.isSignedInteger(32))
     return torch_upstream::ScalarType::Int;
-  if (type.isUnsignedInteger(1))
+  if (type.isSignlessInteger(1))
     return torch_upstream::ScalarType::Bool;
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Utils/Utils.cpp
```diff
@@ -50,7 +50,7 @@ torch_upstream::ScalarType Torch::getScalarTypeForType(Type type) {
     return torch_upstream::ScalarType::Long;
   if (type.isSignedInteger(32))
     return torch_upstream::ScalarType::Int;
-  if (type.isUnsignedInteger(1))
+  if (type.isUnsignedInteger(1) || type.isSignlessInteger(1))
```
cathyzhyi, line 53: Actually, the `type.isUnsignedInteger(1)` can be removed as the dtype can't be `si1`
https://github.com/llvm/torch-mlir/blob/eaf34fa02bd2d90d89e2bee2313c161f417c0709/include/torch-mlir/Dialect/Torch/IR/TorchTypes.td#L155-L167

cathyzhyi (changes requested):
Thanks for the fix!

cathyzhyi:
btw, the new test need to be added to the list here https://github.com/llvm/torch-mlir/blob/d46f169c1a8443caa62b23b331779399f0d0c563/e2e_testing/torchscript/xfail_sets.py#L35

PR #754, opened 2022-04-13
Add device support to TorchMLIR tensor.

makslevental:
Adds support for preserving device throughout eager mode by explicitly checking and copying `elem.device`.


silvasean:
I don't think this is the right factoring. I would suggest having a single TorchMLIRTensor class which has-a: `elem: DeviceTensor` (where `DeviceTensor` is some arbitrary type from the backend), and `backend: TorchMLIREagerBackend`.

TorchMLIREagerBackend would be an abstract base class that would look something like:
```
import abc
from typing import TypeVar, Callable

import torch

from torch_mlir.ir import Module



# An opaque type used to represent tensors on the device.
# Each backend will likely have a different specific type for this,
# but thankfully Python's duck typing is sufficient.
# TODO: This might need to be an ABC too, such as
# to support finding the backend that created the tensor.
DeviceTensor = TypeVar('DeviceTensor')


class TorchMLIREagerBackend(abc.ABC):
    @abc.abstractmethod
    def compile(self, module: Module) -> Callable[[List[DeviceTensor]], List[DeviceTensor]]:
        pass
    @abc.abstractmethod
    def transfer_from_torch_to_device(self, torch.Tensor) -> DeviceTensor:
        pass
    @abc.abstractmethod
    def transfer_from_device_to_torch(self, DeviceTensor) -> torch.Tensor:
        pass
```

makslevental:
> I don't think this is the right factoring. I would suggest having a single TorchMLIRTensor class which has-a: `elem: DeviceTensor` (where `DeviceTensor` is some arbitrary type from the backend), and `backend: TorchMLIREagerBackend`. ....

@silvasean 

I considered this pattern. In my understanding it doesn't work because
1. It doesn't handle building the `torch.Tensor` metadata from a device specific tensor in `__new__` (this is why I made `__new__` abstract in this factoring). In particular, unless you want to have to do `if not isintance(elem, torch.Tensor): backend.transfer_from_device_to_torch(elem)` for every new instance of `TorchMLIRTensor`, you cannot avoid overriding `__new__`. Again, as per the zoo examples, [the wrapper tensor must have all of the same metadata as the wrapped tensor](https://github.com/albanD/subclass_zoo/blob/1566e038f03cd89ab3cc37e670a44e3c2bbc1897/trivial_tensors.py#L90-L92) :
    >The nuance of wrapper tensors is that the external wrapper tensor is still
    required to have all of the metadata that the inner tensor has; this
    includes stride and storage!
2. Forces you to either inject `backend` on construction of `TorchMLIRTensor` (just as PyTorch injects `device`) or set it as some sort of global (module level probably). Workable but eventually unworkable.

silvasean:
For 1., I think that `get_torch_metadata` would be useful. You could also do that computation with meta tensors without querying it from the backend.

For 2., yeah, a global would be good. This is a common pattern in Python. We can always allow an explicitly provided backend but use the default global one.

python/torch_mlir/eager_mode/refbackend.py
```diff
@@ -0,0 +1,87 @@
...
+    np.float32: torch.float32,
+    np.float64: torch.float64,
+    np.complex64: torch.complex64,
+    np.complex128: torch.complex128,
+}
+
+_ref_backend = RefBackendLinalgOnTensorsBackend()
+
+
+class EagerModeRefBackend(TorchMLIREagerBackend):
```
silvasean, line 41: this file can now go in `torch_mlir_e2e_test/eager_backends/refbackend.py`

python/torch_mlir/eager_mode/torch_mlir_eager_backend.py
```diff
@@ -0,0 +1,69 @@
...
+
+from torch_mlir._mlir_libs._mlir.ir import Module
+
+# TODO: This might need to be an ABC too, such as
+# to support finding the backend that created the tensor.
+DeviceTensor = TypeVar("DeviceTensor")
+
+
+@dataclass
+class TensorMetaData:
```
silvasean, line 20: Can you document this? I don't expect all backends will be able to know requires_grad/device/layout/strides all the time -- is there a way that we can factor that out of here? I think that size/dtype is the only thing we should need for now.
Bonus points if you just use meta tensors to compute this, and don't even need this callback at all (except possibly to assert that it matches the meta tensor computation).

python/torch_mlir/eager_mode/torch_mlir_eager_backend.py
```diff
@@ -0,0 +1,69 @@
...
+
+    @abc.abstractmethod
+    def transfer_from_device_to_torch(self, tensor: DeviceTensor) -> torch.Tensor:
+        """If compilation fails for some reason then device specific representations need to be munged into a
+        torch.Tensor representation.
+        """
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def device_array_type(self) -> TypeVar:
```
silvasean, line 57: what is this?

python/torch_mlir/eager_mode/torch_mlir_tensor.py
```diff
@@ -2,103 +2,161 @@
...
+        Deriving classes must provide a way to construct themselves from either their device specific representation
+        or torch.Tensor; the latter is to handle the case that dispatch to PyTorch to recover from an error.
+        """
+        if isinstance(elem, torch.Tensor):
+            r = make_wrapper_subclass_from_torch_tensor(cls, elem, **kwargs)
+            r.elem = backend.transfer_from_torch_to_device(elem)
+        elif isinstance(elem, torch.nn.Parameter):
+            r = make_wrapper_subclass_from_torch_tensor(cls, elem.data, **kwargs)
+            r.elem = backend.transfer_from_torch_to_device(elem.detach().data)
+        elif isinstance(elem, backend.device_array_type()):
```
silvasean, line 78: This doesn't feel right to me. We should always know when we are wrapping a DeviceTensor -- we shouldn't need to do a brittle check on the isinstance/etc.).

silvasean (changes requested):

python/torch_mlir/eager_mode/torch_mlir_eager_backend.py
makslevental: I don't know how much of the metadata determines decisions made by the pytorch dispatcher, so I'm erring on the side of caution, in addition to having in mind the [comment that I linked](https://github.com/albanD/subclass_zoo/blob/1566e038f03cd89ab3cc37e670a44e3c2bbc1897/trivial_tensors.py#L90-L92):

>external wrapper tensor is still required to have all of the metadata that the inner tensor has; this includes **stride and storage**!

Regarding

>Bonus points if you just use meta tensors to compute this

Meta tensor coverage is "spotty", [by PyTorch's own admission](https://github.com/albanD/subclass_zoo/blob/30baf07886c8c8948d465996548295bf5f438d89/python_meta_tensor.py#L17), so this isn't feasible (even if it did give you all of the metadata).

python/torch_mlir/eager_mode/torch_mlir_tensor.py
makslevental: The alternative is just to assume you've DeviceTensor if you've fallen through the previous two cases.

python/torch_mlir/eager_mode/torch_mlir_eager_backend.py
makslevental: just another note here: without `requires_grad` you're not going to get the autodiff graph being constructed, so that one is definitely a non-negotiable.

python/torch_mlir/eager_mode/torch_mlir_tensor.py
makslevental: @silvasean coming back to this: I'm not sure how we know this for a given `elem` since we can enter this constructor from, in principle, anywhere.

python/torch_mlir_e2e_test/eager_backends/refbackend.py
```diff
@@ -0,0 +1,89 @@
...
+        return TensorMetaData(
+            size=tensor.shape,
+            dtype=NUMPY_TO_TORCH_DTYPE_DICT[tensor.dtype.type],
+            requires_grad=tensor.dtype in {np.float, np.float32, np.float64}
+            and kwargs.get("requires_grad", False),
+        )
+
+    def compile(self, imported_module: Module):
+        """Lower the imported TS module to linalg and then further compile for the reference backend and then call."""
+        module_hash = str(imported_module)
```
makslevental, line 61: poor man's compile hash

silvasean:
Hey Maks, I would like for our Eager mode to be tested by the e2e tests before doing further feature work on it.

makslevental:
> Hey Maks, I would like for our Eager mode to be tested by the e2e tests before doing further feature work on it.

We can't run through e2e tests fast enough without a compile hash, and we can't build a compile hash without this refactor.

silvasean (approved):

PR #759, opened 2022-04-15
Add shape transfer function and lowering to linalg for aten.neg

ashay:
This PR contains two commits for `aten.neg`, one for adding the shape transfer function and another for translating the torch operation to the linalg dialect.


powderluv:
Welcome to the community and thanks for the first PR

silvasean:
Looks great! Thanks for contributing :)

FYI, we would typically put both of these in a single commit. I squashed it in the github UI before merging, so no problem this time.

silvasean:
As a next step, can you add an end-to-end test in https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir_e2e_test/test_suite/elementwise.py ?

ashay:
Thanks @silvasean, here [https://github.com/llvm/torch-mlir/pull/760] is the PR for the end-to-end test.

PR #761, opened 2022-04-15
importer: add initial support for loading BFloat16 tensors

ashay:
This patch updates the `torch_mlir::convertTensorToMlirElementsAttr()`
method to enable the creation of tensors whose base type is BFloat16.
This patch also adds a test to validate the IR generation.

I admit that this patch is possibly contentious and/or incomplete.  Let me know if y'all would like to see any additions or changes.  Thanks!

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
@@ -319,6 +319,9 @@ MlirAttribute torch_mlir::convertTensorToMlirElementsAttr(at::Tensor tensor,
   case ScalarType::QInt8:
     return mlirDenseElementsAttrInt8Get(
         shapedType, numElements, static_cast<const int8_t *>(tensorData));
+  case ScalarType::BFloat16:
+    return mlirDenseElementsAttrBFloat16Get(
+        shapedType, numElements, static_cast<const uint16_t *>(tensorData));
   default:
     throwUnsupportedTensorError();
   }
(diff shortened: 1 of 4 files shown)


python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
```diff
@@ -319,6 +319,10 @@ MlirAttribute torch_mlir::convertTensorToMlirElementsAttr(at::Tensor tensor,
   case ScalarType::QInt8:
     return mlirDenseElementsAttrInt8Get(
         shapedType, numElements, static_cast<const int8_t *>(tensorData));
+  case ScalarType::BFloat16:
+    return mlirDenseElementsAttrRawBufferGet(shapedType,
```
silvasean, line 323: I think we should add mlirDenseElementsAttrBFloat16Get upstream.

test/python/importer/jit_ir/ivalue_import/bf16.py
```diff
@@ -0,0 +1,34 @@
+# -*- Python -*-
```
silvasean, line 1: I don't think we need a new file. I think we can just add `self.ones_bf16 = torch.ones(1, dtype=torch.bfloat16)` to the existing tensors.py (and if you could, can you add tests for all the other dtypes supported in convertTensorToMlirElementsAttr?)

silvasean (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
ashay: Thanks, this patch now uses the `mlirDenseElementsAttrBFloat16Get()` function from upstream LLVM.

test/python/importer/jit_ir/ivalue_import/bf16.py
ashay: Good point.  Fixed in the most recent update.

ashay:
@silvasean Requesting a re-review.  Thanks!

silvasean (approved):
:rocket:

PR #764, opened 2022-04-18
Enable support for LTC Input/Output Mapping

henrytwo:
Complements this upstream PyTorch PR: https://github.com/pytorch/pytorch/pull/75828

This PR adds support for mapping input and output tensors which alias each other. (e.g. maps input weight tensor in parameter to the same tensor in output after a training iteration)
```
MLIR: 
func @graph(%arg0: !torch.vtensor<[1,5],f32>, %arg1: !torch.vtensor<[1],si64>, ..., %arg6: !torch.vtensor<[10,5],f32>, %arg7: !torch.vtensor<[10],f32>, ...) {
  ...
  return %arg0, %arg1, %17, %23, ... : !torch.vtensor<[1,5],f32>, !torch.vtensor<[1],si64>, !torch.vtensor<[10,5],f32>, !torch.vtensor<[10],f32>, ...
}

Input/Output Alias Mapping: 
Output: 0 -> Input: 0
Output: 1 -> Input: 1
Output: 2 -> Input: 6
Output: 3 -> Input: 7
```

The `aten::detach` op has also been disabled in this PR to fix the issue of tensors not aliasing properly due to copying.

cc: @antoniojkim @ke1337

### DIFF
--- a/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.h
+++ b/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.h
@@ -41,9 +41,13 @@ public:
 
 class TORCH_API TorchMlirComputation : public torch::lazy::Computation {
 public:
+  using InputOutputAliases = LoweringContext::InputOutputAliases;
+  using InputOutputAlias = LoweringContext::InputOutputAlias;
+
   TorchMlirComputation(
       MlirOperation func_op, MlirContext mlir_context,
-      const std::shared_ptr<torch::jit::Graph>& graph);
+      const std::shared_ptr<torch::jit::Graph>& graph,
+      InputOutputAliases input_output_aliases);
 
   int parameters_size() const override;
 
@@ -67,6 +71,7 @@ private:
   MlirOperation func_op_;
   MlirContext mlir_context_;
   std::shared_ptr<torch::jit::Graph> graph_;
+  InputOutputAliases input_output_aliases_;
   unsigned num_results_;
 };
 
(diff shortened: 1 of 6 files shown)


silvasean (approved):

henrytwo:
@silvasean Keeping a map of aliased output -> input tensors is a feature that we're planning to use internally at Cerebras -- is this something you'd like us to keep in the generic `TorchMlirLoweringContext` too? Not sure if anyone else would find this useful

henrytwo:
> Isn't that something that upstream already provides or could be moved there?

Upstream we're just calling a method in the `LoweringContext` interface to save the alias, so it's up to the lowering context to figure out what to do with it. Other vendors may want to handle this differently than us (e.g. XLA), so that's why we're implementing it on the `torch-mlir` side.

python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.h
```diff
@@ -67,8 +67,12 @@ class TORCH_API TorchMlirLoweringContext : public torch::lazy::LoweringContext {
       const std::vector<int64_t>& param_index,
       bool must_alias = false) override;
 
+  // Check if parameter shape matches result at index.
+  bool CheckResultShape(
+      const BackendDataPtr& parameter_data, int64_t result_idx) override;
+
   // Get the shape of the result tuple component, given by index.
-  c10::optional<torch::lazy::Shape> GetResultShape(size_t index) const override;
+  torch::lazy::Shape GetResultShape(size_t index) const override;
```
antoniojkim, line 75: is `GetResultShape` needed anymore? can it be merged with `CheckResultShape`?

henrytwo: I guess it could, let me check to see if there are any other uses of it.

PR #766, opened 2022-04-19
[REFBACKEND] Add support for returning multiple different return types.

pashu123:
Added the dynamic registration of return function to the execution
engine. This makes sure that  different/multiple return types are supported.

### DIFF
--- a/python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
+++ b/python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
@@ -22,133 +22,85 @@ __all__ = [
 ]
 
 
-def checkArgTypeIsSupported(ty):
+def assert_arg_type_is_supported(ty):
     SUPPORTED = [np.float32, np.float64, np.int32, np.int64, np.bool_]
     assert ty in SUPPORTED, f"Only numpy arrays with dtypes in {SUPPORTED} are supported"
 
 
+memref_type_to_np_dtype = {
+    "mrf32": np.float32,
+    "mrf64": np.float64,
+    "mri1": np.bool_,
+    "mri32": np.int32,
+    "mri64": np.int64
+}
+elemental_type_to_ctype = {
+    "i1": ctypes.c_bool,
+    "i64": ctypes.c_int,
+    "f32": ctypes.c_float,
+    "f64": ctypes.c_double
+}
+
+CONSUME_RETURN_FUNC_PREFIX = "refbackend_consume_func_return_"
+
+
+def get_return_funcs(module):
+    return_prefix_len = len(CONSUME_RETURN_FUNC_PREFIX)
+    return_funcs = []
+    with module.context:
+        for func in module.body:
+            # Returns strings of the form `"refbackend.."` so `"` is deleted.
+            func_name = str(func.attributes["sym_name"]).replace('"', '')
+            if func_name[:return_prefix_len] == CONSUME_RETURN_FUNC_PREFIX:
+                return_funcs.append(func_name)
+
+    return return_funcs
+
+
+def get_ctype_func(func_name):
+    return_prefix_len = len(CONSUME_RETURN_FUNC_PREFIX)
+    ret_types = func_name[return_prefix_len:].split("_")
+    ctypes_arg = [None]
+    for type in ret_types:
+        if type in elemental_type_to_ctype:
+            ctypes_arg.append(elemental_type_to_ctype[type])
+        elif type in memref_type_to_np_dtype:
+            ctypes_arg.append(ctypes.POINTER(UnrankedMemRefDescriptor))
+        else:
+            assert False, f"Not supported type: {type}"
+
+    return ctypes.CFUNCTYPE(*ctypes_arg), ret_types
+
+
 class RefBackendInvoker:
+
     def __init__(self, module):
         self.ee = ExecutionEngine(module)
         self.result = None
 
-        @ctypes.CFUNCTYPE(None, ctypes.POINTER(UnrankedMemRefDescriptor))
-        def consume_return_mri1(a):
-            self.result = unranked_memref_to_numpy(a, np.bool_)
+        return_funcs = get_return_funcs(module)
 
-        @ctypes.CFUNCTYPE(None, ctypes.POINTER(UnrankedMemRefDescriptor))
-        def consume_return_mri32(a):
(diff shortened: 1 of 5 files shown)


pashu123:
@silvasean Do you want me to add test cases? or are the previous test cases enough?

pashu123:
> @silvasean Do you want me to add test cases? or are the previous test cases enough?

Added some tests.

python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
```diff
@@ -24,143 +24,90 @@
...
+      func_name = str(func.attributes["sym_name"])
+      if func_name[1:len(refbackend_return_initials) +
+                   1] == refbackend_return_initials:
+        return_funcs.append(func_name.replace('"', ''))
+
+  return return_funcs
+
+
+def get_ctype_func(func_name):
+  ret_types = func_name.split("_")[4:]
```
cathyzhyi, line 60: nit: it's probably better to get `4` from refbackend_consume_func_return_ rather than hardcode the number

cathyzhyi (changes requested):
I think the indentations are off. Should be 4 spaces instead of 2 spaces.

python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
silvasean: I would recommend something like:

```
>>> s = "refbackend_consume_func_return_mr32"
>>> pre = "refbackend_consume_func_return_"
>>> s.partition(pre)
('', 'refbackend_consume_func_return_', 'mr32')
```

and then split the suffix.

Would recommend `CONSUME_RETURN_FUNC_PREFIX = "refbackend_consume_func_return_"`

python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
```diff
@@ -24,143 +24,90 @@
...
+    return_funcs = get_return_funcs(module)
+
+    for ret_func in return_funcs:
+      ctype_wrapper, ret_types = get_ctype_func(ret_func)
+
+      def consume_return_funcs(*args):
+        self.result = tuple([
+            x if ret_types[i] in elemental_type_to_ctype else
+            unranked_memref_to_numpy(x, memref_type_to_np[ret_types[i]])
+            for x, i in zip(args, range(len(args)))
```
silvasean, line 88: use `enumerate`. or maybe `zip(args, ret_types)`.

python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
```diff
@@ -202,12 +149,13 @@ def invoke(*args):
 
 
 class RefBackendLinalgOnTensorsBackend(LinalgOnTensorsBackend):
-    """Main entry-point for the reference backend."""
-    def __init__(self):
-        super().__init__()
+  """Main entry-point for the reference backend."""
+
```
silvasean, line 153: please use 4 space indent

python/torch_mlir_e2e_test/test_suite/return_types.py
```diff
@@ -0,0 +1,69 @@
...
+import torch
+
+from torch_mlir_e2e_test.torchscript.framework import TestUtils
+from torch_mlir_e2e_test.torchscript.registry import register_test_case
+from torch_mlir_e2e_test.torchscript.annotations import annotate_args, export
+
+# ==============================================================================
+
+
+class TestMultipleMemrefReturn(torch.nn.Module):
```
silvasean, line 15: Memref isn't a torch concept. Call this TestMultipleTensorReturns

python/torch_mlir_e2e_test/test_suite/return_types.py
```diff
@@ -0,0 +1,69 @@
...
+def TestMultipleMemrefReturn_basic(module, tu: TestUtils):
+  module.forward(
+      tu.rand(3, 4).to(torch.float32),
+      tu.rand(2, 3).to(torch.float64),
+      tu.rand(2, 3).to(torch.int32),
+      tu.rand(2, 3).to(torch.int64),
+      tu.rand(2, 3).to(torch.bool))
+
+
+class TestMultipleMemrefAndElementalTypesReturn(torch.nn.Module):
```
silvasean, line 43: same: TestMultipleTensorAndPrimitiveTypeReturns

silvasean (changes requested):
Awesome. I just have a bunch of nits.

python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
```diff
@@ -24,143 +24,90 @@
...
+def get_ctype_func(func_name):
+  ret_types = func_name.split("_")[4:]
+  ctypes_arg = [None]
+  for type in ret_types:
+    if type in elemental_type_to_ctype:
+      ctypes_arg.append(elemental_type_to_ctype[type])
+    elif type in memref_type_to_np:
+      ctypes_arg.append(ctypes.POINTER(UnrankedMemRefDescriptor))
+    else:
+      assert True, f"Not supported type: {type}"
```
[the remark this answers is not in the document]
pashu123: > I think the indentations are off. Should be 4 spaces instead of 2 spaces.

I used this style https://github.com/llvm/torch-mlir/blob/main/.style.yapf (the indent width is mentioned to be 2)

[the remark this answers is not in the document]
pashu123: I have updated the indentation width and set it to 4.

python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
pashu123: I have diverged slightly here, but the idea is the same. Could you take a look and suggest.

python/torch_mlir_e2e_test/linalg_on_tensors_backends/refbackend.py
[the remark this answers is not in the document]
silvasean: I just updated the style file to use pep8 style https://github.com/llvm/torch-mlir/blob/main/.style.yapf

silvasean (approved):

PR #768, opened 2022-04-19
Add oneshot release snapshot for test/ondemand

powderluv:
Add some build scripts to test new release flow based on IREE.
Wont affect current builds, once this works well we can plumb it
in.


silvasean:
Can you do a pass to scrub "IREE" out of the PR (except license stuff obviously) / replace it with the appropriate Torch-MLIR thing?

Seems fine to me otherwise.

.github/workflows/buildManylinux.yml
```diff
@@ -0,0 +1,91 @@
...
+    inputs:
+      release_id:
+        description: 'Release id to upload artifacts to'
+        default: ''
+      python_package_version:
+        description: 'Version to use for creating the Python package'
+        default: ''
+
+jobs:
+  build:
```
silvasean, line 18: what is this doing that makes it "manylinux"? It looks the same as the other one.

Also, let's not do this on every pull request / push at least if it is just for testing / ondemand.

.github/workflows/oneshotSnapshotPackage.yml
```diff
@@ -0,0 +1,64 @@
+name: Release oneshot snapshot package
+
+on:
+  workflow_dispatch:
+
+jobs:
+  release_snapshot_package:
+    name: "Tag snapshot release"
```
silvasean, line 8: What is this doing that is different from releaseSnapshotPackage.yml?

.github/workflows/buildManylinux.yml
powderluv: not yet hooked up. This just sets up the oneshot so we can trigger and iterate with the dummy buildManylinux.yml action. The idea would be only push on release and when it gets called from the scheduled workflow.

.github/workflows/oneshotSnapshotPackage.yml
powderluv: same. So we can build a release snapshot on demand.

stellaraccident (commented):
Ok, I took a pass through and fixed all the issues that I saw. It now builds wheels for me. I tried installing them in a venv that had a conflicting cuda install of Torch, and it indeed failed, working after I uninstalled. I was able to import it, so things seemed minimally working.

A few sharp edges I note in TODOs in the commit:

https://github.com/llvm/torch-mlir/commit/c39a90e6e9a9be2675b8e2f6d3e3c854b4e9b000

I just pushed to the manylinux branch in the main repo because that is what I have access to. Feel free to take my commit, use that branch or whatever.

powderluv:
Builds are successful. Thanks @stellaraccident . We can now create one of release builds. In the next PR I'll switch the nightly to build manylinux

PR #769, opened 2022-04-20
Add a new `torch_mlir.compile` method.

silvasean:
This makes it much easier to convert models and hides all the
ClassAnnotator complexity.

This also adds a new example `torchscript_resnet18_all_output_types.py`
which shows the ResNet18 IR for all output types.

Also,

- This moves `run_pipeline_with_repro_report` to
  `torch_mlir.compiler_utils`.


python/torch_mlir/__init__.py
```diff
@@ -0,0 +1,94 @@
...
+    scripted = torch.jit.script(model)
+
+    if isinstance(example_args, torch.Tensor):
+        example_args = [example_args]
+
+    class_annotator = ClassAnnotator()
+    forward_annotation = [None]
+    for arg in example_args:
+        # Assume that all tensors have value semantics for now.
+        forward_annotation.append((list(arg.shape), arg.dtype, True))
```
cathyzhyi, line 68: does this mean all the arguments will be static shape?

cathyzhyi (approved):

python/torch_mlir/__init__.py
silvasean: Yes,

```
    # TODO: Support dynamic dimension sizes. See `torch.onnx.export`'s
    # `dynamic_axes` for API inspiration, or do something more ergonomic
    # like a tensor wrapper possibly.
```

PR #772, opened 2022-04-20
Pin the Python package to the exact PyTorch nightly.

silvasean:
This avoids issues where PyTorch version drift has made things
incompatible.

One caveat is that you will need to specify
`-f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
--pre` on the command line for pip to know where to find the nightly
packages (there is no way around this) -- this is easiest to do by
simultaneously passing `-r requirements.txt` on the pip command line.


powderluv (approved):
Awesome thanks

PR #774, opened 2022-04-21
[LINALG] Bug fix i64 vs i32 type comparison.

pashu123:
Comparing index type instead of integer types solves the problem.


cathyzhyi (approved):
Thanks for the fix! Since we are here already, can you help to rename `castIndexToInt ` to `castIndexToInt64`?

nirvedhmeshram (approved):

PR #776, opened 2022-04-21
Enable OSX (Intel, Apple Silicon Builds)

powderluv:
Update pinned pytorch version. Will submit a follow on PR to bump.
Also update artifacts directory


powderluv:
not plumbed into anything yet testing the build / upload etc.

Issue #138, opened 2020-12-07
Problems with Torch graph bindings


silvasean:
Running into an issue

```
silvasean@silvasean0:~/pg/mlir-npcomp/mlir-npcomp$ source .env; python frontends/pytorch/test/graph_export/test_script_add3.py
Traceback (most recent call last):
  File "frontends/pytorch/test/graph_export/test_script_add3.py", line 21, in <module>
    def add3(t0, t1, t2):
TypeError: import_function(): incompatible function arguments. The following argument types are supported:
    1. (self: _torch_mlir.ModuleBuilder, arg0: torch::jit::StrongFunctionPtr) -> torch::jit::StrongFunctionPtr

Invoked with: <_torch_mlir.ModuleBuilder object at 0x7f81c3e8d0b0>, <torch.jit.ScriptFunction object at 0x7f81a8cb36d0>
```

1. For some reason, on my system pybind11.h is in two places, and despite our efforts in the CMakeLists.txt file the non-torch one seems to get used for the torch bindings

```
/usr/local/google/home/silvasean/.local/lib/python3.8/site-packages/pybind11/include/pybind11/pybind11.h
/usr/local/google/home/silvasean/.local/lib/python3.8/site-packages/torch/include/pybind11/pybind11.h
```

2. Even after I fix that by manually commenting out the necessary CMakeLists.txt lines, I get the same issue. Unclear why.

stellaraccident:
Ok, I created an isolated repro of the issue and tracked down the root cause. Repro repository here: https://github.com/stellaraccident/ptbindrepro

For these cross module types to work, the computed ABI string for pybind11 using modules must match. The main variable keys that go into the ABI string are (see detail/internals.h):
  `PYBIND11_COMPILER_TYPE`: (_gcc vs _clang)
  `PYBIND11_STDLIB`: (_libcpp vs _libstdcpp)
  `PYBIND11_BUILD_ABI` (`__GXX_ABI_VERSION` on Linux): Typically `cxxabi1011` on official PyTorch binary releases and `cxxabi1013` on my machine.

These are all overridable based on defines for cases where you need to match something else and believe it to be safe.

I wrote [a little script](https://github.com/stellaraccident/ptbindrepro/blob/master/detect_mismatched_pybind.py) that figures this out. For example `detect_mismatched_pybind.py torch ptbindrepro` on my sample repro prints this (with the FIX_FLAG commented out):

```
Importing torch...
  Current pybind11_internals:
[...]

PR #777, opened 2022-04-21
Enhance PyTorch C++ ABI compatibility checking

rdadolf:
Compiling torch-mlir against a source version of PyTorch or an official wheel compiled with the new C++ stdlib ABI fails, as torch-mlir doesn't know how to set compiler flags to remain compatible. This changes the way torch-mlir looks at PyTorch and tries to more closely match the ABI settings, regardless of whether it's the common official wheel or some other version.

Tagging @stellaraccident due to your work on #138.

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/cmake/modules/TorchMLIRPyTorch.cmake
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/cmake/modules/TorchMLIRPyTorch.cmake
@@ -25,37 +25,62 @@ function(TorchMLIRProbeForPyTorchInstall)
 endfunction()
 
 # TorchMLIRConfigurePyTorch
-# Performs configuration of PyTorch flags after CMake has found it to be
-# present. Most of this comes down to detecting whether building against a
-# source or official binary and adjusting compiler options in the latter case
-# (in the former, we assume that it was built with system defaults). We do this
-# conservatively and assume non-binary builds by default.
+# Extensions compiled against PyTorch must be ABI-compatible with PyTorch.
+# On Linux, there are two components to this:
+#   1) Dual ABI settings for libstdc++
+#      See https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html
+# For this, PyTorch helpfully provides a function to check which ABI it was
+# compiled against.
+#   2) C++ ABI compatibility version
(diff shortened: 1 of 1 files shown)


powderluv:
Thanks for your contribution @rdadolf 

Is there any reason we want to support the non -D_GLIBCXX_USE_CXX11_ABI= flow? Could we just simply depend on the the version using CXX11 ABI ?

rdadolf:
> Thanks for your contribution @rdadolf
> 
> Is there any reason we want to support the non -D_GLIBCXX_USE_CXX11_ABI= flow? Could we just simply depend on the the version using CXX11 ABI ?

PyTorch uses the old ABI by default, so if you `pip install` it, you'll get a binary that was compiled with `_GLIBCXX_USE_CXX11=0`. My understanding is that this is due to the age of manylinux, which stretches back to include a gcc version that didn't have the new C++ standard library capability. Since manylinux is required by pypa, PyTorch is forced to use the old one. This is actually an [ongoing issue with PyTorch](https://github.com/pytorch/pytorch/issues/17492).

The C++ ABI compatibility version flag (1011, 1013, etc.) is separate from that.

stellaraccident:
Thanks for digging into this. If you ever see a way to improve this and not rely on the old way, we are just doing our best to match PT and broad Linux compatibility -- and there are no strong preferences beyond that. I look forward to us being firmly in the modern era on this but the last I evaluated the interlocking pieces, I guess we may be still a year or two out :/

rdadolf:
I did also look into the status of the Torch machinery (as you mentioned in those comments), but PyTorch explicitly doesn't support clang on Linux (and are aggressive about checking for it, e.g.: https://github.com/pytorch/pytorch/blob/master/torch/utils/cpp_extension.py#L267). It's possible to force the issue and bypass the check, but we'd still need to generate our own flags for clang, which would leave us more or less back with the current approach. Using `torch.utils.cpp_extension` for gcc compilers and our own thing for clang compilers seems pretty sub-optimal.

if the PyTorch folks decide to expand support, we could revisit the issue. Hence it seemed reasonable to leave your original comments about the future in---they still apply in pretty much the same way as before.

stellaraccident:
Yeah - I think the clang on Linux bit only comes for development: for real builds, we use system compilers (gcc). I'm pretty sketched out in general about deviating because when these things break, it is super subtle and a support nightmare. Personally, I always prefer aligning to whatever upstream convention is in play.

PR #778, opened 2022-04-21
Minor buildsystem fixes

powderluv:
Sets up auto-pinning of latest torch-nightly


silvasean:
What is "auto" about this pinning?

powderluv:
By default if you use "torch" you get torch-cuda which causes build issues. @stellaraccident fixed it with "torch==1.12.0.dev20220420" but I found a workaround with "torch>=1.12.0.dev20220420", that forces getting the CPU version without having to specify the look URL in the requires section.

PR #779, opened 2022-04-21
Some fixes for `torch_mlir.compile` on the MiniLM model.

silvasean:
Recommended review order: commit by commit.


silvasean:
Sorry, abandoning this for now. I need to think more about the refinement/derefinement along function boundaries.

PR #780, opened 2022-04-21
Switch to using the new manylinux+macos Release builds

powderluv:
This switches to using the "Release Build" that builds linux/osx
binaries for the build and release of snapshots.
Moves buildAndTest.yml to be only on CI build + test.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -6,13 +6,6 @@ on:
       - main
   pull_request:
   workflow_dispatch:
-    inputs:
-      release_id:
-        description: 'Release id to upload artifacts to'
-        default: ''
-      python_package_version:
-        description: 'Version to use for creating the Python package'
-        default: ''
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/buildAndTest.yml
```diff
@@ -6,19 +6,12 @@ on:
...
-      python_package_version:
-        description: 'Version to use for creating the Python package'
-        default: ''
 
 jobs:
   build:
     name: Build and Test (Release Asserts)
     # Changes to the name of this job needs to be synced with releaseSnapshotPackage.yml.
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
stellaraccident, line 14: Lets leave this as 20.04. You want to upgrade it on your own time, not because GitHub pushed a new release one morning!

stellaraccident: (and below)

stellaraccident (approved):

.github/workflows/buildAndTest.yml
powderluv: done

PR #781, opened 2022-04-22
llvm: bump tag to e1318078

ashay:
The updated LLVM code includes a patch to create bfloat16 array
attributes, thus enabling a different patch to torch-mlir to flesh out
support for the bfloat16 type.


ashay:
I see, thanks @cathyzhyi!  I'll keep an eye out for those two changes.  In the meantime, I'll patch some more fixes.

cathyzhyi:
The CI are passing now! Thanks a lot for doing this @ashay

cathyzhyi (approved):

ashay:
Thank you @cathyzhyi for the upstream fix!

ashay:
@cathyzhyi Did this change break the main branch bazel build [https://github.com/llvm/torch-mlir/runs/6182181576?check_suite_focus=true]?  Are there any tests I should have run before merging?

powderluv:
I think the Bazel build is informational. @asaadaldien

silvasean:
Correct, the bazel build is allowed to be broken by day-to-day development. It is up to the subcommunity maintaining it to make it green again (we are trying to be consistent with LLVM's policies w.r.t. alternative build systems).

PR #788, opened 2022-04-25
Fix version information in Release builds

powderluv:
env vars seems to be lost in manylinux docker. Use a version file like IREE does.

### DIFF
--- a/.github/workflows/buildRelease.yml
+++ b/.github/workflows/buildRelease.yml
@@ -26,6 +26,8 @@ jobs:
       run: |
         cd $GITHUB_WORKSPACE
         python -m pip install wheel
+        TM_PACKAGE_VERSION=${{ github.event.inputs.python_package_version }}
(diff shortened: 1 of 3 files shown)


.github/workflows/buildRelease.yml
```diff
@@ -26,6 +26,8 @@ jobs:
       run: |
         cd $GITHUB_WORKSPACE
         python -m pip install wheel
+        TM_PACKAGE_VERSION=${{ github.event.inputs.python_package_version }}
+        printf "TORCH_MLIR_PYTHON_PACKAGE_VERSION=%s\n" $TM_PACKAGE_VERSION > ./tm_pkg_version
```
silvasean, line 30: can you call this torch_mlir_package_version. No need to abbreviate.

silvasean (approved):

.github/workflows/buildRelease.yml
powderluv: done

Issue #789, opened 2022-04-25
eager mode tests started to fail from build 410


powderluv:
https://github.com/llvm/torch-mlir/runs/6149438281?check_suite_focus=true

```
FAIL: TORCH_MLIR_PYTHON :: eager_mode/annotate_args.py (2 of 11)
******************** TEST 'TORCH_MLIR_PYTHON :: eager_mode/annotate_args.py' FAILED ********************
Script:
--
: 'RUN: at line 6';   /opt/hostedtoolcache/Python/3.9.12/x64/bin/python /home/runner/work/torch-mlir/torch-mlir/python/test/eager_mode/annotate_args.py | /home/runner/work/torch-mlir/torch-mlir/build/bin/FileCheck /home/runner/work/torch-mlir/torch-mlir/python/test/eager_mode/annotate_args.py
--
Exit Code: 1

Command Output (stderr):
--
/home/runner/work/torch-mlir/torch-mlir/python/test/eager_mode/annotate_args.py:19:10: error: CHECK: expected string not found in input
# CHECK: Torch Tensor (shape=(1, 3, 32, 32), dtype=torch.float32)
         ^
<stdin>:1:1: note: scanning from here
FAIL - simple
^
<stdin>:2:12: note: possible intended match here
Errors: Couldn't normalize args and kwargs
           ^

Input file: <stdin>
Check file: /home/runner/work/torch-mlir/torch-mlir/python/test/eager_mode/annotate_args.py

-dump-input=help explains the following input dump.

Input was:
<<<<<<
            1: FAIL - simple 
check:19'0     X~~~~~~~~~~~~~ error: no match found
            2: Errors: Couldn't normalize args and kwargs 
check:19'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
check:19'1                ?                                possible intended match
            3:  
check:19'0     ~
            4: FAIL - handle_zero_dim 
check:19'0     ~~~~~~~~~~~~~~~~~~~~~~~
            5: Errors: Couldn't normalize args and kwargs 
check:19'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            6:  
check:19'0     ~
            7: FAIL - correctly_order_kwargs 
check:19'0     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
            .
            .
            .
>>>>>>

--

********************
Testing:  0.. 10.. 20.
[...]

PR #790, opened 2022-04-25
Add to eager tests to xfail while they are fixed

powderluv:
Issue tracked at https://github.com/llvm/torch-mlir/issues/789

### DIFF
--- a/python/test/eager_mode/annotate_args.py
+++ b/python/test/eager_mode/annotate_args.py
@@ -21,7 +21,8 @@ from torch_mlir.eager_mode.torch_mlir_dispatch import (
 # CHECK: Torch Tensor (shape=(1, 3, 32, 32), dtype=torch.float32)
 # -----
 # CHECK: PASS - simple
-@run_test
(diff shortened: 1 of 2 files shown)


silvasean:
LGTM

python/test/eager_mode/annotate_args.py
```diff
@@ -21,7 +21,8 @@
 # CHECK: Torch Tensor (shape=(1, 3, 32, 32), dtype=torch.float32)
 # -----
 # CHECK: PASS - simple
```
silvasean, line 23: I think you have to change this line to XFAIL too

silvasean (approved):

python/test/eager_mode/annotate_args.py
makslevental: Yes that's correct (and a weakness of this testing harness that I should think about).

powderluv:
taken over in https://github.com/llvm/torch-mlir/pull/793

PR #791, opened 2022-04-25
[MLIR][Torch] Add e2e support for `div.float`, `ge.float`, `ge.float_int`, `ne.float_int` op

vivekkhandelwal1:


cathyzhyi (changes requested):
looks good. Can you add some FileCheck tests for the folders?

vivekkhandelwal1:
> looks good. Can you add some FileCheck tests for the folders?

Done.

pashu123:
@cathyzhyi I have a minor concern about this PR with 4 commits of trivial ops. Shouldn't we have all the commits squashed up with proper PR description?

silvasean:
No need to squash here. Actually this is more close to the "ideal" LLVM development standard of factored, fine-grained commits. In Torch-MLIR we usually do larger commits because we aren't as strict and move faster than LLVM core, but there's no downside to having it like this.

pashu123:
> No need to squash here. Actually this is more close to the "ideal" LLVM development standard of factored, fine-grained commits. In Torch-MLIR we usually do larger commits because we aren't as strict and move faster than LLVM core, but there's no downside to having it like this.

Sounds good. I was aware of one PR one commit philosophy unless it's a major change. But your point makes sense.

pashu123 (approved):
LGTM!

vivekkhandelwal1:
> No need to squash here. Actually this is more close to the "ideal" LLVM development standard of factored, fine-grained commits. In Torch-MLIR we usually do larger commits because we aren't as strict and move faster than LLVM core, but there's no downside to having it like this.

Yeah, that's why I have factored the PR into four commits which takes a little more effort than squashing and keeping them in a single commit.

vivekkhandelwal1:
@cathyzhyi, should I merge this?

cathyzhyi (approved):
yes, please go ahead and merge.

PR #792, opened 2022-04-25
[LINALG] Decompose static aten.matmul into linalg.batchmatmul.

pashu123:
Decomposes static 4d and above cases of batch_matmul into linalg.batchmatmul
with proper collapsing and expansion.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -105,11 +105,15 @@ public:
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
       return failure();
 
-    unsigned lhsRank = lhs.getType().cast<RankedTensorType>().getRank();
-    unsigned rhsRank = rhs.getType().cast<RankedTensorType>().getRank();
+    auto lhsType = lhs.getType().cast<RankedTensorType>();
+    auto rhsType = rhs.getType().cast<RankedTensorType>();
+
+    unsigned lhsRank = lhsType.getRank();
+    unsigned rhsRank = rhsType.getRank();
 
     Type newResultType = getTypeConverter()->convertType(op.getType());
-    Type elementType = newResultType.cast<TensorType>().getElementType();
+    auto resultType = newResultType.cast<RankedTensorType>();
+    Type elementType = resultType.getElementType();
 
     // The different cases of torch_matmul op is mentioned here:
     // https://pytorch.org/docs/stable/generated/torch.matmul.html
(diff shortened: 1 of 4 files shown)


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -167,11 +171,89 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
...
+
+        auto collapsedResultType =
+            RankedTensorType::get(collapsedResultShape, elementType);
+
+        // Collapse the batch dimension into one dimension. The resultant rank
+        // will always be 3.
+        SmallVector<ReassociationIndices> reassociation(3);
+
+        for (unsigned i = 0, j = 0; i < lhsRank; i++) {
+          if (i < batchRank) {
```
cathyzhyi, line 230: nit: can change to
```
if (i >= batchRank)
    j++;
reassociation[j].push_back(i);
```

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -167,11 +171,89 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
...
+          } else {
+            j += 1;
+            reassociation[j].push_back(i);
+          }
+        }
+        Value collapsedLhs = rewriter.create<tensor::CollapseShapeOp>(
+            op->getLoc(), lhs, reassociation);
+        Value collapsedRhs = rewriter.create<tensor::CollapseShapeOp>(
+            op->getLoc(), rhs, reassociation);
+        Value initTensor0 = createStaticZeroInitTensor(
```
cathyzhyi, line 241: nit: we could just use `createZeroInitTensor` here with `getAsOpFoldResult(getTensorSize` similar to this code
https://github.com/llvm/torch-mlir/blob/cc3a4a58ef7f8390d30b08c6d1b5b6f99a70b85e/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp#L413-L415

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -167,11 +171,89 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
...
+            rewriter, loc, collapsedResultShape, elementType);
+
+        Value batchMatMul =
+            rewriter
+                .create<linalg::BatchMatmulOp>(
+                    loc, collapsedResultType,
```
cathyzhyi, line 247: this could be replaced with `initTensor0.getType` and the calculation for `collapsedResultType` could be removed. This would simplify future changes to support dynamic shapes.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -167,11 +171,89 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
...
 
+    // Fourth Case: When rhs rank and lhs rank is 3, it can be lowered to 
+    // linalg.batch_matmul op.
+    if (lhsRank == 3 && rhsRank == 3) {
+      Value batchDim = getDimOp(rewriter, loc, lhs, 0);
+      Value lhsDim0 = getDimOp(rewriter, loc, lhs, 1);
+      Value lhsDim1 = getDimOp(rewriter, loc, lhs, 2);
+      Value rhsDim0 = getDimOp(rewriter, loc, rhs, 1);
+      Value rhsDim1 = getDimOp(rewriter, loc, rhs, 2);
+      checkDimEqualHelper(rewriter, loc, lhsDim1, rhsDim0);
```
cathyzhyi, line 182: are batch dim required to be the same as well?

cathyzhyi (changes requested):

raikonenfnu:
Can we add mlir-opt + FileCheck test to ensure that the conversion is doing exactly what we want?

PR #796, opened 2022-04-26
[MLIR][TORCH] Add E2E support for `aten.ScalarImplicit` op

vivekkhandelwal1:
This commit adds lowering of `aten.ScalarImplicit` op.

### DIFF
--- a/lib/Dialect/Torch/Transforms/RefineTypes.cpp
+++ b/lib/Dialect/Torch/Transforms/RefineTypes.cpp
@@ -426,6 +426,9 @@ private:
   ChangeResult
   visitBinaryScalarOp(Operation *op,
                       ArrayRef<LatticeElement<ValueKnowledge> *> operands);
+  ChangeResult visitAtenScalarImplicitOp(
+      AtenScalarImplicitOp op,
+      ArrayRef<LatticeElement<ValueKnowledge> *> operands);
 };
 } // namespace
 
@@ -982,6 +985,9 @@ ChangeResult TypeAnalyzer::visitOperation(
     return visitBinaryScalarOp(op, operands);
   }
 
+  if (auto scalarImplicit = dyn_cast<AtenScalarImplicitOp>(op))
+    return visitAtenScalarImplicitOp(scalarImplicit, operands);
+
   // Otherwise, this is an unknown operation. Just mark all results as
   // having reached a pessimistic fixpoint.
   return markAllPessimisticFixpoint(op->getResults());
@@ -1247,6 +1253,19 @@ ChangeResult TypeAnalyzer::visitAten_SoftmaxLikeOp(
         halfToFloat ? Float32Type::get(op->getContext()) : input.dtype;
   }
   return incorporateKnowledge(op.getResult(), knowledge);
+}
+
+ChangeResult TypeAnalyzer::visitAtenScalarImplicitOp(
+    AtenScalarImplicitOp op,
+    ArrayRef<LatticeElement<ValueKnowledge> *> operands) {
+  auto knowledge =
+      ValueKnowledge::getScalarPessimisticValueState(op.getContext());
+  Type dType = operands[0]->getValue().dtype;
+  if (dType.isa<mlir::FloatType>())
+    knowledge.setScalarType(Torch::FloatType::get(op->getContext()));
+  else if (dType.isa<mlir::IntegerType>())
+    knowledge.setScalarType(Torch::IntType::get(op->getContext()));
+  return incorporateKnowledge(op->getResult(0), knowledge);
 }
 
 // -----------------------------------------------------------------------------
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -6158,6 +6158,30 @@ def Torch_AtenIntFloatOp : Torch_Op<"aten.Int.float", [
   }];
 }
 
+def Torch_AtenIntScalarOp : Torch_Op<"aten.Int.Scalar", [
+    AllowsTypeRefinement,
+    HasValueSemantics,
+    ReadOnly
+  ]> {
+  let summary = "Generated op for `aten::Int.Scalar : (Scalar) -> (int)`";
+  let arguments = (ins
+    AnyTorchScalarType:$a
+  );
+  let results = (outs
+    Torch_IntType:$result
+  );
+  let hasCustomAssemblyFormat = 1;
+  let extraClassDefinition = [{
+    ParseResult AtenIntScalarOp::parse(OpAsmParser &parser, OperationState &result) {
+      return parseDefaultTorchOp(parser, result, 1, 1);
+    }
+    void AtenIntScalarOp::print(OpAsmPrinter &printer) {
+      printDefaultTorchOp(printer, *this, 1, 1);
+    }
+  }];
+  let hasFolder = 1;
+}
+
 def Torch_Aten__RangeLengthOp : Torch_Op<"aten.__range_length", [
     AllowsTypeRefinement,
     HasValueSemantics,
@@ -7137,6 +7161,29 @@ def Torch_AtenCeilFloatOp : Torch_Op<"aten.ceil.float", [
     }
   }];
   let hasFolder = 1;
+}
+
+def Torch_AtenScalarImplicitOp : Torch_Op<"aten.ScalarImplicit", [
+    AllowsTypeRefinement,
+    HasValueSemantics,
+    ReadOnly
+  ]> {
+  let summary = "Generated op for `aten::ScalarImplicit : (Tensor) -> (Scalar)`";
+  let arguments = (ins
+    AnyTorchTensorType:$a
+  );
+  let results = (outs
+    AnyTorchScalarType:$result
+  );
+  let hasCustomAssemblyFormat = 1;
+  let extraClassDefinition = [{
+    ParseResult AtenScalarImplicitOp::parse(OpAsmParser &parser, OperationState &result) {
+      return parseDefaultTorchOp(parser, result, 1, 1);
+    }
+    void AtenScalarImplicitOp::print(OpAsmPrinter &printer) {
+      printDefaultTorchOp(printer, *this, 1, 1);
+    }
+  }];
 }
 
 def Torch_Aten_SoftmaxBackwardDataOp : Torch_Op<"aten._softmax_backward_data", [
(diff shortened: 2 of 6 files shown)


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1643,6 +1643,21 @@ class ConvertTensorStaticInfoCastOp
...
 
+namespace {
+class ConvertAtenScalarImplicitOp
+    : public OpConversionPattern<AtenScalarImplicitOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenScalarImplicitOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Value out = rewriter.create<tensor::ExtractOp>(op.getLoc(), adaptor.a());
```
silvasean, line 1654: replaceOpWithNewOp?

silvasean (approved):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
vivekkhandelwal1: The op is failing with the following error message:
```
Unexpected outcome summary:

****** Failed tests - 2 tests
    FAIL - "ScalarImplicitFloatModule_basic"
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 248, in run_tests
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/linalg_on_tensors_backend.py", line 41, in compile
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/compiler_utils.py", line 49, in run_pipeline_with_repro_report
            raise Exception(f"""
        Exception: 
        Lower Torch Backend IR -> Linalg-on-Tensors Backend IR failed with the following diagnostics:
        error: failed to materialize conversion for result #0 of operation 'torch.aten.ScalarImplicit' that remained live after conversion
        note: see current operation: %3 = "torch.aten.ScalarImplicit"(%arg0) : (!torch.vtensor<[],f32>) -> !torch.number
        note: see existing live user here: return %3 : !torch.number


        Error can be reproduced with:
        $ torch-mlir-opt -pass-pipeline='torch-backend-to-linalg-on-tensors-backend-pipeline' /tmp/ScalarImplicitFloatModule.mlir
        Add '-print-ir-after-all -mlir-disable-threading' to get the IR dump for debugging purpose.
```

The result type of this op is a `torch.number`. And the result of the extract op is either an integer or float. Is this the reason behind the failure of e2e tests?

vivekkhandelwal1: @cathyzhyi @silvasean, any suggestions on how to fix this issue?

vivekkhandelwal1: That would be really helpful.

cathyzhyi: FYI, scalar type propagation is supported now in #829. You will need to write a new visit function for aten.ScalarImplicit which reads the `ValueKnowledge.dtype` from the input tensor and fill in the appropriate type into  the `ValueKnowledge.scalarType` of the new knowledge.

cathyzhyi: FYI, scalar type propagation is supported now in #829. You will need to write a new visit function for aten.ScalarImplicit which reads the `ValueKnowledge.dtype` from the input tensor and fill in the appropriate type into  the `ValueKnowledge.scalarType` of the new knowledge.

vivekkhandelwal1: I have added the following code, but that's also giving the error mentioned in the above comment.
```
ChangeResult TypeAnalyzer::visitAtenScalarImplicitOp(
    AtenScalarImplicitOp op,
    ArrayRef<LatticeElement<ValueKnowledge> *> operands) {
  auto knowledge =
      ValueKnowledge::getTensorPessimisticValueState(op.getContext());
  BaseTensorType type = op.a().getType().cast<BaseTensorType>();
  if (type.getDtype().isa<mlir::FloatType>())
    knowledge.scalarType = Torch::FloatType::get(op->getContext());
  else
    knowledge.scalarType = Torch::IntType::get(op->getContext());
  return incorporateKnowledge(op.getResult(), knowledge);
}
```

vivekkhandelwal1: I have updated the PR, if someone wants to try, they can do that. @cathyzhyi

cathyzhyi: the result of `aten.ScalarImplicit` is a scalar type you need use `getScalarPessimisticValueState` to get the initial value.

cathyzhyi:
Note that the method return type can't be `torch.number` so the forward function needs to cast to float or int before return.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1219,6 +1223,20 @@ ChangeResult TypeAnalyzer::visitAten_SoftmaxLikeOp(
   return incorporateKnowledge(op.getResult(), knowledge);
 }
 
+ChangeResult TypeAnalyzer::visitAtenScalarImplicitOp(AtenScalarImplicitOp op) {
+  auto knowledge =
+      ValueKnowledge::getScalarPessimisticValueState(op.getContext());
+  BaseTensorType type = op.a().getType().cast<BaseTensorType>();
+  if (type.getDtype().isa<mlir::FloatType>()) {
+    knowledge.scalarType = type.getDtype();
```
cathyzhyi, line 1231: scalarType should be torch type here.

cathyzhyi (changes requested):

lib/Dialect/Torch/Transforms/RefineTypes.cpp
cathyzhyi: Can you also add unit tests here test/Dialect/Torch/refine-types-ops.mlir?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1655,6 +1655,20 @@ class ConvertTensorStaticInfoCastOp
...
 
+namespace {
+class ConvertAtenScalarImplicitOp
+    : public OpConversionPattern<AtenScalarImplicitOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenScalarImplicitOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    rewriter.replaceOpWithNewOp<tensor::ExtractOp>(op, adaptor.a());
```
cathyzhyi, line 1666: To be safe, need to add `CastOp` to cast back to converted resulted type.

cathyzhyi: oh, nvm, it's not needed.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
vivekkhandelwal1: Here also https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L1068-L1077 the scalarType is a builtin type.

cathyzhyi: ha, I see. yea, those needs to be fixed as well.

cathyzhyi: FYI, the fix and `setScalarType` are added in #838. Whichever PR merged later will need to solve the conflicts.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1249,6 +1253,20 @@ ChangeResult TypeAnalyzer::visitAten_SoftmaxLikeOp(
...
 }
 
+ChangeResult TypeAnalyzer::visitAtenScalarImplicitOp(AtenScalarImplicitOp op) {
+  auto knowledge =
+      ValueKnowledge::getScalarPessimisticValueState(op.getContext());
+  BaseTensorType type = op.a().getType().cast<BaseTensorType>();
+  if (type.getDtype().isa<mlir::FloatType>()) {
+    knowledge.setScalarType(Torch::FloatType::get(op->getContext()));
+    knowledge.kind = torch_upstream::TypeKind::FloatType;
+  } else {
```
cathyzhyi, line 1263: it's safer to check for integer type explicitly on this branch.

cathyzhyi (changes requested):

vivekkhandelwal1:
@cathyzhyi Why's the test `CeilFloatModule` failing on CI? I checked it locally and it passed.

cathyzhyi:
> @cathyzhyi Why's the test `CeilFloatModule` failing on CI? I checked it locally and it passed.

It's because of a nightly torch update. see https://github.com/llvm/torch-mlir/issues/842

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1249,6 +1253,17 @@ ChangeResult TypeAnalyzer::visitAten_SoftmaxLikeOp(
   return incorporateKnowledge(op.getResult(), knowledge);
 }
 
+ChangeResult TypeAnalyzer::visitAtenScalarImplicitOp(AtenScalarImplicitOp op) {
+  auto knowledge =
+      ValueKnowledge::getScalarPessimisticValueState(op.getContext());
+  BaseTensorType type = op.a().getType().cast<BaseTensorType>();
+  if (type.getDtype().isa<mlir::FloatType>())
```
cathyzhyi, line 1260: sry, I missed this in the previous review. The dtype needs to come from `ArrayRef<LatticeElement<ValueKnowledge> *> operands` because the lattice value can have more concrete type information than the operand value itself. Please refer to the other visit functions in this file.

cathyzhyi (changes requested):

cathyzhyi (approved):

PR #797, opened 2022-04-26
Fix ConstantPad2dStaticModule test.

gprateek93:
This commit fixes the `ConstantPad2dStaticModule` test case by adding
the lowering of `aten.pad` operation. Previously the test case
mapped to `aten.constant_pad_nd` operation.
The `aten.pad` now decomposes into `aten.constant_pad_nd` operation.


gprateek93:
> Please add the decompose_complex_ops test.

Thanks! Added the test.

pashu123:
> > Please add the decompose_complex_ops test.
> 
> Thanks! Added the test.

I think you also need to add an e2e test of pad op.

gprateek93:
> > > Please add the decompose_complex_ops test.
> > 
> > 
> > Thanks! Added the test.
> 
> I think you also need to add an e2e test of pad op.

It is already added. That is what this PR fixes. Previously this test case lowered to `aten.constant_pad_nd` but due to latest pytorch version, now it lowers to `aten.pad` operation.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1646,6 +1646,25 @@ class DecomposeAtenNewEmptyOp : public OpRewritePattern<AtenNewEmptyOp> {
...
 
+namespace {
+// Decompose `aten.pad` op into `aten.constant_pad_nd` op.
+class DecomposeAtenPadOp : public OpRewritePattern<AtenPadOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenPadOp op,
+                                PatternRewriter &rewriter) const override {
+
+    Value value = op.value();
+    if (!value)
```
cathyzhyi, line 1657: should check for NoneType here.

Also, need to guard for OptionalType. This is because we can't really handle optional type but only NoneType or FloatType.

cathyzhyi (changes requested):
Shall we change the e2e tests to use the `torch.ops.aten.` operators directly? We could also test the padding value None case.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gprateek93: Added.

silvasean:
Can you summarize why this test suddenly started failing? Was it an upstream change?

pashu123:
> Can you summarize why this test suddenly started failing? Was it an upstream change?

I think it's failing because of `"GeFloatModule_basic"` which you have already reverted. Rebasing the patch would fix the issue.

silvasean:
I mean why did ConstantPad2dStaticModule start failing in the first place (why did we have to fix it?)

gprateek93:
> I mean why did ConstantPad2dStaticModule start failing in the first place (why did we have to fix it?)

Yes it started failing due to upstream change. Previously the `torch.nn.ConstantPad2d` op matched to `aten.constant_pad_nd` operation, but after the introduction of `aten.pad` recently, the op matched to it.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1664,6 +1665,25 @@ class DecomposeAtenIndexPutHackedTwinOp
...
 } // namespace
 
+namespace {
+// Decompose `aten.pad` op into `aten.constant_pad_nd` op.
+class DecomposeAtenPadOp : public OpRewritePattern<AtenPadOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenPadOp op,
+                                PatternRewriter &rewriter) const override {
+
+    Value value = op.value();
```
cathyzhyi, line 1675: need to guard for OptionalType as well. If it's OptionalType need to report error.

cathyzhyi (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gprateek93: Added.

PR #801, opened 2022-04-26
Add a script to build and upload M1 snapshots

powderluv:
Uses the latest snapshot tags and adds the releases to same asset
directories so it can be run on a cronjob without a GH runner.

### DIFF
--- a/build_tools/python_deploy/build_upload_m1_snapshot.sh
+++ b/build_tools/python_deploy/build_upload_m1_snapshot.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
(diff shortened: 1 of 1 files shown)


build_tools/python_deploy/build_upload_m1_snapshot.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+#
+# Licensed under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# build_upload_m1_snapshot.sh
+# This is a wrapper to build_macos_packages.sh to be run on Apple M1 systems
+# since GH Actions don't support M1 runners yet and Universal builds
+# dont work for torch-mlir since we don't have universal PyTorch binaries
```
silvasean, line 10: nit: dont -> don't, and Torch-MLIR instead of torch-mlir.

silvasean (approved):

PR #803, opened 2022-04-27
Fix bazel build post llvm-project bump

asaadaldien:


ashay (approved):
LGTM!  Checked that `bazel build @torch-mlir//...` now runs without errors.

PR #805, opened 2022-04-27
Ensure that imported function input type and block arg types are consistent

silvasean:
Ensure that imported function input type and block arg types are consistent.

I wasn't able to find exactly what frontend situation created it, but
`torch.jit.trace` will sometimes create functions where the
`jit::Block`'s param node has refined tensor types. So we need to adjust
the function's formal param types to those refined types.


silvasean:
I found this when using `torch_mlir.compile` on the BERT model.

cathyzhyi (approved):

PR #806, opened 2022-04-27
[Bazel][Fix] Add missing dependency

asaadaldien:


vivekkhandelwal1 (approved):
Thanks for adding the missing dependency.

PR #809, opened 2022-04-28
Add support for lowering AtenMeanDimOp

ashay:
This PR contains commits for adding support for lowering AtenMeanDimOp operations.  Although this PR contains multiple commits (to aid readability), I will squash them into a single commit before merging, if approved.

Perhaps the last two commits in this PR are the best places to start to understand the key approach of these commits.

### DIFF
--- a/e2e_testing/torchscript/xfail_sets.py
+++ b/e2e_testing/torchscript/xfail_sets.py
@@ -153,4 +153,5 @@ TOSA_PASS_SET = {
     "ConvolutionModule2DStatic_basic",
     "ElementwiseNegModule_basic",
     "TestMultipleTensorReturn_basic",
+    "ReduceMeanDimMultiModule_basic",
 }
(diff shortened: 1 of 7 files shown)


silvasean:
Meta-question: @ashay, are you trying to compile a particular model and finding these ops missing? If it is open-source, I would love if we could add it to https://github.com/llvm/torch-mlir/tree/main/build_tools/torchscript_e2e_heavydep_tests so it is easy for us to collaborate on it / burn down the issues.

silvasean (changes requested):

Shukla-Gaurav:
Hi @ashay, `aten.mean.dim` op has been addressed here https://github.com/llvm/torch-mlir/pull/613. It's been quite long(I should have fixed that long ago). In that PR the op has been addressed in as a part of decomposition. How should we go ahead?

ashay:
Closing because #613 addresses adds the required support.

Issue #807, opened 2022-04-27
`GeFloatModule_basic` test has started to fail. 


powderluv:
Unexpected outcome summary:

****** Failed tests - 1 tests
    FAIL - "GeFloatModule_basic"
        @ trace item #0 - call to "forward"
        @ output of call to "forward"
        ERROR: value (False) is not equal to golden value (True)

Summary:
    Passed: 466
    Failed: 1
    Expectedly Failed: 8

https://github.com/llvm/torch-mlir/runs/6198970107?check_suite_focus=true

gpetters94:
I've dug around and it looks like the problem is arising at the lowering for `AtenFloatTensorOp`. I'm going to put a PR up with a fix for the GeFloat tests while I dig deeper into the issue.

PR #811, opened 2022-04-28
Revert to green

silvasean:
Revert some patches to get the build green again.

Fixes https://github.com/llvm/torch-mlir/issues/807

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -6569,31 +6569,6 @@ def Torch_AtenGtFloatOp : Torch_Op<"aten.gt.float", [
   let hasFolder = 1;
 }
 
-def Torch_AtenGeFloatOp : Torch_Op<"aten.ge.float", [
-    AllowsTypeRefinement,
-    HasValueSemantics,
-    ReadOnly
-  ]> {
-  let summary = "Generated op for `aten::ge.float : (float, float) -> (bool)`";
-  let arguments = (ins
-    Torch_FloatType:$a,
-    Torch_FloatType:$b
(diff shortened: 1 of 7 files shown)


silvasean:
@vivekkhandelwal1 , I appreciate the factoring into smaller commits, but in this case, because they were each very dependent on the previous ones, I had to revert multiple to get the build green again.

pashu123:
Hi, @silvasean I tried running this https://github.com/llvm/torch-mlir/blob/main/build_tools/torchscript_e2e_heavydep_tests/generate_serialized_tests.sh script. Shouldn't we make sure that this script runs to generate serialized scripts?

silvasean:
> Hi, @silvasean I tried running this https://github.com/llvm/torch-mlir/blob/main/build_tools/torchscript_e2e_heavydep_tests/generate_serialized_tests.sh script. Shouldn't we make sure that this script runs to at least generate serialized scripts?

Yes, that would be great. Can you add that?

vivekkhandelwal1:
> Hi, @silvasean I tried running this https://github.com/llvm/torch-mlir/blob/main/build_tools/torchscript_e2e_heavydep_tests/generate_serialized_tests.sh script. Shouldn't we make sure that this script runs to at least generate serialized scripts?

It's failing for two tests: hf_big_bird and timm_resnest. I will remove those tests till they are fixed.

vivekkhandelwal1:
@silvasean, I have to again add these commits once the issue with them is fixed, right?

vivekkhandelwal1:
Hi @silvasean, I have found a fix to this issue. The error was with the test case. Should I create a new PR with all commits squashed or keep them separate?

silvasean:
It's up to you. I would personally squash them due to the mechanical nature, and the fact that they don't apply independently.

PR #815, opened 2022-04-29
Fix LTC Decoupling

antoniojkim:
We previously had to merge in some temporary "hacks" to decouple LTC from the TS backend. Now that the proper fix has been merged into PyTorch master, we need to fix up this branch to use those changes.

Also, added PyTorch submodule for codegen purposes. Previously we were assuming that a pytorch clone was located in a specific place relative to `torch_mlir`. However, that is not an assumption that we could generally make. Adding PyTorch as a submodule fixes this.

CC: @ke1337 @wconstab


python/torch_mlir/csrc/base_lazy_backend/mlir_node.h
```diff
@@ -27,10 +27,63 @@ namespace lazy {
...
+
+  hash_t hash() const override;
+
+  hash_t shapeHash() const override;
 
   virtual TorchMlirOpVector
   Lower(TorchMlirFunction function, TorchMlirLoweringContext* loctx) const;
+
+private:
+  // The hash of the dag WITH size info. Used for shape caching
```
henrytwo, line 53: nit: missing period at end of line

henrytwo (approved):

antoniojkim:
@silvasean Gentle reminder to please review this when you can

python/torch_mlir/csrc/base_lazy_backend/dynamic_ir.cpp
```diff
@@ -0,0 +1,74 @@
...
+// This file is adapted from pytorch/pytorch
+// https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/ts_backend/dynamic_ir.cpp
+//===----------------------------------------------------------------------===//
+
+#include <torch/csrc/lazy/ts_backend/dynamic_ir.h>
+
+namespace torch {
+namespace lazy {
+
+DimensionNode::DimensionNode(OpKind op, OpList operands, hash_t hash_seed)
```
silvasean, line 18: Is there any doc on the strategy here for dynamic shapes?

silvasean (approved):

python/torch_mlir/csrc/base_lazy_backend/dynamic_ir.cpp
antoniojkim: There is, but its very much a work in progress I believe. This class is just a placeholder introduced for now. We'll need to make further changes to this once they finalize the design and implemention in PyTorch master.

PR #816, opened 2022-04-29
[MLIR][TORCH] Add heavydep tests for Bart, BigBird, and GPT2 model

vivekkhandelwal1:
This commit adds e2e heavydep tests for Bart, BigBird, and GPT2 model
via AOT.

### DIFF
--- a/build_tools/torchscript_e2e_heavydep_tests/functorch_inference.py
+++ b/build_tools/torchscript_e2e_heavydep_tests/functorch_inference.py
@@ -1,0 +1,161 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+
+from transformers import AutoModelForMaskedLM, AutoModelForSeq2SeqLM, AutoModelForCausalLM, BartConfig, BigBirdConfig, GPT2Config, DistilBertConfig
+import transformers
+import torch
+import torch.utils._pytree as pytree
+
+from torch_mlir_e2e_test.torchscript.framework import TestUtils
+from torch_mlir_e2e_test.torchscript.registry import register_test_case
+from torch_mlir_e2e_test.torchscript.annotations import annotate_args, export
+from functorch_utils import AOTModule, get_input_annotations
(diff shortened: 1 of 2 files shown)


build_tools/torchscript_e2e_heavydep_tests/functorch_inference.py
```diff
@@ -0,0 +1,161 @@
...
+device = "cpu"
+dtype = torch.float
+
+model = model_type.from_config(config).to(device, dtype=dtype)
+input_ids = torch.randint(0, config.vocab_size, input_size).to(device)
+decoder_ids = torch.randint(0, config.vocab_size, input_size).to(device)
+train_inputs = {"input_ids": input_ids, "labels": decoder_ids}
+
+
+def inference_fn(model, input, labels):
```
pashu123, line 65: Pass the inference function as `with torch.no grad():` Also remove `.loss.sum().backward()` since we are only intereseted in inference graph.

build_tools/torchscript_e2e_heavydep_tests/functorch_inference.py
```diff
@@ -0,0 +1,161 @@
...
```
pashu123, line 141: We can have the same inference function for all these models.

pashu123 (changes requested):

build_tools/torchscript_e2e_heavydep_tests/functorch_inference.py
pashu123:
```
with torch.no_grad(): 
   return model(**inputs)
```

build_tools/torchscript_e2e_heavydep_tests/functorch_inference.py
```diff
@@ -0,0 +1,161 @@
...
+@register_test_case(module_factory=lambda: ts_module_forw)
+def BartSequenceClassification_basic(module, tu: TestUtils):
+    module.forward(*forw_inputs)
+
+
+# ==============================================================================
+# BigBird Model
+
+config = BigBirdConfig(attention_type="original_full")
+model_type = AutoModelForMaskedLM
```
silvasean, line 92: Can you make this a function instead of copying the same code every time?

silvasean (changes requested):

vivekkhandelwal1:
I'm closing this since this is now a stale PR.

PR #820, opened 2022-05-02
[TORCH][MLIR] Add lowering of split like operations.

gprateek93:
Added lowerings for all split like operations. This include
`aten.split`, `aten.split.Sizes`, `aten.split_with_sizes`,
`aten.split.Tensor`  and `aten.chunk` ops. This commit also adds a pass
`DecomposeComplexOpsEarly` which decomposes complex ops like
split ops before shape inference.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/Transforms/Passes.h
+++ b/include/torch-mlir/Dialect/Torch/Transforms/Passes.h
@@ -32,11 +32,15 @@ struct TorchLoweringPipelineOptions
   // If this option is false, only do the bare minimum for correctness.
   Option<bool> optimize{*this, "optimize", llvm::cl::desc("Do optimizations."),
                         llvm::cl::init(true)};
-  
+
+  Option<bool> decomposeEarly{*this, "decompose-complex-ops-early",
+                              llvm::cl::desc("Decompose complex operations."),
+                              llvm::cl::init(true)};
   // If this option is false, decompose complex operations.
   // If this option is true, skip decomposition of complex operations.
-  Option<bool> decompose{*this, "decompose-complex-ops", llvm::cl::desc("Decompose complex operations."),
-                        llvm::cl::init(true)};                      
+  Option<bool> decompose{*this, "decompose-complex-ops",
+                         llvm::cl::desc("Decompose complex operations."),
+                         llvm::cl::init(true)};
 };
 
 /// Creates a pipeline that lowers the object graph IR that is produced by
@@ -67,6 +71,9 @@ std::unique_ptr<OperationPass<func::FuncOp>> createMaximizeValueSemanticsPass();
 std::unique_ptr<OperationPass<ModuleOp>> createRefinePublicReturnPass();
 
 std::unique_ptr<OperationPass<func::FuncOp>> createDecomposeComplexOpsPass();
+
+std::unique_ptr<OperationPass<func::FuncOp>>
+createDecomposeComplexOpsEarlyPass();
 
 std::unique_ptr<OperationPass<ModuleOp>> createPreprocessShapeLibraryPass();
 
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -16,7 +16,9 @@
 #include "torch-mlir/Dialect/Torch/IR/TorchTypes.h"
 #include "torch-mlir/Dialect/Torch/Transforms/Passes.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+#include "llvm/ADT/SmallVector.h"
 #include "llvm/ADT/StringExtras.h"
+#include <cstdint>
 
 using namespace mlir;
 using namespace mlir::torch;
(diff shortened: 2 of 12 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1686,6 +1688,82 @@ class DecomposeAtenPadOp : public OpRewritePattern<AtenPadOp> {
...
+    for (unsigned i = 0; i < tensorType.getSizes().size(); i++)
+      sizes.push_back(-1);
+
+    auto listType = op.getType().template cast<Torch::ListType>();
+    Location loc = op.getLoc();
+    SmallVector<int64_t> endIdx;
+    endIdx.push_back(offsets[0]);
+    for (unsigned i = 1; i < offsets.size(); i++) {
+      endIdx.push_back(offsets[i] + endIdx[i - 1]);
+    }
```
ramiro050, line 1719: nit: you can use `std::partial_sum` (https://en.cppreference.com/w/cpp/algorithm/partial_sum)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1686,6 +1688,82 @@ class DecomposeAtenPadOp : public OpRewritePattern<AtenPadOp> {
...
+    int64_t start = 0;
+    SmallVector<Value> listOfTensors;
+    Value stepVal = rewriter.create<Torch::ConstantIntOp>(
+        loc, rewriter.getI64IntegerAttr(1));
+    for (unsigned i = 0; i < offsets.size(); i++) {
```
ramiro050, line 1724: nit: it seems that `i` is only used to index `endIdx`, so I would do something like 

```cpp
for (int64_t end : endIdx) { ... }
```

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1686,6 +1688,82 @@ class DecomposeAtenPadOp : public OpRewritePattern<AtenPadOp> {
...
+    result.dump();
+    rewriter.replaceOp(op, result);
+    llvm::errs() << "Done till here\n";
+    return success();
+  }
+};
+} // namespace
+
+namespace {
+class DecomposeAtenSplitSizesOp : public OpRewritePattern<AtenSplitSizesOp> {
```
ramiro050, line 1755: Does `patterns.add<DecomposeAtenSplitLikeOp<AtenSplitSizesOp>>(context);` not work?

ramiro050 (commented):
Good work! A few comments

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1409,6 +1411,90 @@ void PrimTupleIndexOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
   });
 }
 
+//===----------------------------------------------------------------------===//
+// AtenSplitSizesOp
+//===----------------------------------------------------------------------===//
+
+void AtenSplitSizesOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                                   MLIRContext *context) {
```
silvasean, line 1419: This is not a canonicalization, but rather a lowering. This needs to go into a new pass called DecomposeOpsEarly.cpp which runs before shape refinement and ReduceOpVariants.

silvasean (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
gprateek93: Added the pass.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -916,6 +916,22 @@ def aten〇index〇Tensor(self: List[int], indices: List[Optional[List[int]]]) -
 def aten〇cat(tensors: List[List[int]], dim: int = 0) -> List[int]:
     return upstream_shape_helpers.cat(tensors, dim)
 
+def aten〇split(self: List[int], split_sizes: List[int], dim: int = 0) -> List[List[int]]:
```
silvasean, line 919: These should be contributed upstream. See https://github.com/llvm/torch-mlir/pull/908

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
```diff
@@ -32,11 +32,15 @@ struct TorchLoweringPipelineOptions
   // If this option is false, only do the bare minimum for correctness.
   Option<bool> optimize{*this, "optimize", llvm::cl::desc("Do optimizations."),
                         llvm::cl::init(true)};
-  
+
+  Option<bool> decomposeEarly{*this, "decompose-complex-ops-early",
```
silvasean, line 36: Keep these both under the "decompose" option. No need for two options.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
gprateek93: Added https://github.com/pytorch/pytorch/pull/79194. Please review. Thanks!

powderluv:
@gprateek93 Should we close this in favor of https://github.com/llvm/torch-mlir/pull/1022 ?

gprateek93:
> @gprateek93 Should we close this in favor of #1022 ?

Okay!

PR #821, opened 2022-05-02
[MLIR][TORCH] Modify aten::dropout op description

vivekkhandelwal1:
This commit adds the op `ValsemVariantAtenDropoutOp` that represents
`AtenDropout_Op` without the underscore. This is needed to
make sure that the `ReduceOpVariants` pass turns the in-place op
into an op that takes value tensors as inputs, otherwise the
semantics correctly.

This commit also adds the lowering of `ValsemVariantAtenDropoutOp` op.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -5547,6 +5547,30 @@ def Torch_AtenDropoutOp : Torch_Op<"aten.dropout", [
   }];
 }
 
+def Torch_AtenDropout_Op : Torch_Op<"aten.dropout_", [
+    IsTrailingUnderscoreInplaceVariant,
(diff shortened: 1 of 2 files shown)


ramiro050:
Do you have an example MLIR function showing the `maximize-value-semantics` failure? I can take a look to see what is going on. In the case of this op, we should not need to add a value-semantics variant because the version without undersore, `AtenDropoutOp`, already exists in pytorch

vivekkhandelwal1:
> I think this line here
> 
> https://github.com/llvm/torch-mlir/blob/52c9987115b39c2f6673315a21f6cdff0d1f51ca/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py#L439
> 
> should have used the `emit_with_mutating_variants` helper.

That worked. Thanks.

ramiro050 (approved):
LGTM!

PR #822, opened 2022-05-02
[WIP][TORCH] Add canonicalization of `aten.transpose.int` op

Shukla-Gaurav:
This commit adds a canonicalization of `aten.tranpose.int` op when its
operand is `aten.permute` op.
`aten.permute` + `aten.transpose.int` -> `aten.permute`


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -949,6 +949,37 @@ OpFoldResult AtenFloatScalarOp::fold(ArrayRef<Attribute> operands) {
...
+
+    int64_t dim0, dim1;
+    if (!matchPattern(op.dim0(), m_TorchConstantInt(&dim0)) ||
+        !matchPattern(op.dim1(), m_TorchConstantInt(&dim1)))
+      return failure();
+
+    SmallVector<Value> dimsList;
+    if (!getListConstructElements(permuteOp.dims(), dimsList))
+      return failure();
+    Value temp = dimsList[dim0];
```
cathyzhyi, line 971: nit: use `std::swap`

cathyzhyi (changes requested):
Can you add a unit test for this?

cathyzhyi:
Also I think this is only an optimization when the `transpose` is the only user of the original `permute`. Otherwise, we would be replacing one `transpose` with a `listcontruct` and an additional `permute`.

Shukla-Gaurav:
> Also I think this is only an optimization when the `transpose` is the only user of the original `permute`. Otherwise, we would be replacing one `transpose` with a `listcontruct` and an additional `permute`. Is this needed for performance optimization?

Yes, this is needed for the performance optimization of BERT. I will check for the users of `permute` op. That's a good point, Thanks!

PR #827, opened 2022-05-03
add no decompose option

denolf:
add option that allows skipping decomposition of complex operation

### DIFF
--- a/include/torch-mlir/Dialect/Torch/Transforms/Passes.h
+++ b/include/torch-mlir/Dialect/Torch/Transforms/Passes.h
@@ -32,6 +32,11 @@ struct TorchLoweringPipelineOptions
   // If this option is false, only do the bare minimum for correctness.
   Option<bool> optimize{*this, "optimize", llvm::cl::desc("Do optimizations."),
                         llvm::cl::init(true)};
+  
+  // If this option is false, decompose complex operations.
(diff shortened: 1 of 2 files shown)


include/torch-mlir/Dialect/Torch/Transforms/Passes.h
```diff
@@ -32,6 +32,11 @@ struct TorchLoweringPipelineOptions
   // If this option is false, only do the bare minimum for correctness.
   Option<bool> optimize{*this, "optimize", llvm::cl::desc("Do optimizations."),
                         llvm::cl::init(true)};
+  
+  // If this option is false, decompose complex operations.
+  // If this option is true, skip decomposition of complex operations.
+  Option<bool> nodecompose{*this, "nodecompose", llvm::cl::desc("No decomposition."),
```
silvasean, line 38: nit: can you call this "decompose-complex-ops" with a default of "true"?

silvasean (changes requested):

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
denolf: sure, done

silvasean (approved):

PR #830, opened 2022-05-04
Use JIT GraphExecutor for execution in example backend

henrytwo:
Previously, we returned dummy tensors in the example backend during its execution phase; however, this resulted in some unexpected bugs since the tensors contained incorrect data. 

To solve this, we will be using `GraphExecutor` with the `jit::Graph` to execute on the CPU, just like how it's done on the TorchScript backend. Of course, this is only for demonstration purposes, as each vendor will be expected to implement a custom execution flow for their device.

cc: @antoniojkim @ke1337


silvasean (approved):
Nice!

PR #831, opened 2022-05-04
Enable BERT support for torch-mlir via LTC

henrytwo:
This PR adds support for lowering a BERT model to MLIR using the Torch MLIR with LTC. As it currently stands, we can extract the forward graph for a [bert-base-cased](https://huggingface.co/bert-base-cased) model.

Many shape inference functions have been implemented, which may eventually be upstreamed to PyTorch.

Current printout using the example backend: https://gist.github.com/henrytwo/ec84705f24ec489e70319deebc97861b

cc: @ke1337 @antoniojkim

### DIFF
--- a/examples/ltc_backend_bert.py
+++ b/examples/ltc_backend_bert.py
@@ -1,0 +1,135 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+"""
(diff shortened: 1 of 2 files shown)


powderluv:
@pashu123 something to try once the backwards graph is also available

antoniojkim:
@silvasean Can you please review? There is another PR waiting on this one

silvasean (approved):
Sorry, this slipped through my review queue (feel free to ping me after 1 business day).

> Many shape inference functions have been implemented, which may eventually be upstreamed to PyTorch.

Why "may"? Is there a reason we would keep them downstream?

henrytwo:
> Why "may"? Is there a reason we would keep them downstream?

Oh yeah I meant we *will* upstream it eventually

PR #833, opened 2022-05-05
Add static shape for scalar tensors

henrytwo:
Previously scalar tensors were denoted with a type of `*` instead of `[]` in MLIR due to bug where an empty shape vector was interpreted as being an unranked tensor. The solution was to "key" the unranked case using a dimension number of -1.

Now that scalar tensor shapes are encoded explicitly, we found a previously unknown bug with JIT tuples. Since we inject shape information into the JIT graph ourselves, any previously constructed tuples will have outdated types. This related PR addresses those issues: https://github.com/pytorch/pytorch/pull/76919 Without it, the MLIR generated is illegal due to the aforementioned conflicting types.

cc: @ke1337 @antoniojkim

### DIFF
--- a/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
+++ b/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
@@ -17,7 +17,6 @@
 #include <torch/csrc/lazy/core/lazy_graph_executor.h>
 
 #include "../../dialects/torch/importer/jit_ir/csrc/function_importer.h"
-#include "../../dialects/torch/importer/jit_ir/csrc/import_options.h"
 #include "../utils/debug.h"
 #include "../utils/exception.h"
 #include "backend_impl.h"
@@ -121,14 +120,11 @@ ComputationPtr TorchMlirLoweringContext::Build() {
   }
 
   // Generate MLIR.
-  torch_mlir::ImportOptions import_options{
-      /*assumeTensorsHaveValueSemantics=*/true,
-      /*assumeZeroRankTensorsAreScalar=*/true};
   MlirOperation func_op = torch_mlir::importJitFunctionAsFuncOp(
       /*context=*/mlir_context_,
       /*function=*/generate_jit_fn().get(),
       /*getArgAttribute=*/[](int) -> MlirAttribute { return {nullptr}; },
-      /*importOptions=*/import_options);
+      /*importOptions=*/{/*assumeTensorsHaveValueSemantics=*/true});
 
   return std::make_shared<TorchMlirComputation>(
       func_op, mlir_context_, graph_, input_output_aliases_);
--- a/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
+++ b/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
@@ -13,7 +13,7 @@
 #include <iostream>
 
 #include <torch/csrc/jit/api/compilation_unit.h>
-#include <torch/csrc/jit/passes/refine_types.h>
+#include <torch/csrc/jit/passes/refine_tuple_types.h>
 #include <torch/csrc/lazy/core/lazy_graph_executor.h>
 
 #include "../../dialects/torch/importer/jit_ir/csrc/function_importer.h"
@@ -110,9 +110,8 @@ ComputationPtr TorchMlirLoweringContext::Build() {
   PRINT_FUNCTION();
 
   // Since we mutated the types of some nodes to insert shape information, we
-  // must perform this pass to ensure other data structures (such as tuples)
-  // have up to date output types.
-  torch::jit::RefineTypes(graph_);
+  // must perform this pass to ensure tuples have up to date output types.
+  torch::jit::RefineTupleTypes(graph_);
 
   // Insert return values into graph.
   for (torch::jit::Value* output : root_tuple_) {
--- a/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
+++ b/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
@@ -13,6 +13,7 @@
 #include <iostream>
 
 #include <torch/csrc/jit/api/compilation_unit.h>
+#include <torch/csrc/jit/passes/refine_types.h>
 #include <torch/csrc/lazy/core/lazy_graph_executor.h>
 
 #include "../../dialects/torch/importer/jit_ir/csrc/function_importer.h"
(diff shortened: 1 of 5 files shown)


python/torch_mlir/dialects/torch/importer/jit_ir/csrc/import_options.h
```diff
@@ -23,6 +23,10 @@ struct ImportOptions {
   // a requirement to use a value-semantic tensor type (!torch.vtensor) in
   // signatures.
   bool assumeTensorsHaveValueSemantics = false;
+  // If this is set to true, then all tensors with zero rank are assumed to be
```
silvasean, line 26: Tensors with 0 rank should already be imported as [] -- if not, then that's just a bug.

python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
```diff
@@ -108,17 +110,25 @@ void TorchMlirLoweringContext::AddParameter(
 ComputationPtr TorchMlirLoweringContext::Build() {
   PRINT_FUNCTION();
 
+  // Since we mutated the types of some nodes to insert shape information, we
```
silvasean, line 113: You might want to check if the static information adjustment I added in https://github.com/llvm/torch-mlir/pull/824 fixes this issue for you. Not opposed to refining the types, but just to know whether it is truly necessary.

silvasean (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/import_options.h
henrytwo: According to [this](
https://github.com/llvm/torch-mlir/blob/torch_mlir_ltc_backend/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp#L143-L147), if `sizes.rank()` is zero, then `optionalSizes=nullptr`. So then over [here](https://github.com/llvm/torch-mlir/blob/main/lib/CAPI/TorchTypes.cpp#L222-L223),  `optionalSizesArrayRef` remains `None`, which I found defaults to `*` instead of `[]`.

If 0 rank is _supposed_ to be `[]`, then the change would just be to default that to an empty array instead?

silvasean: but `sizes.rank()` returns an Optional
```
  c10::optional<size_t> rank() const {
    if(!dims_) {
      return c10::nullopt;
    }
    return dims_->size();
  }
```
The case of "known rank of 0" should not take the branch commented as "unranked".

It might be clearer to rewrite that code as

```
auto &symbolicShape = tensorType->symbolic_sizes();
auto &maybeSizes = symbolicShape.sizes();
if (!maybeSizes.has_value()) { unranked }
auto &sizes = maybeSizes.value();
...
```

henrytwo: Yeah you're right it shouldn't go down that path, but if `sizes.rank()` is zero, then `dims.data()` will be falsy [here](https://github.com/llvm/torch-mlir/blob/torch_mlir_ltc_backend/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp#L158), which will result in a shape of `*` instead of `[]`

silvasean: Yeah, I think I have run into this before -- std::vector doesn't guarantee that dims.data() is a valid pointer. To make this really robust I think we want to key the unranked case on "numSizes == -1" or something more reliable than the pointer being null.

python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
henrytwo: Hm I'm looking at some of the tests [here](https://github.com/silvasean/torch-mlir/blob/b0ad9b21e716f6291e4fb15d8795d14b982b9075/test/python/importer/jit_ir/node_import/tuple.py#L67-L68) and I'm not entirely sure this is what we want?
```
# CHECK-LABEL:   func @__torch__.tuple_construct_arg_needs_refinement(
# CHECK-SAME:                                                         %[[T0:.*]]: !torch.tensor,
# CHECK-SAME:                                                         %[[T1:.*]]: !torch.tensor) -> !torch.tuple<tensor, tensor> {
# CHECK:           %[[T0_REFINED:.*]] = torch.tensor_static_info_cast %[[T1]] : !torch.tensor to !torch.tensor<[4],f32>
# CHECK:           %[[T1_REFINED:.*]] = torch.tensor_static_info_cast %[[T0]] : !torch.tensor to !torch.tensor<[3],f32>
# CHECK:           %[[TUPLE:.*]] = torch.prim.TupleConstruct %[[T0_REFINED]], %[[T1_REFINED]] : !torch.tensor<[4],f32>, !torch.tensor<[3],f32> -> !torch.tuple<tensor<[4],f32>, tensor<[3],f32>>
# CHECK:           %[[DEREFINED:.*]] = torch.derefine %[[TUPLE]] : !torch.tuple<tensor<[4],f32>, tensor<[3],f32>> to !torch.tuple<tensor, tensor>
# CHECK:           return %[[DEREFINED]] : !torch.tuple<tensor, tensor>
```

For our use case, we want to directly change the output type of `TupleConstruct` to reflect the updated one, because otherwise the type is just completely wrong. 

e.g. Before:
```
graph(%a : Float(123), %b : Float(4, 5, 6)):
    c : (Tensor, Tensor) = prim::TupleConstruct(%a, %b)
    return (%c)
```
After:
```
graph(%a : Float(123), %b : Float(4, 5, 6)):
    c : (Float(123), Float(4, 5, 6)) = prim::TupleConstruct(%a, %b)
    return (%c)
```
It looks like your solution does a cast operation before and after, which I feel is unnecessary to solve this issue.

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/import_options.h
henrytwo: Got it, I made that change.

henrytwo:
@silvasean Are there any more changes you need me to make to this PR?

python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
```diff
@@ -108,6 +109,10 @@ void TorchMlirLoweringContext::AddParameter(
...
+  // must perform this pass to ensure tuples have up to date output types.
+  torch::jit::RefineTupleTypes(graph_);
```
ke1337, line 114: Seems this would require https://github.com/pytorch/pytorch/pull/76919 to be landed first?

silvasean:
> @silvasean Are there any more changes you need me to make to this PR?

I think we are good to go!

henrytwo:
Now that https://github.com/pytorch/pytorch/pull/76919 has been merged, I'm going to go ahead and merge this in now.

PR #834, opened 2022-05-06
[TORCH] Add lowering of aten.detach op.

pashu123:
`aten.detach` op is folded and returns the first operand since it's an
identity function(kind of identity just remove the has_grad attribute).


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -615,6 +615,17 @@ OpFoldResult AtenSqueezeDimOp::fold(ArrayRef<Attribute> operands) {
   return nullptr;
 }
 
+//===----------------------------------------------------------------------===//
```
cathyzhyi, line 618: Can you add a unit test in test/Dialect/Torch/canonicalize.mlir?

cathyzhyi (changes requested):

PR #835, opened 2022-05-06
[LINALG] Add support for preserve memory format in aten_empty_like op.

pashu123:
The preserve memory specifies that `If any of the input tensors is in channels_last format,
operator output should be in channels_last format` and hence can be
added as is in aten_empty_like op.


cathyzhyi (changes requested):

cathyzhyi:
Hey @pashu123 canonicalizer is an optimization and doesn't guarantee to run. The transformation is needed for correctness so we still need a conversion pattern for this.

PR #838, opened 2022-05-06
Fix type promotion code for scalar only operations

cathyzhyi:
Fix the type promotion code for scalar only operation to return
TorchType which is the type tracked in ValueKnowledge.scalarType.

- Fix `getPromotedResultScalarType` to return Torch type.
- Add `getBuiltInTypeForTorchScalar` helper to convert scalar type
to builtin type before passing to the next level type promotion
helper `updateResultTypeState`.
- Add `setScalarType` helper to make setting ValueKnowledge.scalarType
  easier.

### DIFF
--- a/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
+++ b/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
@@ -108,8 +108,9 @@ public:
         result.returnOp = returnOp;
       } else {
         return rewriter.notifyMatchFailure(
-            copyToNonValueTensor,
-            "unsupported op encountered during abstract analysis");
+            copyToNonValueTensor, "unsupported op `" +
+                                      user->getName().getStringRef() +
+                                      "` encountered during abstract analysis");
       }
     }
     return result;
--- a/lib/Dialect/Torch/Transforms/RefineTypes.cpp
+++ b/lib/Dialect/Torch/Transforms/RefineTypes.cpp
@@ -114,6 +114,10 @@ static torch_upstream::TypeKind getTypeKind(Type type) {
 /// on the `dtype` from tensors and can't be used on other types like scalar
 /// types.
 static Optional<Type> meetElementTypes(Type lhs, Type rhs) {
+  auto isNullOrBuiltIn = [](Type type) { return !type || isBuiltInType(type); };
+  assert(isNullOrBuiltIn(lhs) && "`lhs` must be a builtin type");
+  assert(isNullOrBuiltIn(rhs) && "`rhs` must be a builtin type");
+
   if (!lhs)
     return rhs;
   if (!rhs)
@@ -166,6 +170,14 @@ struct ValueKnowledge {
                  torch_upstream::TypeKind kind)
       : dtype(dtype), scalarType(scalarType), kind(kind),
         optional(optionalKnowledge) {}
+
+  void setScalarType(Type type) {
+    bool isValidScalarType = type.isa<NumberType, IntType, Torch::FloatType>();
+    assert(isValidScalarType &&
+           "scalarType can only be one of NumberType, IntType and FloatType");
+    scalarType = type;
+    kind = getTypeKind(type);
+  }
 
   // Get the static knowledge intrinsic to `type`.
   static ValueKnowledge getKnowledgeFromType(Type type) {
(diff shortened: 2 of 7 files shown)


lib/Dialect/Torch/Utils/Utils.cpp
```diff
@@ -78,6 +78,18 @@ Type Torch::getTypeForScalarType(
...
 
+Type Torch::getTorchTypeForScalarType(MLIRContext *context,
+                                      torch_upstream::ScalarType dtypeInt) {
+  switch (dtypeInt) {
+  case torch_upstream::ScalarType::Double:
+    return Torch::FloatType::get(context);
+  case torch_upstream::ScalarType::Long:
+    return Torch::IntType::get(context);
+  default:
+    assert(false && "Unsupported scalar type to Torch type conversion");
```
ramiro050, line 89: nit: I think it might be more appropriate to use `llvm::report_fatal_error`, since there are still missing `case`s. The way I understand it, `assert`s are for testing invariants of the function, but in this case the `default` branch can be reached without breaking any invariants, such as by passing `ScalarType::Bool` as the `dtypeInt` argument

lib/Conversion/Utils/Utils.cpp
```diff
@@ -275,6 +275,8 @@ Value convertScalarToDtype(OpBuilder &b, Location loc, Value scalar,
   llvm_unreachable("convertScalarToDtype should handle all the types");
 }
 
+bool isBuiltInType(Type type) { return isa<BuiltinDialect>(type.getDialect()); }
```
ramiro050, line 278: Should this be in `lib/Dialect/Torch/Utils/Utils.cpp` since its prototype is in `include/torch-mlir/Dialect/Torch/Utils/Utils.h`?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -420,7 +432,7 @@ class TypeAnalyzer : public ForwardDataFlowAnalysis<ValueKnowledge> {
 // This is the type rule used for deciding dtype for:
 // 1. A new tensor created from given data.
 // 2. The scalar type for type promotion when a scalar is an operand of a tensor
-// and scalar binary operation.
+// operation (such as AtenMulScalar, AtenAddScalarOp etc)
```
ramiro050, line 435: "AtenMulScalar" -> "AtenMulScalar**Op**"

lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
```diff
@@ -108,8 +108,9 @@ class AbstractlyInterpretCopyToNonValueTensorOpUsersWithinABlock
         result.returnOp = returnOp;
       } else {
         return rewriter.notifyMatchFailure(
-            copyToNonValueTensor,
-            "unsupported op encountered during abstract analysis");
+            copyToNonValueTensor, "unsupported op `" +
+                                      user->getName().getStringRef() +
+                                      "` encountered during abstract analysis ");
```
ramiro050, line 113: nit: extra space at the end of error message

ramiro050 (approved):
LGTM! I just have a few nits

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -448,6 +448,9 @@ def emit_with_mutating_variants(key, **kwargs):
     emit("aten::full : (int[], Scalar, int?, int?, Device?, bool?) -> (Tensor)")
     emit("aten::full_like : (Tensor, Scalar, int?, int?, Device?, bool?, int?) -> (Tensor)")
 
+    # Scalar ops
+    emit("aten::add : (Scalar, Scalar) -> (Scalar)")
```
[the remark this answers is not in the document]
cathyzhyi: right, thanks!!

lib/Conversion/Utils/Utils.cpp
cathyzhyi: yea!

lib/Dialect/Torch/Utils/Utils.cpp
cathyzhyi: Good point. Changed as suggested.

PR #844, opened 2022-05-10
[TORCH][MLIR] Add lowering of `aten.slice_scatter` op and `aten.select_scatter` op

gprateek93:
This commit adds:
1.  Lowering of `aten.slice_scatter` op into `tensor.insert_slice`
op.
2. Decomposes the `aten.select_scatter` op into `aten.slice_scater`
op.


python/torch_mlir_e2e_test/test_suite/slice_like.py
```diff
@@ -232,3 +232,75 @@ def SelectIntModule_basic(module, tu: TestUtils):
...
+    ])
+    def forward(self, x, src):
+        return torch.ops.aten.slice_scatter(x, src, 1, 0, 1, 1)
+
+
+@register_test_case(module_factory=lambda: SliceScatterStaticModule())
+def SliceScatterStaticModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(6, 8), tu.rand(6, 1))
+
+class SelectScatterModule(torch.nn.Module):
```
vivekkhandelwal1, line 272: Could you please add some tests with dim other than zero?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1791,6 +1793,56 @@ class DecomposeAtenToDtypeLayoutOp
...
+    Value src = op.src();
+
+    Value one =
+        rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
+    Value startPlusOne =
+        rewriter.create<AtenAddIntOp>(loc, one.getType(), start, one);
+    BaseTensorType srcTensorType = src.getType().cast<BaseTensorType>();
+    SmallVector<int64_t> sizes;
+    int64_t dimInt = 0;
+    if (srcTensorType.hasSizes()) {
```
cathyzhyi, line 1817: The calculation of result type of the `aten.unsqeeze` can be replaced with `computeReductionType(...self)`

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -29,6 +31,43 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 
+static Value adjustStartOrEndForSlicing(ConversionPatternRewriter &rewriter,
```
ramiro050, line 34: nit: I think we can make the behavior of this function more clear by naming it `clampTorchOptionalInt`. I would also do the following renaming:

- `startOrEndTorchType` -> `torchOptionalInt` (to avoid it being confused with an actual type)
- `startOrEndBuiltin` -> `builtinInt`
- `ValueForNone` -> `defaultValue` (It can be confused with asking for the dialects representation of the `None` value.)
- `dimSize` -> `maxValue` (to make it clear that we are using that value as an upper bound in the clamping)

python/torch_mlir_e2e_test/test_suite/slice_like.py
```diff
@@ -232,3 +232,75 @@ def SelectIntModule_basic(module, tu: TestUtils):
...
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, x, src):
+        return torch.ops.aten.slice_scatter(x, src, 1, 0, 1, 1)
```
cathyzhyi, line 247: nit: add comments for arguments to improve readability.

cathyzhyi, line 247: can you also add cases where step is not 1?

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -29,6 +31,43 @@ using namespace mlir;
...
+                                        Location loc, Value startOrEndTorchType,
+                                        Value startOrEndBuiltin,
+                                        Value valueForNone, Value dimSize) {
+  if (startOrEndTorchType.getType().isa<Torch::NoneType>())
+    return valueForNone;
+  auto dimSizeAsInt = castIndexToInt64(rewriter, loc, dimSize);
+  Value startOrEndToPositive =
```
ramiro050, line 41: Since the behavior of this function is really just about clamping a value to a range, I don't think we need to have the `startOrEnd` parts in the name of the variables in the body of this function. The `startOrEnd` is an implementation detail of the lowerings for the slice ops, not an implementation detail of this function.

cathyzhyi (changes requested):

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -891,3 +891,21 @@ func @torch.aten.to.dtype_layout(%arg0: !torch.vtensor<[?,?],f32>) -> !torch.vte
...
+// CHECK-SAME:  (%[[SELF:.*]]: !torch.vtensor<[?,?],f32>, %[[SRC:.*]]: !torch.vtensor<[?],f32>) -> !torch.vtensor<[?,?],f32> {
+// CHECK-NEXT:    %[[START:.*]] = torch.constant.int 0
+// CHECK-NEXT:    %[[DIM:.*]] = torch.constant.int 1
+// CHECK-NEXT:    %[[STEP:.*]] = torch.constant.int 1
+// CHECK-NEXT:    %[[END:.*]] = torch.aten.add.int %[[START]], %[[STEP]]
+// CHECK-NEXT:    %[[UNSQUEEZE_SRC:.*]] = torch.aten.unsqueeze %[[SRC]], %[[DIM]]
+// CHECK-NEXT:    %[[SLICE_SCATTER:.*]] = torch.aten.slice_scatter %[[SELF]], %[[UNSQUEEZE_SRC]], %[[DIM]], %[[START]], %[[END]], %[[STEP]]
+// CHECK-NEXT:    return %[[SLICE_SCATTER]]
+// CHECK-NEXT:    }
+func @select_scatter_decompose(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.vtensor<[?],f32>) -> !torch.vtensor<[?,?],f32> {
```
ramiro050, line 906: nit: the convention in this file is to name the function after the op being decomposed, so `torch.aten.select_scatter` in this case

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ramiro050, line 1817: Make sure to return early to avoid indentation. In this case I would do

```cpp
if (!srcTensorType.hasSizes())
  return rewriter.notifyMatchFailure(op, "dim is not a valid dim");

// rest of code
```

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1791,6 +1793,56 @@ class DecomposeAtenToDtypeLayoutOp
...
+    Value self = op.self();
+    Value src = op.src();
+
+    Value one =
+        rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
+    Value startPlusOne =
+        rewriter.create<AtenAddIntOp>(loc, one.getType(), start, one);
+    BaseTensorType srcTensorType = src.getType().cast<BaseTensorType>();
+    SmallVector<int64_t> sizes;
+    int64_t dimInt = 0;
```
ramiro050, line 1816: nit: I would move this down to right before the `matchPattern`, since it is not needed before that

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1791,6 +1793,56 @@ class DecomposeAtenToDtypeLayoutOp
...
+    if (srcTensorType.hasSizes()) {
+      ArrayRef<int64_t> srcShape = srcTensorType.getSizes();
+      // `src` has a reduced rank. Hence add 1.
+      int64_t srcRank = srcShape.size() + 1;
+      if (matchPattern(dim, m_TorchConstantInt(&dimInt))) {
+        dimInt = toPositiveDim(dimInt, srcRank);
+        if (!isValidDim(dimInt, srcRank))
+          return rewriter.notifyMatchFailure(op, "dim is not a valid dim");
+
+        sizes.append(srcShape.begin(), srcShape.end());
```
ramiro050, line 1826: nit: You can probably just do `sizes = srcShape;`.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1791,6 +1793,56 @@ class DecomposeAtenToDtypeLayoutOp
...
+        sizes.insert(sizes.begin() + dimInt, 1);
```
ramiro050, line 1827: nit: `sizes[dimInt] = 1;`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1791,6 +1793,56 @@ class DecomposeAtenToDtypeLayoutOp
...
+
+      } else {
+        sizes.resize(srcShape.size() + 1, kUnknownSize);
```
ramiro050, line 1830: what is the reason for adding 1 here?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1791,6 +1793,56 @@ class DecomposeAtenToDtypeLayoutOp
...
+        rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
+    Value startPlusOne =
+        rewriter.create<AtenAddIntOp>(loc, one.getType(), start, one);
+    BaseTensorType srcTensorType = src.getType().cast<BaseTensorType>();
+    SmallVector<int64_t> sizes;
+    int64_t dimInt = 0;
+    if (srcTensorType.hasSizes()) {
+      ArrayRef<int64_t> srcShape = srcTensorType.getSizes();
+      // `src` has a reduced rank. Hence add 1.
+      int64_t srcRank = srcShape.size() + 1;
```
ramiro050, line 1820: This is really just for my own understanding, but why do we add 1 here? What's the expected rank for `src` and why is it originally reduced to one less than the expected rank?

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: Actually, I think naming the function something like `toPositiveValidDim` might be a better name for the function, since it is basically the function `toPositiveDim` plus a few extra calculations to make sure the dim value is within bounds.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gprateek93: Actually according to the `select_scatter` the selection dimension is squeezed out. But in the `slice_scatter` we need the selection/slicing dimension(The reason of adding unsqueezeOp).

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gprateek93: src `dim` dimension is squeezed up. To unsqueeze it 1 is added.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gprateek93: We have to insert here and not assign, as `sizes` does not have the `dimInt` index, or is matched to wrong index due to squeezing of `src`.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
gprateek93: I tried using this, but it is not getting the correct shape which is needed by this particular function.

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: The renaming of the arguments above was not addressed

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: This comment applied to the entire body of the function

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -29,6 +32,62 @@ using namespace mlir;
...
+      loc, startOrEndSgtDimSize, dimSizeAsInt, startOrEndAtLeastZero);
+
+  return castIntToIndex(rewriter, loc, startOrEndBoundedByDimSize);
+}
+
+static void prepareArgumentsForSlicingOp(
+    ConversionPatternRewriter &rewriter, Location loc,
+    SmallVector<Value> &resultShape, SmallVector<Value> &offsets,
+    SmallVector<Value> &strides, Value startTorchType, Value startBuiltinType,
+    Value endTorchType, Value endBuiltinType, int64_t step, Value dimSize,
```
ramiro050, line 64: nit: I find postfixing the `start` and `end` variables with `Type` confusing because it makes it seem like those variables are supposed to be `Type`s. I would flip the order and do something like `torchTypeStart, builtinTypeStart, ...`

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -748,39 +807,14 @@ class ConvertAtenSliceTensorOp : public OpConversionPattern<AtenSliceTensorOp> {
...
-          loc, arith::CmpIPredicate::sgt, startOrEndAtLeastZero, dimSizeAsInt);
-      Value startOrEndBoundedByDimSize = rewriter.create<arith::SelectOp>(
-          loc, startOrEndSgtDimSize, dimSizeAsInt, startOrEndAtLeastZero);
-
-      return castIntToIndex(rewriter, loc, startOrEndBoundedByDimSize);
-    };
+    Value startTorchType = op.start();
+    Value endTorchType = op.end();
+    Value startBuiltinType = adaptor.start();
+    Value endBuiltinType = adaptor.end();
```
ramiro050, line 813: Same comment as in `prepareArgumentsForSlicingOp`, I would place the type info first, and end with `start` or `end`

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1095,6 +1120,79 @@ class ConvertValsemVariantAtenCopyOp
...
+        return op->emitError("unimplemented: step is not constant");
+      step = 1;
+    }
+
+    SmallVector<Value> resultShape;
+    SmallVector<Value> offsets(inputType.getRank(), zero);
+    SmallVector<Value> strides(inputType.getRank(), one);
+    prepareArgumentsForSlicingOp(rewriter, loc, resultShape, offsets, strides,
+                                 startTorchType, startBuiltinType, endTorchType,
+                                 endBuiltinType, step, dimSize, dim, input);
```
ramiro050, line 1175: All this code is the same in the conversion patterns of `AtenSliceTensorOp` and `AtenSliceScatterOp`. If you turn the helper `prepareArgumentsForSlicingOp` into a template, you should be able to perform this entire calculation inside that helper function. This will also simplify the type signature quite a bit

ramiro050 (changes requested):
A few more comments

python/torch_mlir_e2e_test/test_suite/slice_like.py
ramiro050: @gprateek93 , i think what was meant here was to do something like `slice_scatter(x, src, dim=0, start=0, end=1, step=1)`, so that it is clear to the reader what each argument means from just looking at the `forward` function

lib/Conversion/TorchToLinalg/DataMovement.cpp
gprateek93: Done.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1011,6 +1043,58 @@ class ConvertValsemVariantAtenCopyOp
...
+    RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+    RankedTensorType resultType =
+        typeConverter->convertType(op->getResult(0).getType())
+            .cast<RankedTensorType>();
+    Value zero = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+    Value one = rewriter.create<arith::ConstantIndexOp>(loc, 1);
+
+    SmallVector<Value> resultShape;
+    SmallVector<Value> offsets(inputType.getRank(), zero);
+    SmallVector<Value> strides(inputType.getRank(), one);
```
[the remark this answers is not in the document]
gprateek93: Done. Thanks!

gprateek93:
Hi @cathyzhyi, can you please review this PR. I have addressed all the comments.

ramiro050 (approved):
I just have a few small comments and a few unresolved comments, but other than that everything looks good!

ramiro050:
Note: The links to the comments seems to only work if you right click and open in a different tab

lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
```diff
@@ -53,14 +53,14 @@ module {
     return %0 : !torch.list<int>
   }
   func.func @__torch__.torch.jit._shape_functions.adaptive_avg_pool2d(%arg0: !torch.list<int>, %arg1: !torch.list<int>) -> !torch.list<int> {
-    %true = torch.constant.bool true
```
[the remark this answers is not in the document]
gprateek93: I did rebase, But I am getting the same!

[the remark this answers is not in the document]
gprateek93: I think there is the change of order only. No functional change here!

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -29,6 +33,94 @@ using namespace mlir;
...
+      loc, rewriter.getZeroAttr(dimSizeAsInt.getType()));
+  Value predDimSltZero = rewriter.create<arith::CmpIOp>(
+      loc, arith::CmpIPredicate::slt, positiveDim, cst0);
+  Value startOrEndAtLeastZero =
+      rewriter.create<arith::SelectOp>(loc, predDimSltZero, cst0, positiveDim);
+  // startOrEnd > dimSizeAsInt ? dimSizeAsInt : startOrEnd
+  Value startOrEndSgtDimSize = rewriter.create<arith::CmpIOp>(
+      loc, arith::CmpIPredicate::sgt, startOrEndAtLeastZero, dimSizeAsInt);
+  Value startOrEndBoundedByDimSize = rewriter.create<arith::SelectOp>(
+      loc, startOrEndSgtDimSize, dimSizeAsInt, startOrEndAtLeastZero);
```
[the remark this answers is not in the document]
ramiro050: This was not addressed

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -29,6 +33,94 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 
+static Value toPositiveValidDim(ConversionPatternRewriter &rewriter,
+                                Location loc, Value torchType,
+                                Value builtinType, Value valueForNone,
+                                Value dimSize) {
```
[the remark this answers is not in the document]
ramiro050: This was not addressed

ramiro050:
@gprateek93 , there are two comments that were not addressed. Please address them in a new PR. In future PRs, please don't merge until all comments are addressed. 

- https://github.com/llvm/torch-mlir/pull/844#discussion_r870661551 
- https://github.com/llvm/torch-mlir/pull/844#discussion_r906286528

If the comments are unclear, please ask questions.

PR #845, opened 2022-05-11
[TORCH] Fold `aten.contiguous` op

Shukla-Gaurav:
This commit adds folding of `aten.contiguous` op. Since we do not take
memory format into account currently, the `aten.contiguous` op can be
folded away.

Signed-Off-by: Gaurav Shukla


cathyzhyi:
Canonicalizer is an optimization which is not guarantee to run. We would need the conversion pattern for correctness so it can't be deleted.
Why do we need the folder?

Shukla-Gaurav:
@cathyzhyi It seems right to keep the conversion pattern for the sake of completeness. But isn't it good to keep the folder also? It's like if something can be handled earlier in the pipeline, why to keep it long. What do you think? @silvasean @ramiro050

PR #846, opened 2022-05-11
Add -s flag to run e2e tests sequentially

cathyzhyi:
A user might want to avoid the extra layer of multiprocessing libary for
debugging purpose. In such cases, the -s flag can be used to force
sequential execution.


makslevental (approved):
LGTM!

PR #849, opened 2022-05-12
python: [nfc] remove duplicate dialect registration

ashay:
Ever since https://reviews.llvm.org/D88155, the code for the MLIR Python
bindings registers all dialects in
`PyMlirContext::createNewContextForInit()`, so the (secondary) dialect
registration in torch-mlir's `ModuleBuilder` constructor is duplicate.
Although this did not present any major problems until now, we run into
a duplicate operation registration error after updating to the most
recent MLIR build.


ashay:
Closing this PR since this change is included in PR #856.

PR #850, opened 2022-05-12
Use delocate to make portable wheels on OSX

powderluv:
Fix up wheels per python version on OSX


powderluv:
pushing to make the next release cut. will address any comments.

PR #851, opened 2022-05-12
Add shape info to `rand_like` + support for `dtype` flag

ramiro050:
The op `aten.rand_like` was missing a shape function, unit tests, and
the `dtype` argument was being ignored in its decomposition. This
commit fixes all three things.


cathyzhyi (approved):
LGTM!

PR #855, opened 2022-05-13
[MLIR][TORCH] Add decomposition of `aten.adaptive_avg_pool2d` op

vivekkhandelwal1:
This commit adds the decomposition of `aten.adaptive_avg_pool2d` op into
`aten.avg_pool2d` op. The current decomposition only supports cases where
input size is equal to the output size.

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -2218,9 +2218,8 @@ LogicalResult ConvertAtenOp<AtenFlattenUsingIntsOp>::matchAndRewrite(
     newShape.push_back(1);
 
   auto newType = RankedTensorType::get(newShape, selfType.getElementType());
-  auto reshapeOp =
-      rewriter.create<tosa::ReshapeOp>(op.getLoc(), newType, adaptor.self(),
-                                       rewriter.getI64ArrayAttr(newShape));
+  auto reshapeOp = rewriter.create<tosa::ReshapeOp>(
+      op.getLoc(), newType, adaptor.self(), rewriter.getI64ArrayAttr(newShape));
 
   rewriter.replaceOpWithNewOp<tensor::CastOp>(
       op, getTypeConverter()->convertType(op.getType()), reshapeOp);
@@ -2507,7 +2506,8 @@ static Value buildUnitNormalCdf(ConversionPatternRewriter &rewriter,
   auto mean = zero;
   Value xMinusMean = rewriter.create<tosa::SubOp>(loc, outType, x, mean);
   // rsqrt of 2
-  Value rsqrt2 = tosa::getConstTensor<float>(rewriter, op, 0.70710678, {}).getValue();
+  Value rsqrt2 =
+      tosa::getConstTensor<float>(rewriter, op, 0.70710678, {}).getValue();
   Value erfArg = rewriter.create<tosa::MulOp>(loc, outType, xMinusMean, rsqrt2,
                                               /*shift=*/0);
   Value erf = approximateErfOp(rewriter, op, erfArg);
@@ -2623,9 +2623,9 @@ public:
         op, "Unimplemented pooling input parsing function");
   }
 
-  int64_t getOutputDim(int64_t inputDim, int64_t kernelDim, int64_t stride,
-                       int64_t padBefore, int64_t padAfter,
-                       int64_t dilation) const {
+  static int64_t getOutputDim(int64_t inputDim, int64_t kernelDim,
+                              int64_t stride, int64_t padBefore,
+                              int64_t padAfter, int64_t dilation) {
     if (inputDim == ShapedType::kDynamicSize) {
       return ShapedType::kDynamicSize;
     } else {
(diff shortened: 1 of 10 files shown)


vivekkhandelwal1:
The tests which failed in CI, are all passing locally.

ramiro050:
I tested your patch on my machine, and only the `ResNet18StaticModule` and `ResNet18Module` fail. The error message from the test suite is not very useful. I ended up running the ResNet tests using this script:

https://github.com/ramiro050/torch-mlir/blob/73b2db60b5c624a454c9ceb12dd10780905fbb49/examples/torchscript_resnet18_test.py

and I cherry-picked @cathyzhyi LLVM patch for printing runtime errors, and after that I get the following error:

```
unimplemented: only support cases where input and output size are equal
fish: Job 3, 'python ../torch-mlir-scripts/to…' terminated by signal SIGABRT (Abort)
```

Which is an error message in the decomposition of `AtenAdaptiveAvgPool2dOp`

vivekkhandelwal1:
> I tested your patch on my machine, and `ResNet18StaticModule` and `ResNet18Module` are the only tests I get a failure on. The error message from the test suite is not very useful. I ended up running the ResNet tests using this script:
> 
> https://github.com/ramiro050/torch-mlir/blob/73b2db60b5c624a454c9ceb12dd10780905fbb49/examples/torchscript_resnet18_test.py
> 
> and I cherry-picked @cathyzhyi LLVM patch for printing runtime errors ([cathyzhyi/llvm-project@2760f4d](https://github.com/cathyzhyi/llvm-project/commit/2760f4d074cc4ac715f15ca4ef7e9fdef9353232)), and after that I get the following error:
> 
> ```
> unimplemented: only support cases where input and output size are equal
> fish: Job 3, 'python ../torch-mlir-scripts/to…' terminated by signal SIGABRT (Abort)
> ```
> 
> Which is an error message in the decomposition of `AtenAdaptiveAvgPool2dOp`

Thanks for pointing this out @ramiro050. I have fixed it now.

ramiro050:
Do you know why the TOSA Resnet18 test fails?

vivekkhandelwal1:
> Do you know why the TOSA Resnet18 test fails?

Yes, because of the missing support for AtenAvgPool2d op in TorchToTosa. I have added that, and it works fine now.

vivekkhandelwal1:
@ramiro050, please take a look at this updated PR.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -2688,6 +2722,32 @@ class ConvertAtenPoolingBaseOp : public OpConversionPattern<AtenOpT> {
                                           : nhwcToNchw4DTransposeDims);
   }
 
+  Type getOutputType(RankedTensorType inputTy,
```
ramiro050, line 2725: Since this function is defined in the base class for pooling, it should work for all the dependents; however, from what I can tell, this function will not work for `AtenAdaptivePoolingOp`. If this function can't be made to work with that op, then it should be placed outside this base class

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -2636,6 +2636,40 @@ class ConvertAtenPoolingBaseOp : public OpConversionPattern<AtenOpT> {
     }
   }
 
+  // Checks the validity of pooling parameters and stores them in the respective
+  // vector.
+  LogicalResult
+  checkAndGetPoolingParameters(AtenOpT op, ConversionPatternRewriter &rewriter,
```
ramiro050, line 2642: This function does not work for the op `AtenAdaptiveAvgPool2dOp`, so it should not be in this base class.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -2827,50 +2885,66 @@ class ConvertAtenPoolingOp : public ConvertAtenPoolingBaseOp<AtenOpT, TosaOpT> {
...
-            padArray[1], padArray[1], dilationArray[1]);
-    SmallVector<int64_t> outputShape;
-    if (inputRank > 3)
-      outputShape.push_back(inputShape[0]);
-    outputShape.push_back(outputHDim);
-    outputShape.push_back(outputWDim);
-    outputShape.push_back(inputShape[inputRank - 3]);
-    outputTy = RankedTensorType::get(outputShape, inputElemTy);
+    outputTy = ConvertAtenPoolingBaseOp<AtenOpT, TosaOpT>::getOutputType(
+        inputTy, kernelSize, strideArray, padArray, dilationArray);
```
ramiro050, line 2947: All this code is shared by `ConvertAtenMaxPool2dOp` so it should be refactored

ramiro050 (changes requested):
A few initial comments

ramiro050:
@sjarus just a heads up that this PR will make changes to `TorchToTosa.cpp` to add support for `AtenAvgPool2d`, since `AtenAdaptiveAvgPool2d` has been moved to the decomposition pass. Please take a look at the TOSA changes to make sure they align with your plans for pooling ops.

sjarus:
> @sjarus just a heads up that this PR will make changes to `TorchToTosa.cpp` to add support for `AtenAvgPool2d`, since `AtenAdaptiveAvgPool2d` has been moved to the decomposition pass. Please take a look at the TOSA changes to make sure they align with your plans for pooling ops.

Sounds great, thanks for the heads up!

ramiro050 (changes requested):
Looking good. A few comments below

ramiro050 (approved):
LGTM! Thanks for adding this!

PR #858, opened 2022-05-16
Add non-unit groups to aten::convolution

gpetters94:
Adds support for group sizes > 1 for at::convolution (and by extension the ops that decompose into it).

### DIFF
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -663,33 +663,44 @@ public:
       return rewriter.notifyMatchFailure(op,
                                          "only support constant int dilations");
 
-    Value N = getDimOp(rewriter, loc, input, 0);
+    Value inBatch = getDimOp(rewriter, loc, input, 0);
+    Value inChannels = getDimOp(rewriter, loc, input, 1);
     SmallVector<Value> inDims;
     for (size_t i = 2; i < inRank; i++)
       inDims.push_back(getDimOp(rewriter, loc, input, i));
-    Value F = getDimOp(rewriter, loc, weight, 0);
+    Value weightBatch = getDimOp(rewriter, loc, weight, 0);
+    Value weightChannels = getDimOp(rewriter, loc, weight, 1);
     SmallVector<Value> weightDims;
     for (size_t i = 2; i < inRank; i++)
       weightDims.push_back(getDimOp(rewriter, loc, weight, i));
 
-    // Guard unused values (transposed, groups)
-    int64_t group_size;
-    if (!matchPattern(op.groups(), m_TorchConstantInt(&group_size)) ||
-        group_size != 1)
-      return rewriter.notifyMatchFailure(
-          op, "unimplemented: only group size of 1 supported");
+    // Guard unused values (transposed)
     bool transposed = true;
     if (!matchPattern(op.transposed(), m_TorchConstantBool(&transposed)) ||
         transposed)
       return rewriter.notifyMatchFailure(
           op, "unimplemented: only non-transposed convolution supported");
 
-    // Pad the input tensor according to padding.
-    SmallVector<int64_t, 4> paddingIncludingNC = {0, 0};
-    paddingIncludingNC.insert(paddingIncludingNC.end(), paddingInts.begin(),
-                              paddingInts.end());
-    Value paddedInput = torch_to_linalg::getZeroPaddedTensor(
-        op, rewriter, input, paddingIncludingNC);
+    // Checks for valid group size
+    int64_t groupSize;
+    if (!matchPattern(op.groups(), m_TorchConstantInt(&groupSize)))
+      return rewriter.notifyMatchFailure(op,
+                                         "only constant group size supported.");
+    Value groups = castIntToIndex(rewriter, loc, adaptor.groups());
+
+    auto validate = [&](Value toValidate, std::string err) {
+      Value c0 =
+          rewriter.create<arith::ConstantOp>(loc, rewriter.getIndexAttr(0));
+      Value inputValid = rewriter.create<arith::CmpIOp>(
+          loc, arith::CmpIPredicate::eq, c0,
+          rewriter.create<arith::RemSIOp>(loc, toValidate, groups));
+      rewriter.create<cf::AssertOp>(loc, inputValid,
+                                    rewriter.getStringAttr(err));
+    };
+    validate(inChannels,
+             "invalid: groups must divide input channel size evenly.");
+    validate(weightBatch,
+             "invalid: groups must divide weight batch size evenly.");
 
     SmallVector<Value> paddingIntValues =
         getAsConstantIntValues(rewriter, loc, paddingInts);
@@ -698,7 +709,7 @@ public:
     SmallVector<Value> strideIntValues =
         getAsConstantIntValues(rewriter, loc, strideInts);
 
-    SmallVector<Value> outDims{N, F};
+    SmallVector<Value> outDims{inBatch, weightBatch};
     for (size_t i = 0; i < inRank - 2; i++)
       outDims.push_back(torch_to_linalg::getOutputDimForConvOps(
           rewriter, loc, inDims[i], paddingIntValues[i], dilationIntValues[i],
@@ -708,12 +719,12 @@ public:
         rewriter.create<linalg::InitTensorOp>(loc, outDims, elementType);
 
     Value bias = adaptor.bias();
-    Value biasInitTensor;
+    Value outputTensor;
     if (bias.getType().isa<Torch::NoneType>()) {
       Value c0float = rewriter.create<arith::ConstantOp>(
           loc, FloatAttr::get(elementType, 0.0));
-      biasInitTensor = rewriter.create<linalg::FillOp>(loc, c0float, initTensor)
-                           .getResult(0);
+      outputTensor = rewriter.create<linalg::FillOp>(loc, c0float, initTensor)
+                         .getResult(0);
     } else {
       auto biasType = bias.getType().cast<RankedTensorType>();
       if (biasType.getRank() != 1)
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -549,32 +549,51 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+        rewriter.getStringAttr(
+            "invalid: groups must divide input channel size evenly"));
+    Value weightInputValid = rewriter.create<arith::CmpIOp>(
+        loc, arith::CmpIPredicate::eq, F,
+        rewriter.create<arith::MulIOp>(
+            loc, groups, rewriter.create<arith::FloorDivSIOp>(loc, F, groups)));
+    rewriter.create<cf::AssertOp>(
+        loc, weightInputValid,
+        rewriter.getStringAttr(
+            "invalid: groups must divide weight batch size evenly"));
```
ramiro050, line 591: This can be refactored so that the code that checks for divisibility by `groups` can be reused by `IC` above

ramiro050, line 591: I think it would be cleaner to use `arith.remsi` (https://mlir.llvm.org/docs/Dialects/ArithmeticOps/#arithremsi-mlirarithremsiop) to check for divisibility. You can do `arith.remsi(F, groups) == 0`

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -613,29 +632,79 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
-        rewriter
-            .create<linalg::Conv2DNchwFchwOp>(
-                loc, biasInitTensor.getType(), ValueRange{paddedInput, weight},
-                biasInitTensor, stridesAttr, dilationAttr)
-            .getResult(0);
+    Value inputStride = rewriter.create<arith::FloorDivSIOp>(loc, IC, groups);
+    Value weightStride = rewriter.create<arith::FloorDivSIOp>(loc, F, groups);
+
+    SmallVector<Value> offsets(inRank, rewriter.create<arith::ConstantOp>(
+                                           loc, rewriter.getIndexAttr(0)));
```
ramiro050, line 652: nit: `clang-format`

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -613,29 +632,79 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+                                           loc, rewriter.getIndexAttr(1)));
+    SmallVector<Value> outDimSlice(outDims);
+    outDimSlice[1] = weightStride;
+    SmallVector<Value> inputSliceSizes{N, inputStride};
+    inputSliceSizes.append(inDims);
+    SmallVector<Value> weightSliceSizes{weightStride, WC};
+    weightSliceSizes.append(weightDims);
+
+    for (auto group = 0; group < group_size; group++) {
+
```
ramiro050, line 663: nit: extra empty line

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -549,32 +549,51 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
-          op, "unimplemented: only group size of 1 supported");
+    // Guard unused values (transposed)
     bool transposed = true;
     if (!matchPattern(op.transposed(), m_TorchConstantBool(&transposed)) ||
         transposed)
       return rewriter.notifyMatchFailure(
           op, "unimplemented: only non-transposed convolution supported");
 
+    // Checks for valid group size
+    int64_t group_size;
```
ramiro050, line 570: nit: variable should be camel-case

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -549,32 +549,51 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+            loc, groups, rewriter.create<arith::FloorDivSIOp>(loc, F, groups)));
+    rewriter.create<cf::AssertOp>(
+        loc, weightInputValid,
+        rewriter.getStringAttr(
+            "invalid: groups must divide weight batch size evenly"));
+
     // Pad the input tensor according to padding.
     SmallVector<int64_t, 4> paddingIncludingNC = {0, 0};
     paddingIncludingNC.insert(paddingIncludingNC.end(), paddingInts.begin(),
                               paddingInts.end());
```
ramiro050, line 596: Can you also move this down to where you moved `paddedInput` since it is only used there? Also, I think you can do `padingIncludingNC.append(paddingInts)`

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -613,29 +632,79 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
                          rewriter.getAffineDimExpr(1), context),
           rewriter.getMultiDimIdentityMap(resultRank)};
       SmallVector<StringRef> iteratorTypes(resultRank, "parallel");
```
ramiro050, line 634: Can you also replace `"parallel"` with `getParallelIteratorTypeName()`?

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -613,29 +632,79 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
-    Value conv =
-        rewriter
-            .create<linalg::Conv2DNchwFchwOp>(
-                loc, biasInitTensor.getType(), ValueRange{paddedInput, weight},
-                biasInitTensor, stridesAttr, dilationAttr)
-            .getResult(0);
+    Value inputStride = rewriter.create<arith::FloorDivSIOp>(loc, IC, groups);
+    Value weightStride = rewriter.create<arith::FloorDivSIOp>(loc, F, groups);
+
+    SmallVector<Value> offsets(inRank, rewriter.create<arith::ConstantOp>(
```
ramiro050, line 651: nit: I think it would be clearer to call this `zeroOffsets`

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -613,29 +632,79 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+                                           loc, rewriter.getIndexAttr(0)));
+    SmallVector<Value> strides(inRank, rewriter.create<arith::ConstantOp>(
```
ramiro050, line 653: nit: I think it would be clearer to call this `unitStrides`

ramiro050 (changes requested):
Looking good! I have a few initial comments

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -549,32 +549,51 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
                                          "only support constant int dilations");
 
     Value N = getDimOp(rewriter, loc, input, 0);
+    Value IC = getDimOp(rewriter, loc, input, 1);
     SmallVector<Value> inDims;
     for (size_t i = 2; i < inRank; i++)
       inDims.push_back(getDimOp(rewriter, loc, input, i));
     Value F = getDimOp(rewriter, loc, weight, 0);
+    Value WC = getDimOp(rewriter, loc, weight, 1);
```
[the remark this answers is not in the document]
gpetters94: F is the batch size for the weight, as per `linalg::Conv2DNchwFchw`. IC and WC are the same when groups are 1, but they're different otherwise. The constraint in that case is that they are both divisible by groups.

lib/Conversion/TorchToLinalg/Linear.cpp
gpetters94: Do you mean where it is now? I don't want to put it in the loop.

ramiro050: Yeah, that works!

ramiro050: Yeah, that works!

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -612,30 +623,86 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+    SmallVector<Value> inputSliceSizes{inBatch, inputStride};
+    inputSliceSizes.append(inDims);
+    SmallVector<Value> weightSliceSizes{weightStride, weightChannels};
+    weightSliceSizes.append(weightDims);
+
+    // Pad the input tensor according to padding.
+    SmallVector<int64_t, 4> paddingIncludingNC = {0, 0};
+    paddingIncludingNC.append(paddingInts);
+
+    for (auto group = 0; group < groupSize; group++) {
```
silvasean, line 661: I don't think this is the right approach. This can create a really large amount of IR. We should preserve the fact that this is a grouped convolution rather than splitting it into numGroups individual convolutions. If necessary, we might need to add a linalg named op for this.

silvasean: You might want to look at how depthwise convolution is handled here: https://github.com/tensorflow/mlir-hlo/blob/82c3f9953083be894f88f3253d0434b3cd7490e6/lib/Dialect/mhlo/transforms/legalize_to_linalg.cc#L2076

silvasean: For now, handling just non-grouped and depthwise seems fine. More general groups counts will be more complex and will probably require a linalg named op or something -- we need to see what backends want for that.

gpetters94: Ah, I see. Okay, I'll look into the models that depend on this patch and see how pressing the need for non-depthwise grouped convolutions is.

vivekkhandelwal1: > For now, handling just non-grouped and depthwise seems fine. More general groups counts will be more complex and will probably require a linalg named op or something -- we need to see what backends want for that.

Hi @silvasean, can we get this patch merged as of now? Since it will unblock some of the models required by us. @gpetters94 will continue working on your suggestions and modify this patch or add an op in the Linalg dialect.

powderluv: Ouch. Agree it would be good to keep the quality bar and we can carry the patch locally. It would be nice to get this kind of feedback earlier so we don't abandon a patchset after two weeks. Please let us know if there is a better way to front load this kind of feedback. I think @gpetters94 was going off @cathyzhyi reference implementation pointed. I will continue to push for an RFC before an implementation.

gpetters94: I can try my hand at implementing a named op in LLVM. Do you think the same logic I'm using here (splitting grouped convs into groups of convs) will be appropriate at the linalg level?

gpetters94: Adding a special case for depthwise might also need a new op upstream, since the current ops don't support NchwFchw order. I could do three transpositions but I don't think that's the best way to do it.

silvasean: Ah, yeah, then we should add the new op upstream too to avoid the transpositions.

gpetters94:
Okay, most recent commit is confirmed working with my new op in LLVM. Once that lands and we can push torch-mlir's LLVM version up this should be good to go.

gpetters94:
Now that the LLVM op has landed, here's the updated PR using the new logic. I spoke with Mahesh about adding a depthwise case and my concerns that it won't be possible with dynamic tensors because I won't be able to do the necessary calculation at compile time. He suggested that the optimization to depthwise convolution could be done on the backend (i.e. IREE) side. If you think I should add a specific case for depthwise static tensors then I can do that, otherwise this is complete.

silvasean:
> Now that the LLVM op has landed, here's the updated PR using the new logic. I spoke with Mahesh about adding a depthwise case and my concerns that it won't be possible with dynamic tensors because I won't be able to do the necessary calculation at compile time. He suggested that the optimization to depthwise convolution could be done on the backend (i.e. IREE) side. If you think I should add a specific case for depthwise static tensors then I can do that, otherwise this is complete.

I think we should support the special case when enough static info is available (the whole shape doesn't have to be static for us to detect it). It is a very important pattern in practice and we don't want all backends replicating that code.

gpetters94:
> > Now that the LLVM op has landed, here's the updated PR using the new logic. I spoke with Mahesh about adding a depthwise case and my concerns that it won't be possible with dynamic tensors because I won't be able to do the necessary calculation at compile time. He suggested that the optimization to depthwise convolution could be done on the backend (i.e. IREE) side. If you think I should add a specific case for depthwise static tensors then I can do that, otherwise this is complete.
> 
> I think we should support the special case when enough static info is available (the whole shape doesn't have to be static for us to detect it). It is a very important pattern in practice and we don't want all backends replicating that code.

Okay, I've put in a PR at LLVM for the new op here: https://reviews.llvm.org/D128575. It's only for the case where the multiplier is 1, which covers Mobilenet. I can add the non-unit multiplier case if you think it's relevant, but for now I'm going to check for multiplier of 1 and fall through to the grouped op otherwise.

gpetters94:
Testing my patch with other models (i.e. not Mobilenet) exposed an issue in the ordering of the op I wrote in upstream. This latest commit fixes the issue, although it does require a change to the upstream op which I have put up at https://reviews.llvm.org/D128880.

gpetters94:
Both LLVM patches have landed, now this is just waiting for an LLVM bump before it's good to go.

Issue #854, opened 2022-05-13
can we compile a model without shape information?


FullZing:
def compile(model: torch.nn.Module,
            example_args: List[torch.Tensor],
            output_type: OutputType = OutputType.TORCH,
            use_tracing=False):
    """Convert a PyTorch model to MLIR.

    Args:
        model: The PyTorch model to convert.
        example_args: A list of example arguments to use when inferring the
            shapes of the arguments to `forward` method of the model.
            A single tensor is treated as a list of a single tensor.
        output_type: The kind of output to produce. See `OutputType` for more
            details.
        use_tracing: If True, use `torch.jit.trace` to convert the model to
            JIT IR rather than `torch.jit.script`.

compile api needs tensor information

silvasean:
Generally we can't do much without the tensor information. I suppose we could allow None for example_args, but I don't think the output will be very useful. What is your use case?

FullZing:
my cases are RNN models, the input tensor is not fixed
as far as i understand,  tensor information is used for shape inference which locates after ModuleBuilder https://github.com/llvm/torch-mlir/blob/cfc1a6515c00badbf15bff8e3f2e796d3199b8e0/python/torch_mlir/__init__.py#L83
 
why does ModelBuilder https://github.com/llvm/torch-mlir/blob/cfc1a6515c00badbf15bff8e3f2e796d3199b8e0/python/torch_mlir/__init__.py#L80 need tensor information?

silvasean:
We need to distinguish between 1) not knowing any shape information, compared to 2) knowing the number of dimensions (but not knowing the size of the dimensions). The case of RNN's is typically 2) I think. I have posted a PR adding support for that: https://github.com/llvm/torch-mlir/pull/859

The case 1) is rarer, and will prevent Torch-MLIR from doing most of its passes effectively.

PR #859, opened 2022-05-16
torch_mlir.compile: add support for dynamic sizes.

silvasean:
We do this by inroducing a TensorPlaceholder class, which can be used to
specify dynamic sizes. Internally, we canonicalize all example inputs
to TensorPlaceholder's.

This commit also adds some basic testing, which was missing before.

Fixes https://github.com/llvm/torch-mlir/issues/854


silvasean:
cc @FullZing

PR #860, opened 2022-05-16
Update README.md

powderluv:


silvasean:
Thanks. Let me refactor the meeting itself and then I will add it with the right link and stuff.

powderluv:
bump so I can point new folks to the right meeting for next week :D

PR #861, opened 2022-05-18
Add neural_net and bert_training scripts.

pashu123:
1. With the help of `make_fx` we are able to get the full training graph
   with weight updates.
2. NeuralNet_training passes. Bert_training passes after cherry-picking
   https://github.com/llvm/torch-mlir/pull/844.
3. TODO: Remove the functorch's dependency after make_fx moves to
   pytorch core.


build_tools/torchscript_e2e_heavydep_tests/train_models.py
```diff
@@ -0,0 +1,206 @@
...
+                       torch.ops.aten.select_backward
+                   ]))(dict(model.named_parameters()),
+                       dict(model.named_buffers()), inputs)
+    fx_g.graph.set_codegen(torch.fx.graph.CodeGen())
+    fx_g.recompile()
+    fx_g = change_fx_graph_return_to_tuple(fx_g)
+    ts_g = torch.jit.script(fx_g)
+    # TODO: If not saved/load creates some unnecessary functions that
+    # causes problem during mlir-translate.
+    ts_g.save("temp.ts")
```
silvasean, line 78: Can you use tempfile.TemporaryFile instead to avoid creating a bunch of garbage on the filesystem?

silvasean (approved):

pashu123:
> With `make_fx` and decompositions moved into core with [pytorch/pytorch#74360](https://github.com/pytorch/pytorch/pull/74360) and [pytorch/pytorch#76311](https://github.com/pytorch/pytorch/pull/76311), I guess we are ready for (3)?

I see, I am still waiting for this PR to be merged https://github.com/pytorch/pytorch/pull/77474

build_tools/torchscript_e2e_heavydep_tests/train_models.py
pashu123: Just using the temp.TemporaryFile  is giving `  File "/home/prashant/torch-mlir/mlir_venv/lib/python3.9/site-packages/torch/jit/_serialization.py", line 164, in load
    cpp_module = torch._C.import_ir_module_from_buffer(  
RuntimeError: Unrecognized data format ` error. Instead using temp.NamedTemporaryFile and using the .name attribute which stores at `/tmp/...` dir.

byronyi:
Unfortunately, calling `loss.backward()` inside `make_fx` is unsound. 

See the following example:

```python
import torch
from torch.fx.experimental.proxy_tensor import make_fx

def step_fn(x):
    x.backward()
    return x.grad

# Users can register their own hooks to replace grad
# e.g. grad clipping or grad scaling
# clipping: x.register_hook(lambda grad: torch.clamp(grad, -1.0, 1.0))
x = torch.rand((1,), requires_grad=True)
print(step_fn(x))  # tensor([1.])
print(step_fn(x))  # tensor([2.])

x = torch.rand((1,), requires_grad=True)
fx_g = make_fx(step_fn)(x)
print(fx_g(x))  # tensor([1.])
print(fx_g(x))  # tensor([1.])
```

Grad accumulation is a built-in backward hook, and no hooks are in effect inside tracing context.

byronyi:
Looks like it is only sound to capture the graph beneath PyTorch autograd, which implies whole graph capture only makes sense with LazyTensor.

silvasean:
Hey @byronyi, thanks for the info -- that's really not obvious! Is there a specific improvement you would like to see here, or was this just an FYI?

byronyi:
@silvasean Ideally I would like to generate TS for the entire training step using proxy tensor, and this PR is well-shaped demo for the simple case.

However, not able to calling into the backward hook makes it difficult to generalize to realistic training scenarios, where various grad transforms (clipping, normalization, accumulation, and collective communication) is ubiquitous and some of them are even built-in behaviors of PyTorch.

IMHO there are at least two fixes we need for this:

1. carry the backward hooks from args to proxied args (based on work in https://github.com/pytorch/pytorch/pull/77474)
2. add the grad of args as implicit args to the function being traced, to address grad accumulation hook in particular:

```python3
def step_fn(x, x_grad):
    x.grad = x_grad
    x.backward()
    return x.grad
```

Something like this, after initializing `x.grad` in eager (requires calling backward once before make_fx), would generate the correct fx graph:

```
def forward(self, x_1, x_grad_1):
    ones_like_default = torch.ops.aten.ones_like.default(x_1, dtype = torch.float32, layout = torch.strided, device = device(type='cpu'), pin_memory = False, memory_format = torch.preserve_format);  x_1 = None
    add__tensor = torch.ops.aten.add_.Tensor(x_grad_1, ones_like_default);  x_grad_1 = ones_like_default = None
    return add__tensor
```

byronyi:
Gently ping @ezyang @chillee; do you have any advice on capturing backward hooks inside make_fx?

ezyang:
We're aware of the hook deferral problem eliminating pipelining in AOTAutograd. Our short term workaround in our universe is to manually split up the graphs, e.g. as seen in https://github.com/pytorch/torchdynamo/pull/628

As for `make_fx`, you're on the right track. Actually, the problem is not that the hooks are not traceable (they are), but that `make_fx` can't trace things that happen "outside". So for example, if you had done autograd with `torch.autograd.grad`, this would correctly trace out all the gradients, and if you had installed backward hooks on them, those would trace out correctly. You can see that hooks DO trace; take the example in https://pytorch.org/functorch/stable/notebooks/aot_autograd_optimizations.html and edit it to be

```
def fn(a, b, c, d):
    x = a + b + c + d 
    x.register_hook(lambda grad: grad * 2)
    return x.cos().cos()
```

So if you can arrange for everything to happen "inside" make_fx, you can trace it. For backward(), that implies that you have to allocate the leaf variables inside make_fx, and then explicitly return the grad (there's a representation problem, btw: there's no way to say "set the grad on this variable to this value" in the IR, and we are unable to trace this, though maybe we should.)

I think there are some improvements to tracing to make this sort of thing easier to do, but at the end of the day, mediating external side effects is going to require actually writing some more code, along the lines of what we've been calling AOTDispatch

also cc @bdhirsh, this is similar to what we have to do with `copy_` in functionalized graphs.

PR #862, opened 2022-05-18
[TORCH] Add decomposition of `aten.linear` op

Shukla-Gaurav:
This commit adds decomposition of `aten.linear` op.

TODO: remove existing implementation and add a simple test case. Also handle the (3D,2D) case of matmul.

Signed-Off-by: Gaurav Shukla

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -2061,6 +2061,55 @@ public:
 } // namespace
 
 namespace {
+// Decompose `aten.linear` op into `aten.matmul` and `aten.add` ops.
+class DecomposeAtenLinearOp : public OpRewritePattern<AtenLinearOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenLinearOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value input = op.input();
+    Value weight = op.weight();
+    Value bias = op.bias();
+
+    BaseTensorType inputType = input.getType().cast<BaseTensorType>();
+    if (!inputType.hasSizes() || inputType.getSizes().size() < 2)
+      return rewriter.notifyMatchFailure(
+          op, "expected input to be rank 2 or greater");
+
+    BaseTensorType weightType = weight.getType().cast<BaseTensorType>();
+    // `weight` must be a rank 2 matrix.
+    if (!weightType.hasSizes() || weightType.getSizes().size() != 2)
+      return rewriter.notifyMatchFailure(op, "expected weight to be a rank 2");
+
+    SmallVector<int64_t> transposeShape =
+        llvm::to_vector(llvm::reverse(weightType.getSizes()));
+    Type transposeType = weightType.getWithSizesAndDtype(
+        llvm::makeArrayRef(transposeShape), weightType.getDtype());
+    Value transposeWeight =
+        rewriter.create<AtenTOp>(loc, transposeType, weight);
+
+    Value matmul = rewriter.create<AtenMatmulOp>(loc, op.getType(), input,
+                                                 transposeWeight);
+    if (bias.getType().isa<Torch::NoneType>()) {
+      rewriter.replaceOp(op, matmul);
+      return success();
+    }
+
+    BaseTensorType biasType = bias.getType().cast<BaseTensorType>();
+    if (!biasType.hasSizes() || biasType.getSizes().size() != 1)
+      return rewriter.notifyMatchFailure(op, "expected bias to be rank 1");
+
+    Value alpha =
+        rewriter.create<ConstantFloatOp>(loc, rewriter.getF64FloatAttr(1));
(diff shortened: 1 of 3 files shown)


silvasean:
Remove the TorchToLinalg code as well please since it is not not useful.

Shukla-Gaurav:
Once https://github.com/llvm/torch-mlir/pull/951 is merged, this PR will be unblocked wrt (3D, 2D) case of matmul.

Shukla-Gaurav:
@sjarus Could you please take a look into this, `ResNet18Static` is failing in torch to tosa lowering. Thanks!

sjarus:
I'll attempt to repro this early next week . Right now I'm trying to get a pipeline of pending BERT related legalizations out. Do you have detailed verbose output from the test ?

Shukla-Gaurav:
@sjarus There is a case where `aten.matmul` fails to lower to tosa. I quickly looked into it but did not get the solution. Could you please look into it?
`%3 = torch.aten.matmul %arg0, %2 : !torch.vtensor<[?,?,?],f32>, !torch.vtensor<[3,5],f32> -> !torch.vtensor<[?,?,5],f32>`.

sjarus:
What happens when you use 3, 4, 3 instead of -1, -1, -1 in the annotate args ?

Shukla-Gaurav:
@sjarus Did you get a chance to look at it?

sjarus:
Right, sorry for the delay. The experiment result appears aligned to my expectations, but I agree that if dynamic dims are unsupported then it should fail gracefully and not assert. 

It seems the best way to unblock is to keep the test static and refer in a comment to me to update the code so it doesn't assert, or better yet can pass, tho the latter is typically harder. Would that work ?

Shukla-Gaurav:
@sjarus I have made the test case partially static for now in order to pass the tosa test. Should I create a new issue for the same one?
FYI @silvasean

silvasean:
We definitely don't want to regress our testing of dynamic shapes for all backends due a limitation in the TOSA backend, so I would avoid changing the test. We can use the functionality in https://github.com/llvm/torch-mlir/pull/1242 to control the decompositions, but I need to make a PR to centralize that logic first -- will try to do that today.

@Shukla-Gaurav can you also file a bug for the TOSA assertion failure? Ideally include a `torch-mlir-opt` command line invoking the failing pass and a reduced IR snippet.

silvasean:
> We definitely don't want to regress our testing of dynamic shapes for all backends due a limitation in the TOSA backend, so I would avoid changing the test. We can use the functionality in #1242 to control the decompositions, but I need to make a PR to centralize that logic first -- will try to do that today.

See https://github.com/llvm/torch-mlir/pull/1248

Shukla-Gaurav:
@silvasean Thanks for adding support for more control over decomposition based on backend. Could you please review this again?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1978,6 +1978,57 @@ class DecomposeAtenFullOp : public OpRewritePattern<AtenFullOp> {
...
+    if (!inputType.hasSizes() || inputType.getSizes().size() < 2)
+      return rewriter.notifyMatchFailure(
+          op, "expected input to be rank 2 or greater");
+
+    BaseTensorType weightType = weight.getType().cast<BaseTensorType>();
+    // weight must be a rank 2 matrix.
+    if (!weightType.hasSizes() || weightType.getSizes().size() != 2)
+      return rewriter.notifyMatchFailure(op, "expected weight to be a rank 2");
+
+    ArrayRef<int64_t> wShape = weightType.getSizes();
```
silvasean, line 2003: This does not work with dynamic shapes.

silvasean (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
silvasean: Oh wait, yeah it does. Sorry. Usually getSizes() is not good to use for dynamic shapes but in this case it does work.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1978,6 +1978,57 @@ class DecomposeAtenFullOp : public OpRewritePattern<AtenFullOp> {
...
+    Value weight = op.weight();
+    Value bias = op.bias();
+
+    BaseTensorType inputType = input.getType().cast<BaseTensorType>();
+    if (!inputType.hasSizes() || inputType.getSizes().size() < 2)
+      return rewriter.notifyMatchFailure(
+          op, "expected input to be rank 2 or greater");
+
+    BaseTensorType weightType = weight.getType().cast<BaseTensorType>();
+    // weight must be a rank 2 matrix.
```
silvasean, line 1999: nit:
```
`weight`
```
with backticks

silvasean (approved):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
silvasean: nit: I would recommend doing `llvm::to_vector(llvm::reverse(weightType.getSizes())` for this.

Issue #863, opened 2022-05-19
unsupported by backend lowering: torch.prim.TupleIndex


Tengxu-Sun:
I'm trying to export bert model to MLIR with `output_type = LINALG_ON_TENSORS`. But got this error message. 
```
Exception: 
Lowering Torch Backend IR -> Linalg-on-Tensors Backend IR failed with the following diagnostics:
error: unsupported by backend lowering: tensor with unknown rank or dtype
note: see current operation: %789 = "torch.prim.TupleIndex"(%788, %221) : (!torch.tuple<tensor<[1,14,768],f32>, tensor<[1,768],f32>>, !torch.int) -> !torch.vtensor
note: this is likely due to a missing shape transfer function in shape_lib_gen.py
```

When i set `output_type = TORCH`, it convert successfully and the MLIR information in the end is
```
    ......
    %670 = torch.aten.linear %669, %4, %5 : !torch.vtensor<[1,14,3072],f32>, !torch.vtensor<[768,3072],f32>, !torch.vtensor<[768],f32> -> !torch.vtensor<[1,14,768],f32>
    %671 = torch.aten.add.Tensor %670, %result0_136, %int1 : !torch.vtensor<[1,14,768],f32>, !torch.vtensor<[1,14,768],f32>, !torch.int -> !torch.vtensor<[1,14,768],f32>
    %result0_139, %result1_140, %result2_141 = torch.aten.native_layer_norm %671, %224, %2, %3, %float9.999990e-13 : !torch.vtensor<[1,14,768],f32>, !torch.list<int>, !torch.vtensor<[768],f32>, !torch.vtensor<[768],f32>, !torch.float -> !torch.vtensor<[1,14,768],f32>, !torch.vtensor<[1,14,1],f32>, !torch.vtensor<[1,14,1],f32>
    %672 = torch.copy.to_tensor %result0_139 : !torch.tensor<[1,14,768],f32>
    %673 = torch.aten.slice.Tensor %672, %int0, %int0, %int9223372036854775807, %int1 : !torch.tensor<[1,14,768],f32>, !torch.int, !torch.int, !torch.int, !torch.int -> !torch.tensor<[1,14,768],f32>
    %int1_142 = torch.constant.int 1
    %674 = torch.aten.add.int %int0, %int1_142 : !torch.int, !torch.int -> !torch.int
[...]

PR #865, opened 2022-05-19
torch.prim.TupleIndex: Adjust tensor types when folding.

silvasean:
In cases where a refinement/derefinement was needed, we didn't fold.

Fixes https://github.com/llvm/torch-mlir/issues/863

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -1414,14 +1414,20 @@ void PrimTupleIndexOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
     if (i >= (int64_t)tupleConstruct.elements().size())
       return failure();
 
-    Value element = tupleConstruct.elements()[i];
     // TODO: We should have a clear picture of whether we want to consistently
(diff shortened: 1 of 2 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1414,14 +1414,20 @@ void PrimTupleIndexOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
     if (i >= (int64_t)tupleConstruct.elements().size())
       return failure();
 
-    Value element = tupleConstruct.elements()[i];
     // TODO: We should have a clear picture of whether we want to consistently
     // allow refinement, and where. It seems desirable to require precise
     // type equality for TupleConstruct / TupleIndex, but that might break
     // things.
```
ramiro050, line 1420: nit: The "desirable to require precise type equality" part no longer applies

ramiro050 (approved):
LGTM! Just a small nit

lib/Dialect/Torch/IR/TorchOps.cpp
silvasean: I think it actually does still seem desirable, but is not the current state.

PR #867, opened 2022-05-19
[MLIR][TORCH] Add E2E support for `aten.sqrt.int` op

vivekkhandelwal1:
This commit adds lowering of `aten.sqrt.int` op.


ramiro050 (changes requested):

ramiro050 (approved):
LGTM!

PR #868, opened 2022-05-20
[MLIR][TORCH] Add E2E support for `aten.Bool.[float|int]` op

vivekkhandelwal1:
This commit adds lowering of `aten.Bool.float` and `aten.Bool.int` op.

### DIFF
--- a/lib/Conversion/TorchToStd/TorchToStd.cpp
+++ b/lib/Conversion/TorchToStd/TorchToStd.cpp
@@ -225,6 +225,33 @@ public:
 };
 } // namespace
 
+namespace {
+template <typename OpTy, typename CmpOpTy, typename CmpOpPred, CmpOpPred Pred>
+class ConvertAtenBoolLikeOp : public OpConversionPattern<OpTy> {
+public:
+  using OpConversionPattern<OpTy>::OpConversionPattern;
+  using OpAdaptor = typename OpTy::Adaptor;
+  LogicalResult
+  matchAndRewrite(OpTy op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Type inputType = adaptor.a().getType();
+    Value cstZero = rewriter.create<arith::ConstantOp>(
+        loc, rewriter.getZeroAttr(inputType));
+    Value cstTrue =
+        rewriter.create<arith::ConstantOp>(loc, rewriter.getBoolAttr(true));
+    Value cstFalse =
+        rewriter.create<arith::ConstantOp>(loc, rewriter.getBoolAttr(false));
+
+    Value cmpPred;
+    cmpPred = rewriter.create<CmpOpTy>(loc, Pred, adaptor.a(), cstZero);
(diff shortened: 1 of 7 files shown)


lib/Conversion/TorchToStd/TorchToStd.cpp
```diff
@@ -190,6 +190,41 @@ class ConvertTorchConstantOp : public OpConversionPattern<OpTy> {
...
+    Type inputType = adaptor.a().getType();
+    Value cstZero = rewriter.create<arith::ConstantOp>(
+        loc, rewriter.getZeroAttr(inputType));
+    Value cstTrue =
+        rewriter.create<arith::ConstantOp>(loc, rewriter.getBoolAttr(true));
+    Value cstFalse =
+        rewriter.create<arith::ConstantOp>(loc, rewriter.getBoolAttr(false));
+
+    Value pred;
+    if (inputType.isa<mlir::IntegerType>()) {
```
ramiro050, line 212: I'm usually a bit reluctant to have a dynamic cast inside a template because part of the allure of templates is that it avoids the runtime overhead of a dynamic cast. Here I would do something similar to how the comparison ops are lowered. I would add two extra template parameters: one for the comparison op (for example `arith::CmpIOp`) and the other for the comparison predicate.

https://github.com/llvm/torch-mlir/blob/247d05b13c9239da13186da7a7646c5a97cf80b5/lib/Conversion/TorchToStd/TorchToStd.cpp#L119

python/torch_mlir_e2e_test/test_suite/scalar.py
```diff
@@ -151,3 +151,123 @@ def forward(self, lhs, rhs):
...
+        None,
+        ([], torch.float64, True),
+    ])
+    def forward(self, a):
+        return bool(torch.ops.aten.Bool(float(a)))
+
+
+@register_test_case(module_factory=lambda: BoolFloatTrueModule())
+def BoolFloatTrueModule_basic(module, tu: TestUtils):
+    module.forward(torch.rand(()).double())
```
ramiro050, line 195: I think `torch.rand` could in theory output `0.0`, right? I would use `tu.rand(low=0.5)`

python/torch_mlir_e2e_test/test_suite/scalar.py
```diff
@@ -151,3 +151,123 @@ def forward(self, lhs, rhs):
...
+        None,
+        ([], torch.int64, True),
+    ])
+    def forward(self, a):
+        return bool(torch.ops.aten.Bool(int(a)))
+
+
+@register_test_case(module_factory=lambda: BoolIntTrueModule())
+def BoolIntTrueModule_basic(module, tu: TestUtils):
+    module.forward(torch.randint(100, ()))
```
ramiro050, line 255: `torch.randint` by default has `0` as its lower bound. Can you specify a higher lower bound so that it doesn't accidentally output `False`? (For both the `BoolFloatTrue` and `BoolIntTrue`, the test is not gonna fail if the input is zero, but just to be consistent with the name of the test)

ramiro050 (changes requested):
A few small comments below

ramiro050 (approved):
LGTM!

PR #870, opened 2022-05-20
linalg: support more cases in lowering reshape op

ashay:
Prior to this patch, PyTorch's reshape and view operations were lowered
to either `tensor.expand_shape` or `tensor.collapse_shape` ops, which
prevented us from lowering reshape operations for which (1) the source
and destination shape involved both collapsing as well as expansion or
(2) the source and destination shapes had the same rank but not the same
values in each dimension.

A better solution is to use `tensor.reshape`, which handles both of the
above-mentioned cases.  This patch augments the existing lowering of
reshape operations from the torch dialect to the tensor dialect, and it
adds two end-to-end tests to validate the translation.

However, upstream LLVM currently supports lowering reshape operations
only when all shape values are statically known, so this patch adds a
guard condition to ensure that the subsequent lowering to the LLVM
dialect will not fail.


python/torch_mlir_e2e_test/test_suite/reshape_like.py
```diff
@@ -429,3 +429,40 @@ def forward(self, a):
 def ReshapeAliasCollapseModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(2, 4))
 
+# ==============================================================================
+
+class ReshapeSameRankModule(torch.nn.Module):
```
ramiro050, line 434: nit: can you move these two tests up so that they are next to the other `Reshape` tests?

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -135,30 +135,24 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
-    // equal, and the input itself is the result. We do not handle expanding And
-    // collapsing happening at the same time or cases where it's neither
-    // collapsing nor expanding like view of [2,3] for 3x2 tensor.
-    // TODO: For the expanding And collapsing case, we will need to identify
-    // which dimensions are collapsing and which are expanding and do it in two
-    // steps.
-    // TODO: For neither collapsing nor expanding, we could find a intermediate
-    // shape to collapse and then expanded to the target shape. Like [2,3] =>
-    // [6] => [3, 2].
-    if (inputRank == resultRank) {
```
ramiro050, line 155: I think there is still value to keeping this here, since it handles some cases that are not handled by the `ReshapeOp`, namely, the case where sizes are dynamic. I would just update the comment to mention that once `ReshapeOp` handles dynamic sizes, if ever, then this will no longer be needed.

ramiro050 (approved):
LGTM! Just a few small comments

silvasean (changes requested):
I don't think this is the right approach. If there is expansion as well as contraction, then we should lower it to an expand shape and a collapse shape reshape ops.

The tensor.reshape op isn't very well designed because it is effectively non-analyzable. (The correspondents in the `mhlo` dialect have caused significant trouble in the past, it seems that somehow that design leaked upstream). Also, we shouldn't create tensors which represent shapes. That is just a bad design approach and doesn't play well with backends.

Or to put it another way, the tensor.reshape op is not part of our backend contract so we cannot use it in TorchToLinalg.

ashay:
Closing this for now, since it will take a while to implement view ops using just collapse and expand.

PR #871, opened 2022-05-20
build: fix code so that the compiler does not emit warnings

ashay:
When compiling without assertions (i.e. in `NDEBUG` mode), a handful of
statements turn to NOPs, which results in warnings such as missing
return statement or unused variables and function.  This patch replaces
such statements with `llvm_unreachable()`, which informs the compiler
about program termination regardless of the `NDEBUG` mode.  This also
enables torch-mlir to be compiled using the flags `-Wall`, `-Wextra`,
`-Wpedantic`, and `-Werror`.

### DIFF
--- a/externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
+++ b/externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
@@ -403,6 +403,7 @@ bool ScatterOp::payloadUsesValueFromOperand(OpOperand *opOperand) {
     bbArgNumber = 0; // block arg 0 is `update`.
   else {
     bool isValidOperand = operand == indices() || operand == original();
+    (void)isValidOperand;
     assert(isValidOperand &&
            "operand must belong to the current tm_tensor.scatter op");
     return true;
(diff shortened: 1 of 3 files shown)


externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
```diff
@@ -403,8 +403,9 @@ bool ScatterOp::payloadUsesValueFromOperand(OpOperand *opOperand) {
     bbArgNumber = 0; // block arg 0 is `update`.
   else {
     bool isValidOperand = operand == indices() || operand == original();
-    assert(isValidOperand &&
-           "operand must belong to the current tm_tensor.scatter op");
+    if (!isValidOperand)
```
silvasean, line 406: the standard way to deal with these unused-if-no-asserts variables in LLVM is by casting to void such as:

```
bool foo = ...;
(void)foo;
assert(foo && ...)
```

silvasean (changes requested):

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
ashay: Thanks @silvasean!  Fixed in the most recent commit.

ashay:
Rebased with `main`.

silvasean (approved):

PR #874, opened 2022-05-24
Add E2E support for aten.all.bool op

vid-999:


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -641,6 +641,49 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenAllBoolOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    //bool list of type::bool[]
+    Value inputList = op.self();
+
+    SmallVector<Value> inputValues;
```
pashu123, line 656: Please clang-format your code. `git clang-format HEAD~`

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -641,6 +641,49 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+      bool elementVal;
+      if(!matchPattern(element, m_TorchConstantBool(&elementVal))){
+        return rewriter.notifyMatchFailure(
+          op, "could not pattern match the value as a bool!"
+        );
+      }
+      inputValsBool.push_back(elementVal);
+    }
+    
+    //reduce the bool array into a single value
```
pashu123, line 675: Comments should end with a full stop.

pashu123 (changes requested):
Nits.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -641,6 +641,45 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
 };
 } // namespace
 
+namespace {
+class ConvertAtenAllBoolOp : public OpConversionPattern<AtenAllBoolOp> {
```
ramiro050, line 645: The lowering for `AtenAnyBoolOp` was recently added, and it is almost the same as the one needed for this op. Can you refactor this code

https://github.com/llvm/torch-mlir/blob/014a6d16c7891d194e9031acba21b85e2a89453e/lib/Conversion/TorchToStd/TorchToStd.cpp#L206-L223

into a helper static function that takes the reduction function as an argument (in your case `llvm::all_of`) so that both lowerings can share as much of the logic as possible?

lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
```diff
@@ -29,11 +29,11 @@ StringRef mlir::torch::Torch::getShapeLibrary() {
   constexpr StringLiteral shapeLib(R"mlir(
 module {
   func.func @"__torch_mlir_shape_fn.aten.tanh"(%arg0: !torch.list<int>) -> !torch.list<int> {
-    %0 = func.call @__torch__.torch_mlir.dialects.torch.importer.jit_ir.build_tools.upstream_shape_helpers.unary(%arg0) : (!torch.list<int>) -> !torch.list<int>
```
ramiro050, line 32: Do you know what happened here?

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
vid-999: hmm, I don't know it seems like all func.call got changed to call

lib/Conversion/TorchToLinalg/Linear.cpp
vid-999: fixed this in the latest commit.

ramiro050:
There seems to be a lot of changes in this PR that are from other PRs. I think this may have happened from merging main into your feature branch. In general, to get new changes into your feature branch, it is better to do `git rebase main`. Can you please remove the changes that don't belong here?

vivekkhandelwal1:
> > There seems to be a lot of changes in this PR that are from other PRs. I think this may have happened from merging main into your feature branch. In general, to get new changes into your feature branch, it is better to do `git rebase main`. Can you please remove the changes that don't belong here?
> 
> Did the latest forced-push i did address this?

You have to mark the comments as resolved if you have addressed them.

vivekkhandelwal1:
@vid-999, once you address comments and push the latest changes to the PR, please ask for a re-review.

lib/Conversion/TorchToStd/TorchToStd.cpp
```diff
@@ -216,8 +218,17 @@ class ConvertAtenAnyBoolOp : public OpConversionPattern<AtenAnyBoolOp> {
...
+    
+    bool result = true;
+    if (auto getBoolOp = llvm::dyn_cast<AtenAnyBoolOp>(&op)) {
+       result = llvm::any_of(
         inputListBool, [](bool inputListElem) { return inputListElem; });
+    }
+    else if (auto getBoolOp = llvm::dyn_cast<AtenAllBoolOp>(&op)){
+       result = llvm::all_of(
+        inputListBool, [](bool inputListElem) { return inputListElem; });
+    }
```
ramiro050, line 230: I'm usually a bit reluctant to have `dyn_cast` inside of a template because they're design choices that conflict on their goals. One of the main benefits of templates is to allow polymorphism without any runtime overhead. However, a `dyn_cast` does require runtime overhead, so it counteracts the benefits of the template.

For what you're trying to do here there are a couple of common design patterns. The first is to do something like what is used in the https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchToTosa/TorchToTosa.cpp file. Namely, you first create a base template class

```cpp
template <typename AtenOpT>
class ConvertAtenAnyOrAllBoolOp : public OpConversionPattern<AtenOpT> {
  // Method to be implemented by classes that inherit from this class
  virtual bool reduceBoolArray(ArrayRef<bool> boolArray) const;
  // ...
  LogicalResult matchAndRewrite(AtenOpT op, ...) const override {
    // ...
    bool result = reduceBoolArray(inputListBool);
    // ...
  }
}

class ConvertAtenAnyBoolOp : public ConvertAtenAnyOrAllOp<AtenAnyBoolOp> {
  bool reduceBoolArray(ArrayRef<bool> boolArray) const override {
    return llvm::any_of(boolArray, [](bool inputListElem) { return inputListElem; });
  }
}

// same for `AtenAllBoolOp`
```

The other pattern is to create a static helper function instead of a base class. You would do something like 

```cpp
static LogicalResult computeAnyOrAllBoolOp(// whatever argument are needed,
                 function_ref<bool(ArrayRef<bool>, function_ref<bool(bool)>)> reduceBoolArray) {
  return reduceBoolArray(inputListBool, [](bool inputListElem) { return inputListElem; });
}
```

Then define one class for each op and the `matchAndRewrite` body will just be the helper function.

Happy to help with any questions

ramiro050 (changes requested):
A few more comments

vid-999:
> Make sure that LLVM is at the right commit. After this bump https://github.com/llvm/torch-mlir/commit/bb52a460cb101ed63d73357953fbbb544cd62c59, call was changed to func.call. Normally you can do git submodule update and that will set the right LLVM commit. You then should rebuild, then run the ./build_tools/update_shape_lib.sh, and that should get rid of all these changes.

Hi, I changed the llvm tag to https://github.com/llvm/torch-mlir/commit/bb52a460cb101ed63d73357953fbbb544cd62c59, rebuilt it but still I do not see the call changed to fun.call after updating the shape library using the script in build_tools.

ramiro050 (approved):
LGTM! Just a small nit. Also, I would wait until the issue with the `shape` library is sorted out before merging. If the changes are indeed needed, then it should be its own PR since they are unrelated to this PR

vid-999:
> LGTM! Just a small nit. Also, I would wait until the issue with the `shape` library is sorted out before merging. If the changes are indeed needed, then it should be its own PR since they are unrelated to this PR

It looks like it got resolved just now?

PR #885, opened 2022-05-30
torch_mlir.compile: Fix API footgun

silvasean:
use_tracing=True was behaving unexpectedly because the handling of
single arguments was happening after the torch.jit.trace call.


ramiro050 (approved):
LGTM! Thanks for the fix!

Issue #887, opened 2022-05-30
Adding ops without e2e support


henrytwo:
Hi @silvasean, we're now able to trace a Huggingface BERT model with full forward and backwards graph, but we had to add support for a few ops to torch-mlir.

Looking at some existing PRs (e.g. https://github.com/llvm/torch-mlir/pull/570) it seems that adding support for an op usually comes with type refinement, lowering to linalg, and e2e tests (and the e2e tests won't work without the type refinement).

Since we don't need any of this other stuff for LTC, would a PR with only updates to `torch_ods_gen.py` and `GeneratedAtenOps.td` be acceptable?

cc: @antoniojkim @ke1337

silvasean:
Yeah, I guess that would be ok.

henrytwo:
> Can you list which ops are missing so we can add them to the e2e todo list?
> 
> CC @ramiro050 who is tracking that TODO list.

- [aten.native_layer_norm_backward](https://github.com/llvm/torch-mlir/pull/888)
- [aten.embedding_dense_backward](https://github.com/llvm/torch-mlir/pull/889)
- [aten.native_batch_norm_backward](https://github.com/llvm/torch-mlir/pull/890)
- [native_dropout](https://github.com/llvm/torch-mlir/pull/891)
- [native_dropout_backward](https://github.com/llvm/torch-mlir/pull/892)

PR #570, opened 2022-02-09
[TORCH][MLIR] Add E2E support for `aten.native_layer_norm_backward` op.

gprateek93:
This PR adds two commits:

1. adds support for `aten.native_layer_norm_backward` operation. It also adds support for matching constant bools stored in a boolean list.
2. fixes `aten.native_layer_norm`. Previously this operation was not calculating correct shapes for mean and inverted STD. This has been corrected in this commit. Some new helper functions are added to calculate the inverted STD and to broadcast a given input with the help of a broadcast mask.

Signed-Off-By: Prateek Gupta <prateek@nod-labs.com>

PR #888, opened 2022-05-30
Add support for native_layer_norm_backward op

henrytwo:
This PR updates the codegen to allow for the `aten.native_layer_norm_backward` op in MLIR to be generated. The changes are based on PR https://github.com/llvm/torch-mlir/pull/570 by @gprateek93.

This op is used for tracing a full forward and backward graph for a HF BERT model using Lazy Tensor Core. We'd like to have this op in its current high-level form, rather than decomposing it.

e2e tests and type refinement functions have been excluded, since they aren't used by LTC. (Discussion: https://github.com/llvm/torch-mlir/issues/887)

cc: @antoniojkim @ke1337


powderluv:
nice. thank you.

ramiro050 (approved):
LGTM!

PR #889, opened 2022-05-30
Add support for embedding_dense_backward op

henrytwo:
This PR updates the codegen to allow for the `aten.embedding_dense_backward` op in MLIR to be generated.

This op is used for tracing a full forward and backward graph for a HF BERT model using Lazy Tensor Core.

e2e tests and type refinement functions have been excluded, since they aren't used by LTC. (Discussion: https://github.com/llvm/torch-mlir/issues/887)

cc: @antoniojkim @ke1337


ramiro050 (approved):
LGTM!

PR #893, opened 2022-06-01
Replace valsem.aten.zero with aten.zero.functional

henrytwo:
PyTorch now supports functional ops such as `aten.zero.functional`, so torch-mlir specific ops such as `valsem.aten.zero` can be replaced now.

cc: @antoniojkim @ke1337

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -2403,6 +2403,29 @@ def Torch_AtenThresholdBackwardOp : Torch_Op<"aten.threshold_backward", [
   }];
 }
 
+def Torch_AtenZeroFunctionalOp : Torch_Op<"aten.zero.functional", [
+    AllowsTypeRefinement,
+    HasValueSemantics,
+    ReadOnly
+  ]> {
+  let summary = "Generated op for `aten::zero.functional : (Tensor) -> (Tensor)`";
+  let arguments = (ins
+    AnyTorchTensorType:$self
+  );
+  let results = (outs
+    AnyTorchTensorType:$result
+  );
+  let hasCustomAssemblyFormat = 1;
+  let extraClassDefinition = [{
+    ParseResult AtenZeroFunctionalOp::parse(OpAsmParser &parser, OperationState &result) {
+      return parseDefaultTorchOp(parser, result, 1, 1);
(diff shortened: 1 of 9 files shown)


ramiro050:
Do you know if the support for functional ops applies to all underscore ops that don't have a functional variant? If that's the case, then we probably want to change things such that doing `emit_with_mutating_variants("aten::zero.functional : (Tensor) -> (Tensor)")` takes care of everything, similar to what is currently done for ops that have underscore and non-underscore variants

antoniojkim:
> Do you know if the support for functional ops applies to all underscore ops that don't have a functional variant?

It will eventually. But the functionalization pass has not been fully merged into PyTorch `master`. For now, there are a new select ops like `aten::zero_` that have been functionalized

henrytwo:
There shouldn't be any underscore variant for the functional ops, so I don't think we should be using `emit_with_mutating_variants`.

ramiro050:
> There shouldn't be any underscore variant for the functional ops, so I don't think we should be using `emit_with_mutating_variants`.

By underscore variant I meant what is currently happening with `aten.zero_`. `aten.zero_` is essentially the underscore variant of `aten.zero.functional`. When we know that such a pair exists, we can have torch-mlir automatically handle everything with a single `emit`

ramiro050 (approved):
Given that the functionalization pass in pytorch is still a work in progress, this seems like a good solution for now.

henrytwo:
> > There shouldn't be any underscore variant for the functional ops, so I don't think we should be using `emit_with_mutating_variants`.
> 
> By underscore variant I meant what is currently happening with `aten.zero_`. `aten.zero_` is essentially the underscore variant of `aten.zero.functional`. When we know that such a pair exists, we can have torch-mlir automatically handle everything with a single `emit`

Oh yeah I see what you mean now. 

@silvasean, after the functionalization pass is fully implemented in PyTorch, will we still be using the underscore variants in torch-mlir? If we are then I agree with switching over to `emit_with_mutating_variants` to generate the ops in pairs

silvasean:
As long as the ops exist, it makes sense to generate the ODS for them consistently with emit_with_mutating_variants.

The functionalization pass upstream is really an online dispatcher-level rewrite, so it does not help e.g. with importing existing scripted or traced models, or certain FX graphs that contain underscore variants. So we probably will still need some support as long as our TorchScript frontend is a thing (which will be a while).

PR #894, opened 2022-06-01
build: remove manual changes to ShapeLibrary.cpp

ashay:
The patch bumped up the LLVM tag made manual fixes to the code in
`ShapeLibrary.cpp`.  However, since that file is generated by the
`update_shape_lib.sh` script, its contents were reverted each time the
script was run.  This patch fixes the problem by removing the manual
changes to that file.


ramiro050 (approved):
LGTM! Thanks for the quick fix!

Issue #504, opened 2021-12-31
A prototype of jit.ignore support.


jinminxi104:
@ds1231h  and I are working on a prototype of jit.ignore support.
(This issue is similar to #462, and thanks for all helps in #462)
We would love any comments, and see if community is interesting in this prototype.

We are planning to convert jit.ignore function to a `std.call`. 
User can write the jit.ignore function in c/c++ manually, 
and link it with the program generated by torchMLIR

The use cases which we are considering are:
1.  allows user to write c/c++  manually  for TorchScript/TorchMLIR-incompatible python code
2. allows user to link with external library.

### Example

``` python
@torch.jit.ignore    
@detect_type       # detect_type is for getting the rank of foo. please read the design part.
def foo(self, in1, in2):    
    return some_torchscript_incompatible_python_code
    
@export              
@annotate_args([     
    None,            
    ([1024, 4], torch.float32, True),
    ([1024, 4], torch.float32, True),
])
def forward(self, arg1, arg2):
    some codes...
    xxx = self.foo(in1, in2)
    some codes...
    return ret
```

### Design
1.  jit.trace the forward and get the rank/dtype of `foo`.(the `detect_type` decorator in example code)
```python
# get args of forward
for arg in method._torch_mlir_arg_annotations:
  get args in `@annotate_args`

# trace forward
torch.jit.trace(forward, args)
 ```
use a decorator to get rank/dtype of `foo`
``` python
def detect_type(func):
     @wraps(func)
     def wrapper(*args, **kw):
         ret = func(*args, **kw)
         store_tracing_result(ret.dim(), ret.dtype)
         return ret
     return wrapper
```
2. import rank/dtype of `foo` via importer.
    After `importJitFunctionAsFuncOp` in `IValueImporter::importCompilationUnit`, we added rank&dtype to the `foo`
    Then, the IR looks like
    ```
    func @__torch__.IouOfModule.forward(%arg0: !torch.nn.Module<"__torch__.IouOfModule">, %arg1: !torch.tensor {torch.type_bound = !torch.vtensor <[1024,4],f32>}, %arg2: !torch.tensor {torch.type_bound = !torch.vtensor<[1024,4],f32>}) -> !torch.tensor {
        some IRs
        %1 = torch.extern "prim.PythonOp.foo"(%arg0, %0, %arg1) : (!torch.nn.Module<"__torch__.xxx">, !torch.tensor, !torch.tensor) -> !torch.tensor<[?,?],f32>
[...]

Issue #462, opened 2021-12-06
Brainstorm on how to add custom op authoring


ramiro050:
- [Brainstorm on how to add custom op authoring](#brainstorm-on-how-to-add-custom-op-authoring)
- [Adding ops corresponding to lower-level ops](#adding-ops-corresponding-to-lower-level-ops)
  - [Going form TorchScript IR to Torch MLIR](#going-form-torchscript-ir-to-torch-mlir)
- [Inserting MLIR at different points in the lowering process](#inserting-mlir-at-different-points-in-the-lowering-process)

## Brainstorm on how to add custom op authoring

PyTorch offers many ops for performing computations. However, sometimes users want to have more fine-grained control over definitions of custom computations for which the building blocks from PyTorch are not enough. The purpose of this issue is to brainstorm on potential solutions for the problem of giving users a way to author custom ops at the PyTorch level without having to resort to writing C++ in `torch-mlir` or in PyTorch.

## Adding ops corresponding to lower-level ops

One approach that is straight-forward to implement is to add at the `torch` dialect level extra ops corresponding to lower level ops that we want to expose to users. For example, one can expose the `linalg.matmul` op by creating an op `torch.linalg.matmul`. This `torch` dialect op would then have a pretty simple lowering to `linalg`, corresponding to replacing `torch.linalg.matmul` with `linalg.matmul`. The user would then be able to write something like:

```python
class MyModule(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        z = torch.tanh(x)
        t = linalg.init_tensor([3,3], torch.float)
        return linalg.matmul(z, z, t)
```

This would become the following `torch` MLIR:

```mlir
module {
  func private @__torch__.MyModule.forward(%arg0: !torch.nn.Module<"__torch__.MyModule">, %arg1: !torch.tensor {torch.type_bound = !torch.vtensor<[3,3],f32>}) -> !torch.tensor {
    %int6 = torch.constant.int 6
    %int3 = torch.constant.int 3
    %1 = torch.aten.tanh %arg1 : !torch.tensor -> !torch.tensor
    %2 = torch.prim.ListConstruct %int3, %int3 : (!torch.int, !torch.int) -> !torch.list<!torch.int>
    %3 = torch.linalg.init_tensor %2, %int6 : !torch.list<!torch.int>, !torch.int -> !torch.tensor
[...]

PR #895, opened 2022-06-01
Support pytorch extensions

rdadolf:
PyTorch allows new operators to be registered dynamically in modules. Torch-mlir already makes it fairly straightforward to add support for new operators, and this commit just extends that support to allow new PyTorch ops to come from an external module.

Previously, there was no hook to allow extensions to be loaded before querying the op registry, so extension ops would never be generated. Moreover, a bug caused ops imported from an extension to be impossible to support via shape inference.

This does *not* allow ops to be dynamically loaded into torch-mlir. Torch-mlir must still be compiled with support built-in.

This should make tasks like [this one](https://github.com/llvm/torch-mlir/issues/462#issuecomment-989036106) in #462 (and to a lesser extent #504) somewhat easier.

### DIFF
--- a/build_tools/update_shape_lib.sh
+++ b/build_tools/update_shape_lib.sh
@@ -1,13 +1,31 @@
 #!/bin/bash
 # Updates auto-generated shape library files for the `torch` dialect.
-set -e
+#
+# Environment variables:
+#   TORCH_MLIR_EXT_MODULES: comma-separated list of python module names
+#     which register custom PyTorch operators upon being imported.
+#   TORCH_MLIR_EXT_PYTHONPATH: colon-separated list of paths necessary
+#     for importing PyTorch extensions specified in TORCH_MLIR_EXT_MODULES.
+# For more information on supporting custom operators, see:
+#   ${TORCH_MLIR}/python/torch_mlir/_torch_mlir_custom_op_example/README.md
+
+set -eo pipefail
 
 src_dir="$(realpath $(dirname $0)/..)"
 build_dir="$(realpath "${TORCH_MLIR_BUILD_DIR:-$src_dir/build}")"
 torch_transforms_cpp_dir="${src_dir}/lib/Dialect/Torch/Transforms"
 python_packages_dir="${build_dir}/tools/torch-mlir/python_packages"
 
-#ninja -C "${build_dir}"
-PYTHONPATH="${python_packages_dir}/torch_mlir" python \
+pypath="${python_packages_dir}/torch_mlir"
+if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
+  pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
+fi
+ext_module="torch_mlir._torch_mlir_custom_op_example"
+if [ ! -z ${TORCH_MLIR_EXT_MODULES} ]; then
+  ext_module="${ext_module},${TORCH_MLIR_EXT_MODULES} "
+fi
+
+PYTHONPATH="${pypath}" python \
   -m torch_mlir.dialects.torch.importer.jit_ir.build_tools.shape_lib_gen \
+  --pytorch_op_extensions=${ext_module} \
   --torch_transforms_cpp_dir="${torch_transforms_cpp_dir}"
--- a/lib/Conversion/TorchToLinalg/CustomOpExample.cpp
+++ b/lib/Conversion/TorchToLinalg/CustomOpExample.cpp
@@ -1,0 +1,54 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+
+#include "torch-mlir/Conversion/TorchToLinalg/TorchToLinalg.h"
+
+#include "../PassDetail.h"
+#include "PopulatePatterns.h"
+#include "Utils.h"
+#include "mlir/IR/Matchers.h"
+#include "torch-mlir/Conversion/Utils/Utils.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchDialect.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
+#include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+
+using namespace mlir;
+using namespace mlir::torch;
+using namespace mlir::torch::Torch;
+
+namespace {
+class ConvertCustomOpExample
+  : public OpConversionPattern<TorchMlirCustomOpExampleIdentityOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult matchAndRewrite(TorchMlirCustomOpExampleIdentityOp op,
+                                OpAdaptor adaptor,
+                                ConversionPatternRewriter &rewriter
+                                ) const override {
+    // Type checks.
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+    // Since the example op does nothing, we simply replace the uses of the
+    // return value with its argument, then remove the op.
+    rewriter.replaceOp(op, op->getOperands());
+
+    return success();
+  }
+};
+} // namespace
+
+void mlir::torch::torch_to_linalg::populateCustomOpExamplePatternsAndLegality(
+    TypeConverter &typeConverter,
+    RewritePatternSet &patterns,
+    ConversionTarget &target) {
+  MLIRContext *context = patterns.getContext();
+  target.addIllegalOp<TorchMlirCustomOpExampleIdentityOp>();
+  patterns.add<ConvertCustomOpExample>(typeConverter, context);
+}
--- a/python/torch_mlir/_torch_mlir_custom_op_example/torch_mlir_custom_op_example.cpp
+++ b/python/torch_mlir/_torch_mlir_custom_op_example/torch_mlir_custom_op_example.cpp
@@ -1,0 +1,14 @@
+// For writing an extension like this one, see:
+// https://pytorch.org/tutorials/advanced/torch_script_custom_ops.html
+
+#include <torch/script.h> // One-stop header for PyTorch
+
+torch::Tensor identity(torch::Tensor t) {
+  // Do literally nothing.
+  return t;
+}
+
+TORCH_LIBRARY(_torch_mlir_custom_op_example, m) {
+  m.def("identity(Tensor t) -> Tensor");
+  m.impl("identity", &identity);
+}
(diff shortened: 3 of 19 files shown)


rdadolf:
@silvasean Since you originally wrote most of these pieces, I'd appreciated an opinion.

powderluv:
drive-by thank you for the PR :D

silvasean:
This looks awesome. Before committing this, I would like to somehow make it load-bearing for the project so it gets tested/etc. To do that, I think we should add a minimal custom op end-to-end, which I think would consist of:

1. build Python module defining custom torch op
2. add stuff in ods_gen and shape_lib_gen for that op
3. dtype refinement in RefineTypes
4. TorchToLinalg/CustomOpsExample.cpp

I think it could be as simple as a custom unary op -- something for folks to follow the bread crumbs when adding their own. I think the QPyTorch folks are going to want to follow this too.

rdadolf:
I wrote and tested all of that (in fact, it's literally written for a unary no-op---seems we had the same idea), but two things stopped me from including it in the PR:

1. I had no idea where to put it for testing purposes.
2. It means the test op will be exposed to the "real" world, which seemed unclean at the time.

If you folks don't mind, I don't mind.

silvasean:
> * I had no idea where to put it for testing purposes.

What form does the custom op take? Is it just a Python module? Or is there C++ code involved?

> * It means the test op will be exposed to the "real" world, which seemed unclean at the time.

Yeah, I'm sympathetic to this. However, PyTorch's op list is already pretty unclear and has all sorts of random stuff with varying levels of "core-ness" vs "custom-ness". I don't think that `torch_mlir_custom.nop` is that much more unclean than `aten._foo_cudnn_special_thing`

rdadolf:
Sounds good. I'll start adding the rest of the code into the PR.

> What form does the custom op take? Is it just a Python module? Or is there C++ code involved?

My example was a python wrapper around a C++ extension, similar to the [tutorial from the PyTorch docs](https://pytorch.org/tutorials/advanced/torch_script_custom_ops.html). It's small, but it does include a cmake file, requires compilation, etc.

The short version is:

Python side:
```python
import os
import torch

# Register custom.nop as a side-effect of importing this module.
current_dir = os.path.dirname(os.path.abspath(__file__))
lib = os.path.join(*[current_dir, os.pardir, os.pardir, 'custom_nop', 'libnop.so'])
torch.ops.load_library(lib)
```

C++ side:
```c++
#include <torch/script.h> // One-stop header.

#include <iostream>

torch::Tensor nop(torch::Tensor t) {
  std::cout << "no-op executed\n";
  return t;
}

TORCH_LIBRARY(custom, m) {
  m.def("nop(Tensor t) -> Tensor");
  m.impl("nop", &nop);
}
```

silvasean:
Ah, ok. Then I think this can be a standalone build inside build_tools. Perhaps `build_tools/custom_op_example`. And for clarity, let's make sure that the "namespace" of the op is always `torch_mlir_custom_op_example`, e.g. `torch_mlir_custom_op_example.nop`.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -1049,7 +1054,17 @@ def _verify_signature_matches_registry(f, registry: Registry):
     if signature != expected_signature:
         raise ValueError(f"Signature mismatch for {f.__name__!r}: expected {expected_signature!r}, got {signature!r}")
 
+def _maybe_import_op_extensions(args: argparse.Namespace):
+    extension_string = str.strip(args.pytorch_op_extensions)
+    if len(extension_string)>0:
```
silvasean, line 1059: nit: format

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -1049,7 +1054,17 @@ def _verify_signature_matches_registry(f, registry: Registry):
...
+        extension_names = extension_string.split(',')
```
silvasean, line 1060: nit: `"` for strings.

lib/Conversion/TorchToLinalg/CustomOpExample.cpp
```diff
@@ -0,0 +1,55 @@
...
+  LogicalResult matchAndRewrite(_TorchMlirCustomOpExampleNopOp op,
+                                OpAdaptor adaptor,
+                                ConversionPatternRewriter &rewriter
+                                ) const override {
+    // Type checks.
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+    // Since the example op does nothing, we simply replace the uses of the
+    // return value with its argument, then remove the op.
+    op->replaceAllUsesWith(op->getOperands());
```
silvasean, line 40: rewriter.replaceOp is more succinct.

python/torch_mlir/_torch_mlir_custom_op_example/torch_mlir_custom_op_example.cpp
```diff
@@ -0,0 +1,14 @@
+// For writing an extension like this one, see:
+// https://pytorch.org/tutorials/advanced/torch_script_custom_ops.html
+
+#include <torch/script.h> // One-stop header for PyTorch
+
+torch::Tensor nop(torch::Tensor t) {
```
silvasean, line 6: strictly speaking this is an `identity`, not a `nop` -- might make it a bit clearer.

build_tools/update_shape_lib.sh
```diff
@@ -1,13 +1,22 @@
 #!/bin/bash
 # Updates auto-generated shape library files for the `torch` dialect.
```
silvasean, line 2: maybe add a "usage:" in the comment at the top of this file covering the new env vars.

lib/Conversion/TorchToLinalg/CustomOpExample.cpp
rdadolf: Doesn't `replaceOp` require an op argument to replace with? The intent here was to remove the identity op altogether, which means that uses of its return value must instead use the argument to the identity op.

lib/Conversion/TorchToLinalg/CustomOpExample.cpp
```diff
@@ -0,0 +1,55 @@
...
+  }
+};
+} // namespace
+
+void mlir::torch::torch_to_linalg::populateCustomOpExamplePatternsAndLegality(
+    TypeConverter &typeConverter,
+    RewritePatternSet &patterns,
+    ConversionTarget &target) {
+  MLIRContext *context = patterns.getContext();
+  target.addIllegalOp<_TorchMlirCustomOpExampleNopOp>();
```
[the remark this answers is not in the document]
rdadolf: TIL about C++ reserved name formats.

We don't actually control the extension naming scheme. Obviously if someone's extending torch-mlir, they probably have some influence into the upstream extension package, but it still seems a bit wrong to have our code willingly generate illegal names under otherwise legal conditions.

I'll take an alternate view: what's the right thing to do if we *do* encounter a package with a leading underscore? Well, one solution is just to always generate legal code. I.e., when translating the Python names to C++, strip leading underscores. Technically, this collapses `_package_name` and `package_name` to `PackageName`, but... I think that's probably the right answer here, since the alternative is illegal C++ code.

The fix for that is a one-line change to `registry.py`. This has the added benefit of *always* preventing illegal names, which seems positive.

(I'll fix the string delimiters in `registry.py` that I overlooked in a bit. Currently debugging a crash that I can't reproduce locally.)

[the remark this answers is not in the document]
silvasean: What is it on the Torch side that requires the leading _ ? Is it the "namespace" of the extension? The .so file name?

To the degree that the leading _ is mandatory, then removing it should not result in collisions.

lib/Conversion/TorchToLinalg/CustomOpExample.cpp
silvasean: replaceOp takes a list of values (each corresponding to one result of the op being replaced). so you would pass in the operand value. You might be thinking of replaceOpWithNewOp

lib/Conversion/TorchToLinalg/CustomOpExample.cpp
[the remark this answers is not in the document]
rdadolf: > What is it on the Torch side that requires the leading _ ? Is it the "namespace" of the extension? The .so file name?

The extension package name, which ends up being concatenated with the op to give a unique op name. The .so file can be named whatever you wish, since it's entirely controlled by the extension and its build files.

Realistically, nothing about our name translation system is "mandatory", since it's all one-way. Torch-mlir choose how to change Python names into C++ names, and then those C++ names are used internally in torch-mlir. There's nothing that would prevent us from turning `aten.tanh` into `XXMLIRXXatenXXDOTXXtanhXXMLIRXX`, besides driving everyone batty reading it.

So if we think about the whole name translation system as a map, it might look like this:
```
Python Names -> C++ Names
-------------------------
package.op -> PackageOp
package_foo.op_foo -> PackageFooOpFoo
foo.bar.op -> FooBarOp
```

The case we're talking about is this:
```
_package.op -> PackageOp (proposed, currently illegal as _PackageOp)
package.op -> PackageOp
foo._bar.op -> Foo_BarOp
```

Both `_package.op` and `package.op` are technically legal Python names which can register op extensions, so under the current system, they can generate illegal C++ identifiers (without warning). I suppose we could alter the renaming scheme to change leading underscores into something else like `_package.op -> UnderscorePackageOp`, but this seems to violate the principle of least surprise.

At the end of the day, the problem is that this part of the legal Python namespace is a superset of the legal C++ names, so there's always going to be a bit of a rub. We kind of have to choose one of the three:

1. Allow for the possibility of illegal C++ names.
2. Allow for the possibility of namespace collisions.
3. Mangle the python names

My opinion was that 2. is the least bad/intrusive, but I'm happy to reconsider.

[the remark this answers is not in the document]
silvasean: These two quotes seem to be saying opposite things:

1.
> > What is it on the Torch side that requires the leading _ ? Is it the "namespace" of the extension? The .so file name?
> The extension package name, which ends up being concatenated with the op to give a unique op name. 

2.
> Both _package.op and package.op are technically legal Python names which can register op extensions


Is the leading `_` mandatory?

- If it is mandatory, then dropping the leading _ should never result in a collision
- If it is not mandatory, then why don't we just name our package without it?

Either way, I think that your option 2 makes sense.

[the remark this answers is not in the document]
rdadolf: Sorry, I'll try to be a more clear:

PyTorch extensions aren't necessarily written for torch-mlir, so I'm assuming we don't control the package naming scheme in the general case. The leading `_` is not mandatory for *our* example package and we could definitely remove it, but there's nothing that prevents anyone from developing a PyTorch extension that is named that way. If they do, we don't have much choice in what is fed to us. The name will be whatever the name is. So in that sense, we only have a choice about what to do with what's coming down the pipe.

Honestly, I don't have any kind of census about how often leading-underscore names appear in PyTorch op extensions (I suspect it is not common), so this could really all be unimportant either way.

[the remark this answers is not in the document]
rdadolf: Sounds good. Can always come back with another fix if it becomes a problem.

rdadolf:
All set for merge, I think. Happy to take any last comments or requests.

silvasean (approved):
Awesome, thanks Bob!!

Can you squash the history before merging to main?

rdadolf:
> Can you squash the history before merging to main?

Done. I don't have write access to the repo, so I'll need someone to hit the button. Thanks!

silvasean:
> > Can you squash the history before merging to main?
> 
> Done. I don't have write access to the repo, so I'll need someone to hit the button. Thanks!

Just invited you. Can't believe you didn't have it!

PR #898, opened 2022-06-03
[MLIR][TORCH] Add decomposition of `aten.numpy_T` op

vivekkhandelwal1:
This commit adds the decomposition of `aten.numpy_T` op into
`aten.t` or `aten.permute` op.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -1958,6 +1958,29 @@ class DecomposeAtenFloorDivideOp : public OpRewritePattern<AtenFloorDivideOp> {
 } // namespace
 
 namespace {
+// Decompose `aten.numpy_T` op into `aten.permute` op.
+class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenNumpyTOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value self = op.self();
+    int64_t inputRank = getTensorRank(self);
+
+    SmallVector<Value> dimListElements;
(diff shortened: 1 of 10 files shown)


vivekkhandelwal1:
@ramiro050, please review this PR when you find the time.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1941,6 +1941,34 @@ class DecomposeAtenBaddbmmOp : public OpRewritePattern<AtenBaddbmmOp> {
...
+// Decompose `aten.numpy_T` op into `aten.permute` or `aten.t` op.
+class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenNumpyTOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value self = op.self();
+    int64_t inputRank = getTensorRank(self);
+
+    if (inputRank > 2) {
```
ramiro050, line 1954: I think the code using `AtenPermuteOp` works even for input ranks $\leq 2$. If so, then we should just use that version to simplify the decomposition.

ramiro050 (changes requested):
Thanks for this! I just have one comment

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1957,6 +1957,29 @@ class DecomposeAtenFloorDivideOp : public OpRewritePattern<AtenFloorDivideOp> {
 };
 } // namespace
 
+namespace {
+// Decompose `aten.numpy_T` op into `aten.permute` or `aten.t` op.
```
ramiro050, line 1961: nit: comment needs to be updated

ramiro050 (approved):
LGTM! Just one small comment

PR #900, opened 2022-06-03
[MLIR][TORCH] Add E2E support for `aten.div.Tensor_mode` op

vivekkhandelwal1:
This commit adds lowering of `aten.div.Tensor_mode` op.
This commit also fixes formatting for the test file elementwise.py.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -447,11 +447,53 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     Type dtype = converter->convertType(div.getType())
                      .cast<RankedTensorType>()
                      .getElementType();
-    if (!dtype.isa<mlir::FloatType>())
+    if (!dtype.isa<mlir::FloatType>()) {
       div.emitError("unimplemented: non-floating point dtype");
+      return nullptr;
+    }
     Value lhs = convertScalarToDtype(b, loc, payloadArgs[0], dtype);
     Value rhs = convertScalarToDtype(b, loc, payloadArgs[1], dtype);
     return b.create<arith::DivFOp>(loc, lhs, rhs);
+  }
+  if (auto divTensorMode = dyn_cast<AtenDivTensorModeOp>(op)) {
+    AtenDivTensorModeOp::Adaptor adaptor(operands);
+    Type dtype = converter->convertType(divTensorMode.getType())
+                     .cast<RankedTensorType>()
+                     .getElementType();
+    if (!dtype.isa<mlir::FloatType>()) {
+      divTensorMode.emitError("unimplemented: non-floating point dtype");
+      return nullptr;
+    }
+    Value lhs = convertScalarToDtype(b, loc, payloadArgs[0], dtype);
+    Value rhs = convertScalarToDtype(b, loc, payloadArgs[1], dtype);
+    Value div = b.create<arith::DivFOp>(loc, lhs, rhs);
+
+    if (divTensorMode.rounding_mode().getType().isa<Torch::NoneType>())
+      return div;
+
+    std::string roundingMode;
+    if (!matchPattern(divTensorMode.rounding_mode(),
+                      m_TorchConstantStr(roundingMode))) {
+      divTensorMode.emitError("only support constant str rounding mode");
+      return nullptr;
+    }
+    if (roundingMode == "trunc") {
+      // "trunc" - rounds the results of the division towards zero. Equivalent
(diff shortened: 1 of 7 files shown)


python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -18,7 +18,9 @@
 
 # ==============================================================================
 
+
```
ramiro050, line 21: Thanks for the changes to make style consistent! However, in the future can you make these changes as their own PR? It makes it a bit hard to find new changes that are not stylistic

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -442,6 +442,41 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     Value rhs = convertScalarToDtype(b, loc, payloadArgs[1], dtype);
     return b.create<arith::DivFOp>(loc, lhs, rhs);
   }
+  if (auto divTensorMode = dyn_cast<AtenDivTensorModeOp>(op)) {
+    AtenDivTensorModeOp::Adaptor adaptor(operands);
+    Type dtype = converter->convertType(divTensorMode.getType())
+                     .cast<RankedTensorType>()
+                     .getElementType();
+    if (!dtype.isa<mlir::FloatType>())
+      divTensorMode.emitError("unimplemented: non-floating point dtype");
```
ramiro050, line 451: `emitError` does not exit out of the function, so every `emitError` must be followed by a `return nullptr` in this function. I noticed that `AtenDivTensorOp` above also does not return early. Can you fix that one too?

python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -949,28 +1078,34 @@ def forward(self, a):
...
     ])
-
     def forward(self, a):
         return torch.reciprocal(a)
 
 
 @register_test_case(module_factory=lambda: ElementwiseReciprocalIntModule())
 def ElementwiseReciprocalIntModule_basic(module, tu: TestUtils):
-    module.forward(torch.randint(1, 10, (4,), dtype=torch.int32))
+    module.forward(torch.randint(1, 10, (4, ), dtype=torch.int32))
```
ramiro050, line 1101: nit: for one element tuples in python, the convention I've always encountered is the one without the space

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -442,6 +442,41 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    Value lhs = convertScalarToDtype(b, loc, payloadArgs[0], dtype);
+    Value rhs = convertScalarToDtype(b, loc, payloadArgs[1], dtype);
+    Value div = b.create<arith::DivFOp>(loc, lhs, rhs);
+
+    if (divTensorMode.rounding_mode().getType().isa<Torch::NoneType>())
+      return div;
+    else {
```
ramiro050, line 458: No need for the `else` since the previous `if` returns.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -442,6 +442,41 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+        Value cstZero = b.create<arith::ConstantOp>(loc, b.getZeroAttr(dtype));
+        Value pred = b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::ULT,
+                                             div, cstZero);
+        return b.create<arith::SelectOp>(loc, pred, ceil, floor);
+      } else if (roundingMode == "floor") {
+        // "floor" - rounds the results of the division down. Equivalent to
+        // floor division in Python (the // operator)
+        return b.create<math::FloorOp>(loc, div);
+      } else
+        divTensorMode.emitError("invalid rounding mode");
```
ramiro050, line 477: nit: if an `if` branch has braces, all branches must have braces for consistency/readability (https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements). Also, make sure to `return nullptr`.

ramiro050 (changes requested):
Looks mostly good, I just have a few small comments

python/torch_mlir_e2e_test/test_suite/elementwise.py
vivekkhandelwal1: Sure, I will do that.

ramiro050 (approved):
LGTM!

PR #906, opened 2022-06-06
Add Support for aten::scatter_add

antoniojkim:
This PR adds support for generating the `aten.scatter_add` op in MLIR.

Required to support HuggingFace Bert using the LTC backend.

CC: @henrytwo @ke1337


silvasean (approved):
FYI I would encourage you to fold a lot of these op adding PR's into a single commit -- there isn't much value to doing them in separate commits and it adds more review overhead.

PR #912, opened 2022-06-07
E2E HuggingFace Bert using LTC Backend

antoniojkim:
We finally have the full training graph for HuggingFace's implementation of Bert working using the LTC Backend!

The `torch_mlir_ltc_backend` staging branch still won't be ready to merge into `main` after this PR though. We are still waiting on a couple things in upstream PyTorch to be merged in:
- https://github.com/pytorch/pytorch/issues/77445
- https://github.com/pytorch/pytorch/pull/75527

Plus, there are a few more minor changes I'd like to make before we merge it in that will help make it easier to maintain.

CC: @ke1337 @henrytwo


henrytwo:
Note that for this PR, we are executing on the JIT graph (which we confirmed matches exec on CPU) and **not** the emitted MLIR. We're operating under the assumption that torch-mlir is doing its job in emitting MLIR that accurately represents the JIT graph.

In a future PR, we will lower the MLIR to linalg and execute on a reference backend to validate that the MLIR itself is able to produce the correct numerics.

silvasean:
Awesome!!! :partying_face: 

Is there anything in particular you would like me to look at?

PR #914, opened 2022-06-07
Refactor Node Lowering

antoniojkim:
A follow up to https://github.com/pytorch/pytorch/pull/78273 which refactors and provides a longer term, more maintainable way to support lowering for non-native ops.

CC: @ke1337 @henrytwo


antoniojkim:
@silvasean Gentle reminder to please review

PR #915, opened 2022-06-07
Generate underscore variant of functional ops

henrytwo:
As discussed [here](https://github.com/llvm/torch-mlir/pull/893#issuecomment-1146938643), we should be using `emit_with_mutating_variants` to generate both underscore and non-underscore variants for functional ops.

The ODS generator had to be modified to exclude the `functional` suffix during the generation of underscore ops.

Additionally, this PR fixes some formatting issues for the `aten::arange.start_out` and `aten::scatter_add` ops in `GeneratedTorchOps.td`.

cc: @antoniojkim @ke1337 @silvasean

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -2395,6 +2395,51 @@ def Torch_AtenUnsqueeze_Op : Torch_Op<"aten.unsqueeze_", [
   }];
 }
 
+def Torch_AtenZeroFunctionalOp : Torch_Op<"aten.zero.functional", [
+    AllowsTypeRefinement,
+    HasValueSemantics,
+    ReadOnly
+  ]> {
+  let summary = "Generated op for `aten::zero.functional : (Tensor) -> (Tensor)`";
+  let arguments = (ins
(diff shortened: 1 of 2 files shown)


ramiro050:
I think the CI is failing because this class also needs to be updated

https://github.com/llvm/torch-mlir/blob/e1b38e74dd0072f9d6d48f7c90e21912fe3e85cd/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp#L186

Since `aten.zero_` now has the `IsTrailingUnderscoreInplaceVariant` trait, the pattern above is replacing it with `aten.zero`, leading to the error.

henrytwo:
What changes would you recommend I make to that class? Do we _want_ `IsTrailingUnderscoreInplaceVariant` to be applied to `aten.zero_`?

ramiro050:
> What changes would you recommend I make to that class? Do we want to make `IsTrailingUnderscoreInplaceVariant` not apply to the underscore variants of functional ops?

What I had in mind was having a special trait like `hasFunctionalVariant` that applies to underscore ops with functional variants, then we can have a pattern replace the underscore with the functional one in a similar way to what the `ReduceTrailingUnderscoreInplaceVariant` pattern does. But I don't know if this will be too involved.

@silvasean do you have any thoughts on this?

PR #918, opened 2022-06-08
Use double quotes instead of single quotes

henrytwo:
Oops missed some single quotes from the review comment in https://github.com/llvm/torch-mlir/pull/915, so fixing it here


silvasean (approved):
Thanks!

PR #919, opened 2022-06-08
Handle `nn.Linear(..., bias=False)` case for TorchToLinalg

makslevental:
Handles `bias=False` case by checking `bias.getType().isa<Torch::NoneType>()` and initing with zero in this case.


makslevental:
@ramiro050 I'm not sure if I should add a FileCheck test since this doesn't change anything at the IR level.

silvasean (approved):

makslevental:
thanks @silvasean for the quick approval. Anything else you want here (assuming the tests pass, which they should).

PR #925, opened 2022-06-10
Refactor autogen

antoniojkim:
Refactoring the autogen to make organizing and maintaining shape inference functions easier.


antoniojkim:
@silvasean tagging you as an FYI. We did our own internal reviews before this PR was opened so will just be merging this in

PR #934, opened 2022-06-13
[Draft] Implemented __contains__int_list

JakopinA:


include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
```diff
@@ -5742,7 +5740,7 @@ def Torch_AtenScatterAddOp : Torch_Op<"aten.scatter_add", [
     HasValueSemantics,
     ReadOnly
   ]> {
-  let summary = "Generated op for `aten::scatter_add : (Tensor self, int dim, Tensor index, Tensor src) -> (Tensor)`";
+  let summary = "Generated op for `aten::scatter_add : (Tensor, int, Tensor, Tensor) -> (Tensor)`";
```
vivekkhandelwal1, line 5743: You should rebase your local branch over the latest upstream torch-mlir, and then re-run the script `build_tools/update_torch_ods.sh`, these changes should be gone.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -492,6 +493,7 @@ def emit_with_mutating_variants(key, **kwargs):
     emit("aten::Int.Scalar : (Scalar) -> (int)", has_folder=True)
 
     # Primitive ops
+
```
vivekkhandelwal1, line 496: Remove this new line.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -85,6 +85,25 @@ def MmDagModule_basic(module, tu: TestUtils):
...
+    ])
+    def forward(self,):
+        return 3 in [1,2,3]
+
+
+@register_test_case(module_factory=lambda: ContainsIntList())
+def ContainsIntList_basic(module, tu: TestUtils):
+    module.forward()
+
+
```
vivekkhandelwal1, line 106: Did you try using the `torch.ops.aten.__contains__` for this test? Also, you should add a test for the case which evaluates to `false`.

lib/Conversion/TorchToStd/TorchToStd.cpp
```diff
@@ -50,6 +50,20 @@ class ConvertAtenDimOp : public OpConversionPattern<AtenDimOp> {
...
+  LogicalResult
+  matchAndRewrite(RuntimeAssertOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    rewriter.replaceOpWithNewOp<cf::AssertOp>(op, adaptor.condition(),
+                                              adaptor.message());
+    return success();
+  }
+};
+} // namespace
+
```
vivekkhandelwal1, line 66: This code is completely wrong. You have just copy-pasted it from the `RuntimeAssertOp`. You need to modify this to work for the op you are working on. Also add a conversion pattern like this https://github.com/llvm/torch-mlir/blob/8ff03257df6c3a5841034b98bd2feb9e928df71e/lib/Conversion/TorchToStd/TorchToStd.cpp#L318-L319 for the `__contains__.int_list` op.

vivekkhandelwal1 (changes requested):

python/torch_mlir_e2e_test/test_suite/basic.py
JakopinA: Maksim explained to me that the torch.ops.aten__contains function effectively replaces "return 3 in [1,2,3]". If it doesn't, I can replace the code.

lib/Conversion/TorchToStd/TorchToStd.cpp
silvasean: We shouldn't need TorchTo* lowering for this op -- we don't support lists at runtime currently.

vivekkhandelwal1: Then adding the canonicalization/folding pattern in `TorchOps.cpp` would suffice?

python/torch_mlir_e2e_test/test_suite/basic.py
ramiro050: That's probably the case, but for these tests we prefer the `torch.ops.aten...` version because it makes it very clear what is being tested

lib/Conversion/TorchToStd/TorchToStd.cpp
silvasean: Yes.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -129,6 +129,32 @@ def MmDagModule_basic(module, tu: TestUtils):
...
+        ([-1], torch.float32, True)
+    ])
+    def forward(self,key):
+        
+        return torch.ops.aten.__contains__(self.l, key)
+
+
+@register_test_case(module_factory=lambda: ContainsIntList())
+def ContainsIntList_True(module, tu: TestUtils):
+    module.forward(3)
```
ramiro050, line 151: This is not right. The inputs should all be tensors. This is an assumption made by the `annotate_args` above

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -129,6 +129,32 @@ def MmDagModule_basic(module, tu: TestUtils):
...
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1], torch.float32, True)
+    ])
+    def forward(self,key):
+        
+        return torch.ops.aten.__contains__(self.l, key)
```
ramiro050, line 146: The reason the test is failing is that the value that gets generated for `key` is not a `torch.constant.int`. Therefore, when `key` is compared to each of the values in `self.l`, because `self.l` is composed of `torch.constant.int`s, the comparison always fails, resulting in the op not getting folded.

To make sure `key` is a `torch.constant.int` you have to hardcode the value. So do something like `torch.ops.aten.__contains__([1, 2, 3], 3)`

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1512,6 +1521,27 @@ OpFoldResult Aten__Getitem__DictStrOp::fold(ArrayRef<Attribute> operands) {
...
+
+  auto targetKey = item();
+  for (auto l : arrayConstruct.elements()) {
+    if (l == targetKey)
+      return getI1IntegerAttr(getContext(), true);
+  }
+  for (auto l : arrayConstruct.elements()) {
+    if (typeid(l) != typeid(torch.constant.int))
+      return getI1IntegerAttr(getContext(), false);
+  }
```
ramiro050, line 1541: This does not work for the `false` case. An element in the list not being a constant int does not mean that the `targetKey` is not in the list.

Because checking that something is not in the list requires us to determine that `targetKey` is *truly* not equal to any element in the list, we can only achieve this when we have enough information about the list elements. For this particular op, the information we need is that we know the int value of all elements in the list and `targetKey` at compile time. So here you first need to convert the list of `torch.constant.int`s to C++ integers as well as `targetKey`, then you can just loop through the list and check that `targetKey` is not equal to any of the elements. There are helper functions to get the integers. If you do `git grep 'm_TorchConstantInt'` you should find examples for getting a `torch.constant.int` and a list of `torch.constant.int`s

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1512,6 +1521,27 @@ OpFoldResult Aten__Getitem__DictStrOp::fold(ArrayRef<Attribute> operands) {
...
+//===----------------------------------------------------------------------===//
+// Aten__Contains__IntListOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult Aten__Contains__IntListOp::fold(ArrayRef<Attribute> operands) {
+  auto arrayConstruct = getListConstructIfNotModified(l());
+  if (!arrayConstruct)
+    return nullptr;
+
+  auto targetKey = item();
```
ramiro050, line 1533: nit: the convention in this repo is to name variables after their argument name in pytorch. So this should be named `item`

ramiro050 (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1494,6 +1494,15 @@ static PrimDictConstructOp getDictConstructIfNotModified(Value torchDict) {
   return torchDict.getDefiningOp<Torch::PrimDictConstructOp>();
 }
 
+static PrimListConstructOp getListConstructIfNotModified(Value torchList) {
```
ramiro050, line 1497: I think it would make things more clear to call this `static bool isListConstructNotModified(Value torchList)`, since the return value is not used by the fold below

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1512,6 +1521,37 @@ OpFoldResult Aten__Getitem__DictStrOp::fold(ArrayRef<Attribute> operands) {
...
+
+  if (!matchPattern(itemConstruct, m_TorchConstantInt(&item)))
+  {
+    return nullptr;
+  }
+
+  if (!matchPattern(l(), m_TorchConstantIntList(list)))
+  {
+    return nullptr;
+  }
```
ramiro050, line 1545: nit: braces are not needed

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1512,6 +1521,37 @@ OpFoldResult Aten__Getitem__DictStrOp::fold(ArrayRef<Attribute> operands) {
...
+
```
ramiro050, line 1546: nit: extra empty line

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1512,6 +1521,37 @@ OpFoldResult Aten__Getitem__DictStrOp::fold(ArrayRef<Attribute> operands) {
   return nullptr;
 }
 
+//===----------------------------------------------------------------------===//
+// Aten__Contains__IntListOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult Aten__Contains__IntListOp::fold(ArrayRef<Attribute> operands) {
+  auto arrayConstruct = getListConstructIfNotModified(l());
+  auto itemConstruct = item();
```
ramiro050, line 1530: nit: the convention in torch-mlir is to name variables corresponding to arguments of an op after the argument name, so this should be `Value item = item();`

ramiro050 (changes requested):
Looking good! I just have a few small comments below

lib/Dialect/Torch/IR/TorchOps.cpp
JakopinA: What do you mean by this? I already have a variable named item, should I rename it to something else?

Issue #911, opened 2022-06-07
Eliminating dynamic shaped tensors for the static inputs


erman-gurses:
This issue is related to https://github.com/llvm/torch-mlir/pull/935

For the torch IR of miniLM model, we observed dynamic tensors like below even if we provided static input as `torch.randint(2, (1, 128))`

```
    %186 = torch.vtensor.literal(dense<0> : tensor<si64>) : !torch.vtensor<[],si64>
     .
     .
     .
    %190 = torch.prim.NumToTensor.Scalar %int128 : !torch.int -> !torch.vtensor<[],si64>
    %191 = torch.aten.add.Tensor %190, %186, %int1 : !torch.vtensor<[],si64>, !torch.vtensor<[],si64>, !torch.int -> !torch.vtensor<[],si64>
    %192 = torch.aten.Int.Tensor %191 : !torch.vtensor<[],si64> -> !torch.int
    %193 = torch.prim.ListConstruct %int1, %192 : (!torch.int, !torch.int) -> !torch.list<int>
    %194 = torch.aten.ones %193, %int6, %none, %cpu, %false : !torch.list<int>, !torch.int, !torch.none, !torch.Device, !torch.bool -> !torch.vtensor<[1,?],f32>
```
Dynamic shape of %194 makes many of the `torch.aten.add.Tensor` operations also dynamic in the next part of the IR. So solving this code snipped as static might also can solve other dynamic shaped tensor issues. 

I am working on to find a canonical way eliminate those dynamic shaped tensors.

nirvedhmeshram:
So the current plan is to add canonicalization for add.tensor and friends that can handle 0d operands, these operand can be coming either from a  `torch.prim.NumToTensor.Scalar` op or as a 0d `torch.vtensor.literal` below are two example before and after canonicalization
Example 1
```
%0 = torch.prim.NumToTensor.Scalar %int0 : !torch.int -> !torch.vtensor<[],si64>
%1 = torch.prim.NumToTensor.Scalar %int2 : !torch.int -> !torch.vtensor<[],si64>
%2 = torch.aten.add.Tensor %0, %1, %int3 : !torch.vtensor<[],si64>, !torch.vtensor<[],si64>, !torch.int -> !torch.vtensor<[],si64>
```
after canonicalization
``` 
%0 =  aten.add.int(%int0, %int2) : !torch.int, !torch.int -> !torch.int
[...]

PR #935, opened 2022-06-13
Add canonicalization for aten.add.tensor op

erman-gurses:


erman-gurses:
@nirvedhmeshram Yes, I will do that, I am still missing handling this part for now below. Working on it.
%0 = torch.vtensor.literal(dense<0> : tensor<si64>) : !torch.vtensor<[],si64>

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -763,6 +763,43 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+
+void AtenAddTensorOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                                  MLIRContext *context) {
+  patterns.add(+[](AtenAddTensorOp op, PatternRewriter &rewriter) {
+    // The lhs and rhs of the add.tensor op should be 0d tensors for the
+    // canonicalization to be carried out.
+    // `aten.add.tensor(self, other, alpha)` is canonicalized to
+    // `aten.add.int(self, aten.mul.int(other, alpha))`.
+
+    ValueTensorLiteralOp lhsScalarValueTensor = op.self().getDefiningOp<ValueTensorLiteralOp>();
```
silvasean, line 778: This is too specific. We should have a common helper to get the scalar Value (or null) from a tensor Value.

silvasean (changes requested):

vivekkhandelwal1:
The issue that is being addressed with this PR: https://github.com/llvm/torch-mlir/issues/911.

lib/Dialect/Torch/IR/TorchOps.cpp
erman-gurses: @silvasean Sure, I will handle that after I solve the issue regarding extraction value from ElementsAttr.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -98,6 +99,23 @@ static FloatAttr getF64FloatAttr(MLIRContext *context, double value) {
   return FloatAttr::get(Float64Type::get(context), value);
 }
 
+static llvm::Optional<Value> getScalarValue(AtenAddTensorOp op, PatternRewriter &rewriter){
```
pashu123, line 102: This is very specific to AtenAddTensorOp, Can't it be made generic to every op.

pashu123: Or just pass ValueTensorLiteral or PrimNumToTensorOp and extract scalar for it.

test/Dialect/Torch/canonicalize.mlir
```diff
@@ -1267,3 +1267,21 @@ func.func @torch.aten.Bool.int$fold_cst() -> !torch.bool {
   %1 = torch.aten.Bool.int %int : !torch.int -> !torch.bool
   return %1 : !torch.bool
 }
+
+// CHECK-LABEL:   func.func @torch.aten.add.Tensor$canonicalize_0d() -> !torch.vtensor<[],si64> {
```
pashu123, line 1271: Add the test case with ValueTensorLiteralOp.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -763,6 +781,35 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+    llvm::Optional<Value> lhs = getScalarValue(op, rewriter);
+
+    if (!lhs.hasValue())
+      return failure();
+
+    auto rhsScalar = op.other().getDefiningOp<PrimNumToTensorScalarOp>();
+    if (!rhsScalar)
+      return failure();
+    Value rhs = rhsScalar.a();
+    Value mul = rewriter.create<AtenMulIntOp>(op->getLoc(), rhs, op.alpha());
```
silvasean, line 805: do you have to check that the datatypes are actually ints?

erman-gurses: @silvasean, I added type checking for `mul` and `add` variables

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -763,6 +787,58 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+      rhs = getScalarValue(rhsPrimNumToTensorScalarOp, op.getLoc(), rewriter);
+    } else
+      return failure();
+
+    if (!lhs.hasValue() || !rhs.hasValue())
+      return failure();
+
+    Value mul =
+        rewriter.create<AtenMulIntOp>(op->getLoc(), rhs.getValue(), op.alpha());
+    if (!mul.getType().isa<Torch::IntType>())
```
pashu123, line 830: You don't have to check the output type of mul instead the lhs and rhs value passed to mul.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -763,6 +787,58 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+      return failure();
+    Value add =
+        rewriter.create<AtenAddIntOp>(op->getLoc(), lhs.getValue(), mul);
+    if (!add.getType().isa<Torch::IntType>())
```
pashu123, line 834: Same here.

erman-gurses: Sure. I added type checking `lhs` and `rhs` instead of `add` and `mul`

lib/Dialect/Torch/IR/TorchOps.cpp
erman-gurses: Update instead of `mul` and `add` type checking, I did type checking for `lhs` and `rhs`

erman-gurses:
This PR is related to issue https://github.com/llvm/torch-mlir/issues/911

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -98,6 +99,29 @@ static FloatAttr getF64FloatAttr(MLIRContext *context, double value) {
...
 
+static llvm::Optional<Value> getScalarValue(Operation *base_op, Location loc,
+                                            PatternRewriter &rewriter) {
+  llvm::Optional<Value> scalar;
+  if (ValueTensorLiteralOp valueTensorLiteralOp =
+          dyn_cast<ValueTensorLiteralOp>(base_op)) {
+    if (valueTensorLiteralOp &&
+        getTensorRank(valueTensorLiteralOp.getResult()) == 0) {
+      auto val = valueTensorLiteralOp.value()
+                     .cast<ElementsAttr>()
```
ramiro050, line 110: is this cast necessary?

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -98,6 +99,29 @@ static FloatAttr getF64FloatAttr(MLIRContext *context, double value) {
...
+                     .cast<DenseElementsAttr>()
+                     .getSplatValue<int64_t>();
```
ramiro050, line 112: The tensor might not be of type `int`. You should check the type before getting the splat value

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -763,6 +787,58 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+    Operation *rhsValueTensorLiteralOp =
+        op.other().getDefiningOp<ValueTensorLiteralOp>();
+    Operation *rhsPrimNumToTensorScalarOp =
+        op.other().getDefiningOp<PrimNumToTensorScalarOp>();
+    if (rhsValueTensorLiteralOp)
+      rhs = getScalarValue(rhsValueTensorLiteralOp, op.getLoc(), rewriter);
+    else if (rhsPrimNumToTensorScalarOp) {
+      rhs = getScalarValue(rhsPrimNumToTensorScalarOp, op.getLoc(), rewriter);
+    } else
+      return failure();
```
ramiro050, line 823: nit: This is a bit repetitive. I think it would be better to have `getScalarValue` take a `Value input` as the argument rather than an `Operation *`. Then you can replace this with:

```cpp
Value lhs = getScalarValue(op.self(), op.getLoc(), rewriter);
Value rhs = getScalarValue(op.other(), op.getLoc(), rewriter);

if (!lhs || !rhs)
  return failure();
```

Note that there is no need to return an `llvm::Optional`, since `Value` can be `null`.

Then in `getScalarValue` you would just do:

```cpp
if (auto valueTensorLiteralOp = input.getDefiningOp<ValueTensorLiteralOp>()) {
} else if (auto primNumToTensorScalarOp = input.getDefiningOp<PrimNumToTensorScalarOp>()) {
}
```

test/Dialect/Torch/canonicalize.mlir
```diff
@@ -1267,3 +1267,37 @@ func.func @torch.aten.Bool.int$fold_cst() -> !torch.bool {
...
+    return %2 : !torch.vtensor<[],si64>
+}
+
+// CHECK-LABEL:   func.func @torch.aten.add_1.Tensor$canonicalize_0d() -> !torch.vtensor<[],si64> {
+// CHECK:      %[[INT6:.*]] = torch.constant.int 6
+// CHECK:      %[[INT2:.*]] = torch.constant.int 2
+// CHECK:      %[[PR1:.*]] = torch.prim.NumToTensor.Scalar %[[INT2]] : !torch.int -> !torch.vtensor<[],si64>
+// CHECK:      %[[PR2:.*]] = torch.prim.NumToTensor.Scalar %[[INT6]] : !torch.int -> !torch.vtensor<[],si64>
+// CHECK:      return %[[PR2]] : !torch.vtensor<[],si64>
+func.func @torch.aten.add_1.Tensor$canonicalize_0d() -> !torch.vtensor<[],si64> {
```
ramiro050, line 1295: nit: can you name this function `@torch.aten.add.Tensor$canonicalize_0d_vtensor_literal` (note: i've removed the `_1` after `add`)

ramiro050 (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
erman-gurses: Looks like `.cast<ElementsAttr>()` is not necessary. I removed it.

test/Dialect/Torch/canonicalize.mlir
erman-gurses: I see, there is another function for testing as `@torch.aten.add.Tensor$canonicalize_0d_vtensor_literal`. That is the reason I added _1.  Do you want me name them as `@torch.aten.add_1.Tensor$canonicalize_0d_vtensor_literal` and `@torch.aten.add_2.Tensor$canonicalize_0d_vtensor_literal`

lib/Dialect/Torch/IR/TorchOps.cpp
erman-gurses: It would be great, If I get any hint for this. I tried multiple things but not a success.

lib/Dialect/Torch/IR/TorchOps.cpp
erman-gurses: I refactored based on the directions in here

test/Dialect/Torch/canonicalize.mlir
erman-gurses: @ramiro050 I changed their names please take a look

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -763,6 +788,36 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+    // The lhs and rhs of the add.tensor op should be 0d tensors for the
+    // canonicalization to be carried out.
+    // `aten.add.tensor(self, other, alpha)` is canonicalized to
+    // `aten.add.int(self, aten.mul.int(other, alpha))`.
+
+    Value lhs = getScalarValue(op.self(), op.getLoc(), rewriter);
+    Value rhs = getScalarValue(op.other(), op.getLoc(), rewriter);
+
+    if (!lhs || !rhs)
+      return failure();
```
ramiro050, line 807: nit: can you use `rewriter.notifyMatchFailure` with an error message to signal failure? Also for the `failure()` below

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -98,6 +98,31 @@ static FloatAttr getF64FloatAttr(MLIRContext *context, double value) {
...
+        getTensorRank(valueTensorLiteralOp.getResult()) == 0) {
+      auto tensorType =
+          valueTensorLiteralOp.value().getType().cast<RankedTensorType>();
+      if (tensorType.getElementType().isa<mlir::IntegerType>()) {
+        auto val = valueTensorLiteralOp.value()
+                       .cast<DenseElementsAttr>()
+                       .getSplatValue<int64_t>();
+        scalar = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(val));
+        return scalar;
```
ramiro050, line 115: nit: the `return scalar` here and in the `else` branch are not needed because there is a `return scalar` at the very end of this function

ramiro050 (approved):
I just have a couple of small comments, but other than that it LGTM!

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -763,6 +788,36 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+
+void AtenAddTensorOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                                  MLIRContext *context) {
+  patterns.add(+[](AtenAddTensorOp op, PatternRewriter &rewriter) {
+    // The lhs and rhs of the add.tensor op should be 0d tensors for the
+    // canonicalization to be carried out.
+    // `aten.add.tensor(self, other, alpha)` is canonicalized to
+    // `aten.add.int(self, aten.mul.int(other, alpha))`.
+
+    Value lhs = getScalarValue(op.self(), op.getLoc(), rewriter);
```
silvasean, line 803: you should check `if (!lhs)` here before the next call -- this avoids creating IR when a match fails.

erman-gurses: @silvasean I update that part of the code and just want to make sure this is the you wanted.

silvasean: Perfect :)

PR #936, opened 2022-06-14
Add CustomOpExample.cpp to bazel target TorchMLIRTorchToLinalg

fortianyou:
fix undefined symbol:
_ZN4mlir5torch15torch_to_linalg42populateCustomOpExamplePatternsAndLegalityERNS_13TypeConverterERNS_17RewritePatternSetERNS_16ConversionTargetE

### DIFF
--- a/utils/bazel/torch-mlir-overlay/BUILD.bazel
+++ b/utils/bazel/torch-mlir-overlay/BUILD.bazel
@@ -344,6 +344,7 @@ cc_library(
 cc_library(
     name = "TorchMLIRTorchToLinalg",
     srcs = [
+        "lib/Conversion/TorchToLinalg/CustomOpExample.cpp",
         "lib/Conversion/TorchToLinalg/DataMovement.cpp",
         "lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp",
         "lib/Conversion/TorchToLinalg/Linear.cpp",


rdadolf:
Hey there. Thanks for taking these fixes on. I know nearly nothing about the bazel build side, but I'll try to flesh out the highlights of the PR that caused the failures.

The file you caught should be the important one on the transform side, but there's also a demo vehicle which is intended to exercise that path. This is built as part of the `torch_mlir` Python package, and will be loaded when it's imported. It's also used in the end-to-end tests, so if you attempt to run those with a bazel build, it will fail.

`/python/torch_mlir/_torch_mlir_custom_op_example` is the directory in question. The C++ file in that directory needs to be built, and both the resulting SO file and the `__init__.py` file in that directory must be put into a `_torch_mlir_custom_op_example` subdirectory of wherever the `torch_mlir` Python package is installed (`$BUILD/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/`, for instance).

I'm not up to speed on the use case for Bazel, nor how it's being used, but I suspect it installs `torch_mlir` somewhere, and if so, the custom op parts should probably get handled.

Feel free to ask questions, and thanks again for being on top of this.

fortianyou:
> Hey there. Thanks for taking these fixes on. I know nearly nothing about the bazel build side, but I'll try to flesh out the highlights of the PR that caused the failures.
> 
> The file you caught should be the important one on the transform side, but there's also a demo vehicle which is intended to exercise that path. This is built as part of the `torch_mlir` Python package, and will be loaded when it's imported. It's also used in the end-to-end tests, so if you attempt to run those with a bazel build, it will fail.
> 
> `/python/torch_mlir/_torch_mlir_custom_op_example` is the directory in question. The C++ file in that directory needs to be built, and both the resulting SO file and the `__init__.py` file in that directory must be put into a `_torch_mlir_custom_op_example` subdirectory of wherever the `torch_mlir` Python package is installed (`$BUILD/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/`, for instance).
> 
> I'm not up to speed on the use case for Bazel, nor how it's being used, but I suspect it installs `torch_mlir` somewhere, and if so, the custom op parts should probably get handled.
> 
> Feel free to ask questions, and thanks again for being on top of this.

Thanks for replying to this. I think the problem is `populateCustomOpExamplePatternsAndLegality` is used in `lib/Conversion/TorchToLinalg/TorchToLinalg.cpp`. But it's defined in `lib/Conversion/TorchToLinalg/CustomOpExample.cpp`.

rdadolf:
Custom op support has temporarily been removed while a MacOS issue is sorted (#942). It should return, but this PR will be blocked until it does (which is okay, since the thing it targets is no longer breaking bazel).

PR #938, opened 2022-06-14
[MLIR][TORCH] Add support for dtype=torch.bool for `aten.to.dtype` op

vivekkhandelwal1:

### DIFF
--- a/lib/Conversion/Utils/Utils.cpp
+++ b/lib/Conversion/Utils/Utils.cpp
@@ -242,13 +242,30 @@ Value convertScalarToDtype(OpBuilder &b, Location loc, Value scalar,
     return false;
   };
 
-  if (isByteOrChar(scalarType) || isByteOrChar(dtype) ||
-      dtype.isSignlessInteger(1)) {
+  if (isByteOrChar(scalarType) || isByteOrChar(dtype)) {
     // TODO: Handle to-boolean conversion(from-boolean conversion is handled).
     mlir::emitError(loc)
         << "unsupported byte, char or bool type for convertScalarToDtype "
         << scalarType << "(scalar type) -> " << dtype << "(dtype)";
     return nullptr;
+  }
+
+  // If the dtype is i1, i.e., a boolean type.
+  if (dtype.isSignlessInteger(1)) {
+    Type scalarType = scalar.getType();
+    Value cstZero = b.create<arith::ConstantOp>(loc, b.getZeroAttr(scalarType));
+    if (scalarType.isa<mlir::FloatType>()) {
+      return b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UNE, scalar,
+                                     cstZero);
+    } else if (scalarType.isa<mlir::IntegerType>()) {
+      return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ne, scalar,
+                                     cstZero);
+    } else {
+      mlir::emitError(loc)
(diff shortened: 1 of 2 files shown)


ramiro050:
> @ramiro050, what could be the reason behind the Eager Mode tests failing in CI?

In my machine, only the test you added fails in eager mode. For me, it crashes here https://github.com/llvm/torch-mlir/blob/8fd084377dbd9d4b17681256c75db019db81af51/lib/Conversion/TorchToLinalg/DataMovement.cpp#L1080. I think it is because that function does not support `bool` types.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -764,7 +764,21 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+      // If the result dtype is i1, i.e., a boolean type.
+      Type inputElemType = input.getType();
+      Value cstZero =
+          b.create<arith::ConstantOp>(loc, b.getZeroAttr(inputElemType));
+      if (inputElemType.isa<mlir::FloatType>())
+        result = b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UNE, input,
+                                         cstZero);
+      else
+        result = b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ne, input,
+                                         cstZero);
```
ramiro050, line 778: can you make this part of `convertScalarToDtype` instead?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -764,7 +764,21 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    Value result;
+    if (dtype.isa<mlir::IntegerType>() && dtype.isInteger(1)) {
+      // If the result dtype is i1, i.e., a boolean type.
+      Type inputElemType = input.getType();
+      Value cstZero =
+          b.create<arith::ConstantOp>(loc, b.getZeroAttr(inputElemType));
+      if (inputElemType.isa<mlir::FloatType>())
+        result = b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UNE, input,
+                                         cstZero);
+      else
```
ramiro050, line 776: this should be a series if `else if` followed by a branch that results in an error in the end for unsupported input element types

ramiro050 (changes requested):

lib/Conversion/Utils/Utils.cpp
```diff
@@ -242,15 +242,32 @@ Value convertScalarToDtype(OpBuilder &b, Location loc, Value scalar,
...
+                                     cstZero);
+    else if (scalarType.isa<mlir::IntegerType>())
+      return b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ne, scalar,
+                                     cstZero);
+    else {
+      mlir::emitError(loc)
+          << "unsupported scalar type for convertScalarToDtype " << scalarType
+          << "(scalar type) -> " << dtype << "(dtype)";
+      return nullptr;
+    }
```
ramiro050, line 268: nit: If one branch of `if` has braces, they all need to have braces

https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements

ramiro050 (approved):
LGTM. Just one small style comment below. Also, can you update the PR/commit title to something along the lines of "Add support for bool type in `convertScalarToDtyle`"

PR #939, opened 2022-06-14
Only import the LTC backend that's used in example models

henrytwo:
Fixing namespace resolution issue from importing both TS and MLIR libraries into example models. Now we only import the library that we're going to use, based on the user's selection of LTC backend.

cc: @ke1337 @antoniojkim @silvasean 

This PR has already been reviewed internally within Cerebras, so we will be merging immediately.

### DIFF
--- a/examples/ltc_backend_bert.py
+++ b/examples/ltc_backend_bert.py
@@ -17,11 +17,9 @@ import argparse
 import sys
 from typing import List
 
-import ltc_backend.ltc_backend._EXAMPLE_MLIR_BACKEND as ltc_backend
 import torch
 import torch._C
 import torch._lazy
-import torch._lazy.ts_backend
 from datasets import load_dataset
 from datasets.dataset_dict import DatasetDict
 from torch.utils.data import DataLoader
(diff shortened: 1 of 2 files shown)


stellaraccident:
Thanks for the patch! Minor nit on development style in llvm projects: since this is a community project, we try to not rely on offline review/communication within a company and keep reviews out in the open. For a trivial/nfc (non functional change), it is ok to land directly without review. If there is any question, you will find that in the main repo, people will explicitly ask for a "post submit review" from particular people when doing this (as opposed to saying it was reviewed off thread in private).

Not trying to slow you down -- just providing some wording guidance for the future so this project aligns more with the overall development policy.

In this case, the patch LGTM.

antoniojkim:
I think @silvasean was okay with us doing this for now because this is a staging branch. We're very close to being in a place where we can merge in the staging branch into `main`. Once we're working off of `main` instead of the `torch_mlir_ltc_backend` staging branch, we'll for sure make sure to get proper reviews from the community

silvasean:
If it is a large burden to move more of the review onto github I think we can be pragmatic and let's talk about solutions. I know it is "just" public perception, but having PR's like this that feel "dumped" upstream without a community development process (not saying that was your intention -- just that it can be perceived that way) really has a negative impact on the project and attracting contributors.

PR #943, opened 2022-06-15
Change custom op tests to xfail.

rdadolf:
MacOS builds are currently breaking on the custom op library, so it has
been temporarily disabled. This updates the end-to-end tests to xfail
the associated tests.


silvasean:
Hi Bob, this is getting to be a pretty ugly set of workarounds -- do we have an ETA / work plan for the fix? I'm hesitant to adding these workarounds. It seems preferable to disable the Mac builder altogether for now.

@powderluv - thoughts?

rdadolf:
Retracting the PR in place of backing out the custom op support altogether.

PR #944, opened 2022-06-15
Temporarily revert support for custom op extensions.

rdadolf:
The MacOS builders are having linking trouble with the extension library.
Until it's fixed, all support for op extensions is disabled. It should be
easy to restore once the issue is resolved.


rdadolf:
@silvasean @powderluv This is passing on my dev machine, just waiting for the CI tests.

If those pass, I'll merge and wait for the MacOS builders to trigger. Assuming those pass, we should be good.

I've confirmed that the `update_*.sh` scripts run correctly under this reversion, so there shouldn't be any time bombs AFIAK.

silvasean:
Thanks!

PR #949, opened 2022-06-16
emitted aten.norm.ScalarOpt_dim

albertdmath:


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1625,6 +1625,41 @@ class ConvertTensorStaticInfoCastOp
...
 
+namespace {
+class ConvertAtenNormScalarOptDimOp
+    : public OpConversionPattern<AtenNormScalarOptDimOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenNormScalarOptDimOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    //Not sure if this file is where I should place the lowering for this op.
```
ramiro050, line 1636: Yeah, this file is fine, since there are a few other norm ops in here

ramiro050 (changes requested):
A few initial comments

vivekkhandelwal1 (changes requested):

ramiro050:
Can you rebase your branch to resolve the conflicts so that the CI can run?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -953,6 +953,11 @@ def aten〇native_layer_norm(input: List[int], normalized_shape: List[int], weig
     for i in range(num_unreduced_dimensions, len(input)):
         reduction_shape.append(1)
     return input, reduction_shape, reduction_shape
+def aten〇norm〇ScalarOpt_dim(self: List[int], p: Optional[float], dim: List[int], keepdim: bool = False) -> List[int]:
+    # I was told I needed to modify this file to specify how the shape of the result is calculated for this op.
+    # return upstream_shape_functions.unary(input)?
+    # My main question is what is the result tensor size?
```
ramiro050, line 959: Here you can probably use the `upstream_shape_functions.mean_dim` function

https://github.com/llvm/torch-mlir/blob/845e794aabe9c65e33272d8e99ba2180711b801b/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py#L520

python/torch_mlir_e2e_test/test_suite/norm_like.py
```diff
@@ -332,3 +332,24 @@ def forward(self, x):
...
 def LayerNormNormalizeOverAllDimsModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(2, 2, 3))
+
+# ==============================================================================
+
+class LayerNormScalarOptDimModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        #self.ly = torch.norm([2, 2, 3], p = 3)
+        #self.ly.eval()
```
ramiro050, line 342: Please remove commented out code

python/torch_mlir_e2e_test/test_suite/norm_like.py
```diff
@@ -332,3 +332,24 @@ def forward(self, x):
...
+
+    @export
+    @annotate_args([
+        None,
+        ([1, 1, 1, 1], torch.float32, True),
```
ramiro050, line 347: Can you make these dynamic? This allows a more comprehensive test

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2101,6 +2101,32 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
...
+    Location loc = op.getLoc();
+    Value self = op.self();
+    Value p = op.p();
+    Value dim = op.dim();
+    Value keepdim = op.keepdim();
+    Value result  = op.result();
+    BaseTensorType tensorType = self.getType().cast<BaseTensorType>();
+    Value abs = rewriter.create<AtenAbsOp>(loc, tensorType, self);
+    Value exp = rewriter.create<AtenExpOp>(loc, tensorType, p);
+    Value sum = rewriter.create<AtenSumOp>(loc, tensorType, result);
```
ramiro050, line 2121: This is not right. `AtenSumOp` reduces a tensor to a zero-rank tensor, so the return type in general will not be equal to `tensorType`. You need to create a new type with the right shape.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2101,6 +2101,32 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
...
+    : public OpRewritePattern<AtenNormScalarOptDimOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenNormScalarOptDimOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value self = op.self();
+    Value p = op.p();
+    Value dim = op.dim();
+    Value keepdim = op.keepdim();
```
ramiro050, line 2116: These two are currently not used in the decomposition. You should be able to incorporate them by using the `AtenSumDimIntListOp` instead of `AtenSumOp`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2101,6 +2101,32 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
 };
 } // namespace
 
+// Decompose aten.norm.ScalarOpt_dim
```
ramiro050, line 2104: nit: this comment is not necessary

ramiro050 (changes requested):
A few comments below

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2101,6 +2101,29 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
...
+  LogicalResult matchAndRewrite(AtenNormScalarOptDimOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value self = op.self();
+    Value p = op.p();
+    Value result  = op.result();
+    BaseTensorType tensorType = self.getType().cast<BaseTensorType>();
+    Value abs = rewriter.create<AtenAbsOp>(loc, tensorType, self);
+    Value exp = rewriter.create<AtenExpOp>(loc, tensorType, p);
+    Value sum = rewriter.create<AtenSumDimIntListOp>(loc, tensorType, result);
```
ramiro050, line 2118: > This is not right. `AtenSumOp` reduces a tensor to a zero-rank tensor, so the return type in general will not be equal to `tensorType`. You need to create a new type with the right shape.

The comment above was not addressed. After the `sum` reduction, the result type will change. There are some helpers that might be useful (although they might need to be modified to support reduction along multiple dimensions):

https://github.com/llvm/torch-mlir/blob/ab7af5d8ce5e904c1e12fe4b88c48f6d4249cdc5/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L40

https://github.com/llvm/torch-mlir/blob/ab7af5d8ce5e904c1e12fe4b88c48f6d4249cdc5/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L75

ramiro050, line 2118: These three variables are not used anywhere

ramiro050, line 2118: `AtenSumDimIntListOp` takes more arguments. You can see examples by doing `git grep AtenSumDimIntListOp`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2101,6 +2101,29 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
...
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenNormScalarOptDimOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value self = op.self();
+    Value p = op.p();
+    Value result  = op.result();
+    BaseTensorType tensorType = self.getType().cast<BaseTensorType>();
+    Value abs = rewriter.create<AtenAbsOp>(loc, tensorType, self);
+    Value exp = rewriter.create<AtenExpOp>(loc, tensorType, p);
```
ramiro050, line 2117: `ExpOp` is the exponential function $e^x$, I think you want `PowOp`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2101,6 +2101,29 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
...
+    Value sum = rewriter.create<AtenSumDimIntListOp>(loc, tensorType, result);
+    Value div = rewriter.create<AtenDivScalarOp>(loc, tensorType, p, p);
```
ramiro050, line 2119: You can create a `1` by using `Torch::ConstantIntOp`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2101,6 +2101,29 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
 };
 } // namespace
 
+namespace {
+class DecomposeAtenNormScalarOptDimOp
+    : public OpRewritePattern<AtenNormScalarOptDimOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenNormScalarOptDimOp op,
```
ramiro050, line 2109: > These two are currently not used in the decomposition. You should be able to incorporate them by using the `AtenSumDimIntListOp` instead of `AtenSumOp`

This comment I had left on the lines with `op.dim()` and `op.keepdim()` was not addressed.

python/torch_mlir_e2e_test/test_suite/norm_like.py
```diff
@@ -332,3 +332,23 @@ def forward(self, x):
...
+        ([-1, -1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, x):
+        #return torch.ops.aten.norm.ScalarOpt_dim(x, p = 2, dim = 2, keepdim = True)
+        return torch.ops.aten.norm.ScalarOpt_dim(x, p = 2, dim = 3)
+
+
+@register_test_case(module_factory=lambda: LayerNormScalarOptDimModule())
+def LayerNormScalarOptDimModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(1, 1, 1, 1))
```
ramiro050, line 354: can you make a tensor with more than one element? Otherwise the summation aspect of this op is not really getting tested

ramiro050 (changes requested):
There were two comments I had previously made that were not addressed (I have copied them again below). Please address all comments before requesting another review. If my comments are unclear, ask questions.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
albertdmath: It seems like there is no support for AtenPowOp.

ramiro050:
Can you add the `div.int` op in a separate PR? This makes the PRs more self-contained and easier to review

ramiro050:
Hey @albertdmath, you can also run the lit tests locally by doing `cmake --build build --target check-torch-mlir-all`.

albertdmath:
The error I've been getting is that when I use floats the error calls for ints and when I use ints it calls for floats.

lib/Conversion/TorchToArith/TorchToArith.cpp
```diff
@@ -374,6 +374,9 @@ class ConvertTorchToArith : public ConvertTorchToArithBase<ConvertTorchToArith>
     target.addIllegalOp<AtenDivFloatOp>();
     patterns.add<ConvertAtenBinaryOp<AtenDivFloatOp, arith::DivFOp>>(
         typeConverter, context);
+    target.addIllegalOp<AtenDivIntOp>();
+    patterns.add<ConvertAtenBinaryOp<AtenDivIntOp, arith::DivFOp>>(
```
ramiro050, line 378: I don't think this is right. The `arith::DivFOp` takes floats as arguments, while the `AtenDivIntOp` takes ints as arguments. I would instead of doing this do a decomposition of `AtenDivIntOp` into `AtenDivFloatOp`, inserting casts to turn the inputs into floats.

test/Conversion/TorchToArith/basic.mlir
```diff
@@ -180,6 +180,19 @@ func.func @torch.aten.div.float(%arg0: !torch.float, %arg1: !torch.float) -> !to
...
 
+// CHECK-LABEL:  func.func @torch.aten.div.int(
+// CHECK-SAME:                            %[[LHS:.*]]: !torch.float,
+// CHECK-SAME:                            %[[RHS:.*]]: !torch.float) -> !torch.float {
+// CHECK:          %[[LHS_F64:.*]] = torch_c.to_f64 %[[LHS]]
+// CHECK:          %[[RHS_F64:.*]] = torch_c.to_f64 %[[RHS]]
+// CHECK:          %[[SUB:.*]] = arith.divf %[[LHS_F64:.*]], [[RHS_F64:.*]] : f64
+// CHECK:          %[[OUT:.*]] = torch_c.from_f64 %[[SUB:.*]]
+// CHECK:          return %[[OUT:.*]] : !torch.float
+func.func @torch.aten.div.int(%arg0: !torch.int, %arg1: !torch.float) -> !torch.float {
```
ramiro050, line 191: If you decompose the op into `AtenDivFloat` there is no need to have this test. However, you should add a corresponding test in the decompose-complex-ops.mlir file.

ramiro050 (changes requested):
Also, can you create the lowering for `AtenDivInt` in a separate PR, since it is not really related to the goal of this PR.

albertdmath:
Just to confirm, the cast would happen with AtenToDtypeOp?

albertdmath:
I'm also assuming this new approach would mean reverting the changes I made to TorchOps.cpp.

ramiro050:
> Just to confirm, the cast would happen with AtenToDtypeOp?

Can you create the lowering for `AtenDivInt` in a separate PR, since it is not really related to the goal of this PR.

`AtenToDtypeOp` takes a tensor as an input, so that would not work with your op. You should be able to use `AtenFloatScalar` to cast the scalars to float

albertdmath:
Alright, I've created a separate PR for aten.div.int, please let me know if it looks good. Since this op depends on aten.div.int, I first want to get aten.div.int merged in, then take out the work for aten.div.int from this branch and get this merged in.

Vremold:
What's going on with this PR?

PR #951, opened 2022-06-17
[LINALG] Lower `aten.Matmul` to `linalg.BatchMatmul`

Shukla-Gaurav:
This commit lowers `aten.matmul` to `linalg.BatchMatmul` under the
following conditions:
1. The result of matrix multiplication must have batch dimensions,
   i.e., rank greater than 2.
2. The resultant matrix must have at most 1 dynamic batch dimension.

It also handles broadcasting of batch dimensions when batch dimensions
of the matrices are broadcastable.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -164,12 +164,16 @@ public:
 
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
       return failure();
+    auto lhsType = lhs.getType().cast<RankedTensorType>();
+    auto rhsType = rhs.getType().cast<RankedTensorType>();
 
-    unsigned lhsRank = lhs.getType().cast<RankedTensorType>().getRank();
-    unsigned rhsRank = rhs.getType().cast<RankedTensorType>().getRank();
+    // Get the rank of both matrix.
+    unsigned lhsRank = lhsType.getRank();
+    unsigned rhsRank = rhsType.getRank();
 
     Type newResultType = getTypeConverter()->convertType(op.getType());
-    Type elementType = newResultType.cast<TensorType>().getElementType();
+    auto resultType = newResultType.cast<RankedTensorType>();
+    Type elementType = resultType.getElementType();
 
     // The different cases of torch_matmul op is mentioned here:
     // https://pytorch.org/docs/stable/generated/torch.matmul.html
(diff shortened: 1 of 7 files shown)


Shukla-Gaurav:
@anupgangwar @sjarus The CI is failing on TOSA tests, Could you please help me here?

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -228,49 +232,151 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
...
+        collapsedResultShape.push_back(rhsDim1);
+        SmallVector<OpFoldResult> updatedCollapseResultShape =
+            getAsOpFoldResult(collapsedResultShape);
+
+        Value initTensor = rewriter.create<linalg::InitTensorOp>(
+            loc, updatedCollapseResultShape, elementType);
+        Value c0 = rewriter.create<arith::ConstantOp>(
+            loc, rewriter.getZeroAttr(elementType));
+        Value zeroTensor =
+            rewriter.create<linalg::FillOp>(loc, c0, initTensor).getResult(0);
```
ramiro050, line 344: nit: you can use `createZeroInitTensor`

python/torch_mlir_e2e_test/test_suite/matmul.py
```diff
@@ -130,3 +130,62 @@ def Matmul_4d(module, tu: TestUtils):
...
+        ([4, -1, -1, -1], torch.float32, True),
+        ([-1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, lhs, rhs):
+        return torch.matmul(lhs, rhs)
+
+
+@register_test_case(module_factory=lambda: MatmulBroadcastBatchDim())
+def MatmulBroadcastBatchDim_basic(module, tu: TestUtils):
+    module.forward(tu.rand(4, 5, 6, 7), tu.rand(5, 7, 6))
```
ramiro050, line 190: nit: can you make one of the `5`s in this test a `1` to check that the size-1 broadcasting also works as expected?

ramiro050 (approved):
Thanks for this! This is really useful. I just have a few small comments below

ramiro050:
> @anupgangwar @sjarus The CI is failing on TOSA tests, Could you please help me here?

In my experience, these types of errors occur when a single test fails at a runtime assert. For some reason it causes a cascade of failures in the CI, but if you run things locally, you should be able to find the one test that is causing the issues

ramiro050:
> I am unable to reproduce it locally. All the test cases passed locally:
> 
> ```
>     Passed: 139
>     Expectedly Failed: 446
> ```
> 
> fyi @ramiro050 @sjarus

Interesting. I get the same result on my machine. However, I did notice that the test `MatmulSingleDynamicBatchDim_basic` causes an assertion error:

```
python: /usr/local/google/home/ramiroleal/torch-mlir/externals/llvm-project/mlir/include/mlir/IR/StorageUniquerSupport.h:140: static ConcreteT mlir::detail::StorageUserBase<mlir::RankedTensorType, mlir::TensorType, mlir::detail::RankedTensorTypeStorage, mlir::detail::TypeUniquer, Trait, Trait>::get(mlir::MLIRContext *, Args...) [ConcreteT = mlir::RankedTensorType, BaseT = mlir::TensorType, StorageT = mlir::detail::RankedTensorTypeStorage, UniquerT = mlir::detail::TypeUniquer, Traits = <Trait, Trait>, Args = <llvm::ArrayRef<long>, mlir::Type, mlir::Attribute>]: Assertion `succeeded(ConcreteT::verify(getDefaultDiagnosticEmitFn(ctx), args...))' failed.
```

I have a feeling the CI error is related to that.

ramiro050:
Yeah, if you add the test `MatmulSingleDynamicBatchDim_basic` to the TOSA pass set, you get

```
python: /usr/local/google/home/ramiroleal/torch-mlir/externals/llvm-project/mlir/include/mlir/IR/StorageUniquerSupport.h:140: static ConcreteT mlir::detail::StorageUserBase<mlir::RankedTensorType, mlir::TensorType, mlir::detail::RankedTensorTypeStorage, mlir::detail::TypeUniquer, Trait, Trait>::get(mlir::MLIRContext *, Args...) [ConcreteT = mlir::RankedTensorType, BaseT = mlir::TensorType, StorageT = mlir::detail::RankedTensorTypeStorage, UniquerT = mlir::detail::TypeUniquer, Traits = <Trait, Trait>, Args = <llvm::ArrayRef<long>, mlir::Type, mlir::Attribute>]: Assertion `succeeded(ConcreteT::verify(getDefaultDiagnosticEmitFn(ctx), args...))' failed.
FAIL - "MatmulSingleDynamicBatchDim_basic"

Unexpected outcome summary:

****** Failed tests - 1 tests
    FAIL - "MatmulSingleDynamicBatchDim_basic"
        Runtime error: Testing process terminated. Either the compiler crashed or the compiled code crashed at runtime.
```

sjarus:
I'd expect this to fail due to dynamic shapes but without reproducing I don't know where - my guess is TOSA to LinAlg downstream . Can the test be suppressed and then I can turn it on and investigate ?

ramiro050:
> I'd expect this to fail due to dynamic shapes but without reproducing I don't know where - my guess is TOSA to LinAlg downstream . Can the test be suppressed and then I can turn it on and investigate ?

I ran things on my machine using the TOSA backend, and the issue seems to be here

https://github.com/llvm/torch-mlir/blob/ccc858f5311fdcdb7828d691a9db3bd565ac52f5/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L1136

Because the batch dims get multiplied together, the `commonValue` ends up being equal to `-4`. This is then used here 

https://github.com/llvm/torch-mlir/blob/ccc858f5311fdcdb7828d691a9db3bd565ac52f5/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L1200-L1204

causing the runtime error because the type cannot be generated with a size of `-4`. A quick fix for now would be to add after `commonValue` is set, something along the lines of:

```cpp
      // TODO: Fix the case where there are multiple dynamic dims in the batch dims.
      if (commonValue < -1)
        return failure();
```

This should lead to a compilation error rather than an assertion error

lib/Conversion/TorchToLinalg/Linear.cpp
Shukla-Gaurav: I did not use this because here the sizes is a SmallVector<OpFoldResult> which does not fit there.

ramiro050:
> @ramiro050 Thanks for the fix :). Isn't it good to handle this way
> 
> ```
> if(hasDynamicDims)
>         commonValue = kDynamicSize;
> ```
> 
> ?

Yeah, I think so. I did not really understand why when there was a dynamic dim, the sizes were being multiplied. @sjarus maybe you have more context on this 

https://github.com/llvm/torch-mlir/blob/ccc858f5311fdcdb7828d691a9db3bd565ac52f5/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L1134-L1136

sjarus:
> > @ramiro050 Thanks for the fix :). Isn't it good to handle this way
> > ```
> > if(hasDynamicDims)
> >         commonValue = kDynamicSize;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ?
> 
> Yeah, I think so. I did not really understand why when there was a dynamic dim, the sizes were being multiplied. @sjarus maybe you have more context on this
> 
> https://github.com/llvm/torch-mlir/blob/ccc858f5311fdcdb7828d691a9db3bd565ac52f5/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L1134-L1136

Yes I'd say that setting it to KDynamicDim if any dimension part of common dims is dynamic is the right thing to do. Somewhere down the line implementing that logic for the combination of common and non-common broadcasting + dynamic shapes my brain melted...

sjarus:
> > > @ramiro050 Thanks for the fix :). Isn't it good to handle this way
> > > ```
> > > if(hasDynamicDims)
> > >         commonValue = kDynamicSize;
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > ?
> > 
> > 
> > Yeah, I think so. I did not really understand why when there was a dynamic dim, the sizes were being multiplied. @sjarus maybe you have more context on this
> > https://github.com/llvm/torch-mlir/blob/ccc858f5311fdcdb7828d691a9db3bd565ac52f5/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L1134-L1136
> 
> Yes I'd say that setting it to KDynamicDim if any dimension part of common dims is dynamic is the right thing to do. Somewhere down the line implementing that logic for the combination of common and non-common broadcasting + dynamic shapes my brain melted...

Thinking further, it's unclear if the presence of dynamic shapes allows us to reason about the common vs non common shapes in any manner, so the conservative approach is to both the common and non-common parts kDynamicDim if at least one dim or either tensor has a dynamic dim. I didn't really spend much time in this direction, instead validating several permutations of interesting n-D static shaped forms. Handling dynamic dims this way would require a small overhaul of this code but it shouldn't be too terrible.

silvasean:
The "how to broadcast the leading dimensions in the presence of dynamic shapes (including size-1 diemsnions)" should be entirely identical to our handling of elementwise ops, including all the error handling and other cases. I suspect it can be implemented similarly to torch_to_linalg::createElementwiseLinalgGeneric

https://github.com/llvm/torch-mlir/blob/8b9475930399fb70f2cd929a26a47650b6ce0542/lib/Conversion/TorchToLinalg/Utils.cpp#L153

There may be a way to make a lower-level helper that createElementwiseLinalgGeneric calls that is shared with this logic.

Shukla-Gaurav:
@ramiro050 All comments have been addressed. Let me know if everything looks fine and it's ready to merge.
@sjarus I have made some changes to `torch-to-tosa` conversion, please approve if those changes looks fine.
Thanks!

PR #954, opened 2022-06-18
Include Bazel BUILD steps to development.md

sjain-stanford:
Make the bazel build a little more visible to potential users.

cc: @asaadaldien

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,19 +49,6 @@ pip install --pre torch-mlir torchvision -f https://github.com/llvm/torch-mlir/r
 # This will install the corresponding torch and torchvision nightlies
 ```
 
-## Build with Bazel
-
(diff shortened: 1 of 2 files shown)


jpienaar:
Thanks for highlighting this, OOC why add this to the readme rather than development.md alongside the cmake instructions?

sjain-stanford:
> why add this to the readme rather than development.md alongside the cmake instructions?

Thanks for the pointer. I've moved the build instructions to `development.md`. I think the rest of the instructions there are closely tied to the CMake build (e.g. testing, virtualenv setup). I'll send some follow-up PRs to include LIT tests + E2E integration tests using bazel. At that point it may be better to re-factor this file with specific steps (for CMake vs Bazel). @jpienaar WDYT?

silvasean:
This is awesome!

@asaadaldien for review + merging

asaadaldien (commented):
LGTM!

PR #956, opened 2022-06-20
Add decomposition for at::_convolution

gpetters94:
Here's a simple decomposition of `at::_convolution` to `at::convolution`. I'm not sure how much of the flags we want to support.


ramiro050 (approved):
Sorry for the late review, this PR got lost in my notifications. The current support of flags seems fine to me. From the unit tests, the extra flags don't seem to affect the behavior of the op in a way we can measure, so they are probably just for internal use.

PR #958, opened 2022-06-21
Bump LLVM

makslevental:
Bufferization dialect added a new useful op ([`bufferization.alloc_tensor`](https://mlir.llvm.org/docs/Dialects/BufferizationOps/#bufferizationalloc_tensor-mlirbufferizationalloctensorop)) so how about a bump?


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -237,8 +237,8 @@ LogicalResult ClassTypeOp::verify() {
 // PrimLoopOp
 //===----------------------------------------------------------------------===//
 
-OperandRange PrimLoopOp::getSuccessorEntryOperands(unsigned index) {
-  assert(index == 0);
+OperandRange PrimLoopOp::getSuccessorEntryOperands(::llvm::Optional<unsigned int> index) {
+  assert(index.hasValue() && index.value() == 0);
```
makslevental, line 241: cc @silvasean this was the only substantive patch.

makslevental:
hmm gotta adjust some of the filecheck tests...

lib/Dialect/Torch/IR/TorchOps.cpp
silvasean: I would omit the ::llvm but seems fine to me.

test/Dialect/Torch/canonicalize.mlir
```diff
@@ -3,8 +3,8 @@
 // CHECK-LABEL:   func.func @torch.aten.__range_length$fold() -> (!torch.int, !torch.int, !torch.int, !torch.int) {
 // CHECK:           %[[INT1:.*]] = torch.constant.int 1
 // CHECK:           %[[INT2:.*]] = torch.constant.int 2
-// CHECK:           %[[INTM1:.*]] = torch.constant.int -1
 // CHECK:           %[[INT3:.*]] = torch.constant.int 3
+// CHECK:           %[[INTM1:.*]] = torch.constant.int -1
```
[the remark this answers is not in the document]
silvasean: This happens from time to time as the pattern rewriters change under the hood

silvasean (approved):
Thanks for the bump!

PR #960, opened 2022-06-22
Add E2E support for aten.__contains__.int_list

JakopinA:


gpetters94 (approved):
Looks good!

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1529,6 +1538,32 @@ OpFoldResult Aten__Contains__StrOp::fold(ArrayRef<Attribute> operands) {
...
+  auto arrayConstruct = isListConstructNotModified(l());
+  auto itemConstruct = item();
+  if (!arrayConstruct)
+    return nullptr;
+
+  int64_t item;
+  SmallVector<int64_t> list;
+
+  if (!matchPattern(itemConstruct, m_TorchConstantInt(&item)))
+    return nullptr;
```
Shukla-Gaurav, line 1555: You can directly use `item()` here.

Shukla-Gaurav (approved):
Thanks, LGTM! Few minor comments. Please squash your commits before merge.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1494,6 +1494,15 @@ static PrimDictConstructOp getDictConstructIfNotModified(Value torchDict) {
   return torchDict.getDefiningOp<Torch::PrimDictConstructOp>();
 }
 
+static PrimListConstructOp isListConstructNotModified(Value torchList) {
```
ramiro050, line 1497: This function should return a `bool` given the name of the function

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1529,6 +1538,32 @@ OpFoldResult Aten__Contains__StrOp::fold(ArrayRef<Attribute> operands) {
...
 
+//===----------------------------------------------------------------------===//
+// Aten__Contains__IntListOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult Aten__Contains__IntListOp::fold(ArrayRef<Attribute> operands) {
+  auto arrayConstruct = isListConstructNotModified(l());
+  auto itemConstruct = item();
+  if (!arrayConstruct)
+    return nullptr;
```
ramiro050, line 1549: Once the return value of `isListConstructNotModified` is changed to `bool`, this should become

```cpp
if (!isListConstructNotModified(l()))
  return nullptr;
```

ramiro050 (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
ramiro050, line 1497: Also, this function should be moved down to the `Aten__Contains__IntListOp` section, since it is not related to the `PrimTupleUnpackOp`

ramiro050 (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1494,6 +1494,15 @@ static PrimDictConstructOp getDictConstructIfNotModified(Value torchDict) {
   return torchDict.getDefiningOp<Torch::PrimDictConstructOp>();
 }
 
+static PrimListConstructOp isListConstructNotModified(Value torchList) {
+  if (!llvm::all_of(torchList.getUsers(), [](Operation *op) {
+        return isa<Aten__Contains__IntListOp>(op);
+      }))
+    return nullptr;
```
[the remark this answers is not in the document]
ramiro050: The idea is to make sure that there is not an op between the moment the list value is created and the moment the `__contains__` op is called that overwrites the list in place. Checking that the only user is the `__contains__` op is a pretty conservative way of ensuring no mutation has taken place since the list argument was created. However, because it is very conservative, this does mean that we will miss valid cases where a folding could've taken place. For example, if an overwrite happens after the `__contains__` op, this check will prevent the fold from happening.

The structure of this function is borrowed from the one used for dict constructs, which does a slightly more comprehensive check:

https://github.com/llvm/torch-mlir/blob/e143a34948dcc845cddd8d21f5f13862520b1e0e/lib/Dialect/Torch/IR/TorchOps.cpp#L1487-L1495

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1529,6 +1529,40 @@ OpFoldResult Aten__Contains__StrOp::fold(ArrayRef<Attribute> operands) {
...
+// Aten__Contains__IntListOp
+//===----------------------------------------------------------------------===//
+
+static bool isListConstructNotModified(Value torchList) {
+  if (!llvm::all_of(torchList.getUsers(), [](Operation *op) {
+        return isa<Aten__Contains__IntListOp>(op);
+      }))
+    return false;
+
+  return torchList.getDefiningOp<Torch::PrimListConstructOp>();
```
ramiro050, line 1542: This should not compile. `getDefiningOp` returns an `Operation *`, not a `bool`. Please make sure things compile.

ramiro050, line 1542: This can be written as:

```cpp
return llvm::all_of(torchList.getUsers(), ...);
```

lib/Dialect/Torch/IR/TorchOps.cpp
ramiro050: I like this too, since it makes it symmetric with the `matchPattern` below

ramiro050 (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
JakopinA: I don't understand this, all_of is an iterator, not a returnable function, right? How could I return this function as a boolean?

JakopinA: Nevermind, it's an iterator that returns a function, my mistake.

lib/Dialect/Torch/IR/TorchOps.cpp
ramiro050: @JakopinA  ^

ramiro050 (approved):
LGTM!  There is just one comment below that was not addressed.

ramiro050:
> > LGTM! There is just one comment below that was not addressed.
> 
> I just didn't understand what you wanted me to do, I thought it was a compliment about my code.

Gaurav's comment was that you can just inline `item()` instead of creating a variable for it, the same way that `l()` is inlined in the next `matchPattern`

vivekkhandelwal1:
Hi @JakopinA please squash these 3 commits into one. Please make sure to mark the comments/conversations as resolved once you address them. This will simplify the reviewing process for the reviewer.

PR #964, opened 2022-06-22
index_add decomposed into index_put through TM Tensor

lorenzbc:
Combined all the commits from the index-add-to-put_TMTensor branch and move it to a new branch that's rebased

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -1637,6 +1637,56 @@ public:
 };
 } // namespace
 
+
+namespace {
+// Decompose `aten.index_add` op into `valsem.aten.index_put_impl` op. with
+// accumulate set to True
+class DecomposeAtenIndexAddOp : public OpRewritePattern<AtenIndexAddOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenIndexAddOp op,
+                                PatternRewriter &rewriter) const override {
+
+    Value input = op.self();
+    int inputRank = getTensorRank(input);
+    BaseTensorType inputType = input.getType().cast<BaseTensorType>();
+    Value updates = op.source();
+
+    if (inputRank < 0) {
+      return rewriter.notifyMatchFailure(op, "Invalid Input: Unranked tensor");
+    }
+
+    if (getTensorRank(updates) != inputRank) {
+      return rewriter.notifyMatchFailure(
+          op, "Unimplemented: Ranks of input and updates do not match");
+    }
+
+    inputType = input.getType().cast<BaseTensorType>();
+    SmallVector<int64_t> empty;
+    Type expandUpdatesType = inputType.getWithSizesAndDtype(
+        llvm::makeArrayRef(empty), rewriter.getF64Type());
+
+    Value torchCstOne = rewriter.create<Torch::ConstantIntOp>(
+        op.getLoc(), rewriter.getI64IntegerAttr(1));
+
+    Value expandedUpdates = rewriter.create<AtenUnsqueezeOp>(
+        op.getLoc(), expandUpdatesType, updates, torchCstOne);
+
+    Value indicesAsTensorList = rewriter.create<PrimListConstructOp>(
+        op.getLoc(), Torch::ListType::get(updates.getType()),
+        Value(op.index()));
+
+    Value accumulateTrue = rewriter.create<Torch::ConstantBoolOp>(op.getLoc(), true);
+    Value cstFalse = rewriter.create<Torch::ConstantBoolOp>(op.getLoc(), false);
+
+    rewriter.replaceOpWithNewOp<ValsemVariantAtenIndexPutImplOp>(
+        op, op.getType(), op.self(), indicesAsTensorList, expandedUpdates, accumulateTrue,
+        /*unsafe=*/cstFalse);
+    return success();
+  }
+};
+} // namespace
+
 namespace {
 class DecomposeAtenExpandAsOp : public OpRewritePattern<AtenExpandAsOp> {
   using OpRewritePattern::OpRewritePattern;
@@ -2101,6 +2151,8 @@ class DecomposeComplexOpsPass
     target.addIllegalOp<AtenFullLikeOp>();
     patterns.add<DecomposeAtenIndexPutOp>(context);
     target.addIllegalOp<AtenIndexPutOp>();
+    patterns.add<DecomposeAtenIndexAddOp>(context);
+    target.addIllegalOp<AtenIndexAddOp>();
     patterns.add<DecomposeAtenExpandAsOp>(context);
     target.addIllegalOp<AtenExpandAsOp>();
     patterns.add<DecomposeAten_ToCopyOp>(context);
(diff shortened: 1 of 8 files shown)


ramiro050:
You can run the lit tests locally by doing `cmake --build build --target check-torch-mlir`

lorenzbc:
@ramiro050 not sure why but the actual python unit test is failing. What could cause a "failed to legalize operation 'torch.aten.index_add' that was explicitly marked illegal" error?

ramiro050:
> @ramiro050 not sure why but the actual python unit test is failing. What could cause a "failed to legalize operation 'torch.aten.index_add' that was explicitly marked illegal" error?

That error means that your conversion pattern failed, so the op `torch.aten.index_add` stayed in the IR.

From the error message, you can tell that the input does have a rank, so the failure has to be in the `getTensorRank(updates) != inputRank` check. I think this is because the unit test is not right. In your unit test, the input has rank 3 and the `source` tensor has rank 2 according to the annotation.

python/torch_mlir_e2e_test/test_suite/index_put.py
```diff
@@ -818,3 +818,27 @@ def forward(self, input, index, value):
...
+      ([-1, -1], torch.float32, True),
+  ])
+  def forward(self, input, dim, index, value):
+    return torch.ops.aten.index_add(input, dim, index, value)
+
+
+@register_test_case(module_factory=lambda: IndexAddModule())
+def IndexAddModule_basic(module, tu: TestUtils):
+  module.forward(torch.rand((3, 5, 2), dtype=torch.float32),
+          torch.tensor(0.0),
```
ramiro050, line 842: This is a zero-rank tensor, so the annotation should be `[]`.

python/torch_mlir_e2e_test/test_suite/index_put.py
```diff
@@ -818,3 +818,27 @@ def forward(self, input, index, value):
...
+          torch.tensor([0, 1]),
```
ramiro050, line 843: This is an int tensor, which does not match the dtype in the annotation

python/torch_mlir_e2e_test/test_suite/index_put.py
```diff
@@ -818,3 +818,27 @@ def forward(self, input, index, value):
...
+          torch.rand((2, )))
```
ramiro050, line 844: This is a rank-1 tensor, which does not match the rank-2 annotation

ramiro050 (changes requested):

python/torch_mlir_e2e_test/test_suite/index_put.py
```diff
@@ -818,3 +818,27 @@ def forward(self, input, index, value):
...
+  @export
+  @annotate_args([
+      None,
+      ([-1, -1, -1], torch.float32, True),
+      ([-1], torch.float32, True),
+      ([-1], torch.float32, True),
+      ([-1, -1], torch.float32, True),
+  ])
+  def forward(self, input, dim, index, value):
+    return torch.ops.aten.index_add(input, dim, index, value)
```
ramiro050, line 836: can you also add an `alpha` value?

ramiro050 (changes requested):

python/torch_mlir_e2e_test/test_suite/index_put.py
lorenzbc: Can you clarify what you mean here?

python/torch_mlir_e2e_test/test_suite/index_put.py
```diff
@@ -818,3 +818,27 @@ def forward(self, input, index, value):
...
+  def __init__(self):
+    super().__init__()
+
+  @export
+  @annotate_args([
+      None,
+      ([-1, -1, -1], torch.float32, True),
+      ([-1], torch.float32, True),
+      ([-1], torch.float32, True),
+      ([-1, -1], torch.float32, True),
```
vivekkhandelwal1, line 833: These annotations should match with the dtype and shape of the arguments passed in `module.forward()`

vivekkhandelwal1 (changes requested):

python/torch_mlir_e2e_test/test_suite/index_put.py
ramiro050: The tensor has `dtype` int:

```python
In [3]: torch.tensor([0, 1]).dtype
Out[3]: torch.int64
```

which does not match the corresponding annotation here:

https://github.com/llvm/torch-mlir/blob/a8e50a68720ab37e832cc121224d5a8fbb8a78b9/python/torch_mlir_e2e_test/test_suite/index_put.py#L832

lorenzbc:
The op is still marked as illegal despite the changes even though now the annotations match the types and shapes of the arguments

ramiro050:
> The op is still marked as illegal despite the changes even though now the annotations match the types and shapes of the arguments

The test has inputs that are invalid in the decomposition. Here,

https://github.com/llvm/torch-mlir/blob/0c9084f7e14bb1479c7d20088d076b5eae3827a1/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L1659

there is an expectation for the `source` tensor to have the same rank as the `input` tensor, but in your test the `source` tensor has rank 1, while the `input` tensor has rank 3

https://github.com/llvm/torch-mlir/blob/0c9084f7e14bb1479c7d20088d076b5eae3827a1/python/torch_mlir_e2e_test/test_suite/index_put.py#L830-L833

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1637,6 +1637,56 @@ class DecomposeAtenIndexPutOp : public OpRewritePattern<AtenIndexPutOp> {
...
+        op.getLoc(), expandUpdatesType, updates, torchCstOne);
+
+    Value indicesAsTensorList = rewriter.create<PrimListConstructOp>(
+        op.getLoc(), Torch::ListType::get(updates.getType()),
+        Value(op.index()));
+
+    Value accumulateTrue = rewriter.create<Torch::ConstantBoolOp>(op.getLoc(), true);
+    Value cstFalse = rewriter.create<Torch::ConstantBoolOp>(op.getLoc(), false);
+
+    rewriter.replaceOpWithNewOp<ValsemVariantAtenIndexPutImplOp>(
```
navahgar, line 1682: This decomposition seems to ignore the `dim` parameter of `torch.index_add`. AFAICT, there is no way to include that in `index_put`. So, how is this going to work when `dim != 0`? Am I missing something?

navahgar: But I don't see the indices being transformed to the form that you mention here. I believe that should be done using the dim value and I don't see the dim being used anywhere here.

ramiro050: This PR is still a WIP, so there is indeed missing support currently for the `dim` argument.

sjain-stanford:
For the record, I plan to resurrect https://github.com/llvm/torch-mlir/pull/1063 since this effort / PR seems to be stale now. Please feel free to push back if that's not true..

ramiro050:
Yes, I think we can consider this stale, since there hasn't been an update in a year and a half.

riestmo-nxp:
Are your plans to resurrect https://github.com/llvm/torch-mlir/pull/1063 still active @sjain-stanford? I'm currently blocked by this error:

error: failed to legalize operation 'torch.operator' that was explicitly marked illegal
    %204 = torch.operator "torch.aten.index_add"(%165, %int0_249, %182, %203, %int1_250) : (!torch.vtensor<[3,2048],f32>, !torch.int, !torch.vtensor<[?],si64>, !torch.vtensor<[?,2048],f32>, !torch.int) -> !torch.vtensor<[3,2048],f32>

 Therefore I would be happy to see support for aten::index_add added.

PR #965, opened 2022-06-22
linalg: lower `aten.triu` op to `linalg.generic`

ashay:
Prior to this patch, the torch dialect included `AtenTriuOp` for
computing the upper triangular part of the input matrix, but there was
no code for lowering the op to the linalg dialect.

This patch adds code to generate a `linalg.generic` operation that
compares indices (computed using `linalg.index`) to choose between zero
or the original value (using `arith.select`).  The lowering fails if the
number of dimensions are less than two.  This patch also adds a few
end-to-end tests.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -857,6 +857,33 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     return b.create<arith::SelectOp>(loc, mask, fillValue, input);
   }
 
+  if (auto triu = dyn_cast<AtenTriuOp>(op)) {
+    // Check if the rank of the input tensor is valid.
+    AtenTriuOp::Adaptor adaptor(operands);
+    auto inputType = adaptor.self().getType().cast<RankedTensorType>();
+    uint64_t inputRank = inputType.getRank();
+    if (inputRank < 2) {
+      triu.emitError("too few dimensions to compute triangular part of matrix");
+      return nullptr;
+    }
+
+    // Use the indices of the two innermost dimensions.
+    auto rowIndex = b.create<linalg::IndexOp>(loc, inputRank - 2);
+    Value rowIndexI64 = castIndexToInt64(b, loc, rowIndex);
+    auto colIndex = b.create<linalg::IndexOp>(loc, inputRank - 1);
+    Value colIndexI64 = castIndexToInt64(b, loc, colIndex);
+
+    // columnIndex >= rowIndex + diagonal?
+    auto sum = b.create<arith::AddIOp>(loc, rowIndexI64, adaptor.diagonal());
+    auto pred = b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sge,
+                                        colIndexI64, sum);
+
+    Value scalar = payloadArgs[0];
+    Type elementType = inputType.getElementType();
+    Value zero = getConstant(b, loc, 0, elementType);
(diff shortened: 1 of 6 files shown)


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1625,6 +1625,72 @@ class ConvertTensorStaticInfoCastOp
...
+
+    if (inputRank < 2)
+      return op.emitError(
+          "too few dimensions to compute triangular part of matrix");
+
+    // For each dimension, compute the affine expression and the dimension size.
+    SmallVector<AffineExpr> exprs;
+    SmallVector<Value> outputDims;
+    for (auto idx = 0UL; idx < inputRank; idx++) {
+      exprs.push_back(rewriter.getAffineDimExpr(idx));
```
ramiro050, line 1650: `exprs` is not used

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1625,6 +1625,72 @@ class ConvertTensorStaticInfoCastOp
...
+      Value colIndexI64 = castIndexToInt64(rewriter, loc, colIndex);
+
+      // columnIndex >= rowIndex + diagonal?
+      auto sum = rewriter.create<arith::AddIOp>(loc, rowIndexI64, diagonal);
+      auto pred = rewriter.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sge,
+                                                 colIndexI64, sum);
+
+      Value scalar = payloadArgs[0];
+      auto select = rewriter.create<arith::SelectOp>(loc, pred, scalar, zero);
+      rewriter.create<linalg::YieldOp>(loc, select.getResult());
```
ramiro050, line 1671: I think this op can be lowered by just moving this code to the function

https://github.com/llvm/torch-mlir/blob/e91fc3cc1438639554394b8f75753b9d4ad57099/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L119

ramiro050 (changes requested):
Thanks for this! I just have a few comments

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1625,6 +1625,72 @@ class ConvertTensorStaticInfoCastOp
...
+
+    Value outVector =
+        rewriter.create<linalg::InitTensorOp>(loc, outputDims, elementType);
+    auto genericOp =
+        rewriter
+            .create<linalg::GenericOp>(loc, input.getType(), input, outVector,
+                                       indexingMaps, iteratorTypes, bodyBuilder)
+            .getResult(0);
+
+    rewriter.replaceOpWithNewOp<tensor::CastOp>(op, input.getType(), genericOp);
```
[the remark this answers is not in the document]
ashay: I see.  Thanks for the explanation!  Fixed.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
ashay: Woops, thanks for catching that.  Removed.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
ashay: I had missed that.  Rolled all of this into that function and removed the the `ConvertAtenTriuOp` class.

ramiro050 (approved):
LGTM! Thanks for this!

ashay:
Rebased with `main`

ashay:
Rebased with `main`.

PR #967, opened 2022-06-23
Add bazel builds to PR tests

makslevental:
In a "Lemony Snicket's A Series of Unfortunate Events" type move, I merged https://github.com/llvm/torch-mlir/pull/958 and then had to immediately submit https://github.com/llvm/torch-mlir/pull/966 as a patch because it broke [bazel build](https://github.com/llvm/torch-mlir/runs/7016365690?check_suite_focus=true), which slipped past me because it wasn't in the tests. Not sure what your guys budget is for GHA/CI but maybe bazel build should be part of the PR set too?

### DIFF
--- a/.github/workflows/bazelBuildAndTest.yml
+++ b/.github/workflows/bazelBuildAndTest.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+  pull_request:
 
 jobs:
   build:


silvasean:
We are under no obligation to fix the bazel build. Only people from that subcommunity should be getting notifications of breakages there and fixing it.

@asaadaldien -- can you iterate with Maks to debug why he got a notification for this?

makslevental:
> We are under no obligation to fix the bazel build. Only people from that subcommunity should be getting notifications of breakages there and fixing it. This is consistent with LLVM's policy on this -- https://llvm.org/docs/SupportPolicy.html#peripheral-tier
> 
> @asaadaldien -- can you iterate with Maks to debug why he got a notification for this?

oh i didn't get a notif i just saw that it failed and then scrambled to fix it. good to know that we don't have to worry about it!

PR #971, opened 2022-06-23
Add basic support for list of optional tensors in reduce-op-variants

ramiro050:
This commit adds support for lists of type `list<optional<tensor>>`
where each element in the list is either a `!torch.tensor` or a
`!torch.none`.

### DIFF
--- a/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
+++ b/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
@@ -36,6 +36,22 @@ static void createOverwriteTensorContents(PatternRewriter &rewriter,
                                              overwrittenTensor);
 }
 
+static Type getContainerOrTensorTypeWithValueSemantics(Type type) {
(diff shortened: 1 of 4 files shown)


ramiro050:
@silvasean PTAL

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -36,6 +36,20 @@ static void createOverwriteTensorContents(PatternRewriter &rewriter,
                                              overwrittenTensor);
 }
 
+static Type getContainerOrTensorTypeWithValueSemantics(Type type) {
+  if (auto optionalType = type.dyn_cast<OptionalType>()) {
+    Type newContainedType = getContainerOrTensorTypeWithValueSemantics(optionalType.getContainedType());
```
silvasean, line 41: is this 80 cols?

silvasean (approved):

PR #974, opened 2022-06-24
development.md: updating the LLVM submodule

silvasean:
We recently noticed that it is necessary to update the shape lib while
doing an LLVM bump: https://github.com/llvm/torch-mlir/pull/973
We didn't have a place to document that this was a necessary step, so
add a basic section to hold that and the basic instructions for how to
do a bump.


ramiro050 (approved):
Really useful info!

PR #975, opened 2022-06-24
Add E2E support for aten.warn

JakopinA:
Was wondering whether or not lowering makes any sense to have in this function?

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -6272,30 +6272,6 @@ def Torch_AtenDeleteDictStrOp : Torch_Op<"aten.Delete.Dict_str", [
   }];
 }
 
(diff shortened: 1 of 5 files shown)


silvasean:
I don't think this makes sense in this form -- what model are you running that needs this?

ramiro050:
> I don't think this makes sense in this form -- what model are you running that needs this?

I had added this op to the list of torchbench ops because it was being generated when importing `BERT_pytorch` into torch-mlir. However, I've just checked it again, and once the model gets to the maximize-value-semantics pass, the `warn` ops disappear from the graph. So this op might not be needed.

silvasean:
Okay, that is what I would have expected -- these should be in code paths that end up folded away.

vivekkhandelwal1:
> Okay, that is what I would have expected -- these should be in code paths that end up folded away. Same for is_grad_enabled.

And what about `prim.requires_grad`? @ramiro050 @silvasean

silvasean:
prim.requires_grad would probably be similar to is_grad_enabled.

PR #979, opened 2022-06-24
torch: allow torch dialect ops after running drop-shape pass

ashay:
In the `pyhpc_turbulent_kinetic_energy` TorchBench benchmark, the shape
calculation occurs inside loops, but because `DropShapeCalculationsPass`
does not explicitly mark the Torch dialect as legal, the pass execution
fails.

This patch adds Torch to the list of legal dialects, and adds a test to
validate the translation.


silvasean (approved):
Thanks! Btw, the way that our shape refinement works probably won't work for anything with loops (due to not being able to "guess" the type along backedges before looking at the body). So don't get your expectations too high here :)

ashay:
Oh, I see.  Thanks for the heads up!

PR #983, opened 2022-06-27
python: lower `prim::{Load,Store,Enter,Exit}` nodes to torch dialect

ashay:
TorchScript nodes like `prim::Load` and `prim::Store` aren't supported
in torch-mlir because they can't be lowered to backends, but such nodes
can occur in the TorchScript IR.

This patch adds a rudimentary translation from such nodes to
corresponding ops in the Torch dialect.  Since we expected such nodes to
go away during lowering because of the SymbolDCE pass, this patch does
not add code to lower these ops beyond the Torch dialect.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
@@ -549,6 +549,53 @@ def Torch_PrimIfYieldOp : Torch_Op<"prim.If.yield", [
   }];
 }
 
+def Torch_PrimStoreOp : Torch_Op<"prim.Store", []> {
+  let summary = "store operation";
+  let description = [{
+    This op represents a prim::Store node in the Python object graph.
+  }];
+  let arguments = (ins StrAttr:$name, AnyTorchType:$value);
+  let assemblyFormat = [{
+    $name `,` $value attr-dict `:` qualified(type($value))
+  }];
+}
+
+def Torch_PrimLoadOp : Torch_Op<"prim.Load", []> {
+  let summary = "load operation";
+  let description = [{
+    This op represents a prim::Load node in the Python object graph.
+  }];
+  let arguments = (ins StrAttr:$name);
+  let results = (outs AnyTorchType:$result);
+  let assemblyFormat = [{
+    $name attr-dict `:` qualified(type($result))
+  }];
+}
+
+def Torch_PrimEnterOp : Torch_Op<"prim.Enter", []> {
+  let summary = "enter operation";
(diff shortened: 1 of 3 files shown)


include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -1165,4 +1165,41 @@ def Torch_ShapeCalculateYieldShapesOp : Torch_Op<"shape.calculate.yield.shapes",
...
+def Torch_StoreOp : Torch_Op<"store", []> {
+  let summary = "store operation";
+  let description = [{
+    This op represents a prim::Store node in the Python object graph.
+  }];
+  let arguments = (ins StrAttr:$var, AnyTorchType:$value);
+  let assemblyFormat = [{ $var `,` $value attr-dict `:` qualified(type($value)) }];
+}
+
+def Torch_LoadOp : Torch_Op<"load", []> {
```
silvasean, line 1177: please follow the naming convention -- in this case, it would be prim::Load is `torch.prim.load`

test/python/importer/jit_ir/node_import/unimplemented.py
```diff
@@ -0,0 +1,34 @@
+import torch
+import torch_mlir
+
+# RUN: %PYTHON %s | torch-mlir-opt | FileCheck %s
+
+class Inner(object):
```
silvasean, line 6: can you add a test for enter/exit?

silvasean (changes requested):
Thanks!

I'm not sure what you mean by SymbolDCE removing them. SymbolDCE only removes functions. Can you show an IR example with before/after of what you mean?

silvasean:
Also, FYI I would recommend to not make "draft" pull requests -- if you intend for them to be reviewed mark them as non-draft (at least, that has been the convention we have been following).

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
ashay: Thanks!  Fixed by renaming it to `torch.prim.Load` (so that we can reuse some of the existing cases) and moved the ops to be grouped together with other interpreter builtin ops.

ashay:
You're right, SymbolDCE isn't removing individual ops, so the `torch.prim.{Enter,Exit}` ops remain in the code even after running SymbolDCE.  And thanks for the tip about not marking PRs ready for review as draft.  I'll keep that in mind for the future.

silvasean (approved):
Thanks!

PR #988, opened 2022-06-28
Fix the check line.

pashu123:


pashu123:
@silvasean Do we need this fix? I think I think checking for `torch.nn.Module` is enough.

silvasean:
We could do `} : !torch.nn.Module<"{{.*}}">`.

But first I want to understand what changed that made this suddenly be needed. Was it an upstream change?

pashu123:
> We could do `} : !torch.nn.Module<"{{.*}}">`.
> 
> But first I want to understand what changed that made this suddenly be needed. Was it an upstream change?

Yes, the latest upstream.

powderluv:
There is something funky with this setup. I need this PR for all tests to pass but the builder doesn't seem to.

powderluv:
Prashant is offline now so I have https://github.com/llvm/torch-mlir/pull/990 for now

powderluv:
Seems like an upstream fx bug https://github.com/pytorch/pytorch/issues/74058

rdadolf:
(disclaimer: not a torchscript whiz)

Looks like mangling can happen on repeated definitions of methods with the same fully-qualified name [docs](https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/docs/serialization.md#code-object-naming). Could scripting a second tanh module in the same e2e invocation cause this? In which case, the order of declaration could matter (and not possibly not cause errors, if the other instantiation just checked module type and not module name)?

It looks like other folks have run into this: https://github.com/pytorch/pytorch/issues/74058

For reference, here's an example that's counter-intuitive:

```
import torch
import torch_mlir
  
class TanhModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
    def forward(self, x): 
        return torch.ops.aten.tanh(x)
  
tanh_example_input = torch.ones(2, 3)
 
print(torch_mlir.compile(TanhModule(), tanh_example_input, output_type=torch_mlir.OutputType.RAW))
 
class TanhModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
    def forward(self, x): 
        return torch.ops.aten.tanh(x)
 
print(torch_mlir.compile(TanhModule(), tanh_example_input, output_type=torch_mlir.OutputType.RAW))
```
Will emit:
```
...
  } : !torch.nn.Module<"__torch__.TanhModule">
...
  } : !torch.nn.Module<"__torch__.___torch_mangle_0.TanhModule">
```
Commenting out the second declaration will emit:
```
...
  } : !torch.nn.Module<"__torch__.TanhModule">
...
  } : !torch.nn.Module<"__torch__.TanhModule">
```

I guess what I'm saying is that this is almost certainly caused by something else, and I'd wager it has to do with repeated declarations of that module.

If that's the case, though, the original test (and the fix!) are probably too brittle. Maybe the right answer is change the test regex to handle both the mangled and un-mangled cases, since both are technically correct.

powderluv:
``

PR #990, opened 2022-06-28
Update test to pass with  newer versions of tanh

powderluv:
Just taking over  https://github.com/llvm/torch-mlir/pull/988
with Sean's suggestion

TEST: passes on my system that failed previously

### DIFF
--- a/python/test/annotations-sugar.py
+++ b/python/test/annotations-sugar.py
@@ -29,7 +29,7 @@ extract_annotations(module, torch.jit.script(module), annotator)
 print(annotator)
 
 # CHECK: ClassAnnotator {
-# CHECK:   ClassAnnotation('__torch__.MmModule') {
+# CHECK:   ClassAnnotation('{{.*}}.MmModule') {
 # CHECK:     MethodAnnotation('forward') {
 # CHECK:       isExported = true
 # CHECK:       argAnnotations =
--- a/python/test/compile_api/basic.py
+++ b/python/test/compile_api/basic.py
@@ -37,7 +37,7 @@ print(torch_mlir.compile(TanhModule(), placeholder))
 # Basic smoke test for the raw output type.
 print(torch_mlir.compile(TanhModule(), tanh_example_input, output_type=torch_mlir.OutputType.RAW))
 # CHECK: torch.nn_module {
-# CHECK: } : !torch.nn.Module<"__torch__.TanhModule">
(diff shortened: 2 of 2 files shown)


rdadolf:
Migrating from the other PR. Sean wanted to know why. Can't tell you exactly, but I gave some justification why this fix should be safe enough. If you want to be more defensive, `<"{{.*}}.TanhModule">` or `<"__torch__{{.*}}.TanhModule">` should also be mangling-agnostic. Don't do `<"__torch__.{{.*}}.TanhModule">` (two `.`'s outside wildcard).

powderluv:
@rdadolf your suggest regex works well. Updated. Please review.

rdadolf:
Sorry to hold this up: do you mind repeating the same fix to [`python/test/annotations-sugar.py`](https://github.com/llvm/torch-mlir/blob/main/python/test/annotations-sugar.py#L32)? Duplicating that test in the same file (i.e., just copying everything and pasting it again below itself) causes it to fail with the same name-mangling issue. I don't think we need to expand the test to do that---the bug is with our lit check syntax, not anything load-bearing in torch-mlir.

I went ahead and checked our class annotator, too. I think it's safe from name-mangling bugs. And as long as users interact with `.compile()` instead of the individual pieces (like the annotator), I think we're safe from this problem leaking out to bite users in any way.

Besides that, LGTM. Thanks!

silvasean (approved):

powderluv:
@rdadolf I am not sure I understand. Do you want to copy over the same basic test into annotations-sugar.py ?  But isn't it better to keep it as small tests and run all of them for full coverage ? Happy to copy it over if you guys think it is good to duplicate the check.

rdadolf:
Sorry, I was unclear. All I meant was changing this line in `annotations-sugar.py`
```
# CHECK:   ClassAnnotation('__torch__.MmModule') {
```
to use the same regex
```
# CHECK:   ClassAnnotation('{{.*}}.MmModule') {
```

My only point was that if you take the existing annotations-sugar test and duplicate it inside itself, the same name-mangling problem breaks that test, because the class annotator that is generated ends up with a mangled module name.

So in order to protect ourselves from the same problem cropping up again in the future, I figured we should just switch out *all* the places where we explicitly check against module names. Turns out there's only this one other place (AFIAK).

rdadolf (approved):

PR #991, opened 2022-06-28
[tosa] aten.embedding op support

sjarus:
Enables BERT legalization.

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -2609,6 +2609,102 @@ LogicalResult ConvertAtenOp<AtenGeluBackwardOp>::matchAndRewrite(
   return success();
 }
 
+template <>
+LogicalResult ConvertAtenOp<AtenEmbeddingOp>::matchAndRewrite(
+    AtenEmbeddingOp op, OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+
+  Value weight = adaptor.weight();
+  Value indices = adaptor.indices();
+  RankedTensorType outType =
+      typeConverter->convertType(op.getType()).cast<RankedTensorType>();
+
+  auto indicesType = indices.getType().dyn_cast<RankedTensorType>();
+  if (!indicesType || !indicesType.getElementType().isa<IntegerType>())
+    return op.emitError("Indices must be of integer tensor type");
+
+  if (indicesType.getRank() != 2)
+    return op.emitError("indices must be of rank 2");
+
+  auto weightType = weight.getType().cast<RankedTensorType>();
+  if (weightType.getRank() != 2)
+    return op.emitError("weight must be of rank 2");
+
+  // FIXME: padding_idx, scale_grad_by_freq and sparse are not handled yet.
+  int64_t paddingIdx;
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -2609,6 +2609,80 @@ LogicalResult ConvertAtenOp<AtenGeluBackwardOp>::matchAndRewrite(
...
+    return op.emitError("Indices must be of integer tensor type");
+
+  if (indicesType.getRank() != 2)
+    return op.emitError("indices must be of rank 2");
+
+  auto weightType = weight.getType().cast<RankedTensorType>();
+  if (weightType.getRank() != 2)
+    return op.emitError("weight must be of rank 2");
+
+  // FIXME: padding_idx, scale_grad_by_freq and sparse are not handled yet.
```
silvasean, line 2633: can you make sure to bail out on those cases?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3173,11 +3247,11 @@ class ConvertTorchToTosa : public ConvertTorchToTosaBase<ConvertTorchToTosa> {
                                            tosa::AvgPool2dOp);
 #undef INSERT_ADAPTIVE_POOLING_ATEMOP_PATTERN
 
-target.addIllegalOp<AtenMaxPool2dOp>();
-patterns.add<ConvertAtenMaxPool2dOp>(typeConverter, context);
+    target.addIllegalOp<AtenMaxPool2dOp>();
```
silvasean, line 3250: nit: why the indent?

silvasean (approved):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
sjarus: Yeah that's weird. Looks like a preexisting condition that's not covered by my insurance here though. I'm surprised the clang-format didn't fix it, or maybe it's what broke it.

nithinsubbiah:
Hi @sjarus, I tried out this patch to lower minilm to TOSA but it fails with a seg fault specifically an assertion error.
```
torch-mlir/externals/llvm-project/llvm/include/llvm/ADT/SmallVector.h:273: llvm::SmallVectorTemplateCommon::reference llvm::SmallVectorTemplateCommon<TensorShape_t>::operator[](llv
m::SmallVectorTemplateCommon::size_type) [T = TensorShape_t]: Assertion `idx < size()' failed.
```
The script that I used is here (https://gist.github.com/nithinsubbiah/66c3a9137f254f5d327929cfef11d879), I was able to get Torch IR with no dynamic shapes.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
silvasean: Don't worry about it. Yeah, fine to leave it.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
sjarus: Done

silvasean (approved):

PR #993, opened 2022-06-29
torch: clone slot initializers if aliased

ashay:
The IR generated by TorchScript may use the same initializer for
different slots, but doing so violates an invariant in torch-mlir.
This patch adds a workaround to torch-mlir, by cloning the aliased
initializers.  This patch also updates the corresponding MLIR test to
now look for a valid Torch dialect IR (instead of looking for an
invariant failure error message).


silvasean:
This is not a semantically correct transform (i.e. it can miscompile). I think you need to do something more sophisticated here to understand if the object identity is meaningful -- for example, if all uses are getattr's and all uses of those getattr's do not create aliases or mutate (HasValueSemantics can be used as a proxy for this), then it is okay to do this.

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
```diff
@@ -258,9 +295,9 @@ class ObjectGraphInfo {
   // which cannot be used in multiple initializers because their object
   // identity is important.
   DenseSet<Value> objectsWithIdentityAlreadyCopiedIntoInitializers;
-  // Used to keep track of all the used torch slots so that the restrictions can
-  // be applied to those slots only.
-  DenseSet<SlotOp> usedSlots;
+  // Used to keep track of the uses of all torch slots so that we can determine
```
silvasean, line 298: nit: users, not uses

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
```diff
@@ -215,15 +241,26 @@ class ObjectGraphInfo {
     BlockAndValueMapping mapping;
     for (Operation *op : worklist) {
       builder.clone(*op, mapping);
-      for (Value result : op->getResults()) {
+      for (auto result_it : llvm::enumerate(op->getResults())) {
```
silvasean, line 244: resultAndIndex

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
```diff
@@ -142,11 +143,36 @@ class ObjectGraphInfo {
...
+      if (!isa<SlotOp>(use.getOwner()))
+        return checkUsesForValueSemantics(use.get());
+
+      // For `torch.slot` operations, lookup uses of the slot using the
+      // `slotUses` map, before checking whether each use is either free of side
+      // effects (e.g. a list constructor) or has value semantics.
+      auto traitCheck = [](Operation *operation) {
+        return llvm::all_of(operation->getUses(), [](OpOperand &operand) {
+          auto owner = operand.getOwner();
+          return mlir::isSideEffectFree(owner) ||
```
silvasean, line 163: isSideEffectFree isn't the property we want -- an op that isSideEffectFree can still create aliases. Can you check only for HasValueSemantics here?

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
```diff
@@ -142,11 +143,36 @@ class ObjectGraphInfo {
...
+      slotUses[slotIt->getValue()].insert(op);
     });
     return success();
   }
 
+  LogicalResult checkUsesForValueSemantics(Value root) {
+    for (auto &use : root.getUses()) {
+      // For non-slot operations, simply trace the value's uses.
+      if (!isa<SlotOp>(use.getOwner()))
+        return checkUsesForValueSemantics(use.get());
```
silvasean, line 155: isn't `use.get() == root` always? Doesn't this code infinitely recurse?

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
```diff
@@ -142,11 +143,36 @@ class ObjectGraphInfo {
       if (slotIt == nameToSlot.end())
         op->emitError() << "Reference to non-existing module slot " << slotName
                         << "in " << moduleType.getClassName();
-      usedSlots.insert(slotIt->getValue());
+      slotUses[slotIt->getValue()].insert(op);
     });
     return success();
   }
 
+  LogicalResult checkUsesForValueSemantics(Value root) {
```
silvasean, line 151: I would call this `bool isSlotInitializerUsedOnlyWithValueSemantics(Value initializer)`

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
```diff
@@ -215,15 +241,26 @@ class ObjectGraphInfo {
...
                  .second) {
-          return op->emitError() << "potentially-aliased value used to "
-                                    "initialize multiple slots";
+          // Check whether the value's uses ever create aliases or mutate.
+          if (failed(checkUsesForValueSemantics(result)))
+            return op->emitError()
+                   << "unsafe initializer used to initialize multiple slots";
+
+          // Create a deep clone of the `result` value before inserting it.
+          Operation *clonedOp = op->clone();
```
silvasean, line 258: cloning seems like it could be a problem if it is a large weight buffer -- in practice, how big are the things you are cloning? And do some of the clones get deleted?

silvasean (changes requested):

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
ashay: Thanks, fixed.

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
ashay: Fixed.

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
ashay: I see.  The problem is that I have a `torch.prim.TupleConstruct` (see updated test in `//test/Dialect/Torch/GlobalizeObjectGraph/initializers.mlir`) but since that op doesn't have the `HasValueSemantics` trait associated with it, the guard condition fails.  Are list / tuple constructors safe users of `GetAttr`s?

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
ashay: Oh dang, thanks for catching that.  I meant to say `use.getOwner()->getResults()`.  Fixed and updated the test to exercise this code path.

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
ashay: Sadly, for the model that I'm compiling, the buffers are large (some tens of millions of values), but that would, of course, change with the model.  I haven't studied the uniqueness requirement, so pardon by ignorance, but will it be a significant undertaking to remove the non-aliasing requirement for initializers?

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
silvasean: No, they are not safe. You could read out the tensor from the Tuple or List and then mutate it.

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
silvasean: Is there a way for you to construct a reduced test case that I could poke at?

ashay: Sure, here's one starting from Python: https://gist.github.com/ashay/5f26017affc81fe0eaba81b2935bc950.  The comment on that gist shows the error message.  Thanks for offering to look into this!

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
ashay: Oh, that's right.  Makes sense now, thanks!

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
silvasean: Thanks! I have a solution, but it makes torch.global_slot initializers quite a bit more complex for passes to handle. Basically we allow the initializer for one slot to reference another slot. This makes InlineGlobalSlots quite a bit more complex.

So, I'm trying to understand the need for this feature better to weigh it against the increase in complexity of the system. Can you explain a bit more how this occurs "in the wild"? Is this just a pathological example, or is there some common pattern that will end up triggering this from many users?

powderluv:
hi @ashay after this PR please use your personal fork for PRs and use llvm/torch-mlir forks if there are large integration branches required etc. Thanks for all your contributions.

ashay:
> hi @ashay after this PR please use your personal fork for PRs and use llvm/torch-mlir forks if there are large integration branches required etc. Thanks for all your contributions.

Sure, sounds good @powderluv.  Thanks for the heads up!

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
ashay: Sorry, this took a while to comb through the model source code.  It seems as though the key culprit is a reference to `self.parameters()` in the top-level module.  I'm having a hard time reproducing the exact problem in a smaller example, but I'll let you know once I succeed.

ashay: The problem is that `self.parameters()` refers to not just the parameters of the calling `torch.nn.Module` but also the parameters of the member fields.  [Here](https://gist.github.com/ashay/15ecab16886356424e27b60649328da1) is an example, which creates an embedding layer, but the top-level module reaches into the embedding layer to use its weight parameter.

silvasean: Okay, that seems pretty reasonable. I will work on a fix for this.

ashay:
Closing this PR, since cloning slot initializers isn't the best approach.

lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
silvasean: Started work on this in https://github.com/llvm/torch-mlir/pull/1047 PTAL

silvasean:
https://github.com/llvm/torch-mlir/pull/1140 should cover this now.

PR #995, opened 2022-06-29
[MLIR][TORCH] Add decomposition of `aten.var.dim` op

vivekkhandelwal1:
This commit adds the decomposition of `aten.var.dim` op.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -40,8 +40,8 @@ static bool isNoneOrFloatDtype(MLIRContext *context, Value dtype) {
 // `dim` specifies the dimension to reduce and `keepDim` preserves the rank of
 // the input tensor.
 static Type computeReductionType(PatternRewriter &rewriter, Operation *op,
-                                 Value input, Value dim, bool keepDim) {
-  BaseTensorType tensorType = input.getType().cast<BaseTensorType>();
+                                 BaseTensorType tensorType, Value dim,
+                                 bool keepDim) {
   SmallVector<int64_t> sizes;
   int64_t dimInt;
   if (tensorType.hasSizes()) {
@@ -81,7 +81,8 @@ static Value createSumAlongDimension(PatternRewriter &rewriter, Location loc,
       loc, Torch::ListType::get(dim.getType()), dim);
   Value keepDimCst = rewriter.create<ConstantBoolOp>(loc, keepDim);
   Value dtype = rewriter.create<ConstantNoneOp>(loc);
-  Type resultType = computeReductionType(rewriter, op, input, dim, keepDim);
+  Type resultType = computeReductionType(
+      rewriter, op, input.getType().cast<BaseTensorType>(), dim, keepDim);
   if (!resultType)
     return nullptr;
   return rewriter.create<AtenSumDimIntListOp>(loc, resultType, input, dimList,
@@ -94,7 +95,8 @@ static Value createMaxAlongDimension(PatternRewriter &rewriter, Location loc,
                                      bool keepDim) {
   Value keepDimCst = rewriter.create<ConstantBoolOp>(loc, keepDim);
   BaseTensorType valueType =
-      computeReductionType(rewriter, op, input, dim, keepDim)
+      computeReductionType(rewriter, op, input.getType().cast<BaseTensorType>(),
+                           dim, keepDim)
           .cast<BaseTensorType>();
   if (!valueType)
     return nullptr;
@@ -211,7 +213,10 @@ public:
     Value startPlusOne =
         rewriter.create<AtenAddIntOp>(loc, one.getType(), start, one);
     Value slice = rewriter.create<AtenSliceTensorOp>(
-        loc, computeReductionType(rewriter, op, self, dim, /*keepDim=*/true),
+        loc,
+        computeReductionType(rewriter, op,
+                             self.getType().cast<BaseTensorType>(), dim,
+                             /*keepDim=*/true),
         op.self(), dim, start, startPlusOne, /*step=*/one);
 
     // `aten.slice.tensor` doesn't squeeze the dim even when it's size 1 after
@@ -1112,8 +1117,7 @@ public:
 };
 } // namespace
 
-// Decompose aten.var into: sum(square(x - mean))/(numTensorElements-1)
-// for unbiased and mean(square(x - mean)) for biased case.
+// Decompose aten.var into: aten.var.dim op.
 namespace {
 class DecomposeAtenVarOp : public OpRewritePattern<AtenVarOp> {
 public:
@@ -1122,12 +1126,7 @@ public:
                                 PatternRewriter &rewriter) const override {
     Location loc = op.getLoc();
     Value self = op.self();
-    BaseTensorType inputTensorTy = self.getType().cast<BaseTensorType>();
-    if (!inputTensorTy.hasDtype() ||
-        !inputTensorTy.getDtype().isa<mlir::FloatType>()) {
-      return rewriter.notifyMatchFailure(op,
-                                         "Only aten.var support floating type");
-    }
+    unsigned inputRank = getTensorRank(self);
     BaseTensorType rank0FloatTensorTy = op.getType().cast<BaseTensorType>();
     if (!rank0FloatTensorTy.hasSizes() ||
         rank0FloatTensorTy.getSizes().size() != 0) {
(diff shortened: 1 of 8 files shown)


ramiro050:
The decomposition for `AtenVarOp` is very similar to `AtenVarDimOp`. I think we should have `AtenVarDimOp` take the current code for `AtenVarOp` with `AtenMeanOp` replaced with `AtenMeanDimOp` and `AtenSumOp` with `AtenSumDimIntListOp`, and then we can just decompose `AtenVarOp` into `AtenVarDimOp`.

vivekkhandelwal1:
> The decomposition for `AtenVarOp` is very similar to `AtenVarDimOp`. I think we should have `AtenVarDimOp` take the current code for `AtenVarOp` with `AtenMeanOp` replaced with `AtenMeanDimOp` and `AtenSumOp` with `AtenSumDimIntListOp`, and then we can just decompose `AtenVarOp` into `AtenVarDimOp`.

Hi @ramiro050, sorry, I am not able to clearly understand your comment. Do you mean to say we don't need decomposition for `aten.var.dim`? I can't see any decomposition for the `aten.var` op.

ramiro050:
> > The decomposition for `AtenVarOp` is very similar to `AtenVarDimOp`. I think we should have `AtenVarDimOp` take the current code for `AtenVarOp` with `AtenMeanOp` replaced with `AtenMeanDimOp` and `AtenSumOp` with `AtenSumDimIntListOp`, and then we can just decompose `AtenVarOp` into `AtenVarDimOp`.
> 
> Hi @ramiro050, sorry, I am not able to clearly understand your comment. Do you mean to say we don't need decomposition for `aten.var.dim`? I can't see any decomposition for the `aten.var` op.

Here is the decomposition we currently have for `aten.var`

https://github.com/llvm/torch-mlir/blob/d4f1f414354822cace5250889eb738bc5b00579b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L1097

What's nice of using the existing code for `aten.var.dim` is that we also get the support for the `unbiased` flag for free

vivekkhandelwal1:
Hi @ramiro050, I have modified the decompositions for `aten.var` and `aten.var.dim` as suggested by you. Please review the changes.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2101,6 +2078,92 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
...
+    // TODO: Add support for keepdim arg equal to `False` and dimList size not
+    // equal to input rank.
+    if (!keepDim.getType().isa<Torch::NoneType>()) {
+      bool keepdim;
+      if (!matchPattern(keepDim, m_TorchConstantBool(&keepdim)))
+        return rewriter.notifyMatchFailure(
+            op, "unimplemented: keepdim must be a constant");
+      else if (!keepdim && dimListConstruct.elements().size() != inputRank)
+        return rewriter.notifyMatchFailure(
+            op, "unimplemented: for keepdim = false, no. of reduction dims "
```
ramiro050, line 2122: What is the reason for this limitation? I would expect things to work since `AtenMeanDimOp` and `AtenSumDimIntListOp` take care of reducing the dimensions

vivekkhandelwal1: For the cases like, `Input size : [2,3,4], unbiased=False/True, dim=[0,2], keepdim=False`, https://github.com/llvm/torch-mlir/blob/645e7e6fe39a6c1a97de5d01ead2b9d324dbdffe/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L2133 will generate a tensor of size [3] only, which results in an error for https://github.com/llvm/torch-mlir/blob/645e7e6fe39a6c1a97de5d01ead2b9d324dbdffe/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L2135-L2136 because the `self tensor` is of size [2,3,4] and `meanAlongDims` is of size [3]. I think we don't have this kind of broadcasting support for the tensors in these ops lowering. That's why I have added this restriction.

ramiro050: I think `keepDim` should always be `True` in `AtenMeanDimOp`, since at this point in the decomposition we are not yet reducing the rank of the input tensor.

vivekkhandelwal1: In that case, we have to manually create the `resultType` for `AtenMeanDimOp` when `keepDim=False`, since the result type of the Var op will not match the result type of the `AtenMeanDimOp`.

vivekkhandelwal1: Hi @ramiro050, I have removed this limitation from the code via some modifications. Thanks for your suggestions. Please review this PR so that it can be merged.

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -227,91 +227,143 @@ func.func @torch.aten.square(%arg0: !torch.vtensor<[?,?,?],f32>) -> !torch.vtens
...
+// CHECK:           %[[DIM0_0:.*]] = torch.aten.size.int %[[INPUT]], %[[CST0]] : !torch.vtensor<[?,?],f32>, !torch.int -> !torch.int
+// CHECK:           %[[MUL_0:.*]] = torch.aten.mul.int %[[CST1_1]], %[[DIM0_0]] : !torch.int, !torch.int -> !torch.int
+// CHECK:           %[[DIM1_0:.*]] = torch.aten.size.int %[[INPUT]], %[[CST1]] : !torch.vtensor<[?,?],f32>, !torch.int -> !torch.int
+// CHECK:           %[[NUM_ELEMENTS_0:.*]] = torch.aten.mul.int %[[MUL_0]], %[[DIM1_0]] : !torch.int, !torch.int -> !torch.int
+// CHECK:           %[[CST1_2:.*]] = torch.constant.int 1
+// CHECK:           %[[NUM_ELEMENTS_0_SUB_1:.*]] = torch.aten.sub.int %[[NUM_ELEMENTS_0]], %[[CST1_2]] : !torch.int, !torch.int -> !torch.int
+// CHECK:           %[[UNBIASED_VAR:.*]] = torch.aten.div.Scalar %[[SUB_MEAN_SQUARE_SUM]], %[[NUM_ELEMENTS_0_SUB_1]] : !torch.vtensor<[],f32>, !torch.int -> !torch.vtensor<[],f32>
 // CHECK:           return %[[UNBIASED_VAR]] : !torch.vtensor<[],f32>
-func.func @torch.aten.var$unbiased(%arg0: !torch.vtensor<[?,?,?],f32>) -> !torch.vtensor<[],f32> {
+func.func @torch.aten.var$unbiased(%arg0: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[],f32> {
```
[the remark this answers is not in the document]
vivekkhandelwal1: Just to make sure that the length of generated  IR is less. So that it's a bit easy to add the checks.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -37,32 +37,34 @@ static bool isNoneOrFloatDtype(MLIRContext *context, Value dtype) {
...
+          return nullptr;
+        }
+        // The dimension to be reduced is set to 1 when `keepDim` is true else
+        // it is removed.
+        if (keepDim)
+          sizes[dimInt] = 1;
+        else
+          sizes.erase(sizes.begin() + dimInt - 1);
+      } else {
+        sizes.resize(reducedRank, kUnknownSize);
```
ramiro050, line 66: there should be a `break` here to avoid running this computation over and over

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -37,32 +37,34 @@ static bool isNoneOrFloatDtype(MLIRContext *context, Value dtype) {
...
+        if (!isValidDim(dimInt, inputRank)) {
+          (void)rewriter.notifyMatchFailure(op, "dim is not a valid dim");
+          return nullptr;
+        }
+        // The dimension to be reduced is set to 1 when `keepDim` is true else
+        // it is removed.
+        if (keepDim)
+          sizes[dimInt] = 1;
+        else
+          sizes.erase(sizes.begin() + dimInt - 1);
```
ramiro050, line 64: I don't think this is correct. Suppose you have a rank three tensor and `dims = {0, 1, 2}`. After the first two iterations, `sizes.size()` will be 1, which means that the last iteration `sizes.erase(sizes.begin() + 2 - 1)`, will try to remove an element out of bounds.

Also, I know this calculation was already here, but I don't think it is right. If `dimInt = 0`, then this line will try to erase a value out of bounds

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
vivekkhandelwal1: I can do one thing, undo the changes made to this function, and call it from the decomposition class directly in a loop for each dim to be reduced. Would that be fine?

ramiro050: Yeah, since you're always calling it with `keepDim=True`, the approach you mentioned should be fine

ramiro050 (approved):
LGTM! Just a small nit

PR #1001, opened 2022-06-30
Torch-MLIR LTC Backend

antoniojkim:
The LTC staging branch is finally at a point where it is stable enough against latest PyTorch `master` that we can merged the staging branch into `main` and continue working off of `main` instead of off the staging branch.

All of the commits on this branch came from PRs that were individually reviewed internally at Cerebras and then got the stamp of approval to merge in by someone else on the Torch-MLIR project (mainly @silvasean).

Open to any feedback as this is a fairly major change. Having said that, the LTC backend is fairly well decoupled from the rest of Torch-MLIR, so it shouldn't break anything.

CC: @ke1337 @henrytwo


examples/ltc_backend/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+###########################################################################
+# Setup PyTorch
+###########################################################################
+
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/python/torch_mlir/cmake/modules")
+include(TorchMLIRPyTorch)
+TorchMLIRProbeForPyTorchInstall()
```
silvasean, line 7: I think this recently changed with https://github.com/llvm/torch-mlir/pull/955

test/lit.cfg.py
```diff
@@ -67,6 +67,8 @@
 
 if config.enable_bindings_python:
   llvm_config.with_environment('PYTHONPATH', [
-      os.path.join(config.torch_mlir_python_packages_dir, 'torch_mlir'),
-  ],
-                               append_path=True)
+      os.pathsep.join([
+          os.path.join(config.torch_mlir_python_packages_dir, 'torch_mlir'),
+          os.environ['PYTHONPATH'],
```
silvasean, line 72: why is this needed?

silvasean (commented):
A few initial comments. I would like to do another round of review after the docs I mentioned in the other PR have been added so I can orient myself better.

rdadolf:
Not to be "that guy", but is there some way we can rename/disambiguate "backend"? I recognize that you're coming at this from a PyTorch angle where what you've built is called a "backend", but torch-mlir also already has a concept of "backend" that's used quite a bit, and it's unrelated.

It seems like maybe there's room for a new name here. @silvasean maybe you can weigh in?

python/torch_mlir/dialects/torch/importer/jit_ir/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@
 # Setup PyTorch
 #-------------------------------------------------------------------------------
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
+list(APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/python/torch_mlir/cmake/modules")
```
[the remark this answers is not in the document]
antoniojkim: It was required as we moved the `cmake/modules` to the top level `python/torch_mlir`. It no longer lives under `python/torch_mlir/dialects/torch/importer/jit_ir`. Is there a better way to get this to work?

Also, where might Torch-MLIR not be main project? Isn't the intention to make it a pip installable package?

test/lit.cfg.py
antoniojkim: I think this was required to make it work with our local environment and was accidentally included. Will revert

antoniojkim:
> Not to be "that guy", but is there some way we can rename/disambiguate "backend"? I recognize that you're coming at this from a PyTorch angle where what you've built is called a "backend", but torch-mlir also already has a concept of "backend" that's used quite a bit, and it's unrelated.
> 
> It seems like maybe there's room for a new name here. @silvasean maybe you can weigh in?

We've had this discussion before. This is why in the directory structure, we've named it `lazy_base_backend` from its original `ltc_backend` and its intended to be the base for a more concrete backend.

rdadolf:
> We've had this discussion before.

Apologies if I'm re-hashing things. If there's a record of the conversation you could point me to, I'll try to catch up from there.

> This is why in the directory structure, we've named it `lazy_base_backend` from its original `ltc_backend` and its intended to be the base for a more concrete backend.

So I think we may be talking about different things (or maybe I'm just misinterpreting you), because the `lazy_base_` vs. `ltc_` distinction doesn't really relate. I'm talking about the use of the term `backend`. In your case, you're talking about `backend` in the sense that PyTorch uses it, but torch-mlir also uses `backend` to describe the conversation paths from torch to various other MLIR dialects. E.g., everything in `lib/Dialect/TorchConversion/Transforms`. There doesn't seem to be much semantic overlap between torch->tosa, torch->linalg, and the LTC work.

To be clear, this isn't a criticism of LTC---the way you're using "backend" makes total sense. This is more of an issue that stems from two independent projects (PyTorch and torch-mlir) having a name collision. To avoid ambiguity and chaos, though, it's probably up to us to resolve that name collision in a sane way. And maybe that even means renaming all the existing things in torch-mlir instead of LTC. I'm not sure.

antoniojkim:
> > We've had this discussion before.
> 
> Apologies if I'm re-hashing things. If there's a record of the conversation you could point me to, I'll try to catch up from there.

There is some discussion [here](https://github.com/llvm/torch-mlir/pull/724#discussion_r839824865). But mostly to do with making it more clear.

> > This is why in the directory structure, we've named it `lazy_base_backend` from its original `ltc_backend` and its intended to be the base for a more concrete backend.
> 
> So I think we may be talking about different things (or maybe I'm just misinterpreting you), because the `lazy_base_` vs. `ltc_` distinction doesn't really relate. I'm talking about the use of the term `backend`. In your case, you're talking about `backend` in the sense that PyTorch uses it, but torch-mlir also uses `backend` to describe the conversation paths from torch to various other MLIR dialects. E.g., everything in `lib/Dialect/TorchConversion/Transforms`. There doesn't seem to be much semantic overlap between torch->tosa, torch->linalg, and the LTC work.

There isn't meant to be overlap per say. This is why its a "base" lazy backend. I can definitely see us adding a TOSA lazy backend and/or a LinAlg lazy backend that are both derived from the base lazy backend.

To that end, I can kind of see where you're coming from. It isn't an actual backend, only a "base". So, including "backend" in the name might be a bit confusing. Having said that, I'm still in favour of keeping it as is. @henrytwo and I are planning on adding some documentation to make things more clear.

examples/ltc_backend/CMakeLists.txt
powderluv: yes please use TorchMLIRConfigureLibTorch(). We would like to remove the need to probe TorchMLIRProbeForPyTorchInstall() and also remove the need for PyTorch during our build.

.github/workflows/buildAndTest.yml
```diff
@@ -5,6 +5,8 @@ on:
     branches:
       - main
   pull_request:
+    branches:
+      - main
```
powderluv, line 9: I would leave it for all branches (or just ignore branches you don't want like LTC). Since we are starting to use llvm/torch-mlir branches as integration branches (or special test branches) let the CI run on them.

python/torch_mlir/csrc/README.md
```diff
@@ -0,0 +1,19 @@
...
+such as `Compile` were purposefully not overriden as Torch-MLIR does not know
+how to compile the model for the target hardware.
+
+The intent is that vendor hardware specific plugins will subclass the Torch-MLIR
+backend classes and override the remaining pure virtual functions to complete
+the backend.
+
+The Torch-MLIR LTC backend's job is to perform the lowering from ATen to MLIR. A
+hardware vendor's backend job is to take care of the actual compile and
+execution of the lowered MLIR.
```
silvasean, line 19: maybe add a note here about the coding style being different from the rest of the project due to being consistent with LTC itself.

python/torch_mlir/dialects/torch/importer/jit_ir/CMakeLists.txt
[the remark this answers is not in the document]
silvasean: Ah, I think I might have this confused with another CMake variable -- please disregard. I think this is ok.

examples/ltc_backend/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
+###########################################################################
```
silvasean, line 1: this is awesome. is there a way we can get this example running under CI? It has a pretty complex build surface area, and seems prone to breakage. I would personally put it somewhere inside the main tree (not examples) and write some small python unit tests that invoke it.

silvasean (commented):
Overall this looks good to me.

examples/ltc_backend/CMakeLists.txt
henrytwo: > I would personally put it somewhere inside the main tree (not examples)
Do you want the example backend to be moved outside the examples directory? This stuff should already be built automatically with the rest of Torch-MLIR, so I would assume that it should just build with CI (once CI isn't broken at least...)

We previously added e2e tests, which relies on this example backend, so in theory this should be covered by CI tests

silvasean:
Hey folks, is there anything blocking us from merging this?

antoniojkim:
@silvasean @powderluv Any ideas why the CI is failing now? It seems to be unable to find a number of pybind symbols during linking. Not just for the pybind added in the reference lazy backend, but also for the class annotator pybind. I'm unable to reproduce this in my local build, so it must be something to do with the CI build environment itself.

rdadolf:
Looks like Python symbols, probably caused by [the class annotator](https://github.com/llvm/torch-mlir/runs/7361050083?check_suite_focus=true#step:5:477) not being linked to Python ([ref 1](https://github.com/pybind/pybind11/issues/2232), [ref 2](https://github.com/pytorch/pytorch/issues/70882#issuecomment-1016942891)).

Unclear on the root cause, though. Honestly, I've had trouble keeping up with the rapid changes to the build system.

Looks like the cmake find/configure order was changed in #1058. Could that be causing this?

antoniojkim:
> Looks like the cmake find/configure order was changed in #1058 (in the [importer's cmakelists](https://github.com/llvm/torch-mlir/pull/1058/files#diff-2bc9ebcf5c8ff13c8ea9b98dcc8fec16629cea811c921a8ddcca70185b65c81c)). Could that be causing this?

Perhaps @powderluv can comment on this further. But if this was the case, I would expect it to break locally as well. However, it only seems to be breaking in CI

powderluv:
Unfortunately the CI is fragile right now and can break even if it passes locally :( Im testing your branch now to see if I can spot anything.

powderluv:
The right version of PyTorch is found and configured for this build 
https://github.com/llvm/torch-mlir/runs/7361050373?check_suite_focus=true#step:4:355

``` 
-- Found Torch: /opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/torch/lib/libtorch.so (Required is at least version "1.11") 
-- Checking PyTorch ABI settings...
-- PyTorch C++ Dual ABI setting: "0"
-- PyTorch C++ ABI version: "11"
-- TORCH_CXXFLAGS=-D_GLIBCXX_USE_CXX11_ABI=0 -U__GXX_ABI_VERSION -D__GXX_ABI_VERSION=1011 '-DPYBIND11_COMPILER_TYPE="_gcc"' -Wno-pedantic 
```

All the regression tests also passed:

```
[2925/2926] Running the torch-mlir regression tests
-- Testing: 93 tests, 2 workers --
Testing:  0.. 10.. 20.. 30.. 40.. 50.. 60.. 70.. 80.. 90.. 
Testing Time: 20.83s
  Passed: 93
```
  
This looks like some other test failure in LTC. Could still be something build related but not the basic sanity tests / integrations.

antoniojkim:
@powderluv I'm fairly certain the failing tests are due to some unsupported LTC ops. @henrytwo is working on xfailing those tests for LTC.

The issue we were talking about were the python linking errors that are happening in the `Build out-of-tree`

powderluv:
I am trying a test build with the ordering flipped https://github.com/llvm/torch-mlir/actions/runs/2693995666 if it helps the out-of-tree build. If it does we can carry it as part or the LTC branch or I can land it first.

powderluv:
That didn't help so it is unlikely the Pytorch changes. It seems like the Python flags may be getting reset/cleared when generating LTC torchgen.

henrytwo:
@powderluv @silvasean Are there any additional changes we need in this PR before landing in main? If so, could we please get some approval stamps?

As soon as https://github.com/llvm/torch-mlir/pull/1110 goes in we expect CI to be all green here too, since the changes are identical.

silvasean (approved):
Once CI is green, let's land this!

powderluv (approved):
Let's land once CI is green. Thanks for pushing through with this.

powderluv (approved):
Let's land once CI is green. Thanks for pushing through with this.

silvasean:
Great follow-through folks! I'm so excited to see this land!

Issue #994, opened 2022-06-29
macOS e2e tests fail due to multi-processing


powderluv:
build with:
```
(mlir_venv) (base) anush@denali torch-mlir % cmake -GNinja -Bbuild   -DCMAKE_BUILD_TYPE=Release   -DCMAKE_C_COMPILER=clang   -DCMAKE_CXX_COMPILER=clang++   -DPython3_FIND_VIRTUALENV=ONLY   -DLLVM_ENABLE_PROJECTS=mlir   -DLLVM_EXTERNAL_PROJECTS="torch-mlir;torch-mlir-dialects"   -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR=`pwd`   -DLLVM_EXTERNAL_TORCH_MLIR_DIALECTS_SOURCE_DIR=`pwd`/externals/llvm-external-projects/torch-mlir-dialects   -DMLIR_ENABLE_BINDINGS_PYTHON=ON   -DLLVM_TARGETS_TO_BUILD=host externals/llvm-project/llvm  && cmake --build build --target check-torch-mlir-python
```

Fails with the following on an intel mac system

```

********************
********************
Failed Tests (6):
  TORCH_MLIR_PYTHON :: torchscript_e2e_test/basic.py
  TORCH_MLIR_PYTHON :: torchscript_e2e_test/compilation_failure.py
  TORCH_MLIR_PYTHON :: torchscript_e2e_test/error_reports.py
  TORCH_MLIR_PYTHON :: torchscript_e2e_test/non_tensor_values.py
  TORCH_MLIR_PYTHON :: torchscript_e2e_test/runtime_failure.py
  TORCH_MLIR_PYTHON :: torchscript_e2e_test/submodule.py


Testing Time: 2.89s
  Passed: 6
  Failed: 6

```

silvasean:
What is the error message?

powderluv:
The error is: ```
                                                                                                                                                                                                                                             
********************                                                                                                                 
FAIL: TORCH_MLIR_PYTHON :: torchscript_e2e_test/error_reports.py (11 of 12)                                                          
******************** TEST 'TORCH_MLIR_PYTHON :: torchscript_e2e_test/error_reports.py' FAILED ********************                   
Script:                                                                                                                              
--                                                                                                                                   
[...]

PR #1005, opened 2022-07-03
Fix multi-threaded tests on macOS

powderluv:
Issue is tracked in https://github.com/llvm/torch-mlir/issues/994

### DIFF
--- a/python/torch_mlir_e2e_test/torchscript/framework.py
+++ b/python/torch_mlir_e2e_test/torchscript/framework.py
@@ -321,7 +321,7 @@ def run_workers_in_parallel(task_queue: mp.Queue, worker):
 
     processes = []
     for i in range(NUMBER_OF_PROCESSES):
-        p = mp.Process(target=worker, args=(task_queue, ))
+        p = mp.get_context("fork").Process(target=worker, args=(task_queue, ))
         p.start()
         processes.append(p)
     for i in range(NUMBER_OF_PROCESSES):


python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -332,7 +333,9 @@ def run_workers_in_parallel(task_queue: mp.Queue, worker):
 
 def run_tests(tests: List[Test], config: TestConfig, sequential = False) -> List[TestResult]:
     """Invoke the given `Test`'s with the provided `TestConfig`."""
-    if sequential:
+    # Disable OSX multithreading until is fixed
+    # https://github.com/llvm/torch-mlir/issues/994
+    if sequential or platform.system() == 'Darwin':
```
rdadolf, line 338: It looks like this is a common problem from Python 3.8+ on MacOS when Apple swapped to using spawn instead of fork here ([reference](https://github.com/pytest-dev/pytest-flask/issues/104#issuecomment-577908228)). There's two other solutions:

1. Run `mp.set_start_method('fork')` at init time.
2. Explicitly use fork with `Process`: (line 325) `mp.get_context('fork').Process(...)` ([example in the wild](https://github.com/pytorch/pytorch/pull/36542/files#diff-353ea690c435cf81f51b9dde1494eeec64f99a7d342509ecd891094398974784R2217))

The latter seems like it's a better fit for us.

If we do it this way, we retain the ability to operate in parallel and I don't think we actually need to check for a platform since Windows is the only system that would break with `fork` as the mp default, and we don't support it.

Caveat emptor: I haven't tried this in any kind of exhaustive test whatsoever.

rdadolf (changes requested):
See what you think about this alternative. If it doesn't work, then I'm happy to go ahead with disabling parallel tests.

python/torch_mlir_e2e_test/torchscript/framework.py
powderluv: great suggestion. works like a charm. pushed up a fix. PTAL.

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -24,6 +24,7 @@
 from typing import Any, Callable, List, NamedTuple, Optional, TypeVar, Union, Dict
 
 import traceback
+import platform
```
rdadolf, line 27: nit: no longer need this

python/torch_mlir_e2e_test/torchscript/framework.py
```diff
@@ -321,7 +322,7 @@ def run_workers_in_parallel(task_queue: mp.Queue, worker):
 
     processes = []
     for i in range(NUMBER_OF_PROCESSES):
-        p = mp.Process(target=worker, args=(task_queue, ))
+        p = mp.get_context('fork').Process(target=worker, args=(task_queue, ))
```
rdadolf, line 325: nit: I'm an idiot and used `'`'s instead of `"`'s in my suggestion. Personal style leaked in. Mea culpa.

rdadolf (approved):
Great to hear it works. Minor nits added, but this is good to push. Appreciate you tracking down and testing this.

python/torch_mlir_e2e_test/torchscript/framework.py
powderluv: no worries. updated.

python/torch_mlir_e2e_test/torchscript/framework.py
powderluv: ah yes removed now.

PR #1012, opened 2022-07-05
torch: fold prim.dtype(bf16) to integer constant 15

ashay:
A prior patch (63538de2) that added support for bfloat16 type did not
add the canonicalization pattern to fold `torch.prim.dtype` operations
on bfloat16 tensors into the integer constant 15.  This patch fixes the
problem.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -29,6 +29,9 @@ static int64_t getDtypeIntegerFromMlirType(Type dtype) {
   if (dtype.isa<Float32Type>())
     return 6;
 
+  if (dtype.isa<BFloat16Type>())
+    return 15;
+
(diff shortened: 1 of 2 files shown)


ramiro050 (approved):
LGTM!

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -29,6 +29,9 @@ static int64_t getDtypeIntegerFromMlirType(Type dtype) {
   if (dtype.isa<Float32Type>())
     return 6;
 
+  if (dtype.isa<BFloat16Type>())
```
silvasean, line 32: nit: can we use the torch_upstream constants instead of raw integers throughout this function? (separate patch OK)

ashay: PR https://github.com/llvm/torch-mlir/pull/1020 updates the code to make use of the `ScalarType` enum.

PR #1013, opened 2022-07-05
torch: handle `torch.prim.dtype` ops during type refinement

ashay:
The canonicalizer converts `torch.prim.dtype` ops into integer constants
for valid types, but the type may not be known until type refinement is
complete.  However, type refinement cannot make progress until
`torch.prim.dtype` ops have been resolved to their corresponding integer
constants, thus creating a circular dependency.

This patch creates a tight coupling between type refinement and the
lowering of `torch.prim.dtype` ops by handling such ops as they are
encountered during type refinement.  The unit test in this patch aims to
check whether the type refinement pass can now handle chains of
operations that alternate between type construction and type refinement.

### DIFF
--- a/lib/Dialect/Torch/Transforms/RefineTypes.cpp
+++ b/lib/Dialect/Torch/Transforms/RefineTypes.cpp
@@ -363,6 +363,17 @@ public:
                  ArrayRef<LatticeElement<ValueKnowledge> *> operands) final;
 
 private:
+  // Get the MLIR type of the tensor dtype given the dtype integer value and the
+  // input dtype. When DType is None the type is inferred from the input dtype.
+  void fillInDTypeGivenDTypeIntAndInputDType(ValueKnowledge &knowledge,
+                                             Value dtype, Type inputDType);
+
+  // Get the MLIR type of the tensor dtype given the dtype integer value and
+  // data type of torch type. When DType is None the type is inferred from the
+  // data type.
+  void fillInDTypeGivenDTypeAndDataType(ValueKnowledge &knowledge, Value dtype,
+                                        Type dataType);
+
   /// Incorporates `knowledge` into the lattice state of `v`.
   ///
   /// This method should be used instead of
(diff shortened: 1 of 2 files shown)


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -587,24 +598,39 @@ getPromotedResultTypeAssumingNonZeroRank(MLIRContext *context,
...
+  else if (auto primDtypeOp = dyn_cast<PrimDtypeOp>(dtype.getDefiningOp())) {
+    auto element = getLatticeElement(primDtypeOp.a());
+    assert(!element.isUninitialized() && "no refined type for source");
+
+    switch (element.getValue().kind) {
+    case torch_upstream::TypeKind::TensorType:
+      knowledge.dtype = element.getValue().dtype;
+      break;
+
+    case torch_upstream::TypeKind::NumberType:
```
silvasean, line 618: do we need to handle this case? I don't think it is legal to call prim.dtype on a scalar value.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -587,24 +598,39 @@ getPromotedResultTypeAssumingNonZeroRank(MLIRContext *context,
...
+    ValueKnowledge &knowledge, Value dtype, Type inputDType) {
   assert(isBuiltInType(inputDType) && "`inputDType` must be a builtin type");
   int64_t dtypeInt;
   if (dtype.getType().isa<Torch::NoneType>())
     knowledge.dtype = inputDType;
   else if (matchPattern(dtype, m_TorchConstantInt(&dtypeInt)))
     knowledge.dtype = getTypeForDTypeInteger(dtype.getContext(), dtypeInt);
+  else if (auto primDtypeOp = dyn_cast<PrimDtypeOp>(dtype.getDefiningOp())) {
+    auto element = getLatticeElement(primDtypeOp.a());
+    assert(!element.isUninitialized() && "no refined type for source");
```
silvasean, line 611: is it possible to write a test case to hit this, or is it actually "impossible"?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
ashay: You're right, `prim.dtype` should never have a scalar operand.  I'll remove the case.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
ashay: That's a good point, let me think about it.

ashay: Thanks, that makes sense.  Removing that assertion and the unnecessary cases greatly simplified the code!

lib/Dialect/Torch/Transforms/RefineTypes.cpp
ashay: Removed.

silvasean (approved):
Great!

PR #1022, opened 2022-07-07
[MLIR] Add lowering of split like operations

JakopinA:
A fix of PR #820, removes the merge conflicts.


JakopinA:
@gprateek93 Would you mind taking a look at this?

powderluv:
Please remove "Merge Conflict Fix" in the title

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOpsEarly.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+    auto listType = op.getType().template cast<Torch::ListType>();
+    Location loc = op.getLoc();
+
+    // Calculate the end indices of the slices.
+    SmallVector<int64_t> endIdx;
+    endIdx.push_back(splitSizes[0]);
+    for (unsigned i = 1; i < splitSizes.size(); i++) {
+      endIdx.push_back(splitSizes[i] + endIdx[i - 1]);
+    }
```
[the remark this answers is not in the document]
JakopinA: So replace all of that code with ```partial_sum(splitSizes, splitSizes + 1, endIdx);``` ?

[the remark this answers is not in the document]
ramiro050: The first two arguments are the beginning and end of the range of values to partially sum, so it should be `splitSizes.begin(), splitSizes.end()`. The last argument is the iterator to place the result in, so it should be `endIdx.end()`

lib/Dialect/Torch/Transforms/DecomposeComplexOpsEarly.cpp
```diff
@@ -92,12 +92,8 @@ class DecomposeAtenSplitTensorLikeOp
     auto listType = op.getType().template cast<Torch::ListType>();
     Location loc = op.getLoc();
 
-    auto viewOp = dyn_cast<AtenViewOp>(op.self().getDefiningOp());
-    if (!viewOp)
-      return failure();
-
     auto sizesListOp =
-        dyn_cast<PrimListConstructOp>(viewOp.size().getDefiningOp());
+        dyn_cast<PrimListConstructOp>(op.self().getDefiningOp());
```
ramiro050, line 96: This is wrong. `op.self()` will never be a `PrimListConstructOp` because it is a tensor.

lib/Dialect/Torch/Transforms/DecomposeComplexOpsEarly.cpp
```diff
@@ -114,11 +110,6 @@ class DecomposeAtenSplitTensorLikeOp
       if (!matchPattern(op.getOperand(1), m_TorchConstantInt(&splitSize)))
         return rewriter.notifyMatchFailure(op, "split_size must be constant");
       chunks = (dimSize + splitSize - 1) / splitSize;
-    } else if (isa<AtenChunkOp>(op)) {
-      if (!matchPattern(op.getOperand(1), m_TorchConstantInt(&chunks)))
-        return rewriter.notifyMatchFailure(op,
-                                           "number of chunks must be constant");
-      splitSize = dimSize / chunks;
```
ramiro050: This is not the right way of addressing my comment regarding templates and dynamic casts. There is still a dynamic cast in line https://github.com/llvm/torch-mlir/blob/705e4f698cc42d7a2e18eba6188785d8d91e1284/lib/Dialect/Torch/Transforms/DecomposeComplexOpsEarly.cpp#L109. There is also no longer valid support for the `AtenChunkOp`. You need to also add a decomposition of `AtenChunkOp` into `AtenSplitTensorOp`.

lib/Dialect/Torch/Transforms/DecomposeComplexOpsEarly.cpp
```diff
@@ -33,13 +33,13 @@ class DecomposeAtenSplitSizesOp : public OpRewritePattern<AtenSplitSizesOp> {
     SmallVector<int64_t> splitSizes;
     // TODO: Support non static splitSizes.
     if (!matchPattern(op.split_size(), m_TorchConstantIntList(splitSizes)))
-      return failure();
+      return rewriter.notifyMatchFailure(op, "int list does not exist");
```
ramiro050, line 36: The error message is not accurate. The right thing to say is something along the lines of `op.split_sizes() must be a constant int list`. You can do `git grep m_TorchConstantIntList` to see example error messages other people have used.

ramiro050 (changes requested):
A couple of the changes in the latest commit should've made the e2e tests fail (which you can run by doing `python -m e2e_testing.torchscript.main -f "Split|Chunk" -v`). Please make sure the e2e tests added by your PR are passing locally in your machine before requesting a review. I'm happy to help with questions.

JakopinA:
I'm unsure why right now, but the tests are not failing because of the changes I made, in fact every single test fails at the moment.

ramiro050:
> I'm unsure why right now, but the tests are not failing because of the changes I made, in fact every single test fails at the moment.

What errors are you getting?

JakopinA:
When running the torchscript file with verbose, I get ```Runtime error: Testing process terminated. Either the compiler crashed or the compiled code crashed at runtime.``` with every single test, not just the split chunk tests.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -1047,6 +1047,26 @@ def aten〇index〇Tensor(self: List[int], indices: List[Optional[List[int]]]) -
...
+
+def aten〇split_with_sizes(self: List[int], split_sizes: List[int], dim: int = 0) -> List[List[int]]:
+    return upstream_shape_helpers.split(self, split_sizes, dim)
+
+def aten〇split〇Tensor(self: List[int], split_size: int, dim: int = 0) -> List[List[int]]:
+    dim_size = self[dim]
+    chunks = (dim_size + split_size - 1) // split_size
+    split_sizes = [split_size for i in range(chunks)]
+    split_sizes[chunks - 1] = split_size - (split_size * chunks - dim_size)
+    return upstream_shape_helpers.split(self, split_sizes, dim)
```
ramiro050, line 1064: These shape functions are outdated. The module `upstream_shape_helpers` should no longer be used in torch-mlir. See: https://github.com/llvm/torch-mlir/pull/1019. The shape functions should be upstreamed into pytorch, and in this file you should use the corresponding `upstream_shape_functions.` function.

I think this is causing the tests to crash. When I try running a simple lowering I get the following error:

```
-:6600:10: error: 'func.call' op '__torch__.torch_mlir.dialects.torch.importer.jit_ir.build_tools.upstream_shape_helpers.maybe_wrap_dim' does not reference a valid function
-:6600:10: note: see current operation: %4 = "func.call"(%arg2, %3, %2) {callee = @__torch__.torch_mlir.dialects.torch.importer.jit_ir.build_tools.upstream_shape_helpers.maybe_wrap_dim} : (!torch.int, !torch.int, !torch.bool) -> !torch.int
```

qedawkins:
I've been working with this PR and am wondering about the choice to introduce a `DecomposeComplexOpsEarlyPass`. In particular, when trying to compile pytorch-v-diffusion, we end up with the following error:
```
error: failed to legalize operation 'torch.aten._softmax' that was explicitly marked illegal
note: see current operation: %2414 = "torch.aten._softmax"(%2413, %80, %71) : (!torch.vtensor<[2,8,256,256],unk>, !torch.int, !torch.bool) -> !torch.vtensor<[2,8,256,256],unk>
```

Which is thrown because `torch.aten._softmax` fails to decompose due to having an unknown dtype. If we look at the IR for this we see that the loss of shape/type information happens around the `torch.prim.ListUnpack` op at %1410.
```
  %1406 = torch.aten.slice.Tensor %1405, %int1, %int0, %int8, %int1 : !torch.tensor<[2,24,256,64],f32>, !torch.int, !torch.int, !torch.int, !torch.int -> !torch.tensor<[2,8,256,64],f32>
  %1407 = torch.aten.slice.Tensor %1405, %int1, %int8, %int16, %int1 : !torch.tensor<[2,24,256,64],f32>, !torch.int, !torch.int, !torch.int, !torch.int -> !torch.tensor<[2,8,256,64],f32>
  %1408 = torch.aten.slice.Tensor %1405, %int1, %int16, %int24, %int1 : !torch.tensor<[2,24,256,64],f32>, !torch.int, !torch.int, !torch.int, !torch.int -> !torch.tensor<[2,8,256,64],f32>
  %1409 = torch.prim.ListConstruct %1406, %1407, %1408 : (!torch.tensor<[2,8,256,64],f32>, !torch.tensor<[2,8,256,64],f32>, !torch.tensor<[2,8,256,64],f32>) -> !torch.list<tensor>
-------------------------------------------------------------------------
  %1410:3 = torch.prim.ListUnpack %1409 : !torch.list<tensor> -> !torch.tensor, !torch.tensor, !torch.tensor
  %1411 = torch.copy.to_vtensor %1410#0 : !torch.vtensor
  %1412 = torch.aten.mul.Tensor %1411, %67 : !torch.vtensor, !torch.vtensor<[],f64> -> !torch.vtensor
  %1413 = torch.aten.transpose.int %1410#1, %int2, %int3 : !torch.tensor, !torch.int, !torch.int -> !torch.tensor
  %1414 = torch.copy.to_vtensor %1413 : !torch.vtensor
  %1415 = torch.aten.mul.Tensor %1414, %67 : !torch.vtensor, !torch.vtensor<[],f64> -> !torch.vtensor
  %1416 = torch.aten.expand %1412, %1101, %false : !torch.vtensor, !torch.list<int>, !torch.bool -> !torch.vtensor<[?,?,?,?],unk>
  %1417 = torch.aten._reshape_alias %1416, %1103, %1104 : !torch.vtensor<[?,?,?,?],unk>, !torch.list<int>, !torch.list<int> -> !torch.vtensor<[16,256,64],unk>
  %1418 = torch.aten.expand %1415, %1106, %false : !torch.vtensor, !torch.list<int>, !torch.bool -> !torch.vtensor<[?,?,?,?],unk>
  %1419 = torch.aten._reshape_alias %1418, %1108, %1109 : !torch.vtensor<[?,?,?,?],unk>, !torch.list<int>, !torch.list<int> -> !torch.vtensor<[16,64,256],unk>
  %1420 = torch.aten.bmm %1417, %1419 : !torch.vtensor<[16,256,64],unk>, !torch.vtensor<[16,64,256],unk> -> !torch.vtensor<[16,256,256],unk>
  %1421 = torch.aten._unsafe_view %1420, %1112 : !torch.vtensor<[16,256,256],unk>, !torch.list<int> -> !torch.vtensor<[2,8,256,256],unk>
  %1422 = torch.aten._softmax %1421, %int3, %false : !torch.vtensor<[2,8,256,256],unk>, !torch.int, !torch.bool -> !torch.vtensor<[2,8,256,256],unk>

```

Note that %1409 is the last result in the decomposition of an `aten.split_with_sizes` op and %1410 is from the model (it just unpacks the result of the split). I'm not sure if this is a problem caused by the early decomposition introducing a ListConstruct op that we can't handle or if this is a case where we aren't able to handle shape propagation for `torch.prim.ListUnpack`.

My understanding is that the approach of decomposing early was to avoid passing the split-like ops through the shape refinement pass due to returning a list of tensors. Comments from @silvasean or @ramiro050 would be greatly appreciated. If necessary I can open an issue with a MWE; I avoided it in this case because I thought this might be specific to the decomposition strategy used in this PR and thus shouldn't be considered an issue yet.

Issue #999, opened 2022-06-30
[RFC] Adding Torch to MHLO conversion


ZihengJiang:
Hello everyone, we are from the AML (Applied Machine Learning) team at Bytedance. While the TorchMLIR project provides a set of MLIR dialect for Torch, which bridges PyTorch and MLIR ecosystems, in our development, we also have observed interest in supporting the MHLO dialect as a lowering target, which is still missing in the project.
In the RFC, we propose to add conversation from Torch dialect to MHLO (Meta HLO Dialect). The MHLO dialect can be found here (https://github.com/tensorflow/mlir-hlo) and the corresponding operation definition is here (https://www.tensorflow.org/mlir/hlo_ops). We would add this conversion as an independent pass and this could provide an extra path where torch mlir can also be incorporated into MHLO dialect, and thus enrich the applications of TorchMLIR.

## Motivation:
The HLO (High Level Optimizer) IR offers a carefully fixed selected list of operations. MHLO supports a HLO-like compilation pipeline using MLIR and provides a uniform interface to compile and execute these optimized HLO programs with dynamic shape support. 
While the MHLO dialect supports Tensorflow models by nature, if we have the conversion pass from Torch dialect to MHLO, we could convert and lower those different frontend models into a unified IR. This would reduce engineering effort and ease the subsequent backend work.

## Proposal:
- Add mlir-hlo(https://github.com/tensorflow/mlir-hlo) as a dependent submodule.
- Add the conversion as passes within the existing TorchMLIR codebase, e.g. lib/Conversion/TorchToMhlo/*
- Add the new conversion pipeline into https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/TorchConversion/Transforms/Passes.cpp 
- Tests go into test/Conversion/TorchToMhlo/*
- Make the DecomposeComplexOps to be configurable, since different backend may require different decomposition rules.

## Benefits:
- Bridging Torch dialect with MHLO dialect.
- Enabling user to convert and lower those different frontend models (PyTorch, TensorFlow, etc.) into the unified MHLO IR. This would reduce engineering effort and ease the subsequent backend works.
- Provide more accessibility of Torch to other MLIR based projects with this extra path from Torch to MHLO 

## Status:
We have already implemented the whole lowering pipeline and operators conversion to covert BERT and ResNet inference. See the POC in the PR: https://github.com/llvm/torch-mlir/pull/1025 . Welcome everyone interested being part of this effort as well.

[...]

PR #1025, opened 2022-07-07
[MHLO] Init Torch to MHLO conversion.

ZihengJiang:

TODO:
- [ ] Remove dependency for FuncTorch in examples
- [ ] Update BERT and ResNet example
- [ ] Configurable Decomposition
- [ ] Clean and format

Co-authored-by: @byronyi @Vremold Xuanrun Zhang

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "external/llvm-project"]
 	path = externals/llvm-project
 	url = https://github.com/llvm/llvm-project.git
+[submodule "externals/mlir-hlo"]
+	path = externals/mlir-hlo
+	url = https://github.com/tensorflow/mlir-hlo.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,7 +41,13 @@ torch_mlir_add_llvm_external_project(
   TORCH_MLIR_DIALECTS
   ${CMAKE_CURRENT_SOURCE_DIR}/externals/llvm-external-projects/torch-mlir-dialects)
 
+torch_mlir_add_llvm_external_project(
+  mlir-hlo
+  MLIR_HLO
+  ${CMAKE_CURRENT_SOURCE_DIR}/externals/mlir-hlo)
+
 if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+  message(STATUS "Torch-MLIR out-of-tree build.")
   # Out-of-tree build
 
   #-------------------------------------------------------------------------------
@@ -81,7 +87,16 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
   set(TORCH-MLIR_BUILT_STANDALONE 1)
   set(BACKEND_PACKAGE_STRING "LLVM ${LLVM_PACKAGE_VERSION}")
   add_subdirectory(externals/llvm-external-projects/torch-mlir-dialects)
+
+  set(MHLO_BUILD_EMBEDDED ON)
+  add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/externals/mlir-hlo
+                   ${CMAKE_CURRENT_BINARY_DIR}/mlir-hlo
+                   EXCLUDE_FROM_ALL)
+  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/mlir-hlo/include)
+  include_directories(${CMAKE_CURRENT_BINARY_DIR}/mlir-hlo/include)
+  include_directories(${CMAKE_CURRENT_BINARY_DIR})
 else()
+  message(STATUS "Torch-MLIR in-tree build.")
   # In-tree build with LLVM_EXTERNAL_PROJECTS=torch-mlir
   # FIXME: This should really be inherited from the LLVM tree.  In particular,
   # it's going to change when cross-compiling.
@@ -95,6 +110,9 @@ else()
   set(MLIR_INCLUDE_DIR ${LLVM_MAIN_SRC_DIR}/../mlir/include)
   set(MLIR_GENERATED_INCLUDE_DIR ${LLVM_BINARY_DIR}/tools/mlir/include)
   set(MLIR_INCLUDE_DIRS "${MLIR_INCLUDE_DIR};${MLIR_GENERATED_INCLUDE_DIR}")
+  # since mhlo didn't set INTERFACE_DIRECTORIES for their target, we need include mhlo directories globally
+  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/mlir-hlo/include)
+  include_directories(${LLVM_BINARY_DIR}/tools/mlir_hlo/include)
 endif()
 
 set(TORCH_MLIR_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")
--- a/lib/Conversion/TorchToMhlo/BasicOp.cpp
+++ b/lib/Conversion/TorchToMhlo/BasicOp.cpp
@@ -1,0 +1,1022 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+
+#include "torch-mlir/Conversion/TorchToMhlo/TorchToMhlo.h"
+
+#include "../PassDetail.h"
+#include "./MhloLegalizeUtils.h"
+#include "./PopulatePattern.h"
+#include "mlir-hlo/Dialect/mhlo/IR/hlo_ops.h"
+#include "torch-mlir/Conversion/Utils/Utils.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchDialect.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
+#include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+#include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h"
+#include <iostream>
+#include <numeric>
+
+using namespace mlir;
+using namespace mlir::torch;
+using namespace mlir::torch::Torch;
+
+static llvm::Optional<Value> squeeze(Operation *op, Value input,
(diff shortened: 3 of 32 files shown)


scripts/build.sh
```diff
@@ -0,0 +1,33 @@
+# in-tree build
```
powderluv, line 1: probably the scripts in scripts/ are meant to be local scripts and not checked in ?

CMakeLists.txt
```diff
@@ -81,7 +87,16 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
...
   set(BACKEND_PACKAGE_STRING "LLVM ${LLVM_PACKAGE_VERSION}")
   add_subdirectory(externals/llvm-external-projects/torch-mlir-dialects)
+
+  set(MHLO_BUILD_EMBEDDED ON)
+  add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/externals/mlir-hlo
+                   ${CMAKE_CURRENT_BINARY_DIR}/mlir-hlo
+                   EXCLUDE_FROM_ALL)
+  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/mlir-hlo/include)
+  include_directories(${CMAKE_CURRENT_BINARY_DIR}/mlir-hlo/include)
+  include_directories(${CMAKE_CURRENT_BINARY_DIR})
```
powderluv, line 97: is the whole CMAKE_CURRENT_BINARY_DIR required since we are adding it globally ?

.gitmodules
```diff
@@ -1,3 +1,6 @@
 [submodule "external/llvm-project"]
 	path = externals/llvm-project
 	url = https://github.com/llvm/llvm-project.git
```
powderluv, line 3: we probably should land the llvm update and even mhlo as submodules first.

CMakeLists.txt
```diff
@@ -41,7 +41,13 @@ torch_mlir_add_llvm_external_project(
   TORCH_MLIR_DIALECTS
   ${CMAKE_CURRENT_SOURCE_DIR}/externals/llvm-external-projects/torch-mlir-dialects)
 
+torch_mlir_add_llvm_external_project(
```
powderluv, line 44: Can we add a top level TORCH_MLIR_MHLO (name can be anything @silvasean any suggestions ?) CMake flag that can enable / disable the MHLO backend ? This can be the big hammer in case we have to disable it for any reason (broken on macOS etc) .

powderluv (commented):
Thank you. Drive by comments. I know it is only a draft PR but wanted to get comments early.

silvasean:
One thing I would like is if we can add this to our e2e test suite, even if only a few tests pass. We can model it on the Tosa support

Add a case here:
https://github.com/llvm/torch-mlir/blob/f202ae00129e5f03b14e3d04d63e3eca4da17ed5/e2e_testing/torchscript/main.py#L77

Other related files:
python/torch_mlir_e2e_test/torchscript/configs/tosa_backend.py
python/torch_mlir_e2e_test/tosa_backends/linalg_on_tensors.py -- will require using MHLOToLinalg lowerings
e2e_testing/torchscript/xfail_sets.py - add list of expected tests to pass

lib/Conversion/TorchToMhlo/MhloLegalizeUtils.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+  bool do_bcast = (inShape.size() != outShape.size());
+  SmallVector<int64_t> bcastDims;
+  for (size_t i = 0; i < inShape.size(); ++i) {
+    // iterating over the dimension sizes, starting at the trailing dimension
+    size_t outPos = outShape.size() - 1 - i;
+    size_t inPos = inShape.size() - 1 - i;
+    int64_t outDim = outShape[outPos];
+    int64_t inDim = inShape[inPos];
+    if (inDim == outDim) {
```
fortianyou, line 295: The implementation doesn't support implicit broadcast when `inDim == outDim == -1` both unknown?

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -0,0 +1,1022 @@
...
+        rhsTensor = rhs;
+      }
+    }
+
+    // Handle broadcasting. Since we have the output type already, here we
+    // just broodcast operands' shape to output shape.
+    lhsTensor = mhlo::promoteAndBroadcast(rewriter, lhsTensor, outType);
+    rhsTensor = mhlo::promoteAndBroadcast(rewriter, rhsTensor, outType);
+
+    rewriter.replaceOpWithNewOp<mhlo::MulOp>(op, outType, lhsTensor, rhsTensor);
```
fortianyou, line 615: Use chlo::BroadcastMulOp to support implicit/dynamic broadcast

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -0,0 +1,1022 @@
...
+    } else {
+      rhsTensor = rhs;
+    }
+
+    // Handle broadcasting. Since we have the output type already, here we
+    // just broodcast operands' shape to output shape.
+    lhsTensor = mhlo::promoteAndBroadcast(rewriter, lhsTensor, outType);
+    rhsTensor = mhlo::promoteAndBroadcast(rewriter, rhsTensor, outType);
+
+    rewriter.replaceOpWithNewOp<mhlo::DivOp>(op, outType, lhsTensor, rhsTensor);
```
fortianyou, line 663: ditto

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -0,0 +1,1022 @@
...
+  INSERT_UNARY_FPONLY_PATTERN(AtenLogOp, mhlo::LogOp)
+  INSERT_UNARY_FPONLY_PATTERN(AtenExpOp, mhlo::ExpOp)
+  INSERT_UNARY_FPONLY_PATTERN(AtenCloneOp, mhlo::CopyOp)
+  INSERT_UNARY_FPONLY_PATTERN(AtenSqrtOp, mhlo::SqrtOp)
+#undef INSERT_UNARY_FPONLY_PATTERN
+
+#define INSERT_BINARY_ADDSUB_PATTERN(AtenOp, MhloOp)                           \
+  target.addIllegalOp<AtenOp>();                                               \
+  patterns.add<ConvertAtenAddSubOp<AtenOp, MhloOp>>(typeConverter, context);
+  INSERT_BINARY_ADDSUB_PATTERN(AtenAddTensorOp, mhlo::AddOp)
```
fortianyou, line 972: Use chlo::Broadcast[Add]SubOp to support implicit/dynamic broadcast

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -0,0 +1,1022 @@
...
+               std::is_same<AtenOpT, AtenEqScalarOp>()) {
+      compareDirectionAttr = mhlo::ComparisonDirectionAttr::get(
+          op->getContext(), mhlo::ComparisonDirection::EQ);
+    } else if (std::is_same<AtenOpT, AtenNeTensorOp>() ||
+               std::is_same<AtenOpT, AtenNeScalarOp>()) {
+      compareDirectionAttr = mhlo::ComparisonDirectionAttr::get(
+          op->getContext(), mhlo::ComparisonDirection::NE);
+    }
+
+    rewriter.replaceOpWithNewOp<mhlo::CompareOp>(
```
fortianyou, line 151: nit: chlo::BroadcastCompareOp

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -0,0 +1,1022 @@
...
+  RankedTensorType resultType = getTypeConverter()
+                                    ->convertType(op->getResult(0).getType())
+                                    .cast<RankedTensorType>();
+
+  int64_t dimInt;
+  if (!matchPattern(op.dim(), m_TorchConstantInt(&dimInt)))
+    return op->emitError("unimplemented: dim is not constant");
+  uint64_t batchDims = 0;
+
+  rewriter.replaceOpWithNewOp<mhlo::TorchIndexSelectOp>(
```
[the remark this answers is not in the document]
burmako: `mhlo::TorchIndexSelect` was added to simplify lowering `tf.GatherV2` to MHLO in [this commit](https://github.com/tensorflow/tensorflow/commit/18e1ec8a2a127f32a95a6d62897147db1a0017c5) 2.5 years ago. It is modelled after [the client HLO API `TorchIndexSelect`](https://github.com/tensorflow/tensorflow/blob/27465612d3cbab9cc4c1c28735a9b2fa2f391d5c/tensorflow/compiler/xla/client/lib/slicing.cc#L266-L311).

I agree that it doesn't look particularly essential for the MHLO dialect. We've been recently thinking about moving it to CHLO, along with a few other ops which are modelled after client HLO APIs which don't correspond to dedicated HLO ops. (More specifically, these other ops are: `mhlo.broadcast` - keeping broadcast_in_dim in MHLO, `mhlo.create_token`, `mhlo.cross_replica_sum`, `mhlo.dot` - keeping dot_general in MHLO, `mhlo.einsum`, `mhlo.unary_einsum`).

scripts/build.sh
ZihengJiang: Sounds good. I left it here as the example to build the project. Will remove it in the end.

silvasean:
I just found a better reference on coding style for us: https://mlir.llvm.org/getting_started/DeveloperGuide/

lib/Conversion/TorchToMhlo/BasicOp.cpp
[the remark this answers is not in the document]
Vremold: Thanks for your advice, we will change this part's implementation in later commits.

CMakeLists.txt
Vremold: Yeah, it's quite a good suggestion. But it seems our codes are integrated with torch-mlir a little bit tightly. For example, we registered the torch-to-mhlo conversion pass in [/include/torch-mlir/Conversion/Passes.td](https://github.com/llvm/torch-mlir/pull/1025/commits/a20226b6f824d5a1aca64aeec79bd0443eaf4fa8#diff-e378cdef89f6a7ada41d204c78c1580725d8b5bb0d59dfb37b34cf0861fb5361). Do you have any good idea to separate them？

ZihengJiang:
@silvasean @powderluv @fortianyou @Yancey1989 

Thanks to everyone who reviews here. According to the offline meeting between Bytedance and Alibaba. We decide to break the proposal into several PRs:

- Integrate MHLO into TorchMLIR Repo
- Basic conversion pipeline
- Operator conversions
- ResNet and BERT examples

We will address the issue mentions above in the following PRs and track the progress in the RFC https://github.com/llvm/torch-mlir/issues/999

PR #1026, opened 2022-07-08
Add facebook's dlrm model heavy_dep test

vid-999:


python/torch_mlir_e2e_test/test_suite/__init__.py
```diff
@@ -15,6 +15,7 @@
     "ConvolutionModule1D_basic",
     "MaxPool2dWith3dInputModule_basic",
     "MaxPool2dWithIndicesWith3dInputModule_basic",
+    "Dlrm_basic",
```
pashu123, line 18: We don't want to add this here since the tests in the heavydep tests are manual and don't trigger CI.

vid-999: I see, where should it be called then?
Also, this fails when lowering to TORCH, ill make a test for that but where should that be called?

vid-999: should I do an import in main.py in the folder torchscript_e2e_heavydep_tests ?
from . import facebook_dlrm

silvasean: I don't think we have a good solution for this currently :/

vid-999: i see, I removed it from this file for now.

powderluv:
@rdadolf fyi for sparse tensors

build_tools/torchscript_e2e_heavydep_tests/facebook_dlrm.py
```diff
@@ -0,0 +1,839 @@
...
+    def create_emb(self, m, ln, weighted_pooling=None):
+        emb_l = nn.ModuleList()
+        v_W_l = []
+        for i in range(0, ln.size):
+
+            n = ln[i]
+            n = n.item()
+
+            if self.qr_flag and n > self.qr_threshold:
+                EE = QREmbeddingBag(
```
silvasean, line 53: where is this imported from?

build_tools/torchscript_e2e_heavydep_tests/facebook_dlrm.py
```diff
@@ -0,0 +1,839 @@
...
+import torch.nn as nn
+from typing import Dict, List
+from numpy import random as ra
+from torch.utils.data import Dataset
+
+from torch_mlir_e2e_test.torchscript.framework import TestUtils
+from torch_mlir_e2e_test.torchscript.registry import register_test_case
+from torch_mlir_e2e_test.torchscript.annotations import annotate_args, export
+
+class DLRM_Net(nn.Module):
```
silvasean, line 13: can we please use an existing library rather than open-coding this ourselves? We want this to be representative of what users would see.

silvasean (changes requested):

build_tools/torchscript_e2e_heavydep_tests/facebook_dlrm.py
rdadolf: IIRC, this is more or less lifted from the existing reference implementation, no? https://github.com/facebookresearch/dlrm/blob/main/dlrm_s_pytorch.py

As an aside: that probably means it should really be licensed accordingly (i.e., (c) facebook, MIT license), since it's derived code.

silvasean: I'm not specifically familiar, but yes, the code does look extremely similar.

We should use that library since that is what users will presumably be using. We can add that library as a dependency here: https://github.com/llvm/torch-mlir/blob/5bd9362c61828870cf3c4097c718ea34f475b11a/build_tools/torchscript_e2e_heavydep_tests/generate_serialized_tests.sh#L26

vid-999: yes, its taken from the facebook dlrm repo: https://github.com/facebookresearch/dlrm
Apologies, will add the correct license.

build_tools/torchscript_e2e_heavydep_tests/facebook_dlrm.py
vid-999: actually just realized that code isn't even triggered while tracing... so maybe I should remove that or add another test case that triggers that control flow?
its imported from here in the original file: from tricks.qr_embedding_bag import QREmbeddingBag

build_tools/torchscript_e2e_heavydep_tests/facebook_dlrm.py
silvasean: Can you please depend on that DLRM library instead of copying their code?

vid-999: which library ?

silvasean: can you pip install https://github.com/facebookresearch/dlrm ?

vid-999: not that I am aware of. doesn't say on their README.

powderluv: I think @silvasean means can we `pip install facebookdlrm` or equivalent.  I don't think they have a setup.py even just a bunch of .py scripts. I tried to `pip install -e git+https://...` but it wasn't happy without a setup.py

vid-999: I see, also @silvasean @powderluv  I had to make many changes in their source code to be able to get TorchScript via torch.jit.trace. So I don't know how that would work even if it was pip install-able ...

powderluv: I think their code in https://github.com/facebookresearch/dlrm  is meant to be the reference DLRM for the industry since each vendor had different variants (https://github.com/NVIDIA/DeepLearningExamples/tree/master/PyTorch/Recommendation/DLRM and https://github.com/jithunnair-amd/dlrm). Unfortunately the reference DLRM code doesn't lower cleanly without some changes to the model and the authors suggest we make such changes https://github.com/facebookresearch/dlrm/issues/132#issuecomment-708902880

I have been pushing the team to upstream model variants with fixes (just like HuggingFace carries lowering model fixes) into torch-mlir e2e tests but if we want to only use "stock" models that is ok too. We can carry the fixed up / lowerable models downstream in SHARK model/tests suites and report reproductions of issues upstream.

silvasean: My main concern is that I want this to be representative of how a user would actually want to use DLRM via Torch-MLIR -- would they really copy-paste that reference code? It's not really Torch-MLIR's job to vendor a DLRM implementation, let alone having such a vendored implementation only as part of our test-suite and not even exposed to users (requiring yet another copy-paste telephone game).

vid-999: closing this issue and move dlrm to SHARK model/tests suites

PR #1028, opened 2022-07-08
Only set Torch_DIR when TORCH_MLIR_USE_INSTALLED_PYTORCH is off

henrytwo:
Previously, `Torch_DIR` was set regardless of whether `TORCH_MLIR_USE_INSTALLED_PYTORCH` was on or off, which caused some linkage issues with Lazy Tensor Core. This PR modifies the build to only set this variable when `TORCH_MLIR_USE_INSTALLED_PYTORCH` is off.

cc: @antoniojkim

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/CMakeLists.txt
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/CMakeLists.txt
@@ -4,12 +4,13 @@
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 include(TorchMLIRPyTorch)
-set(Torch_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../../../../../../libtorch/share/cmake/Torch")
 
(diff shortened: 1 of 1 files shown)


henrytwo:
While we're on this topic, IIUC we want to be able to toggle between a local PyTorch build or libtorch. For Lazy Tensor Core we also have a PyTorch dependency, so we end up duplicating some of the PyTorch setup in the cmakefile over there too.

Should we have some `TorchMLIRConfigureTorch()` function that calls the logic to check whether `TorchMLIRConfigurePyTorch` or `TorchMLIRConfigureLibTorch` should be used? This should reduce some duplicated code

silvasean:
I think @powderluv is using the libtorch build to iterate on the new Python packaging stuff, but the main flow today AFAIU is the `TORCH_MLIR_USE_INSTALLED_PYTORCH=TRUE` one.

@powderluv is there a reason we can't just delete the libtorch stuff until it is ready to avoid this kind of confusion?

silvasean (approved):

powderluv:
Thanks for the fix. Given the `torchgen` requirement for LTC I plan to simplify / remove the libtorch only build. We will have a src build so we are aligned to a PyTorch version.

PR #1029, opened 2022-07-08
torch_mlir.compile: Allow OutputType as a string.

silvasean:
A lot of code was super verbose with `torch_mlir.OutputType.XYZ`. Now,
you can simply do `"xyz"`. I updated a few examples.


ramiro050 (approved):
LGTM!

PR #1030, opened 2022-07-09
canonicalizer: propagate type information across copy and cast ops

ashay:
Prior to this patch, the canonicalizer for `AtenSizeOp` succeeded only
if the operand's type information included the size of all dimensions.
We can extend the set of optimizable cases by propagating types
across operations whose result type matches the input tensor type.

Specifically, this patch enables the canonicalizer for `AtenSizeOp` to
see past `tensor_static_info_cast`, `copy.to_vtensor`, and
`copy.to_tensor` ops until it reaches the first op whose result type
contains size information for all dimensions.  All other encountered ops
cause the canonicalizer to give up.


ashay:
I am not entirely certain if this patch is kosher, since it uses a heuristic.  If there's a better way to accomplish the same result, please let me know.  Thanks!

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -819,14 +819,33 @@ void AtenAddTensorOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+static FailureOr<BaseTensorType> traceKnownSizeTensorType(Value value) {
+  Type type = value.getType();
+  if (!type.isa<BaseTensorType>())
+    return failure();
+
+  auto baseTensorType = type.cast<BaseTensorType>();
+  if (baseTensorType.areAllSizesKnown())
+    return baseTensorType;
+
+  auto op = value.getDefiningOp();
```
silvasean, line 831: please write this as a while loop, and bail out after "6" (yes, hardcoded ok) iterations. We cannot have an unbounded IR traversal since it can create superlinear compilation time.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -819,14 +819,33 @@ void AtenAddTensorOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
 // AtenSizeOp
 //===----------------------------------------------------------------------===//
 
+static FailureOr<BaseTensorType> traceKnownSizeTensorType(Value value) {
```
silvasean, line 822: Can we generalize this a bit and share it with aten.size.int?

test/Dialect/Torch/canonicalize.mlir
```diff
@@ -1328,3 +1328,18 @@ func.func @torch.aten.add.Tensor$canonicalize_literal_0d() -> !torch.vtensor<[],
     %2 = torch.aten.add.Tensor %0, %1, %int3 : !torch.vtensor<[],si64>, !torch.vtensor<[],si64>, !torch.int -> !torch.vtensor<[],si64>
     return %2 : !torch.vtensor<[],si64>
 }
+
+// CHECK-LABEL:   @torch.aten.size$copy(
+// CHECK-SAME:        %[[ARG:.*]]: !torch.vtensor<[2,3],f32>) -> !torch.list<int> {
+// CHECK:           %[[two:.*]] = torch.constant.int 2
```
silvasean, line 1334: uppercase variable names for consistentcy

silvasean (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
ashay: Ah right, I had not thought of the effect on compilation times due to deep hierarchies.  Fixed.

lib/Dialect/Torch/IR/TorchOps.cpp
ashay: Updated the function to accept `Optional<int64_t>` for the dimension -- if the value is `None`, then the function checks whether all dimension's sizes are known.

test/Dialect/Torch/canonicalize.mlir
ashay: Makes sense.  Thanks, fixed!

ashay:
Rebased with main

PR #1031, opened 2022-07-10
Speed up building libtorch

makslevental:
Speed up building libtorch with some small tricks.


build_tools/build_libtorch.sh
```diff
@@ -102,19 +104,39 @@ fi
...
 checkout_pytorch() {
   if [[ ! -d "$PYTORCH_ROOT" ]]; then
-    git clone https://github.com/pytorch/pytorch "$PYTORCH_ROOT"
+    git clone --depth 1 --single-branch --branch "${PYTORCH_BRANCH}" https://github.com/pytorch/pytorch "$PYTORCH_ROOT"
   fi
   cd "$PYTORCH_ROOT"
-  git fetch --all
-  git checkout origin/"${PYTORCH_BRANCH}"
-  git submodule update --init --recursive
+  for dep in protobuf pocketfft cpuinfo FP16 psimd fmt sleef pybind11 onnx flatbuffers foxi; do
```
makslevental, line 110: minimal number of third party deps (since we're turning off all build flags anyway)

build_tools/build_libtorch.sh
```diff
@@ -102,19 +104,39 @@ fi
...
 }
 
 build_pytorch() {
   cd "$PYTORCH_ROOT"
   # Uncomment the next line if you want to iterate on source builds
-  python setup.py clean
-
+  # python setup.py clean
+
+  sed -i.bak -E 's/^[[:space:]]+check_submodules()/#check_submodules()/g' setup.py
```
makslevental, line 120: `setup.py` will pull submodules even if we don't want them (selective clone above)

build_tools/build_libtorch.sh
```diff
@@ -102,19 +104,39 @@ fi
...
+  # remove after https://github.com/pytorch/pytorch/issues/81178 is fixed
```
makslevental, line 121: dirty dirty hack but necessary until https://github.com/pytorch/pytorch/issues/81178 is resolved

build_tools/build_libtorch.sh
```diff
@@ -102,19 +104,39 @@ fi
...
+  sed -i.bak -E 's/namespace ao {/namespace ao {\n#include <ATen\/native\/ao_sparse\/quantized\/cpu\/packed_params.h>/g' aten/src/ATen/native/ao_sparse/quantized/cpu/qlinear_deserialize.cpp
+  sed -i.bak -E 's/namespace ao {/namespace ao {\n#include <ATen\/native\/ao_sparse\/quantized\/cpu\/packed_params.h>/g' aten/src/ATen/native/ao_sparse/quantized/cpu/qlinear_serialize.cpp
+
+  if ! command -v ninja &> /dev/null
+  then
+    echo "You should probably install Ninja using pip install ninja"
+    USE_NINJA=0
```
makslevental, line 128: I'm not 100% sure if this works - it used to be necessary but now supposedly isn't but still appears in some places (e.g. [here](https://github.com/pytorch/pytorch/blob/c3e089a0472f8ea1d6960d35297fd9c35f260d2d/tools/setup_helpers/cmake.py#L31))

build_tools/build_libtorch.sh
```diff
@@ -127,11 +149,38 @@ build_pytorch() {
...
+  USE_OPENMP=OFF \
+  USE_PYTORCH_QNNPACK=OFF \
+  USE_QNNPACK=OFF \
+  USE_XNNPACK=OFF \
+  python setup.py bdist_wheel -d "$WHEELHOUSE"
 }
 
 package_pytorch() {
-  mkdir -p libtorch/{lib,bin,include,share}
+  rm -rf libtorch/{lib,bin,include,share}
```
makslevental, line 183: I'm not sure about this - on my M1 the `mv` commands below failed because these directories already existed.

build_tools/build_libtorch.sh
```diff
@@ -102,19 +104,39 @@ fi
...
   cd "$PYTORCH_ROOT"
   # Uncomment the next line if you want to iterate on source builds
-  python setup.py clean
-
+  # python setup.py clean
+
+  sed -i.bak -E 's/^[[:space:]]+check_submodules()/#check_submodules()/g' setup.py
+  # remove after https://github.com/pytorch/pytorch/issues/81178 is fixed
+  sed -i.bak -E 's/namespace ao {/namespace ao {\n#include <ATen\/native\/ao_sparse\/quantized\/cpu\/packed_params.h>/g' aten/src/ATen/native/ao_sparse/quantized/cpu/qlinear_deserialize.cpp
+  sed -i.bak -E 's/namespace ao {/namespace ao {\n#include <ATen\/native\/ao_sparse\/quantized\/cpu\/packed_params.h>/g' aten/src/ATen/native/ao_sparse/quantized/cpu/qlinear_serialize.cpp
```
makslevental, line 123: dirty dirty hack but necessary until https://github.com/pytorch/pytorch/issues/81178 is resolved

build_tools/build_libtorch.sh
powderluv: you probably want to clean up with a `git clean` before submodule updates for rebuilds and do a sed again.

build_tools/build_libtorch.sh
powderluv: I moved this from `cp -rv` to `mv` here so the GH runners don't take a long time. Just FYI on why it is different from upstream.

python/CMakeLists.txt
```diff
@@ -80,6 +80,8 @@ if(TORCH_MLIR_ENABLE_JIT_IR_IMPORTER)
       LIBTORCH_CACHE=${LIBTORCH_CACHE}
       LIBTORCH_VARIANT=${LIBTORCH_VARIANT}
       CMAKE_OSX_ARCHITECTURES=${CMAKE_OSX_ARCHITECTURES}
+      CMAKE_C_COMPILER_LAUNCHER=${CMAKE_C_COMPILER_LAUNCHER}
+      CMAKE_CXX_COMPILER_LAUNCHER=${CMAKE_CXX_COMPILER_LAUNCHER}
```
powderluv, line 84: I had some trouble testing passing the compiler flags from outside the build_libtorch.sh build and that is the reason I have a PT_C_COMPILER / PT_CXX_COMPILER. If this works that would be great to get rid of that in follow on commits.

powderluv (approved):
Happy to push and iterate on this since normal devs don't enter this path for now.

build_tools/build_libtorch.sh
powderluv: I hit this problem and the fix works.

build_tools/build_libtorch.sh
powderluv: just hard fail if ninja is not available. We require it to be installed as part of requirements.txt anyway

build_tools/build_libtorch.sh
powderluv: you will need to escape the `{` with `\{`

```
sed -i.bak -E 's/namespace ao \{/namespace ao \{\n#include <ATen\/native\/ao_sparse\/quantized\/cpu\/packed_params.h>/g' aten/src/ATen/native/ao_sparse/quantized/cpu/qlinear_deserialize.cpp
```

powderluv: https://github.com/llvm/torch-mlir/runs/7277257272?check_suite_focus=true#step:4:676

makslevental: yea portability between zsh (on the m1) and bash (on build machines) was what I wasn't sure about.

python/CMakeLists.txt
powderluv: FWIW We have to move all of these out with SET(ENV{LIBTORCH_VARIANT} ${LIBTORCH_VARIANT}) outside of the execute_process() so the builder would pass. I think the env vars set with -E env get lost when running inside a pip wheel sandbox.

PR #1033, opened 2022-07-11
torch: copy uses to prevent iterator invalidation

ashay:
Prior to this patch, the code in the `torch-simplify-shape-calculations`
pass iterated on the uses of an op's result while also modify the uses.
This caused the iterator to get invalidated, thus terminating the loop
early and producing incorrect IR.  This patch first copies the pointers
to all uses before iterating and possibly modifying the uses.


lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp
```diff
@@ -346,12 +346,18 @@ static void refineShapeCalculateResult(ShapeCalculateOp op, int resultNum,
...
+      llvm::map_range(result.getUses(), [](OpOperand &use) { return &use; }));
+
   // Update all the uses of the result type to the new type, if possible. Insert
   // a TensorStaticInfoCastOp for any users that might require the exact
   // previous type.
   Value originalTypedValue;
-  for (OpOperand &use : result.getUses()) {
-    if (use.getOwner()
+
+  for (OpOperand *use : originalUses) {
```
silvasean, line 359: I think llvm::make_early_inc_range would fix this too.

ashay: Ah nice!  It's now just a one-line change.  Thanks!

PR #1044, opened 2022-07-12
Prune xfail e2e LTC tests & fix bugs from functionalization pass

henrytwo:
Changes:
- Pruned number of xfailed e2e LTC tests from 305 to 134
  - Reviewed every failure to ensure the error genuinely warrants an xfail 
- Fixed bug where non-tensor outputs of LTC computation had `.to('cpu')` called, which caused a failure and inflated the xfail count
- Fixed bug with `HBC_basic` test where a constant tensor was created in its constructor without being declared as a buffer, which prevented the device from being updated when the parent `torch.nn.Module` got moved to the `lazy` device
  - Note that this test is still xfail'd due to some unsupported ops. Left a comment about some potential issues that may arise if it gets reenabled in the future 
- Updated autogen `GeneratedTorchOps.td` to reflect the latest set of supported ops
- Renamed `aten.zero.functionalization` to `aten.zero` to reflect upstream PyTorch changes

cc: @ke1337 @antoniojkim


python/torch_mlir_e2e_test/torchscript/configs/lazy_tensor_core.py
```diff
@@ -23,12 +23,33 @@ def run(self, artifact: torch.nn.Module, trace: Trace) -> Trace:
...
                 TraceItem(symbol=item.symbol,
                           inputs=item.inputs,
-                          output=output.to('cpu')))
+                          output=cpu_outputs))
 
         return result
+
+
+def to_device(value, device):
+    """
```
silvasean, line 39: recommend using tree_map from `from torch.utils._pytree import tree_map` -- see example use in python/torch_mlir_e2e_test/torchscript/configs/eager_mode.py

silvasean (approved):
Awesome!

PR #1046, opened 2022-07-12
Fix typo in `inputRank` check of `AtenBatchNormOp`

ramiro050:
The original conversion pattern for `AtenBatchNormOp` required that
the input rank be greater than 2; however, the only
expectation in the conversion pattern and in Pytorch is that the input
rank is greater than 1, since the second dimension of the input must
match the size of the `weight`, `bias`, `runningMean`, and
`runningVar` inputs. This commit fixes the `inputRank` check.


Shukla-Gaurav (approved):
LGTM!

PR #1047, opened 2022-07-12
WIP: Loosen global slot restrictions for aliased values

silvasean:
This is not functional yet. This is just to test if the handling here is
sufficiently general for the integration case of interest.

### DIFF
--- a/lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
+++ b/lib/Dialect/Torch/Transforms/GlobalizeObjectGraph.cpp
@@ -64,6 +64,16 @@ struct LinkageInfo {
   bool isPrivate;
 };
 } // namespace
+
+namespace {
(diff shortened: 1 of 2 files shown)


silvasean:
Hey Ashay, can you verify that your original model can make it past the GlobalizeObjectGraph pass with this modification? If so, then I just need to fix up some tests and op descriptions and InlineGlobalSlots. If not, then I will need to see the reproducer and generalize the logic appropriately.

ashay:
Hey Sean, sadly, it still fails with the same error as before.  I was able to reproduce the problem using [this smaller script](https://gist.github.com/ashay/15ecab16886356424e27b60649328da1), but let me know if you see a different result instead.

silvasean:
Superceded by https://github.com/llvm/torch-mlir/pull/1140

PR #1048, opened 2022-07-13
Blacklist _convolution op

henrytwo:
We need to blacklist `_convolution` since it conflicts with `convolution` during LTC codegen. Underscores are dropped when generating the C++ class name, which results in a compiler error.

`_convolution` should be decomposed to `convolution`, so blacklisting this op will not affect functionality.

This PR also fixes a rebase conflict, which resulted in a duplicated instance of `AtenSelectScatterOp`, `AtenSliceScatterOp`.

cc: @antoniojkim @ke1337


ramiro050 (approved):
LGTM!

PR #1050, opened 2022-07-13
simplify-shape-calc: propagate sizes across copy statements

ashay:
The `CopyToValueTensorOp` and `CopyToNonValueTensorOp` statements don't
change the size of the tensor.  By making the shape refinement passes
aware of this, the compiler now produces fewer unranked tensors, which
results in fewer translation failures downstream.  Specifically, this
patch looks for so-called downcast operations, where the tensor size is
dropped from the source tensor to the result, before relaying the size
of the source tensor to the uses of the downcast op.

The drawback of this change is that it introduces additional
`TensorStaticInfoCastOp` ops, but these ops should become NOPs later in
the pass pipeline.

To prevent a significant increase in compilation times, this patch
limits the shape propagation to at most six uses of the downcast op.

### DIFF
--- a/test/Dialect/Torch/simplify-shape-calculations.mlir
+++ b/test/Dialect/Torch/simplify-shape-calculations.mlir
@@ -444,23 +444,27 @@ func.func @fold_prim_unchecked_cast_op(%arg0: !torch.vtensor, %arg1: !torch.vten
 
 // CHECK:         %[[CAST_0:.*]] = torch.tensor_static_info_cast %arg0 : !torch.vtensor<[2],f32> to !torch.vtensor
 // CHECK:         %[[SHAPE_CALC_0:.*]] = torch.shape.calculate {
-// CHECK:           %[[NEG_0:.*]] = torch.aten.neg %[[CAST_0]] : !torch.vtensor -> !torch.tensor<[],unk>
+// CHECK:           %[[CAST_1:.*]] = torch.tensor_static_info_cast %[[CAST_0]] : !torch.vtensor to !torch.vtensor<[2],f32>
+// CHECK:           %[[NEG_0:.*]] = torch.aten.neg %[[CAST_1]] : !torch.vtensor<[2],f32> -> !torch.tensor<[],unk>
 // CHECK:           torch.shape.calculate.yield %[[NEG_0]] : !torch.tensor<[],unk>
 // CHECK:         } shapes {
 // CHECK:           torch.shape.calculate.yield.shapes %[[SHAPE_LIST]] : !torch.list<int>
 // CHECK:         } : !torch.tensor<[],unk>
-// CHECK:         %[[CAST_1:.*]] = torch.tensor_static_info_cast %[[SHAPE_CALC_0]] : !torch.tensor<[],unk> to !torch.tensor
+// CHECK:         %[[CAST_2:.*]] = torch.tensor_static_info_cast %[[SHAPE_CALC_0]] : !torch.tensor<[],unk> to !torch.tensor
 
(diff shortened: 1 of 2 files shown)


ashay:
This PR contains two commits, the first of which is a non-functional change to use uppercase variable names in the llvm-lit tests.  The second commit contains the key change.

test/Dialect/Torch/simplify-shape-calculations.mlir
```diff
@@ -489,3 +493,36 @@ func.func @shape_calc_with_two_uses(%arg0: !torch.vtensor<[2],f32>) -> !torch.vt
...
+// CHECK:         %[[CAST_2:.*]] = torch.tensor_static_info_cast %[[SHAPE_CALC]] : !torch.tensor<[2],unk> to !torch.tensor
+// CHECK:         return %[[CAST_2]] : !torch.tensor
+// CHECK:       }
+
+func.func @propagate_sizes_through_copy(%arg: !torch.vtensor<[2],f32>) -> !torch.tensor {
+  %cast = torch.tensor_static_info_cast %arg : !torch.vtensor<[2],f32> to !torch.vtensor
+  %tensor = torch.copy.to_tensor %cast : !torch.tensor
+
+  %shape_calc = torch.shape.calculate {
+    %neg = torch.aten.neg %tensor : !torch.tensor -> !torch.tensor
```
silvasean, line 519: Generally we shouldn't be doing shape inference without fully converting the program to value semantics. I think that would make torch.copy.to_tensor go away, and I think that the tensor_static_info_cast should canonicalize away in that case.

ashay: I see, thanks for the tip!  Perhaps I should be focusing on that pass.  There's a `torch.overwrite.tensor.contents` in the IR which is preventing the copies from being removed.  I'll look there first and come back to this.

ashay:
Closing this PR until I review the `torch-maximize-value-semantics` pass to see if it can be improved.

test/Dialect/Torch/simplify-shape-calculations.mlir
silvasean: I've noticed that we've had a few issues surrounding this imperfect conversion to value semantics causing deceptive error messages for developers.

Could you maybe add a pass lib/Dialect/Torch/Transforms/VerifyConversionToValueSemantics.cpp that checks that all tensors have been converted to value semantics, and put it in the pipeline after MaximizeValueSemantics (code can be mostly copypasted from lib/Dialect/TorchConversion/Transforms/VerifyInvariantsBeforeBackendLowering.cpp). I think there might be an edge case around return ops that we need to handle specially, but otherwise I think copying that other pass should be all we need.

ashay: I found that all tensor-returning functions end in the following sequence:

```
%copy = torch.copy.to_tensor %result : !torch.tensor
return %copy : !torch.tensor
```

which causes the invariant to fail.  Would you suggest updating the `GlobalizeObjectGraph` to generate functions that return value tensors instead?

silvasean: No, I would special-case the verification of that construct, or run RefinePublicReturn just before (we will also need to run it again later in the pipeline where it is today though).

ashay: Perfect, thanks for educating me (I had no idea what `RefinePublicReturn` did until you mentioned it)!  The validation pass is in PR https://github.com/llvm/torch-mlir/pull/1052.

PR #1051, opened 2022-07-13
canonicalizer: [nfc] update LIT variable names for consistency

ashay:
A previous patch used lowercase names for LIT variables.  This patch
replaces them with uppercase names to maintain consistency with other
variables.


silvasean (approved):
Thanks for the cleanup!

PR #1052, opened 2022-07-13
torch: add pass to catch non-value tensors

ashay:
This patch adds a new pass `torch-verify-conversion-to-value-semantics`,
which looks for non-value semantics tensors to catch such tensors early
during compilation.

This pass requires `torch-refine-public-return` pass to ensure that
return operations are updated to use value tensors, followed by the
canonicalize pass to remove any dead ops that may use or produce
non-value tensors.

### DIFF
--- a/lib/Dialect/Torch/Transforms/VerifyConversionToValueSemantics.cpp
+++ b/lib/Dialect/Torch/Transforms/VerifyConversionToValueSemantics.cpp
@@ -1,0 +1,64 @@
+//===- VerifyConversionToValueSemantics.cpp ----------------------*- C++-*-===//
+//
+// This file is licensed under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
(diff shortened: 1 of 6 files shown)


lib/Dialect/Torch/Transforms/VerifyConversionToValueSemantics.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "mlir/IR/BuiltinOps.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchTypes.h"
+#include "torch-mlir/Dialect/Torch/Transforms/Passes.h"
+
+using namespace mlir;
+using namespace mlir::torch::Torch;
+
+static LogicalResult checkValueType(Operation *op, Value value) {
+  auto isNotValueTensorType = value.getType().isa<NonValueTensorType>();
+  return isNotValueTensorType ? op->emitError("found non-value tensor type")
```
silvasean, line 21: Can you expand this error message a bit, since it will be what most devs run into? I would recommend adding a note saying "this is likely due to a missing case in the MaximizeValueSemantics pass"

silvasean (approved):
Thanks!

PR #1053, opened 2022-07-14
Fail to install x86_64 linux libtorch.so on other architectures.

u99127:
Found while trying to build torch-mlir on an AArch64 Linux VM, worth
a belts and braces to prevent such cases.

@sjarus

### DIFF
--- a/build_tools/build_libtorch.sh
+++ b/build_tools/build_libtorch.sh
@@ -87,7 +87,12 @@ download_libtorch() {
   fi
 elif [[ $(uname -s) = 'Linux' ]]; then
   echo "Linux detected"
-  DOWNLOAD_URL=${LINUX_X86_URL}
+  if [[ $(uname -m) == 'x86_64' ]]; then
+    DOWNLOAD_URL=${LINUX_X86_URL}
+  else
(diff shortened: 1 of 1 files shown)


u99127:
> thank you for this enhancement. We will be adding source builds soon so you can build on AArch64 linux by default too

Thanks for merging this so promptly @powderluv  - I've managed to get it building from source using -DLIBTORCH_SRC_BUILD=ON but I'm seeing some flakiness in the tests when I run them. Still debugging that and will report probably on the discuss forums once I have a reproducible set of issues .

powderluv:
Great to hear on -DLIBTORCH_SRC_BUILD=ON is working. I have a working branch https://github.com/llvm/torch-mlir/tree/release-src-build that passes all tests. I am trying to test on M1 and x86 mac. I will extract out the build_libtorch.sh script separately so it can be merged and unblock your use case. 

BTW what AArch64 linux system do you use ? Is it publicly available like Graviton etc? Would be nice to verify it works ok or even add a CI if you use that variant.

sjarus:
I believe @u99127 is running Ubuntu in a VM on the M1 Mac platform, not a Graviton instance. But the VM is intended to simulate that eventuality.

PR #1055, opened 2022-07-14
build: fix bazel build after adding VerifyConversionToValueSemantics.cpp

ashay:
A previous patch added a new file
("VerifyConversionToValueSemantics.cpp") to the build, but it did not
add it to the list files known to bazel.  This patch fixes the problem.


ashay:
Checked that running `bazel build @torch-mlir//...` now finishes successfully.

silvasean:
Hi @ashay -- generally it is not the responsibility of Torch-MLIR developers to keep the bazel build running. It is appreciated of course, but not expected in any way (and in fact, somewhat discouraged, to ensure that the subcommunity interested in bazel is truly shouldering the load here).

See:
https://github.com/llvm/torch-mlir/blob/main/development.md#bazel-build

silvasean (approved):

silvasean:
@ashay -- are you folks using the bazel build internally?

ashay:
> are you folks using the bazel build internally?

No, luckily, we use CMake + ninja. :)  I had forgotten our policy w.r.t bazel, but I'll keep that in mind for the future.

PR #1056, opened 2022-07-14
[MLIR] Add lowering of split like operations 

JakopinA:
A fix of PR #820, and a merge of #1022, should work without merge conflicts.


JakopinA:
Hi, I apologize for the mixed up PRs, but I can't re-open the old PR anymore. If you wouldn't mind taking a look at this PR that would be very appreciated.

ramiro050 (changes requested):
There are comments in https://github.com/llvm/torch-mlir/pull/1022 that have not been addressed. Can you please go over the comments in that PR and implement the necessary changes?

JakopinA:
@ramiro050 Hi ramiro, I tried fixing all the issues you mentioned in #1022 outside of this one: https://github.com/llvm/torch-mlir/pull/1022#pullrequestreview-1033179276

Just having some difficulties understanding what I need to do here, that's all. If you wouldn't mind helping, that would be great.

lib/Dialect/Torch/Transforms/DecomposeComplexOpsEarly.cpp
```diff
@@ -0,0 +1,279 @@
...
+  LogicalResult matchAndRewrite(SplitTensorLikeOp op,
+                                PatternRewriter &rewriter) const override {
+    int64_t dim;
+    if (!matchPattern(op.dim(), m_TorchConstantInt(&dim)))
+      return rewriter.notifyMatchFailure(op, "dim must be constant");
+
+    auto listType = op.getType().template cast<Torch::ListType>();
+    Location loc = op.getLoc();
+
+    auto viewOp = dyn_cast<AtenViewOp>(op.self().getDefiningOp());
```
ramiro050, line 95: Currently this conversion pattern only works if the input is the result of an `AtenViewOp`. This is too strict of a requirement. You should be able to make the calculations for the size of the chunks by using `getDimOp` to get the size of the dimension being used for splitting.

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOpsEarly.cpp
JakopinA: So if I'm getting this right, I would remove the auto viewOp, and would replace 
`autosizesListOp =
        dyn_cast<PrimListConstructOp>(viewOp.size().getDefiningOp());` 
with
`auto sizesListOp =
        dyn_cast<PrimListConstructOp>(op.self().getDimOp());`
Is that right?

ramiro050: No, `getDimOp` is not a method of `op.self()`, it's a helper function. You can find example usage by running `git grep 'getDimOp'`. You then have to build a `PrimListConstructOp` with the dimensions rather than doing `dyn_cast<PrimListConstructOp>`

qedawkins: I am wondering something about this pass. This implementation requires that the dim size for the split dimension is static, however because the DecomposeComplexOpsEarlyPass happens before shape refinement, if I understand correctly, we almost never know the shape of this dimension (except when we have something like a defining view op), even if it is something that would be determined statically. I think this was the motivation for requiring an `AtenViewOp` to define the input tensor. @ramiro050 Does this sound right to you?

glebk-cerebras:
@JakopinA do you have any plans to finish this PR?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -1054,6 +1054,41 @@ def aten〇index〇Tensor(self: List[int], indices: List[Optional[List[int]]]) -
...
+    start = 0
+    for i in range(num_indices):
+        end = split_sizes[i]
+        results[i] = upstream_shape_functions.slice(input, dim, start, end, 1)
+        start = end
+    return results
+
+def aten〇split(self: List[int], split_sizes: List[int], dim: int = 0) -> List[List[int]]:
+    return split_fn(self, split_sizes, dim)
+    #return upstream_shape_helpers.split(self, split_sizes, dim)
```
mgehre-amd, line 1070: nit: remove commented code

PR #1058, opened 2022-07-14
Remove libtorch downloads

powderluv:
Remove all the libtorch downloads. If the user sets
-DTORCH_MLIR_USE_INSTALLED_PYTORCH=OFF then just build from src.

Doesn't change developer workflow since we still default to local
PyTorch versions.

TEST: Build and verify all tests (except one xfail quant) pass on linux


makslevental (approved):
LGTM

silvasean (approved):
Thanks!

PR #1059, opened 2022-07-14
Bump LLVM to 889c6f3996769a991a24da957f597e7500d158e7

silvasean:
The biggest change here is to upgrade RefineTypes to the new sparse
dataflow framework.

Smaller changes:
- minor changes to type parsing
- suppress warnings in e2e tests


silvasean:
FYI @Shukla-Gaurav I needed this so I did this myself. the dataflow upgrade was more involved than I expected:

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -619,8 +624,14 @@ void TypeAnalyzer::fillInDTypeGivenDTypeAndDataType(ValueKnowledge &knowledge,
   fillInDTypeGivenDTypeIntAndInputDType(knowledge, dtype, dtypeForDataType);
 }
 
-ChangeResult TypeAnalyzer::visitOperation(
-    Operation *op, ArrayRef<LatticeElement<ValueKnowledge> *> operands) {
+void TypeAnalyzer::visitOperation(Operation *op,
+                                  ArrayRef<const ValueState *> operandsConst,
+                                  ArrayRef<ValueState *> results) {
+  SmallVector<ValueState *> nonConstOperands;
```
sstamenova, line 630: Is there a need to make these non constant other than propagating the `const` to other methods? If not, I think it is cleaner to add `const` instead of doing this.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -886,10 +897,9 @@ ChangeResult TypeAnalyzer::visitOperation(
     Type firstResDtype = operands[0]->getValue().dtype;
     Type secondResDtype =
         IntegerType::get(op->getContext(), 64, IntegerType::Signed);
-    ChangeResult firstRes = visitReductionAlongDimIntOp(
-        maxDim, maxDim.dim(), maxDim.keepdim(), firstResDtype, operands);
-    return firstRes |
-           visitReductionAlongDimIntOp(maxDim, maxDim.dim(), maxDim.keepdim(),
+    visitReductionAlongDimIntOp(maxDim, maxDim.dim(), maxDim.keepdim(),
+                                firstResDtype, operands);
+    return visitReductionAlongDimIntOp(maxDim, maxDim.dim(), maxDim.keepdim(),
```
sstamenova, line 902: Nit: I'd to the same thing here as up above for consistency and add the `return` on a newline.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -172,6 +174,16 @@ struct ValueKnowledge {
...
 
+  void print(raw_ostream &os) const {
+    os << "ValueKnowledge(";
+    if (dtype)
+      os << "dtype=" << dtype;
+    if (scalarType)
+      os << ", scalarType=" << scalarType;
+    if (optional != OptionalKnowledge::unKnown)
+      os << ", optional=" << (int)optional;
+    os << ", kind=" << (int)kind << ")";
```
sstamenova, line 185: Not really necessary, per se, but it would be cleaner to print the name of the kind instead of the int value.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -350,17 +362,32 @@ struct ValueKnowledge {
   // tracking wrapped type info when OptionalKnowledge::unKnown in the future.
   OptionalKnowledge optional;
 };
+} // namespace
 
+using ValueState = dataflow::Lattice<ValueKnowledge>;
```
sstamenova, line 367: Maybe add a `using mlir::dataflow`? It seems to be the pattern for other namespaces in this file.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1509,9 +1514,13 @@ namespace {
 class RefineTypesPass : public RefineTypesBase<RefineTypesPass> {
   void runOnOperation() override {
     auto func = getOperation();
-    TypeAnalyzer analyzer(&getContext());
-    analyzer.run(func);
-    optimize(func, analyzer);
+    DataFlowSolver solver;
+    solver.load<dataflow::DeadCodeAnalysis>();
+    solver.load<dataflow::SparseConstantPropagation>();
```
sstamenova, line 1519: Is the `SparseConstantPropagation` needed?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1509,9 +1514,13 @@ namespace {
...
+    TypeAnalyzer *analyzer = solver.load<TypeAnalyzer>();
```
sstamenova, line 1520: This really should be `TypeAnalysis` to be consistent with mlir.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1509,9 +1514,13 @@ namespace {
...
+    if (failed(solver.initializeAndRun(func)))
+      return signalPassFailure();
+    optimize(func, *analyzer);
```
sstamenova, line 1523: Again, to be consistent with mlir, this should take the `solver` instead of the `analyzer` and then you can remove `getLatticeElementPublic` and instead use `lookupState`, right?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1301,18 +1310,15 @@ static Type getMostRefinedStaticType(Value v, TypeAnalyzer &analyzer) {
         .getWithSizesAndDtype(tensorType.getOptionalSizes(), knowledge.dtype)
         .cast<BaseTensorType>();
   };
-
   if (auto tensorType = v.getType().dyn_cast<BaseTensorType>()) {
-    LatticeElement<ValueKnowledge> *latticeElement =
-        analyzer.lookupLatticeElement(v);
-    if (!latticeElement)
+    ValueState *latticeElement = analyzer.getLatticeElementPublic(v);
+    if (latticeElement->isUninitialized())
```
sstamenova, line 1315: If you use `lookupState`, you can keep the previous logic checking for `null`

sstamenova:
I needed this too and I was just about to send a similar pull request. Perhaps we can add a flow where when someone starts a merge, they create an Issue, so that we can avoid duplicating effort. And yes, that dataflow change was much more involved than expected.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
silvasean: getLatticeElementPublic is mainly for the free functions that TypeAnalysis calls. I changed the other code though.

silvasean:
Thanks @sstamenova for the detailed review!

PR #1062, opened 2022-07-15
[MLIR][TORCH] Add negative dim support for `aten.cat` and `aten.slice` op

vivekkhandelwal1:
This commit adds the support for negative dim cases for `aten.cat`,
`aten.slice.Tensor` and `aten.slice_scatter` op.


lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -76,6 +76,10 @@ LogicalResult prepareArgumentsForSlicingOp(OpTy op, OpAdaptor adaptor,
   if (!matchPattern(op.dim(), m_TorchConstantInt(&dim)))
     return op->emitError("unimplemented: dim is not constant");
 
+  int64_t inputRank = inputType.getRank();
+  if (dim < 0)
```
ramiro050, line 80: nit: use `toPositiveDim` and `isValidDim` with an error message when invalid. Same comment applies below to the other `dim` calculation

ramiro050 (approved):
LGTM, just one small change request

PR #1063, opened 2022-07-15
Add support for aten::index_add

navahgar:
This PR adds support for `aten::index_add` op, including its in-place variant, in mlir.


rdadolf:
I'm not the right one to review this, but as a first step, you'll want to add tests that verify this is working correctly. Thanks for the PR!

ramiro050:
There is currently someone else working on this op: https://github.com/llvm/torch-mlir/pull/964

navahgar:
> There is currently someone else working on this op: #964

Oh, good to know that. Thanks for pointing that out. I will close this one then.

PR #1067, opened 2022-07-15
Remove HasValueSemantics from `is` ops.

silvasean:


silvasean:
Part of the fix for https://github.com/llvm/torch-mlir/pull/993

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/registry.py
```diff
@@ -251,6 +251,11 @@ def has_value_semantics(self):
         # but the alias annotation is empty.
         if self.unique_key == "prim::unchecked_cast : (t) -> (t)":
             return False
+        # The `is` operator compares object identity, so it does not have
```
ashay, line 254: Nit: are you open to listing all overrides in one place?  I imagine we could need to add more in the future.

```python
overrides = [
        # It seems the FunctionSchema of "prim::unchecked_cast : (t) -> (t)" has
        # incorrect alias information. The result can alias with other tensors
        # but the alias annotation is empty.
        "prim::unchecked_cast : (t) -> (t)",

        # The `is` operator compares object identity, so it does not have
        # value semantics.
        "aten::__is__ : (t1, t2) -> (bool)",
        "aten::__isnot__ : (t1, t2) -> (bool)",
]

return not self.unique_key in overrides
```

ashay (approved):
Minor suggestion, but otherwise LGTM!

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/registry.py
silvasean: I think for the cases we have here, it reads a bit better to have them separate (if I put it in a list, I have a tendency to also want to hoist that out to the top level, which hurts reading locality). For sure if this list grows we will want something like you suggest.

PR #1070, opened 2022-07-17
Add linux source builds in CI

powderluv:
This enabled build Pytorch from source for the CI tests. Release
builds will follow


build_tools/build_libtorch.sh
```diff
@@ -68,29 +68,35 @@ build_pytorch() {
...
   USE_BREAKPAD=OFF \
   USE_CUDA=OFF \
+  USE_ITT=OFF \
   USE_DISTRIBUTED=OFF \
   USE_EIGEN_FOR_BLAS=OFF \
-  USE_FBGEMM=OFF \
+  USE_FBGEMM=ON \
   USE_GLOO=OFF \
-  USE_KINETO=OFF \
+  USE_KINETO=ON \
```
makslevental, line 91: just curious: why kineto?

makslevental (approved):
LGTM

build_tools/build_libtorch.sh
powderluv: It fails on x86_64 (only on the CI) without this. battle for another day.

build_tools/build_libtorch.sh
```diff
@@ -68,29 +68,35 @@ build_pytorch() {
...
   USE_MPS=OFF \
   USE_NCCL=OFF \
   USE_NNPACK=OFF \
   USE_OBSERVERS=OFF \
   USE_OPENMP=OFF \
-  USE_PYTORCH_QNNPACK=OFF \
+  USE_PYTORCH_QNNPACK=ON \
   USE_QNNPACK=OFF \
   USE_XNNPACK=OFF \
   ${PYTHON_BIN} setup.py  bdist_wheel -d "$WHEELHOUSE"
```
makslevental, line 102:
```suggestion
  ${PYTHON_BIN} setup.py  bdist_wheel -d "$WHEELHOUSE" | grep -Ev '^(adding|copying)'
```
to silence all of the `adding` and `copying` lines

powderluv: Will add in a follow on commit so we don't have to rerun CI

PR #1073, opened 2022-07-18
Change to notifyMatchFailure

jpienaar:
emitError is intended for error cases and not match failures of
patterns. notifyMatchFailure is intended where pattern reports reason
for not matching.

Op verification should also not happen inside patterns but as part of
verify, so left some of those as is to keep this change small.


stellaraccident (approved):
Thanks

PR #1078, opened 2022-07-18
bump llvm tag to 3580daa

ashay:
This patch makes some rudimentary changes to torch-mlir's use of MLIR
Python bindings to work with the most recent LLVM code.  We can perhaps
do better by being more selective in what we link against, instead of
using `MLIRPythonExtension.RegisterEverything`.


silvasean:
Looks like that change got reverted upstream: https://github.com/llvm/llvm-project/commit/d04c2b2fd916897cadc88f33ab87582f473600ac

PR #1082, opened 2022-07-19
Allow expanding and collapsing in aten::view

qedawkins:
- Supports cases where the view op expands and collapses dims
simultaneously. This does not handle the case where it is neither
expanding nor collapsing (e.g. [2, 3] -> [3, 2])
 - Additionally fixes a previous bug with adding 1-sized dims on both
sides of a tensor with aten.view

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -126,6 +126,7 @@ LogicalResult prepareArgumentsForSlicingOp(OpTy op, OpAdaptor adaptor,
   strides[dim] = rewriter.create<arith::MulIOp>(loc, strides[dim], stepIndex);
   return success();
 }
+
 namespace {
 class ConvertAtenFlattenUsingIntsOp
     : public OpConversionPattern<AtenFlattenUsingIntsOp> {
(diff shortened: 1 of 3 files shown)


qedawkins:
Also @silvasean is there a place I can request write access? This seemed like the best channel to ask in.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -334,139 +318,279 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+                  nextUnchangedOutput, inputShapeVec, outputShape,
+                  outputAssociations.back())))
+            return failure();
+          continue;
+        }
+
+        // Helper to find the minimum set of dims to collapse with the
+        // same number of elements as that of collapseDim. This function assumes
+        // the size of the collapsed dim is never dynamic.
+        auto minimallyCollapseDim = [](AtenViewOp op,
```
ramiro050, line 447: nit: can you make this helper into a `static` function? The lowering for this op keeps getting longer and longer, which makes it harder to read

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -334,139 +318,279 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+
+      bool hasDynamic = false;
+      while (inputDim < nextUnchangedInput && outputDim < nextUnchangedOutput) {
+        inputAssociations.emplace_back();
+        outputAssociations.emplace_back();
+
+        // Helper for filling in the remaining un-collapsed dims when the
+        // input/output dim is next to the next boundary dim. Additionally
+        // computes the size of a collapsed dynamic dim if necessary.
+        auto collapseToSingleDim = [](AtenViewOp op,
```
ramiro050, line 377: Same comment here: can you make this a `static` function?

ramiro050 (changes requested):
Thanks for working on this! Adding functionality to this op is very tricky. Below are some preliminary comments. I will do a more detailed pass in a bit

silvasean:
> Also @silvasean is there a place I can request write access? This seemed like the best channel to ask in.

Added you!

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -184,12 +185,96 @@ class ConvertAtenFlattenUsingIntsOp
...
+  // same number of elements as that of collapseDim. This function assumes
+  // the size of the collapsed dim is never dynamic.
+  static LogicalResult minimallyCollapseDimHelper(AtenViewOp op,
+                                 ConversionPatternRewriter &rewriter,
+                                 int64_t &collapseDim,
+                                 int64_t maxCollapseDim,
+                                 int64_t &startExpandDim,
+                                 int64_t maxExpandDim,
+                                 SmallVector<int64_t> &collapseShape,
+                                 SmallVector<int64_t> &expandShape,
```
ramiro050, line 247: nit: pass by `const` reference

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -184,12 +185,96 @@ class ConvertAtenFlattenUsingIntsOp
...
+
+
+  // Helper to find the minimum set of dims to collapse with the
+  // same number of elements as that of collapseDim. This function assumes
+  // the size of the collapsed dim is never dynamic.
+  static LogicalResult minimallyCollapseDimHelper(AtenViewOp op,
+                                 ConversionPatternRewriter &rewriter,
+                                 int64_t &collapseDim,
+                                 int64_t maxCollapseDim,
+                                 int64_t &startExpandDim,
```
ramiro050, line 244: Same comment as before, I would prefer these two variables not be passed by reference. For `startExpandDim`, you can update it using `expandIndices.back() + 1`

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -334,139 +403,200 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+      while (inputDim < nextUnchangedInput && outputDim < nextUnchangedOutput) {
+        inputAssociations.emplace_back();
+        outputAssociations.emplace_back();
+
+        // outputDim is next to the boundary
+        if (outputDim == nextUnchangedOutput - 1) {
+          if (hasDynamic && inputDim != nextUnchangedInput - 1) {
+            return rewriter.notifyMatchFailure(
+                op,
+                "found ambiguous association of dynamic sizes");
```
ramiro050, line 464: nit: can you have a more descriptive error message? Namely, I think it would be useful to mention that this error happened because `inputDim` did not have the value that was expected. This allows developers to easily find the line where a match failure is generated (currently this error message is identical to one used below)

ramiro050 (changes requested):
I just have a few small comments

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -184,12 +185,96 @@ class ConvertAtenFlattenUsingIntsOp
...
+  // input/output dim is next to the next boundary dim. Additionally
+  // computes the size of a collapsed dynamic dim if necessary.
+  static LogicalResult collapseToSingleDimHelper(AtenViewOp op,
+                                ConversionPatternRewriter &rewriter,
+                                int64_t &collapseDim,
+                                int64_t maxCollapseDim,
+                                int64_t &startExpandDim,
+                                int64_t maxExpandDim,
+                                SmallVector<int64_t> &collapseShape,
+                                SmallVector<int64_t> &expandShape,
```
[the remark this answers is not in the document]
qedawkins: collapseShape is still passed without const because I need to update it for the dynamic shapes case.

qedawkins:
Thanks for the comments! Let me know if it looks like there are still issues.

ramiro050 (approved):
LGTM!

PR #1083, opened 2022-07-19
[MHLO] Init MHLO integration.

ZihengJiang:

Co-authored-by: Bairen Yi <yibairen.byron@bytedance.com>
Co-authored-by: Jiawei Wu <xremold@gmail.com>
Co-authored-by: Tianyou Guo <alibaba-inc.com>
Co-authored-by: Xu Yan <yancey.yx@alibaba-inc.com>
Co-authored-by: Ziheng Jiang <ziheng.jiang@bytedance.com>


ZihengJiang:
ready for review @silvasean @powderluv @fortianyou

python/torch_mlir/__init__.py
```diff
@@ -43,6 +43,8 @@ class OutputType(Enum):
     # Raw output of the JIT IR importer. This is not expected to be useful
     # for end-users, but can be convenient for development or reporting bugs.
     RAW = 3
+    # This output type consists of `mhlo` dialect ops.
```
silvasean, line 46: Nit: please put blank link above this comment. Also, add "It can be thought of
    # as taking the `TORCH` output type and lowering it to MHLO."

lib/Dialect/TorchConversion/Transforms/Passes.cpp
```diff
@@ -118,3 +127,26 @@ void TorchConversion::createTorchBackendToTosaBackendPipeline(
   // correct form.
   pm.addPass(TorchConversion::createVerifyTosaBackendContractPass());
 }
+
+#ifdef ADD_MHLO_TARGET
```
silvasean, line 131: TORCH_MLIR_ENABLE_MHLO

silvasean (changes requested):

lib/Conversion/TorchToMhlo/PopulatePattern.h
```diff
@@ -0,0 +1,22 @@
+//===------------------------------------------------------------*- C++ -*-===//
```
silvasean, line 1: nit: rename file to PopulatePatterns.h for consistency with https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchToLinalg/PopulatePatterns.h

setup.py
```diff
@@ -75,9 +75,13 @@ def run(self):
...
                 f"-DMLIR_ENABLE_BINDINGS_PYTHON=ON",
                 f"-DLLVM_ENABLE_PROJECTS=mlir",
-                f"-DLLVM_EXTERNAL_PROJECTS=torch-mlir;torch-mlir-dialects",
+                f"-DTORCH_MLIR_ENABLE_MHLO=ON",
+                f"-DLLVM_EXTERNAL_PROJECTS=torch-mlir;torch-mlir-dialects;mlir_hlo",
                 f"-DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR={src_dir}",
                 f"-DLLVM_EXTERNAL_TORCH_MLIR_DIALECTS_SOURCE_DIR={src_dir}/externals/llvm-external-projects/torch-mlir-dialects",
+                f"-DLLVM_EXTERNAL_MLIR_HLO_SOURCE_DIR={src_dir}/externals/mlir-hlo",
+                # FIXME: set MLIR_PDLL_TABLEGEN_EXE since mlir-hlo doesn't
+                f"-DMLIR_PDLL_TABLEGEN_EXE=mlir-pdll",
```
powderluv, line 84: where do you get this tool binary from ? Is it built as part of our build ? Is it a host tool ? We cross-compile for arm64 and have to specify the host version like we do for llvm-tblgen.

powderluv (commented):
I think you also need to add -DTORCH_MLIR_ENABLE_MHLO=ON in the .github/workflows to get the CIs to pass

setup.py
powderluv: I think for now you want to point to the mlir-pdll that we built in the CMAKE_BINARY_DIR (or equivalent) or you will likely pick up the system one ? 

I will take care of the host / cross-compile whenever we land that in the future. So just have to confirm this mlir-pdll is the version we just built.

ZihengJiang: I would like to use the mlir-pdll that we built in the CMAKE_BINARY_DIR. I raised a PR to MHLO to remove this config: https://github.com/tensorflow/mlir-hlo/pull/41

lib/Dialect/TorchConversion/Transforms/Passes.cpp
```diff
@@ -20,6 +20,7 @@
 #include "torch-mlir/Conversion/TorchToStd/TorchToStd.h"
 #include "torch-mlir/Conversion/TorchToTMTensor/TorchToTMTensor.h"
 #include "torch-mlir/Conversion/TorchToTosa/TorchToTosa.h"
+#include "torch-mlir/Conversion/TorchToMhlo/TorchToMhlo.h"
```
fortianyou, line 23: Nit: How about adding the `TORCH_MLIR_ENABLE_MHLO` macro guard here as well?

ZihengJiang:
@powderluv I change mhlo's compilation to embedded build so that we don't need to feed the PDLL_EXE option.

ZihengJiang:
@silvasean @powderluv Please review

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+template <typename AtenOpT>
+class ConvertAtenOp : public OpConversionPattern<AtenOpT> {
+public:
+  using OpConversionPattern<AtenOpT>::OpConversionPattern;
+  using OpAdaptor = typename AtenOpT::Adaptor;
+  LogicalResult
+  matchAndRewrite(AtenOpT op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override;
+};
```
silvasean, line 39: nit: anonymous namespace can end here

In LLVM we keep anonymous namespaces as small as possible (one around each class): https://llvm.org/docs/CodingStandards.html#anonymous-namespaces

silvasean (approved):

powderluv (approved):

ZihengJiang:
Merged now. Thanks @silvasean @powderluv @fortianyou

PR #1084, opened 2022-07-19
python: trim registration and loading of dialects and passes

ashay:
In the interest of merging upstream LLVM quickly, a previous patch
(7f08169) updated the torch-mlir build to register all dialects and
passes through Python bindings.  This patch limits the dialects and
passes to only those that are used in torch-mlir.

Key to this change are the removal of
`MLIRPythonExtension.RegisterEverything` and the introduction of a new
Python module (`_mlir_libs/_site_initialize_0.py`), where we register
the dialects and passes used by torch-mlir.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -26,6 +26,7 @@ declare_mlir_python_sources(TorchMLIRPythonSources.TopLevel
   SOURCES
     __init__.py
     compiler_utils.py
+    _mlir_libs/_site_initialize_0.py
 )
 
 declare_mlir_python_sources(TorchMLIRPythonSources.Dialects
@@ -102,16 +103,10 @@ add_subdirectory(torch_mlir/eager_mode)
 ################################################################################
 
 set(_source_components
-  # TODO: Core is now implicitly building/registering all dialects, increasing
-  # build burden by ~5x. Make it stop.
-  # TODO: Reduce dependencies. We need ExecutionEngine and a bunch of passes
-  # for the reference backend, but logically they can be separate. But seemingly
-  # the only way to handle that is to create a separate mlir python package
-  # tree, which seems excessive.
-  MLIRPythonSources
+  MLIRPythonSources.Core
+  MLIRPythonSources.Dialects.func
+  MLIRPythonSources.ExecutionEngine
   MLIRPythonExtension.Core
-  MLIRPythonExtension.RegisterEverything
-  MLIRPythonExtension.ExecutionEngine
   TorchMLIRPythonSources
   TorchMLIRPythonExtensions
 )
--- a/python/TorchMLIRModule.cpp
+++ b/python/TorchMLIRModule.cpp
@@ -9,6 +9,7 @@
 
 #include "mlir-c/Bindings/Python/Interop.h"
 #include "mlir/Bindings/Python/PybindAdaptors.h"
+#include "mlir/CAPI/IR.h"
 #include "torch-mlir-c/Dialects.h"
 #include "torch-mlir-c/Registration.h"
 
@@ -19,14 +20,6 @@ PYBIND11_MODULE(_torchMlir, m) {
 
   m.doc() = "torch-mlir main python extension";
 
-  m.def(
-      "register_dialect",
-      [](MlirContext context, bool load) {
-        MlirDialectHandle handle = mlirGetDialectHandle__torch__();
-        mlirDialectHandleRegisterDialect(handle, context);
-        if (load) {
-          mlirDialectHandleLoadDialect(handle, context);
-        }
-      },
-      py::arg("context"), py::arg("load") = true);
+  m.def("register_required_dialects", torchMlirRegisterRequiredDialects,
+        py::arg("context"));
 }
--- a/include/torch-mlir-c/Registration.h
+++ b/include/torch-mlir-c/Registration.h
@@ -22,6 +22,9 @@ extern "C" {
  */
 MLIR_CAPI_EXPORTED void torchMlirRegisterAllDialects(MlirContext context);
 
+/** Registers upstream (MLIR) dialects used in Torch-MLIR IRs. */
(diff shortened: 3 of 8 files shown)


silvasean:
Nice :)

silvasean (approved):

ashay:
I just now found out that the build error is due to the use of the default (`ld`?).  I am using `lld` on my local machine, which doesn't produce the error.  Still investigating.

python/CMakeLists.txt
```diff
@@ -53,6 +54,23 @@ declare_mlir_python_extension(TorchMLIRPythonExtensions.Main
   EMBED_CAPI_LINK_LIBS
     TorchMLIRCAPI
   PRIVATE_LINK_LIBS
+    MLIRAffineToStandard
```
stellaraccident, line 57: I am somewhat surprised by having this list of C++ deps here... reading more and seeing if I can have a recommendation.

As written, this is linking part of MLIR into the Python extension itself and part into the CAPI .so... that is probably not what we want.

python/CMakeLists.txt
```diff
@@ -102,16 +120,10 @@ add_subdirectory(torch_mlir/eager_mode)
...
-  # TODO: Core is now implicitly building/registering all dialects, increasing
-  # build burden by ~5x. Make it stop.
-  # TODO: Reduce dependencies. We need ExecutionEngine and a bunch of passes
-  # for the reference backend, but logically they can be separate. But seemingly
-  # the only way to handle that is to create a separate mlir python package
-  # tree, which seems excessive.
-  MLIRPythonSources
+  MLIRPythonSources.Core
+  MLIRPythonSources.Dialects.func
+  MLIRPythonSources.ExecutionEngine
```
stellaraccident, line 125: Note for the future: if you ever end up in a world where you don't need the execution engine (or maybe only for source builds/not deployment), the size and number of build artifacts will go way down.

python/TorchMLIRModule.cpp
```diff
@@ -9,24 +9,49 @@
...
+#include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/InitAllPasses.h"
 #include "torch-mlir-c/Dialects.h"
 #include "torch-mlir-c/Registration.h"
 
 namespace py = pybind11;
 
+using namespace mlir;
+
+void registerUpstreamPasses() {
```
stellaraccident, line 22: For reasons, I would recommend against cross linking this directly from the pybind extension. You're going to want to expose this as an entry-point in the torch-mlir CAPI. I would recommend just adding them into the existing torchMlirRegisterAllPasses(). Or introduce a new torchMlirRegisterUpstreamPasses() (but separating probably doesn't buy you much).

python/TorchMLIRModule.cpp
```diff
@@ -9,24 +9,49 @@
...
   m.def(
-      "register_dialect",
-      [](MlirContext context, bool load) {
-        MlirDialectHandle handle = mlirGetDialectHandle__torch__();
-        mlirDialectHandleRegisterDialect(handle, context);
-        if (load) {
-          mlirDialectHandleLoadDialect(handle, context);
-        }
+      "register_required_dialects",
+      [](MlirContext context) {
```
stellaraccident, line 51: Ditto - don't directly use the C++ API here. Until I get all of the dialect handles exposed upstream for individual use, I would recommend creating a torch CAPI (i.e. torchMlirRegisterDialects) that does this. That will keep the layering right and the CAPI self contained.

python/torch_mlir/_mlir_libs/_mlirRegisterEverything.py
```diff
@@ -0,0 +1,4 @@
+from . import _torchMlir
+
+def context_init_hook(context):
```
stellaraccident, line 3: Call this file `_site_initialize_0.py` (example: https://github.com/iree-org/iree/pull/9638/files#diff-27a6b169f4cb66761062a33d5cddb70085a959eb91e971aff58c2a8f2fc129a7)

python/TorchMLIRModule.cpp
stellaraccident: Here is an example of a C API that registers dialects: https://github.com/iree-org/iree/blob/main/compiler/src/iree/compiler/API/Compiler.cpp#L65

stellaraccident (commented):
Thanks. I left some notes. I suspect that the C/C++ cross linking thing is actually what is tripping up ld, since what you are doing is introducing a kind of ODR violation.

I don't have time tonight, but if you struggle with this, let me know and I can straighten it out.

stellaraccident (approved):
Much better - thanks!

ashay:
> I suspect that the C/C++ cross linking thing is actually what is tripping up ld, since what you are doing is introducing a kind of ODR violation.

Thanks so much, that was spot on!  I believe I have separated the C and C++ APIs now, but please let me know if you think more changes are necessary.  Locally, using `ld` now works, so hopefully CI will pass as well.

stellaraccident:
> > I suspect that the C/C++ cross linking thing is actually what is tripping up ld, since what you are doing is introducing a kind of ODR violation.
> 
> Thanks so much, that was spot on! I believe I have separated the C and C++ APIs now, but please let me know if you think more changes are necessary. Locally, using `ld` now works, so hopefully CI will pass as well.

I wish it didn't rely on such esoteric knowledge to spot such things. But glad it helped regardless :)  Thanks for doing the adaptations! Much appreciated.

powderluv:
@ashay could this be causing the nightly failure and also the 
```
 ImportError: dlopen(/Users/kkiningh/Workspace/pytorch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/_mlir_libs/_mlirRegisterEverything.cpython-39-darwin.so, 0x0002): symbol not found 
in flat namespace '_mlirRegisterAllDialects'  
``` 
reported on discord ?

PR #1090, opened 2022-07-21
[MHLO] Init MHLO view like op patterns

fortianyou:

Co-authored-by: Bairen Yi yibairen.byron@bytedance.com
Co-authored-by: Jiawei Wu xremold@gmail.com
Co-authored-by: Tianyou Guo tianyou.gty@alibaba-inc.com
Co-authored-by: Xu Yan yancey.yx@alibaba-inc.com
Co-authored-by: Ziheng Jiang ziheng.jiang@bytedance.com


fortianyou:
This is a follow-up PR of https://github.com/llvm/torch-mlir/pull/1083
Please review for me @silvasean @powderluv @ZihengJiang @byronyi @Yancey1989

lib/Conversion/TorchToMhlo/ViewLikeOps.cpp
```diff
@@ -0,0 +1,132 @@
...
+    // the range of i32(4GiB)
+    std::for_each(dimSizes.begin(), dimSizes.end(), [&](Value& dSize) {
+      // dimSize: cast i64 -> i32
+      dSize = rewriter.create<arith::TruncIOp>(loc, rewriter.getI32Type(), dSize);
+      return dSize;
+    });
+#endif
+
+    Value mhloShape = rewriter.create<tensor::FromElementsOp>(loc, dimSizes);
+    rewriter.replaceOpWithNewOp<chlo::DynamicReshapeOp>(
```
ZihengJiang, line 85: Can we use mhlo::DynamicReshapeOp?

fortianyou: I think we should not use [`mhlo::DynamicReshapeOp`](https://github.com/tensorflow/mlir-hlo/blob/master/include/mlir-hlo/Dialect/mhlo/IR/hlo_ops.td#L1900); because it is not sufficient to represent the reshape from Torch Dialect, since a single new dimension specified as -1 is allowed(the size -1 will be inferred from other dimensions)

[`chlo::DynamicReshapeOp`](https://github.com/tensorflow/mlir-hlo/blob/master/include/mlir-hlo/Dialect/mhlo/IR/chlo_ops.td#L909) is almost the same as  [`mhlo::DynamicReshapeOp`](https://github.com/tensorflow/mlir-hlo/blob/master/include/mlir-hlo/Dialect/mhlo/IR/hlo_ops.td#L1900) , but it allows a single dimension to be specified as -1, and it will be computed by the op to create a valid reshape.

CMakeLists.txt
```diff
@@ -39,6 +39,11 @@ endmacro()
 option(TORCH_MLIR_ENABLE_MHLO "Add mhlo dialect" ON)
 if(TORCH_MLIR_ENABLE_MHLO)
   add_definitions(-DTORCH_MLIR_ENABLE_MHLO)
+  option(TORCH_MLIR_ENABLE_MHLO_TRUNC_DIMSIZE_TO_I32
```
silvasean, line 42: Can you add a "rationale" comment here justifying this flag and why we need it?

silvasean (approved):

silvasean:
I will let you folks iterate on the pattern details. LGTM overall.

lib/Conversion/TorchToMhlo/ViewLikeOps.cpp
ZihengJiang: As an exchangeable IR format, It would be good if we can converge the produced result into a single layer, so that the backend do not need to handle two set of dialects. Could you try to:
- represent the rewriting with pure mhlo like here: https://github.com/tensorflow/mlir-hlo/blob/master/lib/Dialect/mhlo/transforms/chlo_legalize_to_hlo.cc#L1698-L1707  
- Or add a chlo_legalize_to_hlo transform into the TorchToMhlo pipeline?

fortianyou: Thanks for advice, I had eliminated the usages of chlo dialect.

CMakeLists.txt
fortianyou: done

ZihengJiang (approved):
Looks great! Thanks @fortianyou

PR #1095, opened 2022-07-21
[DO NOT MERGE] LTC CI experiments

henrytwo:
PR to experiment with CI so that people don't get spammed with notifications.


silvasean:
ASan is notirously difficult to set up for Python extensions -- you may want to create a standalone C++ .cpp file that links against libtorch to repro it. And I'm not sure how to get an ASan PyTorch build (due to how ASan works, it might catch the bug without an ASanified PyTorch,  but if doesn't catch anything, then you will have to investigate how to do that :/ ).

.github/workflows/buildAndTest.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         mkdir build
         cd build
         cmake $GITHUB_WORKSPACE/externals/llvm-project/llvm -GNinja \
-          -DCMAKE_BUILD_TYPE=Release \
+          -DCMAKE_BUILD_TYPE=Asan \
```
silvasean, line 29: CMake doesn't know about ASan per se. See https://llvm.org/docs/CMake.html - LLVM_USE_SANITIZER:STRING

henrytwo: Ahh, I'm not familiar with ASan so I've just been browsing stack overflow 😅. Thanks for the tip!

henrytwo:
> ASan is notirously difficult to set up for Python extensions -- you may want to create a standalone C++ .cpp file that links against libtorch to repro it. And I'm not sure how to get an ASan PyTorch build (due to how ASan works, it might catch the bug without an ASanified PyTorch, but if doesn't catch anything, then you will have to investigate how to do that :/ ).

I'll try to get around to doing this tomorrow. For now, I'm narrowing down the source from "our side" by making a minimal python script that invokes the LTC backend to see if the same problem can be replicated there. If not, it could indicate something specific to e2e testing

powderluv:
btw feel free to cancel CI builds that are no longer of use. Right now there are five outstanding builds :D

PR #1096, opened 2022-07-22
Barebones libtorch build 

makslevental:
This clocks in at 258 files and builds and links successfully on x86 (at least my workstation). With this minimal path the tests don't pass of course because no libtorch python bindings (or kernels or anything else) are included.

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/ivalue_importer.cpp
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/ivalue_importer.cpp
@@ -11,6 +11,7 @@
 #include "class_annotator.h"
 #include "function_importer.h"
 #include "torch_to_mlir_utils.h"
+#include "packed_params.h"
 
 #include <unordered_map>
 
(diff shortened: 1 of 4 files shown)


python/torch_mlir/dialects/torch/importer/jit_ir/csrc/ivalue_importer.cpp
```diff
@@ -11,6 +11,7 @@
 #include "class_annotator.h"
 #include "function_importer.h"
 #include "torch_to_mlir_utils.h"
+#include "packed_params.h"
```
makslevental, line 14: this header in pytorch is only included if you build quantized infra [quantized/cpu/packed_params.h](https://github.com/pytorch/pytorch/blob/1f29b3130af218847a043e58fdc64511bbe072fe/aten/src/ATen/native/quantized/cpu/packed_params.h) even though from the looks of it there's nothing quantized about it.

powderluv:
This is awesome !!! The Python bindings can be from the released version of pytorch right ? This way this integration can span various versions of PyTorch until there is an api breakage

PR #1097, opened 2022-07-22
add general support for aten.index.Tensor

qedawkins:
- Only offers support for statically sized index tensors when there is more than one
 - Dynamic shape support remains for single indexing tensors
 - Uses op folds in the mlir::shape dialect; shape ops shouldn't ever be generated in IR
 - Currently failing one test case due to a shape function failure


ramiro050 (changes requested):
Looks mostly good, I just have a few comments

qedawkins:
I addressed the comments and refactored out a few unused variables. Another pass would be appreciated to make sure it looks good.

ramiro050 (approved):
LGTM!

PR #1100, opened 2022-07-22
[MLIR][TORCH] Add E2E support for `aten.var.correction` and `aten.ge.int` op 

vivekkhandelwal1:


python/torch_mlir_e2e_test/test_suite/stats.py
```diff
@@ -427,3 +427,116 @@ def forward(self, x):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, x):
+        return torch.ops.aten.var(x, dim=None, correction=2)
```
ramiro050, line 446: Can you also add a test with `dim=[]`. This case should be the same in behavior as `dim=None`, and it would be nice to verify that that is the case in torch-mlir

ramiro050 (changes requested):
Looks mostly good, I just have a few comments below

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2138,6 +2138,89 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
...
+
+  unsigned inputRank = getTensorRank(self);
+  SmallVector<Value> dimListElements;
+  bool isNoneOrEmpty = true;
+  if (!dimList.getType().template isa<Torch::NoneType>()) {
+    if (!getListConstructElements(dimList, dimListElements))
+      return rewriter.notifyMatchFailure(
+          op, "expect dimList to be constructed from list construct");
+    if (dimListElements.size() != 0 || inputRank == 0)
+      isNoneOrEmpty = false;
```
ramiro050, line 2163: A couple of comments/questions on this:

1. Why is the `inputRank == 0` condition needed? If `inputRank == 0`, the `for` loop in line 2166 should not run, so the behavior should be the same.
2. You don't need the `if`. You can do `isNoneOrEmpty = dimListElements.empty();`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2138,6 +2138,89 @@ class DecomposeAtenNumpyTOp : public OpRewritePattern<AtenNumpyTOp> {
...
+  // Divide the square sum by productDimSize - correction.
+  Value squareSum = rewriter.create<AtenSumDimIntListOp>(
+      loc, outputType, square, dimList, keepDim, /*dtype=*/constantNone);
+
+  // `productDimSize` is product of sizes of dimensions to be reduced.
+  Value constantOne =
+      rewriter.create<Torch::ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
+  Value productDimSize = constantOne;
+  for (Value dim : dimListElements) {
+    Value dimSize = rewriter.create<AtenSizeIntOp>(loc, self, dim);
```
ramiro050, line 2202: I would've expected `AtenSizeIntOp` and `AtenMeanDimOp` to handle the case of `dim=[]` rather than creating a special `dimList` in this decomposition, since their python counterparts reduce to a scalar when `dim=[]`. Is this not the case? If so, that should be fixed first.

```python
In [1]: import torch

In [2]: a = torch.rand((1,2,3))

In [3]: torch.ops.aten.sum(a, dim=[])
Out[3]: tensor(2.9630)

In [4]: torch.ops.aten.sum(a, dim=None)
Out[4]: tensor(2.9630)

In [5]: torch.ops.aten.mean(a, dim=[])
Out[5]: tensor(0.4938)
```

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
vivekkhandelwal1: I think you mean to say `AtenSumDimIntListOp` and `AtenMeanDimOp`. Yes, we need to fix the `AtenSumDimIntListOp`, that will automatically fix the `AtenMeanDimOp`. But we still need to create this special `dimList` for `AtenSizeIntOp`, which expects `dim` to be an `integer` not a `list`. Also, we need to fix the shape calculation for `AtenSumDimIntListOp`, and `AtenMeanDimOp`.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
vivekkhandelwal1: For your first comment/question if we don't check for the condition `inputRank == 0`, then it's true that the for loop will not run but https://github.com/llvm/torch-mlir/blob/12ee0d041b15d501ca07d758fe647d62cd8cbc7e/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L2169-L2171 will again create an empty list, which we already have.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -533,13 +535,14 @@ def aten〇max〇dim(self: List[int], dim: int, keepdim: bool = False) -> Tuple[
     return reduced_shape, reduced_shape
 
 def aten〇mean〇dim(self: List[int], dim: List[int], keepdim: bool = False, dtype: Optional[int] = None) -> List[int]:
+    if len(dim)==0:
+        dim = list(range(len(self)))
```
ramiro050, line 539: This is fine for now, but you should upstream this fix to PyTorch as well

ramiro050 (approved):
LGTM!

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
vivekkhandelwal1: Sure, will do.

silvasean:
Just wanted to say thanks for the attention to detail here -- VarCorrectionLargeInputModule_basic caught a legitimate bug downstream where the increased accumulation precision wasn't being respected!

PR #1101, opened 2022-07-22
Add support for `lift_fresh` op

antoniojkim:
In upstream PyTorch, they switched the tensor constructor to use `lift_fresh` instead of `lift` which silently broken LTC. Adding support for this op should fix the examples being broken.

The upstream fix for the TS backend: https://github.com/pytorch/pytorch/pull/81928


antoniojkim:
@silvasean Gentle reminder to please review. Should be quick as its a relatively small change

silvasean (approved):

PR #1105, opened 2022-07-25
Update development.md with source builds

powderluv:
Add source builds and remove deprecated libtorch information.


makslevental (approved):
LGTM

Issue #1075, opened 2022-07-18
[RFC] aten.to.device


gpetters94:
I'm running into calls to `aten.to.device` in a model and I was wondering how we want to handle it. [It looks like](https://pytorch.org/docs/stable/generated/torch.Tensor.to.html) we should just be able to replace it with a conversion but I wanted to double check if we wanted to handle the logic of moving tensors between devices.

PR #1107, opened 2022-07-25
Add lowering for aten.to.device

gpetters94:
As per #1075

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -1997,6 +1997,25 @@ public:
 } // namespace
 
 namespace {
+// Decompose `aten.to.device` op into `aten.to.dtype` op.
+class DecomposeAtenToDeviceOp : public OpRewritePattern<AtenToDeviceOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenToDeviceOp op,
+                                PatternRewriter &rewriter) const override {
+
+    // Device information isn't relevant to torch-mlir, so we can drop that info
+    // here.
+    rewriter.replaceOpWithNewOp<AtenToDtypeOp>(op, op.getType(), op.self(),
+                                               op.dtype(), op.non_blocking(),
+                                               op.copy(), op.memory_format());
+
+    return success();
+  }
+};
+} // namespace
+
+namespace {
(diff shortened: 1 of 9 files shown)


pashu123 (approved):
LGTM.

pashu123:
@gpetters94 Could you add an e2e test case?

vivekkhandelwal1 (commented):
Is the lowering correct? Is it doing what the op is supposed to do?

pashu123:
We can just fold the aten.to.device op. We don't have distributed/device semantics here. Pinging @ramiro050.

gpetters94:
I'm not sure if I can add a test for this op. Its name should be `to.device` but it that's not valid Python syntax, so I'm not sure what I should be calling in the test.

pashu123:
Please squash all your commits into one.

ramiro050:
> I'm not sure if I can add a test for this op. Its name should be `to.device` but it that's not valid Python syntax, so I'm not sure what I should be calling in the test.

You need to specify enough arguments so that the right variant of `.to` gets used. If you do `torch.ops.aten.to(input, device='cpu', dtype=torch.float, non_blocking=False)`, you should get `to.device` generated

ramiro050:
> We can just fold the aten.to.device op. We don't have distributed/device semantics here. Pinging @ramiro050.

This op is a bit strange, because it also has the ability of changing the `dtype` of the tensor, so we can't just fold it away. An option would be to use a canonicalization pattern, but the pattern would have to be conditional on the input being a value-semantics tensor (see https://github.com/llvm/torch-mlir/issues/1075#issuecomment-1187776230), so that it does not get applied too early. I personally think it is simpler to just do a decomposition

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1996,6 +1996,23 @@ class DecomposeAtenToDtypeLayoutOp
 };
 } // namespace
 
+namespace {
+// Decompose `aten.to.device` op into `aten.to.dtype` op.
+class DecomposeAtenToDeviceOp : public OpRewritePattern<AtenToDeviceOp> {
```
ramiro050, line 2001: Drive-by comment: I would add a comment here mentioning that once we get to decompositions, the `device` argument no longer means anything in torch-mlir, so we can drop that information.

pashu123:
@gpetters94 Please rebase and push.

PR #1108, opened 2022-07-25
Export LTC Headers

antoniojkim:
Currently the LTC headers are not being exported in the Python wheel. This means they aren't being distributed. This is something that is needed for vendors that do not intend to clone/build the Torch-MLIR repo in order to use the base lazy backend.

CC: @ke1337 @henrytwo


silvasean:
Fine by me in principle @powderluv  any concerns?

powderluv:
this is ok. I think eventually upstream will have to expose it too. We can also expose it in the source builds.

antoniojkim:
@silvasean @powderluv Can I get a stamp of approval for this PR then?

silvasean (approved):

PR #1110, opened 2022-07-26
Resolve remaining LTC CI failures

henrytwo:
This PR addresses several issues currently faced by CI:
1. Previously, LTC would refuse to build when `TORCH_MLIR_USE_INSTALLED_PYTORCH=OFF` due to LTC not checking this variable. The build scripts have been updated such that LTC uses the appropriate PyTorch build
2. Blacklist any ops that are prefixed by `new_` or end with `_like` in response to https://github.com/pytorch/pytorch/pull/82238
3. Update LTC xfails to reflect new set of supported ops
4. Prune unused Python C++ dependencies
5. Add Shape inference function to `var` op, which is used by a recently added test
6. Due to the aforementioned problems, and linkage issues with out-of-tree builds, this PR also modifies the CI to skip building LTC in the configs which currently do not succeed. This way, we can merge into main to ensure stability before investigating further.
  a. To avoid import issues from the LTC pybind no longer generating, this PR introduces a script to autogen a dummy Python file that allows for a successful import (however throws an error if it's actually used)

cc: @antoniojkim @ke1337


python/CMakeLists.txt
```diff
@@ -13,6 +13,28 @@ set(TORCH_MLIR_PYTHON_ROOT_DIR "${CMAKE_CURRENT_SOURCE_DIR}/torch_mlir")
 # We vendor our own MLIR instance in the `torch_mlir` namespace.
 add_compile_definitions("MLIR_PYTHON_PACKAGE_PREFIX=torch_mlir.")
 
+################################################################################
+# PyTorch
+################################################################################
+
+option(TORCH_MLIR_USE_INSTALLED_PYTORCH "Build from local PyTorch in environment" ON)
+
+if (NOT TORCH_MLIR_USE_INSTALLED_PYTORCH)
```
henrytwo, line 22: This got moved up here, since the LTC backend will also rely on `TORCH_MLIR_USE_INSTALLED_PYTORCH`, in addition to the existing JIT IR

build_tools/autogen_ltc_backend.py
```diff
@@ -236,6 +236,11 @@ def get_opnames(ops):
                 continue
             if base in supported or op in supported:
                 continue
+            # Blacklist new_/_like ops since they are non-differentiable.
+            if re.search('^new_', base) or re.search('^new_', op):
+                continue
+            if re.search('_like$', base) or re.search('_like$', op):
+                continue
```
antoniojkim, line 243: I think this can be simplified a bit
```suggestion
            if any(o.startswith("new_") or o.endswith("_like") for o in (base, op)):
                continue
```

henrytwo:
image
The failure from source build seems to be upstream related? I can see that it exists on main too

vivekkhandelwal1:
> image The failure from source build seems to be upstream related? I can see that it exists on main too
> 
> Based on the timing and errors, I think this is the root cause: [pytorch/pytorch@2bfae07](https://github.com/pytorch/pytorch/commit/2bfae07a796187b047c6113b46d0aabbf319d990)
> 
> We should still be good to merge though, since we saw source build succeed earlier prior to this PyTorch update

Yes, it's failing due to upstream PyTorch changes. https://github.com/llvm/torch-mlir/issues/1121

python/CMakeLists.txt
```diff
@@ -60,31 +82,17 @@ declare_mlir_python_extension(TorchMLIRPythonExtensions.Main
 # Lazy Tensor Core
 ################################################################################
 
-add_subdirectory(torch_mlir/csrc/base_lazy_backend)
+if(TORCH_MLIR_ENABLE_LTC)
+  add_subdirectory(torch_mlir/csrc/base_lazy_backend)
+endif()
+# Reference backend has a separate check for TORCH_MLIR_ENABLE_LTC.
```
antoniojkim, line 88: why have the separate check? Why not consolidate the check here and just not add the subdirectory if not enabled?

henrytwo: I feel like it's cleaner this way, because we need to pull some variables specific to the reference backend for that to work

henrytwo:
Looks like the latest failures are due to some changes on the side of Torch-MLIR -- fix coming shortly

powderluv:
Ok to land as long as CI is green

henrytwo:
> Ok to land as long as CI is green

Thanks, can we get this PR stamped as approved then? I'll land this in tonight once CI passes

silvasean (approved):
:rocket:

PR #1113, opened 2022-07-26
[MLIR][TORCH] Add decomposition for aten.std.dim op

PhaneeshB:
Signed-Off By: Phaneesh Barwaria <phaneesh@nod-labs.com>

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -1216,6 +1216,30 @@ public:
 };
 } // namespace
 
+// Decompose aten.std.dim to sqrt(var.dim(x))
+namespace {
+class DecomposeAtenStdDimOp : public OpRewritePattern<AtenStdDimOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenStdDimOp op,
+                                PatternRewriter &rewriter) const override {
+    Value self = op.self();
+    BaseTensorType inputTensorType = self.getType().cast<BaseTensorType>();
+    if (!inputTensorType.hasDtype() ||
+        !inputTensorType.getDtype().isa<mlir::FloatType>()) {
+      return rewriter.notifyMatchFailure(
+          op, "aten.std.dim expects input tensor of floating-point type");
+    }
+
+    Value varDim =
+        rewriter.create<AtenVarDimOp>(op->getLoc(), op.getType(), self,
+                                      op.dim(), op.unbiased(), op.keepdim());
+    rewriter.replaceOpWithNewOp<AtenSqrtOp>(op, op.getType(), varDim);
+    return success();
+  }
+};
+} // namespace
+
 // Hardsigmoid(x) = max(0, min(1, (x+3)/6))
 namespace {
 class DecomposeAtenHardsigmoidOp : public OpRewritePattern<AtenHardsigmoidOp> {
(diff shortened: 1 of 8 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1206,6 +1206,31 @@ class DecomposeAtenSoftplusOp : public OpRewritePattern<AtenSoftplusOp> {
...
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenStdDimOp op,
+		                PatternRewriter &rewriter) const override {
+    Value self = op.self();
+    BaseTensorType inputTensorType = self.getType().cast<BaseTensorType>();
+    if (!inputTensorType.hasDtype() ||
+        !inputTensorType.getDtype().isa<mlir::FloatType>()) {
+      return rewriter.notifyMatchFailure(op,
+                                         "aten.std.dim only supports floating type");
```
dan-garvey, line 1221: nitpicky comment: I understand it is this way in many other places in the file, but I think a better error message here is something like:
aten.std.dim expects inputTensorType to be an mlir::FloatType. 

or at the very least change to "floating-point type"

dan-garvey (approved):
Looks great to me, may want one more reviewer who more frequently reviews these.

python/torch_mlir_e2e_test/test_suite/stats.py
```diff
@@ -256,6 +256,50 @@ def StdBiasedModule_basic(module, tu: TestUtils):
...
+
+
+@register_test_case(module_factory=lambda: StdDimKeepDimFalseModule())
+def StdDimKeepDimTrueModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5))
+
+
+# ==============================================================================
+
+
```
vivekkhandelwal1, line 302: You should add a test case for `unbiased=False`, because by default it's true.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1206,6 +1206,31 @@ class DecomposeAtenSoftplusOp : public OpRewritePattern<AtenSoftplusOp> {
...
+    }
+
+
```
vivekkhandelwal1, line 1224: Remove this line.

vivekkhandelwal1 (changes requested):

pashu123 (approved):
Add a test case with `unbiased=False` as @vivekkhandelwal1 suggested, rest LGTM!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
PhaneeshB: Changed the error message to `aten.std.dim expects input tensor of floating-point type`

python/torch_mlir_e2e_test/test_suite/stats.py
PhaneeshB: Added a test case for `unbiased=False`.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
PhaneeshB: Line removed

vivekkhandelwal1 (changes requested):
@PhaneeshB Please resolve the conflicts.

PR #1092, opened 2022-07-21
[MHLO] Init MHLO basic op conversion

Vremold:
See  RFC #999 
Co-authored-by: Bairen Yi <yibairen.byron@bytedance.com>
Co-authored-by: Jiawei Wu <xremold@gmail.com>
Co-authored-by: Tianyou Guo <tianyou.gty@alibaba-inc.com>
Co-authored-by: Xu Yan <yancey.yx@alibaba-inc.com>
Co-authored-by: Ziheng Jiang <ziheng.jiang@bytedance.com>

PR #1116, opened 2022-07-27
[MHLO] Add support for dynamic shape in basic op conversion

Vremold:
See RFC #999 

This a patch of PR #1092 . In this patch, we import shape dialect to support dynamic shape when lowering Torch ops to Mhlo ops. 

Co-authored-by: Bairen Yi <yibairen.byron@bytedance.com>
Co-authored-by: Jiawei Wu <xremold@gmail.com>
Co-authored-by: Tianyou Guo <tianyou.gty@alibaba-inc.com>
Co-authored-by: Xu Yan <yancey.yx@alibaba-inc.com>
Co-authored-by: Ziheng Jiang <ziheng.jiang@bytedance.com>

### DIFF
--- a/lib/Conversion/TorchToMhlo/BasicOp.cpp
+++ b/lib/Conversion/TorchToMhlo/BasicOp.cpp
@@ -13,6 +13,8 @@
 #include "./MhloLegalizeUtils.h"
 #include "./PopulatePatterns.h"
 #include "mlir-hlo/Dialect/mhlo/IR/hlo_ops.h"
+#include "mlir/Dialect/Arithmetic/IR/Arithmetic.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "torch-mlir/Conversion/Utils/Utils.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchDialect.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
@@ -125,11 +127,12 @@ public:
 
 } // namespace
 
-// These binary op legalizations are specific to add/sub which have an
+// These binary op legalizations are specific to add/sub(Tensor) which have an
 // alpha multiplier.
+// aten.add(sub).Tensor
 namespace {
 template <typename AtenOpT, typename MhloOpT>
-class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
+class ConvertAtenAddSubTensorOp : public OpConversionPattern<AtenOpT> {
 public:
   using OpConversionPattern<AtenOpT>::OpConversionPattern;
   using OpAdaptor = typename AtenOpT::Adaptor;
@@ -137,16 +140,16 @@ public:
   matchAndRewrite(AtenOpT op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     Value lhs = adaptor.self();
-    TensorType lhsType = lhs.getType().dyn_cast<TensorType>();
+    RankedTensorType lhsType = lhs.getType().dyn_cast<RankedTensorType>();
     Value rhs = adaptor.other();
-    TensorType rhsType = rhs.getType().dyn_cast<TensorType>();
+    RankedTensorType rhsType = rhs.getType().dyn_cast<RankedTensorType>();
 
-    if (!lhsType)
+    if (!lhsType || !rhsType)
       return op.emitError("Only Tensor types supported in MHLO");
 
-    TensorType outType = OpConversionPattern<AtenOpT>::getTypeConverter()
-                             ->convertType(op.getType())
-                             .template cast<TensorType>();
+    RankedTensorType outType = OpConversionPattern<AtenOpT>::getTypeConverter()
+                                   ->convertType(op.getType())
+                                   .template cast<RankedTensorType>();
 
     Type outElemTy = outType.getElementType();
     if (!outElemTy.isIntOrFloat()) {
@@ -154,50 +157,56 @@ public:
           "Only floating-point or integer datatype legalization supported");
     }
 
-    Value rhsAsTensor;
-    if (!rhsType) {
-      if (failed(mhlo::torchScalarToMhloTensor(rewriter, op, op.other(),
-                                               rhsAsTensor, outElemTy,
-                                               outType.getShape())))
-        return op.emitError("Currently only scalar constants are supported for "
-                            "conversion in MHLO operation");
-    }
-    Value lhsTensor = lhs;
-    Value rhsTensor = rhsType ? rhs : rhsAsTensor;
+    Value bcastShape;
 
     // Handle broadcasting. Since we have the output type already, here we
     // just broodcast operands' shape to output shape.
-    lhsTensor = mhlo::promoteAndBroadcast(rewriter, lhsTensor, outType);
-    rhsTensor = mhlo::promoteAndBroadcast(rewriter, rhsTensor, outType);
+    if (lhsType.hasStaticShape() && rhsType.hasStaticShape()) {
+      lhs = mhlo::promoteAndBroadcast(rewriter, lhs, outType);
+      rhs = mhlo::promoteAndBroadcast(rewriter, rhs, outType);
+    } else {
+      lhs = mhlo::promoteType(rewriter, lhs, outType);
+      rhs = mhlo::promoteType(rewriter, rhs, outType);
+      mhlo::broadcastBinaryOperandsDynamic(rewriter, op, lhs, rhs, outType, lhs,
+                                           rhs, bcastShape);
+    }
 
     // Handle alpha.
     Value multTensor;
     if (skipMultiplyAlpha(op.alpha())) {
-      multTensor = rhsTensor;
+      multTensor = rhs;
     } else {
       Value alphaTensor;
       if (failed(mhlo::torchAlphaToMhloTensor(rewriter, op.getOperation(),
                                               op.alpha(), alphaTensor,
-                                              outElemTy, outType.getShape(),
+                                              outElemTy, {},
                                               /*checkForUnity=*/false))) {
         return op.emitError("Currently only scalar constants are supported for "
                             "alpha in conversion to MHLO operation");
       }
-
-      multTensor = rewriter.create<mhlo::MulOp>(op.getLoc(), outType, rhsTensor,
-                                                alphaTensor);
+      if (outType.hasStaticShape()) {
+        alphaTensor = mhlo::broadcastUnaryOperandStatic(rewriter, op,
+                                                        alphaTensor, outType);
+      } else {
+        alphaTensor = rewriter.create<mhlo::DynamicBroadcastInDimOp>(
+            op->getLoc(), outType, alphaTensor, bcastShape,
+            rewriter.getI64TensorAttr({}));
+      }
+      multTensor =
+          rewriter.create<mhlo::MulOp>(op.getLoc(), outType, rhs, alphaTensor);
     }
-
-    rewriter.replaceOpWithNewOp<MhloOpT>(op, outType, lhsTensor, multTensor);
+    rewriter.replaceOpWithNewOp<MhloOpT>(op, outType, lhs, multTensor);
     return success();
   }
 };
 } // namespace
 
-// Binary op legalizations for Mul variants.
+// These binary op legalizations are specific to add/sub(Scalar) which have an
+// alpha multiplier.
+// aten.add(sub).Scalar
 namespace {
-template <typename AtenOpT>
-class ConvertAtenMulOp : public OpConversionPattern<AtenOpT> {
+template <typename AtenOpT, typename MhloOpT>
+class ConvertAtenAddSubScalarOp : public OpConversionPattern<AtenOpT> {
 public:
   using OpConversionPattern<AtenOpT>::OpConversionPattern;
   using OpAdaptor = typename AtenOpT::Adaptor;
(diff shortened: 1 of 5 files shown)


Vremold:
Ready for review. @silvasean @powderluv @fortianyou . Thanks.

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -125,180 +127,219 @@ class ConvertAtenConstPatternOp : public OpConversionPattern<AtenOpT> {
...
-    }
-    Value lhsTensor = lhs;
-    Value rhsTensor = rhsType ? rhs : rhsAsTensor;
+    Value bcastShape;
 
     // Handle broadcasting. Since we have the output type already, here we
     // just broodcast operands' shape to output shape.
-    lhsTensor = mhlo::promoteAndBroadcast(rewriter, lhsTensor, outType);
-    rhsTensor = mhlo::promoteAndBroadcast(rewriter, rhsTensor, outType);
+    if (lhsType.hasStaticShape() && rhsType.hasStaticShape()) {
```
silvasean, line 164: why do we need broadcastBinaryOperandsDynamic separate from promoteAndBroadcast? In the static case shouldn't they canonicalize to the same thing?

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -125,180 +127,219 @@ class ConvertAtenConstPatternOp : public OpConversionPattern<AtenOpT> {
...
-                                              outElemTy, outType.getShape(),
+                                              outElemTy, {},
                                               /*checkForUnity=*/false))) {
         return op.emitError("Currently only scalar constants are supported for "
                             "alpha in conversion to MHLO operation");
       }
-
-      multTensor = rewriter.create<mhlo::MulOp>(op.getLoc(), outType, rhsTensor,
-                                                alphaTensor);
+      if (outType.hasStaticShape()) {
```
silvasean, line 187: same comment here

silvasean (commented):
It would be good if you could handle the static shape as a special case of dynamic, rather than separate code paths when lowering -- is there something preventing that?

silvasean:
Will let @fortianyou give the final LGTM.

lib/Conversion/TorchToMhlo/BasicOp.cpp
Vremold: Yeah, we have tried to handle static shape as a special case of dynamic. But supporting dynamic shape will import shape ops and more nested blocks. We are worrying that it probably does harm to the efficiency when lowering static shaped ops.

fortianyou: Perhaps we can use `mhlo::DimOp` && `tensor::FromElements` instead of `shape::Dialect`?

fortianyou: For example, this is the `AtenBroadcastToOp` currently implemented in BladeDISC:
```
template <>
LogicalResult ConvertAtenOp<AtenBroadcastToOp>::matchAndRewrite(
    AtenBroadcastToOp op,
    OpAdaptor adaptor,
    ConversionPatternRewriter& rewriter) const {
  // Not a tensor type.
  //
  auto selfType = adaptor.self().getType().dyn_cast<TensorType>();
  if (!selfType)
    return op.emitError("Only tensor types are currently supported");

  SmallVector<Value> dimsSize;
  if (!getListConstructElements(adaptor.size(), dimsSize)) {
    return op.emitError("Dims size must be a list of Scalar");
  }

  auto loc = op.getLoc();
  auto rankType = selfType.dyn_cast<RankedTensorType>();
  auto selfRank = rankType ? rankType.getRank() : 0;
  auto newRank = dimsSize.size();
  auto leadingRank = newRank - selfRank;
  for (size_t d = 0; d < newRank; ++d) {
    // !torch.int
    auto dsize = dimsSize[d];
    int64_t dval;
    if (matchPattern(dsize, m_TorchConstantInt(&dval)) && dval == -1) {
      if (d < leadingRank) {
        return op.emitError(
            "Ref: https://pytorch.org/docs/stable/generated/torch.Tensor.expand.html."
            "For the new leading dimensions, the size cannot be set to -1.");
      } else {
        // Passing -1 as the size for a dimension means not changing the size
        // of that dimension.
        //
        // tensor.dim %self -> index
        dsize = rewriter.create<tensor::DimOp>(
            loc, adaptor.self(), d - leadingRank);
      }
    } else {
      // !torch.int -> i64
      dsize = rewriter.create<ToI64Op>(loc, dsize).getResult();
      // i64 -> index
      dsize = rewriter.create<mlir::arith::IndexCastOp>(
          loc, rewriter.getIndexType(), dsize);
    }
    // index -> i32
    dsize = rewriter.create<mlir::arith::IndexCastOp>(
        loc, rewriter.getI32Type(), dsize);
    dimsSize[d] = dsize;
  }

  auto mhloShape = rewriter.create<mlir::tensor::FromElementsOp>(loc, dimsSize);
  auto broadcastDims =
      BuildI64ElementsAttr(rewriter, mhlo::rangeIndices(leadingRank, newRank));
  rewriter.replaceOpWithNewOp<mhlo::DynamicBroadcastInDimOp>(
      op,
      getTypeConverter()->convertType(op.getType()),
      adaptor.self(),
      mhloShape,
      broadcastDims);
  return success();
}
```

fortianyou: Another bonus is that we cat avoid introducing shape dialect.

Vremold: As we want to check whether two dynamic shaped tensor is broadcastable, shape ops is easier to use than `tensor.DimOp` & `tensor.FromElementsOp`

lib/Conversion/TorchToMhlo/MhloLegalizeUtils.cpp
```diff
@@ -265,23 +262,97 @@ LogicalResult torchAlphaToMhloTensor(ConversionPatternRewriter &rewriter,
...
+}
+
+Value broadcastUnaryOperandStatic(PatternRewriter &rewriter, Operation *op,
+                                  Value input, TensorType outType) {
+
+  auto inputType = input.getType().dyn_cast<TensorType>();
+  if (!inputType) {
+    op->emitError("Only ranked tensor type input is supported!");
+  }
+  if (!inputType.hasStaticShape() || !outType.hasStaticShape()) {
```
qiuxiafei, line 344: if `inputType` is null, may it corrupt at this line?

lib/Conversion/TorchToMhlo/MhloLegalizeUtils.cpp
```diff
@@ -265,23 +262,97 @@ LogicalResult torchAlphaToMhloTensor(ConversionPatternRewriter &rewriter,
...
+// Broadcast the input tensor of 0 rank to a tensor with the same shape as the
+// given tensor. Supporting dynamic shape.
+Value broadcastZeroRankTensorToTensorLike(PatternRewriter &rewriter,
+                                          Operation *op, Value input,
+                                          Value tensor, Type elementType) {
+  Location loc = op->getLoc();
+  auto inputType = input.getType().cast<RankedTensorType>();
+  auto tensorType = tensor.getType().cast<RankedTensorType>();
+  if (inputType.getRank() != 0) {
+    op->emitError("Only zero ranked tensor is supported");
```
fortianyou, line 304: Either rewriter.notifyMatchFailure if this is a limitation of the pattern or the check should be in the verifier of the op.
Could the return type be FailureOr<...> ? cc @jpienaar

lib/Conversion/TorchToMhlo/MhloLegalizeUtils.cpp
```diff
@@ -265,23 +262,97 @@ LogicalResult torchAlphaToMhloTensor(ConversionPatternRewriter &rewriter,
...
+  }
+  return input;
+}
+
+Value broadcastUnaryOperandStatic(PatternRewriter &rewriter, Operation *op,
+                                  Value input, TensorType outType) {
+
+  auto inputType = input.getType().dyn_cast<TensorType>();
+  if (!inputType) {
+    op->emitError("Only ranked tensor type input is supported!");
```
fortianyou, line 342: ditto, and other elsewhere

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -125,180 +127,219 @@ class ConvertAtenConstPatternOp : public OpConversionPattern<AtenOpT> {
...
-                                                 outType.getShape())))
-          return op.emitError(
-              "Currently only scalar constants are supported for "
-              "conversion in MHLO operation");
-      } else {
-        rhsTensor = rhs;
+    Value rhs;
+    if (failed(mhlo::torchScalarToMhloTensor(rewriter, op, op.other(), rhs,
+                                             outElemTy, {}, false)))
+      return op.emitError("Currently only scalar constants are supported for "
```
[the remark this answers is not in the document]
Vremold: Ok, we will fix these wrongly-formatted error messages.

lib/Conversion/TorchToMhlo/BasicOp.cpp
fortianyou: Then I would prefer `chlo::BroadcastBinOp` then shape dialect

lib/Conversion/TorchToMhlo/BasicOp.cpp
```diff
@@ -438,12 +489,51 @@ class ConvertAtenBroadcastToOp : public OpConversionPattern<AtenBroadcastToOp> {
   matchAndRewrite(AtenBroadcastToOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
     Value self = adaptor.self();
+    auto selfTy = self.getType().cast<RankedTensorType>();
     auto outType = getTypeConverter()
                        ->convertType(op->getResult(0).getType())
                        .cast<RankedTensorType>();
+    if (selfTy.hasStaticShape()) {
```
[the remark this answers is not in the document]
Vremold: That's good idea! We will try this out after deciding whether to separate codes into different paths according to static and dynamic shapes.

lib/Conversion/TorchToMhlo/MhloLegalizeUtils.cpp
Vremold: Good suggestion, we will change the codes.

lib/Conversion/TorchToMhlo/BasicOp.cpp
Vremold: Since our target is MHLO dialect, introducing CHLO dialect probably needs an additional pipeline to lowering CHLO to MHLO. We are trying to avoid this.

lib/Conversion/TorchToMhlo/MhloLegalizeUtils.cpp
Vremold: Thanks, we forget some "return" sentences here.

lib/Conversion/TorchToMhlo/BasicOp.cpp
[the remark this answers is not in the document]
jpienaar: Thanks!

lib/Conversion/TorchToMhlo/BasicOp.cpp
Vremold: > why do we need broadcastBinaryOperandsDynamic separate from promoteAndBroadcast? In the static case shouldn't they canonicalize to the same thing?
> Yes, I think a better way to support static shapes is to do canonicalization in some passes for static shapes. And static shapes from ShapeType should not be used by default.

Thanks a lot. We will unify the implementation of static shaped and dynamic shaped ops' lowering. And let the canonicalizer to eliminate unnecessary shape ops when involved tensors are totally static shaped.

jpienaar: This is a very good discussion, if you discover some ergonomic/representation/abstraction issues here please let us know. These op sets are not set in stone (some more difficult to change, but not infeasible :-)) and feedback very important to being able to improve them.

ZihengJiang: @fortianyou @Vremold Introducing `chlo` is ok for me, (which might ease our development for dynamic cases), as long as we add a chlo_legalize_to_hlo transform in our lowering pipeline. `ShapeDialect` should be allowed in necessary situation (if chlo is not strong enough), since the `chlo_legalize_to_hlo` also introduces it. Does this sound good to you guys?

Vremold: > @fortianyou @Vremold Introducing `chlo` is ok for me, (which might ease our development for dynamic cases), as long as we add the `chlo_legalize_to_hlo` transform in our lowering pipeline. `ShapeDialect` should be allowed in necessary situation (if chlo is not strong enough), since the `chlo_legalize_to_hlo` also introduces it. Does this sound good to you guys?

Okay for me.

Vremold: It seems that an agreement is reached. I will close this PR and switch to use CHLO dialect in `convert-torch-to-mhlo` pass to handle dynamic cases. While in the whole `torch-backend-to-mhlo-backend-pipeline` pipeline, we plan to reuse `chlo_legalize_to_hlo` pass in mlir-hlo community to finally convert CHLO ops to MHLO ops, which might introduce a little change in current CMake files. What do you think of this plan? @silvasean @ZihengJiang @fortianyou

ZihengJiang: Sounds good. Thanks for the effort :)

silvasean: Sounds good! Great discussion!

fortianyou: Great! Also, I think we can make chlo_legalize_to_hlo configurable :)

PR #1117, opened 2022-07-28
[MLIR][TORCH] Add e2e support for aten.atan2

qedawkins:
Includes an additional conversion from math to libm in the reference backend needed to handle math::atan2.


qedawkins:
This patch adds a pass to the refbackend based on the response found in a related [LLVM discourse thread](https://discourse.llvm.org/t/missing-llvm-lowering-for-math-atan2op/64096/5).

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -722,7 +722,7 @@ void TypeAnalysis::visitOperation(Operation *op,
 
   // Promote the two dtypes assuming possibly-zero rank.
   if (isa<AtenAddTensorOp, AtenSubTensorOp, AtenMulTensorOp, AtenDivTensorOp,
-          AtenDivTensorModeOp, Aten__And__TensorOp, AtenMinimumOp,
+          AtenDivTensorModeOp, Aten__And__TensorOp, AtenMinimumOp, AtenAtan2Op,
```
ramiro050, line 725: Why are there two cases for `AtenAtan2Op` in this file? This makes it a bit harder to know how a single op is being handled

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/RefineTypes.cpp
qedawkins: Oh I forgot to remove the first one.

qedawkins:
Can I also get confirmation that the way I added a pass to the refbackend is correct?

ramiro050:
> Can I also get confirmation that the way I added a pass to the refbackend is correct?

@silvasean PTAL

silvasean:
refbackend lgtm

ramiro050 (approved):

PR #1119, opened 2022-07-28
E2E support for AtenRemainderScalarOp

vid-999:
adding support for AtenRemdinderScalarOp needed for dlrm.


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -791,6 +791,21 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    Type newResultType = converter->convertType(remScalar.getType())
+                             .cast<RankedTensorType>()
+                             .getElementType();
+    Type floatDtype = mlir::FloatType::getF64(context);
+    Value self = convertScalarToDtype(b, loc, payloadArgs[0], floatDtype);
+    Value other = convertScalarToDtype(b, loc, operands[1], floatDtype);
+    Value division = b.create<arith::DivFOp>(loc, self, other);
+    Value floor = b.create<math::FloorOp>(loc, division);
+    Value multiply = b.create<arith::MulFOp>(loc, floor, other);
+    Value subtract = b.create<arith::SubFOp>(loc, self, multiply);
```
ramiro050, line 805: I think you can use `arith::remf` and `arith::remi` https://mlir.llvm.org/docs/Dialects/ArithmeticOps/#arithremf-mlirarithremfop

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -791,6 +791,21 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
     return b.create<arith::DivFOp>(loc, self, other);
   }
+  if (auto remScalar = dyn_cast<AtenRemainderScalarOp>(op)) {
+    MLIRContext *context = op->getContext();
+    Type newResultType = converter->convertType(remScalar.getType())
+                             .cast<RankedTensorType>()
+                             .getElementType();
+    Type floatDtype = mlir::FloatType::getF64(context);
+    Value self = convertScalarToDtype(b, loc, payloadArgs[0], floatDtype);
+    Value other = convertScalarToDtype(b, loc, operands[1], floatDtype);
```
ramiro050, line 801: I think it would be better to not cast to f64. When you cast an integer to a float, you run the risk of losing some information. What I would do is cast both `self` and `other` to `resultType`, and then have an `if-else` depending on if the `resultType` is and integer or a float

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -931,14 +946,14 @@ class ConvertElementwiseOp : public ConversionPattern {
...
-             AtenNeScalarOp, AtenNegOp, AtenMaskedFillScalarOp, AtenLogicalOrOp,
-             AtenTriuOp>(op))
+             AtenLog1pOp, AtenRsqrtOp, AtenDivScalarOp, AtenRemainderScalarOp,
+             AtenAbsOp, AtenReciprocalOp, AtenBitwiseAndTensorOp,
+             AtenGtScalarOp, AtenGeScalarOp, AtenEqScalarOp, AtenLtScalarOp,
+             AtenLeScalarOp, AtenWhereSelfOp, AtenCeilOp, AtenGtTensorOp,
+             AtenEqTensorOp, AtenLtTensorOp, AtenSubScalarOp, AtenAddScalarOp,
+             AtenThresholdOp, AtenThresholdBackwardOp, AtenCloneOp, AtenSinOp,
+             AtenCosOp, AtenNeScalarOp, AtenNegOp, AtenMaskedFillScalarOp,
+             AtenLogicalOrOp, AtenTriuOp>(op))
```
ramiro050, line 956: You also have to add the op to the list of illegal ops

https://github.com/llvm/torch-mlir/blob/a06c582ff8fa8621166e823232ef50d7b0db8426/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L1683

python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -1233,6 +1233,27 @@ def forward(self, x):
...
+
+class ElementwiseRemainderScalarModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1], torch.int64, True),
```
ramiro050, line 1247: Once you add the logic for making the calculation with ints, also add an e2e test that checks that path

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -791,6 +791,26 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    Value self = convertScalarToDtype(b, loc, payloadArgs[0], newResultType);
+    Value other = convertScalarToDtype(b, loc, operands[1], newResultType);
+    Value result; 
+
+    if (newResultType.isa<mlir::FloatType>()){
+      result = b.create<arith::RemFOp>(loc, self, other);
+    }
+    else{
+      result = b.create<arith::RemSIOp>(loc, self, other);
+    }
```
ramiro050, line 809: nit: make sure to run `git clang-format`

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -791,6 +791,26 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+
+    result = convertScalarToDtype(b, loc, result, newResultType);
```
ramiro050, line 811: Is this extra cast needed?

python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -1233,6 +1233,70 @@ def forward(self, x):
 def ElementwiseDivScalarModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 4))
 
+# ==============================================================================
+
+
+class ElementwiseRemainderScalarModule_Int(torch.nn.Module):
```
ramiro050, line 1239: nit: I would call this test `Elementwise..._Int_Float` and the other int one as `Elementwise..._Int`

ramiro050 (approved):
Looks good, just a few small comments/questions

lib/Conversion/TorchToLinalg/Uncategorized.cpp
vid-999: I don't think its needed i forgot to remove it will do that now.

ramiro050 (approved):

vivekkhandelwal1:
@vid-999 We don't merge the PRs until the CI is green. Your patch was failing some tests, and because of this patch getting merged the CI is failing for all other patches.

vid-999:
> @vid-999 We don't merge the PRs until the CI is green. Your patch was failing some tests, and because of this patch getting merged the CI is failing for all other patches.

apologies for that, for some reason, i though they all passed.

Issue #1121, opened 2022-07-29
CI Failure due to change in upstream definition of `mean.dim` op


vivekkhandelwal1:
The CI is failing due to changes in the `mean.dim` op in upstream PyTorch https://github.com/pytorch/pytorch/commit/2bfae07a796187b047c6113b46d0aabbf319d990. Error:
```
Unexpected outcome summary:

****** Failed tests - 1 tests
    FAIL - "MeanDimDtypeModule_basic"
        Compilation error: Traceback (most recent call last):
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 282, in compile_and_run_test
            golden_trace = generate_golden_trace(test)
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 276, in generate_golden_trace
            test.program_invoker(tracer, TestUtils())
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/stats.py", line 105, in MeanDimDtypeModule_basic
            module.forward(tu.rand(3, 4, 5).to(torch.float64))
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 256, in __call__
            output = self.__wrapped__(*args, **kwargs)
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/stats.py", line 100, in forward
            return torch.ops.aten.mean(x, 0, dtype=torch.float32)
          File "/opt/hostedtoolcache/Python/3.9.13/x64/lib/python3.9/site-packages/torch/_ops.py", line 164, in __call__
            return self._op(*args, **kwargs or {})
        RuntimeError: Overloaded torch operator invoked from Python failed to many any schema:
        aten::mean() expected at most 2 argument(s) but received 3 argument(s). Declaration: aten::mean(Tensor self, *, int? dtype=None) -> Tensor

        aten::mean() Expected a value of type 'Optional[List[int]]' for argument 'dim' but instead found type 'int'.
        Position: 1
        Value: 0
[...]

PR #1122, opened 2022-07-29
Fix CI failure due to upstream PyTorch change in aten.mean.dim op

vivekkhandelwal1:
Fixes https://github.com/llvm/torch-mlir/issues/1121


pashu123 (approved):
LGTM!

ramiro050:
Thanks @vivekkhandelwal1 for the quick fix! I would've expected this change in upstream to also change the ODS generated for the `aten.mean.dim` op in torch-mlir. Is this not the case?

PR #1137, opened 2022-08-02
Add decomposition and E2E support for Aten_EmbeddingBag

vid-999:


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2405,6 +2403,42 @@ class DecomposeAtenSelectScatterOp
...
+    Value per_sample_weights = op.per_sample_weights();
+    Value include_last_offset = op.include_last_offset();
+    Value padding_idx = op.padding_idx();
+
+    auto resultType0 = op->getResult(0).getType();
+    auto resultType1 = op->getResult(1).getType();
+    auto resultType2 = op->getResult(2).getType();
+    auto resultType3 = op->getResult(3).getType();
+
+    mlir::TypeRange returnTypes = {resultType0, resultType1, resultType2,
```
ramiro050, line 2430: nit: no need for `=`. You can do `... returnTypes{resultType0,...}`

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -255,8 +254,7 @@ class DecomposeAtenNarrowOp : public OpRewritePattern<AtenNarrowOp> {
 } // namespace
 
 namespace {
-class DecomposeAtenZeroOp
-    : public OpRewritePattern<AtenZeroOp> {
+class DecomposeAtenZeroOp : public OpRewritePattern<AtenZeroOp> {
```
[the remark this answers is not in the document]
vid-999: hmm yeah that seems like the case. Cool makes sense. I'll fix this.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1058,6 +1058,23 @@ void TypeAnalysis::visitOperation(Operation *op,
     return;
   }
 
+  if (auto _embedding_bag = dyn_cast<Aten_EmbeddingBagOp>(op)) {
```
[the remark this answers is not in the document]
vid-999: cool makes sense.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1151,6 +1142,23 @@ void TypeAnalysis::visitAtenLinearOp(AtenLinearOp op,
   incorporateKnowledge(op->getResult(0), knowledge);
 }
 
+template <typename OpTy> void TypeAnalysis::visitEmbeddingBagOp(OpTy op) {
```
ramiro050, line 1145: This doesn't need to be a template. The body of this function works if the input of the function is an `Operation *op`. Removing the `template` also allows you to simplify the code above. You should be able to do:

```c++
if (isa<Aten_EmbeddingBagOp, AtenEmbeddingBagPaddingIdxOp>(op)) {
  visit...;
return;
}
```

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -979,6 +979,25 @@ def aten〇embedding_bag〇padding_idx(weight: List[int], indices: List[int], of
 
     return output_bag_shape, offset2bag_shape, bag_size_shape, max_indices_shape
 
+def aten〇_embedding_bag(weight: List[int], indices: List[int], offsets: List[int], scale_grad_by_freq: bool = False, mode: int = 0, sparse: bool = False, per_sample_weights: Optional[List[int]] = None, include_last_offset: bool = False, padding_idx: int = -1) -> Tuple[List[int], List[int], List[int], List[int]]:
```
ramiro050, line 982: Is there a reason this shape function is simpler than the `embedding_bag_padding_idx` one?

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
vid-999: yes since the mode is always = 0 for this and other some other parameters are constants..

lib/Dialect/Torch/Transforms/RefineTypes.cpp
vid-999: makes sense. will make a change

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
ramiro050: Isn't `mode` also a parameter for this op? In the signature of the shape function it is equal to `0` as the default value, but in general it could have any of the three modes as its value. The same is true for all other parameters

vid-999: I wasn't sure its it was a parameter or a fixed value. If its a parameter I can reuse the same shape function I think

ramiro050: Yeah, that would be great! Just refactor into a helper python function, then you can just use it for both

vid-999: yup makes sense

ramiro050 (approved):
LGTM!

ramiro050:
One of the CI's is failing because you need to add your e2e test to the `XFAIL` list of LTC

vid-999:
> One of the CI's is failing because you need to add your e2e test to the XFAIL list of LTC

added.

PR #1138, opened 2022-08-02
Fix cache-suffix name bug

powderluv:
This should enabling better caching of builds.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -42,7 +42,7 @@ jobs:
         submodules: 'true'
     - uses: ./.github/actions/setup-build
       with:
-        cache-suffix: ${{ matrix.os }}-${{ matrix.targetarch }}-${{ matrix.buildtype }}
+        cache-suffix: ${{ matrix.os }}-${{ matrix.targetarch }}-${{ matrix.llvmtype }}-${{ matrix.llvmbuildtype }}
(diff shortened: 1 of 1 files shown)


makslevental (commented):
LGTM

powderluv:
Do you know what would cause 
```
PASS - "_SoftmaxModule_basic"

Unexpected outcome summary:

****** Failed tests - 1 tests
    FAIL - "Matmul_vecmat"
        @ output of call to "forward"
        ERROR: shape (torch.Size([1, 5])) is not equal to golden shape (torch.Size([5]))


Summary:
    Passed: 653
    Failed: 1
```

In the eager mode tests ?

makslevental:
> Do you know what would cause
> 
> ```
> PASS - "_SoftmaxModule_basic"
> 
> Unexpected outcome summary:
> 
>     FAIL - "Matmul_vecmat"
>         ERROR: shape (torch.Size([1, 5])) is not equal to golden shape (torch.Size([5]))
> 
> 
> Summary:
>     Passed: 653
>     Failed: 1
> ```
> In the eager mode tests ?

@powderluv 

I expect some type system shenanigans symbolic trace sees a `d` element vector but it actually gets turned into a `1*d` matrix for this particular op. Will take a look soon.

powderluv:
CI fix has landed on master.

PR #1139, opened 2022-08-02
Re-enable custom op support

nithinsubbiah:
Re-enable custom op support (https://github.com/llvm/torch-mlir/pull/895) that was reverted due to MacOS builder issue.

### DIFF
--- a/build_tools/update_shape_lib.sh
+++ b/build_tools/update_shape_lib.sh
@@ -16,19 +16,17 @@ build_dir="$(realpath "${TORCH_MLIR_BUILD_DIR:-$src_dir/build}")"
 torch_transforms_cpp_dir="${src_dir}/lib/Dialect/Torch/Transforms"
 python_packages_dir="${build_dir}/tools/torch-mlir/python_packages"
 
+TORCH_MLIR_EXT_PYTHONPATH="${TORCH_MLIR_EXT_PYTHONPATH:-""}"
 pypath="${python_packages_dir}/torch_mlir"
-# TODO: Re-enable once custom op support is back.
-#if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
-#  pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
-#fi
-#ext_module="torch_mlir._torch_mlir_custom_op_example"
-#if [ ! -z ${TORCH_MLIR_EXT_MODULES} ]; then
-#  ext_module="${ext_module},${TORCH_MLIR_EXT_MODULES} "
-#fi
+if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
+  pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
+fi
+TORCH_MLIR_EXT_MODULES="${TORCH_MLIR_EXT_MODULES:-""}"
+if [ ! -z ${TORCH_MLIR_EXT_MODULES} ]; then
+  ext_module="${TORCH_MLIR_EXT_MODULES} "
+fi
 
 PYTHONPATH="${pypath}" python \
   -m torch_mlir.dialects.torch.importer.jit_ir.build_tools.shape_lib_gen \
+  --pytorch_op_extensions=${ext_module} \
   --torch_transforms_cpp_dir="${torch_transforms_cpp_dir}"
-
-# TODO: Add back to shape_lib_gen invocation once custom op support is back.
-#  --pytorch_op_extensions=${ext_module} \
--- a/build_tools/update_torch_ods.sh
+++ b/build_tools/update_torch_ods.sh
@@ -16,20 +16,19 @@ build_dir="$(realpath "${TORCH_MLIR_BUILD_DIR:-$src_dir/build}")"
 torch_ir_include_dir="${src_dir}/include/torch-mlir/Dialect/Torch/IR"
 python_packages_dir="${build_dir}/tools/torch-mlir/python_packages"
 
+TORCH_MLIR_EXT_PYTHONPATH="${TORCH_MLIR_EXT_PYTHONPATH:-""}"
 pypath="${python_packages_dir}/torch_mlir"
-# TODO: Re-enable once custom op support is back.
-#if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
-#  pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
-#fi
-#ext_module="torch_mlir._torch_mlir_custom_op_example"
-#if [ ! -z ${TORCH_MLIR_EXT_MODULES} ]; then
-#  ext_module="${ext_module},${TORCH_MLIR_EXT_MODULES}"
-#fi
+if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
+  pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
+fi
+TORCH_MLIR_EXT_MODULES="${TORCH_MLIR_EXT_MODULES:-""}"
+ext_module="${ext_module:-""}"
+if [ ! -z ${TORCH_MLIR_EXT_MODULES} ]; then
+  ext_module="${TORCH_MLIR_EXT_MODULES}"
+fi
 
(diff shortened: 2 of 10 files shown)


nithinsubbiah:
@rdadolf Python build fails because I haven't included the `__init__.py` that loads the custom_op_example library file (.so) - https://github.com/llvm/torch-mlir/pull/1139/files#diff-f7cdb8efffece9ed974b90846e54110d5d9e1b808a46225538e54e09772877b2L12. When I include it I get the following error: `torch-mlir/python/torch_mlir/_torch_mlir_custom_op_example/__init__.py): Too many levels of symbolic links`. My understanding is that a symbolic link to itself is created somewhere but I can't figure out where or how. If you do see something fishy please let me know, thanks!

python/torch_mlir/_torch_mlir_custom_op_example/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # Setup PyTorch
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../dialects/torch/importer/jit_ir/cmake/modules")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/../cmake/modules")
 include(TorchMLIRPyTorch)
 TorchMLIRProbeForPyTorchInstall()
```
powderluv, line 4: you want to check if it is a LOCAL Pytorch install

nithinsubbiah: I changed this PR to just enable custom op support, will modify this in the following PR which will have this example working

nithinsubbiah:
I've modified this PR to just enable custom op support - ability to register custom ops as torch ops in Torch MLIR. I will enable the example in a follow-up PR (to better resolve the issue)

build_tools/update_shape_lib.sh
```diff
@@ -9,26 +9,27 @@
 # For more information on supporting custom operators, see:
 #   ${TORCH_MLIR}/python/torch_mlir/_torch_mlir_custom_op_example/README.md
 
-set -euo pipefail
+set -eo pipefail
```
powderluv, line 12: why this change ?

build_tools/update_shape_lib.sh
```diff
@@ -9,26 +9,27 @@
...
 
 pypath="${python_packages_dir}/torch_mlir"
-# TODO: Re-enable once custom op support is back.
-#if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
-#  pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
-#fi
-#ext_module="torch_mlir._torch_mlir_custom_op_example"
-#if [ ! -z ${TORCH_MLIR_EXT_MODULES} ]; then
+if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
+ pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
```
powderluv, line 21: two spaces

build_tools/update_torch_ods.sh
```diff
@@ -9,27 +9,28 @@
 # For more information on supporting custom operators, see:
 #   ${TORCH_MLIR}/python/torch_mlir/_torch_mlir_custom_op_example/README.md
 
-set -euo pipefail
+set -eo pipefail
```
powderluv, line 12: why ? can we leave as is ?

build_tools/update_torch_ods.sh
```diff
@@ -9,27 +9,28 @@
...
 
 pypath="${python_packages_dir}/torch_mlir"
-# TODO: Re-enable once custom op support is back.
-#if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
-#  pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
-#fi
-#ext_module="torch_mlir._torch_mlir_custom_op_example"
-#if [ ! -z ${TORCH_MLIR_EXT_MODULES} ]; then
+if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
+ pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
```
powderluv, line 21: two spaces

python/torch_mlir/_torch_mlir_custom_op_example/CMakeLists.txt
```diff
@@ -29,4 +29,3 @@ set_target_properties(torch_mlir_custom_op_example PROPERTIES
 )
 torch_mlir_python_target_compile_options(torch_mlir_custom_op_example)
 mlir_check_all_link_libraries(torch_mlir_custom_op_example)
-
```
powderluv, line 32: white space

build_tools/update_shape_lib.sh
nithinsubbiah: `set -u` throws an unbound variable error if a referenced variable isn't set. When a user doesn't want to add a custom operator `TORCH_MLIR_EXT_PYTHONPATH` and `TORCH_MLIR_EXT_MODULES` variables aren't initialized thus relaxing the need for it to be set.

build_tools/update_torch_ods.sh
nithinsubbiah: same reasoning as above

build_tools/update_shape_lib.sh
powderluv: I think we should set it to default if not set.

python/CMakeLists.txt
```diff
@@ -111,8 +111,8 @@ add_subdirectory(torch_mlir/eager_mode)
 # Required for running the update_torch_ods.sh and update_shape_lib.sh scripts.
 ################################################################################
 
-# TODO: renable once it build on macOS Intel / M1
-#add_subdirectory(torch_mlir/_torch_mlir_custom_op_example)
+#TODO: reenable custom op example
+# add_subdirectory(torch_mlir/_torch_mlir_custom_op_example)
```
powderluv, line 115: i would also remove all these TODOs. I think @silvasean is going to rework custom ops in the future and these may not make sense.

build_tools/update_shape_lib.sh
powderluv: like https://github.com/llvm/torch-mlir/blob/9cf0b6e8ff4ed2400175475c7ca7694688cf17f1/build_tools/python_deploy/build_linux_packages.sh#L43

lib/Conversion/TorchToLinalg/CMakeLists.txt
```diff
@@ -1,6 +1,5 @@
 add_mlir_conversion_library(TorchMLIRTorchToLinalg
-# TODO: Re-enable after MacOS support is fixed for the custom op extension.
-#  CustomOpExample.cpp
+  # CustomOpExample.cpp
```
powderluv, line 2: delete ?

lib/Conversion/TorchToLinalg/PopulatePatterns.h
```diff
@@ -63,7 +63,7 @@ void populateIndirectDataMovementPatternsAndLegality(
 void populateTensorConstructorsPatternsAndLegality(TypeConverter &typeConverter,
                                                    RewritePatternSet &patterns,
                                                    ConversionTarget &target);
-//void populateCustomOpExamplePatternsAndLegality(TypeConverter &typeConverter,
+// void populateCustomOpExamplePatternsAndLegality(TypeConverter &typeConverter,
```
powderluv, line 66: whitespace

lib/Conversion/TorchToLinalg/TorchToLinalg.cpp
```diff
@@ -62,7 +62,7 @@ class ConvertTorchToLinalg
 
     RewritePatternSet patterns(context);
 
-    //torch_to_linalg::populateCustomOpExamplePatternsAndLegality(
+    // torch_to_linalg::populateCustomOpExamplePatternsAndLegality(
```
powderluv, line 65: whitespace

powderluv (commented):
some more whitespace changes

powderluv (approved):

Issue #1074, opened 2022-07-18
Invalid tensor type generated for prim.If during InlineGlobalSlots


kkiningh:
The following model fails during the InlineGlobalSlots pass:

```python
from torch import nn

class Model(nn.Module):
    def __init__(self, n):
        super().__init__()
        self.register_buffer('var', torch.ones(n))

    def forward(self, x):
        if self.training:
            var = x.var(dim=(0))
        else:
            var = self.var
        return var

model = Model(1)
model.eval()
compiled = torch_mlir.compile(model, torch.randn(1, 1), output_type=torch_mlir.OutputType.LINALG_ON_TENSORS)
```

Specifically, the following error message is generated:

```bash
$ ./build/bin/torch-mlir-opt -pass-pipeline='torchscript-module-to-torch-backend-pipeline' -mlir-print-ir-after-failure -mlir-disable-threading /var/folders/98/z_z16my5143_3fbfwmy1kgj40000gn/T/Model.mlir                  
/var/folders/98/z_z16my5143_3fbfwmy1kgj40000gn/T/ipykernel_69244/2867114530.py:11:8: error: 'torch.prim.If' op  along control flow edge from Region #1 to parent results: source type #0 '!torch.tensor<[1],f32>' should match input type #0 '!torch.tensor'
/var/folders/98/z_z16my5143_3fbfwmy1kgj40000gn/T/ipykernel_69244/2867114530.py:11:8: note: see current operation: %6 = "torch.prim.If"(%5) ({
  %7 = "torch.prim.ListConstruct"(%0) : (!torch.int) -> !torch.list<int>
  %8 = "torch.aten.var.dim"(%4, %7, %1, %2) : (!torch.tensor, !torch.list<int>, !torch.bool, !torch.bool) -> !torch.tensor
  "torch.prim.If.yield"(%8) : (!torch.tensor) -> ()
}, {
  %7 = "torch.tensor.literal"() {value = dense<1.000000e+00> : tensor<1xf32>} : () -> !torch.tensor<[1],f32>
  "torch.prim.If.yield"(%7) : (!torch.tensor<[1],f32>) -> ()
}) : (!torch.bool) -> !torch.tensor
```

Looking at the MLIR dump, it seems that the pass is inlining a literal with the type `torch.tensor<1xf32>` when the result of the prim.if op is expecting a bare `torch.tensor` (which I assume is a tensor with no shape information?)

Here's the original MLIR as well in case it helps diagnose the issue:

```
[...]

PR #1140, opened 2022-08-03
Rework how global slot initializers work.

silvasean:
Rather than a per-global-slot initializer region, we now have one for
the whole module. For example, it might look like this:

```
torch.global_slot "private" @tensor : !torch.tensor
torch.global_slot "private" @list : !torch.list<tensor>
torch.global_slot.module_initializer {
  %0 = torch.tensor.literal(dense<0.0> : tensor<f32>) : !torch.tensor
  %1 = torch.prim.ListConstruct %0 : (!torch.tensor) -> !torch.list<tensor>
  torch.initialize.global_slots [
    @tensor(%0 : !torch.tensor)
    @list(%1 : !torch.list<tensor>)
  ]
}
```

This new structure allows GlobalizeObjectGraph to create the initializer in a
much simpler way, avoiding the need to reason about whether different slots
alias each other. Reasoning about whether slots alias each other now is the
responsibility of InlineGlobalSlots, which has to do a much more complicated
analysis, implemented using MLIR's dataflow analysis framework.

Recommended review order:
- Check out the new IR constructs in the .mlir files of various passes
- Op definitions (*.td)
- Changes to GlobalizeObjectGraph pass.
- InlineGlobalSlots pass (~total rewrite)
- Misc changes:
  - Moving torchMlirAdjustStaticInformation for sharing with C++ code.
  - EraseModuleInitializer pass

To make this a bit nicer, it would be good to have a `torch.module` op
with an initializer region attached. That would be more invasive though.

This change has highlighted certain aspects of our project layering
which are worth calling out. None of our backends can handle global
slots, so we enforce that there are no global slots before backend
lowering. At an earlier stage in the project, we had aspirations of
transparently handling mutable global state and such, but for reasons
described below, that is no longer a goal. So really global slots should
be seen as a progressive lowering step as part of inlining all the
IValue's in the original program (GlobalizeObjectGraph is also one such
step).

Over time, with insights from work like IREE-JAX, it has become clear
that there isn't a reliable programming model we can compile for users
where we just transparently handle mutable global state (and some other
things, like lists and dictionaries). There is a need for an "outer
program" that orchestrates more restricted subroutines of the kind we
can handle in our compile flow here. The benefit of that is that it
decouples considerations like shapes, dtypes, etc. from the program
constructs used in the outer program. As long as the outer program can
efficiently invoke (pipelining/async/etc.) high-performance
data-parallel numerical subroutines of the kind we compile in our flow
here, then there is a complete programming model. This is also
consistent with the direction of upstream PyTorch which is becoming more
tracing-based (which inherently loses a lot of program structure, which
then has to be applied back with an "outer program" orchestrating the
traced subroutines).


silvasean:
This fixes https://github.com/llvm/torch-mlir/issues/1074

lib/Dialect/Torch/Transforms/InlineGlobalSlots.cpp
```diff
@@ -6,83 +6,429 @@
...
+  }
+
+  /// Helper for updatating the state with the correct ChangeResult based on the
+  /// safety of a use.
+  ChangeResult
+  incorporateSafetyOfUse(const InlineGlobalSlotsAnalysisState *useState) {
+    // The use is safe, so no need to change anything.
+    if (useState->isSafe)
+      return ChangeResult::NoChange;
+    return setSafe(false);
```
ramiro050, line 135: Is there a reason not to do `return setSafe(useState->isSafe)` here?

ramiro050 (approved):
LGTM! I just have some small questions/comments

lib/Dialect/Torch/Transforms/InlineGlobalSlots.cpp
silvasean: Directly setting the state could violate monotonicity. For example, if don't do the extra `if` to guard the setSafe call, then we could end up doing `setSafe(useState->isSafe/*==true*/)` on something already proved to be safe -- this wil trigger the assertion in `setSafe`.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1936,3 +1979,148 @@ LogicalResult ShapeCalculateYieldShapesOp::verify() {
...
+    if (op.getOperation() != getOperation())
+      return op.emitError("there must be only one global slot initializer");
+  }
+
+  // Collect the relevant symbol names we will verify.
+  DenseSet</*StringAttr*/ Attribute> knownGlobalSlots;
+  for (auto op : module.getOps<GlobalSlotOp>())
+    knownGlobalSlots.insert(op.sym_nameAttr());
+  DenseSet</*StringAttr*/ Attribute> initializedGlobalSlots;
+  auto initialize = cast<InitializeGlobalSlotsOp>(getBody()->getTerminator());
```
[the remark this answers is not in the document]
silvasean: Good catch!! Fixed.

PR #1142, opened 2022-08-03
Fix a small error in shape function for embeddingBag.PaddingIdx op

vid-999:


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -964,7 +964,7 @@ def aten〇embedding_bag〇padding_idx(weight: List[int], indices: List[int], of
     output_bag_shape.append(out_dim1)
 
     offset2bag_shape: List[int] = []
-    if mode == 1:
+    if mode == 0:
```
ramiro050, line 967: nit: can you create variables with the names of the modes, to make this easier to read? i.e.

```python
sum_mode = 0
...

if mode == sum_mode:
  ...
```

ramiro050, line 967: Can you add an e2e test that results in a failure without this change and works with the change?

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
vid-999: umm , for the the mean mode would have to work. Since the mean mode == 1. 
The Sum mode == 0. 
I could implement the mean mode too ?

vid-999: yes

ramiro050: I don't think the mean mode is needed. The sum mode will now go into an `if` that it was not going into before. I would expect this to be a behavior that can be tested using just the sum mode

PR #1144, opened 2022-08-03
[MLIR][TORCH] Fix CI failure

vivekkhandelwal1:
This commit fixes the CI failure by temporarily adding the failing
test to xfail set.


powderluv (approved):
Let's open a bug to track the root cause

PR #1150, opened 2022-08-03
torch_mlir.compile: Allow ignoring traced shapes

### LINKED ISSUE #997: `torch_mlir.compile` error with jit.trace when using `TensorPlaceholder` for example_args (related to dynamic shape support)
I am facing two issues that are likely inter-dependent.

### First Issue: 
This [line of code](https://sourcegraph.com/github.com/llvm/torch-mlir@5888c4f7dc5aad504ef8c77442ebbf3c0393623c/-/blob/python/torch_mlir/__init__.py?L120
) results in an exception when `example_args` is provided as `TensorPlaceholder`.

```
RuntimeError: Type 'Tuple[__torch__.TensorPlaceholder]' cannot be traced. Only Tensors and (possibly nested) Lists, Dicts, and Tuples of Tensors can be trace
```

Is `TensorPlaceholder` not intended to be used for `torch.jit.trace`, and only for the later steps during torch-mlir import (collecting `forward_annotation` for `class_annotator` etc)? If so, how can I use dynamic shapes when tracing, since the shapes are statically annotated during `jit.trace` and those get carried over to the torch dialect upon import? (this leads to the second issue below)
 
### Second Issue:
How can I get the `dynamic_axes` support when `use_tracing=True`?

For this simple model: 
```python3
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()

    def forward(self, x, y):
        x = F.relu(x)
        z1 = x + y
        z2 = torch.sum(z1, 0)
        return z1, z2
```
I first traced (`jit.trace`) with usual `torch.Tensor`s (`example_args = (torch.randn(8, 8), torch.randn(8))`), and modified [this line](https://sourcegraph.com/github.com/llvm/torch-mlir@5888c4f7dc5aad504ef8c77442ebbf3c0393623c/-/blob/python/torch_mlir/__init__.py?L131) to provide `dynamic_axes=[0]` to the `TensorPlaceholder.like` method. This seems to create the right input tensors (with dynamic dims) but then casts back to static shapes:
```
  func.func @forward(%arg0: !torch.vtensor<[?,8],f32>, %arg1: !torch.vtensor<[?],f32>) -> (!torch.vtensor<[8,8],f32>, !torch.vtensor<[8],f32>) {
    %0 = torch.tensor_static_info_cast %arg0 : !torch.vtensor<[?,8],f32> to !torch.vtensor<[8,8],f32>
    %1 = torch.tensor_static_info_cast %arg1 : !torch.vtensor<[?],f32> to !torch.vtensor<[8],f32>
    %2 = torch.aten.relu %0 : !torch.vtensor<[8,8],f32> -> !torch.vtensor<[8,8],f32>
```

@asaadaldien believes these static casts should be dropped from jit_import when using tracing, and only do the casting for constants (e.g. weights which are scripting / tracing invariant).

For reference here are the [full dumps](https://gist.github.com/sjain-stanford/c97dcc2e41e350b04633feb8f96a40a4) for Torchscript (`ts.txt`), raw imported TorchDialect module (`raw.mlir`) and after TS-to-Torch pipeline (`torch.mlir`).

Discord thread: https://discord.com/channels/636084430946959380/742573221882364009/991793938639892490

silvasean:
In some cases, users know that a traced graph is valid for a wider set
of shapes than they originally traced it with. Provide an option for
users to ignore the shapes in the traced graph when they know it is
legal.

Fixes #997

### DIFF
--- a/python/test/compile_api/tracing.py
+++ b/python/test/compile_api/tracing.py
@@ -31,3 +31,24 @@ print(torch_mlir.compile(TanhModule(), (tanh_example_input,), use_tracing=True))
 print(torch_mlir.compile(TanhModule(), [tanh_example_input], use_tracing=True))
 # CHECK-LABEL: @forward
 # CHECK: torch.aten.tanh %{{.*}} : !torch.vtensor<[2,3],f32> -> !torch.vtensor<[2,3],f32>
+
+# TensorPlaceholder support.
+placeholder = torch_mlir.TensorPlaceholder.like(
+    tanh_example_input, dynamic_axes=[1])
+print(torch_mlir.compile(TanhModule(), [placeholder],
+                         use_tracing=True, ignore_traced_shapes=True))
+# CHECK-LABEL: @forward
+# CHECK: torch.aten.tanh %{{.*}} : !torch.vtensor<[2,?],f32> -> !torch.vtensor<[2,?],f32>
+
+try:
+    # CHECK: `ignore_traced_shapes` requires `use_tracing`
+    torch_mlir.compile(TanhModule(), [placeholder], ignore_traced_shapes=True)
+except Exception as e:
+    print(e)
+
+
+try:
+    # CHECK: TensorPlaceholder can only be used with tracing when `ignore_traced_shapes=True`
+    torch_mlir.compile(TanhModule(), [placeholder], use_tracing=True)
+except Exception as e:
+    print(e)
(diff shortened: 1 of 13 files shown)


silvasean:
@sjain-stanford -- PTAL. Also please accept repo invite so I can add you as a "reviewer".

sjain-stanford:
Thanks @silvasean for the fix. LGTM (and I've accepted the repo invite as well).

Issue #1149, opened 2022-08-03
Document CMake hack better


silvasean:
https://github.com/llvm/torch-mlir/blob/48ec3005865cb55b6e70a85ccf67f5e63d3c5b83/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/CMakeLists.txt#L13

Can we get an expanded explanation and use our common "TODO" tag here to track the remaining work. Also, it would be nice if we filed a new issue to track follow-on work post-LTC-landing.

henrytwo:
I'll update the comment to link to the new issue I made, but here's a direct link: https://github.com/llvm/torch-mlir/issues/1154

PR #1155, opened 2022-08-04
Update LTC CMake hack documentation

henrytwo:
Addresses #1149 and clarifies details about LTC CMake hack

cc: @antoniojkim @ke1337


setup.py
```diff
@@ -76,6 +76,7 @@ def run(self):
                 f"-DMLIR_ENABLE_BINDINGS_PYTHON=ON",
                 f"-DLLVM_ENABLE_PROJECTS=mlir",
                 f"-DTORCH_MLIR_ENABLE_MHLO=ON",
+                # TODO(henrytu): Reenable LTC once JIT importer linkage issue is fixed (https://github.com/llvm/torch-mlir/issues/1154)
```
silvasean, line 79: In LLVM/MLIR we just do `TODO` rather than `TODO(username)`.

silvasean (approved):

Issue #647, opened 2022-03-08
Could NOT find Python3


noahchalifour:
I am trying to build the python packages using the commands provided in the readme. I get an error when running: 

```
cmake -GNinja -Bbuild \
  -DCMAKE_C_COMPILER=clang \
  -DCMAKE_CXX_COMPILER=clang++ \
  -DPython3_FIND_VIRTUALENV=ONLY \
  -DLLVM_ENABLE_PROJECTS=mlir \
  -DLLVM_EXTERNAL_PROJECTS="torch-mlir;torch-mlir-dialects" \
  -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR=`pwd` \
  -DLLVM_EXTERNAL_TORCH_MLIR_DIALECTS_SOURCE_DIR=`pwd`/external/llvm-external-projects/torch-mlir-dialects \
  -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
  -DLLVM_TARGETS_TO_BUILD=host \
  external/llvm-project/llvm
```

I have also tried running the above command and adding `-DPython3_EXECUTABLE="$(which python)"`, however I get the same error.

I don't know if the fact that I'm on an M1 MacBook is an issue, any help would be appreciated.

Here is the error:

```
-- Performing Test C_SUPPORTS_MISLEADING_INDENTATION_FLAG - Success
-- Performing Test CXX_SUPPORTS_MISLEADING_INDENTATION_FLAG
-- Performing Test CXX_SUPPORTS_MISLEADING_INDENTATION_FLAG - Success
-- Performing Test LINKER_SUPPORTS_COLOR_DIAGNOSTICS
-- Performing Test LINKER_SUPPORTS_COLOR_DIAGNOSTICS - Failed
-- Looking for os_signpost_interval_begin
-- Looking for os_signpost_interval_begin - found
-- Performing Test macos_signposts_usable
-- Performing Test macos_signposts_usable - Success
-- Found Python3: /Users/noahchalifour/Code/noahchalifour/AI_TEXT_EDITOR_4250/ai-src/torch-mlir/mlir_venv/bin/python (found suitable version "3.8.9", minimum required is "3.6") found components: Interpreter 
-- Linker detection: ld64
-- Performing Test HAS_WERROR_GLOBAL_CTORS
-- Performing Test HAS_WERROR_GLOBAL_CTORS - Success
-- Performing Test LLVM_HAS_NOGLOBAL_CTOR_MUTEX
[...]

PR #1156, opened 2022-08-04
[Docker] Minimal setup to run CMake builds + unit tests

sjain-stanford:
Adds a very minimal (optional) docker setup to quickly get CMake builds + LIT tests running consistently in a reproducible way.

**Context:** While we (Cruise) use the `Bazel` build of torch-mlir internally, we also need the `CMake` build setup on our end to validate our PRs won't break CI (before sending them upstream). Some of us had initial hurdles getting the CMake builds working locally (python venv issues similar to https://github.com/llvm/torch-mlir/issues/647 requiring `python-dev` to be installed). This PR addresses such issues by reducing the (somewhat manual and env dependent) build+test steps to the following:

```shell
# Build an image and launch a docker container
./run_docker.sh

# Run CMake build
./run_cmake_build.sh

# Run CMake unit tests (+ python + dialect unit tests)
./run_cmake_test.sh
```

It is not meant to be used for testing builds on multiple platforms / architectures (aka matrix testing) or for out-of-tree builds.

Please LMK if you'd also like a GitHub actions workflow (non-gating) to test this? Also, feel free to reject this if this is too far from what we should allow in torch-mlir.

cc: @silvasean

(thanks @asaadaldien for the initial `Dockerfile`)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+ARG BASE_IMG=nvidia/cuda:10.2-cudnn8-devel-ubuntu18.04
+FROM ${BASE_IMG} as dev-base
+
+ARG ARCH="x86_64"
+ARG TARGETARCH="amd64"
+ARG BAZEL_VERSION=4.2.1
+
+# Install basic packages
+RUN apt-get update                                              && \
+    apt-get install -y                                             \
+    python3.8                                                      \
+    python3.8-dev                                                  \
+    cmake                                                          \
+    ninja-build                                                    \
+    git                                                            \
+    python3-pip                                                    \
+    wget                                                           \
+    clang                                                          \
+    automake                                                       \
+    libtool                                                        \
+    curl                                                           \
+    make                                                           \
+    unzip
+
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 10
+RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10
+
+# Install bazel
+RUN wget -q https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-${ARCH} -O /usr/bin/bazel \
+    && chmod a+x /usr/bin/bazel
+
+COPY requirements.txt /opt/app/requirements.txt
+WORKDIR /opt/app
+RUN python -m pip install --upgrade pip
+RUN python -m pip install -r requirements.txt
+
+WORKDIR /opt/src/torch-mlir/torch-mlir
--- a/run_cmake_build.sh
+++ b/run_cmake_build.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+# Configure cmake to build torch-mlir in-tree
+cmake -GNinja -Bbuild \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_C_COMPILER=clang \
+  -DCMAKE_CXX_COMPILER=clang++ \
+  -DPython3_FIND_VIRTUALENV=ONLY \
+  -DLLVM_ENABLE_PROJECTS=mlir \
+  -DLLVM_EXTERNAL_PROJECTS="torch-mlir;torch-mlir-dialects" \
+  -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR=`pwd` \
+  -DLLVM_EXTERNAL_TORCH_MLIR_DIALECTS_SOURCE_DIR=`pwd`/externals/llvm-external-projects/torch-mlir-dialects \
+  -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
+  -DLLVM_TARGETS_TO_BUILD=host \
+  externals/llvm-project/llvm
+
+# Build just torch-mlir (not all of LLVM)
+cmake --build build --target tools/torch-mlir/all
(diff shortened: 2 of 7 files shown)


powderluv:
I think it is a good idea to have our own Docker / Dockerfile. We already use Docker for our release builds so we should try to keep them the same across CI  and Release build docker. 

Is there a reason you require CUDA (10.2) Docker ? I would suggest we start with the the manylinux 2014 (maybe even 2_18 had some experiments here: https://github.com/llvm/torch-mlir/compare/main...release-src-build ). 
If we don't require anything special in the Dockerimage everything can be expressed in setup.py / pyproject.toml we could get away without building our own docker image and using the stock manylinux one.

Dockerfile
```diff
@@ -0,0 +1,37 @@
+ARG BASE_IMG=nvidia/cuda:10.2-cudnn8-devel-ubuntu18.04
```
powderluv, line 1: I would make this manylinux

Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+    curl                                                           \
+    make                                                           \
+    unzip
+
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 10
+RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10
+
+# Install bazel
+RUN wget -q https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-${ARCH} -O /usr/bin/bazel \
+    && chmod a+x /usr/bin/bazel
```
powderluv, line 30: The current Release Docker uses bazel-5.1.0 do you need a specific version ?

run_docker.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+docker build -t torch-mlir-cmake:dev .
```
powderluv, line 3: do you want to rebuild docker everytime you run it ?

run_cmake_build.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+# Configure cmake to build torch-mlir in-tree
+cmake -GNinja -Bbuild \
```
powderluv, line 4: This are normal cmake commands so we probably want to avoid this since everyone has a variation of these commands to run.

run_cmake_test.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
powderluv, line 1: same as cmake_build

powderluv (commented):
I would also move everything under build_tools/docker/

silvasean:
We definitely need to dockerize the CI, and hopefully we can write such scripts around that. I'm wary of adding a dockerfile that isn't load bearing for CI (and also I'm not a big fan of adding yet another CI just to test a dockerfile which is not used by CI).

sjain-stanford:
Thanks @powderluv and @silvasean for the feedback. Initially we set out to have this used as an auxiliary (lightweight) mechanism to validate local changes more reproducibly, but I see your valid point in standardizing CI to use the same containers instead of GHA VMs. 

The plan sounds great to me. It does increase the scope of the initial PR (e.g. support for multiple architectures, OS, python versions, in-tree vs out-of-tree, etc) so please allow me sometime to work on dockerizing the different CI workflows we run today. Also, please LMK if anyone would like to help with the effort. I'll leave this PR open for technical discussions / help with issues I encounter.

Dockerfile
sjain-stanford: I was hoping using an nvidia base image would allow testing the GPU builds of PyTorch as well, but we can have multiple Dockerfiles too. For `manylinux` I see [this one](https://hub.docker.com/r/stellaraccident/manylinux2014_x86_64-bazel-5.1.0) from @stellaraccident that comes with bazel pre-installed currently used for release builds - is this the right one to use for CI as well, or one of [these CentOS7 ones](https://github.com/pypa/manylinux#manylinux2014-centos-7-based) from `quay.io`?

Dockerfile
sjain-stanford: Not really, I believe any (relatively new) release should work.

run_docker.sh
sjain-stanford: Yes. Here's why I think this is inexpensive and useful: For the local workflow I expect users would `edit->run_docker->run_test` in a tight loop and because `docker build` uses layered caching (quite aggressively), unless the last image was invalidated (by a change to Dockerfile itself or `docker rmi`) it'd just use the same to spawn a container. For any source code changes, these would be transparent to the docker build step because we're loading the source tree as a virtual mount and not something baked into the image itself. Does that make sense?

Dockerfile
powderluv: If we use the release script we can override to whatever image we want to use.

Dockerfile
powderluv: Hopefully we don't need a Dockerfile then.

run_docker.sh
powderluv: The release script should provide us the trampoline into docker seamlessly

sjain-stanford: I agree, this step becomes moot if we're using the base docker image as-is (without a Dockerfile).

Dockerfile
sjain-stanford: That should work. Thanks!

sjain-stanford:
Let's use https://github.com/llvm/torch-mlir/issues/1162 for further discussion on the "Docker for CI" redesign. Closing this.

PR #1157, opened 2022-08-04
Propagate device data names

antoniojkim:
Need to propagate device data names in order to be able to distinguish between inputs and weights.

An example of its intended usage has been added to the reference backend. The idea is to assign names to tensors on the python side (e.g. perhaps in the dataloader) so that you know which tensors are which in the actual compile

Also, export remaining headers that weren't exported before.

CC: @ke1337 @henrytwo

### DIFF
--- a/python/torch_mlir/csrc/reference_lazy_backend/backend_impl.cpp
+++ b/python/torch_mlir/csrc/reference_lazy_backend/backend_impl.cpp
@@ -113,10 +113,12 @@ public:
       // count number of inputs
       auto name = mlir_data->mlir_info()->name;
       if (startswith(name, "input_")) {
+        // Printing tensor name for testing purposes
         std::cout << "Input tensor: " << name << std::endl;
         ++num_inputs;
       }
     }
+    // Printing number of input tensors for testing purposes
     std::cout << num_inputs << " input tensors found" << std::endl;
     graph_executor.run(stack);
     std::vector<torch::lazy::BackendDataPtr> results;
--- a/python/torch_mlir/csrc/reference_lazy_backend/backend_impl.cpp
+++ b/python/torch_mlir/csrc/reference_lazy_backend/backend_impl.cpp
@@ -18,6 +18,7 @@
 #include <torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.h>
 #include <torch_mlir/csrc/base_lazy_backend/utils/debug.h>
 #include <torch_mlir/csrc/base_lazy_backend/utils/exception.h>
+#include <torch_mlir/csrc/base_lazy_backend/utils/string_utils.h>
 
 #include "backend_impl.h"
 
@@ -88,6 +89,8 @@ public:
     auto mlir_computation =
         static_cast<TorchMlirComputation*>(computation.get());
 
+    int num_inputs = 0;
+
     // Vendor backend specific execution can be inserted here.
     //
     // We don't have a way to execute a computation based on the generated MLIR,
@@ -106,7 +109,14 @@ public:
         at::Tensor tensor = mlir_data->mlir_info()->tensor;
         stack.emplace_back(tensor);
       }
+
+      // count number of inputs
+      auto name = mlir_data->mlir_info()->name;
+      if (startswith(name, "input_")) {
+        ++num_inputs;
+      }
     }
+    std::cout << num_inputs << " input tensors found" << std::endl;
     graph_executor.run(stack);
     std::vector<torch::lazy::BackendDataPtr> results;
     for (torch::jit::IValue component : stack) {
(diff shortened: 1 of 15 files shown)


setup.py
```diff
@@ -85,6 +83,10 @@ def run(self):
                 f"-DCMAKE_C_VISIBILITY_PRESET=hidden",
                 f"-DCMAKE_CXX_VISIBILITY_PRESET=hidden",
             ]
+            enable_ltc = bool(int(os.environ.get("TORCH_MLIR_ENABLE_LTC", 0)))
```
silvasean, line 86: can you document this at the top of the file?

silvasean (changes requested):
can you add a test for this?

silvasean:
I'm not sure I understand what the test is testing. It seems to call the API but I don't see it checking the result.

antoniojkim:
> I'm not sure I understand what the test is testing. It seems to call the API but I don't see it checking the result.

There's unfortunately no good way to check the result. I originally checked that the number if inputs was greater than zero, but this is not true for all ops apparently and thus some ops fail that check.

I also tried saving the input names and checking them against the values that are set, but this also doesn't work as multiple tests are run in parallel and thus there is a race condition in the storage and retrieval of this value.

As such, I decided to only add it as a spot check. Especially considering that this is not something that is needed for all lazy backends (for example, the TorchScript backend has no use for this). I figured its fine to just print the number of inputs in the reference lazy backend so that we can do a visible check of the output to verify that its working as intended.

@silvasean please let me know if my reasoning for this is sufficient

antoniojkim:
@silvasean Gentle reminder to please review again

silvasean:
Sorry, I've been OoO.

What I meant to say is that I see `std::cout << num_inputs << " input tensors found" << std::endl;` but I don't see where you check that that gets actually printed. And I don't think that calling this API in the end-to-end tests is very useful.

I would recommend that you add unit tests like we did with the "eager mode" here: https://github.com/llvm/torch-mlir/tree/main/python/test/eager_mode

I'm imagining a 3 line test where you call `lazy_backend.set_parameter_name(e, f"some_name")` and then either print the graph or print the node or get the parameter name or something and print it out and check that "some_name" is in the output.

antoniojkim:
> What I meant to say is that I see `std::cout << num_inputs << " input tensors found" << std::endl;` but I don't see where you check that that gets actually printed. And I don't think that calling this API in the end-to-end tests is very useful.

There is no concrete check. I only meant for it to be a "check by inspection". Its not required by all backends and I thought it was sufficient to be able to see it in the test logs.

> I would recommend that you add unit tests like we did with the "eager mode" here: https://github.com/llvm/torch-mlir/tree/main/python/test/eager_mode

The problem with doing this is that the tests are run in parallel in the same process. This causes race conditions as there is only a single instance of the static lazy tensor C++ symbols.
 
> I'm imagining a 3 line test where you call `lazy_backend.set_parameter_name(e, f"some_name")` and then either print the graph or print the node or get the parameter name or something and print it out and check that "some_name" is in the output.

This is what I was hoping to add as well, but with the current test infrastructure, unless I'm mistaken, I don't think this is possible due to the reasons I listed above.

antoniojkim:
@silvasean I've added some unit tests. Please review again when you get a chance

python/torch_mlir/csrc/reference_lazy_backend/backend_impl.cpp
```diff
@@ -106,7 +109,15 @@ class ReferenceLazyBackendImpl : public torch::lazy::TorchMlirBackendImpl {
         at::Tensor tensor = mlir_data->mlir_info()->tensor;
         stack.emplace_back(tensor);
       }
+
+      // count number of inputs
+      auto name = mlir_data->mlir_info()->name;
+      if (startswith(name, "input_")) {
```
silvasean, line 115: does this name end up in the graph or any other data structure we can see without doing this? It feels really hacky to print to stdout like this, even for testing.

antoniojkim: No it does not. Plus, this is the reference backend and thus not intended to be used anywhere outside of testing and only as a reference for how a real backend subclass should be implemented.

silvasean: okay. can you make this name more verbose like "only_for_testing_print_this_"

silvasean (approved):

python/torch_mlir/csrc/reference_lazy_backend/backend_impl.cpp
antoniojkim: Is that strictly necessary? I think it's fairly obvious that the printing out is not a necessary step in implementing a lazy backend.

silvasean: I believe so. Otherwise it can be confusing -- is it leftover debugging code? something else?

antoniojkim: okay, I'll add a comment explaining the purpose

Issue #1131, opened 2022-08-01
[RFC] Adding support for non-constant dims for aten.view


gpetters94:
I'm working on lowering OPT, and I'm running into the following:

```
error: failed to legalize operation 'torch.aten.view' that was explicitly marked illegal
note: see current operation: %932 = "torch.aten.view"(%898, %931) : (!torch.vtensor<[1,12,7,64],f32>, !torch.list<int>) -> !torch.vtensor<[12,7,64],f32>
```

Inspecting the lowering of aten.view, it looks like the output shape is `-1, -1, 64` because the first two input dims aren't constants. The solution I'd like to write is to recursively follow the dims up the tree, verifying that all the ops are either constants, no-ops (i.e. NumToTensor), or math ops (i.e. multiplication, addition) and then performing the math statically to determine the output shape. Does this sound like how we want to implement this?

silvasean:
The output shape looks like [12, 7, 64] in your snippet and not [-1, -1, 64]. Can you show the actual IR snippet you are dealing with?

gpetters94:
In the actually processing of `aten.view`, it checks if each input dim is a constant. If not it assigns `kUnknownDim` to it, and in this case the first two inputs are not constants. The code is [here](https://github.com/llvm/torch-mlir/blob/fe3c9f5208ee1f77bdd40807c77ae64ed344dcc9/lib/Conversion/TorchToLinalg/DataMovement.cpp#L322).

silvasean:
Can you show the IR before the pass?

silvasean:
(for future reference, it's usually important to show a reduced, fully valid IR example with any bug reports like this)

gpetters94:
Here's the IR after failure: https://gist.github.com/gpetters94/af96b032acb0e6c6274af9aff62ec5e3

The relevant part is:
```
  %136 = torch.aten.mul.Tensor %123, %71 : !torch.vtensor<[],si64>, !torch.vtensor<[],si64> -> !torch.vtensor<[],si64>
  %137 = torch.aten.Int.Tensor %136 : !torch.vtensor<[],si64> -> !torch.int
  %138 = torch.aten.Int.Tensor %136 : !torch.vtensor<[],si64> -> !torch.int
  %139 = torch.aten.Int.Tensor %136 : !torch.vtensor<[],si64> -> !torch.int
[...]

PR #1165, opened 2022-08-05
Iteratively run the main simplification pipeline.

silvasean:
(ignore the "Rework how global slot initializers work." commit -- this one doesn't trivially commute with it, but is still logically separate)

This introduces a new pass LowerToBackendContract (better name very
welcome) which performs the bulk of the simplifications that we do,
such as
- shape refinement
- dtype refinement
- maximizing value semantics
- inlining global slots
- decomposing complex ops

The key difference from before is that it iterates the set of
transformations, which can help to break a number of "catch-22" issues
where one simplification depends on another, the latest example being
here:
https://github.com/llvm/torch-mlir/issues/1131


silvasean:
@ramiro050 PTAL now that I fixed the CI issue.

silvasean:
+ @ashay for a second pair of eyes on this architectural change.

lib/Dialect/Torch/Transforms/Passes.cpp
```diff
@@ -66,131 +70,82 @@ void mlir::torch::Torch::createTorchScriptModuleToTorchBackendPipeline(
...
   // Convert the bulk of non-ABI-visible !torch.tensor's to !torch.vtensor's.
   pm.addNestedPass<func::FuncOp>(Torch::createMaximizeValueSemanticsPass());
-
-  // Update the return op to return value tensors and remove dead ops.
+  // Update the return op to return value tensors.
   pm.addPass(Torch::createRefinePublicReturnPass());
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
-
-  // Ensure that all tensors have been converted to value semantics.
-  pm.addPass(Torch::createVerifyConversionToValueSemanticsPass());
```
ramiro050, line 153: Is this pass no longer needed?

ramiro050 (approved):
This is great! I just have one question

lib/Dialect/Torch/Transforms/Passes.cpp
silvasean: It is all rolled into the termination condition of the LowerToBackendContract pass.

ramiro050: Should the code for that pass be removed?

ramiro050: Namely, https://github.com/silvasean/torch-mlir/blob/iterate/lib/Dialect/Torch/Transforms/VerifyConversionToValueSemantics.cpp

silvasean: Oh, good call!

PR #1166, opened 2022-08-05
development.md: How to enable ASan

silvasean:


henrytwo (approved):
LGTM. Why would these instructions not work with LTC? Is it because of the PyTorch dependencies?

silvasean:
> LGTM. Why would these instructions not work with LTC? Is it because of the PyTorch dependencies?

Mostly that I haven't thought it through, and generally anything that requires build/shared library considerations requires some special ASan considerations.

PR #1169, opened 2022-08-08
importer: add initial support for loading Float16 tensors

fortianyou:
follow up #761:

This patch updates the `torch_mlir::convertTensorToMlirElementsAttr()`
method to enable the creation of tensors whose base type is Float16.
This patch also adds a test to validate the IR generation, and it
 updates the test for importing tensors of various types.


fortianyou:
@ashay Thanks! :)

PR #1176, opened 2022-08-08
Add decomposition of `aten.masked.tensor` op.

pashu123:
`aten.masked.tensor` op has been decomposed to `aten.masked.scalar` op.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -884,9 +884,9 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
                                           threshold);
     return b.create<arith::SelectOp>(loc, predicate, constantZero, grad);
   }
-  if (auto maskedFill = dyn_cast<AtenMaskedFillScalarOp>(op)) {
+  if (auto maskedFillScalar = dyn_cast<AtenMaskedFillScalarOp>(op)) {
     AtenMaskedFillScalarOp::Adaptor adaptor(operands);
-    Type dtype = converter->convertType(maskedFill.getType())
+    Type dtype = converter->convertType(maskedFillScalar.getType())
                      .cast<RankedTensorType>()
                      .getElementType();
 
@@ -894,6 +894,17 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     Value mask = payloadArgs[1];
     Value fillValue = convertScalarToDtype(b, loc, adaptor.value(), dtype);
 
+    return b.create<arith::SelectOp>(loc, mask, fillValue, input);
+  }
+  if (auto maskedFillTensor = dyn_cast<AtenMaskedFillTensorOp>(op)) {
+    AtenMaskedFillScalarOp::Adaptor adaptor(operands);
+    Type dtype = converter->convertType(maskedFillTensor.getType())
+                     .cast<RankedTensorType>()
+                     .getElementType();
+
+    Value input = payloadArgs[0];
+    Value mask = payloadArgs[1];
+    Value fillValue = convertScalarToDtype(b, loc, payloadArgs[2], dtype);
     return b.create<arith::SelectOp>(loc, mask, fillValue, input);
   }
 
@@ -970,7 +981,7 @@ public:
              AtenEqTensorOp, AtenLtTensorOp, AtenSubScalarOp, AtenAddScalarOp,
              AtenThresholdOp, AtenThresholdBackwardOp, AtenCloneOp, AtenSinOp,
              AtenCosOp, AtenNeScalarOp, AtenNegOp, AtenMaskedFillScalarOp,
-             AtenLogicalOrOp, AtenTriuOp>(op))
+             AtenMaskedFillTensorOp, AtenLogicalOrOp, AtenTriuOp>(op))
       return rewriter.notifyMatchFailure(op, "not a supported elementwise op");
 
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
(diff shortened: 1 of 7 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2445,6 +2445,22 @@ class DecomposeAten_EmbeddingBagOp
...
 
+namespace {
+class DecomposeAtenMaskedFillTensorOp
+    : public OpRewritePattern<AtenMaskedFillTensorOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenMaskedFillTensorOp op,
+                                PatternRewriter &rewriter) const override {
+    Value ScalarFloat =
+        rewriter.create<AtenFloatTensorOp>(op.getLoc(), op.value());
```
ramiro050, line 2456: This only works if the fill value is a float, which is not guaranteed.

I think it would make more sense to add support for this op similar to https://github.com/llvm/torch-mlir/blob/072c2b5aafdd5858bb19c5b296453a9cf0b65113/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L867. This would require the same, if not less, amount of code, and it would have the added benefit that it would work for multiple dtypes for the fill value.

ramiro050 (changes requested):

python/torch_mlir_e2e_test/test_suite/constant_alloc.py
```diff
@@ -1421,3 +1421,24 @@ def forward(self, x, mask):
...
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.int64, True),
+        ([-1, -1], torch.bool, True),
+    ])
+    def forward(self, x, mask):
+        value = torch.tensor(0.01)
```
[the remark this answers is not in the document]
pashu123: Done.

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -1332,3 +1332,23 @@ func.func @torch.aten.std.dim(%arg0: !torch.vtensor<[3,4,5],f32>) -> !torch.vten
...
+// CHECK-SAME:                  %[[ARG0:.*]]: !torch.vtensor<[2,3],f32>, 
+// CHECK-SAME:                  %[[ARG1:.*]]: !torch.vtensor<[2,3],i1>) -> !torch.vtensor<[2,3],f32> {
+// CHECK:  %[[FLOAT1:.*]] = torch.constant.float 1.000000e-02
+// CHECK:  %[[NONE:.*]] = torch.constant.none
+// CHECK:  %[[FALSE:.*]] = torch.constant.bool false
+// CHECK:  %[[FLOAT0D:.*]] = torch.aten.tensor.float %[[FLOAT1]], %[[NONE]], %[[NONE]], %[[FALSE]] : !torch.float, !torch.none, !torch.none, !torch.bool -> !torch.vtensor<[],f32>
+// CHECK:  %[[FLOATSCALAR:.*]] = torch.aten.Float.Tensor %[[FLOAT0D]] : !torch.vtensor<[],f32> -> !torch.float
+// CHECK:  %[[RET:.*]] = torch.aten.masked_fill.Scalar %[[ARG0]], %[[ARG1]], %[[FLOATSCALAR]] : !torch.vtensor<[2,3],f32>, !torch.vtensor<[2,3],i1>, !torch.float -> !torch.vtensor<[2,3],f32>
+// CHECK:  return %[[RET]] : !torch.vtensor<[2,3],f32>
+func.func @torch.aten.masked_fill.tensor(%arg0: !torch.vtensor<[2,3],f32>, %arg1: !torch.vtensor<[2,3],i1>) -> !torch.vtensor<[2,3],f32> {
```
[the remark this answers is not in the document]
pashu123: Not needed anymore.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
pashu123: I see, this makes sense. Thanks.

pashu123:
@ramiro050 PTAL.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -876,6 +876,19 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+  if (auto maskedFillTensor = dyn_cast<AtenMaskedFillTensorOp>(op)) {
+    AtenMaskedFillScalarOp::Adaptor adaptor(operands);
+    Type dtype = converter->convertType(maskedFillTensor.getType())
+                     .cast<RankedTensorType>()
+                     .getElementType();
+
+    Value input = payloadArgs[0];
+    Value mask = payloadArgs[1];
+    Value scalarValue =
+        b.create<tensor::ExtractOp>(loc, adaptor.value(), ValueRange({}));
```
ramiro050, line 888: The `linalg.generic` function created for the op should also be "iterating" over the fill value tensor, so you don't need to extract here. You should be able to do `payloadArgs[2]`

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
pashu123: Done.

pashu123: Thanks!

ramiro050 (approved):
Thanks for this! LGTM

Issue #1154, opened 2022-08-04
LTC Out-of-tree JIT linkage failure


henrytwo:
Currently LTC fails to build using the OOT CI config due to this linkage error: https://github.com/llvm/torch-mlir/runs/7549301528?check_suite_focus=true

Currently we work around this by setting `TORCH_MLIR_ENABLE_LTC=OFF` in cmake, which excludes LTC from compilation and generates a "fake" library in its place.

Additionally, the library type for `TorchMLIRJITIRImporter` is set to `MODULE` instead of `SHARED`. It looks like changing the module type to `SHARED` is the trigger for the aforementioned CI problems. However, we need this library to be `SHARED` in order to link with  LTC.

PR #1177, opened 2022-08-08
Reenable LTC in out-of-tree build

henrytwo:
Addressing #1154 

I'm not entirely sure what changed... but it looks like out-of-tree build works even with LTC enabled now

cc: @antoniojkim @ke1337


powderluv (approved):
Black magic

PR #1181, opened 2022-08-08
python: add explicit dependency for embedding Python into C++

ashay:
Without this patch, the build fails at my end (although I am unsure why
the build works in CI without this explicit dependency).

I confess I am going out on a limb here.  If someone knows if my local
build is broken, please let me know.  Thanks!

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/CMakeLists.txt
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/CMakeLists.txt
@@ -23,21 +23,14 @@ add_library(TorchMLIRJITIRImporter SHARED
   torch_to_mlir_utils.cpp
   )
 
+find_package(Python REQUIRED COMPONENTS Development)
+
 target_link_libraries(TorchMLIRJITIRImporter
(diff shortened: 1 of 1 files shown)


silvasean:
The CI builds only have 2 cores, and usually the resulting lower -j number exposes these missing dependencies which happen to race to completion with more cores.

silvasean:
Oh sorry, that is a different class of "missing dependency fails on CI". This one is truly mystifying. @stellaraccident @powderluv for insight.

powderluv:
Welcome to the land of Python deps.  

Though my experience in this is not as extensive as @stellaraccident couple points to note:

1: The nightly builds use a "static" Python so you can't remove the Python_LIBRARIES check. This is because manylinux ships various statically built python.
2: I am nervous about the find(Python). It is a thorn especially when you are adding LLVM/MLIR Python detection and PyTorch's together. At that point @stellaraccident would b required :D 

I would push this to a branch in torch-mlir and run the "One Shot nightly" to see if the Release build breaks with it (after the normal CI passes).

powderluv:
What OS / Python version are you on ? I can try to recreate it.

Issue #1186, opened 2022-08-09
Some TOSA ops started to fail on CI


powderluv:
https://github.com/llvm/torch-mlir/runs/7737767438?check_suite_focus=true

Going to XFAIL them for now:

```
****** Failed tests - 3 tests
    FAIL - "ElementwiseRemainderScalarModule_Float_basic"
        Compilation error: Traceback (most recent call last):
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 283, in compile_and_run_test
            compiled = config.compile(test.program_factory())
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/tosa_backend.py", line 40, in compile
            run_pipeline_with_repro_report(
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/compiler_utils.py", line 73, in run_pipeline_with_repro_report
            raise TorchMlirCompilerError(trimmed_message) from None
        torch_mlir.compiler_utils.TorchMlirCompilerError: Lower Torch Backend IR -> TOSA Backend IR failed with the following diagnostics:
        error: failed to legalize operation 'torch.constant.float'
        note: see current operation: %0 = "torch.constant.float"() {value = 2.000000e+00 : f64} : () -> !torch.float


        Error can be reproduced with:
        $ torch-mlir-opt -pass-pipeline='torch-backend-to-tosa-backend-pipeline' /tmp/ElementwiseRemainderScalarModule_Float.mlir
        Add '-mlir-print-ir-after-all -mlir-disable-threading' to get the IR dump for debugging purpose.


    FAIL - "ElementwiseRemainderScalarModule_Int_Float_basic"
        Compilation error: Traceback (most recent call last):
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/framework.py", line 283, in compile_and_run_test
            compiled = config.compile(test.program_factory())
          File "/home/runner/work/torch-mlir/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/torchscript/configs/tosa_backend.py", line 40, in compile
[...]

PR #1187, opened 2022-08-09
XFAIL some TOSA tests to unblock CI

powderluv:
Tracked in https://github.com/llvm/torch-mlir/issues/1186


powderluv:
no it was a bad commit. I have reverted it

Issue #1162, opened 2022-08-05
Dockerize CI for robust local reproducers


powderluv:
Opening an issue to discuss the PR #1156 and capture it here so it can live beyond a PR.

powderluv:
I would suggest we modify https://github.com/llvm/torch-mlir/blob/main/build_tools/python_deploy/build_linux_packages.sh
since it is used by the release scripts already. We should be able to expose the additional CI parameters https://github.com/llvm/torch-mlir/blob/31727f81d82537a6d30bb09b9c43d7ae2131ba40/.github/workflows/buildAndTest.yml#L15-L20 via any additional ENV variables. 

Then when we get to run_on_docker() we branch out based on env vars to call the corresponding flavors of CI builds or Release builds. (https://github.com/llvm/torch-mlir/blob/31727f81d82537a6d30bb09b9c43d7ae2131ba40/.github/workflows/buildAndTest.yml#L50-L101 or the release build with https://github.com/llvm/torch-mlir/blob/31727f81d82537a6d30bb09b9c43d7ae2131ba40/build_tools/python_deploy/build_linux_packages.sh#L101-L105 )

Since the docker image itself is a variable you can override the environment variable with any CUDA docker image etc when you run it. 

So locally when you want to build the CI versions you would do something like

```
TM_PYTHON_VERSIONS=cp310-cp310 MANYLINUX_DOCKER_IMAGE=nvidia/cuda:10.2-cudnn8-devel-ubuntu18.04 TORCH_BINARY=OFF ./build_tools/python_deploy/build_linux_packages.sh
```
We can have a list for all targets etc so you don't have to run it one by one. 

Couple points to note: We may have to expose the .ccache directory inside the docker container since our CIs use ccache extensively. 

Let me know if this makes sense if not we can discuss at the community meeting too.  I am trying to get the Release builds to move to PyTorch source builds but have to deal with nested pip sandboxes since we build our pip package and pytorch also with pip but that should be independent of these changes.

sjain-stanford:
[...]

PR #1188, opened 2022-08-09
Include torch-mlir builds in OOT workflow

sjain-stanford:
The matrix jobs in GHA are intended to be setup for checking "build + tests" for in-tree, but "build only" for out-of-tree workflows. However (as discovered [here](https://github.com/llvm/torch-mlir/issues/1162#issuecomment-1208762623)) at the moment building torch-mlir OOT is skipped in CI. We're just "building llvm" and "configuring cmake to build torch-mlir", but the build itself is [gated](https://github.com/llvm/torch-mlir/blob/f85ae9c685d6c9f2d43fe60fdad2ad5d3183ec52/.github/workflows/buildAndTest.yml#L103) for the OOT case with:
```
      if: matrix.llvmtype == 'binary'
```
partly because it is combined with the unit test target which is skipped for OOT. 

As seen below, both build and test steps are skipped for OOT case:
image

This PR splits the build from `check-torch-mlir-all` and moves it into the previous step (which have appropriate gating already).

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -36,7 +36,7 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
-    - name: Get torch-mlir
+    - name: Checkout torch-mlir
       uses: actions/checkout@v2
       with:
         submodules: 'true'
@@ -44,13 +44,13 @@ jobs:
       with:
         cache-suffix: ${{ matrix.os }}-${{ matrix.targetarch }}-${{ matrix.llvmtype }}-${{ matrix.llvmbuildtype }}
 
-    - name: llvm-binary-torch-src-or-binary
+    - name: Configure llvm-binary-torch-src-or-binary
       # Should be the fastest builds for CI and fails fast
       # OSX CMake flags are ignore on Linux
       if: matrix.llvmtype == 'binary'
       run: |
         cd $GITHUB_WORKSPACE
-        cmake $GITHUB_WORKSPACE/externals/llvm-project/llvm -B build -GNinja \
+        cmake -GNinja -Bbuild \
           -DCMAKE_BUILD_TYPE=Release \
           -DCMAKE_LINKER=lld \
           -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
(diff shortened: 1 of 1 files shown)


.github/workflows/buildAndTest.yml
```diff
@@ -66,15 +66,17 @@ jobs:
           -DTORCH_MLIR_USE_INSTALLED_PYTORCH=${{ matrix.torch-binary }} \
           -DCMAKE_OSX_ARCHITECTURES=${{ matrix.taregetarch }} \
           -DMACOSX_DEPLOYMENT_TARGET=10.15 \
-          -DLLVM_TARGETS_TO_BUILD=host
+          -DLLVM_TARGETS_TO_BUILD=host \
+          $GITHUB_WORKSPACE/externals/llvm-project/llvm
+        cmake --build build --target tools/torch-mlir/all
```
sjain-stanford, line 71: This is the in-tree build step that is now co-located with the cmake configuration for in-tree.

.github/workflows/buildAndTest.yml
```diff
@@ -100,8 +102,9 @@ jobs:
           -DTORCH_MLIR_ENABLE_LTC=OFF \
           -DPython3_EXECUTABLE=$(which python) \
           .
+        cmake --build build
```
sjain-stanford, line 105: This is the out-of-tree build step that is now co-located with the cmake configuration for out-of-tree.

powderluv: I would suggest we break the build and test into two and we can choose just the build for all flows

sjain-stanford: Yes, I made these changes this morning (in case you reviewed yesterday). Right now the builds are all colocated in their own step and both unit and integration tests are gated for OOT. Please LMK if I missed anything.

powderluv: I think we should just extract the `cmake --build build` it its own `-name: Build` -- this keeps config step separate from build step. We don't need to specify `--target tools/torch-mlir/all` since that should be default.

sjain-stanford: Ah I see what you mean. Will do.

sjain-stanford: Done. PTAL.

powderluv (approved):
nice.

PR #1195, opened 2022-08-09
Add note about MLIR compiled outputs in dev docs

### LINKED ISSUE #1175: Using Torch-MLIR as a front-end for MLIR files
We want to use different front-ends to look at the IR they produce and see if we can feed them into our optimising pipeline. Torch-MLIR seems to support a lot of the features we want, mainly being able to handle different MLIR dialects for the same models.

The idea is to get a model in Python, run through Torch-MLIR and get the MLIR file in the end. By looking at the tests, I see that there's a `python` folder with the JIT IR exporting MLIR files via `ModuleBuilder`. But it seems it only outputs the Torch dialect.

Further looking at the tests, I see in the `Conversion` folder that there's a `torch-mlir-opt` that converts the Torch dialect into the others. While I can see `MHLO` and `TOSA`, I can also see `Linalg`, `SCF` and `Std` separate from each other. From the docs, I assumed there was one that used all three together.

So I have two questions:
1. Is there a way to use the `ModuleBuilder` to export directly to a particular dialect? If not, is `torch-mlir-opt` the best option?
2. Can I export a full model to `Linalg` with `torch-mlir-opt`? Will it use `SCF` and `Std` for the remaining ops automatically?

@silvasean

rengolin:
Adding an example on how to extract MLIR output from the compilation
process in various different formats to the development documentation.

This should help developers trying to either debug torch_mlir or use it
for the purpose of extracting MLIR outputs for other testing.

Fixes #1175

### DIFF
--- a/development.md
+++ b/development.md
@@ -120,6 +120,28 @@ cmake --build build
 export PYTHONPATH=`pwd`/build/tools/torch-mlir/python_packages/torch_mlir:`pwd`/examples
 ```
 
+## Testing MLIR output in various dialects
+
+To test the compiler's output to the different MLIR dialects, you can use the example `examples/torchscript_resnet18_all_output_types.py`.
+
+Make sure you have activated the virtualenv and set the `PYTHONPATH` above:
+```shell
+source mlir_venv/bin/activate
+export PYTHONPATH=`pwd`/build/tools/torch-mlir/python_packages/torch_mlir:`pwd`/examples
+python examples/torchscript_resnet18_all_output_types.py
+```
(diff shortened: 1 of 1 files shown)


powderluv (commented):
Un-intended mhlo update ?

rengolin:
> Un-intended mhlo update ?

Argh, sorry! I built in-tree, probably changed something? Update now.

powderluv (approved):

silvasean:
Thanks for the PR!

PR #1198, opened 2022-08-09
Flip to C++17

jpienaar:
LLVM now uses C++17.


marbre:
Shouldn't we wait on the LLVM integrate commit before updating this setting?

PR #1200, opened 2022-08-09
E2E support for AtenRemainderScalarOp

vid-999:
trying to fix failing CI
CI fails on Tosa backend.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -803,6 +803,26 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     Value other = convertScalarToDtype(b, loc, operands[1], dtype);
     return b.create<arith::DivFOp>(loc, self, other);
   }
+  if (auto remScalar = dyn_cast<AtenRemainderScalarOp>(op)) {
+    Type newResultType = converter->convertType(remScalar.getType())
+                             .cast<RankedTensorType>()
+                             .getElementType();
+
+    Value self = convertScalarToDtype(b, loc, payloadArgs[0], newResultType);
+    Value other = convertScalarToDtype(b, loc, operands[1], newResultType);
+    Value result;
+
+    if (newResultType.isa<mlir::FloatType>()) {
+      result = b.create<arith::RemFOp>(loc, self, other);
+    } else if (newResultType.isa<mlir::IntegerType>()) {
+      result = b.create<arith::RemSIOp>(loc, self, other);
+    } else {
+      remScalar.emitError(
+          "Unsupported type encountered for AtenRemainderScalarOp.");
+    }
+
+    return result;
+  }
   if (auto reciprocal = dyn_cast<AtenReciprocalOp>(op)) {
     Type dtype = converter->convertType(reciprocal.getType())
                      .cast<RankedTensorType>()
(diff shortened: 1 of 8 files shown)


python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -1304,6 +1304,70 @@ def forward(self, x):
...
+
+class ElementwiseRemainderScalarModule_Int(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.int64, True),
```
ramiro050, line 1361: I don't think TOSA supports int64 inputs. Can you change this to int32?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -803,6 +803,23 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+
+    Value self = convertScalarToDtype(b, loc, payloadArgs[0], newResultType);
+    Value other = convertScalarToDtype(b, loc, operands[1], newResultType);
+    Value result;
+
+    if (newResultType.isa<mlir::FloatType>()) {
+      result = b.create<arith::RemFOp>(loc, self, other);
+    } else {
+      result = b.create<arith::RemSIOp>(loc, self, other);
+    }
```
ramiro050, line 819: The `newResultType` is not guaranteed to be float or int. For example, it could be a `bool`. So, each branch of the `if` needs to check that the type is the right type, and the `else` should be used to `emitError` when the type is not one of the supported types

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
vid-999: yeah that makes sense.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1681,14 +1701,14 @@ void mlir::torch::torch_to_linalg::populateUncategorizedPatternsAndLegality(
...
       AtenSubTensorOp, AtenLerpTensorOp, AtenSigmoidOp, AtenMinimumOp,
-      AtenAtan2Op, AtenMaximumOp, AtenToDtypeOp, AtenClampOp, AtenRsubScalarOp,
-      AtenLogOp, AtenErfOp, AtenSqrtOp, AtenFloorOp, AtenCeilOp,
-      AtenPowTensorScalarOp, AtenLog2Op, AtenLog1pOp, AtenRsqrtOp, AtenAbsOp,
-      AtenReciprocalOp, AtenBitwiseAndTensorOp, AtenGtScalarOp, AtenGeScalarOp,
-      AtenEqScalarOp, AtenLtScalarOp, AtenLeScalarOp, AtenWhereSelfOp,
-      AtenGtTensorOp, AtenEqTensorOp, AtenLtTensorOp, AtenThresholdOp,
-      AtenThresholdBackwardOp, AtenCloneOp, AtenSinOp, AtenCosOp,
-      AtenNeScalarOp, AtenMaskedFillScalarOp, AtenLogicalOrOp, AtenTriuOp>();
+      AtenMaximumOp, AtenToDtypeOp, AtenClampOp, AtenRsubScalarOp, AtenLogOp,
```
vivekkhandelwal1, line 1704: I think you have deleted the `AtenAtan2Op` by mistake from the list.

vivekkhandelwal1 (changes requested):

vivekkhandelwal1:
Also, squash all your commits into one.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
vivekkhandelwal1: From here: https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L941

vid-999: resolved.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -943,14 +963,14 @@ class ConvertElementwiseOp : public ConversionPattern {
...
-             AtenNeScalarOp, AtenNegOp, AtenMaskedFillScalarOp, AtenLogicalOrOp,
-             AtenTriuOp>(op))
+             AtenLog1pOp, AtenRsqrtOp, AtenDivScalarOp, AtenRemainderScalarOp,
+             AtenAbsOp, AtenReciprocalOp, AtenBitwiseAndTensorOp,
+             AtenGtScalarOp, AtenGeScalarOp, AtenEqScalarOp, AtenLtScalarOp,
+             AtenLeScalarOp, AtenWhereSelfOp, AtenCeilOp, AtenGtTensorOp,
+             AtenEqTensorOp, AtenLtTensorOp, AtenSubScalarOp, AtenAddScalarOp,
+             AtenThresholdOp, AtenThresholdBackwardOp, AtenCloneOp, AtenSinOp,
+             AtenCosOp, AtenNeScalarOp, AtenNegOp, AtenMaskedFillScalarOp,
+             AtenLogicalOrOp, AtenTriuOp>(op))
```
ramiro050, line 973: You also need to add your op to the illegal list

https://github.com/llvm/torch-mlir/blob/1abc541e532053484a90813b7216edb9fb800e73/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L1700

ramiro050 (approved):
LGTM. I just have one small change request

lib/Conversion/TorchToLinalg/Uncategorized.cpp
vid-999: resolved.

PR #1204, opened 2022-08-09
mac m1 cross compile

powderluv:
Add support for M1 cross compile. Also compile and test macos x86

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -74,7 +74,7 @@ jobs:
     - name: Configure llvm-binary-torch-src-or-binary
       # Should be the fastest builds for CI and fails fast
       # OSX CMake flags are ignored on Linux
-      if: matrix.llvmtype == 'binary'
+      if: ${{  matrix.llvmtype == 'binary' AND matrix.targetarch == 'x86_64' }}
       run: |
         cd $GITHUB_WORKSPACE
         cmake -GNinja -Bbuild \
@@ -99,7 +99,7 @@ jobs:
     - name: Configure llvm-source-out-of-tree-torch-src-or-binary
       # This build takes a while but is expected to almost always be cached.
       # A cache invalidation occurs when the committed LLVM version is changed.
-      if: matrix.llvmtype == 'source'
+      if: ${{  matrix.llvmtype == 'source' AND matrix.targetarch == 'x86_64' }}
       run: |
         cd $GITHUB_WORKSPACE
         cmake -GNinja -Bllvm-build \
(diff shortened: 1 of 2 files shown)


.github/workflows/buildAndTest.yml
```diff
@@ -44,9 +47,37 @@ jobs:
       with:
         cache-suffix: ${{ matrix.os }}-${{ matrix.targetarch }}-${{ matrix.llvmtype }}-${{ matrix.llvmbuildtype }}
 
+    - name: Configure llvm-cross-compile
+      # libzstd on GH Runners are only x86_64 to remove them.
+      if: matrix.targetarch == 'arm64'
+      run: |
+        sudo rm -rf /usr/local/lib/libzstd.*.dylib
```
jpienaar, line 54: Would -DLLVM_ENABLE_ZSTD =OFF below work also?

powderluv: Ah yes. I was copying what we were doing in IREE. 😀 I think ideally llvm upstream should check for architecture along with presence before linking. I know conda also adds it to the path inadvertently.

makslevental (approved):
CONGRATS

.github/workflows/buildAndTest.yml
```diff
@@ -25,14 +25,17 @@ jobs:
           - llvmtype: binary
             llvmbuildtype: out-of-tree
           # Disable M1 builds until https://github.com/llvm/torch-mlir/issues/1094 is fixed
-          - targetarch: AArch64
+          - targetarch: arm64
+            os: ubuntu-22.04
```
sjain-stanford, line 29: readability nit: maybe flip this to match the other pair of excludes?
```
- os: ubuntu-22.04
  targetarch: arm64

- os: macos-12
  targetarch: x86_64
```

sjain-stanford (approved):

.github/workflows/buildAndTest.yml
sjain-stanford: +1 on using the flag if it works.

sjain-stanford (changes requested):
I think `Run torch-mlir unit tests` step is still using x86_64 build for running tests:
```
cmake --build build --target check-torch-mlir-all
```

You may want to configure this with `_${{ matrix.targetarch }}` as well.

.github/workflows/buildAndTest.yml
powderluv: Will address in follow on

sjain-stanford (commented):
> Yeah we are building both architectures and testing only x86

This makes sense. The only consideration is the 2G ccache limit. As we're building twice in the same job / runner, we may always see low cache hit rate? Is it better to set this up cross compile build as a separate github workflow? This will also maximize parallelization (use a separate runner of which there are no restrictions I presume)? I'm hesitant in increasing the ccache limit of 2G more due to the reasons we discussed on discord.

powderluv:
I think we can drop the macOS x86 build all together. I think we can now reduce the matrix considerably.

sjain-stanford:
> I think we can drop the macOS x86 build all together. I think we can now reduce the matrix considerably.

You mean remove x86 build + tests for macos and only use arm64 build + tests? That would be great indeed :)

Issue #1126, opened 2022-08-01
LTC enablement on various compile flows


powderluv:
Tracking various modes to enable LTC in

* OOT Builds
* Source builds (?) 
* When built from setup.py (on Release builds)
* When build via seutp.py on macOS (cross compile arm64)

henrytwo:
> Source builds (?)

Building LTC from source works, as far as the configs used in CI. Is there an instance where you see it failing still?

powderluv:
Great to hear. Feel add / remove from the list. I thought it was disabled but maybe it wasn't

henrytwo:
Here's a ticket to track the OOT build problem: https://github.com/llvm/torch-mlir/issues/1154

I believe the other build configs hit the same linkage issues as OOT, so it's possible that this may resolve the other problems you listed too.

henrytwo:
@powderluv looks like OOT build is fixed now. I believe release build runs daily, so we'll find out if that works soon, but I'm not sure about running it on macOS. Is this something that can be triggered from CI?

powderluv:
Sorry to report the nightly builds just failed: https://github.com/llvm/torch-mlir/runs/7735569399?check_suite_focus=true

I triggered a "oneshot" nightly that can build Release linux and macOS builds. You can also run it on main or any torch-mlir branch. 

We probably should address the root cause or revert for now.

henrytwo:
😭 I'll revert for now and investigate tomorrow

henrytwo:
FYI I'm going to be working on fixing this problem in the following PR: https://github.com/llvm/torch-mlir/pull/1205

PR #1205, opened 2022-08-09
Reenable LTC in out-of-tree build (for real this time)

henrytwo:
This PR addresses https://github.com/llvm/torch-mlir/issues/1154 and https://github.com/llvm/torch-mlir/issues/1126, which means that LTC can build OOT -- which allows it to be included in our wheels 🎉 🎉 .

To summarize, a problem we were hitting before was that OOT build of LTC in CI would fail due to `TorchMLIRJITIRImporter` being switched to a `SHARED` library, in order to access some functions from LTC. This resulted in some nasty nasty pybind errors.

I was unable to root cause the source of the failure from changing the library type to `SHARED`; however, this PR gets around this problem by breaking up `TorchMLIRJITIRImporter` into two parts: a core library, and the pybind. This way, the pybind can remain ` MODULE`, which will prevent the original problem.

A new library called `TorchMLIRJITIRImporterPybind` has been created to handle the pybind, and `TorchMLIRJITIRImporter` has been converted to a static library -- due to problems linking on `macOS-arm64` with a shared library. Detailed logs are available [here](https://github.com/llvm/torch-mlir/runs/7907158318?check_suite_focus=true),

Even with this, LTC on `macOS-arm64` had been disabled to some other linkage issues, which will be addressed in the future here: https://github.com/llvm/torch-mlir/issues/1253.

This PR also makes it so that LTC is built by default, now that it works for the majority of our CI configs. LTC has been explicitly disabled on macOS-arm64.

cc: @ke1337 @antoniojkim

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -75,7 +75,6 @@ jobs:
           -DLLVM_EXTERNAL_TORCH_MLIR_DIALECTS_SOURCE_DIR="${GITHUB_WORKSPACE}/externals/llvm-external-projects/torch-mlir-dialects" \
           -DLLVM_TARGETS_TO_BUILD=host \
           -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
-          -DTORCH_MLIR_ENABLE_LTC=ON \
           -DTORCH_MLIR_USE_INSTALLED_PYTORCH="${{ matrix.torch-binary }}" \
           -DPython3_EXECUTABLE="$(which python)" \
           $GITHUB_WORKSPACE/externals/llvm-project/llvm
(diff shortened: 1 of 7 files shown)


powderluv (commented):
Looks good. Just to verify can you try the nighty release build to make sure it passes since there is a setup.py change.

henrytwo:
> Looks good. Just to verify can you try the nighty release build to make sure it passes since there is a setup.py change.

Yep, I'm running one now: https://github.com/llvm/torch-mlir/actions/runs/2886449151

henrytwo:
> > Looks good. Just to verify can you try the nighty release build to make sure it passes since there is a setup.py change.
> 
> Yep, I'm running one now: https://github.com/llvm/torch-mlir/actions/runs/2888787539

@silvasean @powderluv Release build ran thru successfully (only commits since then are comments, so it should be good to go)

henrytwo:
@powderluv can I get some stamps on this please, if there's nothing that I need to change? Release build was successful on latest commit

powderluv (approved):

PR #1206, opened 2022-08-09
Disable out-of-tree and PyTorch binary

powderluv:
trim OOT  + PyTorch binary builds

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -24,6 +24,8 @@ jobs:
             llvmbuildtype: in-tree
           - llvmtype: binary
             llvmbuildtype: out-of-tree
+          - llvmbuildtype: out-of-tree
+            torch-binary: OFF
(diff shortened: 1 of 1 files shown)


powderluv:
Remove OOT + PyTorch binary builds to trim down the CI options.

makslevental (approved):
LGTM

sjain-stanford (commented):
I may be missing a nuance but what does the `llvmbuildtype` option do in the workflow? It is unclear where this is used besides the cache key at the moment and seems redundant with `llvmtype`. Should we remove that option altogether, and use `llvmtype: binary` as the gating function here? Or maybe remove `llvmtype` and replace all uses with `llvmbuildtype`, as it is slightly better to read `in-tree` / `out-of-tree` than to read `source` vs `binary` especially when we have the same option for torch.

powderluv:
I think it is a good idea to revisit those options especially once we enable cross compile too. Let's land this and iterate as it reduces build options.

Issue #1182, opened 2022-08-09
Constant folding on Torch Dialect


ZihengJiang:
After converting a PyTorch model to Torch dialect, some parts are converted to this:

```
    %6 = torch.vtensor.literal(dense<5> : tensor<si64>) : !torch.vtensor<[],si64>
    ...
    %451 = torch.prim.NumToTensor.Scalar %int5 : !torch.int -> !torch.vtensor<[],si64>
    %452 = torch.aten.div.Tensor_mode %451, %6, %str : !torch.vtensor<[],si64>, !torch.vtensor<[],si64>, !torch.str -> !torch.vtensor<[],si64>
    %453 = torch.aten.Int.Tensor %452 : !torch.vtensor<[],si64> -> !torch.int
    %454 = torch.prim.ListConstruct %453, %int5, %int-1 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %455 = torch.aten.view %450, %454 : !torch.vtensor<[5,2048],f32>, !torch.list<int> -> !torch.vtensor<[?,5,?],f32>
```

As you can see, most of them can be pre-computed since the inputs are all constant integers, then we can get a static shaped %455. 

Any idea for this?

silvasean:
We already partially support this. It requires an extension for "div". We can generalize the logic here: https://github.com/llvm/torch-mlir/pull/935

ZihengJiang:
This issue is solved by https://github.com/llvm/torch-mlir/pull/1209 . Thanks @Vremold .

PR #1209, opened 2022-08-10
Generalize canonicalization pattern for more aten.sub/div/mul/add op

Vremold:
This PR generalizes the canonicalization pattern in PR #935 to fit more AtenOps (like `AtenSubTensorOp`, `AtenMulScalarOp`, and `AtenDivTensorModeOp`). The basic idea is eliminating these tensor ops when their tensor operands are zero-ranked tensors, which come from `ValueTensorLiteralOp` or `PrimNumToTensorScalarOp`, and replacing them with corresponding scalar ops. 
This PR also helps to solve issue #1182 .

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -843,8 +843,6 @@ LogicalResult rewrite0DBinaryTensorOp(Operation *op,
   auto lhs = getScalarValue(op->getOperand(0), loc, rewriter);
   auto rhs = getScalarValue(op->getOperand(1), loc, rewriter);
   auto outType = op->getResult(0).getType();
-  Value alpha =
-      rewriter.create<Torch::ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
 
   if (!lhs || !rhs) {
     return rewriter.notifyMatchFailure(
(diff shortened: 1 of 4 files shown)


Vremold:
> I see a lot of repeated code in this PR. Can we have a function like this which handles all ops? I think it would make it easier to share all the code and would avoid the templates
> 
> ```
> LogicalResult rewrite0DTensorOp(Operation *op, ...) {
>   // dyn_cast `op` to determine the operation.
> }
> ```

Thanks for this advice. I'll refactor these codes and try to support float dtype as well.

Vremold:
> I see a lot of repeated code in this PR. Can we have a function like this which handles all ops? I think it would make it easier to share all the code and would avoid the templates
> 
> ```
> LogicalResult rewrite0DTensorOp(Operation *op, ...) {
>   // dyn_cast `op` to determine the operation.
> }
> ```

Done!

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -821,41 +826,163 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+  if (auto stringConstruct = s().getDefiningOp<ConstantStrOp>())
+    return getI64IntegerAttr(getContext(),
+                             stringConstruct.valueAttr().getValue().size());
 
   return nullptr;
 }
 
+LogicalResult rewrite0DBinaryTensorOp(Operation *op,
+                                      PatternRewriter &rewriter) {
+  // dyn_cast `op` to determine the operation.
```
silvasean, line 838: nit: This comment is not necessary anymore

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -821,41 +826,163 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+  Location loc = op->getLoc();
+  if (op->getNumOperands() < 2) {
```
silvasean, line 840: Maybe add a note that >2 exists for some cases like aten.add.Scalar

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -821,41 +826,163 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+          op)) {
+    Value alpha = getScalarValue(op->getOperand(2), loc, rewriter);
+    if (!alpha) {
+      return rewriter.notifyMatchFailure(op,
+                                         "only int scalar alpha is supported");
+    }
+    rhs = rewriter.create<AtenMulIntOp>(loc, rhs, alpha);
+  }
+
+  // AtenDivTensorModeOp
```
silvasean, line 861: This  comment isn't necessary

silvasean (approved):

Vremold:
Merged. Thanks! @silvasean

Issue #1207, opened 2022-08-10
clean up llvmtype / buildtype in Github workflows


powderluv:
Tracking work items from the suggestion by @sjain-stanford https://github.com/llvm/torch-mlir/pull/1206#pullrequestreview-1067537720

powderluv:
The hopefully final missing piece of the CI (macOS m1 cross compile) is now up https://github.com/llvm/torch-mlir/issues/1207.  It builds arm64 and x86 separately and tests on x86.  Once that is in we can re-org the matrix types and clean up / reduce superfluous ones

sjain-stanford:
Perfect. I can work on this today. Did you also want me to include the last change we discussed [here](https://github.com/llvm/torch-mlir/pull/1204#issuecomment-1210900565)? 
> remove x86 build + tests for macos and only use arm64 build + tests

powderluv:
yes please. btw just mentioned it on discord -- maybe we can kill one of the Ubuntu builds and leave the macos x86/arm64 since it is redundant. It will probably be longer but worth checking what the runtime it when fully cached.

powderluv:
so if you address those comments I have no more changes in the pipeline. So we end up with three builds

1. macos x86 + arm64 builds and x86 tests (in-tree+pytorch src on macOS  x86 with tests and arm64 builds)
2. ubuntu in tree src build with pytorch binary (most used developer setup)
3. ubuntu out of tree src build with pytorch source (OOT is vendor used + pytorch source tests on Ubuntu)

What do you think ? This should help up us trim all the matrix combinations. Maybe even kill `targetarch` ?

sjain-stanford:
Sounds good except it'd be nice to decouple the mac builds because combining them into one workflow doesn't give us any benefit. The tests are pretty fast so we can run them for both mac builds, it's the build which [hurts us](https://github.com/llvm/torch-mlir/pull/1204#pullrequestreview-1068527446) and combining means we'll require increasing ccache limit to beyond 2G which will hurt overall (given 10G restrictions). 

How about this? IMO this offers reasonable coverage by testing pytorch source with both in-tree and out-of-tree, and also tests macos for both x86 and arm64. WDYT?
```
Current: 
ubuntu - x86_64 - pytorch binary - in-tree - build+test
ubuntu - x86_64 - pytorch source - in-tree - build+test
ubuntu - x86_64 - pytorch binary - out-of-tree - build+test
macos - x86_64 - pytorch binary - in-tree - build + test
[...]

PR #1213, opened 2022-08-11
Simplify matrix configuration for CI workflows

sjain-stanford:
Addresses https://github.com/llvm/torch-mlir/issues/1207. 

#### Provisioned jobs:
```
# ubuntu - x86_64 - llvm in-tree     - pytorch binary - build+test    # most used dev flow and fastest signal
# ubuntu - x86_64 - llvm out-of-tree - pytorch source - build+test    # most elaborate build
# macos  - arm64  - llvm in-tree     - pytorch source - build only    # cross compile, can't test arm64
```

#### Main changes
- [x] Spawn macos builds from a separate matrix (in the same workflow). It made sense to do this as they are fairly different from ubuntu (cross compile, use a different cmake configuration). This simplifies the matrix configuration and exclusions quite a bit, and makes the workflow a bit more tractable and maintenance friendly.
- [x] Remove the submodule md5sum step for ccache config. This was [broken](https://github.com/llvm/torch-mlir/runs/7779288734?check_suite_focus=true#step:3:145) for a while now.
- [x] Removes unused matrix options - `os`, `targetarch`, `python-version`, `llvmtype`.
- [x] Address ZSTD [comment](https://github.com/llvm/torch-mlir/pull/1204#discussion_r942349282) on @powderluv's cross compile [PR](https://github.com/llvm/torch-mlir/pull/1204). 

#### Further improvements (to be addressed in follow-on):
* ubuntu-x86_64 out-of-tree integration tests fail ([error](https://github.com/sjain-stanford/torch-mlir/runs/7781264029?check_suite_focus=true)); only run unit tests for now (tests are excluded in current CI too)

#### Passing workflow:
https://github.com/sjain-stanford/torch-mlir/actions/runs/2840676309
image

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -10,7 +10,7 @@ on:
 # Provisioned Jobs:
 # ubuntu - x86_64 - llvm in-tree     - pytorch binary - build+test    # most used dev flow and fastest signal
 # ubuntu - x86_64 - llvm out-of-tree - pytorch source - build+test    # most elaborate build
-# macos  - arm64  - llvm in-tree     - pytorch binary - build only    # cross compile, can't test arm64
+# macos  - arm64  - llvm in-tree     - pytorch source - build only    # cross compile, can't test arm64
 
 
 jobs:
@@ -144,9 +144,9 @@ jobs:
         llvm-build: [in-tree, out-of-tree]
         torch-binary: [ON, OFF]
         exclude:
-          # Exclude llvm in-tree and pytorch source
+          # Exclude llvm in-tree and pytorch binary
           - llvm-build: in-tree
-            torch-binary: OFF
+            torch-binary: ON
           # Exclude llvm out-of-tree altogether
           - llvm-build: out-of-tree
 
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -1,182 +1,193 @@
 name: Build and Test
 
 on:
-  push:
-    branches:
-      - main
   pull_request:
+    branches: [ main ]
+  push:
+    branches: [ main ]
   workflow_dispatch:
 
+# Provisioned Jobs:
+# ubuntu - x86_64 - llvm in-tree     - pytorch binary - build+test    # most used dev flow and fastest signal
+# ubuntu - x86_64 - llvm out-of-tree - pytorch source - build+test    # most elaborate build
+# macos  - arm64  - llvm in-tree     - pytorch binary - build only    # cross compile, can't test arm64
+
+
 jobs:
-  build-validate:
+  ubuntu-build:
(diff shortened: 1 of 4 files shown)


powderluv:
This is looking great. The arm64 Pytorch src build should work if you remove this one line https://github.com/llvm/torch-mlir/blob/51bfe25c898e5f7a09294ccc012cfbf1327614c9/build_tools/build_libtorch.sh#L134  
It is trying to uninstall the systemwide torch.

.github/workflows/buildAndTest.yml
```diff
@@ -1,182 +1,193 @@
...
+          -DLLVM_EXTERNAL_PROJECTS="torch-mlir;torch-mlir-dialects" \
+          -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR="$GITHUB_WORKSPACE" \
+          -DLLVM_EXTERNAL_TORCH_MLIR_DIALECTS_SOURCE_DIR="${GITHUB_WORKSPACE}/external/llvm-external-projects/torch-mlir-dialects" \
+          -DLLVM_TARGETS_TO_BUILD=AArch64 \
+          -DLLVM_USE_HOST_TOOLS=ON \
+          -DLLVM_ENABLE_ZSTD=OFF \
+          -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
+          -DTORCH_MLIR_ENABLE_LTC=OFF \
+          -DTORCH_MLIR_ENABLE_MHLO=OFF \
+          -DTORCH_MLIR_USE_INSTALLED_PYTORCH="${{ matrix.torch-binary }}" \
```
powderluv, line 186: We can delete this https://github.com/llvm/torch-mlir/blob/51bfe25c898e5f7a09294ccc012cfbf1327614c9/build_tools/build_libtorch.sh#L134 in either this commit or a follow on and switch to src.

sjain-stanford:
> The arm64 Pytorch src build should work if you remove this one line

Thanks, patched. Let's wait to see if the arm64 pytorch source workflow goes through with the fix. If there are more errors, I can revert to pytorch binary and land it for now (to avoid cache evictions the longer this is open). I'll wait for an "all green" CI before landing, but if this looks good otherwise, please feel free to ✅  this.

powderluv (approved):
LGTM. Feel free to switch arm64 pytorch source in a follow on.

powderluv:
nicely done. The silly cache gets generated again when it merges so we got to wait for it again

sjain-stanford:
> The silly cache gets generated again when it merges so we got to wait for it again

Ah I was wondering why it didn't restore from cache after landing because the keys didn't change. Good to know this is normal. Maybe it treats GHA runs on PRs differently than runs on push to main. Oh well...

sjain-stanford:
... and thank you for the help in reviewing it!

PR #1214, opened 2022-08-12
python: trim registration and loading of dialects and passes

ashay:
In the interest of merging upstream LLVM quickly, a previous patch
(7f08169) updated the torch-mlir build to register all dialects and
passes through Python bindings.  This patch limits the dialects and
passes to only those that are used in torch-mlir.

Key to this change are the removal of
`MLIRPythonExtension.RegisterEverything` and the introduction of a new
Python module (`_mlir_libs/_site_initialize_0.py`), where we register
the dialects and passes used by torch-mlir.

Release build now passes for this branch: https://github.com/llvm/torch-mlir/actions/runs/2988625814

### DIFF
--- a/lib/CAPI/CMakeLists.txt
+++ b/lib/CAPI/CMakeLists.txt
@@ -14,6 +14,34 @@ add_mlir_public_c_api_library(TorchMLIRCAPI
 
   LINK_LIBS PUBLIC
   MLIRIR
+  MLIRAffineDialect
+  MLIRArithDialect
+  MLIRBufferizationDialect
+  MLIRControlFlowDialect
+  MLIRFuncDialect
+  MLIRLinalgDialect
+  MLIRMathDialect
+  MLIRMemRefDialect
+  MLIRSCFDialect
+  MLIRTensorDialect
+  MLIRTosaDialect
+  MLIRAffineToStandard
+  MLIRArithToLLVM
+  MLIRArithTransforms
+  MLIRBufferizationTransforms
+  MLIRControlFlowToLLVM
+  MLIRFuncToLLVM
+  MLIRFuncTransforms
+  MLIRLinalgToLLVM
+  MLIRLinalgTransforms
+  MLIRMathToLLVM
+  MLIRMemRefToLLVM
+  MLIRReconcileUnrealizedCasts
+  MLIRSCFToControlFlow
+  MLIRSCFTransforms
+  MLIRTensorTransforms
+  MLIRTosaToArith
+  MLIRTosaToLinalg
   MLIRSupport
   TorchMLIRTorchDialect
   TorchMLIRInitAll
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -50,6 +50,7 @@ declare_mlir_python_sources(TorchMLIRPythonSources.TopLevel
   SOURCES
     __init__.py
     compiler_utils.py
+    _mlir_libs/_site_initialize_0.py
 )
 
 declare_mlir_python_sources(TorchMLIRPythonSources.Dialects
@@ -127,8 +128,8 @@ set(_source_components
   # the only way to handle that is to create a separate mlir python package
   # tree, which seems excessive.
   MLIRPythonSources
-  MLIRPythonExtension.Core
-  MLIRPythonExtension.RegisterEverything
+  MLIRPythonSources.Core
+  MLIRPythonSources.ExecutionEngine
   TorchMLIRPythonSources
   TorchMLIRPythonExtensions
 )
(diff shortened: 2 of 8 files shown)


lib/CAPI/CMakeLists.txt
```diff
@@ -11,9 +11,37 @@ add_mlir_public_c_api_library(TorchMLIRCAPI
   LINK_COMPONENTS
   Core
 
+  DEPENDS
+  MLIRAffinePassIncGen
```
stellaraccident, line 15: Just so I understand, can you comment on how you came to conclude that these needed to be depended on? Ordinarily, I only see explicit dependencies on the generated files from internal to a target and would have expected that we add the actual transform/pass target itself to the LINK_LIBS PUBLIC.

(you may be right to do this but double checking)

stellaraccident: (I would have expected the LINK_LIBS depends below to have transitively insured generation of these files)

lib/CAPI/CMakeLists.txt
```diff
@@ -11,9 +11,37 @@ add_mlir_public_c_api_library(TorchMLIRCAPI
...
+  MLIRArithmeticTransformsIncGen
+  MLIRBufferizationPassIncGen
+  MLIRConversionPassIncGen
+  MLIRFuncTransformsIncGen
+  MLIRLinalgPassIncGen
+  MLIRSCFPassIncGen
+  MLIRTensorTransformsIncGen
+
   LINK_LIBS PUBLIC
```
stellaraccident, line 24: For the dialect headers you include in the corresponding C file, it would be better to include the dialect libraries here (versus counting on transitively getting them from one of these deps)

python/torch_mlir/dialects/torch/__init__.py
```diff
@@ -4,4 +4,4 @@
 # Also available under a BSD-style license. See LICENSE.
 
 from .._torch_ops_gen import *
-from ..._mlir_libs._torchMlir import register_dialect
```
stellaraccident, line 7: With the new scheme, do you think we need this at all? Iiuc, it was mainly here because there was no other way to register things (and now it is automatic).

test/python/smoketest.py
```diff
@@ -4,4 +4,4 @@
 from torch_mlir.dialects import torch
 
 with torch_mlir.ir.Context() as ctx:
-  torch.register_dialect(ctx)
```
stellaraccident, line 7: Depending on what you do above, can remove this part of the test, but it would be nice to maybe create one of the torch types or something (which would just test that things in fact did register).

python/CMakeLists.txt
```diff
@@ -121,15 +122,9 @@ add_subdirectory(torch_mlir/eager_mode)
...
-  # build burden by ~5x. Make it stop.
-  # TODO: Reduce dependencies. We need ExecutionEngine and a bunch of passes
-  # for the reference backend, but logically they can be separate. But seemingly
-  # the only way to handle that is to create a separate mlir python package
-  # tree, which seems excessive.
   MLIRPythonSources
-  MLIRPythonExtension.Core
-  MLIRPythonExtension.RegisterEverything
+  MLIRPythonSources.Core
+  MLIRPythonSources.ExecutionEngine
```
stellaraccident, line 127: At some point (not now), it would be good to remove this and see what impact that has on what needs to be built/deployed size. Based on that data, we could determine if more surgery to split the execution engine is justified.

stellaraccident (approved):
Nice! I glanced through the CI logs and looks reasonable. I expect we are paying a signficant part of the remaining tax for the execution engine for the ref backend. We should get that data at some point and look if there is a faster-to-iterate way to separate it.

stellaraccident (commented):
Looks like something broke the ref backend tests. The output from the failures is pretty slim and that limits any further insights I might have (I wish we biased towards verbose output on failure for such things -- much more CI friendly).

lib/CAPI/CMakeLists.txt
stellaraccident: Ok, fine to be explicit like this for now, but can you add a TODO to that effect so we can better diagnose it?

silvasean:
> Looks like something broke the ref backend tests. The output from the failures is pretty slim and that limits any further insights I might have (I wish we biased towards verbose output on failure for such things -- much more CI friendly).

We actually have good errors here, but we are squelching them when running our e2e tests in parallel due to it being really hard to capture that info for a crash with Python `multiprocessing`. I suppose for the CI we could run in single-threaded mode since we only have 2 cores anyways. But yes, this was a regression when multiprocessing this.

silvasean:
(the usual answer is to run single-threaded with -s, but that doesn't help on the CI or causes an extra iteration step)

powderluv:
You get 4 vCPUs on Linux and 3 for macOS

PR #1216, opened 2022-08-12
Merge matrix runs to fail fast globally

sjain-stanford:
My earlier[ PR](https://github.com/llvm/torch-mlir/pull/1213) had (among other things) decoupled ubuntu and macos builds into separate matrix runs. This is not working well due to limited number of MacOS GHA VMs causing long queue times and backlog. There are two reasons causing this backlog: 

1. macos arm64 builds with pytorch source are getting erratically cancelled due to resource / network constraints. This is addressed with this: https://github.com/llvm/torch-mlir/pull/1215

> "macos-arm64 (in-tree, OFF) The hosted runner: GitHub Actions 3 lost communication with the server. Anything in your workflow that terminates the runner process, starves it for CPU/Memory, or blocks its network access can cause this error."

2. macos runs don't fail-fast when ubuntu runs fail due to being in separate matrix setups. This PR couples them again.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -14,12 +14,11 @@ on:
 
 
 jobs:
-  ubuntu-build:
-    name: ubuntu-x86_64
-    runs-on: ubuntu-22.04
+  build-test:
     strategy:
       fail-fast: true
       matrix:
+        os-arch: [ubuntu-x86_64, macos-arm64]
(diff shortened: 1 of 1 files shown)


sjain-stanford:
@powderluv PTAL. Marking this as `Draft` until we analyze the CI logs to ensure this is doing what we expect it to.

powderluv (approved):
lets merge and iterate.

PR #1218, opened 2022-08-12
build: fix typo in path

ashay:
When we renamed the directory containing submodules from `external` to
`externals`, we accidentally left the original name in the Github
workflow.  This patch fixes the problem.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -65,7 +65,7 @@ jobs:
           -DLLVM_ENABLE_PROJECTS=mlir \
           -DLLVM_EXTERNAL_PROJECTS="torch-mlir;torch-mlir-dialects" \
           -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR="$GITHUB_WORKSPACE" \
-          -DLLVM_EXTERNAL_TORCH_MLIR_DIALECTS_SOURCE_DIR="${GITHUB_WORKSPACE}/external/llvm-external-projects/torch-mlir-dialects" \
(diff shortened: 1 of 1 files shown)


powderluv:
yeah good catch. I don't know how it worked or maybe those dialects are not being tested ?. Good to land.

powderluv (approved):

sjain-stanford:
This is interesting. Should someone look into this? How did it work before this change? If it is not needed, should we remove it?

powderluv:
are there any tests that use it ?

sjain-stanford:
I ran the cmake build + tests in my local docker setup. It works if the path is incorrect, but if I remove the flag altogether, I get some errors due to CMakeLists expecting a dependency: 

```
CMake Error at /opt/src/torch-mlir/CMakeLists.txt:174 (add_dependencies):
  The dependency target "check-torch-mlir-dialects" of target
  "check-torch-mlir-all" does not exist.


CMake Error at cmake/modules/AddLLVM.cmake:521 (add_dependencies):
  The dependency target "TorchMLIRTMTensorDialect" of target
  "obj.TorchMLIRInitAll" does not exist.
Call Stack (most recent call first):
  /opt/src/torch-mlir/externals/llvm-project/mlir/cmake/modules/AddMLIR.cmake:303 (llvm_add_library)
  /opt/src/torch-mlir/lib/CMakeLists.txt:6 (add_mlir_library)
```

I have a feeling this is not being well tested though. As in, it shouldn't expect this flag and not really read from it to build something and then test its use. Any ideas @ashay , @powderluv , @silvasean?

powderluv:
what ? you are taking the weekend off from torch-mlir ?  / just kidding enjoy your weekend.

PR #1220, opened 2022-08-12
Fix unused-variables warnings about EmbeddingBag ops

ramiro050:
According to the documentation for
`torch.embedding_bag` (https://pytorch.org/docs/stable/generated/torch.nn.functional.embedding_bag.html),
the default value for `scale_grad_by_freq` is False.

### DIFF
--- a/lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
@@ -218,10 +218,22 @@ public:
     Value weight = adaptor.weight();
     Value indices = adaptor.indices();
     Value offsets = adaptor.offsets();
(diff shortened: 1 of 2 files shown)


vid-999:
Do you know why the build is failing?

vid-999 (commented):
looks good to me!
Can you look into why the E2E test for embeddingBag is failing?

ramiro050:
> looks good to me! Can you look into why the E2E test for embeddingBag is failing?

I was checking for `if (!scaleGradByFreqBool)` here

https://github.com/llvm/torch-mlir/blob/7d5d068ab94043ca7bce22f40b28a158908e7ad5/lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp#L233

PR #1221, opened 2022-08-13
build: streamline options for enabling LTC and MHLO

ashay:
Prior to this patch, LTC was enabled by default, but it was disabled in
setup.py and in the Github CI rules.  MHLO, on the other hand, was
enabled by default, but also (i.e. redundantly) enabled in setup.py and
in the Github CI rules.

This had the effect of LTC being left enabled unless the build is
triggered by setup.py or Github CI.  It also had the effect of forcing
us to update the option in multiple places, whenever it needed to be
changed.

This patch sets the options' default value at the option definintion,
and it explicitly sets the value only when the default is overriden,
like for disabling MHLO in in-tree builds on macOS.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -75,7 +75,7 @@ jobs:
           -DLLVM_EXTERNAL_TORCH_MLIR_DIALECTS_SOURCE_DIR="${GITHUB_WORKSPACE}/externals/llvm-external-projects/torch-mlir-dialects" \
           -DLLVM_TARGETS_TO_BUILD=host \
           -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
-          -DTORCH_MLIR_ENABLE_MHLO=ON \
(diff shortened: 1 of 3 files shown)


.github/workflows/buildAndTest.yml
```diff
@@ -75,7 +75,6 @@ jobs:
           -DLLVM_EXTERNAL_TORCH_MLIR_DIALECTS_SOURCE_DIR="${GITHUB_WORKSPACE}/externals/llvm-external-projects/torch-mlir-dialects" \
           -DLLVM_TARGETS_TO_BUILD=host \
           -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
-          -DTORCH_MLIR_ENABLE_MHLO=ON \
```
powderluv, line 78: I assume you want a `-DTORCH_MLIR_ENABLE_LTC=ON` here

sjain-stanford (approved):
LGTM, thank you!

PR #1222, opened 2022-08-13
s/external/externals/g

sjain-stanford:
Fix remaining instances of `external/llvm-project`.


sjain-stanford:
@ashay , @powderluv looks like these were also leftover from the rename to `externals`. PTAL.

powderluv:
that explains why it worked.

powderluv (approved):

PR #1228, opened 2022-08-15
Add support for returning more than one copy of the same tensor

ramiro050:
One of the simplifications made by the pass `RefinePublicReturn`
currently only happens if the tensor in question only has one
user. However, the current method of checking this does not correctly
handle the case of a user having multiple uses of the same
tensor. This commit makes sure only unique users are considered.

### DIFF
--- a/lib/Dialect/Torch/Transforms/RefinePublicReturn.cpp
+++ b/lib/Dialect/Torch/Transforms/RefinePublicReturn.cpp
@@ -71,7 +71,9 @@ class RefinePublicReturnPass
           // If the return (or transitively other ops) are not the only users,
           // then we can't be sure that the tensor hasn't been mutated, so stop
           // here.
-          if (!llvm::hasSingleElement(copy->getUsers()))
+          SetVector<Operation *> users(copy->getUsers().begin(),
+                                       copy->getUsers().end());
+          if (users.size() != 1)
             break;
           newOperand = copy.getOperand();
         } else {
--- a/test/Dialect/Torch/refine-public-return.mlir
+++ b/test/Dialect/Torch/refine-public-return.mlir
@@ -59,3 +59,16 @@ func.func @called(%arg0: tensor<*xf32>) -> tensor<*xf32> {
 ^bb2:
   return %arg0 : tensor<*xf32>
 }
+
+// -----
+
+// CHECK-LABEL:   func.func @return_multiple_copies_of_tensor(
+// CHECK-SAME:                                                %[[ARG:.*]]: !torch.vtensor<[],f32>) -> (!torch.vtensor<[],f32>, !torch.vtensor<[],f32>) {
+// CHECK:           %[[CAST:.*]] = torch.tensor_static_info_cast %[[ARG]] : !torch.vtensor<[],f32> to !torch.vtensor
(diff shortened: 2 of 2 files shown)


lib/Dialect/Torch/Transforms/RefinePublicReturn.cpp
```diff
@@ -71,7 +71,10 @@ class RefinePublicReturnPass
           // If the return (or transitively other ops) are not the only users,
           // then we can't be sure that the tensor hasn't been mutated, so stop
           // here.
-          if (!llvm::hasSingleElement(copy->getUsers()))
+          SmallVector<Operation *> usersVector(copy->getUsers());
```
silvasean, line 74: I think this is buggy if there are noncontiguous uses like one use by return, one use by other, another use by return.

Perhaps `SetVector<Operation *> users(copy->getUsers().begin(), copy->getUsers().end); if (users.size() == 1) ...`

silvasean (approved):
Fascinating. I always though that getUsers was deduped.

silvasean:
Did you actually run into this in a real test? I would have thought that by construction in our pipeline we wouldn't hit this situation of multiple uses. If that's the case, then we can change it to a check that there is only one use.

ramiro050:
> Fascinating. I always though that getUsers was deduped.

Yeah, I thought it was weird too.

> Did you actually run into this in a real test? I would have thought that by construction in our pipeline we wouldn't hit this situation of multiple uses. If that's the case, then we can change it to a check that there is only one use.

I ran into this while running the [unit tests](https://github.com/ramiro050/iree-torch/blob/ddd74fb54616b672e7f18f46a6c247d48633ed2a/torchdynamo_poc/torchdynamo_test.py#L79) I had written for my torchdynamo compiler. It is definitely something that can be modified to not return two copies of the same tensor. In real world applications, I imagine this case is not common.

>  I would have thought that by construction in our pipeline we wouldn't hit this situation of multiple uses

What aspects of the pipeline would guarantee this?

silvasean:
> > Did you actually run into this in a real test? I would have thought that by construction in our pipeline we wouldn't hit this situation of multiple uses. If that's the case, then we can change it to a check that there is only one use.
> 
> I ran into this while running the [unit tests](https://github.com/ramiro050/iree-torch/blob/ddd74fb54616b672e7f18f46a6c247d48633ed2a/torchdynamo_poc/torchdynamo_test.py#L79) I had written for my torchdynamo compiler. It is definitely something that can be modified to not return two copies of the same tensor. In real world applications, I imagine this case is not common.

Okay, that is real enough. Let's land this.

> 
> > I would have thought that by construction in our pipeline we wouldn't hit this situation of multiple uses
> 
> What aspects of the pipeline would guarantee this?

I was thinking that something could end up inserting one copy per operand of the return op which would guarantee that the return op always had distinct values. But I guess that isn't the case here since you hit it from "user code".

lib/Dialect/Torch/Transforms/RefinePublicReturn.cpp
ramiro050: Done!

silvasean (approved):

PR #1232, opened 2022-08-16
Use Docker for Bazel Build

sjain-stanford:
A follow-on to https://github.com/llvm/torch-mlir/pull/1225 to launch bazel builds in GHA (as well as locally) inside a docker container. Requires #1225 to land first.

Review just the following files (non bazel diffs are split into #1225 for ease of review):
* `.github/workflows/bazelBuildAndTest.yml`
* `build_tools/docker/run_bazel_build.sh`
* `utils/bazel/.bazelrc`

### DIFF
--- a/build_tools/docker/Dockerfile
+++ b/build_tools/docker/Dockerfile
@@ -26,6 +26,9 @@ RUN apt-get update                                              && \
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 10
 RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10
 
+RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-10 10
+RUN update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-10 10
+
 # Install bazel
 RUN wget -q https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSION}/bazel-${BAZEL_VERSION}-linux-${ARCH} -O /usr/bin/bazel \
     && chmod a+x /usr/bin/bazel
--- a/build_tools/docker/Dockerfile
+++ b/build_tools/docker/Dockerfile
@@ -16,7 +16,7 @@ RUN apt-get update                                              && \
     git                                                            \
     python3-pip                                                    \
     wget                                                           \
-    clang                                                          \
+    clang-10                                                       \
     automake                                                       \
     libtool                                                        \
     curl                                                           \
--- a/build_tools/docker/Dockerfile
+++ b/build_tools/docker/Dockerfile
@@ -1,0 +1,40 @@
+ARG BASE_IMG=ubuntu:18.04
+FROM ${BASE_IMG} as dev-base
+
+ARG ARCH="x86_64"
+ARG TARGETARCH="amd64"
+ARG BAZEL_VERSION=4.2.1
+
+# Install basic packages
+RUN apt-get update                                              && \
+    apt-get install -y                                             \
+    python3.8                                                      \
+    python3.8-dev                                                  \
+    cmake                                                          \
+    ccache                                                         \
+    ninja-build                                                    \
+    git                                                            \
+    python3-pip                                                    \
+    wget                                                           \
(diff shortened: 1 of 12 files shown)


utils/bazel/.bazelrc
```diff
@@ -2,10 +2,10 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-build --action_env=CC=clang
-build --action_env=CXX=clang++
-build --cxxopt=-std=c++17 
-build --host_cxxopt=-std=c++17 
+build --action_env=CC=clang-10
+build --action_env=CXX=clang-cpp-10
```
sjain-stanford, line 6: @asaadaldien PTAL. In the past we'd decided to not send this upstream because clang on GHA VMs was already new, but with Ubuntu 18 container the default clang is pretty old and fails with the AVX errors.

build_tools/docker/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+    apt-get install -y                                             \
+    python3.8                                                      \
+    python3.8-dev                                                  \
+    cmake                                                          \
+    ccache                                                         \
+    ninja-build                                                    \
+    git                                                            \
+    python3-pip                                                    \
+    wget                                                           \
+    clang-10                                                       \
```
sjain-stanford, line 19: @asaadaldien PTAL. Previously we'd decided to not send this upstream because clang on GHA VMs was already new, but with Ubuntu 18 container the default clang is old and fails with the AVX errors.

sjain-stanford: Actually, this breaks CMake workflow because it is looking for `clang` and not `clang-10` :) Looking into a way to symbolic link so as to not update every `-DCMAKE_C_COMPILER=clang` with `-DCMAKE_C_COMPILER=clang-10`.

sjain-stanford: Looks like this is what I need: https://man7.org/linux/man-pages/man1/update-alternatives.1.html (and we're already doing this for python).

build_tools/docker/Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+    libtool                                                        \
+    curl                                                           \
+    make                                                           \
+    unzip
+
+RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 10
+RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10
+
+RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-10 10
+RUN update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-10 10
```
sjain-stanford, line 30: @asaadaldien with this we can avoid having to configure both bazel and cmake builds to pick the specific version.

powderluv:
Do you have a hard requirement for Ubuntu 18.04 ? The reason we moved to 22.04 in the CI is to get newer tools and python version by default

powderluv:
You could add Bazel as an option in https://github.com/llvm/torch-mlir/pull/1234 and then anyone building other CIs / Releases can also test Bazel builds locally via docker.

sjain-stanford:
> You could add Bazel as an option in https://github.com/llvm/torch-mlir/pull/1234 and then anyone building other CIs / Releases can also test Bazel builds locally via docker.

@powderluv we'd like to avoid coupling Bazel CI with the CMake release builds and merge gating CI because it is a very orthogonal flow with almost nothing in common (deps, build, test scripts etc). It adds to the complexity but with little to no ROI. We believe this is acceptable since users do not have to fix a broken bazel build (requiring them to run things locally), but if users want to help fix bazel builds, we'll provide easy-to-run scripts to repro bazel CI locally. Please LMK if you see any strong reasons against this.

sjain-stanford:
Closing as this was coupled with the earlier Dockerize CI PR (#1225 ). Will send a separate one out for dockerizing bazel CI.

silvasean:
+1 to not coupling Bazel. With our current policy, it should definitely be kept as an orthogonal thing.

PR #1235, opened 2022-08-16
[MLIR][TORCH] Fix dynamic cases for aten.index.Tensor

vivekkhandelwal1:


python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -1748,6 +1748,54 @@ def IndexTensorMultiInputOneDim_basic(module, tu: TestUtils):
...
+
+
+@register_test_case(module_factory=lambda: IndexTensorMultiInputNonContiguousOneDimDynamic())
+def IndexTensorMultiInputNonContiguousOneDimDynamic_basic(module, tu: TestUtils):
+    module.forward(tu.rand(5, 4, 3), torch.randint(4, (6, 1)), torch.randint(3, (3,)))
+
+
+# ==============================================================================
+
+
```
qedawkins, line 1798: Can you add a few of the following cases as tests:
1. index1 and index2 are rank 2 with both dims dynamic, and the dim sizes are different
2. index1, index2, and index3 are all rank 1, and index1 is statically sized at [2] (anything except 1) while index2 and index3 are dynamically sized at [2]
3. Same as one of the above test cases, but swap the order of index1 and index2 when performing the index.Tensor op

qedawkins (changes requested):

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -602,12 +602,12 @@ class ConvertAtenIndexTensorOp : public OpConversionPattern<AtenIndexTensorOp> {
       for (auto i : llvm::seq(startIndex, startIndex + maxRank)) {
         auto resultDimSize = refinedResultShape[i];
         if (ShapedType::isDynamic(resultDimSize)) {
-          return rewriter.notifyMatchFailure(
-              op, "unimplemented: index tensors must have static shape if "
-                  "there is more than one index tensor");
+          broadcastedIndexShape.push_back(
+              getDimOp(rewriter, loc, indexTensors[i - startIndex], 0));
+        } else {
+          broadcastedIndexShape.push_back(getConstant(
```
silvasean, line 608: For this type of case analysis, we should be adding .mlir lit tests demonstrating each case -- this makes it easy to refactor the pattern later.

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -602,12 +602,36 @@ class ConvertAtenIndexTensorOp : public OpConversionPattern<AtenIndexTensorOp> {
       for (auto i : llvm::seq(startIndex, startIndex + maxRank)) {
         auto resultDimSize = refinedResultShape[i];
         if (ShapedType::isDynamic(resultDimSize)) {
-          return rewriter.notifyMatchFailure(
-              op, "unimplemented: index tensors must have static shape if "
-                  "there is more than one index tensor");
+          SmallVector<Value> dynamicDims;
+          int64_t staticDim = -1;
```
qedawkins, line 606: nit: I think calling this staticDimSize is a bit clearer

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -602,12 +602,36 @@ class ConvertAtenIndexTensorOp : public OpConversionPattern<AtenIndexTensorOp> {
...
+            RankedTensorType indexTensorType =
+                indexTensor.getType().cast<RankedTensorType>();
+            int64_t indexTensorRank = indexTensorType.getRank();
+            if ((maxRank - indexTensorRank) > (i - startIndex))
+              continue;
+            int64_t dim = i - startIndex - maxRank + indexTensorRank;
+            if (ShapedType::isDynamic(indexTensorType.getShape()[dim]))
+              dynamicDims.push_back(getDimOp(rewriter, loc, indexTensor, dim));
+            else
+              staticDim = indexTensorType.getShape()[dim];
```
qedawkins, line 617: I think there is one case missing here where there could be multiple static dims with one of them being 1 and the other being non-1. This should be the maximum between the current staticDim and new static dim values to handle this case. This also probably warrants an E2E test.

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -602,12 +602,36 @@ class ConvertAtenIndexTensorOp : public OpConversionPattern<AtenIndexTensorOp> {
...
+          }
+          if (dynamicDims.size() >= 2)
+            return rewriter.notifyMatchFailure(
+                op, "aten.index.Tensor: unimplemented case");
```
qedawkins, line 621: nit: This failure message could be clearer. e.g. "unimplemented: index tensors with overlapping dynamic dims"

qedawkins (changes requested):
I left some comments with one possible failure case. @silvasean does this satisfy what you were looking for with the mlir tests? Otherwise LGTM.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -1748,6 +1748,124 @@ def IndexTensorMultiInputOneDim_basic(module, tu: TestUtils):
...
+        return torch.ops.aten.index(x, (index1, None, index2))
+
+
+@register_test_case(module_factory=lambda:
+                    IndexTensorMultiInputNonContiguousMultipleStaticDims())
+def IndexTensorMultiInputNonContiguousMultipleStaticDims_basic(
+        module, tu: TestUtils):
+    module.forward(tu.rand(5, 4, 3, 2), torch.randint(3, (4, 1)),
+                   torch.randint(1, (4, 3)))
+
```
qedawkins, line 1864: This test case doesn't properly test the multiple static dims case I was thinking about. It should be "vertically" more than 1 rather than "horizontally". Basically we need 3 index tensors where one has a static dim of size n, one has a static dim of size 1, and the last has a dynamic dim of size n.

qedawkins (approved):

qedawkins:
LGTM

PR #1238, opened 2022-08-17
Add E2E support for aten.div.int

albertdmath:


python/torch_mlir_e2e_test/test_suite/scalar.py
```diff
@@ -127,6 +127,29 @@ def DivFloatModule_basic(module, tu: TestUtils):
...
 
+class DivIntModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([], torch.float32, True),
```
pashu123, line 138: We need int32 dtype annotations here right? @albertdmath

python/torch_mlir_e2e_test/test_suite/scalar.py
```diff
@@ -127,6 +127,29 @@ def DivFloatModule_basic(module, tu: TestUtils):
...
+        ([], torch.float32, True),
+    ])
+    def forward(self, lhs, rhs):
+        return lhs / rhs
+
+
+@register_test_case(module_factory=lambda: DivIntModule())
+def DivIntModule_basic(module, tu: TestUtils):
+    module.forward(torch.rand(()).int(), torch.rand(()).int())
```
pashu123, line 147: We can add multiple test cases - rand() always limits the range between 0 to 1 and hence test cases are limited.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -658,8 +658,8 @@ void TypeAnalysis::visitOperation(Operation *op,
           AtenZero_Op, AtenIndexTensorOp, ValsemVariantAtenIndexPutImplOp,
           AtenIndexPutOp, ValsemVariantAtenCopyOp, AtenZeroOp,
           AtenIndexPutHackedTwinOp, AtenMaskedFillScalarOp, AtenFlipOp,
-          PrimAbsScalarOp, AtenNumpyTOp, AtenTriuOp, AtenMaskedFillTensorOp>(
-          op)) {
+          PrimAbsScalarOp, AtenNumpyTOp, AtenTriuOp, AtenMaskedFillTensorOp,
+          AtenDivIntOp>(op)) {
```
ramiro050, line 662: I would have expected `AtenDivIntOp` to not require any `RefineTypes` handling since it always takes scalar ints as inputs and produces scalar floats as outputs. Is this change needed?

python/torch_mlir_e2e_test/test_suite/scalar.py
```diff
@@ -127,6 +127,29 @@ def DivFloatModule_basic(module, tu: TestUtils):
...
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([], torch.float32, True),
+        ([], torch.float32, True),
+    ])
+    def forward(self, lhs, rhs):
+        return lhs / rhs
```
ramiro050, line 142: nit: can you use `torch.ops.aten.div`, to make it clear what is being tested? Also, you should be casting the arguments to ints to make sure you're not testing the `div.Tensor` variant, i.e.

```python
     return torch.ops.aten.div(int(lhs), int(rhs))
```

python/torch_mlir_e2e_test/test_suite/scalar.py
```diff
@@ -127,6 +127,29 @@ def DivFloatModule_basic(module, tu: TestUtils):
...
+class DivIntModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([], torch.float32, True),
+        ([], torch.float32, True),
```
ramiro050, line 139: these annotations don't match the input type. They should be `torch.int`s

python/torch_mlir_e2e_test/test_suite/scalar.py
ramiro050, line 147: `torch.rand()` only produces values between zero and one, so this only tests division between ints [0,1]. It would be better to use something like `torch.randint(-100, 100, ())` to test division between a variety of ints

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2462,6 +2462,23 @@ class DecomposeAten_EmbeddingBagOp
 };
 } // namespace
 
+// Decompose aten.div.int into aten.div.float
+namespace {
+class DecomposeAtenDivIntOp : public OpRewritePattern<AtenDivIntOp> {
```
ramiro050, line 2467: Decompositions should also have lit tests in https://github.com/llvm/torch-mlir/blob/main/test/Dialect/Torch/decompose-complex-ops.mlir

ramiro050 (changes requested):
Thanks for this! Below are a few comments

albertdmath:
``` torch_mlir.compiler_utils.TorchMlirCompilerError: Lower Torch Backend IR -> Linalg-on-Tensors Backend IR failed with the following diagnostics:
        error: 'torch_c.from_i64' op operand #0 must be 64-bit signless integer, but got 'i32'
        note: see current operation: %5 = "torch_c.from_i64"(%4) : (i32) -> !torch.int```

File: test/Dialect/Torch/decompose-complex-ops.mlir [hunk 7]
```diff
@@ -1332,3 +1332,17 @@ func.func @torch.aten.std.dim(%arg0: !torch.vtensor<[3,4,5],f32>) -> !torch.vten
...
+
+// -----
+// CHECK-LABEL:  func.func @torch.aten.div.int(
+// CHECK-SAME:                            %[[LHS:.*]]: !torch.int,
+// CHECK-SAME:                            %[[RHS:.*]]: !torch.int) -> !torch.float {
+// CHECK:          %[[LHS_I64:.*]] = torch_c.to_i64 %[[LHS]]
+// CHECK:          %[[RHS_I64:.*]] = torch_c.to_i64 %[[RHS]]
+// CHECK:          %[[SUB:.*]] = arith.divf %[[LHS_I64:.*]], [[RHS_I64:.*]] : i64
+// CHECK:          %[[OUT:.*]] = torch_c.from_i64 %[[SUB:.*]]
+// CHECK:          return %[[OUT:.*]] : !torch.float
```
ramiro050, line 1344: This is not the right IR for the decomposition. The decomposition only creates `torch` dialect ops, but here you have `arith.divf` from the `arith` dialect. You can generate the right checks by using https://github.com/llvm/llvm-project/blob/2dde4ba63974daf59f8ce5c346505f194f920131/mlir/utils/generate-test-checks.py

ramiro050 (changes requested):

test/Dialect/Torch/decompose-complex-ops.mlir
albertdmath: I get the following error when running the script you showed me ```bash: generate-test-checks.py: command not found```
Also, Prashant told me ```There is no lowering of atenFloatScalar op from torch to linalg.``` This could be a problem, but I see it's being used in ```ReifyShapeCalculations.cpp``` does this still mean lowering is needed?

ramiro050: You need to call the script by doing `python generate-test-checks.py ...`. You can get more info on the script by passing the `-h` flag.

For the `AtenFloatScalar`, you should just need a `folder`, handling only the case where the input is a constant int, since your norm op only uses the `AtenDivInt` op to divide 1 by `p`, and `p` can be passed as a constant in the e2e tests

ramiro050: You need to call the script as `python generate-test-checks.py ...` You can get more info on the script by passing the `-h` flag.

Looking a bit closer at what you need from `AtenDivInt` to get the norm op to work, I think all you need is the `fold` function and the tests for the folding. There is already a `fold` function for `AtenFloatScalar`. In your norm op all you need to do is make sure you hard-code the value for `p` rather than passing it as a parameter

PR #1240, opened 2022-08-17
Dockerize and Cache Bazel {Local, CI} Builds

sjain-stanford:
This PR adds:
- A minimal docker wrapper to the bazel GHA workflow to make it reproducible locally
 - Bazel cache to speed up GHA workflows (down to ~5 minutes from ~40+minutes)

This is a no-op for non-bazel workflows and an incremental improvement.

PR #1241, opened 2022-08-17
[Bazel] Check cache directory exists before changing owners

sjain-stanford:
This fixes a seeding issue with the [previous PR](https://github.com/llvm/torch-mlir/pull/1240) where bazel build's GHA cache is not present to begin with and one of the commands (chown) fails on it. Should get the Bazel build back to green.

### DIFF
--- a/.github/workflows/bazelBuildAndTest.yml
+++ b/.github/workflows/bazelBuildAndTest.yml
@@ -31,9 +31,15 @@ jobs:
         path: ~/.cache/bazel
         key: ubuntu_x86_64_torch_mlir_bazel_build_cache
 
+    # Change bazel cache directory to root ownership
+    # to allow writing to it from within the docker container.
+    # If no cache hits, this directory is not present
+    # so don't run chown (will error otherwise).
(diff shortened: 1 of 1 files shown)


silvasean (approved):

sjain-stanford:
Co-incidentally, there's another failure after the rebase to `develop` (after [this commit](https://github.com/llvm/torch-mlir/commit/57681f794764a34c34e2be7f07f7dfbcafa683c1)):
```
bazel-out/k8-fastbuild/bin/external/torch-mlir/_objs/TorchMLIRTorchPasses/Passes.pic.o:Passes.cpp:function (anonymous namespace)::registerLowerToBackendContractPass()::{lambda()#1}::operator()() const: error: undefined reference to 'mlir::torch::Torch::createLowerToBackendContractPass(int, bool)'
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Let me land this now, and send a follow-on PR to include `"lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp",` to the `TorchMLIRTorchPasses` target.

PR #1172, opened 2022-08-08
[MHLO] Add configurable op decomposition rules

Vremold:
See RFC #999 
We register some options in `TorchLoweringPipeline` and `DecomposeComplexOpsPass` to make the decomposition rules configurable.

Co-authored-by: Bairen Yi <yibairen.byron@bytedance.com>
Co-authored-by: Jiawei Wu <xremold@gmail.com>
Co-authored-by: Tianyou Guo <tianyou.gty@alibaba-inc.com>
Co-authored-by: Xu Yan <yancey.yx@alibaba-inc.com>
Co-authored-by: Ziheng Jiang <ziheng.jiang@bytedance.com>

PR #1242, opened 2022-08-17
Add a concept of "backend legal ops".

silvasean:
This is a first step towards formalizing the set of ops in our backend
contract. The goal is to eventually formalize `torch` dialect ops into 3
categories:
1. Legal in backend contract
2. Illegal in backend contract
3. Conditionally legal in backend contract

The "conditionally legal" set are the ops that we can optionally
decompose for backends.

This patch adds relevant pass options for this throughout the compiler,
in preparation for a new set of traits which will formalize this
classification.


silvasean:
@Vremold  This should provide the same functionality as https://github.com/llvm/torch-mlir/pull/1172 but it is more aligned with the direction we are going for the backend contract definition.

ZihengJiang (approved):
LGTM. Thanks @silvasean

Vremold (approved):
Great. Thanks!

PR #1245, opened 2022-08-18
[MHLO] Eliminate explicit dynamic output shape generating in converting AtenSliceTensorOp

Vremold:
This PR passed an additional `output type` arg when converting AtenSliceTensorOp to MHLO dialect. Since mhlo.RealDynamicSliceOp's canonicalization pattern requires static output shape, the benefits of passing output type is to eliminate explicit output shape generating, which introduces dynamic shape even in static patterns, and thus helpful to canonicalize mhlo.RealDynamicSliceOp.

### DIFF
--- a/lib/Conversion/TorchToMhlo/ViewLike.cpp
+++ b/lib/Conversion/TorchToMhlo/ViewLike.cpp
@@ -53,8 +53,8 @@ Value getNormalizedDimSizeInternal(PatternRewriter &rewriter, Operation *op,
 }
 
 Value getDynamicSliceInternal(PatternRewriter &rewriter, Operation *op,
-                              Value input, Value startIndex, Value endIndex,
-                              Value step, size_t dimIndex,
+                              Type outTy, Value input, Value startIndex,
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchToMhlo/ViewLike.cpp
```diff
@@ -192,7 +189,7 @@ LogicalResult ConvertAtenOp<AtenSliceTensorOp>::matchAndRewrite(
   llvm::Optional<Value> step = getOptionalVal(adaptor.step());
 
   FailureOr<Value> sliceInfo =
-      getDynamicSlice(rewriter, op, self, start, end, step, dim);
+      getDynamicSlice(rewriter, op, outTy, self, start, end, step, dim);
   if (failed(sliceInfo))
     return op.emitError("can not create a dynmaic slice");
```
fortianyou, line 195: Nit: L196:L198 can be removed if we had set the same output type as it was before the conversion.

fortianyou (approved):
LGTM

PR #1246, opened 2022-08-18
Drop Python 3.7x from the nightly binary builds

powderluv:
It should still build from anything 3.6+ (until MLIR changes the minimum version). We have seen zero downloads with Python 3.7.


silvasean:
SGTM

silvasean (approved):

sjain-stanford (approved):

PR #1249, opened 2022-08-19
[cleanup] Change OutputType enum values to strings

silvasean:
The use of numbers was arbitrary and was preventing the enum values from
being put in the natural order.


fortianyou (approved):
LGTM! Thanks @silvasean

PR #1251, opened 2022-08-19
[MLIR][TORCH] Add torch.Device type to backend contract scalar types

vivekkhandelwal1:
This error was coming up for various models:
```
  File "v_diffusion_model_only.py", line 208, in <module>
    module = torch_mlir.compile(
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/__init__.py", line 217, in compile
    run_pipeline_with_repro_report(mb.module,
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/compiler_utils.py", line 73, in run_pipeline_with_repro_report
    raise TorchMlirCompilerError(trimmed_message) from None
error: unsupported by backend contract: type '!torch.Device'
note: see current operation: %554 = "torch.constant.device"() {value = "cpu"} : () -> !torch.Device
```
because of `!torch.Device` missing from the list of valid scalar types satisfying backend contract. This commit fixes this issue by adding the type `!torch.Device` to the list.

IR dump of the v-diffusion model which throws this error: https://gist.github.com/e0d8bd737b698868238db176fa054565

### DIFF
--- a/e2e_testing/torchscript/xfail_sets.py
+++ b/e2e_testing/torchscript/xfail_sets.py
@@ -124,6 +124,7 @@ TOSA_PASS_SET = {
     "OnesModuleInt_basic",
     "OnesModuleFloat_basic",
     "OnesModuleFalsePinMemory_basic",
+    "OnesModuleCPUDevice_basic",
     "NewZerosModuleDefaultDtype_basic",
     "NewZerosModuleInt2D_basic",
     "NewZerosModuleInt3D_basic",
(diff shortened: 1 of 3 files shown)


sjain-stanford (commented):
It might be worth adding a test with a minimal repro (that would break if this source fix was reverted).

sjain-stanford (approved):
Thank you for adding the test.

silvasean (approved):

PR #1260, opened 2022-08-22
[MLIR][TORCH] Fix CI failure due to failing tests

vivekkhandelwal1:


powderluv (approved):
Should we file a bug to track the root cause?

vivekkhandelwal1:
> Should we file a bug to track the root cause?

I don't think there's a bug. It's just a new op, and its tests were added here https://github.com/llvm/torch-mlir/commit/c38308f3ef0a93dd0909ac9db86facec5b66f313, but the LTC failing tests list was not updated.

PR #1261, opened 2022-08-22
Iteratively run SimplificationPipeline until code optimization converges

Vremold:
This PR is proposed to solve the problem below. 

```mlir
func.func @torch_module.forward(%arg0: !torch.nn.Module<"torch_module">, %arg1: !torch.tensor) -> !torch.tensor {
  ...
  %str = torch.constant.str "trunc"
  %548 = torch.tensor.literal(dense<5> : tensor<si64>) : !torch.tensor<[],si64>
  %549 = torch.aten.size.int %arg1, %int0 : !torch.tensor, !torch.int -> !torch.int
  %550 = torch.prim.NumToTensor.Scalar %549 : !torch.int -> !torch.tensor
  %551 = torch.aten.div.Tensor_mode %550, %548, %str : !torch.tensor, !torch.tensor<[],si64>, !torch.str -> !torch.tensor
  %552 = torch.aten.Int.Tensor %551 : !torch.tensor -> !torch.int
  %553 = torch.prim.ListConstruct %552, %int5, %int-1 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %554 = torch.aten.view %arg1, %553 : !torch.tensor, !torch.list<int> -> !torch.tensor
  ...
}
```
In the above case, %arg1 has a constant size (5) along the 0-th dimension.It can be deduced that the shape of %554 should be completely static, but the actual output is [?,5,?]. The cause is that before shape refinement pipeline runs, the %551 is not known to be a constant. That is further caused by the canonicalization of AtenDivTensorModeOp fails as it relys on the shape refinement pipeline itself to deduce that one of its operand, %550, is a constant.
So the buggy IR pattern is like this:
1. Shape refinement pipeline reasons out some constants;
2. Based on the constants, we can further canonicalize some ops;
3. The canonicalization produces more constants, which are related to the shape of some tensors;
4. ...

By executing canonicalizer pass and shapeRefinementPipeline multiple times, this PR can help to alleviate this problem.

### DIFF
--- a/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
+++ b/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
@@ -197,6 +197,40 @@ static bool satisfiesBackendContract(ModuleOp module,
   return true;
 }
 
+// Forward declaration
+static llvm::hash_code hashOperation(Operation *);
+static llvm::hash_code hashBlock(Block &block) {
+  llvm::hash_code hash(0);
+  for (Operation &op : block.getOperations()) {
+    llvm::hash_code opHash = hashOperation(&op);
+    hash = llvm::hash_combine(hash, opHash);
+  }
+  return hash;
+}
+
+static llvm::hash_code hashRegion(Region &region) {
+  llvm::hash_code hash(0);
+  for (Block &block : region.getBlocks()) {
+    llvm::hash_code blockHash = hashBlock(block);
+    hash = llvm::hash_combine(hash, blockHash);
+  }
+  return hash;
+}
+
+static llvm::hash_code hashOperation(Operation *op) {
+  llvm::hash_code hash(0);
+  llvm::hash_code opHash = OperationEquivalence::computeHash(
+    op, OperationEquivalence::ignoreHashValue,
+    OperationEquivalence::ignoreHashValue,
+    OperationEquivalence::IgnoreLocations);
+  hash = llvm::hash_combine(hash, opHash);
+  for (auto &region : op->getRegions()) {
+    llvm::hash_code regionHash = hashRegion(region);
+    hash = llvm::hash_combine(hash, regionHash);
+  }
+  return hash;
+}
+
 namespace {
 class LowerToBackendContractPass
     : public LowerToBackendContractBase<LowerToBackendContractPass> {
@@ -217,6 +251,9 @@ public:
     options.backendLegalOps = backendLegalOps;
     createTorchSimplificationPipeline(pm, options);
 
+    bool codeChanged = false;
+    llvm::hash_code moduleHash = hashOperation(module);
+
     int i = 0;
     do {
       if (i++ == maxIterations) {
@@ -234,7 +271,14 @@ public:
 
       if (failed(runPipeline(pm, module)))
         return signalPassFailure();
-    } while (!satisfiesBackendContract(module));
+      
+      llvm::hash_code newModuleHash = hashOperation(module);
+      codeChanged = (moduleHash != newModuleHash);
+      moduleHash = newModuleHash;
+    
+    // Iterate until maxIterations is reached or 
+    // backend contract is satisified and code optimization converges.
+    } while (!satisfiesBackendContract(module) || codeChanged);
     LLVM_DEBUG({
       llvm::dbgs() << "LowerToBackendContractPass: "
                    << "succeeded after " << i


silvasean:
createTorchSimplificationPipeline is already running iteratively inside LowerToBackendContractPass after https://github.com/llvm/torch-mlir/pull/1165 Is this still an issue?

lib/Dialect/Torch/Transforms/Passes.cpp
```diff
@@ -124,12 +124,14 @@ void mlir::torch::Torch::createTorchSimplificationPipeline(
   pm.addNestedPass<func::FuncOp>(Torch::createMaximizeValueSemanticsPass());
   // Update the return op to return value tensors.
   pm.addPass(Torch::createRefinePublicReturnPass());
-  pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
-  // Do shape refinement.
-  // This should be run before RefineTypes (which primarily does dtype
-  // inference), because Torch type promotion rules actually depend on the shape
-  // of the operand.
-  createTorchShapeRefinementPipeline(pm);
+  for (int i = 0; i < options.maxIterations; ++i) {
```
silvasean, line 127: maxIterations controls the iterations in LowerToBackendContractPass and should not be used this way.

Vremold: > createTorchSimplificationPipeline is already running iteratively inside LowerToBackendContractPass after #1165 Is this still an issue?

Yes. In my local development environment, if only run ShapeRefinementPipeline once, the output IR of the pipeline is something like this:
``` mlir
// -----// IR Dump After DropShapeCalculations (torch-drop-shape-calculations) //----- //
...
%5 = torch.vtensor.literal(dense<5> : tensor<si64>) : !torch.vtensor<[],si64>
%210 = torch.prim.NumToTensor.Scalar %int5 : !torch.int -> !torch.vtensor<[],unk>
%211 = torch.aten.div.Tensor_mode %210, %5, %str_4 : !torch.vtensor<[],unk>, !torch.vtensor<[],si64>, !torch.str -> !torch.vtensor<[],unk>
%212 = torch.aten.Int.Tensor %211 : !torch.vtensor<[],unk> -> !torch.int
%213 = torch.prim.ListConstruct %212, %int5, %int-1 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
%214 = torch.aten.view %209, %213 : !torch.vtensor<[5,2048],unk>, !torch.list<int> -> !torch.vtensor<[?,5,?],unk>
...
```

fortianyou: I think the issue is related to RefineTypes. Shall we add `PrimNumToTensorScalar` to https://github.com/llvm/torch-mlir/blob/bb47c166a047ddbc913950cb0f2571594c0fe261/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L638?

Vremold: I think this issue is only about constant shape. Could you explain more on the RefineTypes's influence on this problem? In my experiments, the type inference for `PrimNumToTensorScalarOp` will fail if we do so.
```mlir
// -----// IR Dump After RefineTypes (torch-refine-types) //----- //
%210 = torch.prim.NumToTensor.Scalar %int5 : !torch.int -> !torch.vtensor<[],unk>
```

ZihengJiang:
@silvasean The problem might be `satisfiesBackendContract` returns true after one iteration, but this doesn't mean that the code is optimized to the converged status, which means we can still utilize `ShapeRefinement` and `Canonicalization` to infer some information afterwards.

silvasean:
> @silvasean The problem might be `satisfiesBackendContract` returns true after one iteration, but this doesn't mean that the code is optimized to the converged status, which means we can still utilize `ShapeRefinement` and `Canonicalization` to infer some information afterwards.

Good point. I would recommend that we run until satisfiedBackendContract is true, and then keep iterating until no changes are made to the program (and of course not iterating more than maxIterations times). This can be done by hashing the module. I think OperationEquivalence can help with this: https://mlir.llvm.org/doxygen/structmlir_1_1OperationEquivalence.html

Vremold:
This approach also introduces problems. We need at least one additional SimplificationPipeline execution to make sure that no code change happened. This will result in at least one refine-types pass running after decompose-complex-ops pass of last iteration. But decompose-complex-ops pass probably introduces some ops like torch.aten.empty.memory_format, this is not supported in refine-types pass and will finally caused a compilation failure
```
torch-mlir-opt: /root/share/up/torch-mlir/lib/Dialect/Torch/Transforms/RefineTypes.cpp:1118: void (anonymous namespace)::TypeAnalysis::incorporateKnowledge(mlir::Value, const (anonymous namespace)::ValueKnowledge &): Assertion `updatedKnowledge.has_value() && "IR has contradictory type!"' failed.
...
 #9 0x00000000005bb8ee (anonymous namespace)::TypeAnalysis::visitOperation(mlir::Operation*, llvm::ArrayRef<mlir::dataflow::Lattice<(anonymous namespace)::ValueKnowledge> const*>, llvm::ArrayRef<mlir::dataflow::Lattice<(anonymous namespace)::ValueKnowledge>*>) RefineTypes.cpp:0:0
#10 0x0000000001a64afc mlir::dataflow::AbstractSparseDataFlowAnalysis::visitOperation(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a64afc)
#11 0x0000000001a64569 mlir::dataflow::AbstractSparseDataFlowAnalysis::initializeRecursively(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a64569)
#12 0x0000000001a64675 mlir::dataflow::AbstractSparseDataFlowAnalysis::initializeRecursively(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a64675)
#13 0x0000000001a644c8 mlir::dataflow::AbstractSparseDataFlowAnalysis::initialize(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a644c8)
#14 0x0000000001a54f52 mlir::DataFlowSolver::initializeAndRun(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a54f52)
#15 0x00000000005baa26 (anonymous namespace)::RefineTypesPass::runOnOperation() RefineTypes.cpp:0:0
...
[1]    1836385 abort (core dumped)  torch-mlir-opt -pass-pipeline='torchscript-module-to-torch-backend-pipeline'
```
This problem is already found in the CI failure. I'm not sure if there are other cases where it fails.
Considering that IR optimization mainly happens in ShapeRefinementPipeline and canonicalizer, perhaps a better solution is that we iteratively run these two passes instead of the entire SimplificationPipeline if we want more optimization. This will have minimal impact on the order in which the passes are run.

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -197,6 +197,40 @@ static bool satisfiesBackendContract(ModuleOp module,
...
+    llvm::hash_code blockHash = hashBlock(block);
+    hash = llvm::hash_combine(hash, blockHash);
+  }
+  return hash;
+}
+
+static llvm::hash_code hashOperation(Operation *op) {
+  llvm::hash_code hash(0);
+  llvm::hash_code opHash = OperationEquivalence::computeHash(
+    op, OperationEquivalence::ignoreHashValue,
```
silvasean, line 223: can you add `/*arg=*/` comments here so we know the arg names?

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -197,6 +197,40 @@ static bool satisfiesBackendContract(ModuleOp module,
   return true;
 }
 
+// Forward declaration
+static llvm::hash_code hashOperation(Operation *);
+static llvm::hash_code hashBlock(Block &block) {
+  llvm::hash_code hash(0);
+  for (Operation &op : block.getOperations()) {
+    llvm::hash_code opHash = hashOperation(&op);
```
silvasean, line 205: we should hash the block arguments too (at least their types)

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -234,7 +271,14 @@ class LowerToBackendContractPass
 
       if (failed(runPipeline(pm, module)))
         return signalPassFailure();
-    } while (!satisfiesBackendContract(module));
+      
+      llvm::hash_code newModuleHash = hashOperation(module);
```
silvasean, line 275: can you add a unit test of the hashing logic?

Vremold:
Recently, I'm thinking of the execution order of the passes in TorchSimplificationPipeline. Their are two main requirements:
1. To satisfy backend contract, we need to run the whole TorchSimplificationPipeline iteratively. But it will probably cause problems like in PR #1280 . Will it be better to execute `decompose-complex-ops` pass before shape or type refinement passes? And it also sounds more natural if we decompose complex ops before performing optimizations.
2. To perform more code optimization, we need to run shape-refinement-pipeline and canonicalizer iteratively. If we want minimal effects on the execution order of passes and try to avoid executing TorchSimplificationPipeline multiple times, I think a better solution is that we create another pass which run shape-refinement-pipeline and canonicalizer iteratively until no code change happens or a maxIterations is reached and insert it in the current TorchSimplificationPipeline.
I'm not sure if my idea is feasible and meets for the demands. Thanks for any comments ahead.

PR #1262, opened 2022-08-22
Don't explicitly set MLIR_PDLL_TABLEGEN_EXE

marbre:
With llvm/llvm-project@91b6f76, the variable `MLIR_PDLL_TABLEGEN_EXE` is
set as a cache variable in MLIR upstream.

Likely requires an update of externals/mlir-hlo to
tensorflow/mlir-hlo@4f2a00b or later.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,9 +106,6 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
 else()
   message(STATUS "Torch-MLIR in-tree build.")
   # In-tree build with LLVM_EXTERNAL_PROJECTS=torch-mlir
-  if (TORCH_MLIR_ENABLE_MHLO)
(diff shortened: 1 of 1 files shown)


jpienaar (approved):
Change looks good, are you going to wait until the bump happens? (I'm assuming no presubmits fail here for related reason as this is only enabled for the MHLO and that's not yet part of presubmit).

powderluv (approved):
Thank you

marbre:
> Change looks good, are you going to wait until the bump happens? (I'm assuming no presubmits fail here for related reason as this is only enabled for the MHLO and that's not yet part of presubmit).

I know that a similar fix "broke" IREE but I think in that case the underlying LLVM-integrate was missing. That is not the case here (the LLVM version used includes llvm/llvm-project@91b6f76) though we could probably merge without breaking anything. However, the change might only fully functional *after* MLIR-HLO is up-to-date enough as well.

PR #1265, opened 2022-08-23
Roll Torch-MLIR_Architecture.png into main diagram

powderluv:
Update all references to `torch` dialect
Roll the Torch-MLIR_Architecture.png contents into main diagram
Remove torch_dispatch() which was originally FX lowering


silvasean (changes requested):
The "backend contract" box is super important for the architecture.md use of Torch-MLIR_Architecture.png, but probably not for the README. I think these two diagrams (README and architecture.md) have two slightly different target audiences, so let's keep them as separate files for now.

Also, the Torch dialect is actually used throughout the "purple" boxes in the Torch-MLIR "frontend". It is really deceptive to have it only in the backend.

Overall I think the original diagram is really good, and we just need to remove the torch_dispatch and replace Torch-MLIR Dialct with "Torch Dialect".

sjain-stanford:
Gentle check-in on this :) 

I believe this is good to go with these two boxes removed?

image

powderluv:
> Overall I think the original diagram is really good, and we just need to remove the torch_dispatch and replace Torch-MLIR Dialct with "Torch Dialect".

Updated per @silvasean  comment.  PTAL

silvasean (approved):

sjain-stanford:
> Remove torch_dispatch() which was originally FX lowering

I don't think this was removed in the commit you just force pushed @powderluv .

powderluv:
Ah ok. I missed the first part. Thanks for fixing in the follow on

PR #1269, opened 2022-08-23
Add documentation for adding E2E tests

ramiro050:
This commit adds a file explaining the structure of an E2E test, as
well as useful things to keep in mind when adding new tests.

### DIFF
--- a/docs/adding_an_e2e_test.md
+++ b/docs/adding_an_e2e_test.md
@@ -1,0 +1,169 @@
+# Adding an E2E Test
+
+## Overview
+
+Adding support for a Torch operator in Torch-MLIR should always be accompanied
+by at least one end-to-end test to make sure the implementation of the op
+matches the behavior of PyTorch. The tests live in the
+`torch-mlir/python/torch_mlir_e2e_test/test_suite/` directory. When adding a new
+test, choose a file that best matches the op you're testing, and if there is no
+file that best matches add a new file for your op.
+
+## An E2E Test Deconstructed
+
+In order to understand how to create an end-to-end test for your op, let's break
+down an existing test to see what the different parts mean:
+
+```python
+class IndexTensorModule3dInput(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+        ([-1, -1], torch.int64, True),
+    ])
+    def forward(self, x, index):
+        return torch.ops.aten.index(x, (index,))
+
+
+@register_test_case(module_factory=lambda: IndexTensorModule3dInput())
+def IndexTensorModule3dInput_basic(module, tu: TestUtils):
+    module.forward(tu.rand(5, 4, 3), tu.randint(2, 3, high=3))
+```
+
+### Class Name
+
+
+```python
+class IndexTensorModule3dInput(torch.nn.Module):
+```
+
+The class name should always contain the name of the op that is being
+tested. This makes it easy to search for tests for a particular op. Often times
+an op will require multiple tests to make sure different paths in the
+compilation work as expected. In such cases, it is customary to add extra
+information to the class name about what is being tested. In this example, the
+op is being tested with a rank-3 tensor as an input.
+
+### `__init__` Method
+
+```python
+    def __init__(self):
+        super().__init__()
+```
+
+In most tests, the `__init__` method simply calls the `__init__` method of the
+`torch.nn.Module` class. However, sometimes this method can be used to
+initialize parameters needed in the `forward` method. An example of such a case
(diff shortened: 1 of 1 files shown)


docs/adding_an_e2e_test.md
```diff
@@ -0,0 +1,106 @@
...
+The `@register_test_case` decorator is used to register the test case function. The `module_factory` argument should be a function that when called produces an instance of the test class. This function will be used to create the first argument passed to the test case function.
+
+### Test Case Function
+
+```python
+def IndexTensorModule3dInput_basic(module, tu: TestUtils):
+    module.forward(tu.rand(5, 4, 3), torch.randint(3, (2, 3)))
+```
+
+The convention adopted for the name of the test case function is to have the same name as the test class postfixed by `_basic`. The test function always takes an instance of the test class as the first argument and a [`TestUtils`](https://github.com/llvm/torch-mlir/blob/ba17a4d6c09b4bbb4ef21b1d8d4a93cb056be109/python/torch_mlir_e2e_test/torchscript/framework.py#L167) object as the second argument. The `TestUtils` has some methods that allow the creation of random tensors. However, there is currently no support for generating random integer tensors, so if that is needed use `torch.randint`.
```
ramiro050, line 95: > However, there is currently no support for generating random integer tensors, so if that is needed use `torch.randint`.

This seems like an odd inconsistency to have. @silvasean, should I add support for random ints to the `TestUtils`, or was there a reason this was not done in the past?

silvasean: We should add it. The intent is that TestUtils grows helpers for all the most common generation needs.

ramiro050: Sounds good, I'll push a PR

docs/adding_an_e2e_test.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. Shape of input tensor. Use `-1` for dynamic dimensions
+2. Dtype of the input tensor
+3. Boolean representing whether the input tensor [has value semantics](https://github.com/llvm/torch-mlir/blob/ba17a4d6c09b4bbb4ef21b1d8d4a93cb056be109/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/class_annotator.h#L54-L67). This will always be true for E2E tests, since the [Torch-MLIR backend contract](architecture.md#the-backend-contract) requires all tensors in the IR to eventually have value semantics.
+
+From the structure of the annotations for the arguments other than the `self` argument it is clear that only tensor arguments are supported. This means that if an op requires an input other than a tensor, it should be created in the method body or as a class parameter in the `__init__` method.
```
silvasean, line 69: maybe mention the use of `int(some_tensor)` to get actual ints for ops that need that?

docs/adding_an_e2e_test.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+The convention adopted for the name of the test case function is to have the same name as the test class postfixed by `_basic`. The test function always takes an instance of the test class as the first argument and a [`TestUtils`](https://github.com/llvm/torch-mlir/blob/ba17a4d6c09b4bbb4ef21b1d8d4a93cb056be109/python/torch_mlir_e2e_test/torchscript/framework.py#L167) object as the second argument. The `TestUtils` has some methods that allow the creation of random tensors. However, there is currently no support for generating random integer tensors, so if that is needed use `torch.randint`.
+
+
+## Things to Consider When Creating New Tests
+
+- Do you need negative numbers? If so, [`tu.rand`](https://github.com/llvm/torch-mlir/blob/2374098d71083bce0fd2b665e28ebb621de5779c/python/torch_mlir_e2e_test/torchscript/framework.py#L182) and [`torch.randint`](https://pytorch.org/docs/stable/generated/torch.randint.html) both allow you to specify a lower bound for random number generation
+- Do you need ints? Use `torch.randint` rather than `tu.rand(...).to(dtype=int)`
+- Make sure the annotation of the forward method matches the input types and ranks
```
silvasean, line 102: Can you file a bug for this? This should somehow fail if it is wrong.

docs/adding_an_e2e_test.md
```diff
@@ -0,0 +1,106 @@
+# Adding an E2E Test
```
silvasean, line 1: nit: can you use a markdown formatter to make this 80 columns?

silvasean (approved):
Great! Thanks!

docs/adding_an_e2e_test.md
ramiro050: Here is the issue: https://github.com/llvm/torch-mlir/issues/1294

docs/adding_an_e2e_test.md
ramiro050: Done!

docs/adding_an_e2e_test.md
ramiro050: Done!

PR #1273, opened 2022-08-24
doc: fix instructions for LLVM and MHLO updates

ashay:
Now that we depend on MLIR-HLO, we have a slightly more involved process
for updating the LLVM submodule.  This patch updates the instructions in
`development.md` with both the rationale for the process and the
individual steps of the process.


silvasean (approved):
Great!

PR #1276, opened 2022-08-24
Add TestUtils.randint + replace torch.randint with tu.randint

ramiro050:
This commit adds a method to `TestUtils` that generates random integer
tensors with a similar interface to the `TestUtils.rand`. This commit
also replaces with `tu.randint` all test inputs generated with
`torch.randint`.


vivekkhandelwal1 (approved):
LGTM

silvasean (approved):
Great!

PR #1280, opened 2022-08-25
Fix a bug about torch-refine-types pass when the dtype of output tensor is known

Vremold:
This PR is proposed to solve the problem below.

Recently, I tried to run `torch-refine-types` pass with the following IR and finally got an exception of core dumped. The IR is like:
```
%none = torch.constant.none
%1 = torch.aten.empty.memory_format %0, %none, %none, %none, %none, %none : !torch.list<int>, !torch.none, !torch.none, !torch.none, !torch.none, !torch.none -> !torch.vtensor<[],si64>
```
Here is the exception message:
```
torch-mlir-opt: /root/share/up/torch-mlir/lib/Dialect/Torch/Transforms/RefineTypes.cpp:1118: void (anonymous namespace)::TypeAnalysis::incorporateKnowledge(mlir::Value, const (anonymous namespace)::ValueKnowledge &): Assertion `updatedKnowledge.has_value() && "IR has contradictory type!"' failed.
...
 #9 0x00000000005bb8ee (anonymous namespace)::TypeAnalysis::visitOperation(mlir::Operation*, llvm::ArrayRef<mlir::dataflow::Lattice<(anonymous namespace)::ValueKnowledge> const*>, llvm::ArrayRef<mlir::dataflow::Lattice<(anonymous namespace)::ValueKnowledge>*>) RefineTypes.cpp:0:0
#10 0x0000000001a64afc mlir::dataflow::AbstractSparseDataFlowAnalysis::visitOperation(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a64afc)
#11 0x0000000001a64569 mlir::dataflow::AbstractSparseDataFlowAnalysis::initializeRecursively(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a64569)
#12 0x0000000001a64675 mlir::dataflow::AbstractSparseDataFlowAnalysis::initializeRecursively(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a64675)
#13 0x0000000001a644c8 mlir::dataflow::AbstractSparseDataFlowAnalysis::initialize(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a644c8)
#14 0x0000000001a54f52 mlir::DataFlowSolver::initializeAndRun(mlir::Operation*) (/root/share/up/torch-mlir/build/bin/torch-mlir-opt+0x1a54f52)
#15 0x00000000005baa26 (anonymous namespace)::RefineTypesPass::runOnOperation() RefineTypes.cpp:0:0
...
[1]    1836385 abort (core dumped)  torch-mlir-opt -pass-pipeline='torchscript-module-to-torch-backend-pipeline'
```
This is because in `torch-refine-types` pass, we always invoke function `visitConstantTensorAllocOp` with a none-valued dataType (see these [lines of code](https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L984-L985)), which will be subsequently set `Torch::FloatType` by default. It works well when the dtype of output tensor is unk. But when the output dtype is not unk (like si64 in the above example), dtype mismatch happens and will finally cause a compilation failure like above.
This PR add the dtype check before invoking function `visitConstantTensorAllocOp`, thus avoid the problem of dtype mismatch.

### DIFF
--- a/lib/Dialect/Torch/Transforms/RefineTypes.cpp
+++ b/lib/Dialect/Torch/Transforms/RefineTypes.cpp
@@ -659,8 +659,7 @@ void TypeAnalysis::visitOperation(Operation *op,
           AtenIndexPutOp, ValsemVariantAtenCopyOp, AtenZeroOp,
           AtenIndexPutHackedTwinOp, AtenMaskedFillScalarOp, AtenFlipOp,
           PrimAbsScalarOp, AtenNumpyTOp, AtenTriuOp, AtenMaskedFillTensorOp,
-          AtenRollOp>(
-          op)) {
+          AtenRollOp>(op)) {
     return incorporateKnowledge(op->getResult(0), operands[0]->getValue());
   }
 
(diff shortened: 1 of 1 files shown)


silvasean:
This is not a correct fix. The output dtype should be uniquely determined by the inputs. There should never be a case where we infer "contradictory" types. We need to understand why the dtype was inferred to si64 before but to float after. In this case it seems that whatever inferred si64 originally must have been buggy, assuming that visitConstantTensorAllocOp is correct.

Vremold:
> This is not a correct fix. The output dtype should be uniquely determined by the inputs. There should never be a case where we infer "contradictory" types. We need to understand why the dtype was inferred to si64 before but to float after. In this case it seems that whatever inferred si64 originally must have been buggy, assuming that visitConstantTensorAllocOp is correct.

In fact, torch.aten.empty.memory_format op comes from decompose-complex-ops pass and si64 is set a that time. This is the bug reported in a [comment](https://github.com/llvm/torch-mlir/pull/1261#issuecomment-1225462922) of my last PR. But this PR introduces another bug in python e2e tests, so I turned it to be draft PR.

ramiro050:
> This will finally result in a decompose-complex-op pass followed by a refine-types pass. And the buggy torch.aten.empty.memory_format op and si64 dtype are inserted by the former and finally cause an error in the latter.

I think the error you're getting is because this helper function is not correct

https://github.com/llvm/torch-mlir/blob/e153694c940c9358ee4ce8abb003c1623c134df4/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L132

The `dtype` argument passed to `AtenEmptyMemoryFormatOp` should match the element type of the `resultType`, but currently `dtype=None` is being passed instead.

silvasean:
@Vremold can you link to the PR that fixed this issue?

Vremold:
> @Vremold can you link to the PR that fixed this issue?

As far as I know, this issue is still not solved. But the solution is clear. We should pay attention to the alloc-like op in decompose-complex-op pass and make sure its dtype is properly set rather than simply setting the dtype to none. This PR is kind of outdated, so I closed it.

silvasean:
Okay, let's keep it open then since the problem is not solved.

Vremold:
Thanks @tanyokwok . This issue is solved in PR #1398.

PR #1282, opened 2022-08-25
build: bump PyTorch to 1.12 for base_lazy_backend

xndcn:
torchgen was added in PyTorch 1.12


antoniojkim (approved):

powderluv (commented):
could we also please update all references to find_package Torch to be 1.12 ? We are building with nightly anyway

xndcn:
@powderluv base_lazy_backend seems optonal. If it is not enabled, maybe 1.12 is not required ?

xndcn:
I will close it because I found that although PyTorch 1.12 have torchgen, it does not support "non_native". It seems that base_lazy_backend relies on more recent 1.13 nightly build PyTorch, which is still not stable.

Issue #1271, opened 2022-08-23
ImportError: lib_torch_mlir_ltc.so: cannot open shared object file: No such file or directory


silvasean:
Installing from the nightly packages, I see a missing lib_torch_mlir_ltc.so

Is this intended to work? If not, is there some way to detect that it won't work so that we can avoid this kind of error

```
File "/opt/hostedtoolcache/Python/3.9.[13](https://github.com/iree-org/iree-torch/runs/7970725558?check_suite_focus=true#step:5:14)/x64/lib/python3.9/site-packages/torch_mlir_e2e_test/torchscript/configs/lazy_tensor_core.py", line 6, in <module>
    import torch_mlir.reference_lazy_backend._REFERENCE_LAZY_BACKEND as lazy_backend
ImportError: lib_torch_mlir_ltc.so: cannot open shared object file: No such file or directory
```
https://github.com/iree-org/iree-torch/runs/7970725558?check_suite_focus=true

cc @henrytwo @powderluv

henrytwo:
Hmm, `lib_torch_mlir_ltc.so` should exist under `torch_mlir/base_lazy_backend` in the wheel. Can you confirm that this path is available at runtime + can be used for shared libs?

henrytwo:
I can repro this locally, so I'll investigate further locally

henrytwo:
Hm it looks like setting:
```
export LD_LIBRARY_PATH=venv/lib/python3.8/site-packages/torch_mlir/base_lazy_backend/
```
solves the issue. Is there anyway we can automate this?

silvasean:
We put various shared libraries in `cmake_build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/_mlir_libs` -- I'm not sure how that plays into the bigger picture, but you can probably imitate the incantation we use to import those shared libraries. (libTorchMLIRAggregateCAPI.so in that directory is an example of a non-Python-extension shared library, if that is useful)

PR #1283, opened 2022-08-25
Fix LTC lib_torch_mlir_ltc.so import error

henrytwo:
Addresses https://github.com/llvm/torch-mlir/issues/1271, which will allow LTC to work from the prebuild wheels without import issues.

Previously, we were missing some cmake macros to include `$ORIGIN` to the list of directories searched for shared libraries.

As part of this, the reference LTC backend + base LTC backend were moved to `_mlir_libs` to be consistent with our other libraries (and also because the both of the libs need to be in the same directory for linking).

cc: @ke1337 @antoniojkim

### DIFF
--- a/python/torch_mlir/csrc/reference_lazy_backend/CMakeLists.txt
+++ b/python/torch_mlir/csrc/reference_lazy_backend/CMakeLists.txt
@@ -26,7 +26,7 @@ mlir_configure_python_dev_packages()
 # Library definition
 ###########################################################################
 
-set(LIBRARY_OUTPUT_PATH  "${TORCH_MLIR_PYTHON_PACKAGES_DIR}/torch_mlir/torch_mlir/reference_lazy_backend")
+set(LIBRARY_OUTPUT_PATH  "${TORCH_MLIR_PYTHON_PACKAGES_DIR}/torch_mlir/torch_mlir/_mlir_libs")
 set(OUTPUT_NAME "_REFERENCE_LAZY_BACKEND")
 
 if(TORCH_MLIR_ENABLE_LTC)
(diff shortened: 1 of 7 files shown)


henrytwo:
@silvasean In the future should we add some test that uses the built wheel to ensure we don't hit a similar issue in the future?

It looks like this issue stemmed from the fact that when the lib was used, it wasn't in the same location where it was originally built, which is it why it failed to find the shared library. Piping the binary thru `strings`, I could see some hardcoded paths.

With this PR, `$ORIGIN` was added to the list of paths searched for shared libs, which should solve the problem.

python/torch_mlir/csrc/reference_lazy_backend/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ mlir_configure_python_dev_packages()
 # Library definition
 ###########################################################################
 
-set(LIBRARY_OUTPUT_PATH  "${TORCH_MLIR_PYTHON_PACKAGES_DIR}/torch_mlir/torch_mlir/reference_lazy_backend")
+set(LIBRARY_OUTPUT_PATH  "${TORCH_MLIR_PYTHON_PACKAGES_DIR}/torch_mlir/torch_mlir/_mlir_libs")
```
silvasean, line 29: Maybe add a comment here "We piggyback on the shared library setup/infra used by Torch-MLIR Python bindings for consistency."

silvasean (approved):
Great, thanks!

silvasean:
> @silvasean In the future should we add some test that uses the built wheel to ensure we don't hit a similar issue in the future?

I think we attempt to do that here:

https://github.com/llvm/torch-mlir/blob/a1ace0657d1027a82b09a652a157a4c45e9d62d1/.github/workflows/buildRelease.yml#L25

Maybe we can extend the "smoke test" that is done there to do a basic LTC test?

PR #1285, opened 2022-08-25
Avoid cascading failures when compiler crashes

silvasean:
Change logic so that we never run the multiprocessing codepath with only
1 worker. That configuration was causing all subsequent tests to
spuriously fail if one test failed with a crash (this was easy to see
after sorting the tests). That configuration was the one used by the CI.

Also, sort tests to make output nicer.
Also, make verbose mode more verbose so that it is easy to see in `-s`
mode which test is crashing.


ramiro050 (approved):
Good catch! This will be very helpful

PR #1289, opened 2022-08-26
Fix symint related functionalization ops

antoniojkim:
Upstream PyTorch made some changes to a couple functionalization ops which were causing issues

CC: @ke1337


ramiro050 (approved):
LGTM!

xndcn:
it seems that upsteam PyTorch reverted "SymInt" commit... 
https://github.com/pytorch/pytorch/commit/c7edcd69683f6e3b08305ed0d4621e148fbfbe17

PR #1295, opened 2022-08-27
Remove "torchscript" association from the e2e framework.

silvasean:
Remove "torchscript" association from the e2e framework.

We use it for more than TorchScript testing now. This is a purely
mechanical change to adjust some file paths to remove "torchscript".

The most perceptible change here is that now e2e tests are run with

```
./tools/e2e_test.sh
instead of:
./tools/torchscript_e2e_test.sh
```


ramiro050 (approved):
LGTM!

sjain-stanford (commented):
LGTM (apologies this got buried in my GH notify emails -> will make it a point to pay closer attention to this next time!)

silvasean:
@sjain-stanford FYI there is a chrome extension that I use which notifies on GH stuff: https://chrome.google.com/webstore/detail/notifier-for-github/lmjdlojahmbbcodnpecnjnmlddbkjhnn

PR #1286, opened 2022-08-25
Add CI for generated files.

silvasean:
This ensures that they are always up to date.

This also updates the shape lib to make the new CI actually pass :)

PR #1306, opened 2022-08-29
[MLIR][MHLO] Add AtenFrobeniusNormDimOp and add its conversion pattern to MHLO and linalg

Vremold:
This PR registers a new op, AtenFrobeniusNormDimOp, to the ods and supports converting it to MHLO dialect.


lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
```diff
@@ -3497,60 +3497,65 @@ module {
...
-    %6 = torch.aten.__getitem__.t %arg1, %int0 : !torch.list<int>, !torch.int -> !torch.int
-    %7 = torch.aten.append.t %3, %6 : !torch.list<int>, !torch.int -> !torch.list<int>
-    %8 = torch.aten.__range_length %int2, %2, %int1 : !torch.int, !torch.int, !torch.int -> !torch.int
-    torch.prim.Loop %8, %true, init() {
+    %4 = torch.prim.If %arg6 -> (!torch.int) {
+      torch.prim.If.yield %int1 : !torch.int
+    } else {
+      torch.prim.If.yield %int0 : !torch.int
+    }
+    %5 = torch.aten.__getitem__.t %arg0, %int0 : !torch.list<int>, !torch.int -> !torch.int
```
ZihengJiang, line 3511: why make this change?

lib/Conversion/TorchToMhlo/Reduction.cpp
```diff
@@ -556,6 +569,106 @@ LogicalResult ConvertAtenReductionOp<AtenSumDimIntListOp>::matchAndRewrite(
...
+    return op.emitError(
+        "only float dtype allowed in input tensor of AtenFrobeniusNormDimOp");
+  }
+
+  SmallVector<int64_t> dims;
+  if (!matchPattern(op.dim(), m_TorchConstantIntList(dims))) {
+    return rewriter.notifyMatchFailure(
+        op, "non-const integer `dim` is not supported");
+  }
+  for (auto &dim : dims) {
```
fortianyou, line 598: Nit: How about sorting the dimensions in ascending order so that the conversion can be stable with unordered dims inputs?

test/Conversion/TorchToMhlo/reduction.mlir
```diff
@@ -241,3 +241,64 @@ func.func @torch.aten.max(%arg0: !torch.vtensor<[?,?,?],f32>) -> !torch.vtensor<
   %0 = torch.aten.max %arg0 : !torch.vtensor<[?,?,?],f32> -> !torch.vtensor<[],f32>
   return %0 : !torch.vtensor<[],f32>
 }
+
```
fortianyou, line 244: Can we also add an e2e test for ForbeniusNorm?

lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
Vremold: This is auto changed by running update_shape_lib.sh

test/Conversion/TorchToMhlo/reduction.mlir
silvasean: +1, new ops should always include e2e tests: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#testing

> DO use an end-to-end test if you are implementing a new op or extending the support for an existing op.

lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
ramiro050: Can you rebase to top of main? These changes were already added to torch-mlir in another PR: https://github.com/llvm/torch-mlir/pull/1286

test/Conversion/TorchToMhlo/reduction.mlir
Vremold: aten.frobenius_norm_dim comes from tracing PyTorch function torch.norm(data, p="fro"). Do you have any idea to use tracing mode in e2e tests, without introducing major changes to the current code structure.

fortianyou: Will `torch.norm` decompose into `aten.frobenius_norm_dim` directly through scripting?

Vremold: > Will `torch.norm` decompose into `aten.frobenius_norm_dim` directly through scripting?

There is no decomposition rule for `torch.norm`-like op. If we script 'torch.norm' function, many unnecessary nested regions and unhandled ops will be introduced.

fortianyou: I think it's fine to add tracing mode to e2e tests. And the compile has arguments for it, see: https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/__init__.py#L138

Vremold: > I think it's fine to add tracing mode to e2e tests. And the compile has arguments for it, see: https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/__init__.py#L138

I know it. But it seems for now, our e2e test backend configs all compile test cases by script mode. The simplest solution is that we change the compile function in e2e test backends to use trace mode for all test cases. I just want to figure out if there is a way to run one test case in trace mode but not modify the backend configs much.

ramiro050: You should be able to do `torch.ops.aten.frobenius_norm(input_tensor, dim=..., keepdim=...)`. See: https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#forward-method

lib/Conversion/TorchToMhlo/Reduction.cpp
```diff
@@ -336,7 +343,8 @@ LogicalResult ConvertAtenReductionOp<AtenSumOp>::matchAndRewrite(
   Value input = adaptor.self();
   auto inputTy = input.getType().dyn_cast<RankedTensorType>();
   if (!inputTy) {
-    return rewriter.notifyMatchFailure(op, "only Tensor types supported in MHLO");
+    return rewriter.notifyMatchFailure(op,
```
silvasean, line 346: can you submit this reformatting as a separate PR? It is hard to see the actual change here.

test/Conversion/TorchToMhlo/reduction.mlir
```diff
@@ -241,3 +241,64 @@ func.func @torch.aten.max(%arg0: !torch.vtensor<[?,?,?],f32>) -> !torch.vtensor<
...
+// -----
+// CHECK-LABEL:   func.func @torch.aten.frobenius_norm.dim(
```
silvasean, line 246: I would recommend not adding any .mlir tests. I don't think they add much in this case. Also see our testing philosophy here:

lib/Conversion/TorchToMhlo/Reduction.cpp
Vremold: Sorry for that. I will revert the modification of reformatting.

lib/Conversion/TorchToMhlo/Reduction.cpp
Vremold: Done.

test/Conversion/TorchToMhlo/reduction.mlir
Vremold: Thanks @ramiro050 . Two e2e test cases for AtenFrobeniusNormDim op are added.

test/Conversion/TorchToMhlo/reduction.mlir
Vremold: The mlir is for testing conversion for new op from torch to mhlo. If you don't like it, I will withdraw the changes.

Vremold:
I've add e2e tests for AtenFrobeniusNormDim op, see `python/torch_mlir_e2e_test/test_suite/reduction.py`. For better e2e test coverage, I also added its conversion pattern to Linalg.

fortianyou (approved):
Thanks @Vremold. LGTM

ramiro050:
> I also added its conversion pattern to Linalg.

Thanks for this!

test/Conversion/TorchToMhlo/reduction.mlir
silvasean: We should not submit this until ReduceFrobeniusNormModule_basic and ReduceFrobeniusNormKeepDimModule_basic tests pass. Once those tests pass, I think this .mlir test is irrelevant.

Why do the new E2E tests fail e2e with TorchToMhlo?

silvasean (changes requested):

silvasean:
Thanks for adding TorchToLinalg support for this op. This is very kind, but in the future, this is not necessary -- we can add the tests to REFBACKEND_XFAIL_SET.

test/Conversion/TorchToMhlo/reduction.mlir
Vremold: > Why do the new E2E tests fail e2e with TorchToMhlo?

That's because `hlo-legalize-to-linalg` pass will invoke element-wise ops' (like mhlo.sqrt) reifyReturnTypeShapes function in case that the result type has dynamic shape. See codes [here](https://github.com/tensorflow/mlir-hlo/blob/master/lib/Dialect/mhlo/transforms/legalize_to_linalg_utils.cc#L76-L78). This will introduce ShapeOfOp, which will not be converted or eliminated in the following passes and finally cause an error.

Vremold:
I've made a simple modification to the test cases to avoid dynamic shaped output for mhlo.reduce op. Now, new E2E tests should succeed with TorchToMhlo.

silvasean:
> I've made a simple modification to the test cases to avoid dynamic shaped output for mhlo.reduce op. Now, new E2E tests should succeed with TorchToMhlo.

We try to have all tests be fully dynamic shaped. Do we know why the mhlo.reduce op does not support dynamic shaped output? Is there a bug filed for that at least?

Vremold:
We can solve this problem by adding a `canonicalizer` pass after `hlo-legalize-to-linalg` pass. The changes should happens  [here](https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir_e2e_test/mhlo_backends/linalg_on_tensors.py#L39). In my local environment, it should pass ReduceFrobeniusNormModule_basic after changes. But for ReduceFrobeniusNormKeepDimModule_basic test, we probably also need a little change in current conversion from Torch to MHLO. In detail, we should use `mhlo.reshape` instead of `mhlo.dynamic_reshape` to reshape [?] to [1, 1, ?], as current `hlo-legalize-to-linalg` pass doesn't support `mhlo.dynamic_reshape`. Will it be okay? If so, I will make another commit to implement this all.

Vremold:
We can solve this problem by adding a `canonicalizer` pass after `hlo-legalize-to-linalg` pass. I have made the changes. But for ReduceFrobeniusNormKeepDimModule_basic,  I add it to XFail_sets, because current `hlo-legalize-to-linalg` pass doesn't support `mhlo.dynamic_reshape` which is needed in dim keeping.

sherhut:
> I think we need to fold tensor.extract(shape_of(x)) -> tensor.dim(x). @sherhut @burmako any ideas here do we expect that fold to already happen? See description from @Vremold :

Yes, this [folding already exist](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Shape/IR/ShapeCanonicalization.td#L48). It is required to bridge from the unranked shape representation which uses shape vectors to a ranked shape representation that uses scalars. We rely on it in quite a few places.

silvasean:
> We can solve this problem by adding a `canonicalizer` pass after `hlo-legalize-to-linalg` pass. I have made the changes. But for ReduceFrobeniusNormKeepDimModule_basic, I add it to XFail_sets, because current `hlo-legalize-to-linalg` pass doesn't support `mhlo.dynamic_reshape` which is needed in dim keeping.

@burmako @sherhut any idea how to deal with dynamic_reshape? It seems like somebody must already be using this and lowering it to linalg.

(sorry @Vremold for holding up your PR, but it is a problem if MHLO cannot support end-to-end flows through linalg and I want the MHLO owners/stakeholders to be involved here so it can be fixed)

sherhut:
> > We can solve this problem by adding a `canonicalizer` pass after `hlo-legalize-to-linalg` pass. I have made the changes. But for ReduceFrobeniusNormKeepDimModule_basic, I add it to XFail_sets, because current `hlo-legalize-to-linalg` pass doesn't support `mhlo.dynamic_reshape` which is needed in dim keeping.
> 
> @burmako @sherhut any idea how to deal with dynamic_reshape? It seems like somebody must already be using this and lowering it to linalg.

`dynamic_reshape` in its full power cannot be lowered to 'linalg' because linalg is not expressive enough. We have two solutions that we are using.

1. We have a [symbolic shape optimization](https://cs.opensource.google/tensorflow/tensorflow/+/master:tensorflow/compiler/xla/mlir_hlo/lib/Transforms/symbolic_shape_optimization.cc;l=711;drc=60cce54a98a5687fb535c79cd01c33a95e1b52cd) pass that transforms dynamic reshape into `tensor.collapse_shape` and `tensor.expand_shape` when possible. For the use case here, this might be what is needed. I only briefly looked at the code.

2. You can lower it directly to the `memref` dialect during bufferization.

silvasean:
Thanks Stefan! @Vremold can you try adding SymbolicShapeOptimizationPass to the lowering pipeline? I think that is exactly what is needed.

Vremold:
Thanks @sherhut and @silvasean . This suggestion solved my problem. I've added a `symbolic-shape-optimization` before `hlo-legalize-to-linalg` in the mhlo e2e backend. Many other mhlo e2e tests also passes with this change. All of them are updated in the new `xfails_sets.py`

silvasean (approved):
Awesome!

silvasean:
Thanks for bearing with us as we had this long conversation on MHLO!

Issue #1231, opened 2022-08-16
Investigate failure of `ElementwiseReluModule` and `ResNet18StaticModule` TOSA tests


ashay:
Both `ElementwiseReluModule` and `ResNet18StaticModule` TOSA tests fail after updating LLVM.  See [CI log](https://github.com/llvm/torch-mlir/runs/7849271902?check_suite_focus=true#step:12:680) for details.

```
    FAIL - "ElementwiseReluModule_basic"
        @ trace item #0 - call to "forward"
        @ output of call to "forward"
        ERROR: value (Tensor with shape=[4, 2], dtype=torch.float32, min=+3.403e+38, max=+3.403e+38, mean=+3.403e+38) is not close to golden value (Tensor with shape=[4, 2], dtype=torch.float32, min=+0.0, max=+0.3964, mean=+0.09984)

    FAIL - "ResNet18StaticModule_basic"
        @ trace item #0 - call to "forward"
        @ output of call to "forward"
        ERROR: value (Tensor with shape=[1, 1000], dtype=torch.float32, min=+nan, max=+nan, mean=+nan) is not close to golden value (Tensor with shape=[1, 1000], dtype=torch.float32, min=-1.815, max=+2.328, mean=+0.03187)
```

sjarus:
Without looking at anything in detail, this sounds like something might have changed in TosaToLinAlg/LinAlgNamed/Arith that generates functionally wrong output, since the TOSA content is unchanged here.

silvasean:
@sjarus do you have someone on your side that can investigate this?

sjarus:
Adding @eric-k256 and @u99127, as well as @jpienaar who has been reviewing recent TosaToXXX PRs . I think this might just be a git bisect away from catching which TosaToXX change broke the functional correctness of these models.

jpienaar:
So this failed after the last LLVM update? Could you add the before and after rev here? (ReluN was removed 4 days ago upstream, and https://github.com/llvm/llvm-project/commit/444b4fda17bca581fab084926b91393bd20478e7 seems likely candidate fix given replacement)

eric-k256:
I can confirm that https://github.com/llvm/llvm-project/commit/444b4fda17bca581fab084926b91393bd20478e7 fixes the problem.
I'm not sure what you do in torch-mlir. Does this wait until the next LLVM update, or should we try to move LLVM forward again?

silvasean:
[...]

PR #1307, opened 2022-08-29
build: update llvm tag to 00d648bd

ashay:
- Update MHLO commit to build with LLVM commit hash 00d648bd
 - Update TorchToMhlo code to work with Stablehlo
 - Re-enabled two failing TOSA tests, thus resolving Github Issue #1231


fortianyou:
@ashay I believe this is a bug of MHLO legalization to LinAlg. The error raised here [legalize_to_linalg.cc#L2031]( https://github.com/tensorflow/mlir-hlo/blob/master/lib/Dialect/mhlo/transforms/legalize_to_linalg.cc#L2031), because `reversals` has no value.

silvasean (approved):
Thanks!

PR #1308, opened 2022-08-29
Force double precision in aten.mean.dim

qedawkins:
Accuracy for aten.mean.dim was off for large inputs for the same reason as aten.var.correction (see #1100)

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -1214,9 +1214,18 @@ public:
           op, "expected `dim` to be `None` or constructed from list construct");
     }
 
+    // Upcasting the input tensor to `F64` dtype for higher precision during the
+    // computation of the result.
+    BaseTensorType outputTensorType = outputType.cast<BaseTensorType>();
+    Type outputTypeAsF64 = outputTensorType.getWithSizesAndDtype(
+              outputTensorType.getSizes(), rewriter.getF64Type());;
+    if (inputType.getDtype().getIntOrFloatBitWidth() != 64) {
+      input = convertTensorToDtype(rewriter, loc, input, rewriter.getF64Type());
+    }
+
     // Compute sum along dimensions specified in `dimList`.
     Value sumAlongDims = rewriter.create<AtenSumDimIntListOp>(
-        loc, outputType, input, dimList, keepDim, dtype);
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1214,9 +1214,18 @@ class DecomposeAtenMeanDimOp : public OpRewritePattern<AtenMeanDimOp> {
           op, "expected `dim` to be `None` or constructed from list construct");
     }
 
+    // Upcasting the input tensor to `F64` dtype for higher precision during the
+    // computation of the result.
+    BaseTensorType outputTensorType = outputType.cast<BaseTensorType>();
+    Type newOutputType = outputTensorType.getWithSizesAndDtype(
```
ramiro050, line 1220: drive-by comment: can you call this something like `outputTypeAsF64`? `newOutputType` is a bit confusing since the output type of the op being decomposed is not actually changing

qedawkins: Thanks for the comment! I'm still working on fixing some CI stuff but after looking at the pytorch code (https://github.com/pytorch/pytorch/blob/9e40be3b789cf9cbad58f7de87ac41a157bf4431/aten/src/ATen/native/ReduceOps.cpp#L1542) I think there are probably other ops that are doing this explicit promotion so a more comprehensive set of tests for these large reductions might be warranted.

Also it would probably be worth updating the naming in the calculateVariance helper as well but that might be outside the scope of this PR.

ramiro050: Thanks for digging into it! Yeah, we might eventually need to add double precision support for all reduction ops. And yeah, I would fix `calculateVariance` in a future PR

qedawkins: Thanks for the comment! In that case it might be good to update that in the calculateVariance helper as well but that would probably be outside the scope of this PR.

Also after looking in the pytorch code there are probably other reduction ops that are doing this so a more

powderluv:
Looks like Refbackend fails to run the test on GHA ? Should we skip it by default ?

ramiro050:
Is the error not reproducible locally?

qedawkins:
Closing as this is fixed by #1605

PR #1309, opened 2022-08-29
Implement Expand/Collapse Functionality for Aten.View

JakopinA:
(Currently Unfinished, will also be squashing commits later once the commit is in better shape)

Currently working on making a fix so that a view with dimensions [3,2] can be displayed as [2,3].

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -22,6 +22,14 @@ EAGER_MODE_XFAIL_SET = {
 }
 
 MHLO_PASS_SET = {
+    "ViewDoubleMergeStaticModule_basic",
+    "ViewCollapseOnesMiddleModule_basic",
+    "ViewFiveTestStaticModule_basic",
+    "ViewOffsetTestStaticModule_basic",
+    "ViewTwoFiveThreeStaticModule_basic",
+    "ViewTwoToThreeStaticModule_basic",
+    "ViewExpandOnesMiddleOppModule_basic",
+    "ViewOffsetBackwardTestStaticModule_basic",
     "FlattenStaticModule_basic",
     "FlattenRank0Module_basic",
     "TensorsConcatNegativeDimModule_basic",
@@ -159,6 +167,14 @@ MHLO_PASS_SET = {
 # Write the TOSA set as a "passing" set as it is very early in development
 # and very few tests work yet.
 TOSA_PASS_SET = {
+    "ViewDoubleMergeStaticModule_basic",
+    "ViewCollapseOnesMiddleModule_basic",
+    "ViewFiveTestStaticModule_basic",
+    "ViewOffsetTestStaticModule_basic",
+    "ViewTwoFiveThreeStaticModule_basic",
+    "ViewTwoToThreeStaticModule_basic",
+    "ViewExpandOnesMiddleOppModule_basic",
+    "ViewOffsetBackwardTestStaticModule_basic",
     "ElementwiseUnaryModule_basic",
     "ElementwiseBinaryModule_basic",
     "ElementwiseSigmoidModule_basic",
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -237,26 +237,43 @@ public:
                              int64_t startExpandDim, int64_t maxExpandDim,
                              const SmallVector<int64_t> &collapseShape,
                              const SmallVector<int64_t> &expandShape,
+                             ReassociationIndices &collapseIndices,
                              ReassociationIndices &expandIndices) {
     int64_t collapseDimSize = collapseShape[collapseDim];
+
     int64_t expandedSize = 1;
+    int64_t collapsedSize = collapseDimSize;
 
-    for (auto i : llvm::seq<int64_t>(startExpandDim, maxExpandDim)) {
-      int64_t expandDimSize = expandShape[i];
-      if (expandDimSize == kUnknownSize ||
-          collapseDimSize % (expandedSize *= expandDimSize)) {
-        return rewriter.notifyMatchFailure(
-            op, "desired size is not compatible with the input tensor size");
+    int64_t expandIndex = startExpandDim;
+    int64_t collapseIndex = collapseDim + 1;
+
+    while (expandIndex != maxExpandDim || collapseIndex != maxCollapseDim) {
+
+      if (expandIndex != maxExpandDim && expandedSize <= collapsedSize) {
+        int64_t expandDimSize = expandShape[expandIndex];
+        if (expandDimSize == kUnknownSize)
+          return rewriter.notifyMatchFailure(
+              op, "desired expand size is not compatible with the input tensor "
+                  "size");
+
+        expandedSize *= expandDimSize;
+        expandIndices.push_back(expandIndex);
+        expandIndex++;
+      } else if (collapseIndex != maxCollapseDim &&
+                 collapsedSize < expandedSize) {
+        collapseDimSize = collapseShape[collapseIndex];
+        if (collapseDimSize == kUnknownSize)
+          return rewriter.notifyMatchFailure(
+              op, "desired collapse size is not compatible with the input "
+                  "tensor size");
+
+        collapsedSize *= collapseDimSize;
+        collapseIndices.push_back(collapseIndex);
+        collapseIndex++;
       }
-      expandIndices.push_back(i);
-      if (expandedSize == collapseDimSize)
+
+      if (expandedSize == collapsedSize)
         return success();
-
-      if (expandedSize > collapseDimSize) {
-        return rewriter.notifyMatchFailure(
-            op, "unimplemented: only supports expanding and collapsing "
-                "in view");
-      }
     }
 
     return rewriter.notifyMatchFailure(
@@ -372,7 +389,6 @@ public:
             "is enough static shape information to determine its size, or when "
             "the input tensor is being flattened to a single dimension");
       }
-
       auto productReduceKnownSizes = [](const ArrayRef<int64_t> sizes) {
         auto knownSizes = llvm::make_filter_range(
             sizes, [](int64_t val) { return val != kUnknownSize; });
@@ -475,6 +491,7 @@ public:
                   nextUnchangedOutput, inputShapeVec, outputShape,
                   outputAssociations.back())))
             return failure();
+
           outputDim = nextUnchangedOutput;
           inputDim = nextUnchangedInput;
           continue;
@@ -500,11 +517,12 @@ public:
           if (failed(minimallyCollapseDimHelper(
                   op, rewriter, inputDim, nextUnchangedInput, outputDim,
                   nextUnchangedOutput, inputShapeVec, outputShape,
-                  outputAssociations.back())))
+                  inputAssociations.back(), outputAssociations.back()))) {
             return failure();
+          }
           hasDynamic = false;
           outputDim = outputAssociations.back().back() + 1;
-          inputDim++;
+          inputDim = inputAssociations.back().back() + 1;
           continue;
         }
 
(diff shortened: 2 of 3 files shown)


lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -237,6 +239,70 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+          return rewriter.notifyMatchFailure(
+              op, "desired expand size is not compatible with the input tensor size");
+      }
+
+      if (collapseIndex != maxCollapseDim)
+      {
+        collapseDimSize = collapseShape[collapseIndex];
+        if (collapseDimSize == kUnknownSize)
+          return rewriter.notifyMatchFailure(
+              op, "desired collapse size is not compatible with the input tensor size");
```
ramiro050, line 269: The test `ViewDynamicExpandCollapseModule_basic` fails here. You can see this error message if you run the recreation script given by the test failure error message:

```
torch-mlir-opt -pass-pipeline='torch-backend-to-linalg-on-tensors-backend-pipeline' /tmp/ViewDynamicExpandCollapseModule.mlir
```

adding the `-debug` flag after the command. The last part of the output will look like this:

```
Legalizing operation : 'torch.aten.view'(0x556d8403ab10) {
  %5 = "torch.aten.view"(%arg0, %4) : (!torch.vtensor<[?,4,?,?],f32>, !torch.list<int>) -> !torch.vtensor<[2,1,4,64],f32>

  * Fold {
  } -> FAILURE : unable to fold

  * Pattern : 'torch.aten.view -> ()' {
Trying to match "(anonymous namespace)::ConvertAtenViewOp"
    ** Insert  : 'torch_c.to_i64'(0x556d84078790)
    ** Insert  : 'torch_c.to_i64'(0x556d84078820)
    ** Insert  : 'torch_c.to_i64'(0x556d84078930)
    ** Insert  : 'torch_c.to_i64'(0x556d8407db40)
ImplicitTypeIDRegistry::lookupOrInsert(mlir::InferIntRangeInterface::Trait<Empty>)
    ** Insert  : 'arith.constant'(0x556d840789c0)
    ** Insert  : 'tensor.dim'(0x556d84082e00)
    ** Insert  : 'arith.constant'(0x556d840843a0)
    ** Insert  : 'arith.constant'(0x556d840844b0)
    ** Insert  : 'arith.constant'(0x556d84085a00)
    ** Insert  : 'tensor.dim'(0x556d84084400)
    ** Insert  : 'arith.constant'(0x556d84085a60)
    ** Insert  : 'tensor.dim'(0x556d84085ac0)
    ** Insert  : 'arith.index_cast'(0x556d84085b70)
    ** Insert  : 'arith.cmpi'(0x556d840870f0)
    ** Insert  : 'cf.assert'(0x556d84087270)
    ** Failure : desired collapse size is not compatible with the input tensor size
Input Dim Failure
"(anonymous namespace)::ConvertAtenViewOp" result 0
  } -> FAILURE : pattern failed to match

  * Pattern : 'torch.aten.view -> ()' {
Trying to match "(anonymous namespace)::ConvertAtenScalarToTensorLike"
    ** Failure : not a supported Scalar to Tensor like op
"(anonymous namespace)::ConvertAtenScalarToTensorLike" result 0
  } -> FAILURE : pattern failed to match

  * Pattern : 'torch.aten.view -> ()' {
Trying to match "(anonymous namespace)::ConvertElementwiseOp"
    ** Failure : not a supported elementwise op
"(anonymous namespace)::ConvertElementwiseOp" result 0
  } -> FAILURE : pattern failed to match

  * Pattern : 'torch.aten.view -> ()' {
Trying to match "(anonymous namespace)::ConvertReductionOp"
    ** Failure : not a supported reduce op
"(anonymous namespace)::ConvertReductionOp" result 0
  } -> FAILURE : pattern failed to match
} -> FAILURE : no matched legalization pattern
//===-------------------------------------------===//
```

The message is `** Failure : desired collapse size is not compatible with the input tensor size` while attempting to apply `ConvertAtenViewOp`.

You can use the `-debug` flag to see where the other tests are failing. Happy to help with questions

JakopinA: How exactly did you figure out that the test failed at that line? I've been trying to figure that out for a while and I haven't gotten anywhere with it. With that said, I can see how the code would have failed there, as I had not properly added code considering that a dimension -1 could be passed in.

ramiro050: The failure message of the `ConvertAtenViewOp` pass (when passing the `-debug` flag to `torch-mlir-op`) is `** Failure : desired collapse size is not compatible with the input tensor size`. To find the line of code where the error happened, I just searched for the `notifyMatchFailure` with that same error message

JakopinA:
Bump: I've committed an update that now has every single view-related test pass, including [3,2]->[2,3]. Once I sort out the commits so they are squashed into one commit I will ask for more reviewers.

JakopinA:
Fixed the commit issue, should be ready for reviewing!

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -237,26 +237,45 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
                              int64_t startExpandDim, int64_t maxExpandDim,
                              const SmallVector<int64_t> &collapseShape,
                              const SmallVector<int64_t> &expandShape,
-                             ReassociationIndices &expandIndices) {
+                             ReassociationIndices &collapseIndices,
+                             ReassociationIndices &expandIndices
+                             ) {
+    int64_t expandDimSize = expandShape[startExpandDim];
```
qedawkins, line 243: nit: this variable is unnecessary since you start with expandedSize = 1;
you can just put the declaration down below where you use it.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -500,8 +519,9 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
           if (failed(minimallyCollapseDimHelper(
                   op, rewriter, inputDim, nextUnchangedInput, outputDim,
                   nextUnchangedOutput, inputShapeVec, outputShape,
-                  outputAssociations.back())))
+                  inputAssociations.back(), outputAssociations.back()))) {
             return failure();
+          }
           hasDynamic = false;
           outputDim = outputAssociations.back().back() + 1;
           inputDim++;
```
qedawkins, line 527: I think you need to be properly updating the inputDim here the same way the outputDim is incremented. Ditto for the other call to minimallyCollapseDimHelper.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -513,20 +533,15 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
         inputDim = inputAssociations.back().back() + 1;
         outputDim++;
         continue;
       }
 
-      if (inputDim != nextUnchangedInput || outputDim != nextUnchangedOutput) {
-        return rewriter.notifyMatchFailure(
-            op, "could not match input tensor shape to output shape; "
-                "potentially unsupported view shape");
-      }
```
qedawkins, line 528: I think this check is still important for making sure the view op is valid (think [2, 3] -> [2, 3, 4]). The only reason it says potentially unsupported is because match failures from dynamic dims overlap here and are unimplemented rather than invalid. It would probably be worth updating the error message though. This check should be fine after properly updating the inputDim and outputDim above.

qedawkins (commented):
Looks mostly good! I left a few comments on things that need addressing.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -33,6 +33,8 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
 
+#include <iostream>
```
ramiro050, line 36: Code in torch-mlir should never import `iostream`

https://llvm.org/docs/CodingStandards.html#include-iostream-is-forbidden

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -237,26 +239,43 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+    int64_t collapseIndex = collapseDim + 1;
 
-      if (expandedSize > collapseDimSize) {
-        return rewriter.notifyMatchFailure(
-            op, "unimplemented: only supports expanding and collapsing "
-                "in view");
+    while (expandIndex != maxExpandDim || collapseIndex != maxCollapseDim)
+    {
+
+      if ((expandIndex != maxExpandDim && expandedSize <= collapsedSize)) {
```
ramiro050, line 256: nit: extra parenthesis not needed

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -513,17 +533,19 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
-        outputDim++;
+        outputDim = outputAssociations.back().back() + 1;
         continue;
       }
 
       if (inputDim != nextUnchangedInput || outputDim != nextUnchangedOutput) {
         return rewriter.notifyMatchFailure(
-            op, "could not match input tensor shape to output shape; "
+            op, "could not match inputDim to next Unchanged Input or "
+                "outDim to next Unchanged Output"
```
ramiro050, line 548: nit: this string should end with a period and a space to ensure the right formatting when it gets concatenated with the next string

python/torch_mlir_e2e_test/test_suite/reshape_like.py
```diff
@@ -240,6 +259,158 @@ def ViewDynamicExpandCollapseWithAtenIntModule_basic(module, tu: TestUtils):
...
 # ==============================================================================
 
+class ViewTwoToThreeStaticModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([3,2], torch.float32, True),
```
ramiro050, line 269: nit: the style followed in this file is to have spaces after commas. There are other instances of no space after comma in the changes below.

See: https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#things-to-consider-when-creating-new-tests

> Always follow the style and conventions of the file you're adding a test in.

ramiro050 (changes requested):
Thanks for working on this! This op is very tricky. I have a few style comments below

lib/Conversion/TorchToLinalg/DataMovement.cpp
JakopinA: I believe I fixed this, but I'm not 100% sure. If you wouldn't mind looking at my change that would be very helpful.

e2e_testing/xfail_sets.py
```diff
@@ -159,6 +167,16 @@
...
+    "ViewDoubleMergeStaticModule_basic",
+    "ViewCollapseOnesMiddleModule_basic",
+    "ViewFiveTestStaticModule_basic",
+    "ViewOffsetTestStaticModule_basic",
+    "ViewTwoFiveThreeStaticModule_basic",
+    "ViewTwoToThreeStaticModule_basic",
+    "ViewExpandOnesMiddleOppModule_basic",
+    "ViewOffsetBackwardTestStaticModule_basic",
+    "ViewUnknown1TestStaticModule_basic",
+    "ViewUnknown2TestStaticModule_basic",
```
vivekkhandelwal1, line 179: You need to remove this since the tests are failing for the tosa config.

vivekkhandelwal1 (changes requested):

silvasean:
Can you please revert this? We should not be hardcoding for static shapes.

qedawkins:
> Can you please revert this? We should not be hardcoding for static shapes.

Can you elaborate on this? This looks like an extension of the code pattern that was already there and I thought handling for dynamic shapes will probably include this static shape code anyway, or at least something close to it. I may be misunderstanding what you're asking for though.

silvasean:
We should generally try to avoid doing anything that hardcodes for static shapes. Doing so gives us a false sense of accomplishment and just leaves a landmine for future-us to stumble on. It's not maintainable if we have an irregular patchwork of static-shape vs dynamic shape functional completeness issues. (for TOSA we had to bite the bullet and allow some static shape specializing since TOSA doesn't really support dynamic shapes, but we don't want to do that here because that is a very, very deep pit to dig oneself out of)

Also, this patch doesn't follow the testing guidelines here: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

In particular:

> DO use a unit test if your lowering for an op has multiple cases / logic. This also helps future maintainers of the lowering to see in one place all the different edge cases of the op that you had to handle. (these can be easily reduced out of all the end-to-end tests you added).

qedawkins:
> Also, this patch doesn't follow the testing guidelines here: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

I see this makes sense to me. I need to be more aware of the torch-mlir guidelines when reviewing.

> We should generally try to avoid doing anything that hardcodes for static shapes. Doing so gives us a false sense of accomplishment and just leaves a landmine for future-us to stumble on. It's not maintainable if we have an irregular patchwork of static-shape vs dynamic shape functional completeness issues.

I definitely agree with this, but my thinking for this patch was that handling for the dynamic cases here were somewhat distinct from the static one here but we can revert for now and come back to this with more comprehensive support.

I opened a PR for the revert here #1347. Thank you for bearing with me on this as I'm still acclimating to some of the project guidelines.

PR #1320, opened 2022-08-31
Add lowering for aten::cumsum

gpetters94:
Currently debugging this to try and figure out a segfault. I think I might have set up the Scan wrong, but I'm not sure.

### DIFF
--- a/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
+++ b/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
@@ -14,6 +14,8 @@
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/IR/Builders.h"
+#include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/MLIRContext.h"
 #include "torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorDialect.h"
 #include "torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorOps.h"
@@ -70,6 +72,29 @@ static Value createTMTensorScatterOp(
   Value originalElement = blockArgs[1];
   bodyBuild(regionBuilder, loc, updatesElement, originalElement);
   return scatterOp->getResult(0);
+}
+
+static Value createTMTensorScanOp(
+    OpBuilder &b, Location loc, Value input, Value output, Value accumulator,
+    int64_t dim, bool inclusive,
+    function_ref<void(OpBuilder &, Location, Value, Value)> bodyBuild) {
+  auto inputType = input.getType().cast<RankedTensorType>();
+  auto accType = accumulator.getType().cast<RankedTensorType>();
+  Type elementType = inputType.getElementType();
+  auto scanOp = b.create<TMTensor::ScanOp>(
+      loc, TypeRange{inputType, accType}, input,
+      ValueRange{output, accumulator}, b.getI64IntegerAttr(dim),
+      b.getBoolAttr(inclusive));
+
+  Region &scanOpRegion = scanOp.region();
+  auto &scanOpBlock = scanOpRegion.emplaceBlock();
+  scanOpBlock.addArguments({elementType, elementType}, {loc, loc});
+  OpBuilder regionBuilder(scanOpRegion);
+  auto blockArgs = scanOpBlock.getArguments();
+  Value inputElement = blockArgs[0];
+  Value accElement = blockArgs[1];
+  bodyBuild(regionBuilder, loc, inputElement, accElement);
+  return scanOp->getResult(0);
 }
 
 namespace {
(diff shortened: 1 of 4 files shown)


lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -523,6 +524,48 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+    sizes.erase(sizes.begin() + dim);
+    SmallVector<int64_t> accStatic(resultType.getShape());
+    accStatic.erase(accStatic.begin() + dim);
+    Value acc = createZeroInitTensor(rewriter, loc, sizes, elementType);
+    Type accType = RankedTensorType::get(accStatic, elementType);
+    acc = rewriter.create<tensor::CastOp>(loc, accType, acc);
+
+    Value result =
+        rewriter
+            .create<TMTensor::ScanOp>(loc, TypeRange{resultType, acc.getType()},
```
ramiro050, line 557: For the `TMTensor` ops, you need to create a block inside the region of the op with the computation that needs to be performed as the scan takes place. For example, this is how a `TMTensor::ScatterOp` is created:

https://github.com/llvm/torch-mlir/blob/4d693df63d9dd325a69d22f9180695349cd6771c/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp#L64-L65

The test is currently segfaulting because when the code that lowers the `ScanOp` tries to access the block inside the region of the op, there is no block to access.

You can look at the lit tests of `TMTensor` to see an [example of what the `ScanOp` should look like](https://github.com/llvm/torch-mlir/blob/e52e8868456ce3c99091f501c1c520989785342c/externals/llvm-external-projects/torch-mlir-dialects/test/tmtensor/convert_to_loops.mlir#L69).

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -2769,6 +2769,42 @@ def Aten_EmbeddingBagExample_basic(module, tu: TestUtils):
...
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, val):
+        return torch.ops.aten.cumsum(val, 1)
+
+@register_test_case(module_factory=lambda: CumsumModule())
+def CumsumModule_basic(module, tu: TestUtils):
+    module.forward(torch.randn(2, 7, 4))
```
ramiro050, line 2787: nit: use `tu.rand` to generate random tensors. See: https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#test-case-function. Same comment applies to the other test

ramiro050 (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
gpetters94: Ah, I see. I've added another helper like that for Scan, that seems to have done the trick.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -523,6 +548,50 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
 };
 } // namespace
 
+namespace {
+class ConvertAtenCumsumOp : public OpConversionPattern<AtenCumsumOp> {
```
ramiro050, line 552: This op also takes an optional `dtype` argument, which is currently not being handled here

https://pytorch.org/docs/stable/generated/torch.cumsum.html

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -523,6 +548,50 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+
+    sizes.erase(sizes.begin() + dim);
+    SmallVector<int64_t> accStatic(resultType.getShape());
+    accStatic.erase(accStatic.begin() + dim);
+    Value acc = createZeroInitTensor(rewriter, loc, sizes, elementType);
+    Type accType = RankedTensorType::get(accStatic, elementType);
+    acc = rewriter.create<tensor::CastOp>(loc, accType, acc);
+
+    Value result = createTMTensorScanOp(
+        rewriter, loc, input, output, acc, dim, true,
```
ramiro050, line 580: nit: add a comment `/*inclusive=*/true`

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -523,6 +548,50 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+    if (!matchPattern(op.dim(), m_TorchConstantInt(&dim)))
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only constant dim value is supported");
+
+    sizes.erase(sizes.begin() + dim);
+    SmallVector<int64_t> accStatic(resultType.getShape());
+    accStatic.erase(accStatic.begin() + dim);
+    Value acc = createZeroInitTensor(rewriter, loc, sizes, elementType);
+    Type accType = RankedTensorType::get(accStatic, elementType);
+    acc = rewriter.create<tensor::CastOp>(loc, accType, acc);
```
ramiro050, line 577: Why are the `output` and `acc` casts needed?

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -523,6 +548,50 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+    SmallVector<Value> sizes = getTensorSizes(rewriter, loc, input);
+    Value output = createZeroInitTensor(rewriter, loc, sizes, elementType);
+    output = rewriter.create<tensor::CastOp>(loc, resultType, output);
+
+    int64_t dim;
+    if (!matchPattern(op.dim(), m_TorchConstantInt(&dim)))
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only constant dim value is supported");
+
+    sizes.erase(sizes.begin() + dim);
```
ramiro050, line 572: nit: I think it would improve readability to create a new vector `SmallVector<Value> accSizes(sizes)`, and erase the element from that vector.

ramiro050 (changes requested):
Thanks for this!

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
gpetters94: I was having issues with input being statically sized but `getZeroInitTensor` returning a dynamic tensor, and casting them all to input's sizes (with one less dim in the accumulator's case) was simpler than dealing with both static and dynamic input. If you think it'd be better practice/improve readability I can change that logic.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -523,6 +548,56 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+    Value dtype = op.dtype();
+    if (!dtype.getType().isa<Torch::NoneType>())
+      return rewriter.notifyMatchFailure(
+          op, "unsupported: dtype argument not supported");
+
+    int64_t dim;
+    if (!matchPattern(op.dim(), m_TorchConstantInt(&dim)))
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only constant dim value is supported");
+
```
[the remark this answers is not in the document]
ramiro050: This should be done with the helpers `toPositiveDim` and `isValidDim`

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
ramiro050: I see. Yeah, this should be fine.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -523,6 +548,59 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+    if (!dtype.getType().isa<Torch::NoneType>())
+      return rewriter.notifyMatchFailure(
+          op, "unsupported: dtype argument not supported");
+
+    int64_t dim;
+    if (!matchPattern(op.dim(), m_TorchConstantInt(&dim)))
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only constant dim value is supported");
+    if (!isValidDim(dim, inputRank))
+      dim = toPositiveDim(dim, inputRank);
```
ramiro050, line 574: This should be done the other way. First do `toPositiveDim`, then check if it is valid. If it is not valid, a `notifyMatchFailure` should be returned with an error message. See for example:

https://github.com/llvm/torch-mlir/blob/99093d0623c3d3e6fb961ad0c8fda5ee9fc264fe/lib/Conversion/TorchToLinalg/DataMovement.cpp#L80-L82

ramiro050 (changes requested):

ramiro050 (approved):

PR #1325, opened 2022-08-31
Remove setting ulimit for docker runs

powderluv:
We added both ipc=host and explicit ulimits. This _may_ be causing slow downs on GHA. Remove the ulimit setting still passes all the CI tests locally. `--ipc=host` is still required.

Fixes: https://github.com/llvm/torch-mlir/pull/1325


powderluv:
Speculative. But I have another theory too (the local ccache may also be busted).

Will test on CI before merging.

powderluv:
The test run didn't help (but it also failed on macOS) but I think we should still land this since it gets us closer to where it was yesterday before my changes for the CI. https://github.com/llvm/torch-mlir/actions/runs/2966781198

PR #1327, opened 2022-08-31
Use pre-compiled headers for PyTorch Source builds

powderluv:
This should speed up source builds and ccache. May cause issues on macOS (https://github.com/pytorch/pytorch/issues/80018)


silvasean (approved):

silvasean:
Thanks!

powderluv:
The build time didn't seem to change much on the first run for ccache (1hr 47min) but lets see how it does on subsequent runs. Going to land and monitor.

PR #1330, opened 2022-09-01
Slightly tweak generated file checks

silvasean:
The new logic has the following benefits:
1. It does not clobber the working tree state. We expect testing to not
   change the work tree.
2. It correctly handles the case where a user has changes to the
   generated files, but hasn't checked them in yet (this happens
   frequently when adding new ops).


build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -186,26 +186,45 @@ function build_in_tree() {
   ccache -s
 }
 
+function _check_file_not_changed_by() {
+  # _check_file_not_changed_by <cmd> <file>
+  cmd="$1"
+  file="$2"
+  file_backup="/tmp/$(basename $file)"
+  file_new="/tmp/$(basename $file).new"
```
powderluv, line 194: Minor suggestion is to use the PWD which is more likely to be writable than /tmp/ which may or maynot have perms etc. We noticed similar failures with torch_mlir.compile() and we override writing to /tmp.

build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -186,26 +186,45 @@ function build_in_tree() {
...
+  # Run the command to regenerate it.
+  "$1" || return 1
+  # Save the new generated file.
+  cp "$file" "$file_new"
+  # Restore the original file. We want this function to not change the user's
+  # working tree state.
+  mv "$file_backup" "$file"
+  # We use git-diff as "just a diff program" (no SCM stuff) because it has
+  # nicer output than regular `diff`.
+  if ! git diff --quiet "$file" "$file_new"; then
```
powderluv, line 206: I think we still want @ashay feed back to redirect stderr / stdout to /dev/null.

powderluv (changes requested):

build_tools/python_deploy/build_linux_packages.sh
silvasean: Not sure I understand. --quiet shouldn't output anything.

powderluv: @ashay had suggested redirecting to `&> /dev/null` in this review https://github.com/llvm/torch-mlir/pull/1318 . Maybe it is not required with `--quiet`

build_tools/python_deploy/build_linux_packages.sh
silvasean: Done.

powderluv (approved):

PR #1315, opened 2022-08-30
[MHLO] refactor pass configurations

fortianyou:
Related to https://github.com/llvm/torch-mlir/issues/1227

1. Reduce MHLO #ifdefs
2. Dismiss compilation warnings

PR #1331, opened 2022-09-01
[MHLO] add new options to pipeline

fortianyou:
Follow up on the comments, add enableI32Index and enableStaticShape options in the pipeline:

> We already have this registered there:
> 
> ```
> #ifdef TORCH_MLIR_ENABLE_MHLO
>   mlir::PassPipelineRegistration<Torch::TorchLoweringPipelineOptions>(
>       "torch-backend-to-mhlo-backend-pipeline",
>       "Pipeline lowering torch backend contract to MHLO backend "
>       "contract.",
>       TorchConversion::createTorchBackendToMhloBackendPipeline);
> #endif
> ```
> 
> You can replace `Torch::TorchLoweringPipelineOptions` with a new struct with options parallel to `def ConvertTorchToMhlo : Pass<"convert-torch-to-mhlo", "func::FuncOp"> {` (this is redundant but required in MLIR now).

https://github.com/llvm/torch-mlir/pull/1315#issuecomment-1233552315

### DIFF
--- a/include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
+++ b/include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
@@ -36,9 +36,22 @@ void createTorchBackendToTosaBackendPipeline(
 
 // Do not register the torch-to-mhlo pipeline if mhlo target is disabled
 #ifdef TORCH_MLIR_ENABLE_MHLO
+struct MhloBackendPipelineOptions
+    : public PassPipelineOptions<MhloBackendPipelineOptions> {
+  Option<bool> enableStaticShape{
+      *this, "enable-static-shape",
+      llvm::cl::desc("Enable static shape conversion."), llvm::cl::init(false)};
+  // The i64 calculation is much slower than i32 on some devices, such as
+  // Nvidia GPU. One can truncate from i64 to i32 since dimension sizes
+  // are unlikely to exceed the range of i32(4GiB)
+  Option<bool> enableI32Index{
+      *this, "enable-i32-index",
+      llvm::cl::desc("Enable truncate index from i64 to i32(unsafely)"),
+      llvm::cl::init(false)};
+};
+
 void createTorchBackendToMhloBackendPipeline(
-    OpPassManager &pm,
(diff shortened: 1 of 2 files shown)


include/torch-mlir/Dialect/Torch/Transforms/Passes.h
```diff
@@ -40,6 +40,17 @@ struct TorchLoweringPipelineOptions
   Option<bool> decompose{*this, "decompose-complex-ops",
                          llvm::cl::desc("Decompose complex operations."),
                          llvm::cl::init(true)};
+  Option<bool> enableStaticShape{
```
silvasean, line 43: Can you put this in a separate struct MhloBackendPipelineOptions and put it in TorchConversion/Transforms/Passes.h

silvasean (changes requested):

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
fortianyou: I think this might be needed by another pipeline, since static shape and i32 index is not specific options to MHLO?

silvasean: TorchLoweringPipelineOptions is only for the frontend. It should not be used after the backend contract. (after you submit this I will remove it from the other backend pipelines)

Also these options are specific to MHLO. For example, TCP and Linalg uses `index` instead of i64/i32 and the static shape case is always a special case of dynamic shape, so TCP/Linalg is not irregular in that way. Each backend pipeline can have its own options to deal with its specific situation.

cc @burmako for visbility on MHLO design issues that we want to fix for StableHLO.

ZihengJiang (commented):
I don't have further opinion besides Sean's comments :)

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
fortianyou: Done

silvasean (approved):
Thanks!

ZihengJiang (approved):

ZihengJiang:
Thank you Tianyou!

PR #1332, opened 2022-09-01
Add decomposition to aten.native_layer_norm

fortianyou:

PR #1336, opened 2022-09-02
Delete ConvertAtenNativeLayerNormOp from TorchToLinalg

fortianyou:
The ConvertAtenNativeLayerNormOp is delete because we have decomposition already
see https://github.com/llvm/torch-mlir/pull/1332


silvasean (approved):
:heart: LGTM!

PR #1337, opened 2022-09-02
Add cases for view, type_as, and masked_fill

gpetters94:
This adds a few small changes that are needed for OPT support, namely:

- Support for `aten::view` when the output shape is statically known

- Folding away `torch::type_as` when both arguments are the same type

-  Support for `torch::masked_fill` when the mask is a float type


ramiro050:
Can you split the changes into 3 PRs? They are all quite independent from one another. It would also make the commit titles a lot more descriptive, since each PR could get as a title the description you have in the bullet points above.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -301,6 +301,42 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
           op, "desired size list length mismatches with the result type rank");
     }
 
+    // Case where all sizes (input and output) are statically known
```
ramiro050, line 304: @silvasean PTAL. Normally we try to avoid specializing on static shapes, right?

gpetters94:
I've split the PRs, this one is for masked_fill now.

python/torch_mlir_e2e_test/test_suite/constant_alloc.py
```diff
@@ -1462,3 +1462,25 @@ def forward(self, x, mask, value):
...
+        ([-1, -1], torch.int64, True),
+        ([-1, -1], torch.float, True),
+    ])
+    def forward(self, x, mask):
+        return torch.ops.aten.masked_fill(x, mask, value=0.1)
+
+
+@register_test_case(module_factory=lambda: MaskedFillTensorFloatMaskModule())
+def MaskedFillTensorFloatMaskModule_basic(module, tu: TestUtils):
+    mask = tu.randint(2, 3, low=0, high=1).to(torch.float)
```
ramiro050, line 1484: `high` should be `2`. The range created is `[low, high)`, so currently it will only produce zeros.

ramiro050 (approved):
I just have a small change request, but other than that it LGTM

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -906,6 +906,8 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
 
     Value input = payloadArgs[0];
     Value mask = payloadArgs[1];
+    if (mask.getType().isa<mlir::FloatType>())
+      mask = b.create<arith::ConstantOp>(loc, b.getBoolAttr(false));
```
ramiro050, line 910: I think I missed this the first time I reviewed your changes. Why is `mask` being set to `false` here?

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
gpetters94: That seems to be expected behavior. I was casting to `Int1` at first, but further testing shows that it seems to treat all floats as false. I haven't found anything in the documentation about it.

ramiro050: This seems to be a bug upstream. I would actually expect `float` mask to result in a runtime error, since this is the behavior that `aten.masked_select` has:

https://github.com/pytorch/pytorch/blob/5b58140d1a471b144baf66cc61a45a86746f0215/aten/src/ATen/native/TensorAdvancedIndexing.cpp#L1720-L1721

Can you file a bug upstream for this?

gpetters94: Okay, the bug is up [here](https://github.com/pytorch/pytorch/issues/85063). I'm going to leave this as-is for now in case it's expected behavior but I'll add an assert if it isn't.

PR #1339, opened 2022-09-02
Add bazel as an option to build_linux_packages.sh

powderluv:
Though bazel is not a fully supported flow, this allows for bazel builds to be built with the same interface
as CMake based Release, out-of-tree and in-tree.

Here is a sample run.

TM_PACKAGES="bazel" ./build_tools/python_deploy/build_linux_packages.sh
0.12s user 0.07s system 0% cpu 2:16.63 total

Hopefully this will enable drive-by fixes to be more easily done.

### DIFF
--- a/build_tools/python_deploy/build_linux_packages.sh
+++ b/build_tools/python_deploy/build_linux_packages.sh
@@ -83,6 +83,13 @@ function run_on_host() {
       export USERID=$(id -u)
       export GROUPID=$(id -g)
       ;;
+    bazel)
+      TM_CURRENT_DOCKER_IMAGE=${TM_CI_DOCKER_IMAGE}
+      # CI uses only Python3.10
+      TM_PYTHON_VERSIONS="cp310-cp310"
+      export USERID=$(id -u)
+      export GROUPID=$(id -g)
+      ;;
     *)
       echo "Unrecognized package '$package'"
       exit 1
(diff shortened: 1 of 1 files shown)


sjain-stanford (commented):
For now, if someone wants to help fix bazel builds, they can follow the instructions [here](https://github.com/llvm/torch-mlir/blob/main/docs/development.md#bazel-build), which should repro CI locally.

As [mentioned](https://github.com/llvm/torch-mlir/pull/1232#issuecomment-1218146572) earlier, I'd prefer to not combine the bazel docker build with `build_linux_packages.sh` for a few reasons:
1. Introduces "yet another way" of doing something that already exists
2. Divergent from current bazel CI adding to reproducibility issues / inconsistency
3. Doesn't support an iterative development workflow (current workflow does)
4. Doesn't expose bazel cache (current setup ensures bazel cache is reused both locally and in CI)
5. Couples bazel into CMake builds which as Sean [said](https://github.com/llvm/torch-mlir/pull/1232#issuecomment-1218326062) would be best to keep orthogonal.

I do see an opportunity to combine the dockerfiles and eliminate one. But this also means we would need to add new deps (that are not needed for CMake) to the main dockerfile, making the main docker image unnecessarily heavy. Please LMK your thoughts on this and I can work on this.

powderluv:
Sounds good. Will close this. No need to integrate further. The main docker already supports bazel builds but agreed best to keep it separate.

PR #1342, opened 2022-09-03
WIP: Add some ccache tweaks

powderluv:
In one case it with file clone build was much faster (1.38min vs
2.40min). Have to see how the 4-core GHA behaves.

In general trying to trade ccache space for less compute

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .cache/
 .vscode
 .ccache
+.ccache_tmp
 .env
 *.code-workspace
 .ipynb_checkpoints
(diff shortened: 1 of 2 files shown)


silvasean:
These types of changes can easily become "black magic" and difficult to undo later even if irrelevant. Can you indicate how to reproduce what you measured so that we can know how to test for the potential removal of these flags?

powderluv:
Yeah it is still black magic to me too. Hence WIP and no reviewers 😀. I don't think I'd merge this. I'm trying to understand what these flags do and if it will help.

powderluv:
Didn't seem to have any impact

PR #1343, opened 2022-09-05
build: update llvm tag to d2613d5b

ashay:
Summary of changes:
 - Update the dataflow analysis in RefineTypes.cpp
 - Add tosa-to-arith pass after tosa-to-linalg pass, since
   tosa-to-linalg (and canonicalizations) can produce tosa.const() ops
 - Fixed warning about not marking `matchAndRewrite` as override

### DIFF
--- a/lib/Dialect/Torch/Transforms/RefineTypes.cpp
+++ b/lib/Dialect/Torch/Transforms/RefineTypes.cpp
@@ -383,6 +383,12 @@ public:
   void visitOperation(Operation *op, ArrayRef<const ValueState *> operands,
                       ArrayRef<ValueState *> results) final;
 
+  void setToEntryState(ValueState *lattice) override {
+    auto refType = lattice->getPoint().getType();
+    auto knowledge = ValueKnowledge::getKnowledgeFromType(refType);
+    propagateIfChanged(lattice, lattice->join(knowledge));
+  }
+
 private:
   // Get the MLIR type of the tensor dtype given the dtype integer value and the
   // input dtype. When DType is None the type is inferred from the input dtype.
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -383,6 +383,13 @@ class TypeAnalysis : public dataflow::SparseDataFlowAnalysis<
   void visitOperation(Operation *op, ArrayRef<const ValueState *> operands,
                       ArrayRef<ValueState *> results) final;
 
+  void setToEntryState(ValueState *lattice) override {
+    if (lattice->isUninitialized())
+      return;
+    auto pessimisticState = ValueKnowledge::getPessimisticValueState(nullptr);
+    propagateIfChanged(lattice, lattice->join(pessimisticState));
```
sstamenova, line 390: Just a guess, but should this just be `propagateIfChanged(lattice, lattice->reset())`?

silvasean:
> Hey @silvasean, no rush, whenever you're back tomorrow or later, could you look at the changes in RefineTypes.cpp? There have been some changes to the [data flow analysis framework](https://github.com/llvm/llvm-project/commit/de0ebc5263f391519f3909c4f436afdf8adbc1ad) that I'm having a hard time applying to our code. I've made some partial changes but these cause the tests to fail because the return type of several operations is (incorrectly) inferred as `unk`.

I just checked out the branch and tried building it, but didn't see any unit test failures or e2e test failures. Can you give me more info on how to reproduce?

sjarus:
Tosa changes look fine to me.

silvasean:
> > Can you give me more info on how to reproduce?
> 
> Sorry, I think I managed to find a workaround, but I forgot to update the comment. Just to make sure that my changes aren't incorrect or suboptimal, could you check that the updates to RefineTypes.cpp make sense to you? Thanks much!

LGTM!

PR #1345, opened 2022-09-06
[MLIR][TORCH] Add E2E support for `aten.index.Tensor_hacked_twin` op

vivekkhandelwal1:
This commit adds lowering of `index.Tensor_hacked_twin` op.


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -1122,6 +1122,58 @@ def aten〇index〇Tensor(self: List[int], indices: List[Optional[List[int]]]) -
...
+    # shape is inserted in the location of the consecutive index tensors.
+    result_shape: List[int] = []
+    for i in range(first_index_tensor_location):
+        result_shape.append(unused_dim_sizes[i])
+    for broadcasted_size in broadcasted_shape:
+        result_shape.append(broadcasted_size)
+    for i in range(first_index_tensor_location, len(unused_dim_sizes)):
+        result_shape.append(unused_dim_sizes[i])
+    return result_shape
+
```
ramiro050, line 1176: This is the exact same code used for `aten.index.Tensor`. Please refactor the code into a helper function to avoid repetition.

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -1382,3 +1382,15 @@ func.func @torch.aten.roll(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.int,
...
   return %2 : !torch.vtensor<[?,?],f32>
 }
+
+// -----
+// CHECK-LABEL:   func.func @aten.index.Tensor_hacked_twin(
+// CHECK-SAME:                  %[[INP:.*]]: !torch.vtensor<[?],f32>, %[[INDEX:.*]]: !torch.vtensor<[?,?],si64>) -> !torch.vtensor<[?,?],f32> {
+// CHECK:           %[[INDICES:.*]] = torch.prim.ListConstruct %[[INDEX]] : (!torch.vtensor<[?,?],si64>) -> !torch.list<vtensor>
+// CHECK:           %[[RES:.*]] = torch.aten.index.Tensor %[[INP]], %[[INDICES]] : !torch.vtensor<[?],f32>, !torch.list<vtensor> -> !torch.vtensor<[?,?],f32>
+// CHECK:           return %[[RES]] : !torch.vtensor<[?,?],f32>
+func.func @aten.index.Tensor_hacked_twin(%arg0: !torch.vtensor<[?],f32>, %arg1: !torch.vtensor<[?,?],si64>) -> !torch.vtensor<[?,?],f32> {
```
ramiro050, line 1392: This test is no longer required. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#why-so-much-end-to-end-testing

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
vivekkhandelwal1: Hi @ramiro050, TBH I don't see a way of using a common helper for both the ops, since the type of indices for `index.tensor` is `List[Optional[List[int]]]`, and for `index.tensor_hacked_twin` is `List[List[int]]`.

ramiro050: `List[List[int]]` is a subtype of `List[Optional[List[int]]]`, so if the helper function takes as an argument a `List[Optional[List[int]]]`, it should work for both types

vivekkhandelwal1: I have tried this, but it's not working.
```
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py", line 1356, in <module>
    main(_create_argparse().parse_args())
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py", line 1284, in main
    torch.jit.script(v)
  File "/home/vivek/work/02_07/mlir_venv/lib/python3.8/site-packages/torch/jit/_script.py", line 1343, in script
    fn = torch._C._jit_script_compile(
RuntimeError: 

index_Tensor_Like(int[] self, int[]?[] indices) -> int[]:
Expected a value of type 'List[Optional[List[int]]]' for argument 'indices' but instead found type 'List[List[int]]'.
:
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py", line 1152
])
def aten〇index〇Tensor_hacked_twin(self: List[int], indices: List[List[int]]) -> List[int]:
    index_Tensor_Like(self, indices)
    ~~~~~~~~~~~~~~~~~ <--- HERE
    assert len(indices) <= len(self), "More indices than dimensions to index"
    broadcasted_shape: List[int] = []
```

ramiro050: Hm, that's strange. I think you should be able to get around this by doing something like

```python
def aten〇index〇Tensor_hacked_twin(self: List[int], indices: List[List[int]]) -> List[int]:
    optional_indices: List[Optional[List[int]]] = [x for x in indices]
```

to turn the `indices` to the right type

ramiro050 (approved):
LGTM!

PR #1349, opened 2022-09-07
[MLIR][TORCH] Add E2E support for `aten.bitwise_not` op

vivekkhandelwal1:
This commit adds lowering of `aten.bitwise_not` op.


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -935,6 +936,24 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    if (bitwiseNot.getType()
+            .cast<ValueTensorType>()
+            .getDtype()
+            .isa<mlir::FloatType>()) {
+      bitwiseNot.emitError("Bitwise_Not does not support floating point dtype");
+      return nullptr;
+    }
+    Type elementType = converter->convertType(bitwiseNot.getType())
+                           .cast<RankedTensorType>()
+                           .getElementType();
```
ramiro050, line 949: nit: can you move this up before the `FloatType` check? You should be able to then do `if (elementType.isa<mlir::FloatType>()) { // emit error...` instead

ramiro050 (approved):
Thanks! Just a small nit

PR #1354, opened 2022-09-08
[MLIR][TORCH] Fix shape calculation for `aten::pow.Tensor_Tensor` op

vivekkhandelwal1:


ramiro050 (changes requested):
Can you add an e2e test that would fail without this change?

ramiro050 (approved):

PR #1355, opened 2022-09-08
Add custom op example

nithinsubbiah:
This patch contains an example of how to register a custom op in Torch-MLIR which also acts as a test for custom op extensions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,6 +49,20 @@ else()
     set(ENV{TORCH_MLIR_ENABLE_LTC} 0)
 endif()
 
+option(TORCH_MLIR_USE_INSTALLED_PYTORCH "Build from local PyTorch in environment" ON)
+if(TORCH_MLIR_USE_INSTALLED_PYTORCH)
+    add_definitions(-DTORCH_MLIR_USE_INSTALLED_PYTORCH)
+else()
+    set(TORCH_MLIR_USE_INSTALLED_PYTORCH OFF)
+endif()
+
+option(TORCH_MLIR_CUSTOM_OP_EXAMPLE "Builds custom op example" OFF)
+if(NOT TORCH_MLIR_USE_INSTALLED_PYTORCH)
+    add_definitions(-DTORCH_MLIR_CUSTOM_OP_EXAMPLE)
+    set(TORCH_MLIR_CUSTOM_OP_EXAMPLE ON)
+endif()
+message(STATUS "TORCH_MLIR_CUSTOM_OP_EXAMPLE:" ${TORCH_MLIR_CUSTOM_OP_EXAMPLE})
+
 torch_mlir_add_llvm_external_project(
   torch-mlir-dialects
   TORCH_MLIR_DIALECTS
--- a/build_tools/update_shape_lib.sh
+++ b/build_tools/update_shape_lib.sh
@@ -22,11 +22,16 @@ if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
   pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
 fi
 TORCH_MLIR_EXT_MODULES="${TORCH_MLIR_EXT_MODULES:-""}"
+include_custom_op_example="${INCLUDE_CUSTOM_OP:-OFF}"
+if [ "$include_custom_op_example" == "ON" ]; then
+  ext_module="torch_mlir._torch_mlir_custom_op_example"
+fi
 if [ ! -z ${TORCH_MLIR_EXT_MODULES} ]; then
-  ext_module="${TORCH_MLIR_EXT_MODULES} "
+  ext_module="${ext_module},${TORCH_MLIR_EXT_MODULES} "
 fi
 
 PYTHONPATH="${pypath}" python \
   -m torch_mlir.dialects.torch.importer.jit_ir.build_tools.shape_lib_gen \
   --pytorch_op_extensions=${ext_module:-""} \
-  --torch_transforms_cpp_dir="${torch_transforms_cpp_dir}"
+  --torch_transforms_cpp_dir="${torch_transforms_cpp_dir}" \
+  --include_custom_op_example="${include_custom_op_example}"
--- a/build_tools/update_torch_ods.sh
+++ b/build_tools/update_torch_ods.sh
@@ -22,13 +22,17 @@ if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
   pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
 fi
 TORCH_MLIR_EXT_MODULES="${TORCH_MLIR_EXT_MODULES:-""}"
-ext_module="${ext_module:-""}"
+include_custom_op_example="${INCLUDE_CUSTOM_OP:-OFF}"
+if [ "$include_custom_op_example" == "ON" ]; then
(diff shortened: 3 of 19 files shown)


build_tools/update_shape_lib.sh
```diff
@@ -22,11 +22,15 @@ if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
   pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
 fi
 TORCH_MLIR_EXT_MODULES="${TORCH_MLIR_EXT_MODULES:-""}"
+include_custom_op_example=`grep TORCH_MLIR_CUSTOM_OP_EXAMPLE ${build_dir}/CMakeCache.txt | sed 's/.*=//'`
```
powderluv, line 25: CMakeCache is poisonous :D  Why do we want to grep there ? Is this just to check if it was enabled or not ? Maybe we just add a flag to updated_shape_lib.sh to register custom_op_example ?

build_tools/update_torch_ods.sh
```diff
@@ -22,13 +22,16 @@ if [ ! -z ${TORCH_MLIR_EXT_PYTHONPATH} ]; then
   pypath="${pypath}:${TORCH_MLIR_EXT_PYTHONPATH}"
 fi
 TORCH_MLIR_EXT_MODULES="${TORCH_MLIR_EXT_MODULES:-""}"
-ext_module="${ext_module:-""}"
+include_custom_op_example=`grep TORCH_MLIR_CUSTOM_OP_EXAMPLE ${build_dir}/CMakeCache.txt | sed 's/.*=//'`
```
powderluv, line 25: same

lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
```diff
@@ -6980,6 +6980,10 @@ module {
     %2 = call @__torch__.torch.jit._shape_functions.sum_mean_dim(%arg0, %0, %arg2, %1) : (!torch.list<int>, !torch.optional<list<int>>, !torch.bool, !torch.any) -> !torch.list<int>
     return %2 : !torch.list<int>
   }
+  func.func @"__torch_mlir_shape_fn._torch_mlir_custom_op_example.identity"(%arg0: !torch.list<int>) -> !torch.list<int> {
+    %0 = call @__torch__.torch.jit._shape_functions.unary(%arg0) : (!torch.list<int>) -> !torch.list<int>
+    return %0 : !torch.list<int>
+  }
```
powderluv, line 6986: Will this fail if we check this in but someone builds with custom_op_example ?

python/CMakeLists.txt
```diff
@@ -111,7 +118,9 @@ add_subdirectory(torch_mlir/eager_mode)
 # Required for running the update_torch_ods.sh and update_shape_lib.sh scripts.
 ################################################################################
 
-# add_subdirectory(torch_mlir/_torch_mlir_custom_op_example)
+if(NOT TORCH_MLIR_USE_INSTALLED_PYTORCH)
```
powderluv, line 121: should this be keyed on `TORCH_MLIR_CUSTOM_OP_EXAMPLE`

python/CMakeLists.txt
```diff
@@ -165,4 +174,8 @@ if(TORCH_MLIR_ENABLE_LTC)
   add_dependencies(TorchMLIRPythonModules reference_lazy_backend)
 endif()
 
+if(NOT TORCH_MLIR_USE_INSTALLED_PYTORCH)
```
powderluv, line 177: same should it be `TORCH_MLIR_CUSTOM_OP_EXAMPLE`

tools/e2e_test.sh
```diff
@@ -2,6 +2,10 @@
 set -euo pipefail
 
 src_dir="$(realpath "$(dirname "$0")"/..)"
+build_dir="${src_dir}/build"
+
+include_custom_op_example=`grep TORCH_MLIR_CUSTOM_OP_EXAMPLE ${build_dir}/CMakeCache.txt | sed 's/.*=//'`
```
powderluv, line 7: again anything peeking into CMakeCache == pain

lib/Dialect/Torch/Transforms/ShapeLibrary.cpp
nithinsubbiah: I think you meant what would happen when someone doesn't build custom_op_example? It would not cause any error

build_tools/update_shape_lib.sh
nithinsubbiah: By default custom op example isn't built (assuming PyTorch from binary install). When the build configuration has PyTorch source build enabled, the user has to manually enable the flag before running `update_torch_ods.sh` and `update_shape_lib.sh`. In order to remove this additional step for the user I used grep to get the status. 

Do you have a suggestion which doesn't require an additional step for the user and not use CMakeCache.txt? If not, then I will add a flag here and update the builder scripts too.

CMakeLists.txt
```diff
@@ -49,6 +49,23 @@ else()
...
+option(TORCH_MLIR_CUSTOM_OP_EXAMPLE "Builds custom op example" OFF)
+if(TORCH_MLIR_USE_INSTALLED_PYTORCH)
+    set(ENV{TORCH_MLIR_CUSTOM_OP_EXAMPLE} 0)
+else()
+    message(STATUS "Enabling custom op example")
+    add_definitions(-DTORCH_MLIR_CUSTOM_OP_EXAMPLE)
+    set(ENV{TORCH_MLIR_CUSTOM_OP_EXAMPLE} 1)
+    set(TORCH_MLIR_CUSTOM_OP_EXAMPLE ON)
+endif()
+
```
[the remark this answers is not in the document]
nithinsubbiah: The flag is used in files inside `lib\` so they cannot go inside `python/CMakelists.txt`

build_tools/update_shape_lib.sh
powderluv: I would rather take the additional step and document it than touch CMakeCache.txt.

powderluv:
This looks ok to me once CI passes. Please get sign off from @ramiro050 or @silvasean since it touches the main test files.

silvasean (changes requested):
Let's enable this always to avoid all the "if enable_custom_op_example" all over the codebase.

powderluv:
Unfortunately we can't cross compile custom ops (for now) and that would break apple M1 builds. This is a workaround so we get some CI coverage while the custom op overhaul happens.

silvasean:
I think we should undo all the custom op support on main and put it on a branch, which will exemplify the temporary nature of this solution. We can wait for the longer-term solution here to pan out on the main branch. If this was just a single cmake variable or ifdef it would be fine having it in main, but the way that this is being enabled/disabled is too ad-hoc and not up to the LLVM community standards that we are trying to uphold.

This custom op support design was pretty limited to begin with, and is really hard for users to use -- making them clone a branch to work from isn't a major impediment for those users vs having it in main. The main goal of this design was to have a solution for determined users to make forward progress, and it can serve that role on a branch.

powderluv:
I think a middle ground would be to leave this patch with all the Cmake vars out and put the example in a branch anyone can try. But leave the main branch with the infra for custom ops (pretty much every customer we support requires custom ops) until we rework it.

silvasean:
That seems reasonable.

powderluv:
I created a `custom_example_op` branch with the example.

nithinsubbiah:
Closing this PR since it's been moved to a new branch `custom-op-example` instead.

PR #1356, opened 2022-09-09
[MHLO] Add AtenClampOp conversion pattern to MHLO

Vremold:


test/Conversion/TorchToMhlo/basic.mlir
```diff
@@ -295,3 +295,97 @@ func.func @torch.aten.cat(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.vtenso
   %1 = torch.aten.cat %0, %int0 : !torch.list<vtensor>, !torch.int -> !torch.vtensor<[?,?],f32>
   return %1 : !torch.vtensor<[?,?],f32>
 }
+
+// -----
+
+// CHECK-LABEL:   func.func @torch.aten.clamp(
```
silvasean, line 301: Recommend not including these tests:

Vremold: Okay, I understand the concerns. Sorry for bothering you with the same problem again. I will remove these mlir tests. Should I add e2e test for it?

silvasean: Yes, there must always be an e2e test for new ops.

silvasean: There might already be a e2e test if the ops are supported on one of the other backends.

Vremold:
Please review for me. Thanks @silvasean

silvasean (approved):

PR #1353, opened 2022-09-08
Implement Expand/Collapse Functionality for Aten.View (Re-Opened_

JakopinA:
Second PR for this functionality after being opened with #1309 and closed with #1347 . Issues raised were with the exclusive use of static sizes and lack of unit tests.

PR #1357, opened 2022-09-09
Add static case for aten::view

gpetters94:
Adds a canonicalization to aten::view to use a static size list when the output size is statically known and adds a lowering for when the input and size list are statically known.


lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -301,6 +301,42 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
           op, "desired size list length mismatches with the result type rank");
     }
 
+    // Case where all sizes (input and output) are statically known
```
silvasean, line 304: Why are we doing this? I thought we could flatten to 1d and then expand in the worst case?

silvasean (changes requested):

lib/Conversion/TorchToLinalg/DataMovement.cpp
gpetters94: I don't think I understand the question - that's what the code here is doing, isn't it? Or do you mean this should only happen when the other cases aren't applicable?

silvasean: There should be no static shape limitation.

gpetters94: I didn't think that collapsing from `[?,?] -> [?]` was supported but I can give it a try.

gpetters94: Okay, I've refactored it a bit so it doesn't need a static input shape. Expanding `[?]` to either all static or all dynamic output shapes isn't supported, so I don't think we can do the collapse -> expand thing in an entirely dynamic context.

silvasean: @MaheshRavishankar FYI for expanding with dynamic dimensions

gpetters94: @silvasean Should this get merged then, given that the expansion isn't possible in a dynamic context?

MaheshRavishankar: Thanks Sean... Still dont have an owner for `tensor.expand_shape` with dynamic expanded dimensions. @gpetters94 just to clarify, `[?, ?] -> [?]` is supported with `tensor.collapse_shape` today.

silvasean: @gpetters94 ok let's merge it, but please file a bug on llvm/llvm-project for this and link it from a TODO in the code.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -301,6 +303,48 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
           op, "desired size list length mismatches with the result type rank");
     }
 
+    // Case where output size is all statically known
```
silvasean, line 306: please add .mlir unit tests for this case and all the others in this pattern!

silvasean (changes requested):

lib/Conversion/TorchToLinalg/DataMovement.cpp
qedawkins: Hold on I still have a problem here. This should not be doing a full flatten by default for the static dim case. We only want to do it if the other analysis fails. This seems to regress other static support that we have, assuming I correctly read how the code here is working. The correct approach should be in #1353 which constructs the reassociation indices for expand/collapse considering both dynamic and static shapes. The previous version of #1353 was reverted for issues with dynamic shape support as well as unit tests but that PR has the exact same problem with tensor.expand_shape.

The canonicalization here still makes sense however. #1353 needs to check with this PR to make sure all of the cases here are handled with the canonicalization in place.

gpetters94: @qedawkins @silvasean Do you think a refactor so that the logic in this PR runs only if all the other cases fail would work?

PR #1358, opened 2022-09-09
Fold away type_as

gpetters94:
Folds away aten::type_as when input and output type are the same.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -674,6 +674,20 @@ OpFoldResult AtenSqueezeDimOp::fold(ArrayRef<Attribute> operands) {
 }
 
 //===----------------------------------------------------------------------===//
+// AtenTypeAsOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenTypeAsOp::fold(ArrayRef<Attribute> operands) {
+  Type inType = self().getType();
+  Type newType = other().getType();
+
+  if (inType == newType)
+    return self();
+
+  return nullptr;
+}
+
+//===----------------------------------------------------------------------===//
 // AtenToDtypeOp
 //===----------------------------------------------------------------------===//
 
--- a/python/torch_mlir_e2e_test/test_suite/type_conversion.py
+++ b/python/torch_mlir_e2e_test/test_suite/type_conversion.py
@@ -214,3 +214,23 @@ class ToDtypeBoolLayoutNoneModule(torch.nn.Module):
 @register_test_case(module_factory=lambda: ToDtypeBoolLayoutNoneModule())
 def ToDtypeBoolLayoutNoneModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 5))
+
+
+class TypeAsSameModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, x, y):
+        return torch.ops.aten.type_as(x, y)
+
+
+@register_test_case(module_factory=lambda: TypeAsSameModule())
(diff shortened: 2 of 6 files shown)


python/torch_mlir_e2e_test/test_suite/type_conversion.py
```diff
@@ -214,3 +214,24 @@ def forward(self, x):
...
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, x, y):
+        return torch.ops.aten.type_as(x,
+                                 y)
```
ramiro050, line 231: nit: formatting is off here. This should be a single line

python/torch_mlir_e2e_test/test_suite/type_conversion.py
```diff
@@ -214,3 +214,24 @@ def forward(self, x):
...
+
```
ramiro050, line 232: nit: there is an extra empty line. The convention in this file is to use two empty lines to separate test case from class definition

python/torch_mlir_e2e_test/test_suite/type_conversion.py
```diff
@@ -214,3 +214,24 @@ def forward(self, x):
 @register_test_case(module_factory=lambda: ToDtypeBoolLayoutNoneModule())
 def ToDtypeBoolLayoutNoneModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 5))
+
```
ramiro050, line 217: nit: there should be an extra empty line here. Convention is to use two empty lines to separate test case from class definition

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -673,6 +673,22 @@ OpFoldResult AtenSqueezeDimOp::fold(ArrayRef<Attribute> operands) {
   return nullptr;
 }
 
+//===----------------------------------------------------------------------===//
+// AtenTypeAsOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenTypeAsOp::fold(ArrayRef<Attribute> operands) {
+  Value input = getOperand(0);
+  Value output = getOperand(1);
```
ramiro050, line 682: nit: It's more readable to use `self()` and `other()` to get the operands. Also, it makes more sense to call the second operand `other` rather than `output`, since it does not represent the output of the op.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -673,6 +673,22 @@ OpFoldResult AtenSqueezeDimOp::fold(ArrayRef<Attribute> operands) {
...
+  Type inType = input.getType();
+  Type outType = output.getType();
```
ramiro050, line 684: nit: I would call this `newType`

ramiro050 (changes requested):
This change should also include a lit test to test successful and non-successful foldings here https://github.com/llvm/torch-mlir/blob/main/test/Dialect/Torch/canonicalize.mlir

See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

> With the exceptions above, all changes should include appropriate unit tests, as is standard in the LLVM and MLIR community. This includes full coverage of all canonicalizations, pretty printing, passes, errors, and diagnostics.

gpetters94:
> This change should also include a lit test to test successful and non-successful foldings here https://github.com/llvm/torch-mlir/blob/main/test/Dialect/Torch/canonicalize.mlir
> 
> See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing
> 
> > With the exceptions above, all changes should include appropriate unit tests, as is standard in the LLVM and MLIR community. This includes full coverage of all canonicalizations, pretty printing, passes, errors, and diagnostics.

I'm actually not sure if I can write a lit test for that. The canonicalization depends on static vs. dynamic output shapes which I'm not sure how to specify in a lit test.

ramiro050 (approved):
LGTM!

PR #1360, opened 2022-09-12
Add more info about unit tests.

silvasean:


silvasean:
cc @JakopinA  PTAL

JakopinA:
> cc @JakopinA PTAL
Looking, thank you very much!

PR #1363, opened 2022-09-13
python: add `CHECK-LABEL` statements to localize checks

ashay:
It seems as though an upstream change in PyTorch has caused the module
dump to include not just the module being tested, but also several
seemingly unrelated functions in the `torch._decom.decompositions`
namespace.  The presence of these new functions caused lit to match
variables against incorrect statements (i.e.  statements in the
unrelated functions instead of the module under test).

This patch inserts `CHECK-LABEL` statements in the failing tests so that
lit ignores these unrelated functions and only checks the statements at
or after the test module definition.


silvasean (approved):
Thanks!

PR #1365, opened 2022-09-13
Remove TorchLoweringPipelineOptions from TorchConversion pipelines

silvasean:
TorchLoweringPipelineOptions only applies to the frontend lowering pipeline.


fortianyou (approved):
LGTM

Issue #1305, opened 2022-08-29
Shape simplification failed for AtenArangeOp


Vremold:
I am lowering a TS model to torch dialect. There is a `torch.aten.arange.start` op. And the shape simplification pipeline failed to reason out that its result should have a static shape.
This is the IR before shape simplification pipeline:
``` mlir
%none = torch.constant.none
%int0 = torch.constant.int 0
%int5 = torch.constant.int 5
%0 = torch.aten.arange.start %int0, %int5, %none, %none, %none, %none : !torch.int, !torch.int, !torch.none, !torch.none, !torch.none, !torch.none -> !torch.vtensor
```
And this is the IR after shape simplification pipeline:
```
%none = torch.constant.none
%int0 = torch.constant.int 0
%int5 = torch.constant.int 5
%0 = torch.aten.arange.start %int0, %int5, %none, %none, %none, %none : !torch.int, !torch.int, !torch.none, !torch.none, !torch.none, !torch.none -> !torch.vtensor<[?],unk>
```
The reason seems to be related to the shape-lib for `torch.aten.arange_start` op, which comes from `torch.jit._shape_functions` and accepts `Union[int, float]` as input. The following is the part of IR inserted during the shape calculation:
```
%0 = torch.shape.calculate {
  %2 = torch.aten.arange.start %int0, %int5, %none, %none, %none, %none : !torch.int, !torch.int, !torch.none, !torch.none, !torch.none, !torch.none -> !torch.vtensor<[?],unk>
  torch.shape.calculate.yield %2 : !torch.vtensor<[?],unk>
} shapes {
  %2 = torch.operator "aten.ge"(%float5.000000e00, %int0) : (!torch.float, !torch.int) -> !torch.bool
  torch.prim.If %2 -> () {
    torch.prim.If.yield
  } else {
    torch.prim.RaiseException %str, %none : !torch.str, !torch.none
    torch.prim.If.yield
  }
  %3 = torch.operator "aten.ge"(%float5.000000e00, %float0.000000e00) : (!torch.float, !torch.float) -> !torch.bool
  torch.prim.If %3 -> () {
    torch.prim.If.yield
  } else {
    torch.prim.RaiseException %str, %none : !torch.str, !torch.none
    torch.prim.If.yield
  }
  %4 = torch.operator "aten.sub"(%float5.000000e00, %float0.000000e00) : (!torch.float, !torch.float) -> !torch.number
[...]

PR #1367, opened 2022-09-14
Strength the shape inference for aten.arange-like op

Vremold:
This PR is to solve issue #1305 .
A example IR is like:
```
%none = torch.constant.none
%int0 = torch.constant.int 0
%int5 = torch.constant.int 5
%0 = torch.aten.arange.start %int0, %int5, %none, %none, %none, %none : !torch.int, !torch.int, !torch.none, !torch.none, !torch.none, !torch.none -> !torch.vtensor
```
After shape simplification pipeline, %0 is unexpectedly deduced to have a dynamic shape.
```
...
%0 = torch.aten.arange.start %int0, %int5, %none, %none, %none, %none : !torch.int, !torch.int, !torch.none, !torch.none, !torch.none, !torch.none -> !torch.vtensor<[?],unk>
```
The solution in this PR can be briefly summarized as the following：
1. register aten.sub and aten.ceil.Scalar op, which is need in shape function of aten arange-like ops
2. design folder for aten.sub and aten.ceil.Scalar op
3. register a new constant-like op: Torch::ConstantNumberOp. This is because the shape function of aten arange-like ops requires the return type to be torch.number.
4. design canonicalizer for ConstantNumberOp to unconditionally convert it to Constant{Int,Float}Op and Derefine op.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
@@ -759,7 +759,7 @@ def Torch_ConstantNumberOp : Torch_Op<"constant.number",
   let description = [{
     This op is used as a workaround to the fact that the constant 
     materialization in MLIR must materialize a constant with a single op. 
-    To materialize ops with a static `!torch.number type, we must use this op, 
+    To materialize ops with a static `!torch.number` type, we must use this op, 
     even though we statically know if it is an integer or a float.
 
     Note: This op unconditionally canonicalizes to 
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
@@ -753,6 +753,25 @@ def Torch_ConstantFloatOp : Torch_Op<"constant.float", [
   let hasFolder = 1;
 }
 
+def Torch_ConstantNumberOp : Torch_Op<"constant.number",
+    [ConstantLike, NoSideEffect]> {
+  let summary = "Materialize a constant `number` value.";
+  let description = [{
+    Note: 
+    1. For now, only integer or float input is supported.
+    2. This op will finally be converted to 
+       `torch.constant.{float,int}` + `torch.derefine`
+  }];
+  let arguments = (ins
+    AnyAttrOf<[F64Attr, I64Attr]>:$value
+  );
+  let results = (outs
+    Torch_NumberType:$result
+  );
+  let hasFolder = 1;
+  let hasCanonicalizer = 1;
+}
+
 def Torch_ConstantBoolOp : Torch_Op<"constant.bool", [
     ConstantLike,
     NoSideEffect,
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
@@ -757,10 +757,13 @@ def Torch_ConstantNumberOp : Torch_Op<"constant.number",
     [ConstantLike, NoSideEffect]> {
   let summary = "Materialize a constant `number` value.";
   let description = [{
-    Note: 
-    1. For now, only integer or float input is supported.
-    2. This op will finally be converted to 
-       `torch.constant.{float,int}` + `torch.derefine`
+    This op is used as a workaround to the fact that the constant 
+    materialization in MLIR must materialize a constant with a single op. 
+    To materialize ops with a static `!torch.number type, we must use this op, 
+    even though we statically know if it is an integer or a float.
+
+    Note: This op unconditionally canonicalizes to 
+    `torch.constant.{float,int}` + `torch.derefine`
   }];
   let arguments = (ins
     AnyAttrOf<[F64Attr, I64Attr]>:$value
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -1594,27 +1594,29 @@ void Torch::ConstantFloatOp::getAsmResultNames(
 //===----------------------------------------------------------------------===//
 // ConstantNumberOp
 //===----------------------------------------------------------------------===//
+
 OpFoldResult Torch::ConstantNumberOp::fold(ArrayRef<Attribute> operands) {
   return valueAttr();
 }
 
 void Torch::ConstantNumberOp::getCanonicalizationPatterns(
-  RewritePatternSet &patterns, MLIRContext *context) {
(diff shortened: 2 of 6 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1948,6 +1974,41 @@ OpFoldResult AtenMulIntOp::fold(ArrayRef<Attribute> operands) {
...
+
+OpFoldResult AtenSubOp::fold(ArrayRef<Attribute> operands) {
+  int64_t intLhs, intRhs;
+  if (matchPattern(getOperand(0), m_TorchConstantInt(&intLhs)) && 
+      matchPattern(getOperand(1), m_TorchConstantInt(&intRhs))) {
+    return getI64IntegerAttr(getContext(), intLhs - intRhs);
+  }
+  
+  double floatLhs, floatRhs;
+  if (matchPattern(getOperand(0), m_TorchConstantFloat(&floatLhs)) &&
```
silvasean, line 1989: can we share this with sub.int/sub.float? E.g. call `AtenSubIntOp::fold` and `AtenSubFloatOp::fold`? Something like:

```
if (!operands[0] || !operands[1])
  return nullptr;
if (operands[0].isa<IntegerAttr>() && operands[1].isa<IntegerAttr>())
  return AtenSubIntOp::fold(operands);
...
```

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1948,6 +1974,41 @@ OpFoldResult AtenMulIntOp::fold(ArrayRef<Attribute> operands) {
...
+  return nullptr;
+}
+
+//===----------------------------------------------------------------------===//
+// AtenCeilScalarOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenCeilScalarOp::fold(ArrayRef<Attribute> operands) {
+  int64_t intVal;
+  if (matchPattern(getOperand(), m_TorchConstantInt(&intVal))) {
```
silvasean, line 2002: Does the constant int case happen in practice?

silvasean (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1591,6 +1591,32 @@ void Torch::ConstantFloatOp::getAsmResultNames(
   setNameFn(getResult(), StringRef(buf.data(), buf.size()));
 }
 
+//===----------------------------------------------------------------------===//
+// ConstantNumberOp
+//===----------------------------------------------------------------------===//
+OpFoldResult Torch::ConstantNumberOp::fold(ArrayRef<Attribute> operands) {
```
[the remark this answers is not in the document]
Vremold: Done.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -735,6 +735,25 @@ def Torch_ConstantFloatOp : Torch_Op<"constant.float",
   let hasFolder = 1;
 }
 
+def Torch_ConstantNumberOp : Torch_Op<"constant.number",
+    [ConstantLike, NoSideEffect]> {
+  let summary = "Materialize a constant `number` value.";
+  let description = [{
+    Note: 
+    1. For now, only integer or float input is supported.
+    2. This op will finally be converted to
```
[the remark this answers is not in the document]
Vremold: Done.

lib/Dialect/Torch/IR/TorchOps.cpp
Vremold: `fold` is not static method of Operation class. We can not invoke this function without constructing a concrete operation.

silvasean: Okay then let's factor out helpers and share the code.

silvasean: It looks like we need to change atenBinaryIntOperatorFoldHelper and also adapt it to use `ArrayRef<Attribute> operands` rather than matchPattern

Vremold: @silvasean  I've reimplemented the `atenBinaryIntOperatorFoldHelper` and added `atenBinaryFloatOperatorFoldHelper` to handle   float cases. In this PR, I don't want to change other codes too much, so I kept the old `atenBinaryIntOperatorFoldHelper` and its usages. If the reimplementation is what you want, I will open another PR to refactor the codes.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -735,6 +735,28 @@ def Torch_ConstantFloatOp : Torch_Op<"constant.float",
   let hasFolder = 1;
 }
 
+def Torch_ConstantNumberOp : Torch_Op<"constant.number",
+    [ConstantLike, NoSideEffect]> {
+  let summary = "Materialize a constant `number` value.";
+  let description = [{
+    This op is used as a workaround to the fact that the constant 
+    materialization in MLIR must materialize a constant with a single op. 
+    To materialize ops with a static `!torch.number type, we must use this op,
```
silvasean, line 744: nit: missing "`"

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1887,8 +1915,9 @@ OpFoldResult Aten__Contains__IntListOp::fold(ArrayRef<Attribute> operands) {
 
 using BinaryIntOperatorFn = std::function<int64_t(int64_t, int64_t)>;
 template <typename OpTy>
-static OpFoldResult atenBinaryIntOperatorFoldHelper(OpTy op,
-                                                    BinaryIntOperatorFn f) {
+static std::enable_if_t<!std::is_same_v<OpTy, ArrayRef<Attribute>>,
+                        OpFoldResult>
+atenBinaryIntOperatorFoldHelper(OpTy op, BinaryIntOperatorFn f) {
```
silvasean, line 1920: You should be able to unify these with just `ArrayRef<Attribute> operands` (this will require changing some of the other folders)

silvasean (changes requested):

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
Vremold: Done.

lib/Dialect/Torch/IR/TorchOps.cpp
Vremold: Done.

silvasean (approved):
LGTM! Thanks!

Vremold:
Merged after rebasing. Thanks @silvasean .

PR #1368, opened 2022-09-14
[MLIR][TORCH] Add support for bool element type for aten.sum[.dim_Int…

vivekkhandelwal1:
…List] op

This commit adds bool element type support for `aten.sum` and `aten.sum.dim_IntList` op.

### DIFF
--- a/python/torch_mlir_e2e_test/test_suite/reduction.py
+++ b/python/torch_mlir_e2e_test/test_suite/reduction.py
@@ -49,6 +49,25 @@ def ReduceSumDtypeFloatModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
+class ReduceSumElementTypeBoolModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.bool, True),
+    ])
+    def forward(self, a):
+        return torch.sum(a)
+
+
+@register_test_case(module_factory=lambda: ReduceSumElementTypeBoolModule())
+def ReduceSumElementTypeBoolModule_basic(module, tu: TestUtils):
+    module.forward(tu.randint(3, 4, 5, high=2).to(torch.bool))
+
+# ==============================================================================
+
 class ReduceSumDimIntListFloatModule(torch.nn.Module):
     def __init__(self):
         super().__init__()
@@ -122,6 +141,25 @@ class ReduceSumDimIntListEmptyDimModule(torch.nn.Module):
 @register_test_case(module_factory=lambda: ReduceSumDimIntListEmptyDimModule())
 def ReduceSumDimIntListEmptyDimModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 4, 5))
+
+# ==============================================================================
+
+class ReduceSumDimIntListElementTypeBoolModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.bool, True),
+    ])
+    def forward(self, a):
+        return torch.sum(a, dim=(-1), keepdim=False)
+
+
+@register_test_case(module_factory=lambda: ReduceSumDimIntListElementTypeBoolModule())
+def ReduceSumDimIntListElementTypeBoolModule_basic(module, tu: TestUtils):
+    module.forward(tu.randint(1, 128, high=2).to(dtype=torch.bool))
 
 # ==============================================================================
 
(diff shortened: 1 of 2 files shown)


python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -49,6 +49,25 @@ def ReduceSumDtypeFloatModule_basic(module, tu: TestUtils):
...
+        None,
+        ([-1, -1, -1], torch.bool, True),
+    ])
+    def forward(self, a):
+        return torch.sum(a).to(torch.bool)
+
+
+@register_test_case(module_factory=lambda: ReduceSumElementTypeBoolModule())
+def ReduceSumElementTypeBoolModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5).to(torch.bool))
```
ramiro050, line 67: When converting floats to bool, PyTorch only assigns `False` to the value `0.0`, so `tu.rand` is almost guaranteed to only produce `True` values.

To generate a bool tensor, you should do `tu.randint(3, 4, 5, high=2).to(torch.bool)`

python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -125,6 +144,25 @@ def ReduceSumDimIntListEmptyDimModule_basic(module, tu: TestUtils):
...
+        None,
+        ([1, 128], torch.bool, True),
+    ])
+    def forward(self, a):
+        return torch.sum(a, dim=(-1), keepdim=False).to(dtype=torch.bool)
+
+
+@register_test_case(module_factory=lambda: ReduceSumDimIntListElementTypeBoolModule())
+def ReduceSumDimIntListElementTypeBoolModule_basic(module, tu: TestUtils):
+    module.forward(tu.randint(1, 128).to(dtype=torch.bool))
```
ramiro050, line 162: Can you set `high=2` in `tu.randint` to have a roughly even distribution of `True` and `False`? (`tu.randint` defaults to `high=10`, so this will produce many more `True`s than `False`)

python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -49,6 +49,25 @@ def ReduceSumDtypeFloatModule_basic(module, tu: TestUtils):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.bool, True),
+    ])
+    def forward(self, a):
+        return torch.sum(a).to(torch.bool)
```
ramiro050, line 62: I think it's better to not have the `to.(torch.bool)` here, since this conversion will turn any non-zero number to `True`, preventing us from knowing if the summation is actually working correctly. The same applies to the other test

python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -125,6 +144,25 @@ def ReduceSumDimIntListEmptyDimModule_basic(module, tu: TestUtils):
...
 # ==============================================================================
 
+class ReduceSumDimIntListElementTypeBoolModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([1, 128], torch.bool, True),
```
ramiro050, line 154: Is there a reason to use static shapes for this test?

lib/Conversion/TorchToLinalg/Reduction.cpp
```diff
@@ -400,8 +400,17 @@ class ConvertReductionOp : public ConversionPattern {
...
+      // Check if the element type is bool.
+      if (elemType.isInteger(1)) {
+        result = createLinalgPayloadForReduceOp(
+            builder, loc, payloadArgs, op, operands,
+            mlir::IntegerType::get(op->getContext(), 64));
+        result = convertScalarToDtype(rewriter, loc, result, elemType);
+      } else {
+        result = createLinalgPayloadForReduceOp(builder, loc, payloadArgs, op,
+                                                operands, elemType);
+      }
```
ramiro050, line 413: I don't think this file needs any modifications. It seems to me that the issue is with `RefineTypes`:

https://github.com/llvm/torch-mlir/blob/2bb5f4d8fe1f91715716f46b9b8b895aec078021/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L932

The result type should be `int64` whenever the input is a bool tensor.

vivekkhandelwal1: Thanks, Ramiro for your valuable suggestions.

vivekkhandelwal1:
Hi @ramiro050 could you please review this patch?

ramiro050 (approved):
LGTM! Thanks!

PR #1369, opened 2022-09-14
Publish Python 3.7 packages.

dellis23:
This is the version Colab uses.

### DIFF
--- a/build_tools/python_deploy/build_linux_packages.sh
+++ b/build_tools/python_deploy/build_linux_packages.sh
@@ -44,7 +44,7 @@ TM_RELEASE_DOCKER_IMAGE="${TM_RELEASE_DOCKER_IMAGE:-stellaraccident/manylinux201
 # ./build_tools/docker/Dockerfile
 TM_CI_DOCKER_IMAGE="${TM_CI_DOCKER_IMAGE:-powderluv/torch-mlir-ci:latest}"
 # Version of Python to use in Release builds. Ignored in CIs.
-TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp39-cp39 cp310-cp310}"
+TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp37-cp37m cp39-cp39 cp310-cp310}"
 # Location to store Release wheels
 TM_OUTPUT_DIR="${TM_OUTPUT_DIR:-${this_dir}/wheelhouse}"
 # What "packages to build"


silvasean:
cc @powderluv

silvasean:
I'm personally fine with this. Will let @powderluv make the final call.

powderluv:
As much I don't like old Python versions :) this is a valid use case. Thanks for the PR.

build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -44,7 +44,7 @@ TM_RELEASE_DOCKER_IMAGE="${TM_RELEASE_DOCKER_IMAGE:-stellaraccident/manylinux201
 # ./build_tools/docker/Dockerfile
 TM_CI_DOCKER_IMAGE="${TM_CI_DOCKER_IMAGE:-powderluv/torch-mlir-ci:latest}"
 # Version of Python to use in Release builds. Ignored in CIs.
-TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp39-cp39 cp310-cp310}"
+TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp37-cp37 cp39-cp39 cp310-cp310}"
```
powderluv, line 47: This should be `cp37-cp37m`

silvasean: It also says "Ignored in CI's" -- do we need to update somewhere else for the nightly packages?

dellis23: Done, thanks.  Is there a reference somewhere I can look at to understand this?

powderluv: Yeah we don't use TM_PYTHON_VERSIONS for CIs since our CI run on Ubuntu 22.04 with Python 3.10. These versions come from the naming scheme using my manylinux (and they have it as cp37-cp37m).  The `m` says it is Python 3.7 with the pymalloc extension (https://peps.python.org/pep-3149/#proposal). Later versions have it integrated by default.

powderluv:
hold on -- can we use Python 3.9 / 3.10 in Colab ? https://www.datasciencelearner.com/change-python-version-in-google-colab-steps/ 

Saves us building (every night) and publishing 3.7m 200MB files every night.

dellis23:
> hold on -- can we use Python 3.9 / 3.10 in Colab ? https://www.datasciencelearner.com/change-python-version-in-google-colab-steps/
> 
> Saves us building (every night) and publishing 3.7m 200MB files every night.

I tried this earlier but ran into some issues.  One of the issues I got past was an internal vs external colab issue.  However, I'm still hitting the same issues people ran into in [this](https://stackoverflow.com/questions/68657341/how-can-i-update-google-colabs-python-version) thread, namely that while this does install the new Python at the system level, it doesn't work for the actual colab notebook cell execution.

powderluv:
The CI (running on GHA) is frail so if we find a way to run the colab with 3.9 or 3.10 lets document and use that.  Adding another Python version can potentially lead to timeouts of the builds etc (which is why I dropped them earlier).

dellis23:
I'll keep looking into it, though it doesn't seem promising.  Another idea I had: if we built a 3.7 version, would it be compatible with 3.9?  If so, we could maintain one version pinned to the version colab uses, and one cutting edge version on the latest version of Python.

dellis23:
The only reliable workaround I've found so far is to not use a colab runtime at all, instead opting for a local one.  See here: https://github.com/iree-org/iree-torch/pull/41

That being said, it would be nice if people could just install this package in a vanilla colab notebook, so I do think we should still try to figure this out.

jpienaar:
> The CI (running on GHA) is frail so if we find a way to run the colab with 3.9 or 3.10 lets document and use that. Adding another Python version can potentially lead to timeouts of the builds etc (which is why I dropped them earlier).

Would an option here be to have this be a separate action to not affect the CI? If the actual colab cells don't use newly installed python version then that's not very useful.

dellis23:
Thank you!

PR #1370, opened 2022-09-14
shape-lib: generate string as multiple lines to work with MSVC

ashay:
As @oroppas identified, literal strings that are over 16,380 characters
cause the MSVC compiler to throw an error, eventually causing the
Windows build of Torch-MLIR to fail because the length of the generated
MLIR for the shape library crosses the allowed threshold.

This patch fixes the problem by making the Python script generate one
literal string per line to satisfy the MSVC compiler.

Thanks to @oroppas for the bulk of the effort required to resolve this!

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
@@ -1251,6 +1251,17 @@ def main(args):
     # Put the `〇` back to a regular `.`.
     asm = asm.replace("\\E3\\80\\87", ".")
 
+    # We're about to put quotes around the string, so escape the `"` characters.
+    asm = asm.replace("\"", "\\\"")
+
+    # Instead of dumping one big chunk of text that is several thousand lines
+    # long (and which causes MSVC to error out), split it into multiple lines.
+    # See MSVC Compiler Error C2026
+    # [https://docs.microsoft.com/en-us/cpp/error-messages/compiler-errors-1/compiler-error-c2026?view=msvc-170]
+    # for details.
+    multiple_lines = asm.replace("\n", "\\n\"\n\"")
+    asm = f"\"{multiple_lines}\""
+
     # Write out the shape library .cpp file.
     shape_lib_cpp_file = os.path.join(
         args.torch_transforms_cpp_dir, "ShapeLibrary.cpp")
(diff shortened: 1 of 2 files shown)


powderluv (commented):
awesome thanks. Just in time for me to build on Windows. 

BTW if you build with VS2022's clang I had to specify the LLVM target triple also. 

Will let @silvasean give approval on this but LGTM

powderluv:
Thanks for sharing the command line. I used the same except I also try specifying `-DCMAKE_C_COMPILER=clang` and clang++ so it runs in the VS2022 clang to shake out anything with it. 

Thanks for the pointer for ``--config Release`` .  I noticed the same with IREE on windows. 

Excited about it. Lets add a Windows CI once we get it to work.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -1276,19 +1284,16 @@ def main(args):
...
 StringRef mlir::torch::Torch::getShapeLibrary() {{
-// TODO: Find a way to embed this string nicely.
-// It is currently too long, and will probably break MSVC builds if anyone
-// attempts that.
-// We want to preserve the legibility of the shape library as a checked in file,
-// since that is sometimes useful for debugging / diffing.
-// Probably the ideal outcome is to have the shape library be a .mlir file
-// that is checked in, and then we embed it as part of the build process.
+#ifndef _MSC_VER
 #pragma clang diagnostic push
```
silvasean, line 1288: I don't think we need the pragmas now that we are doing the string line-wise

silvasean:
This seems fine in principle. Is there a way to not wrap stuff like this?

```
         "  func.func "
         "@__torch__.torch._decomp.decompositions.nll_loss_backward(%arg0: "
         "!torch.tensor, %arg1: !torch.tensor, %arg2: !torch.tensor, %arg3: "
         "!torch.optional<tensor>, %arg4: !torch.int, %arg5: !torch.int, "
         "%arg6: !torch.tensor) -> !torch.tensor {\n"
```
?

Having that be a single string would reduce the number of lines churned and make it a bit more readable.

silvasean (approved):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
silvasean: That's fun. Clang must be considering the preprocessor-concatenated size rather than the individual string literal size, and MSVC only cares about the individual string literals? :rofl:

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -1251,6 +1251,14 @@ def main(args):
     # Put the `〇` back to a regular `.`.
     asm = asm.replace("\\E3\\80\\87", ".")
 
+    # We're about to put quotes around the string, so escape the `"` characters.
+    asm = asm.replace("\"", "\\\"")
+
+    # Instead of dumping one big chunk of text that is several thousand lines
```
silvasean, line 1257: nit: can you add a link to a doc about the MSVC limitation?

Issue #1364, opened 2022-09-13
LTC not in CI is causing problems


silvasean:
When building Torch-MLIR with LTC on Linux I'm now seeing this error, which seems it would have been caught on CI: https://gist.github.com/silvasean/8e8f3cad89d6c97070e1657c0d5db347

It seems we have LTC disabled in CI due to a mac issue: https://github.com/llvm/torch-mlir/blob/a9e1014fc70ae85e90205f1d9ed25defb65dc466/.github/workflows/buildAndTest.yml#L69

Do we have a plan for how to re-enable this or what the support plan is for LTC in Torch-MLIR?

silvasean:
And here is where we have it disabled in the Linux build:

https://github.com/llvm/torch-mlir/blob/a9e1014fc70ae85e90205f1d9ed25defb65dc466/build_tools/python_deploy/build_linux_packages.sh#L180

antoniojkim:
> Do we have a plan for how to re-enable this or what the support plan is for LTC in Torch-MLIR?
Yes. There is another PyTorch error regarding symints that is preventing me from pushing a fix for this. I'm in touch with some of the PyTorch LTC folks to fix this. Once that is done, I'll put up a PR that fixes the error and re-enables LTC

antoniojkim:
As an FYI, this being the blocking issue
https://github.com/pytorch/pytorch/issues/84987

silvasean:
Can you please re-enable the tests as well?

antoniojkim:
@silvasean where is it disabled? I only see it disabled for the `mac-arm64` which I thought was disabled for a different reason.

silvasean:
This is the line that enables it on the CI: https://github.com/llvm/torch-mlir/blob/4ef6e69ed49f309aa70ea40f12bf7488cd8e3434/build_tools/python_deploy/build_linux_packages.sh#L243

PR #1373, opened 2022-09-15
Fix symint ops and blacklist `lift_fresh_copy`

antoniojkim:
Fix LTC issues caused by changes in upstream PyTorch
- A new `symint` key was added to the native functions yaml
- The `lift_fresh_copy` op was blacklisted in LTC as it is not yet supported

This PR depends on this upstream PR getting in first: https://github.com/pytorch/pytorch/pull/85029

Issue: https://github.com/llvm/torch-mlir/issues/1364


silvasean (approved):
Great to see this fixed!!

Issue #1366, opened 2022-09-14
[RFC] Bootstrapping the TCP Dialect in Torch-MLIR


sjarus:
We are announcing the effort to bootstrap the early development and technological demonstration of the TCP dialect ([public specification](https://docs.google.com/document/d/1Twyph8jU_f1QDoBInr8OkUXxcyRw-KqFdyclkK1UQbc/edit#heading=h.vrsunsna8bm), [technical discussion](https://docs.google.com/document/d/1f3KVsXA4xm6W7gd2cKx9ThGA52XMfZKsPV3ZOlsztC4/edit#heading=h.1ivbw5pi0o8e)) under [llvm-external-projects](https://github.com/llvm/torch-mlir/tree/main/externals/llvm-external-projects) within Torch-MLIR. 

The TCP dialect intends to offer strong support for backend compilation connectivity to LinAlg, Tensor and other dialects, compiler friendly support for dynamic shapes, broadcasting and other capabilities, and the Torch-MLIR pathway offers an opportunity to demonstrate these dialect capabilities to augment the existing backend paths.

The initial effort will focus on connectivity both into and out of the TCP dialect, emphasizing the technical value proposition, and we welcome input into the mechanics of the op construct and connectivity.

@navahgar @sanjoy @sjain-stanford @silvasean @jpienaar @stellaraccident

silvasean:
Overall I'm supportive, but I think the MLIR community needs to make a final decision here. I would recommend that we elaborate a bit on the details of the bringup plan, expected timelines, "unknowns", scoping, etc. and make a new RFC thread on MLIR discourse. I think that having a demo branch with Mlp2LayerModule_basic running e2e as prep work for the RFC would give folks some concrete code to rally around and really accelerate things.

Especially Chris's point about "what's *not* part of TCP" is really useful to define (at least as a draft to rally around).

stellaraccident:
Hey Sean, I obviously don't want to go against the community consensus here, but I do wonder if we are holding a bit of a different standard vs the rest of development that happens in torch-mlir (for which, it seems like this group is "the community"). For example, discussions about ONNX, MHLO connectivity, etc all happened here vs via LLVM side RFC.

But it also sounds like you would like to see some more definition, and I don't want to discount that and assume anything. I'm running behind from vacation and think I missed some back and forth. Would you mind deep linking to the question you would like to see answered?

[...]

PR #1223, opened 2022-08-14
[MHLO] Init end-to-end unit tests

fortianyou:
See RFC https://github.com/llvm/torch-mlir/issues/999

Co-authored-by: Bairen Yi yibairen.byron@bytedance.com
Co-authored-by: Jiawei Wu xremold@gmail.com
Co-authored-by: Tianyou Guo tianyou.gty@alibaba-inc.com
Co-authored-by: Xu Yan yancey.yx@alibaba-inc.com
Co-authored-by: Ziheng Jiang ziheng.jiang@bytedance.com

PR #1375, opened 2022-09-15
[Tcp] Add boilerplate for TCP dialect

navahgar:
This PR adds the initial boilerplate necessary for the TCP dialect. It includes:
- a dummy op in TCP. 
- a compilation flag, `-DTORCH_MLIR_DIALECTS_ENABLE_TCP` to enable TCP. This will be `OFF` by default.


externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -0,0 +1,65 @@
...
+  );
+
+  let results = (outs
+    Tcp_Tensor:$out
+  );
+
+  let assemblyFormat = "$in attr-dict `:` type($in) `->` type($out)";
+}
+
+def Tcp_AddOp : Tcp_BinaryOp<"add", [NoSideEffect]> {
```
silvasean, line 27: shape inference fns, infer type op interface.

cannot be marked NoSideEffect since it has UB (or do we want a separate Speculatable trait? IIRC this was broken in MLIR core). Sanjoy is the expert on this stuff.

externals/llvm-external-projects/torch-mlir-dialects/test/Conversion/TcpToTosa/binary_ops.mlir
```diff
@@ -0,0 +1,7 @@
+// RUN: torch-mlir-dialects-opt <%s -convert-tcp-to-tosa | FileCheck %s
+
+func.func @test_add(%arg0 : tensor<10x20xf32>, %arg1 : tensor<10x20xf32>) -> tensor<10x20xf32> {
+  // CHECK: %{{.*}} = "tosa.add"(%{{.*}}, %{{.*}}) : (tensor<10x20xf32>, tensor<10x20xf32>) -> tensor<10x20xf32>
+  %0 = tcp.add %arg0, %arg1 : tensor<10x20xf32>, tensor<10x20xf32> -> tensor<10x20xf32>
+  return %0 : tensor<10x20xf32>
```
silvasean, line 6: include linalg lowerings too so we can exercise the dynamic shape design

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sanjoy: Yes, this cannot be `NoSideEffect` so I'd remove it from now.  We'll eventually have to figure out a more nuanced way to express this trait.

sjain-stanford:
Changed base branch from `main` to `mlir-tcp` based on @stellaraccident 's [suggestion](https://github.com/llvm/torch-mlir/issues/1366#issuecomment-1248786577).

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/Passes.td
```diff
@@ -0,0 +1,25 @@
...
+#define TORCH_MLIR_DIALECTS_CONVERSION_PASSES
+
+include "mlir/Pass/PassBase.td"
+
+#ifdef TORCH_MLIR_DIALECTS_ENABLE_TCP
+//===----------------------------------------------------------------------===//
+// TcpToTosa
+//===----------------------------------------------------------------------===//
+
+def ConvertTcpToTosa
```
sanjoy, line 11: This pass is mainly here to help bootstrap right (and it will go away at some point)?

If yes, please add some commentary here to that effect.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/Passes.td
```diff
@@ -0,0 +1,25 @@
...
+      : Pass<"convert-tcp-to-tosa", "func::FuncOp"> {
+  let summary = "Lower TCP to TOSA";
+  let description = [{
+    Pass that converts TCP operations to the equivalent operations in TOSA.
+  }];
+
+  let constructor = "mlir::torch::tcp::createConvertTcpToTosa()";
```
sanjoy, line 18: The convention seems to be `createConvertTcpToTosaPass`?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
```diff
@@ -0,0 +1,29 @@
...
+  let name = "tcp";
+  let cppNamespace = "::mlir::torch::tcp";
+  let description = [{
+    Tensor Compute Primitives (TCP) dialect.
+
+    TCP is a mid-level dialect in MLIR that enables transformations that
+    complement those in existing dialects, by supporting ops at a
+    high-level of abstraction.
+
+    Status: This is being boostrapped under Torch-MLIR Dialects.
```
sanjoy, line 16: Is this necessary here?  This is more about the roadmap than the technical content of the dialect right?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -0,0 +1,65 @@
...
+
+def Tcp_TanhOp : Tcp_UnaryOp<"tanh", [NoSideEffect]> {
+  let summary = "Computes tanh of input, elementwise";
+
+  let description = [{
+    Elementwise unary op to compute tanh.
+  }];
+
+  let arguments = (ins
+    Tcp_Tensor:$in
```
sanjoy, line 17: This looks too loose -- tanh should only accept floating point tensors right?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sanjoy, line 27: As @silvasean said, binary ops should not be `NoSideEffect` since they have UB on mismatching shapes.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpInterfaces.td
```diff
@@ -0,0 +1,30 @@
...
+  let description = [{
+    Implemented by all binary ops in TCP.
+  }];
+}
+
+def TcpTernaryOpInterface : OpInterface<"TcpTernaryOpInterface"> {
+  let description = [{
+    Implemented by all ternary ops in TCP.
+  }];
+}
```
sanjoy, line 28: Couple of points:

1. Please document the broadcasting semantics here.
2. The description itself is circular -- IMO this is a good place to define what "binary op" means.
3. Should this op interface define operations like `lhs` and `rhs` for binary ops?
4. Should this be a trait instead?  If not, I suggest adding a oneliner stating why not.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -0,0 +1,65 @@
...
+  );
+
+  let assemblyFormat = "$in1 `,` $in2 attr-dict `:` type($in1) `,` type($in2) `->` type($out)";
+}
+
+def Tcp_MatmulOp : Tcp_Op<"matmul", [NoSideEffect]> {
+  let summary = "Computes 2D / 3D matrix multiplication";
+
+  let description = [{
+    Matrix multiplication of two 2D / 3D tensors.
```
sanjoy, line 50: This is a good place to define the semantics of this op in more detail (enough detail for someone to implement a reference interpreter).

Matmul semantics are probably obvious, but this op will set the tone for other less obvious ops that come after it.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToTosa/TcpToTosa.cpp
```diff
@@ -0,0 +1,113 @@
...
+class ConvertElementwiseTcpOpToTosaOp : public OpConversionPattern<TcpOpT> {
+public:
+  using OpConversionPattern<TcpOpT>::OpConversionPattern;
+  using OpAdaptor = typename TcpOpT::Adaptor;
+  LogicalResult matchAndRewrite(
+            TcpOpT op,
+            OpAdaptor adaptor,
+            ConversionPatternRewriter& rewriter) const override {
+    llvm::SmallVector<mlir::Type> newTypes;
+    for (auto res : op->getResultTypes()) {
```
sanjoy, line 36: Unnecessary `{`.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToTosa/TcpToTosa.cpp
```diff
@@ -0,0 +1,113 @@
...
+    typeConverter.addConversion([](TensorType tensorType) {
+      // TODO: Do we need to convert this to tosa.TensorType?
+      return tensorType;
+    });
+
+    patterns.add<ConvertElementwiseTcpOpToTosaOp<TanhOp, tosa::TanhOp>>(typeConverter, context);
+    patterns.add<ConvertElementwiseTcpOpToTosaOp<AddOp, tosa::AddOp>>(typeConverter, context);
+    patterns.add<ConvertMatmulTcpOpToTosaOp<MatmulOp, tosa::MatMulOp>>(typeConverter, context);
+
+    if (failed(applyPartialConversion(getOperation(), target, std::move(patterns)))) {
```
sanjoy, line 99: Unnecessary `{`?  Same elsewhere.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToTosa/TcpToTosa.cpp
```diff
@@ -0,0 +1,113 @@
...
+};
+
+class ConvertTcpToTosa : public ConvertTcpToTosaBase<ConvertTcpToTosa> {
+
+public:
+  void runOnOperation() override {
+    MLIRContext* context = &getContext();
+    RewritePatternSet patterns(context);
+    ConversionTarget target(*context);
+    target.addIllegalDialect<TcpDialect>();
```
sanjoy, line 77: This will eventually have to go right since TCP won't fully convert to TOSA?  If yes, I suggest adding a comment here stating that.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToTosa/TcpToTosa.cpp
```diff
@@ -0,0 +1,113 @@
...
+  LogicalResult matchAndRewrite(
+            TcpOpT op,
+            OpAdaptor adaptor,
+            ConversionPatternRewriter& rewriter) const override {
+    auto lhsTy = op.getLhs().getType().template cast<RankedTensorType>();
+    auto rhsTy = op.getRhs().getType().template cast<RankedTensorType>();
+    // TODO: Support 2D inputs.
+    if (lhsTy.getRank() != 3 || rhsTy.getRank() != 3) {
+      return op.emitError("tcp.matmul with inputs of rank != 3 cannot be legalized");
+    }
```
sanjoy, line 59: Should this be caught by the IR verifier and just `assert`ed here?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -0,0 +1,65 @@
...
+  }];
+
+  let arguments = (ins
+    Tcp_Tensor2Dor3D:$lhs,
+    Tcp_Tensor2Dor3D:$rhs
```
sanjoy, line 55: Please add a comment here or in the `description` on why we're restricting to rank 2 and 3 only.

sanjoy (changes requested):

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
silvasean: It is honestly unclear to me if we want to support `(2D,3D)` mixed cases or require both to be the same rank. Or something more like dot_general

silvasean:
Posting this here for persistence: https://github.com/llvm/torch-mlir/pull/1223 -- here is an example of adding the e2e framework for mhlo, we should do something similar for TCP. See rationale here for the critical importance of e2e testing in this space: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#why-so-much-end-to-end-testing

sjarus:
nit: I recommend renaming the commit prefix to remote the all caps which looks a little "shouty". Just [tcp] as a required prefix should be sufficient to identify every push.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
burmako: From the spec, I remember that complex element types were also on the menu, so that would be floating point and complex tensors?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -0,0 +1,65 @@
...
+  let description = [{
+    Elementwise unary op to compute tanh.
+  }];
+
+  let arguments = (ins
+    Tcp_Tensor:$in
+  );
+
+  let results = (outs
+    Tcp_Tensor:$out
```
burmako, line 21: What is the relationship between operands and results of unary and binary ops? Do we want them to be exactly the same or just compatible, for some definition of "compatible"?

In MHLO, we allow "compatible" operand and result types via [isCompatibleReturnTypes](https://github.com/tensorflow/mlir-hlo/blob/00beaaeb7909637b60864addda55506696bc964d/include/mlir-hlo/Dialect/mhlo/IR/hlo_ops.td#L146-L154) and [HLO_CompatibleOperandsAndResultType](https://github.com/openxla/stablehlo/blob/31b8d00a7b37716e9e0d1a780301b52d4b78abf1/stablehlo/dialect/Base.td#L185) which are both implemented via [isCompatibleForHloTypeInference](https://github.com/openxla/stablehlo/blob/31b8d00a7b37716e9e0d1a780301b52d4b78abf1/stablehlo/dialect/Base.cpp#L67-L82), which roughly speaking allows things like:

```
func @dynamism(%arg0: tensor<?xf32>, %arg1: tensor<1xf32>) {
  %0 = "mhlo.add"(%arg0, %arg0) : (tensor<?xf32>, tensor<?xf32>) -> tensor<?xf32>
  %1 = "mhlo.add"(%arg0, %arg0) : (tensor<?xf32>, tensor<?xf32>) -> tensor<1xf32>
  %2 = "mhlo.add"(%arg0, %arg1) : (tensor<?xf32>, tensor<1xf32>) -> tensor<?xf32>
  %3 = "mhlo.add"(%arg0, %arg1) : (tensor<?xf32>, tensor<1xf32>) -> tensor<1xf32>
  %4 = "mhlo.add"(%arg1, %arg0) : (tensor<1xf32>, tensor<?xf32>) -> tensor<?xf32>
  %5 = "mhlo.add"(%arg1, %arg0) : (tensor<1xf32>, tensor<?xf32>) -> tensor<1xf32>
  %6 = "mhlo.add"(%arg1, %arg1) : (tensor<1xf32>, tensor<1xf32>) -> tensor<?xf32>
  %7 = "mhlo.add"(%arg1, %arg1) : (tensor<1xf32>, tensor<1xf32>) -> tensor<1xf32>
  return
}
```

I wasn't around when this functionality was added to MHLO, but I've been told that it was motivated by similar functionality in the TF dialect, which itself was motivated by the desire to support progressive shape refinement during `--tf-shape-inference`. If MHLO didn't support these relaxed compatibility checks, then `--tf-shape-inference` would need to introduce casts when doing shape refinement, which was deemed undesirable.

Are we planning to do shape refinement in TCP, or that's expected to be the job of higher-level layers? If it's the former (which I think could be the right choice), then does anyone have a specific shape inference facility in mind that we could use? In the recent ODM, I remember someone said that there are better tools than `--tf-shape-inference` (which also isn't available in upstream, so we cannot use it anyway in TCP), and I'm eager to learn more.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToTosa/TcpToTosa.cpp
burmako: If something is checked by the verifier, I think we shouldn't even assert on that everywhere else to simplify code.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -0,0 +1,65 @@
...
+  );
+
+  let assemblyFormat = "$in1 `,` $in2 attr-dict `:` type($in1) `,` type($in2) `->` type($out)";
+}
+
+def Tcp_MatmulOp : Tcp_Op<"matmul", [NoSideEffect]> {
```
burmako, line 46: I think it would be good to implement verification + type inference for this op right away. This will force us to face interesting questions, in addition to the 2D/3D item which is already discussed. E.g.:
  1) Do we allow result element type to be different from operand element type? I think we should, if we want to support quantization well.
  2) If yes, what do we do with type inference? Do we: a) say that matmul doesn't support it, b) add something like `preferred_element_type` attribute to enable type inference, c) infer result element type equal to operand element type and then work around via isCompatibleReturnTypes?

externals/llvm-external-projects/torch-mlir-dialects/test/Conversion/TcpToTosa/binary_ops.mlir
burmako: +1

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Good point. Removed.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/Passes.td
navahgar: That was the initial plan. But after the recent discussions, it felt like this pass might be something the community might find useful. We can probably take a call on this later. I have left a comment mentioning that this will be supported in the longer term if there is a usecase.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/Passes.td
navahgar: Done.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
navahgar: Removed.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Good point regarding float and complex. Updated.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToTosa/TcpToTosa.cpp
navahgar: I didn't realized this was the convention being followed in MLIR. I saw both usages actually, so, not quite sure. Anyways, will follow this from now. Thanks for pointing that out.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToTosa/TcpToTosa.cpp
navahgar: Yes, thats correct. Actually, I don't see a need for adding this, when we add the individual ops as legal and illegal below. So, removing this.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToTosa/TcpToTosa.cpp
navahgar: This op actually accepts 2D and 3D tensors. The 2D tensor case is not handled yet, because that requires reshaping into 3D (since Tosa only accepts 3D tensors). The IR verifier will allow only 2D and 3D tensors. This is a temporary check to error out for the 2D case. This will be removed once we support that case.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Added

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
asaadaldien: IMO because something more general like `torch.tensordot` or `tf.tensordot` can be mapped to `tcp.matmul` with sequence of operand transposes & reshapes, and the bottom consumers transform for backends (codegen or lib-calls) generally prefers lower ranks matmuls/batch_matmul contracts `e.g match(AB), match(A^TB), match(AB^T)..etc`. In this case given that how TCP is positioned in the stack a lower rank fixed contract `reduce_add(A[i,m,k] * B[i,k,n], k)` make sense.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/Passes.td
sanjoy: Makes sense.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
```diff
@@ -0,0 +1,20 @@
...
+
+namespace mlir {
+
+#define GEN_PASS_DECL_CONVERTTCPTOTOSA
+#include "torch-mlir-dialects/Conversion/Passes.h.inc"
+
+namespace torch {
+namespace tcp {
+
+std::unique_ptr<Pass> createConvertTcpToTosaPass();
```
sanjoy, line 14: This is auto-generated right?  I'm looking here: https://github.com/llvm/llvm-project/blob/e854c17b02f8cd82a303d223ba5f3b0d87579cd7/mlir/tools/mlir-tblgen/PassGen.cpp#L127

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpDialect.h
```diff
@@ -0,0 +1,9 @@
+#ifndef TORCH_MLIR_DIALECTS_DIALECT_TCP_IR_TCPDIALECT_H_
+#define TORCH_MLIR_DIALECTS_DIALECT_TCP_IR_TCPDIALECT_H_
+
+#include "mlir/IR/Dialect.h"
+#include "mlir/IR/OpDefinition.h"
+
+#include "torch-mlir-dialects/Dialect/Tcp/IR/TcpDialect.h.inc"
+
+#endif // TORCH_MLIR_DIALECTS_DIALECT_TCP_IR_TCPDIALECT_H_
```
sanjoy, line 9: Nit: Github flags these files as not ending with a newline, not sure if that's a problem.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
```diff
@@ -0,0 +1,30 @@
...
+include "mlir/IR/OpBase.td"
+
+def Tcp_Dialect : Dialect {
+  let name = "tcp";
+  let cppNamespace = "::mlir::torch::tcp";
+  let description = [{
+    Tensor Compute Primitives (TCP) dialect.
+
+    TCP is a mid-level dialect in MLIR that enables transformations that
+    complement those in existing dialects, by supporting ops at a
```
sanjoy, line 13: This isn't complete IMO -- we also supports ops at a lower level of abstraction than, say, the Torch dialect.  This also doesn't mention anything about the domain (ML & ML adjacent domains).

My 2c is to keep the description here bold and brief:

> TCP is a mid-level transformation oriented IR for deep learning & similar applications.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
```diff
@@ -0,0 +1,30 @@
...
+    high-level of abstraction.
+  }];
+
+  let dependentDialects = [];
+}
+
+def Tcp_Scalar : AnyTypeOf<[AnyFloat, AnySignlessIntegerOrIndex, AnyComplex]>;
```
sanjoy, line 20: This will likely become clear as we design more ops, but IMO we should keep the `index` type distinct from other non-index types.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -0,0 +1,70 @@
...
+  let summary = "Computes 2D / 3D matrix multiplication";
+
+  let description = [{
+    Matrix multiplication of two 2D / 3D tensors.
+
+    This op supports the base case for matrix multiplication with 2D tensors
+    and stacked matrix multiplication with 3D tensors. All other cases are
```
sanjoy, line 53: By "stacked matrix multiplication with 3D tensors" do you mean a batch GEMM?

sanjoy (changes requested):

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
silvasean: There are a variety of ways to perform shape inference. Linalg does a very good job despite the "exact match of the static type" requirement by having patterns that push the casts around to a fixed-point, which typically results in them being absorbed somewhere.

I wrote some more general shape inference thoughts here, though I think a lot of what is described there doesn't apply to TCP/MHLO-like design points [link](https://discourse.llvm.org/t/rfc-symbolic-shape-analysis/62879/17?u=_sean_silva).

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
silvasean: +1,

Also, why not have separate matmul and batch_matmul ops? My experience is that with these "variant" ops it's much easier to `isa<BatchMatmulOp>` than `if (getOperand(0).cast<RankedTensorType>().getRank() == 3)`

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
silvasean: +1. The one exception where `tensor<index>` should exist is for "gather/scatter indices". We definitely don't want to ever use `tensor<?xindex>` as a "shape" which happens in MHLO and is the source of much brittleness.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Yes, that's what I meant.

sjain-stanford:
In case anyone is wondering, I switched base branches to main and back (to mlir-tcp) to flush the extraneous commit history that was showing earlier, likely due to a rebase on main (while base branch is mlir-tcp). When rebasing, we may want to do mlir-tcp on main first, then this branch tcp_1 on mlir-tcp, which should do it cleanly hopefully.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
navahgar: I tried removing this and it doesn't auto generate for me. Not sure if I'm missing something in cmake files to do this.

externals/llvm-external-projects/torch-mlir-dialects/test/Conversion/TcpToTosa/binary_ops.mlir
navahgar: I'm working on a follow up PR with Linalg lowerings.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpInterfaces.td
navahgar: Good point regarding trait vs interface. I'll spend some more time trying to figure out what works best for our usecase here. For now, I'm removing all interfaces.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Good point regarding the mixed 2D-3D case. In the form I had, that case is not allowed and we will have to error out.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: I'm working on a document to summarize the different design choices for Matmul (2 ops vs 1 op, 3D and 2D cases, etc.). So, I'm removing it from this PR. Will send a separate PR for matmul later.

navahgar: > why not have separate matmul and batch_matmul ops?

I'm working on a document to summarize the different design choices for Matmul (2 ops vs 1 op, 3D and 2D cases, etc.). So, I'm removing matmul op from this PR. Will send a separate PR for matmul later.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpDialect.h
navahgar: I'm not sure either. Added newline in all files anyways.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
navahgar: Done.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
navahgar: Good points regarding `index`. However, in addition to gather / scatter, we should probably allow some binary operations on index types, for example, max, min. Wdyt?

Updated this PR to not include `index` types.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Thanks for those points. That is very useful to know.

Will address shape inference in a separate PR.

navahgar (commented):
Thanks for the review. Addressed the comments. PTAL.

I'm moving some of the stuff discussed here to follow up PRs:
  * Linalg lowering
  * e2e tests
  * matmul op
  * shape inference
 
Thanks for pointers regarding those.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToTosa/TcpToTosa.cpp
navahgar: Removed the matmul op from the PR. Working on a document summarizing the design choices. Will be sending a separate PR for that later.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
sanjoy: > we should probably allow some binary operations on index types, for example, max, min

Yes, but I'm not convinced that we need these operations on tensors containing index types.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
sanjoy: Can you check if other passes have this?  Maybe you need to do something extra to put it in the `torch::tcp` namespace?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -0,0 +1,46 @@
+#ifndef TORCH_MLIR_DIALECT_TCP_OPS
+#define TORCH_MLIR_DIALECT_TCP_OPS
+
+include "torch-mlir-dialects/Dialect/Tcp/IR/TcpOpsBase.td"
+
+include "mlir/IR/OpBase.td"
+include "mlir/Interfaces/SideEffectInterfaces.td"
+
+def Tcp_TanhOp : Tcp_UnaryOp<"tanh", [NoSideEffect]> {
```
sanjoy, line 9: Should the `NoSideEffect` be factored into `Tcp_UnaryOp` somehow?

sanjoy (approved):
Minor nits inline, LGTM otherwise.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
navahgar: I think what you pointed to is only generating code that goes into `Passes.h.inc` file. For example, that file contains the following:
```
inline void registerConvertTcpToTosa() {
  ::mlir::registerPass([]() -> std::unique_ptr<::mlir::Pass> {
    return mlir::torch::tcp::createConvertTcpToTosaPass();
  });
}
```
I think that is what is being generated here.

navahgar: @silvasean Do you know if this file could be generated automatically?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
navahgar: There was a comment in one of our docs that seemed to be talking about a use-case for this. I might be mistaken though. 

Anyways, I have removed it in this PR. We can table this discussion until we have a real use-case with index type beyond Gather / Scatter.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Thats a good idea. Updated that. Also added `Elementwise` to both `Tcp_UnaryOp` and `Tcp_BinaryOp`.

externals/llvm-external-projects/torch-mlir-dialects/test/Conversion/TcpToTosa/binary_ops.mlir
silvasean: The initial PR should be a "scale model" for what we want the system to grow into, so we should include the linalg lowerings from the beginning.

silvasean: This also includes wiring up the e2e tests.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/Passes.h
```diff
@@ -0,0 +1,13 @@
+#ifndef TORCH_MLIR_DIALECTS_CONVERSION_PASSES_H
```
silvasean, line 1: all these files need license headers

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
silvasean: createConvertTcpToTosaPass needs to be declared here. It is not autogenerated. See e.g. https://github.com/llvm/torch-mlir/blob/main/include/torch-mlir/Conversion/TorchToArith/TorchToArith.h

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
```diff
@@ -0,0 +1,20 @@
+#ifndef TORCH_MLIR_DIALECTS_CONVERSION_TCPTOTOSA_TCPTOTOSA_H_
+#define TORCH_MLIR_DIALECTS_CONVERSION_TCPTOTOSA_TCPTOTOSA_H_
+
+#include "mlir/Pass/Pass.h"
+
+namespace mlir {
+
+#define GEN_PASS_DECL_CONVERTTCPTOTOSA
```
silvasean, line 8: Why do we need this GEN_PASS_DECL_CONVERTTCPTOTOSA? I don't think we do this in the main Torch-MLIR codebase. I think we just have a single PassDetail.h which only need to be included by the .cpp files.

silvasean (changes requested):

externals/llvm-external-projects/torch-mlir-dialects/test/Conversion/TcpToTosa/binary_ops.mlir
silvasean: Or we can submit this with no ops (or just a dummy op (literally call it DummyOp)) to get the build system boilerplate and then we can evaluate new ops one by one.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
navahgar: Aah okay. Thanks for pointing that out.

externals/llvm-external-projects/torch-mlir-dialects/test/Conversion/TcpToTosa/binary_ops.mlir
navahgar: > The initial PR should be a "scale model" for what we want the system to grow into, so we should include the linalg lowerings from the beginning.

Okay, fair enough.

I have removed all the ops in this PR and added only a dummy op to get the boilerplate code in first.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/Passes.h
navahgar: Added

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
burmako: @silvasean `GEN_PASS_CLASSES` [has been recently deprecated](https://discourse.llvm.org/t/psa-deprecation-warning-per-pass-autogenerated-macros-and-pass-options/64998), and will be removed in the near future. `GEN_PASS_DECL_PASSNAME` is the recommended way to approach this moving forward. E.g. see [this MLIR-HLO commit](https://github.com/tensorflow/mlir-hlo/commit/7596fd322d2408bf88d4c0e2748d95a3f760aefe) for an example.

By the way, functions like `createConvertTcpToTosaPass` can be autogenerated, but only if you don't specify `let constructor` in Passes.td, e.g. see [this StableHLO PR](https://github.com/openxla/stablehlo/pull/176) for an example.

externals/llvm-external-projects/torch-mlir-dialects/README.md
```diff
@@ -9,3 +9,15 @@ This project is intended to be used via LLVM's external projects setup:
...
+
+## TCP Dialect
+Tensor Compute Primitives (TCP) aims to be a mid-level dialect in MLIR that enables transformations that complement those in existing dialects, by supporting ops at a high-level of abstraction.
+
+TCP is being bootstrapped under Torch-MLIR Dialects.
+
+All technical discussions regarding TCP will be done in the discourse [TCP-WG](https://discourse.llvm.org/c/mlir/mlir-tcp-wg/36) category.
+### References
+* Discourse [thread](https://discourse.llvm.org/t/rfc-proposal-for-a-high-level-ml-dialect-in-mlir/64249) regarding TCP proposal.
+* TCP [proposal document](https://docs.google.com/document/d/1f3KVsXA4xm6W7gd2cKx9ThGA52XMfZKsPV3ZOlsztC4/edit?usp=sharing)
```
burmako, line 21: Drop `?usp=sharing` from the link? I don't think it's essential. (Same for the spec draft link below).

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
```diff
@@ -0,0 +1,43 @@
...
+  }];
+
+  let dependentDialects = [];
+}
+
+//===----------------------------------------------------------------------===//
+// Tcp Type Definitions.
+//===----------------------------------------------------------------------===//
+
+def Tcp_Scalar : AnyTypeOf<[AnyFloat, AnySignlessInteger, AnyComplex]>;
```
burmako, line 32: The spec draft says `AnySignlessIntegerOrIndex`. I think `AnySignlessInteger` is a more recent development, so the spec needs to be updated?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
```diff
@@ -0,0 +1,43 @@
...
+include "mlir/IR/OpBase.td"
+
+def Tcp_Dialect : Dialect {
+  let name = "tcp";
+  let cppNamespace = "::mlir::torch::tcp";
+  let description = [{
+    Tensor Compute Primitives (TCP) dialect.
+
+    TCP is a mid-level transformation oriented IR for deep learning & similar
+    applications.
```
burmako, line 22: Consider synchronizing this text with the definition of TCP in `externals/llvm-external-projects/torch-mlir-dialects/README.md`.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
```diff
@@ -0,0 +1,43 @@
...
+//===----------------------------------------------------------------------===//
+
+#ifndef TORCH_MLIR_DIALECT_TCP_BASE
+#define TORCH_MLIR_DIALECT_TCP_BASE
+
+include "mlir/IR/OpBase.td"
+
+def Tcp_Dialect : Dialect {
+  let name = "tcp";
+  let cppNamespace = "::mlir::torch::tcp";
```
burmako, line 17: `::mlir::tcp` perhaps, given the plans to be applicable more widely than to PyTorch?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
```diff
@@ -0,0 +1,43 @@
...
+
+//===----------------------------------------------------------------------===//
+// Tcp Type Definitions.
+//===----------------------------------------------------------------------===//
+
+def Tcp_Scalar : AnyTypeOf<[AnyFloat, AnySignlessInteger, AnyComplex]>;
+def Tcp_Tensor : RankedTensorOf<[Tcp_Scalar]>;
+
+//===----------------------------------------------------------------------===//
+// Tcp Operator.
```
burmako, line 36: "Tcp Operators"?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.h
```diff
@@ -0,0 +1,25 @@
...
+#define TORCH_MLIR_DIALECTS_DIALECT_TCP_IR_TCPOPS_H_
+
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/Dialect.h"
+#include "mlir/IR/OpDefinition.h"
+#include "mlir/Interfaces/ControlFlowInterfaces.h"
+#include "mlir/Interfaces/SideEffectInterfaces.h"
+
+#include "torch-mlir-dialects/Dialect/Tcp/IR/TcpInterfaces.h.inc"
```
burmako, line 20: Where is this file coming from?

burmako (approved):
Reviewed prose and .td files. Sean will undoubtedly have a better perspective on the conventions for .h/CMakeLists.txt files. At a glance, everything checks out.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
navahgar: Given that we are bootstrapping TCP under TorchMLIR externals, I assumed it has to use the `::mlir::torch::tcp` namespace. 

@silvasean Is it okay to use `::mlir::tcp` for this instead?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
silvasean: Oh neat. I wasn't aware of that.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
silvasean: Given the plans to make it more widely applicable, mlir::tcp seems fine.

externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/dummy_op.mlir
```diff
@@ -0,0 +1,8 @@
+// RUN: torch-mlir-dialects-opt %s | torch-mlir-dialects-opt | FileCheck %s
```
silvasean, line 1: nit: file name should be dummy-op.mlir for consistency.

externals/llvm-external-projects/torch-mlir-dialects/README.md
navahgar: Removed

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
navahgar: Done

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
navahgar: Thanks for clarifying. Updated it to `mlir::tcp`

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
navahgar: Updated it to "Tcp Ops Base", which is more appropriate here.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.h
navahgar: This is an artifact of some code I had earlier. Removed this. Thanks for catching that.

externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/dummy_op.mlir
navahgar: Done

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Conversion/TcpToTosa/TcpToTosa.h
navahgar: > By the way, functions like createConvertTcpToTosaPass can be autogenerated, but only if you don't specify let constructor in Passes.td, e.g. see https://github.com/openxla/stablehlo/pull/176 for an example.

Good to know that. Thanks.

burmako:
Super pumped to see this PR landing!! 🎉 🎉 🎉

PR #1376, opened 2022-09-16
Miscellaneous fixes for Windows builds

ashay:
This patch contains multiple commits that make it possible to build and test Torch-MLIR on Windows.  See commit messages for descriptions.

1. test: allow spaces in path to Python executable
2. python: remove header file that causes Windows build failures
3. python: drop `TORCH_API` from function defined in Torch-MLIR
4. python: make output of class anotations deterministic
5. test: use Python3_EXECUTABLE as interpreter path for consistency
6. test: fix RUN string
7. python: port parallel test framework to Windows

### DIFF
--- a/python/torch_mlir_e2e_test/framework.py
+++ b/python/torch_mlir_e2e_test/framework.py
@@ -22,12 +22,13 @@ compiling or TorchScript'ing).
 
 import abc
 from typing import Any, Callable, List, NamedTuple, Optional, TypeVar, Union, Dict
+from itertools import repeat
 
 import sys
 import traceback
 
 import torch
-import torch.multiprocessing as mp
+import multiprocess as mp
 
 TorchScriptValue = Union[int, float, List['TorchScriptValue'],
                          Dict['TorchScriptValue',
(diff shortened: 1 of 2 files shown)


powderluv:
woah. thank you. I had to reboot that Windows machine to test some linux driver stuff on that GPU so it will be a few days before I get back to Windows. Thank for sharing.

test/Conversion/TorchToMhlo/basic.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: torch-mlir-opt <%s -convert-torch-to-mhlo -split-input-file -verify-diagnostics | FileCheck %s
+// RUN: torch-mlir-opt%exeext <%s -convert-torch-to-mhlo -split-input-file -verify-diagnostics | FileCheck %s
```
sstamenova, line 1: This should not be necessary, lit supports having this "tool" selected correctly on all platforms. Have a look at how `mlir-opt` is handled, for example.

test/python/importer/jit_ir/ivalue_import/submodules.py
```diff
@@ -7,7 +7,7 @@
 import torch
 from torch_mlir.dialects.torch.importer.jit_ir import ModuleBuilder
 
-# RUN: %PYTHON %s | torch-mlir-opt | FileCheck %s
+# RUN: %PYTHON %s | torch-mlir-opt%exeext | FileCheck %s
```
sstamenova, line 10: The python path might need some extra handling in the lit files. I am not sure if it's configured correctly, but in the past we've run into issues where the default configured path does not support installations that contain spaces (which happens on Windows). Have a look at how this is handled for llvm/lldb/clang (I think mlir is not correct)

powderluv (approved):
Thank you for this. Happy to test it out once it lands.

powderluv:
cool Thanks for the heads up. Will wait for the monday roll.

powderluv:
should we add a Windows CI next ?

sstamenova:
> should we add a Windows CI next ?

I think that would be great to make sure we don't accidentally regress after all the work that was put into getting windows to work.

PR #1378, opened 2022-09-16
[torch-mlir][Tosa] fix during torch.max.dim lower to tosa the reshape…

lipracer:
…'s new shape attr mismatch reshape's result type


powderluv:
Thank you for the contribution

sjarus:
Thanks for this fix!

sjarus (approved):

PR #1380, opened 2022-09-18
Disable libzstd detection to fix M1 builds

powderluv:
Previously we `sudo rm -f` the non-universal zstd installed in the GHA. The CI has this fix but it doesn't take effect in the Release builds without this change.


makslevental (approved):
LGTM

PR #1387, opened 2022-09-20
[MLIR][TORCH] Add a test for sum.dim_IntList op working for tosa

vivekkhandelwal1:


vivekkhandelwal1:
I have added this test case so that we know that the TorchToTosa lowering for the `aten.sum.dim_IntList` op is working for some cases.

sjarus (approved):
Thanks for adding this test!

Issue #1374, opened 2022-09-15
Packages can't be fetched from releases page automatically


silvasean:
It appears that Github changed their HTML for the [releases page](https://github.com/llvm/torch-mlir/releases) so that it loads the links via javascript.

This means that `pip install -f https://github.com/llvm/torch-mlir/releases` no longer sees any package links on that page, which makes it fail to install `torch-mlir`.

IREE has a similar packaging system, so roping in @GMNGeoffrey @stellaraccident since any solution there is probably shareable.

ramiro050:
Github, as of yesterday, seems to have gone back to generating a `releases` page with all the links in the html, so this should no longer be an issue. As far as I know, this issue did not affect IREE's releases page.

It seems that this was affecting some repositories but not all of them. See: https://github.com/Homebrew/brew/issues/13853

silvasean:
Thanks Ramiro, let's close this then! @dellis23 you should be able to go back to using the releases page for the install in your colab then.

ramiro050:
> I just had to reopen [Homebrew/brew#13853](https://github.com/Homebrew/brew/issues/13853) due to this issue reappearing. Not sure if you all are affected at the moment, but it's something to be cognizant of.

Ha! I just tried it again right now and it is once again not working. Thanks for the heads-up

powderluv:
We did a temporary workaround in the SHARK repo to find the expanded_assets link:

https://github.com/nod-ai/SHARK/blob/main/setup_venv.sh#L74-L79

Anyway to message Github about it?

silvasean:
@asl does the LLVM Foundation have a way for us to message github about this issue?

asl:
@silvasean Unfortunately, no :(

ramiro050:
@GMNGeoffrey, this is now affecting IREE as well:

```shell
$ pip install iree-compiler --find-links https://github.com/iree-org/iree/releases --no-index
Looking in links: https://github.com/iree-org/iree/releases
ERROR: Could not find a version that satisfies the requirement iree-compiler (from versions: none)
ERROR: No matching distribution found for iree-compiler
```

The HTML for the `releases` page no longer has download links in it.

stellaraccident:
We need to acquire/craft a script that generates a pip index file and then publish that. It is probably no more than ~a hundred lines of code. But I don't just have it handy...

stellaraccident:
Index specification: https://peps.python.org/pep-0503/

[...]

PR #1389, opened 2022-09-20
Create github action for creating pip-compatible releases page

dellis23:
This runs a github actions cron job every hour to scrape the releases API and generate a static releases page inside the `github-pages` branch.  This page is available at https://llvm.github.io/torch-mlir/ and can be used as a package index for pip.

See https://github.com/llvm/torch-mlir/issues/1374 for background.

### DIFF
--- a/.github/workflows/gh-pages-releases.yml
+++ b/.github/workflows/gh-pages-releases.yml
@@ -1,0 +1,40 @@
+# See: https://github.com/llvm/torch-mlir/issues/1374
+name: Publish releases page
+
+on:
+  schedule:
+    - cron: '0 * * * *'
+
+  workflow_dispatch:
+
+jobs:
+  scrape_and_publish_releases:
+    name: "Scrape and publish releases"
+    runs-on: ubuntu-20.04
+
+    # Don't run this in everyone's forks.
+    if: github.repository == 'llvm/torch-mlir'
+
+    steps:
+      - name: Checking out repository
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.WORKFLOW_INVOCATION_TOKEN }}
+      - name: Run scrape releases script
+        run: python ./build_tools/scrape_releases.py llvm torch-mlir > /tmp/index.html
+        shell: bash
+      - run: git fetch --all
+      - run: git switch github-pages
+      - run: git config --global user.email "none@none.com"
+      - run: git config --global user.name "torch-mlir"
+      - run: mv /tmp/index.html package-index/index.html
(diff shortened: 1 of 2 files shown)


.github/workflows/gh-pages-releases.yml
```diff
@@ -0,0 +1,40 @@
...
+    # Don't run this in everyone's forks.
+    if: github.repository == 'llvm/torch-mlir'
+
+    steps:
+      - name: Checking out repository
+        uses: actions/checkout@v2
+        with:
+          token: ${{ secrets.WORKFLOW_INVOCATION_TOKEN }}
+      - name: Run scrape releases script
+        run: python ./tools/scrape_releases.py llvm torch-mlir > /tmp/index.html
```
silvasean, line 24: nit: let's put this in build_tools since it is a "meta" tool and "not part of the compiler"

silvasean (approved):
Great! Nice to see that this is so minimal :)

.github/workflows/gh-pages-releases.yml
```diff
@@ -0,0 +1,40 @@
...
+        shell: bash
+      - run: git fetch --all
+      - run: git switch github-pages
+      - run: git config --global user.email "none@none.com"
+      - run: git config --global user.name "torch-mlir"
+      - run: mv /tmp/index.html index.html
```
silvasean, line 30: nit: "index.html" is kind of like a special "main" IIRC for websites ([link](https://www.pcmag.com/encyclopedia/term/indexhtml) found through quick googling). Let's use a slightly more verbose name `package_index.html` to dodge that special name.

.github/workflows/gh-pages-releases.yml
dellis23: Done.

.github/workflows/gh-pages-releases.yml
dellis23: I'm suddenly feeling very old remembering when the transition happened to no longer having html extensions in URLs and realizing that was a while ago haha.  I've moved everything to `package-index` in an index file there so we still get the nice URL.  (Should be visible at https://llvm.github.io/torch-mlir/package-index/)

silvasean: nice :)

minor nit, can you add some newlines/<br>'s so that when opening in the browser it is one link per line. Sometimes I have looked at PyTorch's equivalent of this page in the browser for various reasons.

dellis23: Yup!  Good idea, thanks.  Might as well make it readable for humans too, since I am one.

PR #1392, opened 2022-09-20
Strips overloads to fix test

dan-garvey:
Do you think we should see if there is a strip_overloads in upstream torch and then try and move this out of heavydep? I assume its only here because of functorch?


ramiro050:
As far as I know, you can only access `strip_overloads` through functorch

build_tools/torchscript_e2e_heavydep_tests/README.md
```diff
@@ -13,7 +13,7 @@ self-contained virtual environment. It can be used like so:
 # serialized test artifacts in the other specified directory.
 # This command is safe to re-run if you have already built the virtual
 # environment and just changed the tests.
-build_tools/e2e_heavydep_tests/generate_serialized_tests.sh \
+build_tools/torchscript_e2e_heavydep_tests/generate_serialized_tests.sh \
```
ramiro050, line 16: The `torchscript_e2e_heavydep_tests` directory should be renamed to `e2e_heavydep_tests`. This seems to have been missed in https://github.com/llvm/torch-mlir/commit/e16b43e20b770ebbdfb536ff539bec8847537b59

ramiro050 (approved):
LGTM! I just have one small change request

powderluv:
btw no change required for this PR but in future please fork in your repo and push the PR instead of upstream branches

ramiro050:
@dan-garvey I think maybe you forgot to stage some of the changes. All I see now in this PR is the deletion of the `torchscript_e2e_heavydep_tests` directory

Issue #1361, opened 2022-09-13
Investigate failure of two TOSA end-to-end tests after LLVM tag update


ashay:
In PR #1359 for updating the LLVM tag to 4d4ca6c9, we see the `AvgPool2dFloatModule_basic` and `AvgPool2dCeilModeTrueModule_basic` fail due to segfaults.  The stack trace for both tests is identical:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff7dee859 in __GI_abort () at abort.c:79
#2  0x00007ffff7e5926e in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff7f83298 "%s\n") at ../sysdeps/posix/libc_fatal.c:155
#3  0x00007ffff7e612fc in malloc_printerr (str=str@entry=0x7ffff7f85a78 "malloc(): invalid next size (unsorted)") at malloc.c:5347
#4  0x00007ffff7e6410c in _int_malloc (av=av@entry=0x7ffff7fb8b80 <main_arena>, bytes=bytes@entry=936) at malloc.c:3739
#5  0x00007ffff7e66299 in __GI___libc_malloc (bytes=936) at malloc.c:3066
#6  0x00000000005a5697 in PyType_GenericAlloc ()
#7  0x00000000005a9def in ?? ()
#8  0x00007ffff779a3b5 in ?? () from /usr/lib/python3.8/lib-dynload/_ctypes.cpython-38-x86_64-linux-gnu.so
```

To reproduce the failure, build the code in branch of that PR, and run either the `AvgPool2dCeilModeTrueModule_basic` or the `AvgPool2dFloatModule_basic` test with the TOSA backend:

```
$ python3 -m e2e_testing.main --config=tosa -v -s --filter=AvgPool2dCeilModeTrueModule_basic
Compiling AvgPool2dCeilModeTrueModule_basic...
Running AvgPool2dCeilModeTrueModule_basic...
malloc(): invalid size (unsorted)
Aborted (core dumped)
```

silvasean:
@sjarus do you know who is the maintainer for these tests now?

sjarus:
Adding @eric-k256 (Eric Kunze)

eric-k256:
I'll take a look this afternoon. I don't think average pool has changed lately, but will find out more.

sjarus:
[...]

PR #1395, opened 2022-09-21
Add a way to turn off crashing tests

silvasean:
This adds a very long and obnoxious option to disable crashing tests. The right fix here is to use the right multiprocessing techniques to ensure that segfaulting tests can be XFAILed like normal tests, but we currently don't know how to implement "catch a segfault" in Python (patches or even just ideas welcome).

Motivated by #1361, where we ended up removing two tests from *all* backends due to a failure in one backend, which is undesirable.


silvasean:
> LGTM. This is better than what I had in mind, which was to run each XFAIL test in a separate subprocess and to check the return code of the subprocess.

That's what I would like us to do, but I literally couldn't find a way to do it. If you have ideas for how to do that in Python I would like to know (or see a patch :) )

Issue #1396, opened 2022-09-21
Build failing with "no matching function for call to 'convert'"


dellis23:
See here: https://github.com/llvm/torch-mlir/actions/runs/3098310406/jobs/5016066081#step:4:41827

I _think_ this is the relevant error.

```
/main_checkout/torch-mlir/libtorch/include/c10/util/Optional.h:786:29: error: no matching function for call to 'convert'
    return *this ? **this : detail_::convert<T>(constexpr_forward<V>(v));
                            ^~~~~~~~~~~~~~~~~~~
/main_checkout/torch-mlir/build_oot/python/torch_mlir/csrc/base_lazy_backend/generated/LazyNativeFunctions.cpp:726:89: note: in instantiation of function template specialization 'c10::optional<long>::value_or<at::Tensor>' requested here
        LazyTensorPtr lazy_storage_offset = torch::lazy::TryGetLtcTensor(storage_offset.value_or(at::Tensor()));
```

dellis23:
I think this particular issue is resolved now.  We may still be having occasionally other build issues, but I don't think the primary discussion is happening here.

Other issues I know of:

 * https://github.com/llvm/torch-mlir/issues/1434
 * https://github.com/llvm/torch-mlir/issues/1364

silvasean:
We are now pinning the PyTorch revision and so this should happen a lot less.

PR #1400, opened 2022-09-22
Disable LTC again

silvasean:
https://github.com/llvm/torch-mlir/issues/1396


fortianyou (approved):
LGTM, thanks @silvasean

dellis23 (approved):

dellis23:
Thanks!

antoniojkim:
I would very highly prefer that we fix the issue instead of just disabling LTC again. I have a PR up to fix the issue here
https://github.com/llvm/torch-mlir/pull/1401

silvasean:
Great! Thanks for the quick fix.

The general feeling in LLVM community development is to get HEAD back to a green state as fast as possible, rather than waiting for "fixing forward". See e.g. our patch reversion policy https://llvm.org/docs/DeveloperPolicy.html#patch-reversion-policy
In this case, it's awesome that we were able to get a fix so quickly, however in my experience usually these things can take longer, so it's always "safe" to go back to green and generally preferable to get HEAD green again immediately.

Issue #1328, opened 2022-08-31
pin pytorch source builds nightly to speed up CI


powderluv:
Pin the Pytorch bump to a nightly builder -- that also populates the ccache for the day.

silvasean:
The other option here is to do like we do for LLVM, with a deliberate ~weekly bump of the PyTorch dependency.

powderluv:
I am hoping the infra to bump will be the same -- and we can adjust how frequently (nightly / weekly) etc. When it fails it will still need someone to fix it up but hopefully we get the green commit moving without hand holding. 

We will also have to disable CIs from uploading their CI caches.

powderluv:
@ashay I have a WIP on how I think we could pin PyTorch and roll only when our tests pass. So we get the best of being very close to PyTorch automatically like we do -- but we don't roll into a broken master. 

I have a pseudo-code branch here: https://github.com/llvm/torch-mlir/tree/RollPyTorch if you want to give it a crack. 

Basically we add a nightly that runs our current OOT+PyTorch source build with Top of Master Pytorch. If it passes we commit the torch_mlir_pytorch_version file.  

And on the CI side if we load the pinned SHA.  Happy to discuss any other suggestions just wanted to dump what I had in my mind. 

FYI @sjain-stanford @makslevental

powderluv:
Oh thank you for the commit.  I think pinning the binary is good too, but I think we still need pure source builds for downstream customers. They will have long supported PyTorch forks (long past the PyTorch nightlies are removed) to enable a hermetic build environment without an external binary. 

Do you know anyway we can find the SHA of that particular binary you got with 
`python -m pip index versions -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html --pre torch | grep "Available versions" | tr ' ' '\n' | sort -nr | head -n1 | tr -d ','` ? 

We could use that SHA for the source builds CI and also pin the exact same binary for binary builds. 

Feel free to add any commits to the branch. We can clean up / review / commit once we figure it out.

powderluv:
Thanks @ashay 

[...]

PR #1401, opened 2022-09-22
Fix as_strided symint

antoniojkim:
Upstream PyTorch pushed out a change to `as_strided_copy` and `as_strided_scatter` for their Symbolic Integer effort. This caused a break in LTC.

Adding those ops to the symint list should fix things.

Also adding support for `hardtanh` to pass tests

https://github.com/llvm/torch-mlir/issues/1396


silvasean (approved):
Thanks for the quick fix!

antoniojkim:
I built it locally and looked to see what op the error was happening for
```
/main_checkout/torch-mlir/build_oot/python/torch_mlir/csrc/base_lazy_backend/generated/LazyNativeFunctions.cpp:726
```
It happened to be `as_strided_copy` which lined up with this commit added yesterday to upstream PyTorch
https://github.com/pytorch/pytorch/commit/3eb27229dd74dd0bea434326c471f16c50e558a4

antoniojkim:
> I built it locally and looked to see what op the error was happening for
> 
> ```
> /main_checkout/torch-mlir/build_oot/python/torch_mlir/csrc/base_lazy_backend/generated/LazyNativeFunctions.cpp:726
> ```
> 
> It happened to be `as_strided_copy` which lined up with this commit added yesterday to upstream PyTorch [pytorch/pytorch@3eb2722](https://github.com/pytorch/pytorch/commit/3eb27229dd74dd0bea434326c471f16c50e558a4)

And reading through that commit, they changed `as_strided_copy/scatter` to have symint support, so all we needed to do was make sure we were generating the symint versions of those ops too

silvasean:
"e2e_testing.torchscript.main" should be "e2e_testing.main" to fix the python issue.

dellis23:
Can we come up with a generally applicable flow for these situations?  I lean toward what Sean proposed of get the "rollback done" (though it's not exactly a rollback in this case) and then fix separately.

Is the worry that we might end up committing additional code while the LTC tests are disabled that would break LTC more and make fixing even worse?  Or would we also potentially generate releases that are broken for LTC?

powderluv:
IMHO Revert first. Disable second. Fix third to get a tree green.

powderluv:
@ashay yeah I think it is time to Pin Pytorch and roll only when tests pass. There is a tracker for it https://github.com/llvm/torch-mlir/issues/1328 . I had a WIP idea I will post in there to discuss and if you are willing to take over to land it please do.

silvasean:
> @ashay yeah I think it is time to Pin Pytorch and roll only when tests pass. There is a tracker for it #1328 . I had a WIP idea I will post in there to discuss and if you are willing to take over to land it please do.

Yeah, I think that is a good idea. I've added a note to discuss this in the community meeting two weeks from now. We can discuss at the developer hour too.

antoniojkim:
> I think it is time to Pin Pytorch and roll only when tests pass.

I'm in favour of this idea as well

PR #1403, opened 2022-09-22
Add build flag for out-of-tree builds

qedawkins:
Needed for external builds when we have an out-of-tree build (e.g. onnx-mlir)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,7 +54,9 @@ torch_mlir_add_llvm_external_project(
   TORCH_MLIR_DIALECTS
   ${CMAKE_CURRENT_SOURCE_DIR}/externals/llvm-external-projects/torch-mlir-dialects)
 
-if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+option(TORCH_MLIR_OOT_BUILD "Specifies an out of tree build" OFF)
+
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR OR TORCH_MLIR_OOT_BUILD)
   message(STATUS "Torch-MLIR out-of-tree build.")
   # Out-of-tree build
 


powderluv (approved):

CMakeLists.txt
```diff
@@ -54,7 +54,9 @@ torch_mlir_add_llvm_external_project(
   TORCH_MLIR_DIALECTS
   ${CMAKE_CURRENT_SOURCE_DIR}/externals/llvm-external-projects/torch-mlir-dialects)
 
-if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR)
+option(TORCH_MLIR_OOT_BUILD "Specifies an out of tree build" OFF)
+
+if(CMAKE_SOURCE_DIR STREQUAL CMAKE_CURRENT_SOURCE_DIR OR TORCH_MLIR_OOT_BUILD)
```
silvasean, line 59: Can we reliably detect this without the option? I thought that mlir-hlo or torch-mlir-external-dialects had a way to do this.

@marbre

silvasean: Can you rename this to TORCH_MLIR_BUILD_EMBEDDED for consistency?

Also, generally you should not merge PR's without resolving all conversations.

qedawkins: Sure. Was there an unresolved conversation here though?

silvasean: Yes, this one is marked unresolved.

qedawkins: but this conversation happened after it was approved/merged though?

powderluv: It is ok folks :)

BTW Is `TORCH_MLIR_BUILD_EMBEDDED` really what this is ? I assume embedded build in MHLO is different since there is no external deps vs for a Torch-MLIR Embedded build IMHO should be a standalone library of torch-mlir that we can embedded into other projects (with or without IREE)

qedawkins: That would probably be preferred in the long run because when building embedded we shouldn't need the llvm submodule nor mhlo. I didn't see an easy way to do that with the current Torch-MLIR repo.

powderluv: Please open an issue to track that feature request and we can see if there is a more general interest.

silvasean: Can we at least rename the flag OUT_OF_TREE instead of OOT?

qedawkins: I opened #1416. Apologies if I've been difficult here. How should I avoid this in the future, as having code-review level problems show up in a merged PR is definitely not something that should be happening often?

marbre: > Can we reliably detect this without the option? I thought that mlir-hlo or torch-mlir-external-dialects had a way to do this.
> 
> cc @marbre

Sry, i am currently oof (and sick), but I can take a look after I return. I think we can handle this more consistently across the projects.

PR #1410, opened 2022-09-23
[MHLO] Add torch-to-mhlo e2e support for aten.gather op

Vremold:


e2e_testing/xfail_sets.py
```diff
@@ -24,6 +24,8 @@
 
 MHLO_PASS_SET = {
     "BroadcastToIdentityCaseStaticModule_basic",
+    "GatherStaticModule_basic",
+    "GatherModule_basic",
```
tanyokwok, line 28: Could you please add more dynamic shapes e2e tests to cover the following cases:
1. Both input and index are randomly generated; (The index could be un-ordered)
2. Test data with different ranks

tanyokwok (commented):
Thanks @Vremold ! It LGTM with a minor comment on the e2e test.

e2e_testing/xfail_sets.py
Vremold: Done.

tanyokwok (approved):
LGTM

Vremold:
Merged. Thanks @tanyokwok @silvasean !

PR #1412, opened 2022-09-24
Fix Base Lazy Backend Type Conversion

GlebKazantaev:
### Description
This PR introduces several changes related to type conversions in base lazy backend.
* Fixed `c10::prim::Constant` conversion to use import options.
* Fixed `ivalue_importer` to use import options.
* Added CAPI for torch-mlir transformations.
* Updated base lazy backend to execute next passes:
  * `RefineTypes` - is a generic pass that performs type refinement so all backends should benefit from it. For instance, we have a real case with `aten.ScalarImplicit` operation which after conversion has `!torch.number` output type, so in our backend we would have to duplicate logic from `RefineTypes` pass to convert `!torch.number` to `!torch.float` or `!torch.int`.
  * `SatisfiesBackendContract` - is a validation pass that checks that after model import model satisfies backend contract. In case of failure it prints out the error message and returns `signalPassFailure()` status. This pass helps us to prevent backend failures due to unexpected IR structure.
* Updated `ModuleBuilder::Function` to use ImportOptions.

@antoniojkim @ke1337


test/python/importer/jit_ir/ivalue_import/annotations/arg-tensor-type-bound.py
```diff
@@ -32,6 +32,8 @@ def forward(self, a, b):
     ((), torch.float, True),
 ])
 
+import_options = ImportOptions()
+import_options.assumeTensorsHaveValueSemantics = False
```
silvasean, line 36: shouldn't this be the default?

test/python/importer/jit_ir/ivalue_import/tensors_non_value_semantics.py
```diff
@@ -5,7 +5,7 @@
 import typing
```
silvasean, line 5: I think you can just leave this one as tensors.py

python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
```diff
@@ -140,6 +141,32 @@ ComputationPtr TorchMlirLoweringContext::Build() {
...
+  TORCH_CHECK(
+      !mlirModuleIsNull(moduleOp),
+      "MlirModule is null.");
+
+  // Apply passes.
+  MlirPassManager pm = mlirPassManagerCreate(mlir_context_);
+  MlirOpPassManager nestedFuncPm = mlirPassManagerGetNestedUnder(
+    pm, mlirStringRefCreateFromCString("func.func"));
+
+  MlirPass refineTypesPass = mlirCreateRefineTypes();
```
silvasean, line 158: This breaks the architecture we are trying to maintain: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#overview

In particular, we should already know precise types in LTC and not need to run RefineTypes. Is there a way to fix this on the PyTorch side?

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -158,7 +158,7 @@ static bool satisfiesBackendContract(ModuleOp module,
   // We just check for the GlobalSlotModuleInitializerOp since its verifier
   // ensures that the set of global slots matches those initialized by the
   // module initializer.
-  auto walkResult0 = module.walk([&](Torch::GlobalSlotModuleInitializerOp op) {
+  auto walkResult0 = moduleOp.walk([&](Torch::GlobalSlotModuleInitializerOp op) {
```
silvasean, line 161: nit: let's not rename this variable

silvasean (changes requested):

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
GlebKazantaev: I change it because `module` is a C++20 keyword. https://en.cppreference.com/w/cpp/language/modules
Yes, we don't use this standard but I think it is a good practice not to overlap with any keywords.

test/python/importer/jit_ir/ivalue_import/tensors_value_semantics.py
```diff
@@ -0,0 +1,68 @@
...
+
+mb = ModuleBuilder()
+
+class TestModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        # TODO: Test (and make work) tensors that alias each other.
+        self.ones = torch.ones(1)
+        self.ones_i32 = torch.ones(1, dtype=torch.int32)
+        self.ones_i64 = torch.ones(1, dtype=torch.int64)
```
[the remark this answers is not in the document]
GlebKazantaev: Sure, I can keep single instance for `torch.ones`, `torch.quantize_per_tensor` and `torch.nn.Parameter`. I hope in future we won't have any value-semantic if-else statements inside single type.

test/python/importer/jit_ir/ivalue_import/annotations/arg-tensor-type-bound.py
GlebKazantaev: removed.

test/python/importer/jit_ir/ivalue_import/tensors_value_semantics.py
```diff
@@ -0,0 +1,68 @@
+# -*- Python -*-
```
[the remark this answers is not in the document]
GlebKazantaev: changed.

test/python/importer/jit_ir/ivalue_import/tensors_non_value_semantics.py
GlebKazantaev: changed.

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
GlebKazantaev: changed.

python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
GlebKazantaev: I found that after we apply `Scalar::Lower` and create an operation with `Tensor` output type [here](https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/csrc/base_lazy_backend/mlir_node_lowering.cpp#L284), so then PyTorch [is trying to align types](https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/frontend/schema_matching.cpp#L131-L137) and insert `Int/Float/Complex/Scalar-ImpliIcit` ops. And in my case I get this graph:
```
  ...
  %1 : Long(requires_grad=0, device=cpu) = prim::Constant[value={1}]()
  %2 : Long(requires_grad=0, device=cpu) = prim::Constant[value={1}]()
  ...
  %23 : Scalar = aten::ScalarImplicit(%2)
  %24 : Scalar = aten::ScalarImplicit(%1)
  %25 : Half(2, 10) = aten::addmm(%22, %p2, %12, %23, %24)
```
Based on `aten::addmm` spec it expects `Scalar` type (NumberType) for the 4th and 5th inputs, so that is why we get `ScalarImplicit` instead of `Int/Float-Implicit`.

But I guess that during type alignment in PyTorch for the case when destination type is `NumberType` we can deduce precise type from `Tensor` like this:
```cpp
...
} else if (concrete_number) {
        auto tensorType = value->type()->cast<TensorType>();
        if (auto scalarType = tensorType->scalarType()) {
            switch (*scalarType) {
              // This is just an example. Need to add more types.
              case at::ScalarType::Byte:
              case at::ScalarType::Char:
              case at::ScalarType::Short:
              case at::ScalarType::Int:
              case at::ScalarType::Long:
                value = graph.insert(aten::IntImplicit, {value}, {}, loc);
                break;
              default:
                value = graph.insert(aten::ScalarImplicit, {value}, {}, loc);
            }
        } else {
          value = graph.insert(aten::ScalarImplicit, {value}, {}, loc);
        }
      }
...
```

Using this patch I was able to get correct JIT graph with refined types:
```
  ... 
  %1 : Long(requires_grad=0, device=cpu) = prim::Constant[value={1}]()
  %2 : Long(requires_grad=0, device=cpu) = prim::Constant[value={1}]()
  ... 
  %23 : int = aten::IntImplicit(%2)
  %24 : int = aten::IntImplicit(%1)
  %25 : Half(2, 10) = aten::addmm(%22, %p2, %12, %23, %24)
```
and torch-mlir:
```
  ...
  %0 = torch.vtensor.literal(dense<1> : tensor<si64>) : !torch.vtensor<[],si64>
  %1 = torch.vtensor.literal(dense<1> : tensor<si64>) : !torch.vtensor<[],si64>
 ...
  %7 = torch.aten.IntImplicit %1 : !torch.vtensor<[],si64> -> !torch.int
  %8 = torch.aten.IntImplicit %0 : !torch.vtensor<[],si64> -> !torch.int
  %9 = torch.aten.addmm %6, %arg2, %4, %7, %8 : !torch.vtensor<[10],f16>, !torch.vtensor<[2,16],f16>, !torch.vtensor<[16,10],f16>, !torch.int, !torch.int -> !torch.vtensor<[2,10],f16>
```

On the one hand there is nothing criminal in that we make type more precise, but on the other hand the `Scalar` type was also correct. So I think it is better do discuss this change with PyTorch team. 
Is there any chance that we keep `RefineTypes` transformation for now in base_lazy_backend until we push this fix to the PyTorch?

silvasean: Adding RefineTypes here significantly breaks the architecture of Torch-MLIR, so we do not want to add it, even temporarily.

glebk-cerebras: Okay, I'll try to make a fix on PyTorch side.
BTW, is there a list of requirements in PyTorch that the LTC satisfies? I mean guaranties for value semantics, type refinement etc.

GlebKazantaev: Okay, I'll try to make a fix on PyTorch side.
BTW, is there a list of requirements in PyTorch that the LTC satisfies? I mean guaranties for value semantics, type refinement etc.

silvasean: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#the-backend-contract -- it should be checked by the SatisfiesBackendContract pass you added.

See also this for reference: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#lazytensorcore

include/torch-mlir-c/Transforms.h
```diff
@@ -0,0 +1,22 @@
+//===-- torch-mlir-c/Transforms.h - Helpers for C API to Core MLIR ------*- C -*-===//
```
silvasean, line 1: nit: 80 columns

include/torch-mlir/Dialect/Torch/Transforms/Passes.td
```diff
@@ -329,4 +329,16 @@ def LowerToBackendContract
   let dependentDialects = ["func::FuncDialect"];
 }
 
+def SatisfiesBackendContract
```
silvasean, line 332: Nit: can you name this VerifySatisfiesBackendContract / verify-satisfies-backend-contract -- we like passes to be named with verbs the the beginning.

python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
```diff
@@ -140,6 +141,25 @@ ComputationPtr TorchMlirLoweringContext::Build() {
       /*getArgAttribute=*/[](int) -> MlirAttribute { return {nullptr}; },
       /*importOptions=*/{/*assumeTensorsHaveValueSemantics=*/true});
 
+  MlirLocation loc = mlirLocationUnknownGet(mlir_context_);
+  MlirModule moduleOp = mlirModuleCreateEmpty(loc);
+  MlirBlock moduleBody = mlirModuleGetBody(moduleOp);
+  mlirBlockInsertOwnedOperation(moduleBody, 0, func_op);
```
silvasean, line 147: So it appears that you insert the function as owned, but then you throw away the module -- that seems like it would create a use-after-free. I think you will need to remove the function from the module, or switch TorchMlirComputation to accepting a module

test/python/importer/jit_ir/ivalue_import/tensors.py
```diff
@@ -55,10 +55,14 @@ def __init__(self):
 # CHECK:   torch.slot "ones_qint8", %[[ONES_QINT8]] : !torch.tensor<[1],!torch.qint8>
 # CHECK:   torch.slot "ones_quint8", %[[ONES_QUINT8]] : !torch.tensor<[1],!torch.quint8>
 # CHECK: }
-
-
 test_module = TestModule()
 recursivescriptmodule = torch.jit.script(test_module)
+
+import_options = ImportOptions()
+import_options.assumeTensorsHaveValueSemantics = False
```
silvasean, line 62: can we make this the default (i.e. retain the old behavior) so that we don't need to add custom import options everywhere?

test/python/importer/jit_ir/node_import/prim.py
```diff
@@ -162,3 +162,34 @@ def prim_max(x: int):
...
+# CHECK-LABEL:   func.func @__torch__.prim_Constant_scalar() -> !torch.number {
+# CHECK:           %[[A:.*]] = torch.tensor.literal
+# CHECK:           %[[RET:.*]] = torch.aten.ScalarImplicit
+# CHECK:           return %[[RET]] : !torch.number
+import_options = ImportOptions()
+import_options.assumeTensorsHaveValueSemantics = False
+mb.import_function(create_script_function("__torch__.prim_Constant_scalar", """
+graph():
+  %0 : Long(requires_grad=0, device=cpu) = prim::Constant[value={1}]()
+  %1 : Scalar = aten::ScalarImplicit(%0)
```
silvasean, line 175: are these tests still relevant?

silvasean (changes requested):

test/python/importer/jit_ir/node_import/prim.py
GlebKazantaev: Yes as they test node_import code https://github.com/llvm/torch-mlir/pull/1412/files#diff-05424e18718175ef3b778a9eb98fe120702705e14d57eab13674ebf84df8e28bR201
And test/python/importer/jit_ir/ivalue_import/tensors-value-semantics.py test ivalue_importer.

test/python/importer/jit_ir/ivalue_import/tensors.py
GlebKazantaev: done.

include/torch-mlir-c/Transforms.h
GlebKazantaev: done

include/torch-mlir/Dialect/Torch/Transforms/Passes.td
GlebKazantaev: done

GlebKazantaev:
I have some concerns regarding VerifySatisfyBackendContract pass execution. I think that the best option now is not execute VerifySatisfyBackendContract pass in base lazy backend. And the reason for that is that with this change we won't be able to fix problems quickly if something breaks in PyTorch LTC. For an instance, the case with !torch.number type which is prohibited by backend contract. It will take time to make a fix on PyTorch side and probably it won't be included into the next release. Especially it make things worse when we stick to some particular version of PyTorch. To be able to temporary work around PyTorch LTC issues we should pass torch-mlir model as is to the backend without any additional validation (like we do right now).

I also want to keep CAPI for torch-mlir transformations as we can reuse this passes for conversion needs in lazy backend.

@silvasean I resolved your comments. Please let me know what you think about this ^^^

silvasean:
> I have some concerns regarding VerifySatisfyBackendContract pass execution. I think that the best option now is not execute VerifySatisfyBackendContract pass in base lazy backend. And the reason for that is that with this change we won't be able to fix problems quickly if something breaks in PyTorch LTC. For an instance, the case with !torch.number type which is prohibited by backend contract. It will take time to make a fix on PyTorch side and probably it won't be included into the next release. Especially it make things worse when we stick to some particular version of PyTorch. To be able to temporary work around PyTorch LTC issues we should pass torch-mlir model as is to the backend without any additional validation (like we do right now).
> 
> I also want to keep CAPI for torch-mlir transformations as we can reuse this passes for conversion needs in lazy backend.
> 
> @silvasean I resolved your comments. Please let me know what you think about this ^^^

We should include the VerifySatisfyBackendContract pass. It is intended that breaking the backend contract should immediately bring development to a halt. It is like if we produce verifier-invalid MLIR -- there is not point in continuing because the invariants of the system are broken already at the input level, and the producer has to be fixed. We will need to work with upstream PyTorch to improve CI or other signals and development processes if breaking the backend contract becomes common.

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
```diff
@@ -109,13 +111,23 @@ std::unique_ptr<OperationPass<ModuleOp>>
 createLowerToBackendContractPass(int maxIterations, bool decompose,
                                  ArrayRef<std::string> backendLegalOps);
 
+std::unique_ptr<OperationPass<ModuleOp>>
+createVerifySatisfiesBackendContractPass();
```
silvasean, line 115: Nit: I'm sorry for asking you to change this, but could we call this VerifyBackendContract? I don't think "satisfies" adds much.

glebk-cerebras:
> We should include the VerifySatisfyBackendContract pass.
Yes, I agree, and that was my initial intention. But I've never considered the fact that we just started to run large models using LTC and we don't have enough test coverage to be sure that LTC satisfies backend contract for most basic/common cases.
Enabling this pass by default right now will shoot ourselves in the foot. As I said it make things worse when we or customer stick to some particular release version of PyTorch/TorchMLIR, so we don't have a chance to temporarily fix this issues on backend side. So we either have to wait until next PyTorch release or ask customers to update on custom version of PyTorch which is not possible in most cases.

The solution Im currently thinking is this:
We never executed VerifySatisfyBackendContract pass in lazy backend before, and to make transition smoothly we can use TorchMLIR transformation CAPI to run this validation pass in lazy backends and see which cases doesn't work. This will give us enough flexibility at the beginning. So we can easily move forward, fix PyTorch LTC issues and create sufficient test coverage to enable this verification pass by default in future.

@silvasean, @antoniojkim what do you think?

silvasean:
Can you explain more this thing about the different PyTorch versions? As far as I know, Torch-MLIR LTC must be built against PyTorch HEAD. Building it against released versions is not supported.

Can you do an initial assessment of the impact of adding this pass? If LTC is not already creating IR that follows the backend contract then that is a big problem, because it means that our LTC frontend won't compose with our MHLO/TOSA/Linalg backends. We should have a quantified workstream for aligning LTC with the backend contract if it doesn't already.

glebk-cerebras:
> Can you explain more this thing about the different PyTorch versions?

For example, for our next release we will require the latest PyTorch version (which is going to be 1.13), so we can build torch-mlir against this PyTorch version and distribute as a part of our env. But if we will find out some PyTorch LTC issues after PyTorch 1.13 released, we won't be able to fix it on PyTorch side, so that's why we need this flexibility to work around LTC issues on our Cerebras lazy backend side.

> Can you do an initial assessment of the impact of adding this pass? 
The only issue I currently observe is related to !torch.number type which is unsupported by backend contract https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L146
The root cause is the type alignment in PyTorch. I described it in details here: https://github.com/llvm/torch-mlir/pull/1412#discussion_r980316798
I haven't observed any other issues before, but I've just started to explore larger models, so I can't guaranty that we won't find smth else.

glebk-cerebras:
@silvasean is there any chance that we can merge this changes and continue our discussion on github issues or discord?
I removed all unnecessary changes from this PR, so it doesn't change current logic in the base_lazy_backend.

_PS. I just need to rebase and rename `VerifySatisfyBackendContract`_

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
glebk-cerebras: Done.

GlebKazantaev: Done.

silvasean:
> For example, for our next release we will require the latest PyTorch version (which is going to be 1.13), so we can build torch-mlir against this PyTorch version and distribute as a part of our env. But if we will find out some PyTorch LTC issues after PyTorch 1.13 released, we won't be able to fix it on PyTorch side, so that's why we need this flexibility to work around LTC issues on our Cerebras lazy backend side. Otherwise our model enabling process will be blocked until the next PyTorch release version which release date should be also considered for our Cerebras release planing which make things too complicated.

It sounds like you will need to backport any Torch-MLIR LTC changes to PyTorch 1.13 in that case, which means a custom build of Torch-MLIR. You would be able to comment out the VerifyBackendContract pass as part of those if necessary, right?



> The only issue I currently observe is related to !torch.number type which is unsupported by backend contract https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L146
> The root cause is the type alignment in PyTorch. I described it in details here: [#1412 (comment)](https://github.com/llvm/torch-mlir/pull/1412#discussion_r980316798)
> I haven't observed any other issues so far, but I've just started to explore larger models, so I can't guaranty that we won't find smth else.

VerifyBackendContract is not the only thing that could fail. The regular IR verifier could also fail, or the torch::jit::Graph could fail to be imported by the node_importer. If you expect there to be issues related to invalid IR due to PyTorch updates, then you will need to be prepared for appropriate backporting/patching -- it's not just the VerifyBackendContract pass. I don't think that a single instance of breakage here is enough to justify not enforcing a fundamental architectural invariant of the system.

silvasean (approved):

silvasean:
I really think you should add the VerifyBackendContract pass to the base_lazy_backend before submitting this. That is a big improvement to the system and would have saved a bunch of time already. I don't think that speculation about future violations of the backend contract is enough to justify not having it, since it is a fundamental architectural invariant of the system.

GlebKazantaev:
> VerifyBackendContract is not the only thing that could fail. The regular IR verifier could also fail, or the torch::jit::Graph could fail to be imported by the node_importer.

That's a good point!

> I really think you should add the VerifyBackendContract pass to the base_lazy_backend before submitting this.

To enable `VerifyBackendContract` pass I also have either to figure out how to detach MlirOperation from MlirModule after pass manager execution or update TorchMlirComputation and some other places to work with MlirModule. Not to block the generic changes in this PR and track remaining work I've created a separate [issue](https://github.com/llvm/torch-mlir/issues/1460).

python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
GlebKazantaev: I think switching to `TorchMlirComputation` is the best option. I'll try to work on this within https://github.com/llvm/torch-mlir/issues/1460

GlebKazantaev:
@ashay @silvasean is there any way to grant me some rights, so the workflows will start running automatically?

image

GlebKazantaev:
> Unfortunately, I am the wrong person, since I lack the power to make that happen.
No problem, I'll ask on slack.

silvasean:
> @ashay @silvasean is there any way to grant me some rights, so the workflows will start running automatically?
> 
> image

Just sent you an invite!

PR #1413, opened 2022-09-25
[MLIR] Add canonicalizer for aten.slice.t op

Vremold:
`aten.slice.t` op lacks a canonicalizer. It will cause the shape-simplification pipeline fails to infer %3 to have a static shape in the following IR:
```
%0 = torch.tensor.literal(dense_resource<__elided__> : tensor<4x300xf32>) : !torch.tensor<[4,300],f32>
%1 = torch.prim.ListConstruct %int16, %int75, %int2, %int2 : (!torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
%2 = torch.aten.slice.t %1, %int0, %int-2, %int1 : !torch.list<int>, !torch.int, !torch.int, !torch.int -> !torch.list<int>
%3 = torch.aten.view %0, %2 : !torch.tensor<[4,300],f32>, !torch.list<int> -> !torch.tensor<[?,?],f32>
```
This PR adds a canonicalizer for `aten.slice.t` op. The basic idea is replacing `aten.slice.t` op with a new `prim.ListConstruct` op, when the former's operand comes from `prim.ListConstruct` op and all its parameters (i.e. `start`, `end` and `step`) are constant integers (or none if possible) .

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -1730,22 +1730,35 @@ void AtenSliceTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
   patterns.add(+[](AtenSliceTOp op, PatternRewriter &rewriter) {
     auto valueList = op.l();
     auto listConstructOp = valueList.getDefiningOp<PrimListConstructOp>();
-    if (!listConstructOp) {
+    if (!listConstructOp || isListPotentiallyMutated(listConstructOp)) {
       return failure();
     }
 
     SmallVector<Value> listElements =
         llvm::to_vector<4>(listConstructOp.elements());
+    int64_t size = static_cast<int64_t>(listElements.size());
+
     int64_t st;
     int64_t ed;
     int64_t step;
-    if (!matchPattern(op.start(), m_TorchConstantInt(&st)) ||
-        !matchPattern(op.end(), m_TorchConstantInt(&ed)) ||
-        !matchPattern(op.step(), m_TorchConstantInt(&step))) {
+    if (op.start().getType().isa<Torch::NoneType>()) {
+      st = 0;
+    } else if (!matchPattern(op.start(), m_TorchConstantInt(&st))) {
       return failure();
     }
-    st = st >= 0 ? st : st + listElements.size();
-    ed = ed >= 0 ? ed : ed + listElements.size();
+    if (op.end().getType().isa<Torch::NoneType>()) {
+      ed = listElements.size();
+    } else if (!matchPattern(op.end(), m_TorchConstantInt(&ed))) {
+      return failure();
+    }
+    if (!matchPattern(op.step(), m_TorchConstantInt(&step))) {
+      return failure();
+    }
+
+    st = st >= 0 ? st : st + size;
+    st = st >= 0 ? st : 0;
+    ed = ed >= 0 ? ed : ed + size;
+    ed = ed < size ? ed : size;
     SmallVector<Value> newListElements;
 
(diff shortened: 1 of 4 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1721,6 +1721,43 @@ void AtenAddTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+        llvm::to_vector<4>(listConstructOp.elements());
+    int64_t st;
+    int64_t ed;
+    int64_t step;
+    if (!matchPattern(op.start(), m_TorchConstantInt(&st)) ||
+        !matchPattern(op.end(), m_TorchConstantInt(&ed)) ||
+        !matchPattern(op.step(), m_TorchConstantInt(&step))) {
+      return failure();
+    }
+    st = st >= 0 ? st : st + listElements.size();
```
silvasean, line 1747: This is not enough to guarantee that the value is positive. The operand could be `torch.constant.int -1000`.Please add .mlir test for this case as well.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1721,6 +1721,43 @@ void AtenAddTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+//===----------------------------------------------------------------------===//
+// AtenSliceTOp
+//===----------------------------------------------------------------------===//
+
+void AtenSliceTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                               MLIRContext *context) {
+  patterns.add(+[](AtenSliceTOp op, PatternRewriter &rewriter) {
+    auto valueList = op.l();
+    auto listConstructOp = valueList.getDefiningOp<PrimListConstructOp>();
+    if (!listConstructOp) {
```
silvasean, line 1733: You need to check isListPotentiallyMutated as well

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1721,6 +1721,43 @@ void AtenAddTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
   });
```
silvasean, line 1721: please add .mlir tests

silvasean (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
Vremold: Thanks for the advice. I'v made the changes in the latest commit. Please review.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1721,6 +1721,56 @@ void AtenAddTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+    auto listConstructOp = valueList.getDefiningOp<PrimListConstructOp>();
+    if (!listConstructOp || isListPotentiallyMutated(listConstructOp)) {
+      return failure();
+    }
+
+    SmallVector<Value> listElements =
+        llvm::to_vector<4>(listConstructOp.elements());
+    int64_t size = static_cast<int64_t>(listElements.size());
+
+    int64_t st;
```
silvasean, line 1741: nit: call these "start" and "end"

silvasean (approved):
Great tests!

lib/Dialect/Torch/IR/TorchOps.cpp
Vremold: Done.

PR #1415, opened 2022-09-26
build: update llvm tag to bebc9695

ashay:
Summary of changes:
  - Renamed OptionalArrayRefParameter since the name conflicts with an
    upstream symbol that has a different meaning
    (https://reviews.llvm.org/D133819)
  - Removed extraneous dependency between TorchMLIRTorchToMhlo and
    ChloOps, since the existing dependency on MhloDialect is sufficient
  - Fixed code to prevent warnings related to comparisons between signed
    and unsigned values

### DIFF
--- a/lib/Conversion/TorchToMhlo/Linear.cpp
+++ b/lib/Conversion/TorchToMhlo/Linear.cpp
@@ -102,13 +102,13 @@ RankedTensorType castContractingDim(PatternRewriter &rewriter, Operation *op,
   }
   SmallVector<int64_t> outShape;
   // set batch dims, will skip invalid dimensions
-  for (size_t k = 0; k < lhsShape.size(); ++k) {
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToMhlo/Linear.cpp
```diff
@@ -102,13 +102,13 @@ RankedTensorType castContractingDim(PatternRewriter &rewriter, Operation *op,
   }
   SmallVector<int64_t> outShape;
   // set batch dims, will skip invalid dimensions
-  for (size_t k = 0; k < lhsShape.size(); ++k) {
+  for (int64_t k = 0; k < static_cast<int64_t>(lhsShape.size()); ++k) {
```
silvasean, line 105: it seems weird that you changed the signedness of the iteration variable, and also had to insert a cast as well.

silvasean (approved):

PR #1419, opened 2022-09-26
Github action to update PyTorch version

ashay:
This PR contains four commits:

 - build: allow upgrading pip modules (that adds the `--upgrade` flag to pip)
 - build: [nfc] clarify source versus binary PyTorch builds (minor variable name change)
 - build: create RollPyTorch to update PyTorch version in Torch-MLIR
 - build: checkout Pytorch using either branch name or commit hash

### DIFF
--- a/pytorch-requirements.txt
+++ b/pytorch-requirements.txt
@@ -1,1 +1,3 @@
--f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html --pre torch==1.13.0.dev20220927
+-f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+--pre
+torch==1.13.0.dev20220927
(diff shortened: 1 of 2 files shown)


powderluv (approved):
As discussed on discord we can auto-approve when tests pass so there is no human in the loop when tests are good. When it fails we will need it to be fixed and rolled.

pytorch-requirements.txt
```diff
@@ -0,0 +1 @@
+-f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html --pre torch==1.13.0.dev20220926
```
powderluv, line 1: lets change this to last nights build.

powderluv: oh nvm I think you already have a passing version here but you may still want last nights so nothing breaks.

PR #1420, opened 2022-09-26
Propagate parameter names

antoniojkim:
We need to propagate parameter names into the MLIR graph in order to create proper tensor mappings for the Cerebras backend.

CC: @ke1337 @glebk-cerebras @vaibhav-cerebras @prateekgu-cerebras


silvasean (changes requested):
This changes the semantics of the external calling convention of the MLIR module, which is one of the most complicated parts of the compiler (or any compiler). We currently provide guarantees about positional arguments. Is there no way for you to pass this information directly to the lower-level compiler you are calling into, since Torch-MLIR doesn't use this information, and we guarantee the order of positional arguments?

antoniojkim:
@silvasean This will not affect the MLIR graph generated by Torch-MLIR LTC unless the parameter names are explicitly set
https://github.com/antoniojkim/torch-mlir/blob/5b41855c4d71d885751583700f1f33b9d51bb77d/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp#L252-L254

If not set it will return a null attribute same as before
https://github.com/antoniojkim/torch-mlir/blob/5b41855c4d71d885751583700f1f33b9d51bb77d/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp#L164-L166

> Also, conceptually, I don't understand why an LTC backend would need this information. Can you explain what your backend does with it?

This is something specifically required for the Cerebras LTC backend. We require the parameter names be present in the MLIR graph in order to be able to generate our internal tensor mappings.

silvasean:
Can you handle this outside Torch-MLIR? E.g. after lowering the Torch-MLIR graph you will add your Cerebras-specific attributes afterward.

silvasean:
Yes, this is a huge concern since it changes the external calling convention of the module. Please find a way to do it outside Conceptually this seems like it should be not very difficult -- it should be possible to build a map associating the parameter names to function argument numbers, and then after you lower the graph with Torch-MLIR, you apply those attributes on the parameters.

antoniojkim:
@silvasean I was able to attach the parameter name attribute after lowering into MLIR, so I've cleaned up and removed that change. Please review again to see if the external calling conventions are now preserved

silvasean (approved):
LG -- make sure to update the commit message to the latest version of the patch

PR #1424, opened 2022-09-27
[LINALG] Add lowering for the aten.upsample_nearest2d op.

pashu123:


ramiro050:
Can you resolve the conflicts so that the CI can run?

python/torch_mlir_e2e_test/test_suite/conv.py
```diff
@@ -12,6 +12,7 @@
 
 
 class Conv2dNoPaddingModule(torch.nn.Module):
+
```
ramiro050, line 15: Can you make all the style changes in this file in a separate PR? It makes it hard to know which changes are truly related to this PR

ramiro050 (changes requested):

python/torch_mlir_e2e_test/test_suite/conv.py
pashu123: Sure.

pashu123: Done.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -152,6 +153,133 @@ class ConvertAtenFlipOp : public OpConversionPattern<AtenFlipOp> {
...
+    auto inputRank = input.getType().cast<RankedTensorType>().getRank();
+    Type elementType =
+        input.getType().cast<RankedTensorType>().getElementType();
+
+    SmallVector<Value> dims = getTensorSizes(rewriter, loc, input);
+    SmallVector<Value> scaleFactorsInt;
+
+    if (!adaptor.scale_factors().getType().isa<Torch::NoneType>()) {
+      SmallVector<double> scaleFactors;
+      if (!matchPattern(adaptor.scale_factors(),
```
ramiro050, line 221: As far as I can tell, the lowering for this op does not need this to be a constant float list. It seems like all the floats are converted back to values. You should use `getListConstructElements` instead.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -152,6 +153,133 @@ class ConvertAtenFlipOp : public OpConversionPattern<AtenFlipOp> {
 };
 } // namespace
 
+// Returns updated output dimension and scale_factor converted to integer.
+static std::pair<Value, Value> getScaledDims(OpBuilder &builder, Location loc,
```
ramiro050, line 157: Getting the `scale_factor` converted to an int seems to be unrelated to the goal of this function, which is to scale the dims. I think it would be clearer to have this function simply return the scaled dim, and have the conversion to `int` of `scale_factor` happen in the lowering itself (since it's just one line of code)

python/torch_mlir_e2e_test/test_suite/conv.py
```diff
@@ -705,3 +705,80 @@ def forward(self, inputVec, weight):
...
+class UpSampleNearest2dStaticSize(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([1, 1, 4, 4], torch.float32, True),
```
ramiro050, line 717: Is there a reason not to have all tests be fully dynamic?

python/torch_mlir_e2e_test/test_suite/conv.py
```diff
@@ -705,3 +705,80 @@ def forward(self, inputVec, weight):
...
+    ])
+    def forward(self, inputVec):
+        return torch._C._nn.upsample_nearest2d(inputVec,
+                                               output_size=[11, 11],
+                                               scale_factors=None)
+
+
+@register_test_case(module_factory=lambda: UpSampleNearest2dStaticSize())
+def UpSampleNearest2dStaticSize_basic(module, tu: TestUtils):
+    module.forward(torch.randn(1, 1, 4, 4))
```
ramiro050, line 727: use `tu.rand`. See:

https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#test-case-function

ramiro050 (changes requested):

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
```diff
@@ -150,6 +150,37 @@ m_TorchConstantIntList(SmallVectorImpl<int64_t> &bind_values) {
   return detail::torch_list_construct_op_binder(bind_values);
 }
 
+namespace detail {
+/// Matches the constant integers stored in a `torch.ListConstruct`.
```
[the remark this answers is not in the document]
pashu123: Done.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -152,6 +153,133 @@ class ConvertAtenFlipOp : public OpConversionPattern<AtenFlipOp> {
...
+                  ConversionPatternRewriter &rewriter) const override {
+
+    Location loc = op->getLoc();
+    Value input = adaptor.input();
+
+    Type resultType = getTypeConverter()->convertType(op.getResult().getType());
+
+    auto inputRank = input.getType().cast<RankedTensorType>().getRank();
+    Type elementType =
+        input.getType().cast<RankedTensorType>().getElementType();
```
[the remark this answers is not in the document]
pashu123: Done.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -152,6 +153,133 @@ class ConvertAtenFlipOp : public OpConversionPattern<AtenFlipOp> {
...
+
+    SmallVector<Value> dims = getTensorSizes(rewriter, loc, input);
+    SmallVector<Value> scaleFactorsInt;
+
+    if (!adaptor.scale_factors().getType().isa<Torch::NoneType>()) {
```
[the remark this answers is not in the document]
pashu123: Done.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -152,6 +153,133 @@ class ConvertAtenFlipOp : public OpConversionPattern<AtenFlipOp> {
...
+      SmallVector<double> scaleFactors;
+      if (!matchPattern(adaptor.scale_factors(),
+                        m_TorchConstantFloatList(scaleFactors)))
+        return rewriter.notifyMatchFailure(op,
+                                           "only constant dim lists supported");
+      for (unsigned i = 0; i < scaleFactors.size(); i++) {
+        auto scaledPair =
+            getScaledDims(rewriter, loc, dims[2 + i], scaleFactors[i]);
```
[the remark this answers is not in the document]
pashu123: Done.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -152,6 +153,133 @@ class ConvertAtenFlipOp : public OpConversionPattern<AtenFlipOp> {
...
+        scaleFactorsInt.push_back(scaledPair.first);
+        dims[2 + i] = scaledPair.second;
+      }
+    } else {
+
+      SmallVector<int64_t> outputDims;
+      if (!matchPattern(adaptor.output_size(),
```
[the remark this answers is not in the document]
pashu123: I see, yes. Thanks for the suggestion.

python/torch_mlir_e2e_test/test_suite/conv.py
pashu123: I have added 2 static and 2 dynamic tests. If you want me to add all static I can do that.

ramiro050: I would just keep the 2 dynamic ones and remove the static ones. The static case should just be a special case of the dynamic case, so if it works dynamically, it should work statically.

ramiro050 (approved):
Other than the removal of static tests and code to match float lists, this LGTM! Thanks!

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
```diff
@@ -150,6 +150,37 @@ m_TorchConstantIntList(SmallVectorImpl<int64_t> &bind_values) {
   return detail::torch_list_construct_op_binder(bind_values);
 }
 
+namespace detail {
+/// Matches the constant floats stored in a `torch.ListConstruct`.
+struct torch_list_construct_float_op_binder {
```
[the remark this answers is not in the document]
pashu123: Makes sense. Done.

python/torch_mlir_e2e_test/test_suite/conv.py
pashu123: Done.

pashu123: Converted the static ones to dynamic ones as they were covering different cases (same scaled_h, scaled_w vs different).

pashu123:
@ramiro050 Any ideas why it's failing with the error message         ```
RuntimeError: Unimplemented Error: 
        	/main_checkout/torch-mlir/build/tools/torch-mlir/python/torch_mlir/csrc/base_lazy_backend/generated/shape_inference.cpp:79    std::vector<torch::lazy::Shape> torch::lazy::compute_shape_upsample_nearest2d(const at::Tensor &, at::OptionalIntArrayRef, c10::optional<at::ArrayRef<double>>)
```

PR #1427, opened 2022-09-27
Bump shape library

ramiro050:

PR #1428, opened 2022-09-27
Add `--no-index` to CI's git-diff check on generated files

ramiro050:
`git diff` does not work by default on untracked files. Since the function `_check_file_not_changed_by` stores the new generated file in an untracked file, `git diff` was not catching any modifications in the new generated file. This commit adds the flag `--no-index` to make `git diff` work with untracked files.

### DIFF
--- a/build_tools/python_deploy/build_linux_packages.sh
+++ b/build_tools/python_deploy/build_linux_packages.sh
@@ -211,12 +211,12 @@ function _check_file_not_changed_by() {
   mv "$file_backup" "$file"
   # We use git-diff as "just a diff program" (no SCM stuff) because it has
   # nicer output than regular `diff`.
-  if ! git diff --quiet "$file" "$file_new"; then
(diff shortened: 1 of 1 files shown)


ramiro050:
The CI should fail if this fix is right, since there are new changes made to the shape library: https://github.com/llvm/torch-mlir/pull/1427

silvasean:
I thought that when you pass two files it directly compares them without any VCS behavior. I use that a lot -- git-diff as just a regular diffing tool.

ramiro050:
> I thought that when you pass two files it directly compares them without any VCS behavior. I use that a lot -- git-diff as just a regular diffing tool.

From [`git-diff`s docs](https://git-scm.com/docs/git-diff):

> git diff [<options>] --no-index [--] <path> <path>
> This form is to compare the given two paths on the filesystem. You can omit the --no-index option when running the command in a working tree controlled by Git and at least one of the paths points outside the working tree, or when running the command outside a working tree controlled by Git. This form implies --exit-code.

PR #1429, opened 2022-09-28
[tosa] Add support for some cases of aten.broadcast_to op

vivekkhandelwal1:
This commit adds support for TorchToTosa lowering of `aten.broadcast_to` op for cases:
1.) When the rank of input and output tensor is equal.
2.) When the rank of input tensor is zero.

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -23,7 +23,6 @@ EAGER_MODE_XFAIL_SET = {
 }
 
 MHLO_PASS_SET = {
-    "BroadcastToIdentityCaseStaticModule_basic",
     "GatherStaticModule_basic",
     "GatherModule_basic",
     "Gather2DInputModdule_basic",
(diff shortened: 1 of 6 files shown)


e2e_testing/xfail_sets.py
```diff
@@ -454,7 +453,8 @@
     "_LogSoftmaxModuleStable_basic",
     "LiftFreshCopyModule_basic",
     "ReduceSumDimIntListKeepDimNegativeDimStaticModule_basic",
-    "BroadcastToIdentityCaseStaticModule_basic",
+    "BroadcastToSameRankStaticModule_basic",
```
sjarus, line 456: I'm curious - why is the MHLO pass set impacted by a TOSA change ? Is it related to something changed in Utils ?

sjarus (approved):

e2e_testing/xfail_sets.py
vivekkhandelwal1: It's because I added a new test which passes.

Issue #1425, opened 2022-09-27
Unable to compile functions with dictionary inputs


dellis23:
I've tried a few approaches.  Not sure if any of these _should_ work or if I'm missing the correct way altogether.

```python
import torch
from torch import Tensor
import torch_mlir
from functorch import make_fx
from typing import Dict

def dict_fn(x: Dict[str, Tensor]):
  return x['a'] * 2.0

print(dict_fn({'a': torch.tensor(3.0)}))

linalg_on_tensors_mlir = torch_mlir.compile(
    make_fx(dict_fn)({'a': torch.tensor(3.0)}),
    {'a': torch.tensor(3.0)},
    output_type=torch_mlir.OutputType.LINALG_ON_TENSORS)
```

[Output](https://gist.github.com/dellis23/5e661e642961732471010392a76883ea)

```python
import torch
from torch import Tensor
import torch_mlir
from functorch import make_fx
from typing import Dict

@torch.jit.script
def dict_fn(x: Dict[str, Tensor]):
  return x['a'] * 2.0

print(dict_fn({'a': torch.tensor(3.0)}))

linalg_on_tensors_mlir = torch_mlir.compile(
    make_fx(dict_fn)({'a': torch.tensor(3.0)}),
    {'a': torch.tensor(3.0)},
    output_type=torch_mlir.OutputType.LINALG_ON_TENSORS)
```

[Output](https://gist.github.com/dellis23/1a64b686934166bec4cbaf80215bf3bc)

```python
import torch
from torch import Tensor
import torch_mlir
from functorch import make_fx
from typing import Dict

def dict_fn(x: Dict[str, Tensor]):
  return x['a'] * 2.0

print(dict_fn({'a': torch.tensor(3.0)}))

linalg_on_tensors_mlir = torch_mlir.compile(
    make_fx(dict_fn)({'a': torch.tensor(3.0)}),
    {'a': torch.tensor(3.0)},
    output_type=torch_mlir.OutputType.LINALG_ON_TENSORS,
    use_tracing=True)
```

[Output](https://gist.github.com/dellis23/1293ee6459ddfb9b52a983aa2a1091bb).  This seems like it's the closest to correct, and perhaps `torch-mlir` isn't passing along the necessary annotations.  I'm not sure though.  This was the same error torch itself gave when trying to trace before I added the type annotations.

I've tried many other variations, but I'm not sure they're sensical.

ramiro050:
[...]

PR #1430, opened 2022-09-28
Reject dictionary inputs when tracing.

dellis23:
The underlying error message was misleading.  See https://github.com/llvm/torch-mlir/issues/1425

### DIFF
--- a/python/test/compile_api/tracing.py
+++ b/python/test/compile_api/tracing.py
@@ -9,9 +9,8 @@ import torch
 
 import torch_mlir
 
+
 class TanhModule(torch.nn.Module):
-    def __init__(self):
-        super().__init__()
     def forward(self, x):
         return torch.ops.aten.tanh(x)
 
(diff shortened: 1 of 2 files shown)


dellis23:
One question: is it safe to have an allowlist that only allows `Tensor` and `TensorPlaceholder`?  If so, I figured this approach would be more flexible and easier to maintain.

ramiro050:
> One question: is it safe to have an allowlist that only allows `Tensor` and `TensorPlaceholder`? If so, I figured this approach would be more flexible and easier to maintain.

Yeah, I think it would be better to throw an error in the beginning of the function if `example_inputs` doesn't match the type `Union[_example_arg, Sequence[_example_arg]]`, rather than having special handling for `dict`s. 

You can check if something is a `Sequence` by doing:

```python
from collections.abc import Sequence

if not isinstance(example_args, Sequence):
    raise ...
```

dellis23:
Okay, I've updated it.  Unfortunately, doing it this way means we have to do the check in a couple of places, since we allow both sequences and non-sequences as inputs.

ramiro050 (approved):
LGTM! Thanks

PR #1432, opened 2022-09-28
Update releases pages after release builds

powderluv:
Remove the periodic update since there are no changes after the nightly and it just clutters the Github Actions.


dellis23:
Thanks! I didn't think about the clutter on that page.

dellis23 (approved):

dellis23:
Unfortunately I think there's a race condition here, whereby the mac builds finish first and prematurely update the releases page, leading to our linux builds being late by a day every time this runs.

image

powderluv:
ah I see. thanks for the heads up. We could update right after each build.

PR #1433, opened 2022-09-28
build: use Github Actions for creating PR

ashay:
Prior to this patch, the CI script used the Github CLI (`gh`) to create
a PR, but doing so had the disadvantage that the created PR was
attributed to Sean.  This patch uses the
`peter-evans/create-pull-request@v4` Github action which allows us to
customize various parts of the PR, including the committer's name.  By
using a specific commiter name, we can also set the PyTorch version
update PRs to be auto-approved.

### DIFF
--- a/.github/workflows/RollPyTorch.yml
+++ b/.github/workflows/RollPyTorch.yml
@@ -1,7 +1,6 @@
 name: Roll PyTorch
 
-on:
-  workflow_dispatch:
+on: workflow_dispatch
 
 jobs:
   build_linux:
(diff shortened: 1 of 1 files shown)


.github/workflows/RollPyTorch.yml
```diff
@@ -1,7 +1,6 @@
 name: Roll PyTorch
 
-on:
-  workflow_dispatch:
+on: workflow_dispatch
```
powderluv, line 3: We probably want to add a schedule for this -- maybe couple hours after the latest PyTorch nightlies are published ?

powderluv: Perfect sounds good to do that in a follow on

powderluv (approved):

silvasean (approved):
Great! Thanks for iterating on this folks :) This is going to really improve our development velocity and quality of life!

PR #1438, opened 2022-09-29
Final changes necessary to auto-update PyTorch version

ashay:
This PR contains two commits.  The first commit pushes the PyTorch version update directly to the main branch, so that we avoid the PR + CI + Approval + Merging steps.  The second commit schedules this workflow to be scheduled daily.

---

I checked that the [workflow](https://github.com/llvm/torch-mlir/actions/runs/3153913606) runs as expected and pushes the change to the main branch if the tests pass (see commit 15e1c45ed63b04cc408361d30c67196679fa353e).

### DIFF
--- a/.github/workflows/RollPyTorch.yml
+++ b/.github/workflows/RollPyTorch.yml
@@ -1,6 +1,9 @@
 name: Roll PyTorch
 
-on: workflow_dispatch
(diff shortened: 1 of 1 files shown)


powderluv (approved):
thank you

Issue #1423, opened 2022-09-27
Lowering failed: unsupported by backend contract: tensor with unknown rank


dellis23:
I'm trying to develop a training example using `functorch`.  I'm starting with a small regression.  My first step before trying to get training working was to make sure I was able to compile the forward pass.

This is the example I'm using (training steps omitted for brevity):

```python
import numpy as np
import torch
from sklearn.datasets import make_regression
from sklearn.model_selection import train_test_split
import functorch

import torch_mlir
from functorch import make_fx

# Dataset
X, y = make_regression(n_features=3)
X, X_test, y, y_test = train_test_split(X, y)

# Weights
params = {
    'w': torch.zeros(X.shape[1:], dtype=torch.double),
    'b': torch.tensor(0., dtype=torch.double),
}

def forward(w, b, X):
    return torch.matmul(X, w) + b

linalg_on_tensors_mlir = torch_mlir.compile(
    make_fx(forward)(params['w'], params['b'], torch.tensor(X)),
    (params['w'], params['b'], torch.tensor(X)),
    output_type=torch_mlir.OutputType.LINALG_ON_TENSORS,
    use_tracing=True)
```

This gives me the following error:

```python
TorchMlirCompilerError: Lowering TorchScript IR -> Torch Backend IR failed with the following diagnostics:
error: unsupported by backend contract: tensor with unknown rank
note: see current operation: %1 = "torch.tensor_static_info_cast"(%arg0) : (!torch.vtensor<[3],f64>) -> !torch.vtensor<*,f64>
note: this is likely due to a missing shape transfer function in shape_lib_gen.py
```

silvasean:
@ramiro050 can you PTAL at this?

dellis23:
I stripped it down a bit more to make it easier to reproduce:

```python
def forward(w, X):
    return torch.matmul(X, w)

args = (torch.tensor([1., 3., 4.], dtype=torch.double),
        torch.ones((75, 3), dtype=torch.double))
print(forward(*args))

print("Compiling with Torch-MLIR")
linalg_on_tensors_mlir = torch_mlir.compile(
    make_fx(forward)(*args),
    args,
    output_type=torch_mlir.OutputType.LINALG_ON_TENSORS,
    use_tracing=True)
```

ramiro050:
[...]

PR #294, opened 2021-08-30
Sigmoid lowering to linalg

dan-garvey:
Adds a pretty straightforward lowering for Sigmoid following the existing conventions for relu and tanh. 

Would it be appropriate to mirror the lit tests for tanh in ./test/Conversion/TorchToLinalg/elementwise.mlir and /test/Dialect/Torch/refine-types.mlir? Any other tests required for this kind of patch?

PR #1444, opened 2022-09-30
Add support for mv.

dellis23:
See #1423.

Please let me know if there are better places to put anything or conventions I missed.  One thing I'm unclear on: I see in the codebase a lot of explicit `__init__`s with calls to super that do nothing else.  Is there a reason for this I'm missing?

Also, I ran into issues with what I assume is an op that was deleted (frobenius norm).  I had to do a lot of manual surgery to get things working again.  Is there a better procedure for this?  I can do that in a separate commit if so and rebase this off of it so this is cleaner.

Also: should I do the same for other dialects?  I see a number of other dialects for matmul, but I only see `TorchToLinalg` in [the PR I was referencing](https://github.com/llvm/torch-mlir/pull/294/).  I figured it would be good to get feedback on this first before doing it.

I tried to move the new test case to its own file, but I'm getting an import error when I add an entry to the import list in `test_suite/__init__.py`:

```
ImportError: cannot import name 'mv' from 'torch_mlir_e2e_test.test_suite' (/usr/local/google/home/danielellis/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/__init__.py)
```

I really have no clue what it could be.  The imports are the exact same as the other files in the directory and don't appear circular.  The file is definitely present on disk.  I also tried changing the name of the file.  It's the same error I get when I put in a bogus name.  Is there some magic I'm missing?


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -415,6 +415,42 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
 };
 } // namespace
 
+namespace {
+class ConvertAtenMvOp : public OpConversionPattern<AtenMvOp> {
```
ramiro050, line 419: I think it would make more sense to add a decomposition of this op into `AtenMatmulOp`, since that op performs this same handling:

https://github.com/llvm/torch-mlir/blob/9dd5ae8239e2d45fd8959b8b014bf0c76be16d31/lib/Conversion/TorchToLinalg/Linear.cpp#L217

The decomposition would happen in the `DecomposeComplexOps.cpp` file

ramiro050 (changes requested):

ramiro050:
> One thing I'm unclear on: I see in the codebase a lot of explicit __init__s with calls to super that do nothing else. Is there a reason for this I'm missing?

This is is probably due to someone adding the `__init__` at some point on a test and everyone else just copying to match style. If they are not needed, we can remove all of them in a separate PR.

> Also, I ran into issues with what I assume is an op that was deleted (frobenius norm). I had to do a lot of manual surgery to get things working again. Is there a better procedure for this? I can do that in a separate commit if so and rebase this off of it so this is cleaner.

Can you elaborate more on the error you were seeing? This op should be working.

> should I do the same for other dialects?

Adding support for other dialects is always optional. The expectation is that people simply add what they need. (However, if you add it as a decomposition, other dialects that already support `AtenMatmulOp` will get your op for free!)

> I tried to move the new test case to its own file, but I'm getting an import error when I add an entry to the import list in test_suite/__init__.py

The [main](https://github.com/llvm/torch-mlir/blob/9dd5ae8239e2d45fd8959b8b014bf0c76be16d31/e2e_testing/main.py#L10) file is in a different directory than the test suite, so the import of `torch_mlir_e2e_test` is made from the python bindings generated when you build torch-mlir rather than the `torch_mlir_e2e_test` directory in the source code. You need to rebuild torch-mlir to have your file copied over to the python bindings directory.

dellis23:
> Can you elaborate more on the error you were seeing? This op should be working.

When I ran one of the scripts (I believe it was `update_torch_ods.sh`), it removed `AtenFrobeniusNormDimOp` from the output file.  Then when I tried building / continuing on, I kept getting compilation errors.  I resolved those by removing the various parts related to that op.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -399,7 +400,6 @@ def emit_with_mutating_variants(key, **kwargs):
     emit("aten::nll_loss_backward : (Tensor, Tensor, Tensor, Tensor?, int, int, Tensor) -> (Tensor)")
     emit("aten::bincount : (Tensor, Tensor?, int) -> (Tensor)")
     emit("aten::linalg_vector_norm : (Tensor, Scalar, int[]?, bool, int?) -> (Tensor)")
-    emit("aten::frobenius_norm.dim : (Tensor, int[], bool) -> (Tensor)")
```
ramiro050, line 402: Did you remove this line before or after running `./build_tools/update_torch_ods.sh`? This is the line that generates the ODS for the op. Also, when I run thing locally on main, I don't see this op disappear.

dellis23: I hadn't removed anything; I didn't have any good reason to pick on frobenius :P

Perhaps it was something temporary with us running against head pytorch?  If it's working fine for you at head now I'll try clearing it all out and rerunning once I'm able to move over to the decomposition approach.

ramiro050: Ah, maybe it was a temporary thing with PyTorch. There is a commit that makes changes to `frobenius_norm` that would appear when running `update_torch_ods.sh`: https://github.com/pytorch/pytorch/commit/1c0f0b33a0e013d6ec162cf488ff7643c4ffa33e

The commit has been merged into PyTorch and reverted twice, so this might come back in the future.

dellis23:
Closing this in favor of https://github.com/llvm/torch-mlir/pull/1453

PR #1455, opened 2022-10-03
LTC aten.where support

GlebKazantaev:
Removed `where` op from ltc blacklist. `torch.where` codegen successfully compiles and converts to torch-mlir.

@antoniojkim @ke1337

### DIFF
--- a/test/python/importer/jit_ir/node_import/prim.py
+++ b/test/python/importer/jit_ir/node_import/prim.py
@@ -147,6 +147,19 @@ def prim_min(x: int):
 def prim_max(x: int):
     return max(x), max(x,x), max(x, x, x)
 
+# CHECK-LABEL:   func.func @__torch__.prim_where(
+# CHECK-SAME:                             %[[ARG:.*]]: !torch.tensor) -> !torch.tensor {
+# CHECK:           %[[A:.*]] = torch.constant.int 1
+# CHECK:           %[[B:.*]] = torch.constant.float 1.000000e+00
+# CHECK:           %[[SCALAR:.*]] = torch.aten.lt.Scalar %[[ARG]], %[[A]] : !torch.tensor, !torch.int -> !torch.tensor
+# CHECK:           %[[RET:.*]] = torch.aten.where.ScalarOther %[[SCALAR]], %[[ARG]], %[[B]] :
+# CHECK-SAME:           !torch.tensor, !torch.tensor, !torch.float -> !torch.tensor
+# CHECK:           return %[[RET]] : !torch.tensor
+@mb.import_function
+@torch.jit.script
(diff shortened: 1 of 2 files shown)


test/python/importer/jit_ir/node_import/prim.py
```diff
@@ -147,6 +147,19 @@ def prim_min(x: int):
...
+# CHECK-SAME:                             %[[ARG:.*]]: !torch.tensor) -> !torch.tensor {
+# CHECK:           %[[A:.*]] = torch.constant.int 1
+# CHECK:           %[[B:.*]] = torch.constant.float 1.000000e+00
+# CHECK:           %[[SCALAR:.*]] = torch.aten.lt.Scalar %[[ARG]], %[[A]] : !torch.tensor, !torch.int -> !torch.tensor
+# CHECK:           %[[RET:.*]] = torch.aten.where.ScalarOther %[[SCALAR]], %[[ARG]], %[[B]] :
+# CHECK-SAME:           !torch.tensor, !torch.tensor, !torch.float -> !torch.tensor
+# CHECK:           return %[[RET]] : !torch.tensor
+@mb.import_function
+@torch.jit.script
+def prim_where(x):
```
silvasean, line 160: You don't need this test. all the "aten.*" ops are handled systematically so there is no need for a special test. (this file is for testing "prim.*" ops, which would look like `torch.prim.*` when imported (note this one is `torch.***aten***.where.ScalarOther`)

silvasean (approved):

test/python/importer/jit_ir/node_import/prim.py
GlebKazantaev: I removed the test. 
What does it mean `handled systematically` ? Is there any automated validation for regular aten ops?

GlebKazantaev:
@silvasean please start validation

GlebKazantaev:
@ke1337 @antoniojkim @ashay could you please merge this PR? Look like I don't have rights because Im a first time contributor to this repo.

test/python/importer/jit_ir/node_import/prim.py
GlebKazantaev: @silvasean could you please clarify about `handled systematically`?

silvasean: 99% of ops are imported by this line: https://github.com/llvm/torch-mlir/blob/d3cc3f1affe3c58d03feb07911938e83a3faa37e/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/node_importer.cpp#L109

Which basically just takes the op name as a string and puts it in an MLIR operation. So we don't need separate testing for each op.

PR #1456, opened 2022-10-03
Disable auto-update of PyTorch version until CI script stabilizes

ashay:
Instead of letting the auto-update script either fail because of script
errors or letting it commit bad versions, this patch makes the update
process manual, for now.  Once the script stabilizes, I will its
re-enable periodic execution.

This PR also reverts fixes several problems in CI.


powderluv (approved):

pashu123:
@ashay Can we switch to this version https://github.com/llvm/torch-mlir/commit/f32869b854534b36dcd0fc9ac33813309d1b39d6 the build started passing for this commit.

Issue #1445, opened 2022-09-30
publish torch-mlir binaries to PyPi


powderluv:
Now that the CI is mature we can turn on auditwheel in the nightly builder and publish to PyPi. Someone with admin rights (@silvasean or @dellis23 or @stellaraccident ) will have to put in the PyPI API key in as a Github secret and it is only a few lines of code to add to the release builder.

dellis23:
Okay, we have a pypi token as a github actions secret (`PYPI_API_TOKEN`).  Do you want to add it to the release builder @powderluv since you're more familiar with that process?  If not, I can look into it.

powderluv:
Please do since I wont be able to test end to end without PyPi access. I think this https://github.com/marketplace/actions/pypi-publish would help. 

We will have to uncomment this line to run audit_wheel:

https://github.com/llvm/torch-mlir/blob/eda18e351c7f0a45dd78bf8fc43efb570301c3b3/build_tools/python_deploy/build_linux_packages.sh#L145

dellis23:
Alright, I've made a lot of progress on this, but I've hit a snag we'll probably have to discuss together:

```shell
WARNING  Error during upload. Retry with the --verbose option for more details. 
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         File too large. Limit for project 'torch-mlir' is 100 MB. See          
         https://pypi.org/help/#file-size-limit for more information.     
```

See [here](https://github.com/llvm/torch-mlir/actions/runs/3250589985/jobs/5334464719#step:11:115).  It looks like the file is `216.6 MB`.

PR #1461, opened 2022-10-04
Publish releases to PyPI after build

dellis23:


dellis23:
See https://github.com/llvm/torch-mlir/issues/1445

powderluv (approved):
I haven't tested with audit wheel before so hopefully it just works -- but if not we may need to fix to make sure it passes.

.github/workflows/buildRelease.yml
```diff
@@ -51,6 +56,11 @@ jobs:
         GITHUB_TOKEN: ${{ secrets.WORKFLOW_INVOCATION_TOKEN }}
       with:
         release_id: ${{ github.event.inputs.release_id }}
+    - name: Publish to PyPI
+      if: github.event.inputs.release_id != ''
+      uses: pypa/gh-action-pypi-publish@v1.5.1
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
powderluv, line 63: Do we have to specify the .whl package ? We don't want to push the torch package ? Just the torch-mlir one.

dellis23:
> I haven't tested with audit wheel before so hopefully it just works -- but if not we may need to fix to make sure it passes.

Looks like I need to do something else to get it to work:

```
usage: auditwheel [-h] [-V] [-v] command ...
auditwheel: error: cannot access /wheelhouse/torch_mlir-*-cp39-cp39-linux_x86_64.whl. No such file
```

.github/workflows/buildRelease.yml
dellis23: The only other argument I see is `packages_dir`.  I can try that, though I'm not sure yet what the directory structure looks like on disk and whether these packages will just be in the same directory.  On a side note, the iteration time on this is _really_ slow, so this might take a while.

dellis23:
@powderluv Do you have any tips on testing the release process in a PR?  The issue I'm having: I'm doing a full build, but because `release_id` isn't set, it skips the publishing part.  See [here](https://github.com/llvm/torch-mlir/actions/runs/3192242298/jobs/5209494585).  I could just take a risk and merge / iterate on `main`, but I'm worried about this making the build red (well, _more_ red) for everyone.

powderluv:
just use the "Oneshot build" and select the `pypi-release` branch. It will do a full build for you and do a release.

powderluv (approved):

powderluv:
For the large file issue -- we can comment out auditwheel for now and see if you can upload it. Auditwheel copies the dylibs from pytorch so the file size increases. 

Long term we have to shed the weak-link into Pytorch and build statically just what we need (not all Aten ops)

powderluv (approved):
Eventually we need to find how to get large packages via PyPi (like Pytorch has a 2GB cuda install)

dellis23:
It looks like there's a way to request an exemption on the file size limit: https://pypi.org/help/#file-size-limit

I can file a request.  We need to include the total size of the project (I assume 200MB, based on the output I had from before?) and a reason for the additional size.  Do you have a compelling statement I could use for the latter part?  I don't know enough about everything going into the build to be able to say anything other than "the project is big."

powderluv:
How about : 

`torch-mlir` depends on C++ libraries that are included as part of its wheel to be able to provide maximum compatibility flexibility for the end users.

powderluv (approved):

powderluv:
Thank you

PR #1465, opened 2022-10-05
Publish releases page after both mac and linux builds.

dellis23:
Mac was finishing first, causing linux releases to be lagged a day behind.

Related discussion in https://github.com/llvm/torch-mlir/pull/1432


dellis23:
I think the steps outlined [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) will let us avoid a bit of clutter and only publish once.

silvasean (approved):

PR #1466, opened 2022-10-05
[LINALG] Make `AtenMaxDimOp` use `arith.maxf` to calculate maximum

ramiro050:
This commit updates the linalg conversion of `AtenMaxDimOp` to use `arith.maxf` instead of `arith.select` to calculate the maximum. This allows better vectorization further downstream, since the operation can be converted to a simple max reduction when the `indices` result
is not used. See: https://github.com/iree-org/iree/issues/10666.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Reduction.cpp
+++ b/lib/Conversion/TorchToLinalg/Reduction.cpp
@@ -152,12 +152,10 @@ public:
               nestedLoc, oldIndex.getType(),
               rewriter.create<linalg::IndexOp>(loc, dim));
 
-          Value predicate;
-          if (inElementType.isa<mlir::FloatType>())
-            predicate = rewriter.create<arith::CmpFOp>(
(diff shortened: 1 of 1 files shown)


lib/Conversion/TorchToLinalg/Reduction.cpp
```diff
@@ -152,12 +152,10 @@ class ConvertAtenMaxDimOp : public OpConversionPattern<AtenMaxDimOp> {
               nestedLoc, oldIndex.getType(),
               rewriter.create<linalg::IndexOp>(loc, dim));
 
-          Value predicate;
-          if (inElementType.isa<mlir::FloatType>())
```
ramiro050, line 156: This `if` is not needed since it is already checked in the beginning of the function:

https://github.com/llvm/torch-mlir/blob/c14ce36f80b536e4306f5d4a5b71177d2d20c53c/lib/Conversion/TorchToLinalg/Reduction.cpp#L83

qedawkins (approved):
LGTM!

PR #1469, opened 2022-10-06
[MLIR][TORCH] Add E2E support for `aten.[div.int|bitwise_or.Tensor]` ops

Shukla-Gaurav:
This commit adds lowering of `aten.div.int` and `aten.bitwise_or.Tensor` ops. Both these ops are required in order to support bloom_560m model.


test/Conversion/TorchToArith/basic.mlir
```diff
@@ -167,6 +167,21 @@ func.func @torch.aten.mul.int(%arg0: !torch.int, %arg1: !torch.int) -> !torch.in
...
+// CHECK-SAME:                            %[[LHS:.*]]: !torch.int,
+// CHECK-SAME:                            %[[RHS:.*]]: !torch.int) -> !torch.float {
+// CHECK:          %[[LHS_I64:.*]] = torch_c.to_i64 %[[LHS]]
+// CHECK:          %[[RHS_I64:.*]] = torch_c.to_i64 %[[RHS]]
+// CHECK:          %[[LHS_F64:.*]] = arith.sitofp %[[LHS_I64]] : i64 to f64
+// CHECK:          %[[RHS_F64:.*]] = arith.sitofp %[[RHS_I64]] : i64 to f64
+// CHECK:          %[[DIV:.*]] = arith.divf %[[LHS_F64]], %[[RHS_F64]] : f64
+// CHECK:          %[[OUT:.*]] = torch_c.from_f64 %[[DIV]]
+// CHECK:          return %[[OUT:.*]] : !torch.float
+func.func @torch.aten.div.int(%arg0: !torch.int, %arg1: !torch.int) -> !torch.float {
```
ramiro050, line 180: This test is not needed. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

ramiro050 (approved):
I just have a few style comments, but other than that LGTM

vivekkhandelwal1:
@Shukla-Gaurav You need to add the test `DivIntModule_basic` to `LTC_XFAIL_SET`

PR #1470, opened 2022-10-09
[MLIR][TORCH] Add e2e support for `aten.Mish` op

Abhishek-Varma:
-- This commit adds e2e support for `aten.Mish` op.
-- `aten.Mish` op is decomposed as following :-
    Mish(x) = x * Tanh(Softplus(x))

Signed-off-by: Abhishek Varma <avarma094@gmail.com>


Abhishek-Varma:
Hi @powderluv @silvasean - I'm a first time contributor here so can you point me to a set of maintainers/reviewers who can initiate the workflows/reviews?
Thanks. :)

python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -826,6 +826,29 @@ def ElementwiseMulTensorIntModule_basic(module, tu: TestUtils):
...
+        super().__init__()
+        self.mish = torch.nn.Mish()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, x):
+        return self.mish(x)
```
pashu123, line 841: We can directly call the aten op instead of initializing the nn.module. I guess it's `torch.ops.aten.mish(x)`.

pashu123 (approved):
Nice work! Thanks.

python/torch_mlir_e2e_test/test_suite/elementwise.py
Abhishek-Varma: Thanks for this - I was initially trying `torch.ops.aten.Mish(x)` hence wasn't able to get through.

vivekkhandelwal1 (approved):
LGTM!

python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -826,6 +826,28 @@ def ElementwiseMulTensorIntModule_basic(module, tu: TestUtils):
...
+        None,
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, x):
+        return torch.ops.aten.mish(x)
+
+
+@register_test_case(module_factory=lambda: ElementwiseMishModule())
+def ElementwiseMishModule_basic(module, tu: TestUtils):
+    module.forward(2 * tu.rand(5, 3) - 0.5)
```
ramiro050, line 845: nit: use `low=` and `high=` to specify range for `tu.rand`. See: https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#things-to-consider-when-creating-new-tests

ramiro050 (changes requested):

python/torch_mlir_e2e_test/test_suite/elementwise.py
Abhishek-Varma: Hi @ramiro050 , so I referred the said link and in the example [above](https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#test-case-function) I see the use of `tu.rand` similar to the way I used here in the PR and other test cases we see in `elementwise.py` - since we were to follow similar convention/style of the file we're adding the test case into, I tried doing the same.

Most of the `tu.rand` invocations we see in the test file are using the default value for `low` and `high` parameters.

I believe unless we use the non-default values we shouldn't have to explicitly mention the parameters and their values while invoking the function - a similar invocation of `tu.randint` can be seen in the example [here](https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#test-case-function).

Let me know if a change is still warranted else we can aim to get this PR in.

ramiro050: Hi @Abhishek-Varma, yeah this file contains some of earliest tests added to torch-mlir before there was support for specifying `low` and `high` in `tu.rand`, so people resorted to multiplying and subtracting to change the range. However, for new tests, it is better to use `low` and `high` than the external linear transformation -- `2 * x - 0.5` in this case -- since it makes it much easier to tell what range is being tested.

Abhishek-Varma: I understand, thanks @ramiro050 - I've addressed this now.
You may check and approve/initiate the workflow if all looks good. :)

python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -826,6 +826,28 @@ def ElementwiseMulTensorIntModule_basic(module, tu: TestUtils):
...
+        None,
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, x):
+        return torch.ops.aten.mish(x)
+
+
+@register_test_case(module_factory=lambda: ElementwiseMishModule())
+def ElementwiseMishModule_basic(module, tu: TestUtils):
+    module.forward(2 * tu.rand(5, 3, low=-3.0, high=3.0) - 0.5)
```
ramiro050, line 845: I think my comment might not have been clear. Sorry about that. What I meant was that we should remove the `2 *` and the `- 0.5` and simply use the `low` and `high` to specify the range. (the `2 *` and `- 0.5`) are simply modifying the range to be twice as big and shifted 0.5 to the left, which are things that can be achieved with with `low` and `high`.

In short, this line should be

```python
module.forward(tu.rand(5, 3, low=-3.0, high=3.0))
```

ramiro050 (changes requested):

python/torch_mlir_e2e_test/test_suite/elementwise.py
Abhishek-Varma: Oh all right, I understand what you meant to say now and have updated the test case - you may re-review. :) Thanks.

ramiro050 (approved):
LGTM! Thanks!

Abhishek-Varma:
Can someone merge this in? I don't have write access to this repository as of now. Thanks. :)

PR #1472, opened 2022-10-10
Disable LTC and update PyTorch

ashay:
When built using PyTorch's master branch, the LTC code has been failing         
to build for a few days.  As a result, the PyTorch version referenced by         
Torch-MLIR is stalled to the one from October 4th.

This PR contains two patches: the first disables LTC and the second
advances the PyTorch version.


powderluv:
if required we can xfail to roll it forward with an issue tracker

powderluv (approved):
@antoniojkim We want to try and get tonights build successful. Happy to enable LTC again once it is fixed.

powderluv:
@pashu123 please re-enable the tests once it is fixed.

pashu123:
Thanks, @ashay for filing it as an issue. I will take a look at it.

PR #1475, opened 2022-10-11
build: update llvm tag to 438e5918

ashay:
Replaced `linalg::InitTensorOp` with `tensor::EmptyOp`
(https://reviews.llvm.org/D135129)


ramiro050:
The issue with `_LogSoftmaxModuleStable_basic` seems to be caused by a bug in new changes made to the `TosaToLinalg` pass in LLVM. I've made a fix PR for it here: https://reviews.llvm.org/D135828. Unfortunately, this means the LLVM bump in this PR will not be possible unless we disable that test, but I personally don't like the idea of new changes breaking already supported and tested paths. @silvasean, any thoughts on this?

cc: @jpienaar

silvasean:
Waiting until Monday seems fine to me. Thanks for the fixes Ramiro!

powderluv:
FYI downstream IREE has moved past this point in LLVM so we would need this next LLVM bump for iree-torch, SHARK etc to work.

PR #1476, opened 2022-10-11
[LINALG] Add lowering for aten::round op.

pashu123:
-- Added the lowering for aten::round op.
-- Added the folding for integer cases.


ramiro050 (approved):
LGTM!

PR #1481, opened 2022-10-11
torch_mlir.compile: allow custom backend_legal_ops set

### LINKED ISSUE #1418: [FeatureRequest] Allow setting legal ops
Currently the legal op set is binding with backend(https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/__init__.py#L131), however, even for the same IR backend like MHLO, two developer groups may have their own legal op set. It would be better to allow user to set the legal op directly via the `compile` API instead of backend-based.

silvasean:
Allow customizing `backend_legal_ops` for "torch" output type, since we don't know which backend will be used (it might be a custom backend). We don't allow customizing the `backend_legal_ops` for the other output types (Linalg, TOSA, MHLO) since those backends control their set of legal ops directly.

Fixes #1418


ZihengJiang (approved):
LGTM. Thanks @silvasean !

PR #1482, opened 2022-10-11
Fix build release workflow.

dellis23:


powderluv (approved):
how did it work before ? :D

dellis23:
Ha, I guess I just didn't notice the failure?  It worked in the PR originally because I had them under both the linux and mac section.  But when I moved to try to "optimize" it, there was no longer a `runs-on` section.

PR #1485, opened 2022-10-11
Replace linear transformation with `low` and `high` in test inputs

ramiro050:
This commit replaces test inputs that were being linearly transformed by multiplying and adding/subtracting to the input tensor with inputs that use the `low` and `high` keyword arguments instead.


python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -351,7 +351,7 @@ def forward(self, x):
 
 @register_test_case(module_factory=lambda: ConstantPad2dStaticModule())
 def ConstantPad2dStaticModule_basic(module, tu: TestUtils):
-    module.forward(tu.rand(1, 1, 20, 20) - 0.5)
+    module.forward(tu.rand(1, 1, 20, 20, low=-1))
```
ramiro050, line 354: I'm aware that this replacement results in a different range. However, the only thing special about the `-0.5` is that it generated an equal number on average of negative and positive values. Using `low=-1` achieves this same goal without having to also specify `high=`, which would be the case if I had used `low=-0.5` instead.

vivekkhandelwal1 (approved):
LGTM! But the CI is failing, you need to fix the test.

ramiro050:
> LGTM! But the CI is failing, you need to fix the test.

Yeah, i'm currently investigating it. Seems like a lack of precision issue

ramiro050:
I've made the range for the `gelu` test to be from `-0.5` to `0.5` to make it pass. I've opened an issue https://github.com/llvm/torch-mlir/issues/1497 to track the precision error, since that problem is outside the scope of this PR.

PR #1487, opened 2022-10-12
Remove unused input tensor from linalg.generic in aten.convolution

ramiro050:
This commit removes the `weight` tensor from the inputs of one of the `linalg.generic` ops generated by the `aten.convolution` linalg lowering, since the indexed values are not actually used by the body of the `linalg.generic`. Moreover, in general the `weight` tensor does not have the same shape as the output tensor of the `linalg.generic`, so both tensors being indexed by the same indexing maps is wrong.


qedawkins (approved):
LGTM

PR #1489, opened 2022-10-13
build: retain object files after building PyTorch from source

ashay:
With the goal of improving time spent in CI, this patch retains the
object files built while compiling PyTorch from source.  By doing so,
the build system would recognize that the object files are newer than
the source files, thus skipping a full rebuild.
    
Although this patch does not make any _functional_ changes, it still is
speculative.  So if we do not observe any significant improvement in
time spent in CI over the next few days, I will revert this patch so
that we can save some disk space.

### DIFF
--- a/build_tools/build_libtorch.sh
+++ b/build_tools/build_libtorch.sh
@@ -122,13 +122,13 @@ package_pytorch() {
   fi
 
   # Copy over all of the cmake files
-  mv build/lib*/torch/share     libtorch/
-  mv build/lib*/torch/include   libtorch/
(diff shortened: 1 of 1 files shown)


powderluv:
I think the same disk being attached is remote -- however I do think it is worth trying this out in case the `mv` vs `cp` has some impact. Originally it was a `cp` and I made it `mv` to reduce clutter in the logfiles. If we can make the `cp` quiet that would be even better.

powderluv:
I think it is still worth running the CI run fully to test if the cp vs mv does anything to the atime to affect the cache given you already have the change.

PR #998, opened 2022-06-30
Integrate Functionalization Pass

antoniojkim:
The functionalization pass PR has finally landed in upstream PyTorch master. The changes in this PR are what's required to integrate them into the MLIR LTC backend.

@silvasean This is perhaps the last PR required to be merged into the staging branch before we can start considering merging the staging branch into `main`

CC: @henrytwo @ke1337

PR #1494, opened 2022-10-14
New ops support & enhancements

GlebKazantaev:
### Description
* Fixed functionalization issue with `torch.arange(...,device='lazy').expand(...)` sequence of operations. This case appears inside positional embedding layers during indices initialization. The problem is that in PyTorch LTC they have limited number of primitives they currently support for functionalization. So to have this case working we have to follow TS approach with `arange.start_out` primitive that covers all possible `torch.arange` variations. That's why all arange primitives in `ltc_backend.yaml` are in blacklist, but `arange.start_out` marked for codegen. 
* Fixed functionalization issue for `torch.ones`. The reason why it was disabled in config is the same as for torch.aten. But in case of `torch.ones` it fallback to `fill_` primitive.
* Moved back support for non-native expand operation. Unfortunately this operation is still in use in TS and [torch-mlir backend](https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/csrc/base_lazy_backend/mlir_native_functions.cpp#L82). For the reference: [old version](https://github.com/llvm/torch-mlir/pull/998/files#diff-4d898b4339dac5a2d259c9eceb9ed44bf1fbc52c92e4854b73f43aa34db3721bL279)
* Added code generation for stack operation. Unfortunately it is not enough to make it work because of the bug in PyTorch `c10::ListType::isSubtypeOfExt` implementation. [PR to the PyTorch](https://github.com/Cerebras/pytorch/pull/18).
* Added missing `aten.FloatImplicit` primitive which is required by [this change in PyTorch](https://github.com/pytorch/pytorch/pull/86692). Also it is good to have all Implicit like ops in torch-mlir.
* Added shape inference for `aten.where` operation.


python/torch_mlir/csrc/base_lazy_backend/shape_inference.cpp
```diff
@@ -42,6 +42,12 @@ std::vector<torch::lazy::Shape> compute_shape_hardtanh(
   return {Shape(self.scalar_type(), self.sizes().vec())};
 }
 
+std::vector<torch::lazy::Shape> compute_shape_where(
```
silvasean, line 45: Is there a reason we are not upstreaming these? This is supposed to be temporary per the TODO above -- is there a plan to upstream these?

silvasean (approved):

python/torch_mlir/csrc/base_lazy_backend/shape_inference.cpp
antoniojkim: the TS backend does not support the `where` op and thus we can't upstream it. Only when the TS backend does support the `where` op would we consider upstreaming it. Or better yet, if they do provide support for the `where` op, then this will probably mean that they already have an implementation for `compute_shape_where` and we can deprecate this one

GlebKazantaev: Originally the plan was to upstream these functions. But in PyTorch they are now working on transition to structured kernels which should replace all this custom shape functions. So I believe that in nearest future we will just remove all these functions.

PR #1503, opened 2022-10-18
[docs] Add long-term roadmap

silvasean:
Add a roadmap covering expected project evolution over the next 1-2 years.


silvasean:
cc @dellis23 @eric-k256

silvasean:
Hey folks, here is a long-term roadmap which hopefully covers some of the main ways that we expect the project will evolve in the future. I would love to hear your feedback!

powderluv:
Really nice write up. Thanks for putting it together.

powderluv (approved):

docs/long_term_roadmap.md
```diff
@@ -0,0 +1,221 @@
...
+programs into the compiler.
+
+- [TorchDynamo](https://github.com/pytorch/torchdynamo) - TorchDynamo uses
+  tracing-JIT-like techniques and program slicing to extract traces of tensor
+  operations, which can then be passed to lower-level compilers. It works
+  seamlessly with unmodified user programs.
+- [FuncTorch](https://github.com/pytorch/functorch) - FuncTorch is basically JAX
+  for PyTorch. It requires manual program tracing and slicing, but that is
+  actually important for users since it gives them direct control over various
+  important transformations, such as `grad` and `vmap`.
```
antoniojkim, line 71: Does this mean that LTC support will eventually be deprecated in Torch-MLIR?

silvasean: As long as LTC is not deprecated by PyTorch and we have users that rely on it, then I think we should keep supporting it.

But as Will said [here](https://dev-discuss.pytorch.org/t/skipping-dispatcher-with-lazytensor/634/2?u=_sean_silva), it seems like on the 1-2yr time horizon LTC will be more an implementation detail under TorchDynamo for users that already have compilers written using LTC. That is, LTC is basically just a way to convert a TorchDynamo FX graph into LTC graphs, for users that have toolchains written against LTC graphs. But that won't make much sense for us, because we convert to MLIR in the end no matter what. That is, in the future going `TorchDynamo FX graph -> LTC Graph -> MLIR` can just be replaced by the direct `TorchDynamo FX graph -> MLIR` path. So I predict that in the 1-2yr time horizon we will be able to delete our LTC support.

The only blocker there would be if end-users have `device='lazy'` hardcoded into their code. That will require a migration plan for current LTC-based toolchains onto TorchDynamo. This migration will improve the end-user experience since TorchDynamo is more seamless, but it is a end-user-impacting migration nonetheless and we will want to phase it appropriately with the community. Do you have a sense of how difficult such a migration would be for you folks?

silvasean: I added this to the doc.

antoniojkim: > That is, in the future going TorchDynamo FX graph -> LTC Graph -> MLIR can just be replaced by the direct TorchDynamo FX graph -> MLIR path. So I predict that in the 1-2yr time horizon we will be able to delete our LTC support.

If we can get the MLIR graph without any decompositions, then we can probably make this work.

> Do you have a sense of how difficult such a migration would be for you folks?

Having said that, our infrastructure is pretty heavily reliant on the LTC structure. It will take time for us to experiment with what it might take to be able to use TorchDynamo as our frontend instead. For that reason, I think keeping the `TorchDynamo FX graph -> LTC Graph -> MLIR` path would be very useful until such a time

silvasean:
CC @Chillee -- could you PTAL at the "refactoring the frontend" part of long_term_roadmap.md?

docs/long_term_roadmap.md
silvasean: Yes, it will require design work for sure, and we will want to keep the lights on until the migration can happen.

docs/long_term_roadmap.md
```diff
@@ -0,0 +1,242 @@
...
+  alignment with the StableHLO effort. One byproduct of TCP that is expected to
+  be very valuable is to incorporate the robust dynamic shape strategy from
+  Linalg into an MHLO-like dialect, and there is a strong desire from StableHLO
+  developers to adopt this once proven in TCP.
+-
+  [PrimTorch](https://dev-discuss.pytorch.org/t/tracing-with-primitives-update-0/577) -
+  is an effort on the PyTorch side to decompose PyTorch operators into a
+  smaller set of primitive ops. This effort could effectively reduce the op
+  surface area at the Torch-MLIR level a lot, which would make the duplication
+  across backends less of anissue. But it still leaves open a lot of
```
ramiro050, line 133: "anissue" -> "an issue"

docs/long_term_roadmap.md
```diff
@@ -0,0 +1,242 @@
...
+  variance on the MLIR side. This means that MLIR will be the "more frozen" of
+  the two major dependencies (PyTorch and MLIR).
+- We would like Torch-MLIR to be hooked into the PyTorch CI systems, and
+  generally be more tightly integrated with the PyTorch development process
+  (this includes things like packaging as well).
+
+### Co-design
+
+Many users of MLIR are developing advanced hardware or software systems, and
+often these require information from the frontend that beyond what PyTorch
```
ramiro050, line 202: "that beyond" -> "that is beyond"

ramiro050 (approved):
This is great!

docs/long_term_roadmap.md
```diff
@@ -0,0 +1,242 @@
...
+
+- Future A: We concentrate the bottleneck step in the "Backend contract ->
+  StableHLO/MHLO/TCP" lowering path. This gets us a stable output for most
+  things. The cascaded/transitive lowerings then let us do O(100) lowerings from
+  then on down. (exact details are not worked out yet, and depend on e.g. TCP
+  adoption, etc.)
+- Future B: PrimTorch concentrates the bottleneck step on the PyTorch side.
+
+These two efforts synergize, but the need for cascaded lowerings is much less if
+PrimTorch solves the decomposition problem on the PyTorch side.
```
Chillee, line 154: So, what are the main concerns for us solving this problem on the PyTorch side? Or is it just the lack of coverage/maturity of this project?

Chillee (commented):
Looks good!

Excited that much of the work we're doing now is so useful for you guys :)

Some components that still seem quite in flux (i.e. we should monitor/discuss them further).

1. Control-flow operators: We're starting to add control-flow operators in PyTorch. However, there's a lot of questions surrounding them, particularly wrt dynamic shapes (can you branch on shape computation? Can the different branches have different shapes? etc.)

2. Data-dependent output shapes/shape constraints: I'd like to know more about how the MLIR ecosystem plans on handling data-dependent output shapes. I know a bit about XLA's plan, but what about the rest of the ecosystem?

docs/long_term_roadmap.md
```diff
@@ -0,0 +1,242 @@
...
+infrastructure that bridges from the PyTorch eager world into compiler-land.
+PyTorch has two main projects that together cover almost all user use cases and
+provide a technically sound, high quality-of-implementation path from user
+programs into the compiler.
+
+- [TorchDynamo](https://github.com/pytorch/torchdynamo) - TorchDynamo uses
+  tracing-JIT-like techniques and program slicing to extract traces of tensor
+  operations, which can then be passed to lower-level compilers. It works
+  seamlessly with unmodified user programs.
+- [FuncTorch](https://github.com/pytorch/functorch) - FuncTorch is basically JAX
```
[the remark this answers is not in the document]
silvasean: Thanks for the info! From the perspective of this diagram, we are mostly focusing on the ways that users interact with the compiler, the two being 1) Seamless with TorchDynamo or 2) JAX-like explicit capture with FuncTorch. Going into the specifics of mentioning AOTAutograd is probably one level too deep for this doc.

docs/long_term_roadmap.md
silvasean: I think it is coverage/maturity, together with the right kinds of supported, documented user-facing (user = toolchain implementor here) controls for decompositions. @ngimel mentioned that such tools for controlling decompositions were planned (or already existing?), so I think we are all aligned here.

silvasean:
> 2. Data-dependent output shapes/shape constraints: I'd like to know more about how the MLIR ecosystem plans on handling data-dependent output shapes. I know a bit about XLA's plan, but what about the rest of the ecosystem?

Do you mean things like bounded dynamic shapes? Unless PyTorch supports that as a first-class concept on its symbolic sizes, we are unlikely to support it in Torch-MLIR (since we try to align closely with PyTorch's "native" infra). (MLIR's builtin tensor type doesn't natively support bounded dynamic shapes, so most backends do not even have a way of us giving them bounded dynamic shape info)

In general, an operation with data-dependent output shapes is no different at this level than one with regular dynamic shapes. It creates a tensor with a `?` size, and we are unable to refine it further based on static type information (which is fine -- often input tensors have `?` sizes and so much of the graph remains with `?`'s for dimension sizes even after propagating shapes). When it matters is backend-ward from Torch-MLIR, for example in a system like IREE.

@benvanik -- can you comment on ways to handle data-dependent shapes at the IREE level so that the PyTorch devs can understand how we think about it? Is there any info that we can give you from the frontend to make things easier?

docs/long_term_roadmap.md
byronyi: Original program semantics lost in decomposition should be considered harmful. 

Early milestone of a compiler project is usually focused on operator coverage, so decomposing is good in terms of supporting more models: hey, at least I only need to cover the primitives, and I let framework decomposes the rest. However, one does not simply make a compiler (and hardware) just to support more models. Users want better performance, otherwise they could just use plain CPU (or CUDA w/o fancy stuff like fp16 tensor core or transformer engine). With more hard-wired accelerators built into these chips, or even with handcrafted custom op kernels, I often find myself stuck in a weird situation: a large number of local patterns need to be matched to recover those computation semantics (some w/ decent perf improvement opportunities) decomposed in upper level of the stack.

Alternatively, I find myself lazy (or incompetent) in writing all these pattern matches, and I simply save the effort by ditch a particular set of decompositions. With a specific model and backend, I could figure out the optimal decomposition configuration. But with more models, more hardware, and an evolving compiler stack? The complexity only multiplies.

No one likes such a hardware lottery, but facing the end of Moore's law, there is no other way out.

docs/long_term_roadmap.md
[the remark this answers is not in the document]
Chillee: @silvasean In the medium term, we expect that "explicit capture" will also sit behind Dynamo, just a mode where Dynamo will error (or bake in assumptions) instead of producing multiple graphs.

docs/long_term_roadmap.md
```diff
@@ -0,0 +1,242 @@
...
+tools, and it is also very error-prone. We are awaiting movement on this front
+from the PyTorch core team. There is some inspiration from systems like
+[IREE-JAX](https://github.com/iree-org/iree-jax) in the JAX space for how to do
+this, but ultimately this will depend on what the PyTorch core team decides on
+for edge deployments. It is our responsibility to stay connected with them and
+make sure that what they are building suits our needs.
+
+### Project Governance / Structure
+
+Torch-MLIR is current an
```
sstamenova, line 183: Nit: currently

ZihengJiang:
Thanks for the great writeup. I agree that we should leave many features like decompositions at the PyTorch side, which can be more friendly and flexible. And I am open for StableHLO in terms of MHLO lowering.

One thing that I still don't feel very clear about is the input format of the TorchMLIR. In the diagram it shows that we are going to support multiple frontends like TorchDynamo, FuncTorch. But will we convert them both into torch script firstly or we are considering more in-depth support for them?

silvasean:
> Thanks for the great writeup. I agree that we should leave many features like decompositions at the PyTorch side, which can be more friendly and flexible. And I am open for StableHLO in terms of MHLO lowering.
> 
> One thing that I still don't feel very clear about is the input format of the TorchMLIR. In the diagram it shows that we are going to support multiple frontends like TorchDynamo, FuncTorch. But will we convert them both into torch script firstly or are we considering more in-depth support for them?

Good point, I added a note about that:

> Note: Because both TorchDynamo and FuncTorch are TorchFX-based, we could write a direct TorchFX -> MLIR importer, and delete the TorchScript importer. This would remove the need for Torch-MLIR to build its own custom Python extension -- Torch-MLIR would be a pure-Python user of the standard MLIR Python bindings. There is no immediate rush for this though, since TorchFX can be converted to TorchScript (this may become lossy as the dynamic shape support in PyTorch gets more advanced).

ZihengJiang:
Another thing is training. According to my previous experience of applying torch-mlir in training scenario, it is much less mature compared with inference. Do we have plan to raise the priority of training in long term?

docs/long_term_roadmap.md
```diff
@@ -0,0 +1,250 @@
...
+Smaller blockers are related to general API stability and usability of the
+various pieces of PyTorch infra.
+
+These blockers are expected to be addressed by the PyTorch core devs over time.
+Torch-MLIR's role here is to communicate our requirements to PyTorch core and
+align their roadmap and ours. We do this by maintaining connections with the
+PyTorch core developers and being "good-citizen power users" of their latest
+technology (i.e. trying things out, surfacing bugs, providing feedback, etc.).
+
+Note: Because both TorchDynamo and FuncTorch are TorchFX-based, we could write a
```
qiuxiafei, line 100: TorchFX lacks support for dynamic control flow, but TorchScript does. Deleting the TorchScript importer may prevent backend compilers from seeing control flow, if they could do any optimization.

byronyi: I don't think he means to drop the top-level TorchScript importer; he probably means the lowering path torch.fx -> TorchScript -> MLIR. We should deprecate it in favor for direct lowering from torch.fx -> MLIR.

silvasean:
> Another thing is training. According to my previous experience of applying torch-mlir in training scenario, it is much less mature compared with inference. Do we have plan to raise the priority of training in long term?

Training is only difficult today because it is currently awkward to get backwards graphs for real models. TorchDynamo (+ AOTAutograd) and FuncTorch make that easy. So in the future stack, training will just be "more ops support", and we have already proved Torch-MLIR's scalability in that dimension. So I think that better training support will come naturally without any "steering" as users try it and file bugs on missing features.

qiuxiafei:
> 1. We're starting to add control-flow operators in PyTorch

@Chillee , do you mean fx by 'PyTorch' ?

> ... what about the rest of the ecosystem

In BladeDISC, to resolve data-dependent dynamic shape issue, an upper bound of output tensor is estimated and a extra output indicating actual result shape is added if there isn't one. You may take [tf.Where](https://github.com/alibaba/BladeDISC/blob/main/tao_compiler/mlir/disc/IR/hlo_disc_ops.td#L303) and [tf.SparseFillEmptyRows](https://github.com/alibaba/BladeDISC/blob/main/tao_compiler/mlir/disc/IR/hlo_disc_ops.cc#L517-L519) as examples.

PR #1504, opened 2022-10-18
Revert PyTorch update since new PyTorch is not available for macOS

ashay:
Upstream PyTorch nightly page
https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html]
somehow dropped the link for torch-1.14.0.dev20221018 for macOS but not
for Linux or Windows, whereas our RollPyTorch action assumes that if the
nightly version is available for Linux, it is also available for macOS.
This reverts the commit that changed the PyTorch version.


powderluv (approved):

powderluv:
Here is the PyTorch HUD https://hud.pytorch.org/hud/pytorch/pytorch/nightly

Usually they are fast to revert the breakages.

Issue #1446, opened 2022-09-30
Supporting torch's cuda version.


pashu123:
For some of the torch model we have fp16 native support in torch's cuda that doesn't work on cpu.
For example(in stable diffusion): 

```
import onnx
import torch
from diffusers import UNet2DConditionModel

YOUR_TOKEN = "SECRET"


class UnetModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.unet = UNet2DConditionModel.from_pretrained(
            "CompVis/stable-diffusion-v1-4",
            torch_dtype=torch.float16,
            revision="fp16",
            subfolder="unet",
            use_auth_token=YOUR_TOKEN,
        )

    def forward(self, x, y, z):
        return self.unet.forward(x, y, z, return_dict=False)[0]


unet = UnetModel()
unet.cuda()

with torch.inference_mode(), torch.autocast("cuda"):
    inputs = (
        torch.randn(2, 4, 64, 64, dtype=torch.half, device="cuda"),
        torch.randn(1, dtype=torch.half, device="cuda"),
        torch.randn(2, 77, 768, dtype=torch.half, device="cuda"),
    )
    traced_fn = torch.jit.trace(unet, inputs)
```

ramiro050:
There was recently a [PR](https://github.com/llvm/torch-mlir/commit/940959589b186369925a9d53a8d0f6d39fdafac6) that added support for `u/i8` types. I think adding support for `f16` should follow a similar set of changes.

pashu123:
@ramiro I could compile torch-mlir with torch's Cuda version with no additional changes. Also, I obtained the stable diffusion's f16 IR https://gist.github.com/pashu123/84a3a2130e6db9f97c670caf42256208 (elided version) This is failing in the refine types pass which I am taking a look at.

silvasean:
This was fixed, right? @powderluv

pashu123:
Yes, @silvasean. We can close the isse.

powderluv:
We have some build infra work to remove the restriction during install

PR #1505, opened 2022-10-19
Relax the need for only CPU versions of PyTorch

powderluv:
This allows installing corresponding PyTorch CUDA / ROCM versions and using torch-mlir.


powderluv:
follow on to https://github.com/llvm/torch-mlir/issues/1446

PR #1508, opened 2022-10-19
[MLIR][MHLO] emit avg_pool1d and adaptive_avg_pool1d op and add their conversion pattern to MHLO

Vremold:
This PR mainly includes:
1. emit avg_pool1d and adaptive_avg_pool1d
2. decompose adaptive_avg_pool1d op to avg_pool1d op
3. add conversion pattern for avg_pool1d from torch dialect to mhlo dialect.
Besides, this PR also fix a little bug in shape function signature generation. That is if the default arg is a list containing exactly one element, e.g., '[1]', the original strategy will convert it to something like '(1)', which will not be considered as a tuple in Python.


ZihengJiang (approved):
Looks great to me :)

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -734,12 +734,58 @@ def avg_pool2d(input: List[int], kernel_size: List[int], stride: List[int], padd
   else:
     return [nbatch, nInputPlane, outputHeight, outputWidth]
 
+def avg_pool1d(input: List[int], kernel_size: List[int], stride: List[int], padding: List[int], ceil_mode: bool, count_include_pad: bool):
```
silvasean, line 737: Please add new shape functions / fix existing shape functions upstream: https://github.com/llvm/torch-mlir/blob/main/docs/adding_a_shape_function.md#step-by-step-guide

silvasean (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
Vremold: > Please add new shape functions / fix existing shape functions upstream: https://github.com/llvm/torch-mlir/blob/main/docs/adding_a_shape_function.md#step-by-step-guide

The new commit add missing unit mlir test cases for decomposing aten.adaptive_avg_pool2d op. Later, I will propose another PR to pytorch upstream to add all resting shape functions of avg_pool realated ops in torch-mlir. And you will be the coauthor. Will that be okay?

silvasean: I think it makes more sense to move the shape functions to PyTorch first, because otherwise you will need to make another PR later that changes Torch-MLIR to call the upstream shape functions.

Issue #1491, opened 2022-10-13
Throw better error message when not returning a tensor or tuple of tensors


dellis23:
See https://github.com/llvm/torch-mlir/issues/1471 for background

dellis23:
@silvasean 

Re: the case we talked about today, of returning a single-item tuple.  The following error is thrown:

```python
Traceback (most recent call last):
  File "/usr/local/google/home/danielellis/torch-mlir/repro.py", line 40, in <module>
    main()
  File "/usr/local/google/home/danielellis/torch-mlir/repro.py", line 32, in main
    linalg_on_tensors_mlir = torch_mlir.compile(
  File "/usr/local/google/home/danielellis/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/__init__.py", line 274, in compile
    run_pipeline_with_repro_report(
  File "/usr/local/google/home/danielellis/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/compiler_utils.py", line 73, in run_pipeline_with_repro_report
    raise TorchMlirCompilerError(trimmed_message) from None
torch_mlir.compiler_utils.TorchMlirCompilerError: Lowering TorchScript IR -> Torch Backend IR failed with the following diagnostics:
error: failed to legalize unresolved materialization from '!torch.tuple<tensor>' to '!torch.tensor' that remained live after conversion
note: see current operation: %2 = "builtin.unrealized_conversion_cast"(%1) : (!torch.tuple<tensor>) -> !torch.tensor
note: see existing live user here: "func.return"(%2) : (!torch.tensor) -> ()
```

We talked about putting another check just before the success [here](https://github.com/llvm/torch-mlir/blob/0af55781aede053f32d2b618f0f6d36d875fd908/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp#L261).  We never actually reach this line, however.  We end up failing in [the check before](https://github.com/llvm/torch-mlir/blob/0af55781aede053f32d2b618f0f6d36d875fd908/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp#L258-L260), in the call to `applyPartialConversion`.

ramiro050:
I think the issue is this type conversion:

https://github.com/llvm/torch-mlir/blob/0af55781aede053f32d2b618f0f6d36d875fd908/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp#L196-L201

[...]

PR #1509, opened 2022-10-19
Add better error message for single-tensor tuple returns.

dellis23:
See https://github.com/llvm/torch-mlir/issues/1491

### DIFF
--- a/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
+++ b/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
@@ -189,13 +189,10 @@ public:
 } // namespace
 
 static bool isValidNonContainerResultType(Type resultType) {
-  if (resultType.isa<Torch::BaseTensorType>() ||
-      resultType.isa<Torch::FloatType>() ||
-      resultType.isa<Torch::IntType>() ||
-      resultType.isa<Torch::BoolType>()) {
-    return true;
-  }
-  return false;
+  return resultType.isa<Torch::BaseTensorType>() ||
+         resultType.isa<Torch::FloatType>() ||
+         resultType.isa<Torch::IntType>() ||
+         resultType.isa<Torch::BoolType>();
 }
 
 static LogicalResult validateReturns(func::FuncOp func) {
--- a/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
+++ b/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
@@ -298,9 +298,8 @@ static LogicalResult adjustCallingConventions(func::FuncOp func,
   // We don't know how to rewrite it, so mark it as illegal.
   target.addIllegalOp<func::CallIndirectOp>();
   if (failed(applyPartialConversion(func.getOperation(), target,
-                                    std::move(patterns)))) {
+                                    std::move(patterns))))
     return failure();
-  }
   return success();
 }
 
--- a/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
+++ b/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
@@ -196,8 +196,9 @@ static bool isValidNonContainerResultType(Type resultType) {
 }
 
 static LogicalResult validateReturns(func::FuncOp func) {
-  // getResultTypes can only ever be one item.  Multiple return values are
-  // stored in a tuple.
+  assert(func.getResultTypes().size() <= 1 &&
+         "getResultTypes can only ever be one item. Multiple return values are "
+         "stored in a tuple.");
   const auto& resultType = func.getResultTypes().front();
 
   // Allow single tensor, scalar, and bool returns
--- a/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
+++ b/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
@@ -216,8 +216,9 @@ static LogicalResult validateReturns(func::FuncOp func) {
   }
 
   return func->emitError(
-    "Functions must return a tensor, scalar, bool, multiple tensors, scalars, "
-    "or bools, or a tuple of multiple tensors, scalars, or bools.");
+    "Functions must return a single value of a valid type, multiple valid "
+    "types, or a tuple of more than one valid type. Valid types: tensor, "
+    "scalar, and bool.");
 }
 
 static LogicalResult adjustCallingConventions(func::FuncOp func,
--- a/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
+++ b/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
@@ -203,12 +203,12 @@ static LogicalResult validateReturns(func::FuncOp func) {
   // stored in a tuple.
   const auto& resultType = func.getResultTypes().front();
 
-  // Allow single tensor returns
+  // Allow single tensor, scalar, and bool returns
   if (isValidNonContainerResultType(resultType)) {
     return success();
   }
 
-  // Allow multi-tensor tuple returns
+  // Allow multi-tensor/scalar/bool tuple returns
   if (auto tuple = resultType.dyn_cast<Torch::TupleType>()) {
     bool containsOnlyTensors = true;
     const auto& containedTypes = tuple.getContainedTypes();
@@ -223,8 +223,8 @@ static LogicalResult validateReturns(func::FuncOp func) {
   }
 
   return func->emitError(
-    "Functions must return a tensor, multiple tensors, "
-    "or a tuple of multiple tensors.");
+    "Functions must return a tensor, scalar, bool, multiple tensors, scalars, "
+    "or bools, or a tuple of multiple tensors, scalars, or bools.");
 }
 
 static LogicalResult adjustCallingConventions(func::FuncOp func,
--- a/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
+++ b/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
@@ -19,6 +19,7 @@
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Transforms/Passes.h"
 
+
 using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::Torch;
@@ -187,8 +188,49 @@ public:
 };
 } // namespace
 
+static bool isValidNonContainerResultType(Type result_type) {
+  if (result_type.isa<Torch::BaseTensorType>() ||
+      result_type.isa<Torch::FloatType>() ||
+      result_type.isa<Torch::IntType>() ||
+      result_type.isa<Torch::BoolType>()) {
+    return true;
+  }
+  return false;
+}
+
+static LogicalResult validateReturns(func::FuncOp func) {
+  // getResultTypes can only ever be one item.  Multiple return values are
+  // stored in a tuple.
+  const auto& result_type = func.getResultTypes().front();
+
+  // Allow single tensor returns
+  if (isValidNonContainerResultType(result_type)) {
+    return success();
+  }
+
+  // Allow multi-tensor tuple returns
+  if (auto tuple = result_type.dyn_cast<Torch::TupleType>()) {
+    bool contains_only_tensors = true;
+    const auto& contained_types = tuple.getContainedTypes();
+    for (const auto& contained_type : contained_types) {
+      if (!isValidNonContainerResultType(contained_type)) {
+        contains_only_tensors = false;
+      }
+    }
+    if (contained_types.size() >= 2 && contains_only_tensors) {
+      return success();
+    }
+  }
+
+  return func->emitError(
+    "Functions must return a tensor, multiple tensors, "
+    "or a tuple of multiple tensors.");
+}
+
 static LogicalResult adjustCallingConventions(func::FuncOp func,
                                               TypeBoundMap &typeBoundMap) {
+  if (failed(validateReturns(func)))
+      return failure();
   MLIRContext *context = func.getContext();
   RewritePatternSet patterns(context);
   TypeConverter typeConverter;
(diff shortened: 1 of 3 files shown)


dellis23:
Note: I'm not sure why `UpSampleNearest2dDynamicFactor_basic` started passing.  I'm assuming a synchronization issue that will go away if I re-sync to main?

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -187,8 +188,49 @@ class AdjustCallingConventionForReturn
...
+      }
+    }
+    if (contained_types.size() >= 2 && contains_only_tensors) {
+      return success();
+    }
+  }
+
+  return func->emitError(
+    "Functions must return a tensor, multiple tensors, "
+    "or a tuple of multiple tensors.");
```
ramiro050, line 227: This error message is not completely accurate. Returning values other than tensors is supported. Also, I think we can improve the specificity of the error messages by returning early once an invalid return type is encountered. In other words, rewriting things as:

```cpp
static LogicalResult validateReturns(func::FuncOp func) {
  // ...
  if (auto tuple = resultType.dyn_cast<Torch::TupleType>()) {
    // ...
    if (containedTypes.size() < 2)
      return func->emitError("Error about tuple size");
    // ...
    if (!containsValidTypes)
      return func->emitError("Error about tuple not containing valid types");
  }

  if (!isValidNonContainerResultType(resultType)) {
    return func->emitError("Error about resultType not being a valid type or tuple of valid types");
  }

  return success();
}
```

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
dellis23: @silvasean's preference was an allowlist rather than a denylist, so that's what I went with here.  Will see what he thinks.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -187,8 +188,49 @@ class AdjustCallingConventionForReturn
 };
 } // namespace
 
+static bool isValidNonContainerResultType(Type result_type) {
+  if (result_type.isa<Torch::BaseTensorType>() ||
```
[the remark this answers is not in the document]
dellis23: Done.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -187,8 +188,49 @@ class AdjustCallingConventionForReturn
...
+      result_type.isa<Torch::IntType>() ||
+      result_type.isa<Torch::BoolType>()) {
+    return true;
+  }
+  return false;
+}
+
+static LogicalResult validateReturns(func::FuncOp func) {
+  // getResultTypes can only ever be one item.  Multiple return values are
+  // stored in a tuple.
```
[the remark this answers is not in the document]
silvasean: This is true in our pipeline, but you could still hit the assert with verifier-valid IR. `assert` should only be used for scenarios where verifier-valid IR could never hit it.

I do think we should emit an error diagnostic in this case though, since it should hold for our pass pipelines.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
silvasean: Yes, I think an allowlist is better here, since we really should know exactly everything. The errors should be great nonetheless.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -187,8 +188,49 @@ class AdjustCallingConventionForReturn
...
+
+  // Allow multi-tensor tuple returns
+  if (auto tuple = result_type.dyn_cast<Torch::TupleType>()) {
+    bool contains_only_tensors = true;
+    const auto& contained_types = tuple.getContainedTypes();
+    for (const auto& contained_type : contained_types) {
+      if (!isValidNonContainerResultType(contained_type)) {
+        contains_only_tensors = false;
+      }
+    }
```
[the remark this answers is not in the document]
dellis23: Thanks.  I had a feeling there was a nice mapping function somewhere to simplify this.  Done.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -187,8 +188,49 @@ class AdjustCallingConventionForReturn
...
+  const auto& result_type = func.getResultTypes().front();
+
+  // Allow single tensor returns
+  if (isValidNonContainerResultType(result_type)) {
+    return success();
+  }
+
+  // Allow multi-tensor tuple returns
+  if (auto tuple = result_type.dyn_cast<Torch::TupleType>()) {
+    bool contains_only_tensors = true;
```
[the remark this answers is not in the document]
dellis23: Thanks, definitely better.  (All of this evolved from me starting with assuming tensors were the only thing we could return and then fixing test cases.  Forgot to figure out / update the naming.)

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -187,8 +188,49 @@ class AdjustCallingConventionForReturn
...
+  }
+  return false;
+}
+
+static LogicalResult validateReturns(func::FuncOp func) {
+  // getResultTypes can only ever be one item.  Multiple return values are
+  // stored in a tuple.
+  const auto& result_type = func.getResultTypes().front();
+
+  // Allow single tensor returns
```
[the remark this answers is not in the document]
dellis23: I've tried to update it.  It's quite verbose now, but more accurate.  Let me know if you have better ideas.  Happy to nitpick the phrasing to get it right, since it's what the user will see (for the error message).

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
dellis23: Resolving.  Updated error message, though I still don't like it a ton.  If you see a potential improvement please let me know.

ramiro050:
The CI seems to be failing on this lit test:

https://github.com/llvm/torch-mlir/blob/4a776be1560ec3127aae681e36325b547c270314/test/Dialect/Torch/adjust-calling-conventions.mlir#L35

Given the new allowlist, the tests that return `None` will have to be adjusted to check for the error message instead. Here's an example of how to do that:

https://github.com/llvm/torch-mlir/blob/4a776be1560ec3127aae681e36325b547c270314/test/Dialect/Torch/reduce-op-variants-error.mlir#L7

You can run the tests locally by doing `cmake --build build --target check-torch-mlir`

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -187,8 +187,46 @@ class AdjustCallingConventionForReturn
...
+  return resultType.isa<Torch::BaseTensorType>() ||
+         resultType.isa<Torch::FloatType>() ||
+         resultType.isa<Torch::IntType>() ||
+         resultType.isa<Torch::BoolType>();
+}
+
+static LogicalResult validateReturns(func::FuncOp func) {
+  if (func.getResultTypes().size() > 1) {
+    return func->emitError(
+      "Functions can only ever can only ever return one item. Multiple return "
```
ramiro050, line 200: "~can only ever~ can only ever"

dellis23: 🤦 Done.

ramiro050:
> > You can run the tests locally by doing `cmake --build build --target check-torch-mlir`
> 
> Ah thanks, I was running the e2e tests but didn't know about this one.

Oh also, the convention seems to be to have all error cases in their own `{name_of_pass}_error.mlir` file, so you'll have to create a new file. Make sure to include this line at the top

```
// RUN: torch-mlir-opt -torch-adjust-calling-conventions -allow-unregistered-dialect -split-input-file -verify-diagnostics %s
```

dellis23:
> > > You can run the tests locally by doing `cmake --build build --target check-torch-mlir`
> > 
> > 
> > Ah thanks, I was running the e2e tests but didn't know about this one.
> 
> Oh also, the convention seems to be to have all error cases in their own `{name_of_pass}_error.mlir` file, so you'll have to create a new file. Make sure to include this line at the top
> 
> ```
> // RUN: torch-mlir-opt -torch-adjust-calling-conventions -allow-unregistered-dialect -split-input-file -verify-diagnostics %s
> ```

I'm not sure what's going on here.  See the error [here](https://github.com/llvm/torch-mlir/actions/runs/3291007757/jobs/5424609519#step:6:6194).  Shortened snippet:

```shell
FileCheck error: '<stdin>' is empty.
FileCheck command line:  FileCheck /main_checkout/torch-mlir/test/Dialect/Torch/adjust-calling-conventions.mlir
```

I do already see a `test/Dialect/Torch/adjust-calling-conventions.mlir` file.  I tried also making a file in the same directory named `adjust-calling-conventions-error.mlir` and `adjust-calling-conventions_error.mlir` with the contents you mentioned, but neither fix the problem.

I'm guessing you're saying I need to create this file because we have a new failure?  I'm having trouble seeing how the error message implies that though.  The best I can read from it is that it can't even find the file that already does exist?

ramiro050:
So there's two things separate things happening here. The first is the crash in the CI. It's failing on an assert: https://github.com/llvm/torch-mlir/actions/runs/3291007757/jobs/5424609519#step:6:6200

This is due to the fact that functions can return nothing (this is a valid case). So this check

https://github.com/llvm/torch-mlir/blob/7a8f1d14d5945e917672aa227c336e663e03a411/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp#L199

should handle the case when there are no return types before calling `front` on them.

The second thing is related to my comments about adding an `adjust-calling-conventions-error.mlir` file. I don't think we want the return of `None` types to be valid. The current backend contract does not expect backends to handle dynamic computations on `None`s

https://github.com/llvm/torch-mlir/blob/7a8f1d14d5945e917672aa227c336e663e03a411/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L37-L43

So the new file `adjust-calling-conventions-error.mlir` should have the tests for returning `None`s with checks for the error message.

dellis23:
> So there's two separate things happening here. The first is the crash in the CI. It's failing on an assert: https://github.com/llvm/torch-mlir/actions/runs/3291007757/jobs/5424609519#step:6:6200
> 
> This is due to the fact that functions can return nothing (this is a valid case). So this check
> 
> https://github.com/llvm/torch-mlir/blob/7a8f1d14d5945e917672aa227c336e663e03a411/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp#L199
> 
> should handle the case when there are no return types before calling `front` on them.
> 
> The second thing is related to my comments about adding an `adjust-calling-conventions-error.mlir` file. I don't think we want the return of `None` types to be valid. The current backend contract does not expect backends to handle dynamic computations on `None`s
> 
> https://github.com/llvm/torch-mlir/blob/7a8f1d14d5945e917672aa227c336e663e03a411/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L37-L43
> 
> So the new file `adjust-calling-conventions-error.mlir` should have the tests for returning `None`s with checks for the error message.

Talked with Ramiro about this over chat.  He suggested getting @silvasean's feedback to verify.

> So there's two separate things happening here. The first is the crash in the CI. It's failing on an assert: https://github.com/llvm/torch-mlir/actions/runs/3291007757/jobs/5424609519#step:6:6200
> 
> This is due to the fact that functions can return nothing (this is a valid case). So this check
> 
> https://github.com/llvm/torch-mlir/blob/7a8f1d14d5945e917672aa227c336e663e03a411/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp#L199
> 
> should handle the case when there are no return types before calling `front` on them.
> 
> The second thing is related to my comments about adding an `adjust-calling-conventions-error.mlir` file. I don't think we want the return of `None` types to be valid. The current backend contract does not expect backends to handle dynamic computations on `None`s
> 
> https://github.com/llvm/torch-mlir/blob/7a8f1d14d5945e917672aa227c336e663e03a411/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L37-L43
> 
> So the new file `adjust-calling-conventions-error.mlir` should have the tests for returning `None`s with checks for the error message.

Okay, we spoke about this over video.  I think the current understanding is that we _do_ need to support `None` and empty returns.  There's a question about what to do for tuples that contain `None`s though, but this was messy in our testing so I'm going to create another issue after this is done to address that separately.  @ramiro050 feel free to add more if I missed anything.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -187,8 +187,47 @@ class AdjustCallingConventionForReturn
...
+         resultType.isa<Torch::FloatType>() ||
+         resultType.isa<Torch::IntType>() ||
+         resultType.isa<Torch::BoolType>() ||
+         resultType.isa<Torch::NoneType>();
+}
+
+static LogicalResult validateReturns(func::FuncOp func) {
+  if (func.getResultTypes().size() > 1) {
+    return func->emitError(
+      "Functions can only ever return one item. Multiple return values are "
```
silvasean, line 201: This is not really precise. I would phrase this as "Functions directly imported from Python should..."

dellis23: Updated.  LMK what you think of the new phrasing.  I also added a comment above the part where we return if nothing has been returned (as distinct from `None`).  Let me know if there's a better way to phrase that comment.

ramiro050 (approved):
LGTM!

silvasean:
Hey Dan, this patch should be tested in test/Dialect/Torch/adjust-calling-conventions.mlir (you can use test/Dialect/Torch/refine-public-return.mlir as an example -- the main difference is the `-split-input-file -verify-diagnostics` and use of `// -----` along with the `expected-error` directives).

When you add those tests, can you also remove the end-to-end testing for this patch?

This patch falls into the "With the exceptions above, all changes should include appropriate unit tests, as is standard in the LLVM and MLIR community. This includes full coverage of all canonicalizations, pretty printing, passes, errors, and diagnostics." part of that.

PR #1513, opened 2022-10-20
[MLIR][TORCH] Add E2E support for `aten.mse_loss` op

vivekkhandelwal1:
This commit adds decomposition for the `aten.mse_loss` op.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2845,6 +2845,63 @@ class DecomposeAtenIndexTensorHackedTwinOp
...
+      return rewriter.notifyMatchFailure(
+          op, "Expected a constant integer value for reduction");
+
+    Location loc = op.getLoc();
+    BaseTensorType inputType = op.self().getType().cast<BaseTensorType>();
+    BaseTensorType resultType = op.getType().cast<BaseTensorType>();
+    BaseTensorType subType =
+        inputType
+            .getWithSizesAndDtype(
+                !inputType.hasSizes()
```
ramiro050, line 2870: You should return a `notifyMatchFailure` if the input type does not have sizes, since the input type having sizes is needed to create the right decomposition when the reduction type is `Mean`. The `Mean` case assumes that the input tensor has a rank.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2845,6 +2845,63 @@ class DecomposeAtenIndexTensorHackedTwinOp
...
+
+    if (reductionType >= 1) {
+      int64_t inputRank = getTensorRank(op.self());
+      SmallVector<Value> dimListElements;
+      for (int64_t i = 0; i < inputRank; i++)
+        dimListElements.push_back(rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(i)));
+      Value dimList = rewriter.create<PrimListConstructOp>(
+          loc, Torch::ListType::get(Torch::IntType::get(op->getContext())),
+          dimListElements);
```
ramiro050, line 2887: I don't think there is a need for this. Both `AtenMeanDimOp` and `AtenSumDimIntListOp` should reduce along all dimensions when `dim=None`

python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -601,3 +601,61 @@ def forward(self, a):
...
+        ([-1 , -1], torch.float32, True),
+        ([-1 , -1], torch.float32, True),
+    ])
+
+    def forward(self, x, y):
+        return torch.ops.aten.mse_loss(x, y, reduction=0)
+
+@register_test_case(module_factory=lambda: MseLossNoReductionModule())
+def MseLossNoReductionModule_basic(module, tu: TestUtils):
+    module.forward(torch.randn(2, 4), torch.randn(2, 4))
```
ramiro050, line 623: This applies to all three test cases: use `tu.rand` to generate tensors. See: https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#test-case-function

> all random inputs should be generated through the TestUtils object.

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -991,3 +991,19 @@ func.func @torch.aten.roll(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.int,
...
+
+// CHECK-LABEL:   func.func @torch.aten.mse_loss(
+// CHECK-SAME:                                   %[[SELF:.*]]: !torch.vtensor<[?,?],f32>,
+// CHECK-SAME:                                   %[[TARGET:.*]]: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[?,?],f32> {
+// CHECK:         %[[REDUCTION:.*]] = torch.constant.int 0
+// CHECK:         %[[ALPHA:.*]] = torch.constant.float 1.000000e+00
+// CHECK:         %[[SUB:.*]] = torch.aten.sub.Tensor %[[SELF]], %[[TARGET]], %[[ALPHA]] : !torch.vtensor<[?,?],f32>, !torch.vtensor<[?,?],f32>, !torch.float -> !torch.vtensor<[?,?],f32>
+// CHECK:         %[[SQUARE_SUB:.*]] = torch.aten.mul.Tensor %[[SUB]], %[[SUB]] : !torch.vtensor<[?,?],f32>, !torch.vtensor<[?,?],f32> -> !torch.vtensor<[?,?],f32>
+// CHECK:         return %[[SQUARE_SUB]] : !torch.vtensor<[?,?],f32>
+func.func @torch.aten.mse_loss(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[?,?],f32> {
```
ramiro050, line 1005: Can you add tests for each reduction type? See:  https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

> DO use a unit test if your lowering for an op has multiple cases / logic. This also helps future maintainers of the lowering to see in one place all the different edge cases of the op that you had to handle. (these can be easily reduced out of all the end-to-end tests you added).

ramiro050 (changes requested):
Thanks for this!

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -991,3 +991,59 @@ func.func @torch.aten.roll(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.int,
...
+
+// CHECK-LABEL:   func.func @torch.aten.mse_loss(
+// CHECK-SAME:                                   %[[SELF:.*]]: !torch.vtensor<[?,?],f32>,
+// CHECK-SAME:                                   %[[TARGET:.*]]: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[?,?],f32> {
+// CHECK:         %[[REDUCTION:.*]] = torch.constant.int 0
+// CHECK:         %[[ALPHA:.*]] = torch.constant.float 1.000000e+00
+// CHECK:         %[[SUB:.*]] = torch.aten.sub.Tensor %[[SELF]], %[[TARGET]], %[[ALPHA]] : !torch.vtensor<[?,?],f32>, !torch.vtensor<[?,?],f32>, !torch.float -> !torch.vtensor<[?,?],f32>
+// CHECK:         %[[SUB_SQUARE:.*]] = torch.aten.mul.Tensor %[[SUB]], %[[SUB]] : !torch.vtensor<[?,?],f32>, !torch.vtensor<[?,?],f32> -> !torch.vtensor<[?,?],f32>
+// CHECK:         return %[[SUB_SQUARE]] : !torch.vtensor<[?,?],f32>
+func.func @torch.aten.mse_loss(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[?,?],f32> {
```
ramiro050, line 1005: nit: can you give each function a unique postfix to know what each is doing? Such as `@torch.aten.mse_loss$mean`

ramiro050 (approved):
Two small comments, but other than that LGTM

PR #1514, opened 2022-10-20
[MLIR][TORCH] Add -convert-torch-to-tosa-custom pass for undecomposed selective ops

AmosLewis:
Use tosa::custom op to represent softmax op which we deliberately choose not to decompose into simple ops.The reason to do this is in this RFC here: https://github.com/llvm/torch-mlir/issues/1519

### DIFF
--- a/include/torch-mlir/Conversion/Passes.td
+++ b/include/torch-mlir/Conversion/Passes.td
@@ -114,6 +114,41 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "func::FuncOp"> {
   let constructor = "mlir::torch::createConvertTorchToTosaPass()";
 }
 
+def ConvertTorchToTosaCustom : Pass<"convert-torch-to-tosa-custom", "func::FuncOp"> {
+        let summary = "Convert Torch ops to TOSA custom ops";
+  let description = [{
+    The purpose to use tosa::custom is handle complex ops when we do not
+    want to decompose them into simple ops.
+    The aten op name will be used to construct a StringAttr as the identifier attribute for tosa::CustomOp.
+    So in the output the users know where the tosa::CustomOp is coming from by the StringAttr id.
+    Each input arg from Aten Dialect has to be converted to a tensor of number values as the
+    operand of tosa::CustomOp op.
+    The contract to follow:
+        AnyTorchTensorType 	->	TensorType<AnySizexAnyType>  of tosa::ConstOp
+        Torch_IntType		->      RankedTensorType<1xi64>  of tosa::ConstOp
+        Torch_FloatType		->      RankedTensorType<1xf32>  of tosa::ConstOp
+        ...
+        TODO:
+          1. Support more types: Torch_BoolType, Torch_QInt8Type, Torch_QUInt8Type
+              Torch_NoneType, Torch_StringType, Torch_OptionalType, Torch_ListType
+              Torch_NumberType, Torch_DictType, Torch_DeviceType, Torch_GeneratorType,
+              Torch_LinearParamsType, Torch_TupleType, Torch_UnionType,
+              Torch_AnyType, AnyTorchOptionalType,
+          2. Support multi result types
+    The examples are in the test/Conversion/TorchToTosa/custom.mlir
+  }];
+  let options = [
+    ListOption<"customOps", "custom-ops", "std::string",
+               "List of operation names that should be converted to tosa::custom",
+               "llvm::cl::ZeroOrMore">
+  ];
+
+  let constructor = [{
+      mlir::torch::createConvertTorchToTosaCustomPass(
+          /*customOps=*/{})
+  }];
+}
+
 def ConvertTorchToTMTensor : Pass<"convert-torch-to-tmtensor", "func::FuncOp"> {
   let summary = "Convert recognized Torch ops to TMTensor/Linalg ops";
   let description = [{
--- a/include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
+++ b/include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
@@ -28,7 +28,15 @@ void createTorchBackendToLinalgOnTensorsBackendPipeline(OpPassManager &pm);
 
 /// Creates a pipeline that lowers from the torch backend contract to the
 /// TOSA backend contract.
-void createTorchBackendToTosaBackendPipeline(OpPassManager &pm);
+struct TosaBackendPipelineOptions
+    : public PassPipelineOptions<TosaBackendPipelineOptions> {
+  ListOption<std::string> customOps{
+      *this, "custom-ops",
+      llvm::cl::desc("List of ops to be converted to tosa::CustomOp."),
+      llvm::cl::ZeroOrMore};
+};
+void createTorchBackendToTosaBackendPipeline(
+    OpPassManager &pm, const TosaBackendPipelineOptions &options);
 
 // Do not register the torch-to-mhlo pipeline if mhlo target is disabled
 #ifdef TORCH_MLIR_ENABLE_MHLO
--- a/lib/Conversion/TorchToTosa/TorchToTosaCustom.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosaCustom.cpp
@@ -1,0 +1,150 @@
+//===----------------------------------------------------------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// Also available under a BSD-style license. See LICENSE.
+//
+//===----------------------------------------------------------------------===//
+
+#include "torch-mlir/Conversion/TorchToTosa/TorchToTosa.h"
+#include "torch-mlir/Conversion/TorchToTosa/TosaLegalizeCommon.h"
+#include "torch-mlir/Conversion/TorchToTosa/TosaLegalizeUtils.h"
+
+#include "../PassDetail.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
+#include "mlir/Dialect/Tosa/IR/TosaOps.h"
+#include "mlir/Dialect/Traits.h"
+#include "mlir/IR/Matchers.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+#include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionDialect.h"
+#include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
+#include <unordered_map>
+#include <iostream>
+
+using namespace mlir;
+using namespace mlir::torch;
+using namespace mlir::torch::Torch;
+
+namespace {
+
+class ConvertSelectiveOpToTosaCustom : public ConversionPattern {
+public:
+  ArrayRef<std::string>  customOps;
+  ConvertSelectiveOpToTosaCustom(TypeConverter &typeConverter, MLIRContext *context,
+                                 ArrayRef<std::string> customOps)
+      : ConversionPattern(typeConverter, MatchAnyOpTypeTag(), /*benefit=*/1,
+                          context){
+    this->customOps = customOps;
+  }
+
+  LogicalResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const override {
+    std::string op_name = op->getName().getStringRef().str();
+    for (auto customOpName : customOps) {
+      if (customOpName == op_name) {
+        int num_operands = operands.size();
+        std::vector<mlir::Value> inputs_vec;
+        for (int i = 0; i < num_operands; i++) {
+          auto operand = op->getOperands()[i];
+          auto operand_type = operand.getType();
+          // type convert for operands
+          if (operand_type.template isa<torch::Torch::IntType>()) {
+            int64_t operand_tosa;
+            if (!matchPattern(operand, m_TorchConstantInt(&operand_tosa)))
+              return rewriter.notifyMatchFailure(
+                  op, "unimplemented: operand should be a torch.constant.int");
+            auto operand_tensor_int =
+                tosa::getConstTensor<int64_t>(rewriter, op, operand_tosa, {1});
+            inputs_vec.push_back(operand_tensor_int.value());
+          } else if (operand_type.template isa<torch::Torch::FloatType>()) {
+            double operand_tosa;
+            if (!matchPattern(operand, m_TorchConstantFloat(&operand_tosa)))
+              return rewriter.notifyMatchFailure(
+                  op,
+                  "unimplemented: operand should be a torch.constant.float");
+            auto operand_tensor_float =
+                tosa::getConstTensor<float>(rewriter, op, operand_tosa, {1});
+            inputs_vec.push_back(operand_tensor_float.value());
+          } else if (operand_type
+                         .template isa<torch::Torch::ValueTensorType>()) {
+            inputs_vec.push_back(operands[i]);
+          } else {
+            // TODO Handle more types like !torch.list<...>, !torch.device,
+            // !torch.string, !torch.none, !torch.generator.
+            return rewriter.notifyMatchFailure(
+                op, "unimplemented: inputs type. The input has to be int/float/tensor ");
+          }
+        }
+        // Create operands for tosa::CustomOp
+        llvm::ArrayRef<mlir::Value> ref(inputs_vec.data(), inputs_vec.size());
+        ValueRange custom_inputs(ref);
+        // Create output type for tosa::CustomOp
+        auto outType =
+            getTypeConverter()->convertType(op->getResult(0).getType());
+
+        rewriter.replaceOpWithNewOp<tosa::CustomOp>(
+            op, outType, op->getName().getStringRef(), custom_inputs);
+        return success();
+      }
+    }
+    return failure();
+  }
+};
+
+} // namespace
+
+// -----------------------------------------------------------------------------
+// TorchToTosaCustom Pass
+// -----------------------------------------------------------------------------
+
+namespace {
+class ConvertTorchToTosaCustom
+    : public ConvertTorchToTosaCustomBase<ConvertTorchToTosaCustom> {
+public:
+  ConvertTorchToTosaCustom() = default;
+  ConvertTorchToTosaCustom(ArrayRef<std::string> customOps) {
+    this->customOps = customOps;
+  }
+  void getDependentDialects(DialectRegistry &registry) const override {
+    registry.insert<tosa::TosaDialect>();
+    registry.insert<tensor::TensorDialect>();
+    registry.insert<arith::ArithDialect>();
+    TorchConversion::getBackendTypeConversionDependentDialects(registry);
+  }
+
+  void runOnOperation() override {
+    MLIRContext *context = &getContext();
+    ConversionTarget target(*context);
+    target.addLegalDialect<tosa::TosaDialect, tensor::TensorDialect,
+                           arith::ArithDialect>();
+
+    TypeConverter typeConverter;
+    typeConverter.addConversion([](Type type) { return type; });
+    TorchConversion::setupBackendTypeConversion(target, typeConverter);
+
+    RewritePatternSet patterns(context);
+
+    patterns.add<ConvertSelectiveOpToTosaCustom>(typeConverter,
+                                                 context, customOps);
+
+    for (std::string opName : customOps) {
+      target.addIllegalOp(OperationName(opName, context));
+    }
+
+    if (failed(applyPartialConversion(getOperation(), target,
+                                      std::move(patterns))))
+      return signalPassFailure();
+  }
+};
+} // namespace
+
+std::unique_ptr<OperationPass<func::FuncOp>>
+mlir::torch::createConvertTorchToTosaCustomPass(
+    ArrayRef<std::string> customOps) {
+  return std::make_unique<ConvertTorchToTosaCustom>(customOps);
+}
--- a/lib/Dialect/TorchConversion/Transforms/Passes.cpp
+++ b/lib/Dialect/TorchConversion/Transforms/Passes.cpp
@@ -47,7 +47,7 @@ void mlir::torch::registerTorchConversionPasses() {
       "contract.",
       TorchConversion::createTorchBackendToLinalgOnTensorsBackendPipeline);
 
-  mlir::PassPipelineRegistration<>(
+  mlir::PassPipelineRegistration<TorchConversion::TosaBackendPipelineOptions>(
       "torch-backend-to-tosa-backend-pipeline",
       "Pipeline lowering torch backend contract to TOSA backend "
       "contract.",
@@ -96,7 +96,11 @@ void TorchConversion::createTorchBackendToLinalgOnTensorsBackendPipeline(
 }
 
 void TorchConversion::createTorchBackendToTosaBackendPipeline(
-    OpPassManager &pm) {
+    OpPassManager &pm, const TosaBackendPipelineOptions &options) {
+  if (!options.customOps.empty()) {
+    pm.addNestedPass<func::FuncOp>(
+        createConvertTorchToTosaCustomPass(options.customOps));
+  }
   pm.addNestedPass<func::FuncOp>(createConvertTorchToTosaPass());
   // Perform rank broadcasting so TosaToLinalg pass works
   pm.addNestedPass<func::FuncOp>(createTosaMakeBroadcastablePass());
--- a/include/torch-mlir/Conversion/TorchToTosa/TorchToTosa.h
+++ b/include/torch-mlir/Conversion/TorchToTosa/TorchToTosa.h
@@ -17,6 +17,9 @@
 namespace mlir {
 namespace torch {
 std::unique_ptr<OperationPass<func::FuncOp>> createConvertTorchToTosaPass();
+
+std::unique_ptr<OperationPass<func::FuncOp>>
+createConvertTorchToTosaCustomPass(ArrayRef<std::string> customOps);
 }
 } // namespace mlir
 
--- a/lib/Conversion/TorchToTosa/CMakeLists.txt
+++ b/lib/Conversion/TorchToTosa/CMakeLists.txt
@@ -1,5 +1,6 @@
 add_mlir_conversion_library(TorchMLIRTorchToTosa
   TorchToTosa.cpp
+  TorchToTosaCustom.cpp
   TosaLegalizeUtils.cpp
   TosaLegalizeCommon.cpp
 
--- a/test/Conversion/TorchToTosa/custom.mlir
+++ b/test/Conversion/TorchToTosa/custom.mlir
@@ -1,0 +1,39 @@
+// RUN: torch-mlir-opt <%s -convert-torch-to-tosa-custom="custom-ops=torch.aten.softmax.int,torch.aten.rsub.Scalar" -split-input-file -verify-diagnostics | FileCheck %s
+
+// CHECK-LABEL:   func.func @torch.aten.softmax.int$cst_dim(
+// CHECK-SAME:                                              %[[VAL_0:.*]]: !torch.vtensor<[2,3],f32>) -> !torch.vtensor<[2,3],f32> {
+// CHECK:           %[[VAL_1:.*]] = torch_c.to_builtin_tensor %[[VAL_0]] : !torch.vtensor<[2,3],f32> -> tensor<2x3xf32>
+// CHECK:           %[[VAL_2:.*]] = torch.constant.int 1
+// CHECK:           %[[VAL_3:.*]] = torch.constant.int 1
+// CHECK:           %[[VAL_4:.*]] = "tosa.const"() {value = dense<1> : tensor<1xi64>} : () -> tensor<1xi64>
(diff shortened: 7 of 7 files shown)


ramiro050:
The CI error is due to this patch. The program is crashing when creating the `tosa::CustomOp`:

https://github.com/llvm/torch-mlir/actions/runs/3292098619/jobs/5427059644#step:6:6218

ramiro050:
I don't think this is the right approach for what this patch is trying to achieve. As a user of torch-mlir, I would expect that running the `torch-to-tosa` pass on any set of supported ops will result in standard tosa ops. Users should not be expected by default to handle arbitrary custom ops. To use custom ops should be a choice made by the user.

I think this requires a more general solution where users can specify which torch-ops to pass down as custom ops, and have the conversion to `tosa::CustomOp` happen automatically. Since this will require some non-trivial changes to several parts of torch-mlir, can you open an RFC issue with what you're tryinig to achieve, and we can brainstorm on the best solution.

AmosLewis:
> The CI error is due to this patch. The program is crashing when creating the `tosa::CustomOp`:
> 
> https://github.com/llvm/torch-mlir/actions/runs/3292098619/jobs/5427059644#step:6:6218

The interesting thing is when I run [https://github.com/llvm/torch-mlir/actions/runs/3292098619/jobs/5427059644#step:6:6218:~:text=6191-,%2D%2D,%2D%2D,-6194](https://github.com/llvm/torch-mlir/actions/runs/3292098619/jobs/5427059644#step:6:6218:~:text=6191-,%2D%2D,%2D%2D,-6194) on my local machine, everything looks good. Here is my local run cmd print:
```
➜  torch-mlir git:(tosa-custom-softmax) ✗ torch-mlir-opt test/Conversion/TorchToTosa/basic.mlir -convert-torch-to-tosa -split-input-file -verify-diagnostics | FileCheck  test/Conversion/TorchToTosa/basic.mlir
➜  torch-mlir git:(tosa-custom-softmax) ✗ 
```

AmosLewis:
> 

Yes. The RFC is on the way by my partner. I just try to do something before the RFC is done. This is the initial step for demonstrating custom op could be used to handle complex ops that we don't want to decompose.  There should be a pass/list where we add ops and only ops from this list should go through the custom op route, the rest of op should be lower normally.

silvasean:
Let's wait for the RFC to discuss more. Per-op custom patterns are definitely not the right way to do this. Maybe we could have a pass `torch-to-tosa-convert-to-custom{op-list=aten.softmax.int,...}` which systematically converts any Torch op to the appropriate tosa.custom.

sjarus:
This is an interesting PR. Integer domain softmax _could_ be done using a custom op, but need not be. There is a working and bit accurate decomposition of TFLite softmax that is fairly detailed but works: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/tosa/transforms/legalize_common.cc#L1375

Please note that the legalize_common.cc there is the same as TosaLegalizeCommon.cc - the goal here is to eventually move these builders into the llvm-project TOSA codebase. 

The straightforward solution here is to copy the convertSoftMaxOp() into TosaLegalizeCommon.cpp, then just call that. It should do all the work for you. This approach is used by several other TorchToTosa legalizations that invoke calls to TosaLegalizeCommon to implement legalizations using previously solved pattern replacements.

sjarus (changes requested):
Please see previous comment.

AmosLewis:
> Let's wait for the RFC to discuss more. Per-op custom patterns are definitely not the right way to do this. Maybe we could have a pass `torch-to-tosa-convert-to-custom{op-list=aten.softmax.int,...}` which systematically converts any Torch op to the appropriate tosa.custom.

The RFC is ready here. [https://github.com/llvm/torch-mlir/issues/1519](https://github.com/llvm/torch-mlir/issues/1519) @ramiro050 @sjarus

sjarus:
Adding @eric-k256

AmosLewis:
Here is my most recent local result:
```
➜  torch-mlir git:(tosa-custom-softmax) ✗ torch-mlir-opt -convert-torch-to-tosa="custom-ops=torch.aten.softmax.int"    test/Conversion/TorchToTosa/custom.mlir  | externals/llvm-project/mlir/utils/generate-test-checks.py

// NOTE: Assertions have been autogenerated by utils/generate-test-checks.py

// The script is designed to make adding checks to
// a test case fast, it is *not* designed to be authoritative
// about what constitutes a good test! The CHECK should be
// minimized and named to reflect the test intent.



// CHECK-LABEL:   func.func @torch.aten.softmax.int$cst_dim(
// CHECK-SAME:                                              %[[VAL_0:.*]]: !torch.vtensor<[2,3],f32>) -> !torch.vtensor<[2,3],f32> {
// CHECK:           %[[VAL_1:.*]] = torch_c.to_builtin_tensor %[[VAL_0]] : !torch.vtensor<[2,3],f32> -> tensor<2x3xf32>
// CHECK:           %[[VAL_2:.*]] = torch.constant.none
// CHECK:           %[[VAL_3:.*]] = torch.constant.int 1
// CHECK:           %[[VAL_4:.*]] = "tosa.custom"(%[[VAL_1]]) {dim = 1 : i64, identifier = "softmax"} : (tensor<2x3xf32>) -> tensor<2x3xf32>
// CHECK:           %[[VAL_5:.*]] = torch_c.from_builtin_tensor %[[VAL_4]] : tensor<2x3xf32> -> !torch.vtensor<[2,3],f32>
// CHECK:           return %[[VAL_5]] : !torch.vtensor<[2,3],f32>
// CHECK:         }
```

AmosLewis:
Just got a E2E test bug about the option I added:
```
python -m e2e_testing.main --config=tosa -v
: CommandLine Error: Option 'custom-ops' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```

include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
```diff
@@ -28,7 +28,19 @@ void createTorchBackendToLinalgOnTensorsBackendPipeline(OpPassManager &pm);
...
+    : public PassPipelineOptions<TosaBackendPipelineOptions> {
+  // If this option is true, custom complex operations.
+  // If this option is false, skip decomposition of complex operations.
+  Option<bool> custom{*this, "custom-ops",
+                      llvm::cl::desc("custom complex operations."),
+                      llvm::cl::init(true)};
+  ListOption<std::string> customOps{
+      *this, "custom-ops",
+      llvm::cl::desc("List of ops to be converted to the backend."),
+      llvm::cl::ZeroOrMore};
```
ramiro050, line 41: I think the build issue is here. There are two options both with the same name `"custom-ops"`

AmosLewis: Make sense. Thanks, Ramiro. I have this fixed, could you review and approved this patch if everything else makes sense?

AmosLewis:
> Please see previous comment.
Please review this patch again. The purpose of this patch is to work with the op that we choose to not decompose.

AmosLewis:
> Please see previous comment.

Please review this patch again. The purpose of this patch is to work with the op that we choose to not decompose.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3854,6 +3902,9 @@ class ConvertTorchToTosa : public ConvertTorchToTosaBase<ConvertTorchToTosa> {
     INSERT_ATENOP_PATTERN(AtenSliceTensorOp);
     INSERT_ATENOP_PATTERN(AtenBroadcastToOp);
     INSERT_ATENOP_PATTERN(AtenWhereSelfOp);
+    if(!customOps.empty()){
```
eric-k256, line 3905: This feels awkward, and isn't checking the value of the customOps to make sure that softmax is in the customOp list. I think a better option is to always insert the pattern, and then at the top of the softmax matchAndRewrite, search for torch.aten.softmax.int in the string, and use that to determine whether to create the custom op. 

Some checking that the customOp list is valid would be good. With this change, someone could put any op in the customOp list and get no warning that the pass will not do what is being asked. At least having a list of ops that support legalization to tosa.custom could be used to warn if an op was passed that wasn't expected.

include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
```diff
@@ -28,7 +28,19 @@ void createTorchBackendToLinalgOnTensorsBackendPipeline(OpPassManager &pm);
 
 /// Creates a pipeline that lowers from the torch backend contract to the
 /// TOSA backend contract.
-void createTorchBackendToTosaBackendPipeline(OpPassManager &pm);
+struct TosaBackendPipelineOptions
+    : public PassPipelineOptions<TosaBackendPipelineOptions> {
+  // If this option is true, custom complex operations.
+  // If this option is false, skip decomposition of complex operations.
+  Option<bool> enableCustomOps{*this, "enableCustomOps",
```
eric-k256, line 35: I don't see any other reference to this, is this option needed, or is the custom-ops ListOption sufficient?

eric-k256 (commented):
This feels halfway between being a one-off solution for torch.aten.softmax.int, and providing a flexible way to handle many ops in this manner. With some relatively small changes, I think it can be an example use of the flexible mechanism.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3028,6 +3028,50 @@ LogicalResult ConvertAtenOp<AtenWhereSelfOp>::matchAndRewrite(
 }
 
 
+template <>
+LogicalResult ConvertAtenOp<AtenSoftmaxIntOp>::matchAndRewrite(
```
ramiro050, line 3032: I think there is still more work to be done before I would be comfortable merging this PR. The first reason is that this creates a precedent of allowing per-op custom patterns for converting to `tosa::CustomOp`. Other contributors of torch-mlir who might need a different op lowered to `tosa::CustomOp` might follow this approach, and we definitely don't want future PRs adding more per-op custom patterns to `tosa::CustomOp`.

The second reason is that this is not yet a solution that can be generalized nicely to other ops. Two main reasons come to mind at the moment:

1.  How would this approach work if there was also a conversion pattern for `AtenSoftmaxIntOp` that converts to standard tosa ops? Which pattern would get applied when both patterns are active? This is where having a separate pass for custom ops simplifies things.

2. This does not address one of the issues raised in the RFC, which is the handling of types such as `!torch.none`. This PR simply ignores the third argument of `AtenSoftmaxIntOp`. If users start depending on this convention for this op, it will make it harder to change things in the future. We need to first come up with a good design for the rules of conversion for all types that users can then depend on. Essentially, we need to clearly define a `tosa::CustomOp` contract. For example, are we always going to convert constant `int`s to rank-1 tensors as is done in this PR, or should we use rank-0 tensors or encode the int as an attribute? This is crucial to get right in the beginning so that users can have something stable to depend on.

I think we should continue iterating on this.

ramiro050 (changes requested):

sjarus:
> This feels halfway between being a one-off solution for torch.aten.softmax.int, and providing a flexible way to handle many ops in this manner. With some relatively small changes, I think it can be an example use of the flexible mechanism.

Also, isn't custom() being deprecated or at least being replaced with interface() aligned with ml_program ?

I think that torch softmax quantized is better off leveraging the existing functional lowering into TosaLegalizeCommon that is bit accurate for tfl.softmax.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: Can we split it into small pieces with several PR? We are not trying to implement everything of https://github.com/llvm/torch-mlir/issues/1519 in one PR.

AmosLewis: > I think there is still more work to be done before I would be comfortable merging this PR. The first reason is that this creates a precedent of allowing per-op custom patterns for converting to `tosa::CustomOp`. Other contributors of torch-mlir who might need a different op lowered to `tosa::CustomOp` might follow this approach, and we definitely don't want future PRs adding more per-op custom patterns to `tosa::CustomOp`.
> 
> The second reason is that this is not yet a solution that can be generalized nicely to other ops. Two main reasons come to mind at the moment:
> 
> 1. How would this approach work if there was also a conversion pattern for `AtenSoftmaxIntOp` that converts to standard tosa ops? Which pattern would get applied when both patterns are active? This is where having a separate pass for custom ops simplifies things.
> 2. This does not address one of the issues raised in the RFC, which is the handling of types such as `!torch.none`. This PR simply ignores the third argument of `AtenSoftmaxIntOp`. If users start depending on this convention for this op, it will make it harder to change things in the future. We need to first come up with a good design for the rules of conversion for all types that users can then depend on. Essentially, we need to clearly define a `tosa::CustomOp` contract. For example, are we always going to convert constant `int`s to rank-1 tensors as is done in this PR, or should we use rank-0 tensors or encode the int as an attribute? This is crucial to get right in the beginning so that users can have something stable to depend on.
> 
> I think we should continue iterating on this.

Can we split it into small pieces with several PR? We are not trying to implement everything of https://github.com/llvm/torch-mlir/issues/1519 in one PR.

AmosLewis: Can we split it into small pieces with several PR? We are not trying to implement everything of https://github.com/llvm/torch-mlir/issues/1519 in one PR.

ramiro050: Sure! It's fine to implement in chunks with unimplemented paths resulting in an error with a good error message. Given that the first PR will lay down the structure that future PRs will build on, for this RFC the first PR should include a document describing the `tosa::CustomOp` contract as well as create the `TorchToTosaCustom` pass, and have a simple working example of an op being systematically converted to `tosa::CustomOp`.

include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
AmosLewis: Deleted

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: Cool, I just create a new ConvertTorchToTosaCustom pass and add the document in the description. The example is also there.

include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
```diff
@@ -28,7 +28,14 @@ void createTorchBackendToLinalgOnTensorsBackendPipeline(OpPassManager &pm);
 
 /// Creates a pipeline that lowers from the torch backend contract to the
 /// TOSA backend contract.
-void createTorchBackendToTosaBackendPipeline(OpPassManager &pm);
+struct TosaBackendPipelineOptions
+    : public PassPipelineOptions<TosaBackendPipelineOptions> {
+  ListOption<std::string> customOps{
+      *this, "custom-ops",
+      llvm::cl::desc("List of ops to be converted to the backend."),
```
ramiro050, line 35: nit: I think it's a bit clearer to say "List of ops to be converted to tosa::CustomOp"

lib/Conversion/TorchToTosa/TorchToTosaCustom.cpp
```diff
@@ -0,0 +1,143 @@
...
+public:
+  using OpConversionPattern<AtenOpT>::OpConversionPattern;
+  using OpAdaptor = typename AtenOpT::Adaptor;
+  LogicalResult
+  matchAndRewrite(AtenOpT op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override;
+};
+
+template <>
+LogicalResult ConvertAtenOp<AtenSoftmaxIntOp>::matchAndRewrite(
```
ramiro050, line 45: This is not a systematic conversion to `tosa::CustomOp`. By systematic, I mean that the pattern should apply to every op in the IR and it should convert to `tosa::CustomOp` without using any op-specific knowledge. In other words, the transformations should be done on `Operation *` objects.

lib/Dialect/TorchConversion/Transforms/Passes.cpp
```diff
@@ -96,7 +96,11 @@ void TorchConversion::createTorchBackendToLinalgOnTensorsBackendPipeline(
 }
 
 void TorchConversion::createTorchBackendToTosaBackendPipeline(
-    OpPassManager &pm) {
+    OpPassManager &pm, const TosaBackendPipelineOptions &options) {
+  if(!options.customOps.empty()){
```
ramiro050, line 100: nit: run `git clang-format`. Spacing is off

include/torch-mlir/Conversion/Passes.td
```diff
@@ -114,6 +114,32 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "func::FuncOp"> {
   let constructor = "mlir::torch::createConvertTorchToTosaPass()";
 }
 
+def ConvertTorchToTosaCustom : Pass<"convert-torch-to-tosa-custom", "func::FuncOp"> {
+        let summary = "Convert Torch ops to TOSA custom ops";
+  let description = [{
+    The purpose to use tosa::custom is handle complex ops when we donnot
```
ramiro050, line 120: This is not a contract. A contract is something that describes in detail exactly what the expectations are before or after some transformation happens. For example, the backend contract in torch-mlir describes what the expectations are after conversion to torch backend is finished: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#the-backend-contract.

In this case, the contract should outline what each of the different torch dialect types will get turned into when this pass is finished. It should also outline exactly in what way attribute strings will be used (if they are used at all) to encode arguments. This document is meant to help users understand exactly what will happen when this pass takes place from just looking at the aten op signatures. Otherwise, there is no way for users to reliably know how to further lower `tosa::CustomOp` to their backend.

ramiro050 (dismissed):

ramiro050:
> > This feels halfway between being a one-off solution for torch.aten.softmax.int, and providing a flexible way to handle many ops in this manner. With some relatively small changes, I think it can be an example use of the flexible mechanism.
> 
> Also, isn't custom() being deprecated or at least being replaced with interface() aligned with ml_program ?
> 
> I think that torch softmax quantized is better off leveraging the existing functional lowering into TosaLegalizeCommon that is bit accurate for tfl.softmax.

@AmosLewis, can you address this?

lib/Conversion/TorchToTosa/TorchToTosaCustom.cpp
AmosLewis: I thought we can do the #define INSERT_ATENOP_PATTERN(AtenOp) ...     kind of thing for every op just like what the TorchToTosa.cpp has done.

include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
AmosLewis: Done

lib/Dialect/TorchConversion/Transforms/Passes.cpp
AmosLewis:
```
(mlir_venv) nod% git clang-format .
clang-format did not modify any files
```

ramiro050: `git clang-format` will only modify staged files. To modify the files already committed, you need to specify the range of commits. For example: `git clang-format HEAD~1`

lib/Conversion/TorchToTosa/TorchToTosaCustom.cpp
ramiro050: No, we don't want to write a new pattern for every possible PyTorch op (this is what I meant in my previous comment when I mentioned that we don't want "per-op custom patterns for converting to tosa::CustomOp"). This type of transformation should be achievable with a single pattern that applies to all ops.

ramiro050: > And Since the !torch.none kind of type conversion is not handled yet, and 'tosa.custom' op operand must be a tensor of number values.

If this PR does not yet plan on handling `!torch.none`, then you should choose a simpler op that does not take `!torch.none`s as inputs for the first PR.

> I don't think there is a general way to handle all ops.

If there is no general way of converting a good subset of PyTorch ops to `tosa::CustomOp`, then we should improve the interface to `tosa::CustomOp` first, otherwise this will bring little value to users. However, I do think you should already be able to handle a good subset of ops, namely, all ops that at compile time take either tensors as input or constants. I would focus on this and reject all other ops for now.

eric-k256:
> 

We had thoughts of making bigger changes to tosa.custom, and calling it tosa.interface. That hasn't happened, but I do have a proposed change to tosa.custom in the dialect here: https://reviews.llvm.org/D137133
I think that makes the generic Operation* mapping simpler, as you can now encode attributes into the implementation_attrs 

The other point Suraj was making is that we do have a legalization for softmax int into a series of TOSA ops that could be used to work aorund this specific problem. The larger problem of conversion from torch to tosa.custom remains, which the pass described by Ramiro aims to solve.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: Done

lib/Conversion/TorchToTosa/TorchToTosaCustom.cpp
AmosLewis: Any suggestion how to handle the **!torch.none**?

lib/Dialect/TorchConversion/Transforms/Passes.cpp
AmosLewis: Done

include/torch-mlir/Conversion/Passes.td
AmosLewis: Done

AmosLewis:
@ramiro050  I just added a generic MatchAndRewrite for torch-to-tosa-custom pass. No need to add write new MatchAndRewrite for new ops. It worked for the 2 examples in custom.mlir. Please review. We can add more type conversion support to it later. Please review.

include/torch-mlir/Conversion/Passes.td
```diff
@@ -114,6 +114,35 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "func::FuncOp"> {
...
```
ramiro050, line 120: nit: `donnot` -> `do not`

include/torch-mlir/Conversion/Passes.td
```diff
@@ -114,6 +114,35 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "func::FuncOp"> {
...
+    want to decompose them into simple ops.
+    The aten op name will used to construct a StringAttr as the identifier attribute for tosa::CustomOp.
```
ramiro050, line 122: `will used` -> `will be used`

include/torch-mlir/Conversion/Passes.td
```diff
@@ -114,6 +114,35 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "func::FuncOp"> {
...
+    So in the output the users know where is the tosa::CustomOp is coming from by the StringAttr id.
```
ramiro050, line 123: `where is the` -> `where the`

include/torch-mlir/Conversion/Passes.td
```diff
@@ -114,6 +114,35 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "func::FuncOp"> {
...
+    Each input arg from Aten Dialect has to be converted to a tensor of number values as the
+    operand of tosa::CustomOp op. After convert, use ValueRange/SmallVector to include
+    all operands as the final input operands for tosa::CustomOp.
+    The contract to follow:
+        AnyTorchTensorType 	->	TensorType<?xAnyType>  of tosa::ConstOp
```
ramiro050, line 128: nit: I would replace `?` with `AnySize`, since `?` means the type has a single dynamic dimension

include/torch-mlir/Conversion/Passes.td
```diff
@@ -114,6 +114,35 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "func::FuncOp"> {
...
+def ConvertTorchToTosaCustom : Pass<"convert-torch-to-tosa-custom", "func::FuncOp"> {
+        let summary = "Convert Torch ops to TOSA custom ops";
+  let description = [{
+    The purpose to use tosa::custom is handle complex ops when we donnot
+    want to decompose them into simple ops.
+    The aten op name will used to construct a StringAttr as the identifier attribute for tosa::CustomOp.
+    So in the output the users know where is the tosa::CustomOp is coming from by the StringAttr id.
+    Each input arg from Aten Dialect has to be converted to a tensor of number values as the
+    operand of tosa::CustomOp op. After convert, use ValueRange/SmallVector to include
+    all operands as the final input operands for tosa::CustomOp.
```
ramiro050, line 126: I don't think this last sentence is necessary. It's more of an implementation detail.

include/torch-mlir/Conversion/Passes.td
```diff
@@ -114,6 +114,35 @@ def ConvertTorchToTosa : Pass<"convert-torch-to-tosa", "func::FuncOp"> {
...
+    The contract to follow:
+        AnyTorchTensorType 	->	TensorType<?xAnyType>  of tosa::ConstOp
+        Torch_IntType		->      RankedTensorType<1xi64>  of tosa::ConstOp
+        Torch_FloatType		->      RankedTensorType<1xf32>  of tosa::ConstOp
+        ...
```
ramiro050, line 131: I would also add a TODO explaining the things that are currently unsupported.

ramiro050 (changes requested):

include/torch-mlir/Conversion/Passes.td
AmosLewis: Done

include/torch-mlir/Conversion/Passes.td
AmosLewis: Done

lib/Conversion/TorchToTosa/TorchToTosaCustom.cpp
```diff
@@ -0,0 +1,150 @@
...
+        inputs_vec.push_back(operand_tensor_float.value());
+      } else if (adaptor_operand_type
+                     .isa<mlir::TensorType>()) { // Torch::ValueTensorType
+        inputs_vec.push_back(*adaptor.getODSOperands(i).begin());
+      } else {
+        // TODO Handle more types like !torch.list<...>, !torch.device,
+        // !torch.string, !torch.none, !torch.generator.
+        return rewriter.notifyMatchFailure(
+            op,
+            "unimplemented: inputs type. The input has to be int/float ");
```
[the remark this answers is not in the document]
AmosLewis: Done

lib/Conversion/TorchToTosa/TorchToTosaCustom.cpp
```diff
@@ -0,0 +1,150 @@
...
+        auto operand_tensor_int = tosa::getConstTensor<int64_t>(
+            rewriter, op.getOperation(), operand_tosa, {1});
+        inputs_vec.push_back(operand_tensor_int.value());
+      } else if (adaptor_operand_type
+                     .isa<mlir::FloatType>()) { // Torch::ConstantFloatOp
+        double operand_tosa;
+        if (!matchPattern(operand, m_TorchConstantFloat(&operand_tosa)))
+          return rewriter.notifyMatchFailure(
+              op, "unimplemented: operand should be a torch.constant.float");
+        auto operand_tensor_float = tosa::getConstTensor<int64_t>(
```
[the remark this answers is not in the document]
AmosLewis: Done

lib/Conversion/TorchToTosa/TorchToTosaCustom.cpp
```diff
@@ -0,0 +1,150 @@
...
+  using OpConversionPattern<AtenOpT>::OpConversionPattern;
+  using OpAdaptor = typename AtenOpT::Adaptor;
+  LogicalResult
+  matchAndRewrite(AtenOpT op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    ValueRange adaptor_operands = adaptor.getOperands();
+    int num_operands = adaptor_operands.size();
+    std::vector<mlir::Value> inputs_vec;
+    for (int i = 0; i < num_operands; i++) {
+      auto operand = *op.getODSOperands(i).begin();
```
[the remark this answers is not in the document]
AmosLewis: Done

include/torch-mlir/Conversion/Passes.td
AmosLewis: Done

include/torch-mlir/Conversion/Passes.td
AmosLewis: Done

include/torch-mlir/Conversion/Passes.td
AmosLewis: Done

include/torch-mlir/Conversion/Passes.td
AmosLewis: Done

Svoch:
Folks, I believe the approach I have in my recent PR (which is in reference to the discussion in [this Selective Op decomposition RFC](https://github.com/llvm/torch-mlir/issues/1519)) addresses the outstanding comments in this PR.

@AmosLewis - do you mind if we move the discussion over to [the other PR](https://github.com/llvm/torch-mlir/pull/1563)?

AmosLewis:
> Folks, I believe the approach I have in my recent PR (which is in reference to the discussion in [this Selective Op decomposition RFC](https://github.com/llvm/torch-mlir/issues/1519)) addresses the outstanding comments in this PR.
> 
> @AmosLewis - do you mind if we move the discussion over to [the other PR](https://github.com/llvm/torch-mlir/pull/1563)?
Sure, Let's relocate to your patch.

PR #1523, opened 2022-10-24
[Bazel] Update to Ubuntu-22.04 and clang-16 for Bazel build docker

sjain-stanford:
* Update Ubuntu and clang in the docker container
* Specifically build just `@torch-mlir//:torch-mlir-opt`

Triggered GHA run:
https://github.com/sjain-stanford/torch-mlir/actions/runs/3317006870/jobs/5479411204

### DIFF
--- a/utils/bazel/docker/Dockerfile
+++ b/utils/bazel/docker/Dockerfile
@@ -1,35 +1,40 @@
-ARG BASE_IMG=ubuntu:18.04
+ARG BASE_IMG=ubuntu:22.04
 FROM ${BASE_IMG} as dev-base
 
-ARG ARCH="x86_64"
-ARG BAZEL_VERSION=5.2.0
-
 # Install basic packages
-RUN apt-get update                                              && \
-    apt-get install -y                                             \
-    clang-10                                                       \
-    curl                                                           \
-    git                                                            \
-    python3-pip                                                    \
-    python3.8                                                      \
-    python3.8-dev                                                  \
-    wget                                                           \
+RUN apt-get update && \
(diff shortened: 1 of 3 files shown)


utils/bazel/docker/Dockerfile
```diff
@@ -1,35 +1,40 @@
...
     unzip
 
-RUN update-alternatives --install /usr/bin/python python /usr/bin/python3.8 10
-RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.8 10
-
-RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-10 10
-RUN update-alternatives --install /usr/bin/clang++ clang++ /usr/bin/clang++-10 10
+# Install clang
+ARG REPO_NAME="deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy main"
+RUN echo $REPO_NAME >> /etc/apt/sources.list.d/llvm.list && \
```
asaadaldien, line 17: IIRC clang-14 was available on this distro, curious why do we need a newer version ?

asaadaldien (approved):

utils/bazel/docker/Dockerfile
sjain-stanford: This is basically mirroring the [clang setup](https://github.com/llvm/torch-mlir/blob/6c31b06922efd5f0e1063aebbd63b6a322c64da9/build_tools/docker/Dockerfile#L29) from the regular (cmake) docker builds for consistency. LMK if you see a concern with using the newer version.

PR #1524, opened 2022-10-25
[MHLO] reimplement linear lowering torchToMhlo

zzpmiracle:
1. aten.linear input can be any rank, weight should be 2-D tensor, according to https://pytorch.org/docs/stable/generated/torch.nn.Linear.html
2. currently, use bmm to address linear not have good performance, use reshape and mhlo.dot to get better performance.

### DIFF
--- a/lib/Conversion/TorchToMhlo/Linear.cpp
+++ b/lib/Conversion/TorchToMhlo/Linear.cpp
@@ -373,10 +373,10 @@ public:
     auto lhsRank = lhsTy.getRank();
     auto rhsRank = rhsTy.getRank();
 
-    if (lhsRank != 2 && lhsRank != 3)
-      return op.emitError("aten.Linear called but input rank not 2 or 3");
-    if (rhsRank != 2 && rhsRank != 3)
-      return op.emitError("aten.Linear called but weight rank not 2 or 3");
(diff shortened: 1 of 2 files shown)


Vremold (approved):
LGTM. Are there more linear-related mhlo e2e tests, e.g., Mlp1LayerModule_basic, which pass after this PR?

li-plus:
@zzpmiracle @tanyokwok This PR is causing compile error with the latest mlir version. Would you revert and fix that?

```
./torch-mlir/lib/Conversion/TorchToMhlo/Linear.cpp:427:24: error: no member named 'kDynamicSize' in 'mlir::ShapedType'; did you mean 'kDynamic'?
          {ShapedType::kDynamicSize, lhsTy.getShape()[lhsRank - 1]},
           ~~~~~~~~~~~~^~~~~~~~~~~~
                       kDynamic
./torch-mlir/build/tools/mlir/include/mlir/IR/BuiltinTypeInterfaces.h.inc:106:30: note: 'kDynamic' declared here
    static constexpr int64_t kDynamic =
                             ^
1 error generated.
```

tanyokwok:
OK, I will revert this: https://github.com/llvm/torch-mlir/pull/1744

PR #1526, opened 2022-10-26
[MLIR][TORCH] Add E2E support for `aten.addcmul_` and `aten.addcdiv_` op

vivekkhandelwal1:

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -482,6 +482,8 @@ TOSA_PASS_SET = {
     "HardTanhIntModule_basic",
     "AtenRoundIntModule_basic",
     "MseLossNoReductionModule_basic",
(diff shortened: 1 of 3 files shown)


python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -1360,6 +1360,54 @@ def AddCDivModule_basic(module, tu: TestUtils):
 # ==============================================================================
 
 
+class AddCMul_Module(torch.nn.Module):
```
ramiro050, line 1363: No need to add these tests. These underscore ops will get replaced with the corresponding non-underscore variant before reaching any of the backends, so the tests for the non-underscore variants are sufficient.

ramiro050 (approved):

Issue #1492, opened 2022-10-13
Some iteration-related operations fail with a confusing error message


dellis23:
The following error occurs when I try to do a number of operations involving iteration:

```python
torch_mlir.compiler_utils.TorchMlirCompilerError: Lowering TorchScript IR -> Torch Backend IR failed with the following diagnostics:
error: unsupported by backend contract: tensor with unknown rank
note: see current operation: %6 = "torch.tensor_static_info_cast"(%arg0) : (!torch.vtensor<[3],f32>) -> !torch.vtensor<*,f32>
note: this is likely due to a missing shape transfer function in shape_lib_gen.py
```

Here's a script that reproduces various cases that cause the failure, and some that do not:

```python
import functorch
import torch
import torch_mlir

from functorch._src.compile_utils import strip_overloads


# Fails
def tuple_stuff(w, grad_w):
    for i, grad in enumerate(grad_w):
        w[i] = w[i] + grad
    return w


# Fails
def tuple_stuff_2(w):
    for i, item in enumerate(w):
        return w[i]


# Fails
def tuple_stuff_3(w):
    for i in w:
        return i
        

# Fails
def tuple_stuff_4(w):
    for i in w:
        return torch.tensor(3.0) * i


# Passes
def tuple_stuff_5(w):
    size = len(w)
    for i in range(size):
        return w[i] * 2.


# Fails
def tuple_stuff_6(w):
    size = len(w)
    for i in range(size):
        w[i] = w[i] * 2.
    return w


def main():
    w = torch.tensor([1., 2., 3.])
    grad_w = torch.tensor([1., 1., 1.])
    graph = functorch.make_fx(tuple_stuff_6)(w)
    strip_overloads(graph)
    linalg_on_tensors_mlir = torch_mlir.compile(
        graph,
        (w,),
        output_type=torch_mlir.OutputType.LINALG_ON_TENSORS,
        use_tracing=False)


if __name__ == "__main__":
    main()
```

It's unclear to me exactly what conditions cause things to break.

I talked to Ramiro about this, and it looks like the following operation is the culprit:

```python
    %24 = torch.operator "aten.unbind.int"(%23, %int0) : (!torch.tensor<*,f32>, !torch.int) -> !torch.list<tensor>
```

A few questions:

 * What exactly is the cause or causes of the failures?
 * Are they something we can fix or support?
 * If not, can we throw a better error message to communicate to the user what the problem is and how to fix it?

dellis23:
[...]

PR #1527, opened 2022-10-26
Rewrite unbind + getitem to simple tensor index op

dellis23:
See: https://github.com/llvm/torch-mlir/issues/1492

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -6442,6 +6442,29 @@ def Torch_AtenTypeAsOp : Torch_Op<"aten.type_as", [
   let hasFolder = 1;
 }
 
+def Torch_AtenUnbindIntOp : Torch_Op<"aten.unbind.int", [
+    AllowsTypeRefinement,
+    ReadOnly
+  ]> {
+  let summary = "Generated op for `aten::unbind.int : (Tensor, int) -> (Tensor[])`";
+  let arguments = (ins
+    AnyTorchTensorType:$self,
+    Torch_IntType:$dim
+  );
+  let results = (outs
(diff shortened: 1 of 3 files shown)


dellis23:
@silvasean: @ramiro050 and I have been talking about this and aren't sure if this is the best approach.  We're having trouble with dead code elimination not actually eliminating the unbind even after all references are removed.  Marking it as pure seems to fix that, but we only see one other op marked pure.

One alternative we considered: every time we check a `getitem` and find an `unbind` preceding it, see if it's the last consumer of the `unbind`, and if so, remove it.  Just feels a bit hacky and like the dce pass should handle this.

Also, this PR still needs:

 1) To be updated to handle the dim argument to `unbind` correctly
 2) Tests

But I wanted to figure out the approach we should take first.

ramiro050:
> Marking it as pure seems to fix that, but we only see one other op marked pure.

To add to this, marking it as `Pure` does not seem right to me since this op in PyTorch can raise an exception if the `dim` argument is invalid. I suppose because most PyTorch ops can raise exceptions, we will probably not be able to have a simple general approach to dead code elimination.

> One alternative we considered: every time we check a getitem and find an unbind preceding it, see if it's the last consumer of the unbind, and if so, remove it. Just feels a bit hacky and like the dce pass should handle this.

Another approach that feels a bit nicer is to add a folder for the `unbind.int` op that removes the op if the arguments are valid and the op does not have any users. As far as I can tell, this is the only valid approach for dead code elimination of torch dialect ops: we first need to verify that the inputs are valid before we can proceed with removing it. I personally think it's fine to treat `unbind.int` in a special way by adding this folder to it since ops like this one (ops that return lists of tensors) are rare.

The general approach would be to have a pass that runs a special verification function on each op and removes the ones that pass the verification and have no users. However, this would require a lot of work (defining custom verifiers for each op) with not much in return.

PR #718, opened 2022-03-30
[MLIR][TORCH] Add E2E support for aten.avg_pool2d op

vivekkhandelwal1:
This commit adds lowering of `aten.avg_pool2d` op.

Signed-Off By: Vivek Khandelwal <vivek@nod-labs.com>

PR #1530, opened 2022-10-27
[MLIR][TORCH] Fix wrong parameter name "supportFPInputOnly"

xndcn:
The parameter "supportFPInputOnly" of function createPoolingOp() is supposed to be "supportNonFPInput", which was added to distinguish between "MaxPool2d" and "AvgPool2d" op in #718


ramiro050 (approved):
Good catch! LGTM

PR #1531, opened 2022-10-27
Remove all but one of valsem ops + move fill.Scalar to elementwise

ramiro050:
This commit removes almost all of the valsem ops, since the value semantics version of the ops now exist in PyTorch. The only op missing is `aten.bernoulli_.float`. In addition, this commit also simplifies the implementation of `aten.fill.Scalar` by moving it to the pattern that converts elementwise ops.


build_tools/autogen_ltc_backend.yaml
```diff
@@ -23,6 +24,7 @@ blacklist:
 - arange
 - arange.start
 - arange.start_step
+- fill.Scalar
```
ramiro050, line 27: @antoniojkim, after adding support for `aten.fill.Scalar` in torch-mlir, some of the tests that use the op `aten.fill_.Scalar` (note the underscore), started to fail with the error:

```
        Runtime error: Traceback (most recent call last):
          File "/usr/local/google/home/ramiroleal/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 302, in compile_and_run_test
            trace = config.run(compiled, golden_trace)
          File "/usr/local/google/home/ramiroleal/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/configs/lazy_tensor_core.py", line 35, in run
            output = getattr(artifact, item.symbol)(*lazy_inputs)
          File "/usr/local/google/home/ramiroleal/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/constant_alloc.py", line 364, in forward
            memory_format=torch.preserve_format).fill_(0)
        RuntimeError: Unimplemented Error: 
        	/usr/local/google/home/ramiroleal/torch-mlir/build/tools/torch-mlir/python/torch_mlir/csrc/base_lazy_backend/generated/shape_inference.cpp:35    std::vector<torch::lazy::Shape> torch::lazy::compute_shape_fill(const at::Tensor &, const at::Scalar &)
```

My guess is that LTC was converting the ops to `aten.fill.Scalar`, causing issues. I noticed that `aten.fill.Scalar` is currently not supported in the [TS backend](https://github.com/pytorch/pytorch/blob/master/aten/src/ATen/native/ts_native_functions.yaml), so I added the op to this list for consistency. This fixed the test failures.

Let me know if this change is right or if something else needs to be modified.

vivekkhandelwal1 (approved):
LGTM

build_tools/autogen_ltc_backend.yaml
antoniojkim: Yep, the change looks right to me

PR #1533, opened 2022-10-28
Drop 3.9 binaries to keep under 6hrs build

powderluv:
Since we added 3.7 for Collab drop 3.9 to keep build times under check.


dellis23:
Thanks!

PR #1535, opened 2022-10-29
Set proper version info for Windows

powderluv:
Set the correct package version for Windows release builds


powderluv:
Here is a successful build with the correct version

https://github.com/llvm/torch-mlir/releases/tag/oneshot-20221030.65

vivekkhandelwal1 (approved):

PR #1537, opened 2022-10-31
mlir: add missing dependency on TableGen targets

ashay:
lib/Dialect/Torch/Utils/Utils.cpp includes TorchOps.h, which, by way of
included header files, refers to both TorchOps.h.inc as well as
TorchTypes.h.inc.  However, the build rules do not specify the
dependency of the `TorchMLIRTorchUtils` target on the TableGen generated
header files, causing spurious build errors.
    
This patch fixes the problem by adding `MLIRTorchOpsIncGen` and
`MLIRTorchTypesIncGen` to the list of dependencies of
`TorchMLIRTorchUtils`.


ashay:
This change is aimed at preventing errors like this one: https://github.com/llvm/torch-mlir/actions/runs/3357951008/jobs/5564259776.

ashay:
Rebased with main.

ramiro050 (approved):
Thanks!

PR #1538, opened 2022-10-31
ci: simplify workflow by not using matrix builds for different platforms

ashay:
Our platform-specific build instructions have diverged substantially, so
there exists little in common among the CI steps for each platform.
Moreover, among the 2x3x4 = 24 combinations that we start with when
using the build matrix, we only use four, so our exclude list has grown
long and almost each step has a conditional statement that limits it to
specific platforms, making it difficult to mentally trace the set of
statements that execute for each platform.
    
This patch replaces the matrix build with four parallel builds, one for
each case.  The resulting file is only 13 lines longer than before.
However, one downside of this patch is that GitHub will no longer stop
the concurrent builds if any individual platform's build fails.  The
build matrix halted concurrent builds because we used the `fast-fail`
flag, but no such flag exists for non-matrix builds.

---

The resulting builds look like https://github.com/ashay/torch-mlir/actions/runs/3359070766 in the GitHub UI.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -15,115 +15,129 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
-
-# Provisioned Jobs:
-# ubuntu/docker - x86_64 - llvm in-tree     - pytorch binary - build+test    # most used dev flow and fastest signal
-# ubuntu/docker - x86_64 - llvm out-of-tree - pytorch source - build+test    # most elaborate build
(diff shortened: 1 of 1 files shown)


ashay:
I am not sure why GitHub still shows the old CI tests as required.  Marking PR as Draft until I figure out the reason.

powderluv:
Just for some background we went with the matrix build exactly for those two features -- fast-fail and terminate old builds if we have a new push for the PR. We were seeing about 10-15 builds running for hours since someone decided to push a few patches in quick succession.

ashay:
> Just for some background we went with the matrix build exactly for those two features

Oh, I see.  Thanks for confirming.  Closing this PR now.

PR #1539, opened 2022-10-31
build: update llvm tag to 74fb770d

tanyokwok:
This commit makes the following changes needed to update bump LLVM:

-  replace usages of `tensor::createPadScalarOp`; see https://reviews.llvm.org/D136493


ashay (approved):
Thanks for the quick update!

ashay (approved):
That's some serious renaming!  Do you use some kind of a script, or did you have to make the changes manually?

tanyokwok:
> That's some serious renaming! Do you use some kind of a script, or did you have to make the changes manually?

I had a script offline that will update the filecheck if error occurs.

PR #1543, opened 2022-11-01
[MLIR][TORCH] Add E2E support for `aten.upsample_nearest2d_backward.vec` op

vivekkhandelwal1:
Hi, I'm working on adding the lowering for the op `aten::upsample_nearest2d_backward.vec` op. I have added the lowering but it's giving the wrong results. The logic is as follows:
```
grad_output: Tensor of size [n, c, oh, ow]
result: Tensor of size [n, c, ih, iw], initialized with zero
kh = ceil(oh/ih), kw = ceil(ow/iw)

for i in range(n):
    for j in range(c):
        for p in range(ih):
            for q in range(iw):
                for x in range(kh):
                    for y in range(kw):
                        result[i, j, p, q] += grad_output[i, j, (p*kh)+x, (q*kw)+y]

```
I think the cause of the wrong answer lies in this accumulation. Because when I print the result and the input tensors, I see only the last value of the kernel or window (kh x kw) getting stored. I have kept the iterator type for the first four loops as `parallel` and for the last two loops as `reduction`.

Is this logic not implementable in Torch-MLIR?

@ramiro050 @silvasean @pashu123


vivekkhandelwal1:
Please ignore the above comment, it worked by changing the `args[0] -> args[2]` here: https://github.com/vivekkhandelwal1/torch-mlir/blob/0bb19c22ed01d70823d8cb59e0b7581c2107daca/lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp#L1037
Thanks to @pashu123!

vivekkhandelwal1:
Hi @ramiro050, could you please review this PR?

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -896,6 +896,179 @@ class ConvertAtenUpsampleNearest2dVecOp
...
+        loc,
+        getAsOpFoldResult(
+            castIntVectorToIndexVector(rewriter, loc, scaleFactorsIntValues)),
+        elementType);
+
+    SmallVector<AffineExpr> affineExprs;
+    for (unsigned i = 0; i < gradOutputRank; i++)
+      affineExprs.push_back(rewriter.getAffineDimExpr(i));
+
+    AffineMap inputMap = AffineMap::get(gradOutputRank + 2, /*symbolCount=*/0,
```
ramiro050, line 1027: Try to avoid hardcoding constants. This should be `gradOutputRank + scaleFactorsIntValues.size()`

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -896,6 +896,179 @@ class ConvertAtenUpsampleNearest2dVecOp
...
+                                        affineExprs, op->getContext());
+
+    affineExprs.clear();
+    for (unsigned i = 0; i < 2; i++)
```
ramiro050, line 1031: Try to avoid hardcoding constants. You should do `i < scaleFactorsIntValues.size()`

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -896,6 +896,179 @@ class ConvertAtenUpsampleNearest2dVecOp
...
+      affineExprs.push_back(rewriter.getAffineDimExpr(gradOutputRank + i));
+
+    AffineMap kernelMap = AffineMap::get(gradOutputRank + 2, /*symbolCount=*/0,
```
ramiro050, line 1034: Try to avoid hardcoding constants. You should do `gradOutputRank + scaleFactorsIntValues.size()`

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -896,6 +896,179 @@ class ConvertAtenUpsampleNearest2dVecOp
...
+                  Value c = rewriter.create<linalg::IndexOp>(loc, 1);
+                  Value ih = rewriter.create<linalg::IndexOp>(loc, 2);
+                  Value iw = rewriter.create<linalg::IndexOp>(loc, 3);
+                  Value kh = rewriter.create<linalg::IndexOp>(loc, 4);
+                  Value kw = rewriter.create<linalg::IndexOp>(loc, 5);
+                  Value accValue = getGradOutputValue(
+                      rewriter, loc, gradOutput, elementType, n, c, ih, iw, kh,
+                      kw, gradOutputSizeIndexValues, scaleFactorsIntValues);
+                  Value result =
+                      rewriter.create<arith::AddFOp>(loc, args[2], accValue);
```
ramiro050, line 1061: nit: can you give `args[2]` a name?

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -896,6 +896,179 @@ class ConvertAtenUpsampleNearest2dVecOp
...
+    SmallVector<StringRef> iteratorTypes(gradOutputRank,
+                                         getParallelIteratorTypeName());
+    iteratorTypes.push_back(getReductionIteratorTypeName());
+    iteratorTypes.push_back(getReductionIteratorTypeName());
+
+    Value finalRes =
+        rewriter
+            .create<linalg::GenericOp>(
+                loc, outTensor.getType(), ValueRange{outTensor, kernelTensor},
+                ValueRange{outTensor},
```
ramiro050, line 1047: I don't think you need to pass `outTensor` as an input. You have access to `outTensor` as the output of the linalg::GenericOp

ramiro050 (changes requested):
Thanks for this!

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
vivekkhandelwal1: In that case, I need to create another dummy tensor which would be passed as an input. Because the iteration is over the `outTensor`, which gives these indices: https://github.com/vivekkhandelwal1/torch-mlir/blob/0c388c18e1a9260811d53c8d957a84661a876f5d/lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp#L1051-L1054

ramiro050: I'm not sure I understand. You should be able to get the indices for the `outTensor` without needing to pass `outTensor` as an input. The indices are generated using the information from the `AffineMap`s, which in this case have a domain of 6 dimensions. Since the output tensor of the `linalg::GenericOp` uses the `inputMap` as it's indexing map, then the indices should work without having to pass `outTensor` as an input.

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -896,6 +896,177 @@ class ConvertAtenUpsampleNearest2dVecOp
...
+  Value pred = builder.create<arith::AndIOp>(loc, predH, predW);
+  Value result = builder.create<arith::SelectOp>(
+      loc, pred, gradOutputValue,
+      convertScalarToDtype(builder, loc, constantZero, gradOutputElemType));
+
+  return result;
+}
+
+namespace {
+class ConvertAtenUpsampleNearest2dBackwardVecOp
```
ramiro050, line 945: Can you add the python pseudo code you have in the PR description as a comment here? I find it makes understanding the code much easier.

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
vivekkhandelwal1: Thanks, I figured this out.

vivekkhandelwal1:
> LGTM!

Thanks!

PR #1545, opened 2022-11-01
Revert "ci: build PyTorch before building Torch-MLIR"

ashay:
Reverts llvm/torch-mlir#1542 since it's causing post-merge CI failures.
See https://github.com/llvm/torch-mlir/actions/runs/3373219101/jobs/5597790989 for an example.

### DIFF
--- a/build_tools/python_deploy/build_linux_packages.sh
+++ b/build_tools/python_deploy/build_linux_packages.sh
@@ -187,17 +187,7 @@ function run_in_docker() {
 function build_in_tree() {
   local torch_from_bin="$1"
   local python_version="$2"
-
-  if [ "${torch_from_bin}" == "OFF" ]
-  then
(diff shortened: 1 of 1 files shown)


ashay:
@powderluv  Sorry for the noise.  It doesn't look like an easy to reproduce error, so I am reverting this first, while I understand why it's happening.

powderluv (approved):

powderluv:
I think the Pytorch build has to be built with the same cxx abi to avoid those failures. Seems familiar from earlier this summer when I was wrangling Pytorch builds (and have blacked out in my memory )

ashay:
> I think the Pytorch build has to be built with the same cxx abi to avoid those failures.

Ah, that explains it.  Thanks!  I wonder why we didn't see this problem in the PR tests, but I'll come back to it after the cache issues are resolved.

PR #1547, opened 2022-11-02
CI script improvements

ashay:
This PR contains four commits (see commit messages for details):

 - ci: update versions of external actions
 - ci: replace deprecated action with bash commands
 - ci: use smaller ccache artifacts to reduce evictions
 - ci: use consistent platform identifiers

@dellis23 I added you as a reviewer because I changed one of your workflows in the last commit so that the workflow runs on ubuntu-22.04 for consistency with the rest of the workflows.  Let me know if you'd like me to revert the change.  Thanks!

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -14,7 +14,7 @@ runs:
 
   steps:
   - name: Set up Python
-    uses: actions/setup-python@v2
+    uses: actions/setup-python@v4
     with:
       python-version: '3.10'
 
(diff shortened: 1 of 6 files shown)


dellis23:
22.04 LGTM

dellis23 (approved):

powderluv:
I like `-latest` to get autoupdated until it fails but I understand why it has to be pinned :D  LGTM

powderluv (approved):

ashay:
> I like `-latest` to get autoupdated until it fails but I understand why it has to be pinned :D LGTM

Thanks, that makes sense.  I don't mind using `-latest`, as long as it is consistent.  Let me update the last commit to switch all of the platforms to `-latest`.

PR #1548, opened 2022-11-03
ci: enable ccache on Windows

ashay:
This patch makes a few small, but key, changes to enable ccache on
Windows.  First, it replaces the hendrikmuhs/ccache-action action with
command line invocations to the ccache binary, since the action has two
bugs, one of which causes CI to refer to different ccache artifacts
before versus after the build on Windows whereas the other bug can
sometimes cause the action to incorrectly infer that the cache is empty.

Second, this patch slightly alters the cache key, so that our old cache
artifacts, which have grown too big, are eventually discarded in favor
of the new, smaller cache artifacts.  Along the way, this patch also
keeps the RollPyTorch's cache artifact separate from the regular build's
cache artifact so as to keep these artifacts small, and also because the
RollPyTorch action is off the critical path for most contributors.

Third, this patch makes small changes to the CMake file so that on
Windows, the ccache binary is added as a prefix, as recommended on the
[ccache Wiki](https://github.com/ccache/ccache/wiki/MS-Visual-Studio).

Finally, as an anciallary change, this patch drops the PowerShell
scripts for Windows builds and instead moves the CMake invocation into
the workflow file.  This patch doesn't invoke tests on Windows (since
the workflow prior to this patch also did not invoke tests), but I will
add the test invocation in a subsequent patch.

---

Time consumed for Windows builds:
1. [Cold cache run](https://github.com/ashay/torch-mlir/actions/runs/3381876509/jobs/5616223165):  2 hours and 30 minutes.
2. [Exact cache hit](https://github.com/ashay/torch-mlir/actions/runs/3381876509/jobs/5617680570): 25 minutes
3. [Imprecise cache hit](https://github.com/ashay/torch-mlir/actions/runs/3382941863): 25 minutes.

* Imprecise cache hits occur when there was no cache entry for the precise commit hash, so the caching layer uses the most recent cache artifact.

This change does not address the recent macOS cache regression, where an imprecise cache hit causes CI to run a full rebuild of Torch-MLIR.


powderluv:
Thank you for the great work on the CI. It may be worth leaving the Powershell script since we would want to "recreate the CI" locally and it is better to have a single entry point

powderluv (commented):
LGTM but may be worth having the .ps1 scripts around to test like the cI

ashay:
> It may be worth leaving the Powershell script since we would want to "recreate the CI" locally and it is better to have a single entry point

Excellent point.  I've added it back in, and tried to match the inlined CMake invocation.  Hopefully I got the syntax right. :)

powderluv:
oh but it is a bash script. I think we want it windows native .ps1 no ? Anyway not a big deal we can readd that later whenever required so we don't hold up your cache enhancements.

ashay:
> I think we want it windows native .ps1 no ?

Indeed, we do, but I struggled with the PowerShell syntax until I finally went back to bash.  I need to read up on variable substitution and the `Start-Process` thing before I come back to PowerShell.

> we can readd that later whenever required so we don't hold up your cache enhancements.

I think so, too.  I'll merge this PR and have a separate PR for switching to PowerShell.

PR #1552, opened 2022-11-03
Remove MakeView from IR Builder

antoniojkim:
Once https://github.com/pytorch/pytorch/pull/88445 lands in upstream PyTorch, Torch-MLIR's LTC will break because of a change in the API. Getting ahead of this by prefetching the fix.


powderluv (approved):

powderluv:
I think it is probably merged now https://github.com/llvm/torch-mlir/actions/runs/3419211001/jobs/5692446164

powderluv (approved):

PR #1554, opened 2022-11-04
[MLIR][TORCH] Add E2E support aten.convolution_backward op

vivekkhandelwal1:
This commit adds the decomposition for the `aten.convolution_backward` and `aten.convolution_backward_overrideable` op.


vivekkhandelwal1:
@ramiro050 Could you please review this PR?

gpetters94 (approved):
LGTM

python/torch_mlir_e2e_test/test_suite/backprop.py
```diff
@@ -56,7 +58,148 @@ def TanhBackward_basic(module, tu: TestUtils):
...
+            transposed=False,
+            output_padding=[0],
+            groups=1,
+            output_mask=[True, True, True])
+
+
+@register_test_case(module_factory=lambda: ConvolutionBackwardModule1D())
+def ConvolutionBackwardModule1D_basic(module, tu: TestUtils):
+    with torch.backends.mkldnn.flags(enabled=False):
+        module.forward(torch.randn(3, 3, 3), torch.randn(3, 3, 3),
```
ramiro050, line 92: use` `tu.rand`. Same applies to all other tests

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -647,11 +645,23 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+            op, rewriter, input, paddingIntValues, /*unpaddedDims=*/2);
+      } else {
+        // Case: Dynamic Padding.
+        paddingIntValues = getTypeConvertedValues(
+            rewriter, loc, getTypeConverter(), paddingIntValues);
+        SmallVector<int64_t, 4> paddingIncludingNC = {0, 0};
+        paddingIncludingNC.insert(paddingIncludingNC.end(), paddingInts.begin(),
+                                  paddingInts.end());
+        paddedInput = torch_to_linalg::getZeroPaddedTensor(op, rewriter, input,
+                                                           paddingIncludingNC);
```
ramiro050, line 663: What is the reason for adding a special static case?

lib/Conversion/TorchToLinalg/Utils.h
```diff
@@ -30,6 +30,12 @@ Value getPaddedTensor(Operation *op, OpBuilder &b, Value &input,
 Value getZeroPaddedTensor(Operation *op, OpBuilder &b, Value &input,
                           SmallVectorImpl<int64_t> &paddingInts);
 
+// Helper function that adds dynamic padding to a tensor, ignoring the first two
+// dimensions
```
ramiro050, line 34: Is the `ignoring the first two dimensions` accurate? It seems that the ignored dimensions depends on `unpaddedDims`

ramiro050 (changes requested):
Sorry for the late review. I somehow missed this PR. Below are a few comments/questions

lib/Conversion/TorchToLinalg/Linear.cpp
gpetters94: It's been a bit, I'm not sure. Let me double check.

gpetters94: @vivekkhandelwal1 It looks like only the static case is necessary, I guess the dynamic code was left over from the previous PR.

vivekkhandelwal1:
@ramiro050 Could you please review this patch?

ramiro050 (approved):
LGTM!

PR #1555, opened 2022-11-04
importer: fix usage after PyTorch update

ashay:
Unless requested otherwise, PyTorch no longer installs most of the
header files under the caffe2 directory (see
https://github.com/pytorch/pytorch/pull/87986).  This breaks our
importer code since we need to use the `MakeGuard()` function to execute
statements in the event of exceptions.

To fix this issue, this patch implements a rudimentary version of
PyTorch's ScopeGuard, where once the class variable goes out of scope,
it executes a predefined method.

### DIFF
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/ivalue_importer.cpp
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/ivalue_importer.cpp
@@ -22,7 +22,6 @@
 #include "torch-mlir-c/TorchTypes.h"
 
 #include "ATen/native/quantized/PackedParams.h"
-#include "caffe2/core/scope_guard.h"
 
 using namespace torch_mlir;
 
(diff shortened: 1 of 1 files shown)


ashay:
Btw, do we need do something special to conform with the PyTorch license if we add one of their files into our repository?

powderluv (approved):

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/ivalue_importer.cpp
```diff
@@ -153,6 +152,22 @@ class IValueImporter {
 };
 } // namespace
 
+// RAII pattern to insert an operation before going out of scope.
+class InserterGuard {
+private:
+  MlirBlock _importBlock;
```
silvasean, line 158: nit: no leading `_` per style guide. Yes this is weird but it works :)

Issue #1519, opened 2022-10-24
[RFC] Selective Op decomposition: Propagate down backend-legal Ops as Custom Ops


Svoch:
Hi folks, I'd like to get your feedback on my proposed design of selective decomposition of operations in Torch-MLIR.

I am building mainly on the notion of "backend-legal Ops" (merged in [this PR](https://github.com/llvm/torch-mlir/pull/1242)) and the ideas proposed in the design for the custom Ops support (mentioned in [this RFC](https://github.com/llvm/torch-mlir/issues/1462)).

When an operation is marked as legal in the backend, then it will not be decomposed in the "DecomposeComplexOpsPass", and is propagated down to the backend. Next there would be two possibilities:
1. A lowering pattern for the undecomposed Op exists in the backend: In this case the compilation will succeed.
2. There is no lowering pattern for the undecomposed Op in the backend: In this case the compilation fails. For example, marking an Op like `AtenLayerNormOp` and compiling will yield an IR that cannot be lowered into the backend dialects.

To handle the second case, we can mark such operations during (or before) the decomposition pass and have the backend conversion passes lower them into custom Ops. For example, consider Torch module below:
```
class SimpleModel(nn.Module):
    def __init__(self) -> None:
        super().__init__()
        self.layernorm = nn.LayerNorm(normalized_shape=4)

    def forward(self, input_tensor):
        model_out = self.layernorm(input_tensor)
        return model_out
```
When compiling this to a backend, say TOSA, using the commands below
```
torch_mlir.BACKEND_LEGAL_OPS[torch_mlir.OutputType.TOSA].append("torch.aten.layernorm")
module = torch_mlir.compile(SimpleModel(), torch.randn(4, 4), output_type=torch_mlir.OutputType.TOSA)
```
The compilation fails due to not being able to legalize unhandled Ops like the `aten.layer_norm.int` and its attributes present as `constant.int` in the Torch IR:
```
====================
Torch Backend IR
module attributes {torch.debug_module_name = "SimpleModel"} {
  func.func @forward(%arg0: !torch.vtensor<[4,4],f32>) -> !torch.vtensor<[4,4],f32> {
    %0 = torch.vtensor.literal(dense<0.000000e+00> : tensor<4xf32>) : !torch.vtensor<[4],f32>
    %1 = torch.vtensor.literal(dense<1.000000e+00> : tensor<4xf32>) : !torch.vtensor<[4],f32>
    %int4 = torch.constant.int 4
[...]

PR #1563, opened 2022-11-07
Add support to lower backend-legal operations as TOSA CustomOps

Svoch:
This change systematically lowers any Torch operation marked as "backend-legal" for the TOSA backend to a `tosa.Custom` operation.
To do so,
- Propagate the list of backend-legal ops down to the TOSA backend passes pipeline (i.e., `TorchBackendToTosaBackendPipeline`)
- Add a new pass to the TOSA backend pipeline to convert backend-legal Torch operations to custom operations in TOSA dialect.
- Within the `TorchBackendLegalToTosaCustom` conversion pass, add a generic conversion pattern to match to the input operands of the Torch operation:
  - Torch Constant operations like `torch.constant.int`, `torch.constant.float`, `torch.constant.bool`, `torch.prim.ListConstruct` match to TOSA constant Ops created based on their values.
  - Torch `NoneType` is passed down by creating a TOSA CustomOp representing "None".
  -  Any other operation is passed down as is so that the subsequent Torch to TOSA conversion pass will convert it as normal.

The backend contract for Torch to TOSA conversion of the backend-legal operations is described in detail here:

>     The pass convert-torch-backend-legal-to-tosa-custom extends the selective
>     decomposition patterns to TOSA backend, such that it will convert operations
>     marked as backend-legal by the user for the TOSA backend to CustomOps in
>     TOSA dialect.
>     The backend-legal operations will not be decomposed or lowered through
>     the existing Torch to TOSA conversion patterns, even if such patterns
>     exist in the subsequent Torch to TOSA conversion pass; instead a TOSA
>     CustomOp will be created with the specifications below:
>       1. The 'identifier' attribute in the CustomOp will be the operation
>          name of the corresponding ATen operation.
>       2. All inputs to the ATen operation will be converted into legal TOSA
>          operations. There will be no distiction between inputs and constant
>          attributes; all will be treated as inputs to the CustomOp. It is up
>          to the consuming backend to deduce them based on the ATen operation
>          semantics.
>       3. Since TOSA conversion pattern of Torch operations are responsible
>          to consume and convet the constant attributes (such as 'axis' for a
>          reduction operation), the TOSA CustomOp conversion will also match
>          and rewrite these attributes as TOSA ConstOps as well. Specifically:
>            - torch.constant.bool        -> tosa.ConstOp (of tensor type i1)
>            - torch.constant.int         -> tosa.ConstOp (of tensor type i64)
>            - torch.constant.float       -> tosa.ConstOp (of tensor type f32)
>            - torch.prim.ListConstruct   -> tosa.ConstOp (of tensor type i64)
>            - torch.constant.none        -> tosa.CustomOp (of tensor type i1)
>                                            The 'identifier' attribute of this
>                                            CustomOp is 'torch.constant.none'
>          All other Torch ATen operations will be lowered to TOSA by the Torch
>          to TOSA conversion pass after this one.
>       4. The order of the input operands of the backend-legal Torch operation
>          preserved and the TOSA CustomOp will have the same order.
   

Remaining TODO items:
1.  Extend the contract for other Torch constant operations such as `torch.constant.device`, `torch.constant.number` and `torch.constant.str`
2. Update this pass to populate the 'config' attribute of the TOSA CustomOp to establish a proper operand mapping scheme between the backend-legal Torch operation and the TOSA CustomOp. This work requires the LLVM uplift in Torch-MLIR to include the change in TOSA dialect to extend `tosa.Custom` operation with the `config` attribute (See the commit at https://github.com/llvm/llvm-project/commit/d94ee70f4f01e4d9eec49e02eff57a5655618401)

Related to https://github.com/llvm/torch-mlir/issues/1519
cc @silvasean @ramiro050 @eric-k256 @powderluv


test/Conversion/TorchToTosa/selective_decomposition.mlir
```diff
@@ -0,0 +1,19 @@
+// RUN: torch-mlir-opt <%s -convert-torch-backend-legal-to-tosa-custom="custom-ops=torch.aten._softmax" -split-input-file -verify-diagnostics | FileCheck %s
+
+// CHECK-LABEL:   func.func @torch.aten._softmax(
+// CHECK-SAME:      %[[VAL_0:.*]]: !torch.vtensor<[4,4],f32>) -> !torch.vtensor<[4,4],f32> {
+// CHECK:           %[[VAL_1:.*]] = torch_c.to_builtin_tensor %[[VAL_0]] : !torch.vtensor<[4,4],f32> -> tensor<4x4xf32>
+// CHECK:           %[[VAL_2:.*]] = torch.constant.int -1
+// CHECK:           %[[VAL_3:.*]] = torch.constant.bool false
+// CHECK:           %[[VAL_4:.*]] = "tosa.const"() {value = dense<-1> : tensor<i64>} : () -> tensor<i64>
+// CHECK:           %[[VAL_5:.*]] = "tosa.const"() {value = dense<0> : tensor<i64>} : () -> tensor<i64>
+// CHECK:           %[[VAL_6:.*]] = "tosa.custom"(%[[VAL_1]], %[[VAL_4]], %[[VAL_5]]) {identifier = "torch.aten._softmax"} : (tensor<4x4xf32>, tensor<i64>, tensor<i64>) -> tensor<4x4xf32>
```
silvasean, line 10: A small nit here: let's have the "identifier" omit the "torch." but have the "config" added in https://reviews.llvm.org/D137133 be set to `torch_mlir`.

Svoch: Sure! I can change it, though the current LLVM version in Torch-MLIR does not have the commit you shared ([link to TOSA CustomOp arguments](https://github.com/llvm/llvm-project/blob/74fb770de9399d7258a8eda974c93610cfde698e/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td#L1798)).
I can update the `identifier` to omit the "torch." part for now and we can either wait for an LLVM uplift to bring in 'config' or keep having it marked as TODO. Either way makes sense to me.

Svoch: I updated the 'identifier' attribute of the TOSA CustomOp, also added _a_ `config` attribute. Please note that TOSA CustomOps in this version of LLVM still do not have the `config` attribute, so it's a monkey-patch right now. Added a TODO to update the builder call after the LLVM uplift.

python/torch_mlir/__init__.py
```diff
@@ -141,6 +141,7 @@ def compile(model: torch.nn.Module,
             use_tracing: bool = False,
             ignore_traced_shapes = False,
             backend_legal_ops: Optional[Sequence[str]] = None,
+            backend_custom_ops: Optional[Sequence[str]] = None,
```
Svoch, line 144: @silvasean - I understand that `BACKEND_LEGAL_OPS` construct above is used to identify which operations are legal for each of the backends, but it does not distinguish between the operations that should be converted into custom ops and those that only the decomposition step will be skipped.

One option I had was to extend the `BACKEND_LEGAL_OPS` to have two lists for each backend: "decompose" and "custom". Where for the ops in the "decompose" list only the decomposition step will be skipped and for those in "custom", on top of skipping the decomposition, the backend will convert them to CustomOps.

Another option was using another list (like this patch). I am not sure if this aligns with your original idea regarding marking backend-legal and backend-custom operations.

include/torch-mlir/Conversion/Passes.td
```diff
@@ -105,6 +105,59 @@ def ConvertTorchToLinalg : Pass<"convert-torch-to-linalg", "func::FuncOp"> {
...
+         to consume and convet the constant attributes (such as 'axis' for a
+         reduction operation), the TOSA CustomOp conversion will also match
+         and rewrite these attributes as TOSA ConstOps as well. Specifically:
+           - torch.constant.bool        -> tosa.ConstOp (of tensor type i64)
+           - torch.constant.int         -> tosa.ConstOp (of tensor type i64)
+           - torch.constant.float       -> tosa.ConstOp (of tensor type f32)
+           - torch.prim.ListConstruct   -> tosa.ConstOp (of tensor type i64)
+           - torch.constant.none        -> tosa.CustomOp (of tensor type i1)
+                                           The 'identifier' attribute of this
+                                           CustomOp is 'torch.constant.none'
```
eric-k256, line 138: If we know that these are constant, this is what the new implementation_attrs attribute to tosa.custom is meant to hold. It's a string, which means you would need to encode the attributes as a big string, but if we find that is unworkable, we could make it more flexible to contain an ArrayAttr. It's a bit less convenient on my side, as I'd have to come up with an encoding for a serialization pass we have, but it's doable if it makes the op more useful overall.

python/torch_mlir/__init__.py
silvasean: This is getting into power user territory. I don't think we should directly support this in `torch_mlir.compile`. We can add a `torch_mlir.tosa` module that builds this functionality on top of the base torch_mlir.compile API though (we could call the function `torch_mlir.tosa.compile`). It would mostly be a passthrough to torch_mlir.compile but would need to set the backend_legal_ops correctly and use `torch_mlir.compile` to compile to the backend contract, and then appropriately invoke the torch-backend-to-tosa-backend-pipeline.

test/Conversion/TorchToTosa/selective_decomposition.mlir
AmosLewis: replace with this should work.
`// CHECK:           %[[VAL_6:.*]] = "tosa.custom"(%[[VAL_1]], %[[VAL_4]], %[[VAL_5]]) {config = "torch_mlir", identifier = "aten._softmax"} : (tensor<4x4xf32>, tensor<i64>, tensor<i64>) -> tensor<4x4xf32>`

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,201 @@
...
+                                  Operation *backendLegalOp) const {
+    // Get the Torch Op to find the constant attributes attached to the
+    // backendLegalOp
+    Value torchInputOperand = inputOperand;
+    if (auto unrealizedCastOp = dyn_cast_or_null<UnrealizedConversionCastOp>(
+            inputOperand.getDefiningOp())) {
+      torchInputOperand = unrealizedCastOp.getInputs()[0];
+    }
+    // Handle the special case where input operand is an argument to the module
+    // function
```
AmosLewis, line 53: How about we add TODO here
// TODO supports more types: torch.constant.device, torch.constant.number, torch.constant.str, etc.

include/torch-mlir/Conversion/Passes.td
```diff
@@ -105,6 +105,59 @@ def ConvertTorchToLinalg : Pass<"convert-torch-to-linalg", "func::FuncOp"> {
...
+         to the consuming backend to deduce them based on the ATen operation
+         semantics.
+      4. Since TOSA conversion pattern of Torch operations are responsible
+         to consume and convet the constant attributes (such as 'axis' for a
+         reduction operation), the TOSA CustomOp conversion will also match
+         and rewrite these attributes as TOSA ConstOps as well. Specifically:
+           - torch.constant.bool        -> tosa.ConstOp (of tensor type i64)
+           - torch.constant.int         -> tosa.ConstOp (of tensor type i64)
+           - torch.constant.float       -> tosa.ConstOp (of tensor type f32)
+           - torch.prim.ListConstruct   -> tosa.ConstOp (of tensor type i64)
```
ramiro050, line 135: I think it would be clearer to use the types rather than the ops here. This will allow being explicit about which lists are supported and which are not. I'm assuming support for lists here is just for `!torch.list<int>`, right?

include/torch-mlir/Conversion/Passes.td
```diff
@@ -105,6 +105,59 @@ def ConvertTorchToLinalg : Pass<"convert-torch-to-linalg", "func::FuncOp"> {
...
+                                           The 'identifier' attribute of this
+                                           CustomOp is 'torch.constant.none'
+         All other Torch ATen operations will be lowered to TOSA by the Torch
+         to TOSA conversion pass after this one.
+        TODO: Extend the contract for other Torch constant operations such as:
+          - torch.constant.device
+          - torch.constant.number
+          - torch.constant.str
+      5. The order of the input operands of the backend-legal Torch operation
+         preserved and the TOSA CustomOp will have the same order.
```
ramiro050, line 146: nit: "The ~order of the~ input operands of the backend-legal Torch operation ~preserved~ and the TOSA CustomOp will have the same order."

include/torch-mlir/Conversion/Passes.td
```diff
@@ -105,6 +105,59 @@ def ConvertTorchToLinalg : Pass<"convert-torch-to-linalg", "func::FuncOp"> {
...
+        TODO: Update this pass to populate the 'config' attribute of the TOSA
+              CustomOp to establish a proper operand mapping scheme between
+              the backend-legal Torch operation and the TOSA CustomOp.
+              (See the commit at https://github.com/llvm/llvm-project/commit/d94ee70f4f01e4d9eec49e02eff57a5655618401)
+  }];
+  let options = [ListOption<
+      "customOps", "custom-ops", "std::string",
+      "List of operations considered backend-legal and should be converted "
+      "to converted to CustomOps in TOSA dialect.",
```
ramiro050, line 155: "List of operations considered backend-legal ~and~ **that** should be converted ~to converted~ to CustomOps in TOSA dialect."

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,201 @@
...
+        .Default([&](Operation *defaultOperand) { return inputOperand; });
+  }
+
+  LogicalResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const final {
+
+    for (auto customOpName : customOps) {
+      if (customOpName == op->getName().getStringRef()) {
+        SmallVector<Value, 4> customOpInputOperands;
```
ramiro050, line 114: nit: This vector can have an arbitrary size, so no need for the `4`. See: https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,201 @@
...
+          return noneCustomOp.getResult(0);
+        })
+        .Default([&](Operation *defaultOperand) { return inputOperand; });
+  }
+
+  LogicalResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const final {
+
+    for (auto customOpName : customOps) {
```
ramiro050, line 112: This pattern gets applied one op at a time, so there should not be a need for iterating through all of `customOps`. Instead, `customOps` should be stored as a set, and this pattern should begin with:

```c++
if (!customOps.contains(op->getName().getStringRef())
```

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,201 @@
...
+                op,
+                "failed to match the constant operand of the backend-legal Op");
+          }
+          customOpInputOperands.push_back(customOpInputOperand);
+        }
+        SmallVector<Type> customOpResultTypes;
+        auto convertTypesResult = getTypeConverter()->convertTypes(
+            op->getResultTypes(), customOpResultTypes);
+        if (convertTypesResult.failed())
+          return failure();
```
ramiro050, line 130: nit: use `notifyMatchFailure` with an error message

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,201 @@
...
+public:
+  ArrayRef<std::string> customOps;
+
+  ConvertBackendLegalAtenOp(TypeConverter &typeConverter, MLIRContext *context,
+                            ArrayRef<std::string> customOps)
+      : ConversionPattern(typeConverter, MatchAnyOpTypeTag{}, 0, context) {
+    this->customOps = customOps;
+  }
+
+  Value matchCustomOpOperandValue(Value inputOperand, PatternRewriter &rewriter,
```
ramiro050, line 43: I think it would be clearer to call this `convertOperandToTensor` and have the op return `nullptr` if the final value is not a tensor type. At the moment if an argument is an unsupported type, for example a string, there is no checks that would result in an error. All paths left as TODOs should result in errors with clear error messages

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,201 @@
...
+#include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
+#include "llvm/ADT/TypeSwitch.h"
+
+using namespace mlir;
+using namespace mlir::torch;
+using namespace mlir::torch::Torch;
+
+namespace {
+
+class ConvertBackendLegalAtenOp : public ConversionPattern {
```
ramiro050, line 33: nit: I think its clearer to call this pass `ConvertToCustomOp`

ramiro050 (changes requested):

include/torch-mlir/Conversion/Passes.td
Svoch: I see. I have a question on how to deduce wether an input is a constant attribute vs. some input that happens to be constant. Does it suffice for them to have the types listed here to consider them an attribute that needs to be listed in `implementation_attrs`? 
Also, do you have a schema in mind about the encoding?

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
Svoch: Done, thanks.

test/Conversion/TorchToTosa/selective_decomposition.mlir
Svoch: @eric-k256 @silvasean - Rebased and updated the CustomOp with `config` attribute set to "torch_mlir".
@AmosLewis - Done, thanks.

python/torch_mlir/__init__.py
Svoch: Done, added a module under `backends.tosa` and have it call `torch_mlir.compile` API. The `backend_custom_ops` argument is only present in the `torch_mlir.backends.tosa.compile` call.

include/torch-mlir/Conversion/Passes.td
Svoch: Done, updated to use types instead of operations. It is explicit that only list of integers are supported (i.e., `!torch.list<int>`)

include/torch-mlir/Conversion/Passes.td
Svoch: Done.

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
Svoch: Updated, thanks.

include/torch-mlir/Conversion/Passes.td
Svoch: Done, thanks.

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
Svoch: Done, updated `customOps` to be a `SetVector` instead of an `ArrayRef` to do look-ups and avoid looping on the custom operations instead.

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
Svoch: Done, thanks.

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
Svoch: So this is a little tricky, since any operand that is not one of the special constant cases (like `torch.constant.bool` or `torch.constant.int`) that should be matched and absorbed by the rewriter needs to pass through. The reason is, if an operand is a result of another operation then it should be handled by the Torch to TOSA conversion pass.

Now to handle these special operations, I think we need to have all of them specifically listed here in the switch-case. I updated the switch-case to include all of the operation types that _should be_ handled here. This way, any unhandled constant operand will return `nullptr` and the rewriter will `notifyMatchFailure`; otherwise it will pass through and the expectation is that Torch to TOSA conversion pass will need to convert it.

Does this look reasonable?

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
Svoch: Done, updated to `ConvertBackendLegalAtenOpToCustomOp` to be explicit about what this pattern converts to. I also kept the "BackendLegalAtenOp" to be consistent with the naming of the conversion patterns in Torch to TOSA in terms of what the pattern is converting from.
Let me know if you prefer the `ConvertToCustomOp` over this.

python/torch_mlir/backends/tosa/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+import torch
+import torch_mlir
+
+from ...compiler_utils import run_pipeline_with_repro_report
+
+def compile(model: torch.nn.Module,
+            example_args: Union[torch_mlir._example_arg, Sequence[torch_mlir._example_arg]],
+            use_tracing: bool = False,
+            ignore_traced_shapes = False,
+            backend_legal_ops: Optional[Sequence[str]] = None,
```
silvasean, line 17: remove backend_legal_ops argument. The whole point of this function is that it avoids the redundancy between the two.

python/torch_mlir/backends/tosa/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+    """Convert a PyTorch model to TOSA IR in MLIR.
+
+    Args:
+        model: The PyTorch model to convert.
+        example_args: A list of example arguments to use when inferring the
+            shapes of the arguments to `forward` method of the model.
+            A single tensor is treated as a list of a single tensor.
+            A TensorPlaceholder object is also allowed in the place of any
+            Tensor.
+        output_type: The kind of output to produce. See `OutputType` for more
```
silvasean, line 29: use **kwargs to forward these to torch_mlir.compile. No need to duplicate all the docs and signature here.

silvasean (changes requested):

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,220 @@
...
+              return nullptr;
+            })
+        .Default([&](Operation *defaultOperand) { return inputOperand; });
+  }
+
+  LogicalResult
+  matchAndRewrite(Operation *op, ArrayRef<Value> operands,
+                  ConversionPatternRewriter &rewriter) const final {
+
+    if (customOps.contains(op->getName().getStringRef())) {
```
ramiro050, line 135: nit: return early to avoid indentation

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+  ConvertBackendLegalAtenOpToCustomOp(TypeConverter &typeConverter,
+                                      MLIRContext *context,
+                                      ArrayRef<std::string> customOps)
+      : ConversionPattern(typeConverter, MatchAnyOpTypeTag{}, 0, context) {
+    this->customOps = SetVector<StringRef>(customOps.begin(), customOps.end());
+  }
+
+  Value convertOperandToTensor(Value inputOperand, PatternRewriter &rewriter,
+                               Operation *backendLegalOp) const {
```
ramiro050, line 46: nit: I find the name `backendLegalOp` a bit confusing, since backend could mean different things. I would just call it `op`, since it is the op we are currently matching on

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
ramiro050: > So this is a little tricky, since any operand that is not one of the special constant cases (like `torch.constant.bool` or `torch.constant.int`) that should be matched and absorbed by the rewriter needs to pass through. The reason is, if an operand is a result of another operation then it should be handled by the Torch to TOSA conversion pass.

I'm not sure I understand. Isn't the result type of every value returned by this function supposed to be a `RankedTensorType`?

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,220 @@
...
+    // backendLegalOp
+    Value torchInputOperand = inputOperand;
+    if (auto unrealizedCastOp = dyn_cast_or_null<UnrealizedConversionCastOp>(
+            inputOperand.getDefiningOp())) {
+      torchInputOperand = unrealizedCastOp.getInputs()[0];
+    }
+    // Handle the special case where input operand is an argument to the module
+    // function
+    if (!torchInputOperand.getDefiningOp())
+      return inputOperand;
```
ramiro050, line 57: This needs to check that the `inputOperand` is a tensor

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+    return TypeSwitch<Operation *, Value>(torchInputOperand.getDefiningOp())
```
ramiro050, line 59: nit: I feel that using a `TypeSwitch` in this particular case makes things more verbose and harder to read than just using `if` statements. For example, for all the `torch.constant`s, the type switch first dynamic casts to see which case to apply, then inside the cases the same dynamic cast is used inside the `matchPattern` to get the constant value. We can simply use the `matchPattern` to see if the input is a `torch.constant`. Moreover, all the `return nullptr` lines can be factored out if using `if` statements

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
```diff
@@ -0,0 +1,220 @@
...
+
+  Value convertOperandToTensor(Value inputOperand, PatternRewriter &rewriter,
+                               Operation *backendLegalOp) const {
+    // Get the Torch Op to find the constant attributes attached to the
+    // backendLegalOp
+    Value torchInputOperand = inputOperand;
+    if (auto unrealizedCastOp = dyn_cast_or_null<UnrealizedConversionCastOp>(
+            inputOperand.getDefiningOp())) {
+      torchInputOperand = unrealizedCastOp.getInputs()[0];
+    }
```
ramiro050, line 53: Is there a reason this is needed instead of using `op->getOperand(...)` to get `inputOperand`?

ramiro050 (changes requested):

lib/Conversion/TorchToTosa/TorchBackendLegalToTosaCustom.cpp
ramiro050: Ah I see why, you need both the torch dialect operand and also the converted operand for the default case. However, `UnrealizedConversionCastOp` is an implementation detail of how partial conversions are performed, so I don't think it's a good idea to depend on it. I would prefer passing the operand index as an argument, then doing `op->getOperand(operandIndex)` to get the torch dialect operand.

ZihengJiang:
We observed the same demand on the MHLO side. Looking forward to this PR being landed.

PR #1565, opened 2022-11-09
[torch_mlir.compile] Add support for multiple exported methods

### LINKED ISSUE #1557: Add support for multiple functions when exporting a model built with functorch
`torch-mlir` supports exporting multiple functions via `nn.Modules`.  When building a model with functorch (necessary for training), however, we can only export a single function.

silvasean:
For AoT deployments models often have multiple exported methods. This patch enables something like this:

```
class TwoMethodsModule(torch.nn.Module):
    def sin(self, x):
        return torch.ops.aten.sin(x)

    def cos(self, x):
        return torch.ops.aten.cos(x)

example_args = torch_mlir.ExampleArgs()
example_args.add_method("sin", torch.ones(2, 3))
example_args.add_method("cos", torch.ones(2, 4))
print(torch_mlir.compile(TwoMethodsModule(), example_args))
```

In the
[long-term](https://github.com/llvm/torch-mlir/blob/main/docs/long_term_roadmap.md#tools-for-advanced-aot-deployments) we will need to reconcile this with our story for stateful models and the backend contract being purely functional. For now, this provides some basic infra that seems harmless. Arguably, we could tighten up the backend contract even more to only allow a single compiled function which would prohibit this or require building out a layer above.

Fixes #1557

### DIFF
--- a/python/test/compile_api/tracing.py
+++ b/python/test/compile_api/tracing.py
@@ -59,14 +59,14 @@ class DictModule(torch.nn.Module):
 
 
 try:
-    # CHECK: Only Tensors, TensorPlaceholders, or a sequences of Tensors and TensorPlaceholders are supported as inputs.
+    # CHECK: Only Tensors, TensorPlaceholder's, or sequences of Tensors and TensorPlaceholder's are supported as example args for method inputs. Got '{'a': tensor(3.)}'
     torch_mlir.compile(DictModule(), {'a': torch.tensor(3.0)}, use_tracing=True)
 except Exception as e:
     print(e)
 
 
 try:
-    # CHECK: Only Tensors, TensorPlaceholders, or a sequences of Tensors and TensorPlaceholders are supported as inputs.
+    # CHECK: Only Tensors, TensorPlaceholder's, or sequences of Tensors and TensorPlaceholder's are supported as example args for method inputs. Got '{'a': tensor(3.)}'
     torch_mlir.compile(DictModule(), [{'a': torch.tensor(3.0)}], use_tracing=True)
 except Exception as e:
     print(e)
--- a/python/torch_mlir/__init__.py
+++ b/python/torch_mlir/__init__.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 # Also available under a BSD-style license. See LICENSE.
 
-from typing import Optional, Sequence, Union, List
+from typing import Optional, Sequence, Union, List, Dict, Tuple
 from enum import Enum
 
 import sys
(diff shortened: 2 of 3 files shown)


python/test/compile_api/tracing.py
```diff
@@ -59,14 +59,14 @@ def forward(self, x):
 
 
 try:
-    # CHECK: Only Tensors, TensorPlaceholders, or a sequences of Tensors and TensorPlaceholders are supported as inputs.
+    # CHECK: Only Tensors, TensorPlaceholder's, or sequences of Tensors and TensorPlaceholder's are supported as example args for method inputs. Got '{'a': tensor(3.)}'
```
dellis23, line 62: Nit: an apostrophe here implies it's possessive, not plural

python/torch_mlir/__init__.py
```diff
@@ -119,27 +118,138 @@ def like(tensor: torch.Tensor, dynamic_axes: List[int] = None):
...
+            self, for chaining.
+        """
+        assert method_name not in self._example_args
+        self._example_args[method_name] = ExampleArgs._canonicalize_args(
+            example_args)
+        return self
+
+    @staticmethod
+    def get(example_args: _example_args) -> "ExampleArgs":
+        """Gets an ExampleArgs from allowed way to specify one.
```
dellis23, line 158: I'm finding the phrasing of this docstring to be a bit confusing.

dellis23 (approved):
Cool!  Curious to see what this will look like for functorch-based functions.

python/test/compile_api/tracing.py
silvasean: I'm in the camp where making sure that the identifier boundaries are correct and greppable takes precedence over the grammatical rules of ordinary English prose. In Markdown I would typically write `TensorPlaceholder`s, but this is a plaintext message.

python/torch_mlir/__init__.py
silvasean: It was generated by copilot (Github AI autocomplete) lol. Fixed.

silvasean:
For the record, it would arguably be more Pythonic and consistent with torch.jit.trace_module to accept a dict here `{"forward": [my_tensor]}`, but we specifically have [a test case](https://github.com/llvm/torch-mlir/blob/9a73b9e6c7dbb0d4227cfb287f864bae70c35d0e/python/test/compile_api/tracing.py#L63) where a user passed such a dict and was expecting a different behavior. The more type-ful approach here with ExampleArgs is a bit cleaner and easier to get right anyway.

python/test/compile_api/tracing.py
dellis23: Makes sense. Should Tensors be changed to "Tensor's" too then?

silvasean: Oh duh, yes lol

silvasean: https://github.com/llvm/torch-mlir/pull/1569

PR #1568, opened 2022-11-09
Move single-tensor-tuple-return test to mlir unit test.

dellis23:


silvasean:
Also cover the error "Functions directly imported from Python should only ever return one item. Multiple return values are returned as a tuple"

dellis23:
> Also cover the error "Functions directly imported from Python should only ever return one item. Multiple return values are returned as a tuple"

Done

PR #1572, opened 2022-11-11
[MLIR][TORCH] Add E2E support for `aten.randint.low` op

vivekkhandelwal1:


test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -1047,3 +1047,40 @@ func.func @torch.aten.mse_loss$sum_reduction(%arg0: !torch.vtensor<[?,?],f32>, %
...
+// CHECK:         %[[VAL_1:.*]] = torch.aten.broadcast_to %[[VAL_0]], %[[SIZE]] : !torch.vtensor<[],si64>, !torch.list<int> -> !torch.vtensor<[1024,1024],si64>
+// CHECK:         %[[INT7:.*]] = torch.constant.int 7
+// CHECK:         %[[VAL_2:.*]] = torch.aten.to.dtype %[[VAL_1]], %[[INT7]], %[[FALSE]], %[[FALSE]], %[[NONE_0]] : !torch.vtensor<[1024,1024],si64>, !torch.int, !torch.bool, !torch.bool, !torch.none -> !torch.vtensor<[1024,1024],f64>
+// CHECK:         %[[UNIFORM:.*]] = torch.aten.uniform %[[VAL_2]], %[[FLOAT1]], %[[FLOAT1000]], %[[NONE_0]] : !torch.vtensor<[1024,1024],f64>, !torch.float, !torch.float, !torch.none -> !torch.vtensor<[1024,1024],f64>
+// CHECK:         %[[INT4_4:.*]] = torch.constant.int 4
+// CHECK:         %[[RES:.*]] = torch.aten.to.dtype %[[UNIFORM]], %[[INT4_4]], %[[FALSE]], %[[FALSE]], %[[NONE_0]] : !torch.vtensor<[1024,1024],f64>, !torch.int, !torch.bool, !torch.bool, !torch.none -> !torch.vtensor<[1024,1024],si64>
+// CHECK:         return %[[RES]] : !torch.vtensor<[1024,1024],si64>
+func.func @torch.aten.randint.low() -> !torch.vtensor<[1024,1024],si64> {
```
ramiro050, line 1077: No need for this test. This falls under:

> DON'T use a unit test if your lowering pattern could be described as a trivial "macro expansion" of one op into another op or set of ops.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2896,6 +2896,60 @@ class DecomposeAtenMseLossOp : public OpRewritePattern<AtenMseLossOp> {
...
+
+    BaseTensorType floatResultType =
+        resultTensorType
+            .getWithSizesAndDtype(resultTensorType.getSizes(),
+                                  rewriter.getF64Type())
+            .cast<BaseTensorType>();
+    emptyTensor = rewriter.create<AtenToDtypeOp>(
+        loc, floatResultType, emptyTensor,
+        getDtypeIntValueForType(rewriter, loc, rewriter.getF64Type()), cstFalse,
+        cstFalse, none);
```
ramiro050, line 2938: I don't think this `AtenToDtypeOp` is needed. You should be able to get the right type for the `emptyTensor` by passing the right result type and `dtype` arguments to `AtenFullOp`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2896,6 +2896,60 @@ class DecomposeAtenMseLossOp : public OpRewritePattern<AtenMseLossOp> {
...
+
+    Value none = rewriter.create<ConstantNoneOp>(loc);
+    Value cstFalse = rewriter.create<ConstantBoolOp>(loc, false);
+    Value zero =
+        rewriter.create<ConstantFloatOp>(loc, rewriter.getF64FloatAttr(0.0));
+    Value low = rewriter.create<Torch::ConstantFloatOp>(
+        loc, rewriter.getF64FloatAttr((double)cstLow));
+    Value high = rewriter.create<Torch::ConstantFloatOp>(
+        loc, rewriter.getF64FloatAttr((double)cstHigh));
+    Value emptyTensor = rewriter.create<AtenFullOp>(
```
ramiro050, line 2926: I think its better to create an `AtenEmptyMemoryFormatOp`. The entire tensor contents will be overwritten by `uniform` so no need to fill it with a value

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2896,6 +2896,60 @@ class DecomposeAtenMseLossOp : public OpRewritePattern<AtenMseLossOp> {
...
+        cstFalse, none);
+    Value result =
+        rewriter.create<AtenUniformOp>(loc, floatResultType, emptyTensor,
+                                       /*from=*/low,
+                                       /*to=*/high,
+                                       /*generator=*/none);
+    rewriter.replaceOpWithNewOp<AtenToDtypeOp>(
+        op, resultType, result,
+        getDtypeIntValueForType(rewriter, loc, resultTensorType.getDtype()),
+        cstFalse, cstFalse, none);
```
ramiro050, line 2947: can you add comments for the `false, false, none` arguments?

ramiro050 (changes requested):
Nice! This is a much simpler approach than what I had in mind

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2896,6 +2896,55 @@ class DecomposeAtenMseLossOp : public OpRewritePattern<AtenMseLossOp> {
...
+    Value high = rewriter.create<Torch::ConstantFloatOp>(
+        loc, rewriter.getF64FloatAttr((double)cstHigh));
+
+    BaseTensorType floatResultType =
+        resultTensorType
+            .getWithSizesAndDtype(resultTensorType.getSizes(),
+                                  rewriter.getF64Type())
+            .cast<BaseTensorType>();
+    Value emptyTensor = rewriter.create<AtenEmptyMemoryFormatOp>(
+        loc, floatResultType, op.size(), /*dtype=*/none, /*layout=*/none,
```
ramiro050, line 2931: Either the `dtype` argument needs to be specified (default dtype is f32 but result type is f64), or the result type should be changed to `f32`.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2896,6 +2896,55 @@ class DecomposeAtenMseLossOp : public OpRewritePattern<AtenMseLossOp> {
...
+        /*device=*/none, /*pin_memory=*/none, /*memory_format=*/none);
```
ramiro050, line 2932: You should be passing `layout, device, pin_memory` from the `AtenRandIntLowOp`

python/torch_mlir_e2e_test/test_suite/rng.py
```diff
@@ -276,3 +276,43 @@ def forward(self, x):
...
+class RandIntLowDtypeModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        a = torch.ops.aten.randint(low=1, high=1000, size=[128, 256, 512], dtype=torch.float32)
```
ramiro050, line 311: Why is it needed to turn the tensor to `float32`?

PR #1576, opened 2022-11-11
Fix non-`const Node*` usages

antoniojkim:
https://github.com/pytorch/pytorch/pull/88773 changed the LTC backend interface which breaks our implementation as well.

Fixing the non-`const Node*` usages to match the latest interface


ke1337 (approved):

powderluv (approved):

antoniojkim:
This is already fixed now in https://github.com/llvm/torch-mlir/pull/1580

PR #1578, opened 2022-11-11
[MLIR][TORCH] Fix aten.unsqueeze op

AmosLewis:
The range of the unsqueeze dim is: [-input.dim() - 1, input.dim() + 1), the bug forgets to add 1. 
`https://pytorch.org/docs/stable/generated/torch.unsqueeze.html`


eric-k256 (approved):
Looks good to me.

PR #1581, opened 2022-11-14
[MLIR][TORCH] Add E2E support for `aten.randn` op

vivekkhandelwal1:


lib/Conversion/TorchToLinalg/Random.cpp
```diff
@@ -154,6 +160,104 @@ class ConvertAtenUniformOp : public OpConversionPattern<AtenUniformOp> {
...
+                  Value uOne = getUniformRandomNumber(b, loc, temp, min, max,
+                                                      resultElemTy);
+                  for (int i = 0; i < resultRank; i++) {
+                    Value index = b.create<linalg::IndexOp>(loc, i);
+                    temp = getNextTemp(b, loc, index, temp, multiplier,
+                                       incrementStep);
+                  }
+                  Value uTwo = getUniformRandomNumber(b, loc, temp, min, max,
+                                                      resultElemTy);
+                  // res = sqrt(-2ln(u1)) * cos(2*pie*u2)
```
ramiro050, line 241: I think it would be simpler to do a decomposition. Something like this should work:

```python
def randn_decomp(size):
    u1 = torch.empty(size).uniform_()
    u2 = torch.empty(size).uniform_()
    return torch.sqrt(-2 * torch.log(u1)) * torch.cos(2 * 3.14159 * u2)
```

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/Random.cpp
vivekkhandelwal1: That's not working. I have tried this, but it's giving the wrong result.

vivekkhandelwal1: This is the code that I added:
```
namespace {
class DecomposeAtenRandnOp : public OpRewritePattern<AtenRandnOp> {
public:
  using OpRewritePattern::OpRewritePattern;
  LogicalResult matchAndRewrite(AtenRandnOp op,
    Location loc = op.getLoc();
    Type resultType = op.getType();

    Value none = rewriter.create<ConstantNoneOp>(loc);
    Value low = rewriter.create<Torch::ConstantFloatOp>(
        loc, rewriter.getF64FloatAttr((double)0.0));
    Value high = rewriter.create<Torch::ConstantFloatOp>(
        loc, rewriter.getF64FloatAttr((double)1.0));
    Value cstMinusTwo = rewriter.create<Torch::ConstantFloatOp>(
        loc, rewriter.getF64FloatAttr((double)-2.0));
    Value cstTwoPie = rewriter.create<Torch::ConstantFloatOp>(
        loc, rewriter.getF64FloatAttr((double)(2.0 * 3.14159)));

    Value emptyTensorA = rewriter.create<AtenEmptyMemoryFormatOp>(
        loc, resultType, op.size(), /*dtype=*/none, /*layout=*/op.layout(),
        /*device=*/op.device(), /*pin_memory=*/op.pin_memory(),
        /*memory_format=*/none);
    Value emptyTensorB = rewriter.create<AtenEmptyMemoryFormatOp>(
        loc, resultType, op.size(), /*dtype=*/none, /*layout=*/op.layout(),
        /*device=*/op.device(), /*pin_memory=*/op.pin_memory(),
        /*memory_format=*/none);

    Value uOne = rewriter.create<AtenUniformOp>(loc, resultType, emptyTensorA,
                                                /*from=*/low,
                                                /*to=*/high,
                                                /*generator=*/none);

    Value uTwo = rewriter.create<AtenUniformOp>(loc, resultType, emptyTensorB,
                                                /*from=*/low,
                                                /*to=*/high,
                                                /*generator=*/none);

    Value logUOne = rewriter.create<AtenLogOp>(loc, resultType, uOne);
    Value minusTwoLogUOne =
        rewriter.create<AtenMulScalarOp>(loc, resultType, logUOne, cstMinusTwo);
    Value r = rewriter.create<AtenSqrtOp>(loc, resultType, minusTwoLogUOne);
    Value theta =
        rewriter.create<AtenMulScalarOp>(loc, resultType, uTwo, cstTwoPie);
    Value cosTheta = rewriter.create<AtenCosOp>(loc, resultType, theta);
    rewriter.replaceOpWithNewOp<AtenMulTensorOp>(op, op.getType(), r, cosTheta);
    return success();
  }
};
} // namespace
```

ramiro050: Yeah, this is strange. I've opened an issue to track this: https://github.com/llvm/torch-mlir/issues/1608

lib/Conversion/TorchToLinalg/Random.cpp
```diff
@@ -95,14 +120,6 @@ class ConvertAtenUniformOp : public OpConversionPattern<AtenUniformOp> {
         loc, rewriter.getI64IntegerAttr(6364136223846793005));
     Value incrementStep = rewriter.create<arith::ConstantOp>(
         loc, rewriter.getI64IntegerAttr(1442695040888963407));
```
ramiro050, line 122: I would move 

https://github.com/llvm/torch-mlir/blob/10cf74cd381fe91c057fb71b75308d974b9d13e1/lib/Conversion/TorchToLinalg/Random.cpp#L111-L122

to `getNextTemp`, since it's all related. This also avoid repeating the multiplier and increment values multiple times in this file

lib/Conversion/TorchToLinalg/Random.cpp
```diff
@@ -154,6 +160,104 @@ class ConvertAtenUniformOp : public OpConversionPattern<AtenUniformOp> {
...
+            .create<linalg::GenericOp>(
+                loc, initTensor.getType(), /*inputs=*/ValueRange{},
+                /*outputs=*/initTensor, indexingMaps, iteratorTypes,
+                [&](OpBuilder &b, Location loc, ValueRange args) {
+                  Value temp = initialSeed;
+                  for (int i = 0; i < resultRank; i++) {
+                    Value index = b.create<linalg::IndexOp>(loc, i);
+                    temp = getNextTemp(b, loc, index, temp, multiplier,
+                                       incrementStep);
+                  }
```
ramiro050, line 231: I would include this `for` loop in `getNextTemp`. It makes sense for the entire LCG algorithm to be in a single function.

lib/Conversion/TorchToLinalg/Random.cpp
```diff
@@ -154,6 +160,104 @@ class ConvertAtenUniformOp : public OpConversionPattern<AtenUniformOp> {
...
+      return failure();
+    Location loc = op.getLoc();
+    RankedTensorType resultType = getTypeConverter()
+                                      ->convertType(op->getResult(0).getType())
+                                      .cast<RankedTensorType>();
+    Type resultElemTy = resultType.getElementType();
+
+    // The pin_memory should be either `False` or `none`.
+    bool pinMemory;
+    if (!op.pin_memory().getType().template isa<Torch::NoneType>() &&
```
ramiro050, line 182: You should also check for `layout` and `device`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3135,6 +3135,25 @@ class DecomposeAtenVarMeanCorrectionOp
 };
 } // namespace
 
+namespace {
+// Decompose `aten.randn.generator` op into `aten.randn` op.
+class DecomposeAtenRandnGeneratorOp
```
ramiro050, line 3140: I think we should flip this. `AtenRandnOp` is a special case of `AtenRandnGeneratorOp`, so it makes more sense to decompose the special case to the general case. This will simplify things if we ever add support for generators

ramiro050 (changes requested):
I think we can move ahead with this implementation while the issue with the RNG algorithm is sorted out, since in this case not using a decomposition only affects simplicity of code and not performance.

vivekkhandelwal1:
@ramiro050 Can we merge this now?

ramiro050:
> @ramiro050 Can we merge this now?

After talking to @silvasean, we think it would be better to wait until the RNG issue is resolved and then go with the decomposition implementation to avoid adding code that will be deleted soon after. The decomposition also serves as a good test of the RNG algorithm. Does this work for you?

Apologies for the confusion.

ramiro050:
Can you test the decomposition with this patch: https://github.com/llvm/torch-mlir/pull/1633?

vivekkhandelwal1:
> Can you test the decomposition with this patch: #1633?

Hi @ramiro050, I have added the decomposition. Can you review this PR now?

vivekkhandelwal1:
Hi @ramiro050, just a reminder for you to review this PR.

PR #1585, opened 2022-11-14
Enable torch-mlir LIT tests in Bazel

sjain-stanford:
Adds support to run `.mlir` LIT tests in bazel. 

```
bazel test @torch-mlir//test/...
```

Follow-on PR will contain these updates:
- Add tests to GHA CI workflow
- Add `.py` LIT tests to bazel


sjain-stanford:
> Are other backends besides RefBackend intended to be supported later ?

We're mirroring / enabling the `.mlir` LIT tests that are supported with the regular CMake flow. It seems right now we only have `Conversion`, `Dialect` and `RefBackend` directories in `test/`, and (I might be wrong but) other backends are checked under `Conversion/TorchTo***`, which we've enabled as well.

PR #1590, opened 2022-11-15
[MLIR][TORCH] Fix aten.upsample_nearest2d op

Abhishek-Varma:
-- `aten.upsample_nearest2d.vec` op is not present owing to [upsample_*.vec PR](https://github.com/pytorch/pytorch/pull/85638).
-- So this commit adds a lowering on aten.upsample_nearest2d.
-- It also fixes `aten.upsample_nearest2d` -> LINALG lowering.


Abhishek-Varma:
In the CI failure for `ubuntu-x86_64, in-tree, ON` I see :- 
```
****** Unexpectedly Passed tests - 2 tests
16249    XPASS - "UpSampleNearest2dDynamicSize_basic"
16250    XPASS - "UpSampleNearest2dStaticSize_basic"
```

As per [LLVM lit](https://llvm.org/docs/CommandGuide/lit.html) :- 
```
XPASS:
    The test succeeded, but it was expected to fail. This is used for tests which were specified as expected
    to fail, but are now succeeding (generally because the feature they test was broken and has been fixed).
```

But the implementation in this PR really doesn't add any new feature - it tries to account for the update in signature. Let me know where am I going wrong in my understanding of the same.

Am I supposed to delete these test cases? Doesn't seem like so because I see `ubuntu-x86_64, out-of-tree, OFF` build passing so the `in-tree` shouldn't be any different, right?

Can someone help me here ? @vivekkhandelwal1 @ramiro050

python/torch_mlir_e2e_test/test_suite/conv.py
```diff
@@ -737,48 +759,10 @@ def __init__(self):
...
-    @annotate_args([None, ([-1, -1, -1, -1], torch.float32, True)])
-    def forward(self, inputVec):
-        return torch._C._nn.upsample_nearest2d(inputVec,
-                                               output_size=None,
-                                               scale_factors=[2.3, 4.7])
-
-
-@register_test_case(module_factory=lambda: UpSampleNearest2dDiffFactor())
-def UpSampleNearest2dDynamicFactor_basic(module, tu: TestUtils):
-    module.forward(tu.rand(2, 3, 2, 2))
```
vivekkhandelwal1, line 763: What's the reason to remove this test?

lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp
```diff
@@ -819,48 +819,50 @@ class ConvertAtenUpsampleNearest2dVecOp
...
-        scaleFactorsInt.push_back(intVal);
-      }
-
-      for (unsigned i = 0; i < scaleFactorsFloatValues.size(); i++)
-        dims[hDimOffset + i] = getScaledDims(
-            rewriter, loc, dims[hDimOffset + i], scaleFactorsFloatValues[i]);
-
+      scaleFactorsInt.push_back(intVal);
+      dims[hDimOffset + 0] = getScaledDims(
+            rewriter, loc, dims[hDimOffset + 0], adaptor.scales_h());
```
vivekkhandelwal1, line 839: `hDimOffset + 0` -> `hDimOffset`

vivekkhandelwal1 (changes requested):

python/torch_mlir_e2e_test/test_suite/conv.py
Abhishek-Varma: `output_size` is not OPTIONAL anymore so these tests aren't really valid anymore.
Let me know your take here.

vivekkhandelwal1: Okay, I'm looking for a test with scaling factors similar to this.

Abhishek-Varma: I've addressed this - you may review again.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -1213,8 +1213,8 @@ def aten〇linalg_vector_norm(self: List[int], ord: float = 2, dim: Optional[Lis
 def aten〇frobenius_norm〇dim(self: List[int], dim: List[int], keepdim: bool = False) -> List[int]:
     return upstream_shape_functions.sum_mean_dim(self, dim, keepdim, 0)
 
-def aten〇upsample_nearest2d〇vec(input: List[int], output_size: Optional[List[int]], scale_factors: Optional[List[float]]) -> List[int]:
-    return upstream_shape_functions.upsample_nearest2d(input, output_size, scale_factors)
```
ramiro050, line 1217: Can you upstream the fixes to this function? Here is an example PR: https://github.com/pytorch/pytorch/pull/76889.

Abhishek-Varma: Sure @ramiro050 I'm working on it - meanwhile can we aim to get this PR merged in ?

python/torch_mlir_e2e_test/test_suite/conv.py
Abhishek-Varma: Have added `UpSampleNearest2dDynamicFactor_basic` to `EAGER_MODE_XFAIL_SET`.

ramiro050 (approved):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
ramiro050: Sure! Do update this once your Pytorch PR lands

PR #1594, opened 2022-11-16
[custom op] Generalize shape library logic to work with dtypes

ramiro050:
This commit generalizes the shape library logic, so that dtype rules
for ops can also be expressed using the same mechanism. In other
words, each op can now have a shape function and a dtype function
specified in Python that is imported during lowering to calculate the
shapes and dtypes throught a program. For more information about how
to specify a dtype function, see the updated
`docs/adding_a_shape_and_dtype_function.md`.

For those not familiar with how the shape library works, the file
`docs/calculations_lib.md` provides an overview.

To make the reviewing a bit easier, I suggest the following review
order:

1. Get familiar with the overall architecture by reading
`docs/calculations_lib.md`
2. New op declarations
   - `include/torch-mlir/Dialect/Torch/IR/TorchOps.td`
   - `lib/Dialect/Torch/IR/TorchOps.cpp`
3. New passes
   - `include/torch-mlir/Dialect/Torch/Transforms/Passes.td`
   - `include/torch-mlir/Dialect/Torch/Transforms/Passes.h`
   - `lib/Dialect/Torch/Transforms/Passes.cpp`
   - `lib/Dialect/Torch/Transforms/ReifyDtypeCalculations.cpp`
   - `lib/Dialect/Torch/Transforms/SimplifyDtypeCalculations.cpp`
   - `lib/Dialect/Torch/Transforms/ReifyCalculationsUtils.cpp`
   - `lib/Dialect/Torch/Transforms/ReifyCalculationsUtils.h`
   - `lib/Dialect/Torch/Transforms/SimplifyCalculationsUtils.cpp`
   - `lib/Dialect/Torch/Transforms/SimplifyCalculationsUtils.h`

   The `*Utils.*` files include logic that is shared by dtype and
   shape passes.
4. Tests
   - `test/Dialect/Torch/ops.mlir`
   - `test/Dialect/Torch/reify-dtype-calculations.mlir`
   - `test/Dialect/Torch/simplify-dtype-calculations.mlir`
5. Introduce `torch_mlir_promote_dtypes`
   - `python/torch_mlir/dialects/torch/importer/jit_ir/csrc/node_importer.cpp`
   - `python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/library_generator.py`
6. Simple refactoring/generalizing
   - `include/torch-mlir/Dialect/Torch/Utils/Utils.h`
   - `lib/Dialect/Torch/Utils/Utils.cpp`
   - `lib/Dialect/Torch/Transforms/DropCalculations.cpp`
   - `lib/Dialect/Torch/Transforms/RefineTypes.cpp`
   - `lib/Dialect/Torch/Transforms/ReifyShapeCalculations.cpp`
   - `lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp`
   - `python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/registry.py`
   - `python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/testing_framework.py`
   - `python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/calculations_lib_gen.py`
   - `lib/Dialect/Torch/Transforms/CalculationsLibrary.cpp`
7. The rest of the files include minor changes
   - Replace `shape` with `calculations`
   - Replace `m_TorchConstantIntList` with
   `m_TorchListOfConstantInts` (needed to avoid ambiguity with new
   pattern `m_TorchListOfOptionalConstantInts`)


.github/workflows/RollPyTorch.yml
```diff
@@ -64,14 +64,14 @@ jobs:
         echo "PT_RELEASE=${PT_RELEASE}" >> ${GITHUB_ENV}
         echo "PT_HASH_CHANGED=${PT_HASH_CHANGED}" >> ${GITHUB_ENV}
 
-    - name: Build and test (in-tree), also update ODS and shape library
+    - name: Build and test (in-tree), also update ODS and calculations library
```
ramiro050, line 67: I'm open to suggestions on a better name for the library. I thought of `refinement library`, but I wouldn't consider decompositions a refinement.

test/Dialect/Torch/simplify-dtype-calculations.mlir
```diff
@@ -0,0 +1,282 @@
...
+// CHECK:             {{.*}} = torch.aten.add.Tensor {{.*}} -> !torch.vtensor<[1],f64>
+func.func @promote_dtypes$tensor_tensor_same_category_different_width(%arg0: !torch.vtensor<[1],f32>, %arg1: !torch.vtensor<[1],f64>, %arg2: !torch.float) {
+  %int7 = torch.constant.int 7
+  %int6 = torch.constant.int 6
+  %int1 = torch.constant.int 1
+  %0 = torch.dtype.calculate {
+    %1 = torch.aten.add.Tensor %arg0, %arg1, %arg2 : !torch.vtensor<[1],f32>, !torch.vtensor<[1],f64>, !torch.float -> !torch.vtensor<[1],unk>
+    torch.dtype.calculate.yield %1 : !torch.vtensor<[1],unk>
+  } dtypes {
+    %1 = torch.prim.ListConstruct %int1, %int1 : (!torch.int, !torch.int) -> !torch.list<optional<int>>
```
silvasean, line 37: can you explain what %int1/etc. represent? Even better, get the dtype directly off of a tensor with prim::dtype.

silvasean (changes requested):
First round of comments. Overall looks good. Will probably have a few more nits after these are addressed but not much more than that.

lib/Conversion/TorchToMhlo/Pooling.cpp
```diff
@@ -93,18 +93,19 @@ LogicalResult ConvertAtenOp<AtenMaxPool2dOp>::matchAndRewrite(
...
+                     m_TorchListOfConstantInts(kernelSize)))) {
     return rewriter.notifyMatchFailure(
         op, "non-const int kernel size unsupported!");
   }
-  if (!(matchPattern(op.stride(), m_TorchConstantIntList(stride)))) {
+  if (!(matchPattern(op.stride(), m_TorchListOfConstantInts(stride)))) {
     return rewriter.notifyMatchFailure(op, "non-const int stride unsupported!");
   }
-  if (!(matchPattern(op.padding(), m_TorchConstantIntList(padding)))) {
+  if (!(matchPattern(op.padding(), m_TorchListOfConstantInts(padding)))) {
```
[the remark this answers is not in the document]
ramiro050: Yeah, I actually kept my entire commit history in another branch, so extracting these changes was pretty easy. PR is here: https://github.com/llvm/torch-mlir/pull/1601

.github/workflows/RollPyTorch.yml
ramiro050: Yeah, that's a much better name

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/testing_framework.py
```diff
@@ -0,0 +1,313 @@
...
+    args_with_split_rank_and_dtype = []
+    for argument in arguments:
+        if contains_tensor(argument):
+            args_with_split_rank_and_dtype += [(argument, "rank"),
+                                               (argument, "dtype")]
+        else:
+            args_with_split_rank_and_dtype.append((argument, None))
+
+    result = []
+    for arg, arg_type in args_with_split_rank_and_dtype:
```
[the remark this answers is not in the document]
ramiro050: No particular reason :laughing: I was mirroring what I wrote in C++, however, I just realized this can be done in a single loop in C++ too.

lib/Dialect/Torch/Transforms/ReifyDtypeCalculations.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/Transforms/Passes.h"
+
+using namespace mlir;
+using namespace mlir::torch;
+using namespace mlir::torch::Torch;
+
+// Used to keep track of whether a library function argument corresponds to the
+// rank or dtype of a tensor, or to some other value.
+enum class FunctionArgType { rank, dtype, other };
```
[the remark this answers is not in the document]
ramiro050: I'm no longer using this enum

ramiro050:
@silvasean, I've addressed the comments in separate commits. Also, I had to make this change to make all tests pass: https://github.com/llvm/torch-mlir/pull/1594/commits/a24b7c6ee8a4863de38859865da701276d9fe8e1. The commit message (pasted below) has the explanation:

> This commit adds back the ops that use the new dtype refinement
pass. This is needed to avoid the catch-22 that results from ops in
the new dtype refinement pass needing dtype information from ops in
the RefineTypes pass, and ops in the RefineTypes pass needing dtype
information from ops in the new dtype refinement pass.
>
> The reason this catch-22 problem is not handled by the iterative
application of passes when lowering to the backend contract is because
the DecomposeComplexOps pass is not currently designed to run more
than once, since every op with a decomposition gets marked illegal
after the pass is done. Marking ops as legal results in no
decomposition patterns being applied because the graph is already in a
legal state.
>
> Adding back the ops to RefineTypes seems like the simplest solution to
this problem while the rest of the ops get relocated to use the new
dtype refinement pass.

silvasean:
> @silvasean, I've addressed the comments in separate commits. Also, I had to make this change to make all tests pass: [a24b7c6](https://github.com/llvm/torch-mlir/commit/a24b7c6ee8a4863de38859865da701276d9fe8e1). The commit message (pasted below) has the explanation:
> 
> > This commit adds back the ops that use the new dtype refinement
> > pass. This is needed to avoid the catch-22 that results from ops in
> > the new dtype refinement pass needing dtype information from ops in
> > the RefineTypes pass, and ops in the RefineTypes pass needing dtype
> > information from ops in the new dtype refinement pass.
> > The reason this catch-22 problem is not handled by the iterative
> > application of passes when lowering to the backend contract is because
> > the DecomposeComplexOps pass is not currently designed to run more
> > than once, since every op with a decomposition gets marked illegal
> > after the pass is done. Marking ops as legal results in no
> > decomposition patterns being applied because the graph is already in a
> > legal state.
> > Adding back the ops to RefineTypes seems like the simplest solution to
> > this problem while the rest of the ops get relocated to use the new
> > dtype refinement pass.

I think that before we land this we should fix this so that the iterative lowering behaves as intended here (perhaps as easy as switching DecomposeComplexOps to use the greedy rewriter? and then perhaps move the final legality check into satisfiesBackendContract somehow?). That seems like an independently useful improvement as well as the right thing to do here.

I would generally bias pretty heavily against keeping the old path in RefineTypes alive unless absolutely necessary -- when we migrate an op, we should delete the old support. That guarantees monotonic progress towards the new system by making the new system load-bearing. If the old system still works for these ops, it is easy to add support for ops in the new system, but the e2e test is still ends up using the old system for some reason and so weren't even testing the new code -- thus we accumulate buggy, untested code in the new system.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/library_generator.py
```diff
@@ -0,0 +1,174 @@
...
+    # `torch.tensor(scalar).to(type(scalar)).dtype` does not work because
+    # `torch.tensor` needs to know the type of the input at compile time
+    # and there is no `jit.script` support for Python's `type`.
+    #
+    # TODO: A better way to handle this would be to add support for
+    # `isinstance` in torch-mlir, which requires adding a new torch dialect
+    # op.
+    return torch.ops.prim.NumToTensor(scalar).dtype
+
+@torch.jit.ignore
```
silvasean, line 36: can you comment on why torch.jit.ignore is needed here?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/library_generator.py
```diff
@@ -0,0 +1,174 @@
...
+    # jit.scripted, all math ops that work on scalars becomes
+    # (Scalar, Scalar) -> (Scalar) ops. So to get the right result
+    # dtype, we use the tensor-tensor promotion rules.
+    return _promote_tensor_tensor(0, lhs_dtype, 0, rhs_dtype)
+
+def promote_dtypes(ranks: List[Optional[int]],
+                   dtypes: List[torch.dtype]) -> torch.dtype:
+    return __torch_mlir_internal_promote_dtypes(ranks, dtypes)
+
+def __torch_mlir_internal_promote_dtypes(ranks: List[Optional[int]],
```
silvasean, line 74: comment on why this function is defined here but also imported magically?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/library_generator.py
```diff
@@ -0,0 +1,174 @@
...
+            lhs_dtype = _promote_scalar_tensor(
+                rhs_dtype, lhs_optional_rank, lhs_dtype)
+        elif lhs_optional_rank is not None and rhs_optional_rank is not None:
+            lhs_dtype = _promote_tensor_tensor(
+                lhs_optional_rank, lhs_dtype, rhs_optional_rank, rhs_dtype)
+            lhs_optional_rank = max(lhs_optional_rank, rhs_optional_rank)
+    return lhs_dtype
+
+def not_present_in_registry(f):
+    """Decorator for shape functions not present in the shape registry.
```
silvasean, line 96: nit: shape -> abstract interpretation

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/library_generator.py
```diff
@@ -0,0 +1,174 @@
...
+def _verify_signature_matches_registry(f, registry: Registry):
+    source = inspect.getsource(f)
+    signature = None
+    for line in source.splitlines():
+        if line.startswith("def "):
+            signature = line
+            break
+    assert signature is not None, f"Could not find signature for {f.__name__}"
+    assert "〡" in signature, f"Malformed signature {signature}. Signature missing the character `〡`"
+    f_name, overload = f.__name__.split("〡")
```
silvasean, line 122: overload has another meaning (it is the 3rd atom in f_name). So perhaps function_kind could be better

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/registry.py
```diff
@@ -142,54 +182,92 @@ def uppercase_first_letter(s):
...
-                default = f" = {default_debug}"
-            parameter_name = arg["name"]
-            if parameter_name == "from":
-                parameter_name = "from_" # Avoid using a Python keyword.
-            parameter_decls.append(f"{parameter_name}: {pytype}{default}")
-        ret_decls = []
-        for ret in self.returns:
-            pytype = _pytype_to_shape_fn_pytype(ret["pytype"])
-            ret_decls.append(f"{pytype}")
+        def_name += f"〡{function_overload}"
```
silvasean, line 192: nit: function kind

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/testing_framework.py
```diff
@@ -0,0 +1,303 @@
...
+#     ErrorInvocation(TensorOfShape(2, 3, 4), dim=-4), # `dim` out of bounds.
+#     ErrorInvocation(TensorOfShape(2, 3, 4), dim=3), # `dim` out of bounds.
+# ])
+# ```
+# Each `Invocation` takes a list of args/kwargs which will be passed to both the
+# shape function and the real op and the results compared.
+# We expect both the successful and error cases to be tested.
+#
+# The typical iteration flow is to add invocations to the list and then re-run
+# `build_tools/update_shape_lib.sh` to re-run the tests.
```
silvasean, line 43: nti: update_shape_lib needs to be renamed

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/testing_framework.py
```diff
@@ -0,0 +1,303 @@
...
+class Invocation:
+    """Representation of a single op invocation (i.e. list of args to the op).
+
+    This class is used to represent a single invocation of an op in a way that
+    we can use to both invoke the abstract interpretation function and invoke
+    the actual op, which have slightly different signatures.
+
+    Specifically, this class has special knowledge of `TensorOfShape` and
+    translates it appropriately to either a tensor (for the real op) or a
+    `List[int]` (for the shape function).
```
silvasean, line 127: also mention the dtype function

test/Dialect/Torch/drop-abstract-interp-calculations.mlir
```diff
@@ -22,6 +22,27 @@ func.func @basic(%arg0: !torch.vtensor<[2,?],unk>) -> !torch.vtensor {
...
+// CHECK:           %[[TANH:.*]] = torch.aten.tanh %[[ARG]] : !torch.vtensor<*,f32> -> !torch.vtensor<*,f32>
+// CHECK:           %[[CAST:.*]] = torch.tensor_static_info_cast %[[TANH]] : !torch.vtensor<*,f32> to !torch.vtensor
+// CHECK:           return %[[CAST]] : !torch.vtensor
+func.func @basic$dtype_calculate(%arg0: !torch.vtensor<*,f32>) -> !torch.vtensor {
+  %int6 = torch.constant.int 6
+  %0 = torch.dtype.calculate {
+    %2 = torch.aten.tanh %arg0 : !torch.vtensor<*,f32> -> !torch.vtensor<*,f32>
+    torch.dtype.calculate.yield %2 : !torch.vtensor<*,f32>
+  } dtypes {
+    %2 = torch.aten.dim %arg0 : !torch.vtensor<*,f32> -> !torch.int
```
silvasean, line 37: why is torch.aten.dim needed here?

test/Dialect/Torch/simplify-dtype-calculations.mlir
```diff
@@ -0,0 +1,284 @@
...
+  } dtypes {
+    torch.dtype.calculate.yield.dtypes %dtype : !torch.int
+  } : !torch.number
+  return %0 : !torch.number
+}
+
+// -----
+
+// If result type is already refined (even if wrong, as is the case here), don't make any changes
+// to result type.
```
silvasean, line 251: maybe add a TODO here (or appropriately in the code) to raise an error if the inferred type is not a supertype of the already-present result type. An op like you have written here is actually statically invalid (nasal demons UB) so we probably don't want it here long-term.

docs/adding_abstract_interpretation_functions.md
```diff
@@ -0,0 +1,90 @@
+# Adding Abstract Interpretation Functions
+
+## Overview
+
+As part of adding support for a Torch operator in Torch-MLIR, it is usually
+necessary to define a shape and dtype function so that the compiler can infer
+the shapes and dtypes of result tensors for the operator. We use the 
+[abstract interpretation library](abstract_interp_lib.md) for this process.
```
silvasean, line 8: I don't see docs/abstract_interp_lib.md in the PR

docs/adding_abstract_interpretation_functions.md
```diff
@@ -0,0 +1,90 @@
...
+
+2. Paste the function signature into `abstract_interp_lib_gen.py` in an
+   appropriate place (ideally near other functions with a similar
+   functions). Note that `abstract_interp_lib_gen.py` will check that
+   these signatures are verbatim identical with the ones given in
+   `JITOperatorRegistryDump.txt` -- this ensures that the functions
+   don't get outdated if Torch changes an operator signature.
+
+3. Fill in the body of the function. Ideally this will just be a call
+   into a helper function from
```
silvasean, line 37: can you add a sentence here giving guidance for type functions?

docs/adding_abstract_interpretation_functions.md
```diff
@@ -0,0 +1,90 @@
...
+- You might find that there is an `Optional` value that you would
+  expect to be resolved to either a concrete value or `None`. You can
+  then look at the calculation that produces the optional value and
+  see what folds or canonicalizations are missing.
+
+See [this video](https://www.youtube.com/watch?v=E5epCJOtrf8) for general
+guidance on debugging Torch-MLIR.
+
+As a last resort, you can rewrite the function using constructs that
+`torch-simplify-shape/dtype-functions` can handle (look at other
```
silvasean, line 88: nit: I'm generally not a fan of these kinds of shortenings with `/` -- the reason is that then if you do a verbatim search for `torch-simplify-dtype-functions` you won't find this string. Even if longer, writing `torch-simplify-shape-functions/torch-simplify-dtype-functions` is superior.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/testing_framework.py
```diff
@@ -0,0 +1,303 @@
...
+# ==============================================================================
+# Shape, dtype, and decomposition function testing infrastructure.
+# ==============================================================================
+
+# We expect all functions to be adequately tested. For functions
+# implemented with upstream helpers, additional testing is usually not needed.
+# But for functions that are authored/maintained by the Torch-MLIR
+# project, we expect adequate testing.
+#
+# To do this, we provide decorators `@check_shape/dtype/decomposition_function`
```
silvasean, line 20: nit: same comment here about using the full identifiers to aid greppability/sedability

lib/Dialect/Torch/Transforms/Passes.cpp
```diff
@@ -137,23 +138,39 @@ void mlir::torch::Torch::createTorchSimplificationPipeline(
   }
 }
 
-void mlir::torch::Torch::createTorchShapeRefinementPipeline(OpPassManager &pm) {
-  // Reify the shape functions for each op that is present in the shape library.
-  pm.addPass(Torch::createReifyShapeCalculationsPass());
+template <typename ReifyCalculations, typename SimplifyCalculations>
+static void
+createRefinementPipeline(mlir::OpPassManager &pm,
+                         ReifyCalculations reifyCalculationsPass,
```
silvasean, line 144: can you use function_ref here instead of template arguments for clarity?

lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.h
```diff
@@ -0,0 +1,67 @@
...
+void importLibraryFunctions(ModuleOp module, ModuleOp library,
+                            SmallVector<std::string> functionsNeeded);
+
+// Recursively adjust `operand` to match `desiredType`.
+//
+// This function by default handles a few types such as `UnionType`,
+// `OptionalType`, and `ListType`, to name a few. However, the user
+// can intercept any type and adjust it in a custom way by defining
+// `preProcessOperand`, which gets called at the beginning of each
+// recursive call.
```
silvasean, line 58: maybe preprocessOperand -> "baseTransformation" and say "the transformtion applied to base element types. This function can be thought of as mapping baseTransformation across UnionType/OptionalType/ListType"

lib/Dialect/Torch/Transforms/ReifyDtypeCalculations.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "mlir/Parser/Parser.h"
+#include "mlir/Transforms/DialectConversion.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/Transforms/Passes.h"
+
+using namespace mlir;
+using namespace mlir::torch;
+using namespace mlir::torch::Torch;
+
+static bool isTensorTypeOrWrappedTensorType(Type type) {
```
silvasean, line 22: for someone reading without context, it seems like tuple should be handled here -- should we assert or something? (I don't think the PyTorch signatures ever have tuple arguments so it should be safe)

lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp
```diff
@@ -175,13 +172,15 @@ class AbstractlyInterpretListOpsWithinABlock
       }
       if (auto setItem = dyn_cast<Aten_SetItemTOp>(user)) {
         if (!setItem.use_empty())
-          return failure();
+          return rewriter.notifyMatchFailure(
+              op, "Expected `Aten_SetItemTOp` to not have userse");
```
silvasean, line 176: nit: userse typo

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -0,0 +1,1080 @@
...
+    return upstream_shape_functions.unary(self)
+
+def aten〇lt〇Scalar〡shape(self: List[int], other: float) -> List[int]:
+    return upstream_shape_functions.unary(self)
+
+def aten〇add〇Scalar〡shape(self: List[int], other: float, alpha: float = 1) -> List[int]:
+    return upstream_shape_functions.unary(self)
+
+@check_dtype_function([
+    Invocation(TensorOfShape(1, 1, 1, dtype=torch.float32), other=0),
```
silvasean, line 231: Is there a reason you're using 1,1,1 instead of just 1 or 1,1 or []?

You may want to define helpers NonZeroDTensorWithDtype(torch.int64) and ZeroDTensorWithDtype(torch.int64) to make the testing intent for the dtype functions clearer. (that is, not even have a place to write the shape, since it is immaterial to the test)

silvasean (changes requested):

test/Dialect/Torch/drop-abstract-interp-calculations.mlir
ramiro050: Not needed.

ramiro050:
I've replaced two of the ops I was using (`AtenAddScalarOp` and `AtenAddTensorOp`) for e2e testing this patch to ops that are less commonly encountered in large workloads: https://github.com/llvm/torch-mlir/pull/1594/commits/634dd9ac41a7dd4cc2040ebc4c5d9b8237c84e99 https://github.com/llvm/torch-mlir/pull/1594/commits/11ec64d87c9668eba8fcad2d49682d2d4ea39e29. The new ops maintain the same level of test coverage that the previous ops were doing.

While running `decomposeComplexOps` multiple times did fix the catch-22 I was encountering before, I would've still required increasing the number of iterations run in `lowerToBackendContractPass` to get tests like `Resnet18` and `mobilenet` to pass, since they use a lot of `add`s, noticeably affecting the time it took torch-mlir to lower to a particular backend.

The best approach for moving those common ops into the `abstract_lib_gen.py` file is to do a good chunk of them in a single go, which should be straight forward.

silvasean (approved):
Nice :) Let's do this!

PR #1597, opened 2022-11-16
[MLIR][TORCH] Add lowering for `prims::squeeze` op

vivekkhandelwal1:
Hi, I'm working on adding the lowering for the `prims.squeeze` op. While running the test case, I'm getting the following error:
```
****** Failed tests - 1 tests
    FAIL - "PrimsSqueezeModule_basic"
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 290, in compile_and_run_test
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/configs/linalg_on_tensors_backend.py", line 33, in compile
            module = torch_mlir.compile(
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/__init__.py", line 347, in compile
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/compiler_utils.py", line 73, in run_pipeline_with_repro_report
            raise TorchMlirCompilerError(trimmed_message) from None
        error: unsupported by backend contract: tensor with unknown rank
        note: see current operation: %2 = "torch.tensor_static_info_cast"(%arg0) : (!torch.vtensor<[?,?,?,?],f32>) -> !torch.vtensor<*,f32>
        note: this is likely due to a missing shape transfer function in shape_lib_gen.py
```
Somehow, the shape calculation is not happening. Since the op belongs to the `prims` namespace, do I need to do something additional to add the lowering of this op?


vivekkhandelwal1:
@ramiro050 @silvasean PTAL at this issue as well. This is the kind of op I have encountered for the first time.

ramiro050:
I don't think anything special needs to be done to make this work. Can you paste the IR after the failure?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -919,6 +919,13 @@ def aten〇squeeze〇dim(self: List[int], dim: int) -> List[int]:
 def prim〇NumToTensor〇Scalar(a: float) -> List[int]:
     return []
 
+def prims〇squeeze(a: List[int], dimensions: List[int]) -> List[int]:
```
ramiro050, line 922: This needs to be upstreamed. Also, can you add tests to it?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -919,6 +919,13 @@ def aten〇squeeze〇dim(self: List[int], dim: int) -> List[int]:
...
+    out = a
+    num_dims = len(dimensions)
+    for i in range(num_dims):
+        out = upstream_shape_functions.squeeze(out, dimensions[num_dims - i - 1])
```
ramiro050, line 926: You need to sort `dimensions` first, right? You will probably need to add the canonicalization pattern for `sort` from you other PR here: https://github.com/llvm/torch-mlir/blob/1d949f3ac27e89232ec376fe2d3b2ea6b5c33674/lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp#L421

You also need to handle duplicate dimensions. I think you can do this by keeping track of the dimension used in the previous iteration, then checking if it equals the new dimension (after sorting, of course)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3135,6 +3135,41 @@ class DecomposeAtenVarMeanCorrectionOp
...
+      return rewriter.notifyMatchFailure(
+          op, "all dimensions must be constant ints");
+
+    std::sort(dimensions.begin(), dimensions.end());
+    std::reverse(dimensions.begin(), dimensions.end());
+
+    BaseTensorType inputTensorTy = input.getType().cast<BaseTensorType>();
+    Type squeezeDimResultTy = inputTensorTy;
+    Value squeeze = input;
+    for (unsigned i = 0; i < dimensions.size(); i++) {
```
ramiro050, line 3158: You need to first remove duplicate dimensions

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
silvasean: The upstream ops should already have lots of tests. I think we should just add it upstream.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
vivekkhandelwal1: Actually, duplicate dimensions are allowed for this op.

ramiro050: Right, but duplicate dimensions need to only be squeezed once. In other words, you need to remove the duplicates.

```python
In [12]: a = torch.rand((1,2,1,4,5,6))

In [13]: torch.ops.prims.squeeze(a, [0,2,2,0,0]).size()
Out[13]: torch.Size([2, 4, 5, 6])
```

Also, can you add an e2e test that checks this behavior and the dimensions out of order behavior?

vivekkhandelwal1:
Closing this PR now, since it's not required.

PR #1601, opened 2022-11-16
`m_TorchConstant{Int/Bool}List` -> `m_TorchListOfConstant{Int/Bool}s`

ramiro050:
This commit renames the patterns used to match on lists of constant values to `m_TorchListOfConstant{valueType}s`. This is needed to avoid ambiguity for when `valueType` has `Optional` in it. In particular, it makes it clear whether the values in the list are optional or the list itself is optional.


silvasean (approved):

silvasean:
nit: typo Toch in PR title

PR #1604, opened 2022-11-17
[TOSA] Add aten.Scalar si64 type support

AmosLewis:
This issue is found when converting deberta model https://github.com/nod-ai/SHARK/issues/493


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -228,9 +228,9 @@ class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
                                          "Only Tensor types supported in TOSA");
 
     if (auto lhsElemTy = lhsType.getElementType().dyn_cast<IntegerType>()) {
-      if (lhsElemTy.getWidth() > 32)
+      if (lhsElemTy.getWidth() > 64)
```
eric-k256, line 231: TOSA doesn't support 64-bit integers. Is this really a 64-bit value, or just defaulting to 64-bit?

eric-k256 (changes requested):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: I trace the ops, looks like we create torch.constant.int to int64_t in ConvertAtenOp<PrimNumToTensorScalarOp>
```
%int0 = torch.constant.int 0
%230 = torch.prim.NumToTensor.Scalar %int0 : !torch.int -> !torch.vtensor<[],si64>
%232 = torch.aten.broadcast_to %230, %231 : !torch.vtensor<[],si64>, !torch.list<int> -> !torch.vtensor<[128,128],si64>
%233 = torch.aten.copy %232, %229, %false : !torch.vtensor<[128,128],si64>, !torch.vtensor<[128,128],si64>, !torch.bool -> !torch.vtensor<[128,128],si64>
%234 = torch.aten.slice.Tensor %233, %int1, %int0, %int9223372036854775807, %int1 : !torch.vtensor<[128,128],si64>, !torch.int, !torch.int, !torch.int, !torch.int -> !torch.vtensor<[128,128],si64>
%235 = torch.aten.unsqueeze %234, %int0 : !torch.vtensor<[128,128],si64>, !torch.int -> !torch.vtensor<[1,128,128],si64>
%678 = torch.aten.unsqueeze %235, %int1 : !torch.vtensor<[1,128,128],si64>, !torch.int -> !torch.vtensor<[1,1,128,128],si64>
%722 = torch.aten.neg %678 : !torch.vtensor<[1,1,128,128],si64> -> !torch.vtensor<[1,1,128,128],si64>
%723 = torch.aten.add.Scalar %722, %int256, %int1 : !torch.vtensor<[1,1,128,128],si64>, !torch.int, !torch.int -> !torch.vtensor<[1,1,128,128],si64>
```

AmosLewis: Is there any possibility we change tosa to support 64-bit?

eric-k256: With TOSA, we're trying to support hardware where si64 either doesn't exist, or is very expensive. Adding support for si64 as a default for torch constant ints would not be desirable.
If we really have to, we could consider adding it to the dialect, and then a validation pass that verifies i32 capability for hardware to use. That's a second choice for me.

silvasean: This is a recurring issue I hear from the IREE developers as well regarding f64/i64 support. Technically, !torch.int and !torch.float are si64 and f64 respectively. So truncating them would require a special user flag to say that they don't care about it, since it is possible to write code that does care. 

I filed an issue about this: https://github.com/llvm/torch-mlir/issues/1615

AmosLewis: @Svoch

eric-k256: I think the cast is okay for now. We've got the torch issue that Sean filed, perhaps we should file one here to note that this needs a long term fix, possibly finding alignment for the groups like IREE that also have this problem.

eric-k256 (approved):

AmosLewis:
There are 2 e2e test fail with this cast. Any idea to fix it? @eric-k256 
FAIL - "TensorLiteralModule_basic"
 torch_mlir.compiler_utils.TorchMlirCompilerError: Lowering TOSA to Linalg-on-Tensors failed with the following diagnostics:
        error: 'arith.addi' op requires the same type for all operands and results
        note: see current operation: %5 = "arith.addi"(%arg0, %arg1) : (i32, i64) -> i64

FAIL - "TensorOpaqueLiteralModule_basic"
torch_mlir.compiler_utils.TorchMlirCompilerError: Lowering TOSA to Linalg-on-Tensors failed with the following diagnostics:
        error: 'arith.addi' op requires the same type for all operands and results
        note: see current operation: %5 = "arith.addi"(%arg0, %arg1) : (i32, i64) -> i64

AmosLewis (commented):
@eric-k256  The one makes me feel weird is that in torchScalarToTosaTensor, how could it create an int64 constTensor if tosa dosenot support int64?
lib/Conversion/TorchToTosa/TorchToTosa.cpp Line178
`tosaTensor = tosa::getConstTensor<int64_t>(rewriter, op, {d}, dshape).value();`

eric-k256:
It was a mistake that someone made that wasn't caught, and presumably nothing used the tensor in a form where it caused errors to appear. Presumably it should have created an int32 const tensor, preferably notifying if any of the values would overflow.

AmosLewis:
> Looking at it again, you are only casting one side of the op. You'll need to cast both lhs and rhs of the op, and make sure that the result you are generating is also an i32, rather than the i64 that is coming in. That should solve the arith dialect problem, but you will also need to make sure the next downstream op is ready to receive an i32.

Add the casting of rhs type. The last 2 e2e bug fixed. But got 3 now e2e bug.
```
FAIL - "ElementwiseAddScalarInt64Module_basic"
	error: 'arith.muli' op operand #1 must be signless-integer-like, but got 'f32'
        note: see current operation: %24 = "arith.muli"(%arg1, %arg2) : (i32, f32) -> f32
FAIL - "ElementwiseAddScalarIntModule_basic"
	error: 'arith.muli' op operand #1 must be signless-integer-like, but got 'f32'
        note: see current operation: %18 = "arith.muli"(%arg1, %arg2) : (i32, f32) -> f32
FAIL - "ElementwiseSubScalarIntModule_basic"
	torch_mlir.compiler_utils.TorchMlirCompilerError: Lowering Torch Backend IR -> TOSA Backend IR failed with the following diagnostics:
        error: failed to legalize operation 'torch.aten.sub.Scalar' that was explicitly marked illegal
        note: see current operation: %2 = "torch.aten.sub.Scalar"(%arg0, %1, %0) : (!torch.vtensor<[?,?],si32>, !torch.float, !torch.int) -> !torch.vtensor<[?,?],f32>
```

AmosLewis:
Finally, fix all of the e2e bugs. Please review @eric-k256

AmosLewis:
Finally, fix all of the e2e bugs. Please review @eric-k256
```
locally xpass:
    "ElementwiseAddScalarInt64Module_basic",
    "TensorLiteralModule_basic",
    "TensorOpaqueLiteralModule_basic",
    "TypePromotionDifferentCategoryModule_basic",
    "TypePromotionSameCategoryDifferentWidthModule_basic",
    "TypePromotionZeroRankHigherCategoryModule_basic",
```

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -245,40 +251,92 @@ class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
           op, "Only floating-point or integer datatype legalization supported");
     }
 
+    Type rhsAlphaMulElemType;
+    if (outElemTy.isa<mlir::FloatType>()) {
+      rhsAlphaMulElemType = outElemTy;
+    } else {
+      // if output type is 64, input type should also be 64
+      rhsAlphaMulElemType = rewriter.getIntegerType(32);
```
eric-k256, line 259: This comment doesn't make sense in this context. Perhaps a comment at the top of the function describing what you are trying to achieve for i64 would be helpful.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -245,40 +251,92 @@ class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
...
         return rewriter.notifyMatchFailure(
             op, "Currently only scalar constants are supported for "
                 "conversion in TOSA operation");
+    } else if (rhsType.getElementType() != rhsAlphaMulElemType) {
+      // right is tensor, rhsType == tensor<i32/i64/f32>
+      // right must be cast to same type as the alpha, so MulOp success
+      rhs = rewriter.create<tosa::CastOp>(
+          op->getLoc(),
+          RankedTensorType::get(rhsType.getShape(), rhsAlphaMulElemType), rhs);
+      // reinitialize left value type to tensor<i32/f32>
```
eric-k256, line 277: You say left value, but then change the rhsType.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -245,40 +251,92 @@ class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
...
+      rhsType = rhs.getType().dyn_cast<TensorType>();
+      ;
```
eric-k256, line 279: delete this extra ; line

eric-k256 (changes requested):
I think your code is generally okay, but your comments make me question what is happening. Can you try to clean them up. An explanation of what combinations of input are valid would help.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: Done

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: Done

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: Done

AmosLewis:
> I think your code is generally okay, but your comments make me question what is happening. Can you try to clean them up. An explanation of what combinations of input are valid would help.

This is valid input type. I list at the beginning of the conversion.
```
    // left  : tensor: tensor<i32/i64/f32>
    // right : scalar: i32/i64/f32
    //         tensor: tensor<i32/i64/f32>
    // alpha : scalar: i32/i64/f32
    // output: tensor: tensor<i32/i64/f32>
```

eric-k256 (approved):

PR #1605, opened 2022-11-17
[MLIR][TORCH] Fix mean and mean.dim op for large-sized inputs

vivekkhandelwal1:
This commit fixes the aten.mean and aten.mean.dim op decomposition for supporting large-sized inputs.
This commit also fixes the formatting for the file stats.py

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -1366,11 +1366,18 @@ public:
     Value input = op.self();
     Value output = op.result();
     BaseTensorType outputTensorType = output.getType().cast<BaseTensorType>();
-    Value sum =
-        rewriter.create<AtenSumOp>(loc, outputTensorType, input, op.dtype());
+    Type outputTensorTypeAsF64 = outputTensorType.getWithSizesAndDtype(
+        outputTensorType.getSizes(), rewriter.getF64Type());
+    Value sum = rewriter.create<AtenSumOp>(
+        loc, outputTensorTypeAsF64, input,
+        rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(
+                     (int)getScalarTypeForType(rewriter.getF64Type()))));
     Value numTensorElements = rewriter.create<AtenNumelOp>(loc, input);
-    rewriter.replaceOpWithNewOp<AtenDivScalarOp>(op, outputTensorType, sum,
-                                                 numTensorElements);
+    Value mean = rewriter.create<AtenDivScalarOp>(loc, outputTensorTypeAsF64,
+                                                  sum, numTensorElements);
+    rewriter.replaceOp(op, convertTensorToDtype(rewriter, loc, mean,
+                                                outputTensorType.getDtype()));
     return success();
   }
 };
@@ -1390,7 +1397,10 @@ public:
     Value dimList = op.dim();
     Value keepDim = op.keepdim();
     Value dtype = op.dtype();
-    Type outputType = op.getType();
+    BaseTensorType outputTensorType =
+        op.result().getType().cast<BaseTensorType>();
+    Type outputTensorTypeAsF64 = outputTensorType.getWithSizesAndDtype(
+        outputTensorType.getSizes(), rewriter.getF64Type());
     MLIRContext *context = op.getContext();
 
     BaseTensorType inputType = input.getType().cast<BaseTensorType>();
@@ -1409,7 +1419,7 @@ public:
 
     // Compute sum along dimensions specified in `dimList`.
     Value sumAlongDims = rewriter.create<AtenSumDimIntListOp>(
-        loc, outputType, input, dimList, keepDim, dtype);
+        loc, outputTensorTypeAsF64, input, dimList, keepDim, dtype);
 
     // `productDimSize` is product of sizes of dimensions to be reduced.
     Value productDimSize;
@@ -1425,8 +1435,11 @@ public:
             rewriter.create<AtenMulIntOp>(loc, productDimSize, dimSize);
       }
     }
-    rewriter.replaceOpWithNewOp<AtenDivScalarOp>(op, outputType, sumAlongDims,
-                                                 productDimSize);
+    Value meanDim = rewriter.create<AtenDivScalarOp>(
+        loc, outputTensorTypeAsF64, sumAlongDims, productDimSize);
+    rewriter.replaceOp(op, convertTensorToDtype(rewriter, loc, meanDim,
+                                                outputTensorType.getDtype()));
+
     return success();
   }
 };
@@ -2752,7 +2765,7 @@ static LogicalResult calculateVariance(OpTy op, PatternRewriter &rewriter,
   BaseTensorType inputTensorTy = self.getType().cast<BaseTensorType>();
   Type outputType = op.getType();
   BaseTensorType outputTensorType = outputType.cast<BaseTensorType>();
-  Type newOutputType = outputTensorType.getWithSizesAndDtype(
+  Type outputTensorTypeAsF64 = outputTensorType.getWithSizesAndDtype(
       outputTensorType.getSizes(), rewriter.getF64Type());
   if (!inputTensorTy.hasDtype() ||
       !inputTensorTy.getDtype().isa<mlir::FloatType>()) {
@@ -2802,8 +2815,9 @@ static LogicalResult calculateVariance(OpTy op, PatternRewriter &rewriter,
   Value square = rewriter.create<AtenSquareOp>(loc, inputTensorTy, subMean);
 
   if (!unbiased) {
-    Value result = rewriter.create<AtenMeanDimOp>(
-        loc, newOutputType, square, dimList, keepDim, /*dtype=*/constantNone);
+    Value result = rewriter.create<AtenMeanDimOp>(loc, outputTensorTypeAsF64,
+                                                  square, dimList, keepDim,
+                                                  /*dtype=*/constantNone);
     result = convertTensorToDtype(rewriter, loc, result,
                                   outputTensorType.getDtype());
     rewriter.replaceOp(op, result);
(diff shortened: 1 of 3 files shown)


pashu123 (approved):
LGTM!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1366,11 +1366,18 @@ class DecomposeAtenMeanOp : public OpRewritePattern<AtenMeanOp> {
     Value input = op.self();
     Value output = op.result();
     BaseTensorType outputTensorType = output.getType().cast<BaseTensorType>();
-    Value sum =
-        rewriter.create<AtenSumOp>(loc, outputTensorType, input, op.dtype());
+    Type newOutputType = outputTensorType.getWithSizesAndDtype(
```
ramiro050, line 1369: nit: I would call this `outputTensorTypeAsF64`, since this type does not represent the actual output type of the `AtenMeanOp`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1390,7 +1397,10 @@ class DecomposeAtenMeanDimOp : public OpRewritePattern<AtenMeanDimOp> {
     Value dimList = op.dim();
     Value keepDim = op.keepdim();
     Value dtype = op.dtype();
-    Type outputType = op.getType();
+    BaseTensorType outputTensorType =
+        op.result().getType().cast<BaseTensorType>();
+    Type newOutputType = outputTensorType.getWithSizesAndDtype(
```
ramiro050, line 1402: Same comment here: `outputTensorTypeAsF64`

python/torch_mlir_e2e_test/test_suite/stats.py
```diff
@@ -66,8 +71,31 @@ def forward(self, x):
...
+        None,
+        ([-1, -1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, x):
+        return torch.ops.aten.mean(x)
+
+
+@register_test_case(module_factory=lambda: MeanLargeInputModule())
+def MeanLargeInputModule_basic(module, tu: TestUtils):
+    module.forward(100 + tu.rand(3, 4, 1024, 8192))
```
ramiro050, line 93: nit: use `low, high` to specify custom range rather than adding. This applies to the other tests below as well

python/torch_mlir_e2e_test/test_suite/stats.py
```diff
@@ -695,7 +752,6 @@ def VarCorrectionKeepDimModule_basic(module, tu: TestUtils):
 
 
 class VarCorrectionNoneModule(torch.nn.Module):
-
```
ramiro050, line 698: There are a lot of stylistic changes here unrelated to this PR. Can you make these in a separate PR?

ramiro050 (changes requested):

pashu123 (changes requested):
This is in terms of f16 support → We shouldn't be up casting f16's to f64's, casting to f32's should be enough.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
pashu123, line 1402: This will upcast int32 or int64 type to f64 type, which is not the main motivation regarding this PR. We should cast when there is a need, i.e, check dtype and then upcast. We should rather create a static function upcast that checks the dtype and casts accordingly.

pashu123: I see for `aten.mean` the output dtype is limited to float and complex.

ramiro050:
> This is in terms of f16 support → We shouldn't be up casting f16's to f64's, casting to f32's should be enough.

@vivekkhandelwal1 @pashu123, is this still needed?

vivekkhandelwal1:
> > This is in terms of f16 support → We shouldn't be up casting f16's to f64's, casting to f32's should be enough.
> 
> @vivekkhandelwal1 @pashu123, is this still needed?

Not right now. Since the fp16 tests don't work on the refbackend, this support would be added in a separate PR.

vivekkhandelwal1:
@ramiro050 Is this patch good to merge?

ramiro050 (approved):
LGTM!

tanyokwok:
Hi @vivekkhandelwal1, we found a great performance decay with the commit since all mean/sum such reduction computations are now in f64.

I am curious why we force the computation precision to f64. Can we use an option to enable f64 in your scenario and leave f32 computation by default?

cc @ramiro050 @silvasean 

Affected issue: https://github.com/alibaba/BladeDISC/issues/799

silvasean:
AFAICT PyTorch uses f32 for this computation instead of f64. Why do we need f64? Since it is causing a major regression I think we should revert this patch.

We don't want an option to control f32 or f64 here. We should just be consistent with PyTorch's behavior.

vivekkhandelwal1:
> Hi @vivekkhandelwal1, we found a great performance decay with the commit since all mean/sum such reduction computations are now in f64.
> 
> I am curious why we force the computation precision to f64. Can we use an option to enable f64 in your scenario and leave f32 computation by default?
> 
> cc @ramiro050 @silvasean
> 
> Affected issue: [alibaba/BladeDISC#799](https://github.com/alibaba/BladeDISC/issues/799)

If we revert this conversion to `f64` then the tests with the large input size fails.

silvasean:
I tried reverting the .cpp file changes in this patch locally and all the e2e test cases in the patch continue to pass. We should just revert this patch since it does not appear to contain an e2e test that fails before but passes after.

qedawkins:
In the past I referenced [this](https://github.com/pytorch/pytorch/blob/011452a2a1c745d4b12f83f89eca039f482d134b/aten/src/ATen/native/ReduceOps.cpp#L1682) to see how PyTorch was doing the mean and variance computation on CPU  when debugging a correctness issue for var.correction. This culminated in the VarCorrectionLargeInputModule_basic test added in #1100, but if the large input tests pass both with and without double precision then the tests must not be adequately testing the precision differences.

Also summation order by the backend matters here, e.g. say if we want to sum 1000 elements, we might first sum together the elements in groups of 4 and then fully reduce using these partial sums. This can avoid underflow problems with lower precision, but doesn't mean that summing with single precision in this lowering is correct. My guess is that PyTorch + CUDA won't do double precision accumulation though.

> AFAICT PyTorch uses f32 for this computation instead of f64. Why do we need f64? Since it is causing a major regression I think we should revert this patch.

> We don't want an option to control f32 or f64 here. We should just be consistent with PyTorch's behavior.

In short, agreed that PyTorch's behavior should have the final say. If they aren't doing double precision accumulation for these ops then we can happily drop this patch. For context, this precision issue first showed up for aten.var.correction with V-Diffusion, I'm unsure which model had problems for the mean.dim op though.

tanyokwok:
> Also summation order by the backend matters here, e.g. say if we want to sum 1000 elements, we might first sum together the elements in groups of 4 and then fully reduce using these partial sums.

I agree with you that the summation order and backend implementations do matter. Accumulation and computation types are chosen according to backends and algorithms, and data types. The example case you mentioned is just a special case; it's not general(Since it is only selected when reduction on all dimensions). And cuda's implementation is here: [ReduceMomentKernel.cu#L13-L42](https://github.com/pytorch/pytorch/blob/011452a2a1c745d4b12f83f89eca039f482d134b/aten/src/ATen/native/cuda/ReduceMomentKernel.cu#L13-L42). My two cents:

1. I think Torch-MLIR conversions on torch-dialect won't tell one which summation order and accumulation types should be used. It should be selected by backends.
2. The updates in the PR changed the accumulation and computation types of AtenMeanOp, which is inconsistent with what PyTorch does.

vivekkhandelwal1:
> I tried reverting the .cpp file changes in this patch locally and all the e2e test cases in the patch continue to pass. We should just revert this patch since it does not appear to contain an e2e test that fails before but passes after.

Hi @silvasean, if you run the following test:
```
class MeanLargeInputModule(torch.nn.Module):
    def __init__(self):
        super().__init__()

    @export
    @annotate_args([
        None,
        ([-1, -1, -1, -1], torch.float32, True),
    ])
    def forward(self, x):
        return torch.ops.aten.mean(x)


@register_test_case(module_factory=lambda: MeanLargeInputModule())
def MeanLargeInputModule_basic(module, tu: TestUtils):
    module.forward(tu.rand(3, 4, 256, 1024, low=100, high=200))
```
without this patch, then it won't pass. If you still want this patch to be reverted I can do that, or I can add this test for the other ops too.

vivekkhandelwal1:
> Yes, I think we want to revert this patch. Does the test pass on CUDA?

How can it be tested on CUDA?

silvasean:
We don't have a pre-built e2e config for CUDA, but just write a standalone script that runs the one op on cuda and CPU and compares them.

powderluv:
Is the revert for lack of CUDA e2e tests (and lack of a GPU CI ?) or the earlier regression by moving to f64 ?  
(Either way ok to revert until we figure out the way forward but trying to understand how we want to support CUDA models that are a mainstay of any SOTA model).

silvasean:
My understanding is that CUDA and CPU backends have provide different precision guarantees, and this test enforces the CPU precision guarantee. We generally do not want e2e tests that would fail if comparing PyTorch's native CPU and CUDA. When we have a GPU CI we can add a CUDA e2e test config which enforces this.

I documented the rationale for removing this here:

https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#special-kinds-of-tests

```
The testing of functions with special numerical precision considerations can also be tricky. Our rule of thumb is that if a test would fail across two upstream PyTorch backends (e.g. CPU and CUDA) due to different numerical precision choices, then it should not be included in our e2e test suite. See https://github.com/llvm/torch-mlir/pull/1605 for context.
```

vivekkhandelwal1:
@silvasean @tanyokwok https://github.com/llvm/torch-mlir/pull/1692 is merged now.

PR #1612, opened 2022-11-18
[MLIR][TORCH] Add aten.sort.int op

vivekkhandelwal1:

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -1346,6 +1346,40 @@ OpFoldResult AtenIntScalarOp::fold(ArrayRef<Attribute> operands) {
 }
 
 //===----------------------------------------------------------------------===//
+// AtenSortIntOp
+//===----------------------------------------------------------------------===//
+
+void AtenSortIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                                MLIRContext *context) {
+  patterns.add(+[](AtenSortIntOp op, PatternRewriter &rewriter) {
+    SmallVector<int64_t> listElements;
+    if (!matchPattern(op.self(), m_TorchListOfConstantInts(listElements)))
+      return rewriter.notifyMatchFailure(
+          op, "all input list elements must be constant ints");
+    bool reverse;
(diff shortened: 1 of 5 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1345,6 +1345,40 @@ OpFoldResult AtenIntScalarOp::fold(ArrayRef<Attribute> operands) {
...
+    SmallVector<Value> sortedListElements;
+    for (int64_t elem : listElements)
+      sortedListElements.push_back(rewriter.create<Torch::ConstantIntOp>(
+          op->getLoc(), rewriter.getI64IntegerAttr(elem)));
+    Value result = rewriter.create<Torch::PrimListConstructOp>(
+        op->getLoc(), Torch::ListType::get(rewriter.getType<Torch::IntType>()),
+        sortedListElements);
+
+    op.self().replaceAllUsesWith(result);
+    rewriter.eraseOp(op);
```
ramiro050, line 1377: I think you can just do `rewriter.replaceOp(...)`

ramiro050 (approved):
LGTM, just a small nit

lib/Dialect/Torch/IR/TorchOps.cpp
vivekkhandelwal1: That won't work because the op doesn't have any return value.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1345,6 +1345,40 @@ OpFoldResult AtenIntScalarOp::fold(ArrayRef<Attribute> operands) {
...
 
+//===----------------------------------------------------------------------===//
+// AtenSortIntOp
+//===----------------------------------------------------------------------===//
+
+void AtenSortIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                                MLIRContext *context) {
+  patterns.add(+[](AtenSortIntOp op, PatternRewriter &rewriter) {
+    SmallVector<int64_t> listElements;
+    if (!matchPattern(op.self(), m_TorchListOfConstantInts(listElements)))
```
silvasean, line 1356: you need to check that the list is not mutated.

PR #1617, opened 2022-11-21
[MLIR][TORCH] Add vision model tests to xfail set

vivekkhandelwal1:
The three vision model tests namely, `MobilenetV3Module_basic` `ResNet18Module_basic`, `ResNet18StaticModule_basic` are failing in eager_mode execution with the latest PyTorch release.


pashu123:
@vivekkhandelwal1 Is there a quick way we can resolve these tests instead of adding them to x-fail set?

vivekkhandelwal1:
> @vivekkhandelwal1 Is there a quick way we can resolve these tests instead of adding them to x-fail set?

I don't think so.

silvasean:
@powderluv @makslevental What is the plan here? Not even being able to run ResNet seems like a major regression.

makslevental:
close in favor of https://github.com/llvm/torch-mlir/pull/1624 which fixes

PR #1619, opened 2022-11-21
[MLIR][TORCH] Add E2E support for `prims.convert_element_type` op

vivekkhandelwal1:


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -730,6 +730,25 @@ OpFoldResult AtenToDtypeOp::fold(ArrayRef<Attribute> operands) {
   return getOperand(0);
 }
 
+//===----------------------------------------------------------------------===//
```
pashu123, line 733: We don't need this. There is a canonicalize pass after decompose complex ops pass. https://github.com/llvm/torch-mlir/blob/25ab8fcc1f5307258b7f3dbc2a91424e56a0f735/lib/Dialect/Torch/Transforms/Passes.cpp#L136 The canonicalize pass will trigger the canonicalization of `aten.to.dtype` which is already covered.

pashu123 (changes requested):

pashu123 (approved):
LGTM!

silvasean:
Hey Vivek, can you give more context on where you are seeing all these prims ops from? I wasn't aware that upstream was using those ops yet.

vivekkhandelwal1:
> Hey Vivek, can you give more context on where you are seeing all these prims ops from? I wasn't aware that upstream was using those ops yet.

I'm seeing these ops during the stable diffusion backward graph generation. Even for the DistilGPT2 model, this op has appeared in the IR.

Issue #1614, opened 2022-11-18
MHLO E2E tests should run as part of CI


silvasean:
We need to add a `--config=mhlo` line here: https://github.com/llvm/torch-mlir/blob/1d949f3ac27e89232ec376fe2d3b2ea6b5c33674/build_tools/python_deploy/build_linux_packages.sh#L262

When I run the MHLO e2e tests locally, it does not finish successfully. We have XFAIL's, but the compiler crashes (which cannot be caught by our xfail system) and needs to be fixed before enabling in the CI. E.g.:
```
./tools/e2e_test.sh -c mhlo -s -v -f AddSizeIntNegDimModule_basic
```

tanyokwok:
The test crashed because we didn't handle a negative input dimension in the AtenSizeIntOp conversion correctly. I will fix it in the PR https://github.com/llvm/torch-mlir/pull/1620

PR #1620, opened 2022-11-21
fix mhlo e2e ci crashes

tanyokwok:
Fix https://github.com/llvm/torch-mlir/issues/1614


lib/Conversion/TorchToMhlo/Basic.cpp
```diff
@@ -488,13 +488,23 @@ LogicalResult ConvertAtenOp<AtenSizeIntOp>::matchAndRewrite(
...
-  auto dim = rewriter.create<arith::IndexCastOp>(
-      op.getLoc(), rewriter.getIndexType(), adaptor.dim());
+
+  Value dim;
+  int64_t dimInt;
+  if (matchPattern(op.dim(), m_TorchConstantInt(&dimInt))) {
+    dimInt = toPositiveDim(dimInt, selfType.getRank());
+    dim = rewriter.create<arith::ConstantIndexOp>(op.getLoc(), dimInt);
+  } else {
+    dim = rewriter.create<arith::IndexCastOp>(
```
silvasean, line 498: I think you need to emit IR for the calculation of toPositiveDim here for the dynamic case?

silvasean (approved):
Thanks for the quick fix!

lib/Conversion/TorchToMhlo/Basic.cpp
tanyokwok: done

PR #1625, opened 2022-11-22
Fix modulus operation in LCG algorithm used for torch::GetNextSeed

ramiro050:
There seems to have been a typo in the implementation of `lowerGetNextSeed`. The LCG algorithm uses the recurrence relation $X_{n+1} = (aX_n + c) \mod m$ to generate the next seed. For the values of $a$ and $c$ used in the current implementation, $m$ is supposed to be $2^{64}$. See Donald Knuth's values for MMIX in https://en.wikipedia.org/wiki/Linear_congruential_generator .

### DIFF
--- a/lib/RefBackend/RefBackend.cpp
+++ b/lib/RefBackend/RefBackend.cpp
@@ -258,9 +258,10 @@ static Value lowerGetNextSeed(OpBuilder &b, Location loc) {
   // temp = multiplier * currentSeed + incrementStep
   Value mul = b.create<arith::MulIOp>(loc, currentSeed, multiplier);
   Value temp = b.create<arith::AddIOp>(loc, mul, incrementStep);
-  // temp mod 64 = temp & 63
-  Value cst127 = b.create<arith::ConstantOp>(loc, b.getI64IntegerAttr(127));
-  Value nextSeed = b.create<arith::AndIOp>(loc, temp, cst127);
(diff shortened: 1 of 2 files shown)


lib/RefBackend/RefBackend.cpp
```diff
@@ -258,9 +258,10 @@ static Value lowerGetNextSeed(OpBuilder &b, Location loc) {
   // temp = multiplier * currentSeed + incrementStep
   Value mul = b.create<arith::MulIOp>(loc, currentSeed, multiplier);
   Value temp = b.create<arith::AddIOp>(loc, mul, incrementStep);
-  // temp mod 64 = temp & 63
-  Value cst127 = b.create<arith::ConstantOp>(loc, b.getI64IntegerAttr(127));
-  Value nextSeed = b.create<arith::AndIOp>(loc, temp, cst127);
+  // temp mod 2**64 = temp & (2**64 - 1) = temp & i64MaxValue
+  Value i64Max = b.create<arith::ConstantOp>(
+      loc, b.getIntegerAttr(b.getIntegerType(64), APInt::getMaxValue(64)));
+  Value nextSeed = b.create<arith::AndIOp>(loc, temp, i64Max);
```
ramiro050, line 264: Since all the calculations are done with `i64`s here, I would expect this modulus operation to not be needed, right?

ramiro050:
Closing in favor of https://github.com/llvm/torch-mlir/pull/1633

Issue #1599, opened 2022-11-16
Model lowering crashes during the `RefineTypes` pass


vivekkhandelwal1:
Hi, I'm working on lowering a model, which fails during the `torchscript-module-to-torch-backend-pipeline` phase. It crashes at this line of code
https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L649
 with the given stack trace:
```
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.	Program arguments: /home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/bin/torch-mlir-opt -pass-pipeline=builtin.module(torchscript-module-to-torch-backend-pipeline) ../vivekkhandelwal1-SHARK/sd_backward_torchscript_ir_elided_16_11.mlir
 #0 0x000000000779c76a llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) /home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/externals/llvm-project/llvm/lib/Support/Unix/Signals.inc:569:11
 #1 0x000000000779c93b PrintStackTraceSignalHandler(void*) /home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/externals/llvm-project/llvm/lib/Support/Unix/Signals.inc:636:1
 #2 0x000000000779aeab llvm::sys::RunSignalHandlers() /home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/externals/llvm-project/llvm/lib/Support/Signals.cpp:103:5
 #3 0x000000000779d0b1 SignalHandler(int) /home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/externals/llvm-project/llvm/lib/Support/Unix/Signals.inc:407:1
 #4 0x00007f4808e5a420 __restore_rt (/lib/x86_64-linux-gnu/libpthread.so.0+0x14420)
 #5 0x00000000029d9192 mlir::TypeStorage::getAbstractType() /home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/externals/llvm-project/llvm/../mlir/include/mlir/IR/TypeSupport.h:129:5
[...]

PR #1630, opened 2022-11-22
[MLIR][TORCH] Fix refine types crash

vivekkhandelwal1:
This commit fixes https://github.com/llvm/torch-mlir/issues/1599.


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -947,7 +948,7 @@ void TypeAnalysis::visitOperation(Operation *op,
   if (auto sumDimIntList = dyn_cast<AtenSumDimIntListOp>(op)) {
     Type defaultDtype = operands[0]->getValue().dtype;
     // If the input dtype is bool, the result type should be i64.
```
silvasean, line 950: I don't think this is correct. You need to bail out if you don't know the defaultDtype.

silvasean (changes requested):

silvasean (approved):

Issue #1608, opened 2022-11-17
Torch-MLIR's RNGs result in highly correlated random tensors


ramiro050:
The current implementation of RNGs in Torch-MLIR creates random tensors that are highly correlated to one another. For example, running the script below that produces two uniform tensors results in the following correlation coefficient matrix for the two tensors:

```
tensor([[1.0000, 0.9701],
        [0.9701, 1.0000]], dtype=torch.float64)
```

In contrast, PyTorch produces the following matrix:

```
tensor([[ 1.0000e+00, -4.2704e-05],
        [-4.2704e-05,  1.0000e+00]], dtype=torch.float64)
```

My suspicion is that because the `getNextSeed` op and the implementation of `ConvertAtenUniformOp` use the same multiplier and increment to generate the next random number, this results in a high degree of correlation from one uniform tensor to the next.

This correlation prevents [certain decompositions](https://github.com/llvm/torch-mlir/pull/1581#discussion_r1023602897) in Torch-MLIR from working correctly.

## Script to recreate issue

```python
import torch
import torch_mlir
from torch_mlir_e2e_test.linalg_on_tensors_backends.refbackend \
    import RefBackendLinalgOnTensorsBackend

class MyModule(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x, y):
        u1 = x.uniform_()
        u2 = y.uniform_()
        return u1, u2

example_inputs = [torch.empty((1000000), dtype=torch.float64),
                  torch.empty((1000000), dtype=torch.float64)]
module = MyModule()
script_module = torch.jit.script(module, example_inputs)
torch_mlir_module = torch_mlir.compile(script_module, example_inputs,
                                       output_type="linalg-on-tensors")

backend = RefBackendLinalgOnTensorsBackend()
compiled = backend.compile(torch_mlir_module)
jit_module = backend.load(compiled)

print("Running TorchScript Graph")
print('Correlation coefficient matrix:')
script_results = script_module.forward(*example_inputs)
print(torch.corrcoef(torch.stack(script_results, 0)))
print("Running Compiled Graph")
print('Correlation coefficient matrix:')
numpy_inputs = list(map(lambda x: x.numpy(), example_inputs))
jit_results = tuple(map(lambda x: torch.tensor(x), jit_module.forward(*numpy_inputs)))
print(torch.corrcoef(torch.stack(jit_results, 0)))
```

ramiro050:
[...]

PR #1633, opened 2022-11-23
Replace LCG algorithm with squares64 algorithm in AtenUniformOp

ramiro050:
This commit replaces the LCG algorithm that was being used by the `TorchToLinalg` lowering of `AtenUniformOp` to generate random numbers with the `squares64` algorithm, for the LCG algorithm was producing tensors that were highly correlated with one another.

Squares64 algorithm: https://arxiv.org/abs/2004.06278

Closes https://github.com/llvm/torch-mlir/issues/1608

### DIFF
--- a/lib/Conversion/TorchToLinalg/Random.cpp
+++ b/lib/Conversion/TorchToLinalg/Random.cpp
@@ -57,6 +57,61 @@ public:
 };
 } // namespace
 
+static Value toLinearIndex(OpBuilder &b, Location loc,
+                           ArrayRef<Value> indicesIntValues,
+                           ArrayRef<Value> shapeIntValues) {
+  assert(indicesIntValues.size() == shapeIntValues.size() &&
+         "Expected `indices` and `shape` to have the same size");
+  Value result =
+      b.create<arith::ConstantOp>(loc, b.getZeroAttr(b.getI64Type()));
+  for (auto [index, stride] : llvm::zip(indicesIntValues, shapeIntValues)) {
+    assert(index.getType().isa<mlir::IntegerType>() &&
+           stride.getType().isa<mlir::IntegerType>() &&
+           "Input arrays to `toLinearIndex` must only contain values of type "
+           "`mlir::IntegerType`");
+    Value mul = b.create<arith::MulIOp>(loc, result, stride);
+    result = b.create<arith::AddIOp>(loc, mul, index);
+  }
+  return result;
+}
+
+// Squares64 Algorithm for generating 64-bit random numbers.
+// See: https://arxiv.org/abs/2004.06278
+static Value randomUniformUInt(OpBuilder &b, Location loc, Value ctr,
+                               Value key) {
+  auto mul = [&](Value lhs, Value rhs) -> Value {
+    return b.create<arith::MulIOp>(loc, lhs, rhs);
+  };
+  auto add = [&](Value lhs, Value rhs) -> Value {
+    return b.create<arith::AddIOp>(loc, lhs, rhs);
+  };
+  Value cst32 = b.create<arith::ConstantOp>(loc, b.getI64IntegerAttr(32));
+  auto shiftRight32 = [&](Value val) -> Value {
+    return b.create<arith::ShRUIOp>(loc, val, cst32);
+  };
+  auto swapLoHi = [&](Value val) -> Value {
+    Value leftShift = b.create<arith::ShLIOp>(loc, val, cst32);
+    Value rightShift = shiftRight32(val);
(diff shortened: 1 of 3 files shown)


python/torch_mlir_e2e_test/test_suite/rng.py
```diff
@@ -52,9 +52,9 @@ def __init__(self):
     @export
     @annotate_args([
         None,
-        ([256, 512, 8], torch.float64, True),
-        ([512, 1024, 4], torch.float64, True),
-        ([512, 256, 4], torch.float64, True),
+        ([256, 512, 12], torch.float64, True),
```
silvasean, line 55: nit: Is there a particular reason why this changed?

lib/Conversion/TorchToLinalg/Random.cpp
```diff
@@ -124,11 +158,16 @@ class ConvertAtenUniformOp : public OpConversionPattern<AtenUniformOp> {
...
                   for (int i = 0; i < resultRank; i++) {
-                    Value index = b.create<linalg::IndexOp>(loc, i);
-                    temp = getNextTemp(b, index, temp);
+                    indicesIntValues.push_back(castIndexToInt64(
+                        b, loc, b.create<linalg::IndexOp>(loc, i)));
                   }
+
+                  Value linearIndex =
+                      toLinearIndex(b, loc, indicesIntValues, sizesIntValues);
+                  Value randomVal = squares64(b, loc, linearIndex, key);
```
silvasean, line 169: nit: maybe be more specific here randomUniformUint?

silvasean (approved):
Awesome, thanks!!

I wonder if we can add a basic e2e test that tests the correlation coefficient of the random values to avoid regressions in the future?

ramiro050:
> Awesome, thanks!!
> 
> I wonder if we can add a basic e2e test that tests the correlation coefficient of the random values to avoid regressions in the future? (perhaps the 3 cases from my script in the PR: u1 = uniform(shape).flatten() u2 = uniform(shape).flatten() stack = np.stack([u1, u2]) print("two independent uniforms\n", np.corrcoef(stack)) larger = uniform((2,) + shape).reshape([2, -1]) print("split major dimension of large buffer\n", np.corrcoef(larger)) larger = uniform((2,) + shape).transpose(list(reversed(range(len(shape) + 1)))).reshape([2, -1]) print("split minor dimension of large buffer\n", np.corrcoef(larger)) )
> 
> I don't know if we have "corrcoef" supported in Torch-MLIR, but perhaps there is a way to open-code it with exisitng functions we do have?

Yeah, I thought of this, but we don't have support for `stack` or `corrcoef`. Although, `stack` can be replaced with `cat`. I'll add support for `corrcoef`

ramiro050:
Not sure why the CI is failing when running the ResNet test. I'm unable to reproduce the error locally.

vivekkhandelwal1:
> Not sure why the CI is failing when running the ResNet test. I'm unable to reproduce the error locally.

I had also seen this kind of failure before in the CI. This is probably an OOM issue. The new tests that you have added are pretty large in size. When I faced such an issue I reduced the test case size and it worked.

python/torch_mlir_e2e_test/test_suite/rng.py
```diff
@@ -76,9 +76,64 @@ def forward(self, x, y, z):
...
+        # Split second dimension of large buffer
+        corr_minor = self.correlation(larger.t().reshape(2, -1))
+
+        # This is hacky. The problem with returning just the correlations
+        # is that `torch.allclose` becomes stricter the closer values are to
+        # zero. Since the correlations are on the order of 1E-3, `rtol=1E-3`,
+        # and `atol=1E-7`, the correlations have to have a difference smaller
+        # than `atol + rtol * correlation = 1E-6`, which is too strict.
+        # By subtracting from 1, the correlations are expected to have a
+        # difference of around 1E-3.
```
silvasean, line 126: another possibly equally hacky (but perhaps more clear, and decouple it from the global rtol)is something like:

```
torch.where(torch.abs(corr) < 0.001, 1, 2)
```

python/torch_mlir_e2e_test/test_suite/rng.py
silvasean: Oh, then can we just make the sizes dynamic if they aren't important?

(not a big deal either way)

python/torch_mlir_e2e_test/test_suite/rng.py
ramiro050: That's so much better, thanks!

silvasean (approved):

PR #1636, opened 2022-11-23
[MLIR][TORCH] Add e2e support for `aten.amax` + fix a bug in computeReductionType API

Abhishek-Varma:
This PR consists of 2 commits (to maintain linear history of logically grouped tasks) :-

1. Commit 1 -> Fixes a bug in `computeReductionType` API.
    The API removes a `dim` from the size/shape if `keepdim` is `false`.
    The previous implementation was removing the wrong dimension and was crashing when
    `dim` was set to 0.

2. Commit 2 -> Adds e2e support for `aten.amax` op.


Abhishek-Varma:
Hi @vivekkhandelwal1 @ramiro050 @powderluv - this can be reviewed now.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -167,6 +167,40 @@ static Value createSoftmaxBackwardCommonKernel(PatternRewriter &rewriter,
   return sub;
 }
 
+namespace {
```
vivekkhandelwal1, line 170: Can you add a short comment explaining the decomp?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -568,6 +568,19 @@ def _reduce_along_dim(self: List[int], dim: int, keepdim: bool):
...
+        wrapped_dims.append(upstream_shape_functions.maybe_wrap_dim(dim, len(self)))
+    out: List[int] = []
+    for i, self_dim in enumerate(self):
+        if i in wrapped_dims:
+            if keepdim:
+                out.append(1)
+        else:
+            out.append(self_dim)
+    return out
+
```
vivekkhandelwal1, line 583: You have to upstream this in the PyTorch. Like https://github.com/pytorch/pytorch/pull/80860.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -589,6 +602,10 @@ def aten〇max〇dim(self: List[int], dim: int, keepdim: bool = False) -> Tuple[
     reduced_shape = _reduce_along_dim(self, dim, keepdim)
     return reduced_shape, reduced_shape
 
+def aten〇amax(self: List[int], dim: List[int] = (), keepdim: bool = False) -> List[int]:
+    return _reduce_along_dims(self, dim, keepdim)
+
+
```
vivekkhandelwal1, line 608: Remove this extra line.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -167,6 +167,40 @@ static Value createSoftmaxBackwardCommonKernel(PatternRewriter &rewriter,
...
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: dims not list of Scalar");
+
+    bool keepDim;
+    if (!matchPattern(op.keepdim(), m_TorchConstantBool(&keepDim)))
+      return rewriter.notifyMatchFailure(
+          op, "Expected a boolean value for keepDim");
+
+    Value input = op.self();
+    for (unsigned i = 0; i<dims.size(); i++) {
```
vivekkhandelwal1, line 188: Use clang-format.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
Abhishek-Varma: Yes, I'll keep this in mind and work on adding this in PyTorch along with `aten.upsample_nearest2d`'s shape function as mentioned by @ramiro050 [here](https://github.com/llvm/torch-mlir/pull/1590#discussion_r1025437892).

vivekkhandelwal1 (approved):
LGTM!

Abhishek-Varma:
Hi @ramiro050 - can you review/approve/merge this PR?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -167,6 +167,67 @@ static Value createSoftmaxBackwardCommonKernel(PatternRewriter &rewriter,
...
+///       make an adjustment (refer the code comment) before calling
+///       aten.max.dim.
+class DecomposeAtenAmaxOp : public OpRewritePattern<AtenAmaxOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenAmaxOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    SmallVector<Value> dims;
+    if (!getListConstructElements(op.dim(), dims))
```
ramiro050, line 198: nit: I would use a `matchPattern` to match on `m_TorchListOfConstantInts`, since you need the ints in the `for` loop. This prevents generating IR when the inputs are invalid

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -167,6 +167,67 @@ static Value createSoftmaxBackwardCommonKernel(PatternRewriter &rewriter,
...
+    // For every dimension included in `dim` of the op, we create a call to
+    // aten.max.dim.
+    for (unsigned i = 0; i < dims.size(); i++) {
+      int64_t cstDim = -1;
+      if (!matchPattern(dims[i], m_TorchConstantInt(&cstDim)))
+        return rewriter.notifyMatchFailure(
+            op, "unimplemented: dim must be constant");
+      // We adjust the reducing dimension in case `keepDim` is unset.
+      if (!keepDim)
+        cstDim -= i;
```
ramiro050, line 217: This does not handle all inputs correctly. Namely, it does not take into account repeated dim values and dim values out of order. I would recommend first sorting the `dims` argument and removing any duplicates, then iterating over the `dims` argument in reverse order. In reverse order no adjustment needs to take place. 

 You should add e2e tests for those two cases as well

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
```diff
@@ -589,6 +602,9 @@ def aten〇max〇dim(self: List[int], dim: int, keepdim: bool = False) -> Tuple[
     reduced_shape = _reduce_along_dim(self, dim, keepdim)
     return reduced_shape, reduced_shape
 
+def aten〇amax(self: List[int], dim: List[int] = (), keepdim: bool = False) -> List[int]:
+    return _reduce_along_dims(self, dim, keepdim)
```
ramiro050, line 606: I would've expected `upstream_shape_functions.sum_mean_dim` to work here. Is that not the case?

python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -462,6 +462,60 @@ def ReduceMaxUnsignedIntModule_basic(module, tu: TestUtils):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, a):
+        return torch.ops.aten.amax(a, (0,2))
```
ramiro050, line 493: nit: add space after comma

python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -462,6 +462,60 @@ def ReduceMaxUnsignedIntModule_basic(module, tu: TestUtils):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, a):
+        return torch.ops.aten.amax(a, (0,2), True)
```
ramiro050, line 511: nit: space afer comma, and use `keepdim=True` to make it clear what `True` corresponds to

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
Abhishek-Varma: This was really helpful @ramiro050 ! It irked me when the "adjustment" I was making was exactly same as the "inverse" of the loop IV! Should've seen that it's effectively doing what reverse iterating would've. Haha. Thank you so much for this! :D

Regarding the two test cases:
I did address it in the code and added two initially but the case with duplicate/repeated dim values bails out at the PyTorch level itself.
So, I've taken off the unnecessary check of the duplicate values from the code and now the update revision only adds one more test case to the bunch which tests for random ordering of the supplied dim values.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
Abhishek-Varma: Yes, it works - I thought of following the norm for other surrounding ops which base their shape inference on a locally created function.

I believe as a follow-up action item I can create a "properly" named API in upstream PyTorch - for instance using `sum_mean_dim` for `amax` seems unnatural at the first glance. Let me know if that action item sounds good to you. :D

I've currently incorporated `upstream_shape_functions.sum_mean_dim` in this updated revision - you may take a look.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -20,6 +20,7 @@
 #include "llvm/ADT/StringExtras.h"
 #include <cstdint>
 
+#include "llvm/Support/Debug.h"
```
ramiro050, line 23: Is this include needed?

ramiro050 (approved):
LGTM! Just one small comment

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/shape_lib_gen.py
ramiro050: Yeah, the `sum_mean_dim` name is a bit strange, but if a shape function can be expressed in terms of an upstream function, that's always preferred. In this case, `amax`, `sum`, and `mean` are all reduction functions, so it's not too odd to see `sum_mean_dim` in the shape function of `amax`. 

> I believe as a follow-up action item I can create a "properly" named API in upstream PyTorch - for instance using sum_mean_dim for amax seems unnatural at the first glance. Let me know if that action item sounds good to you. :D

Sure!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
Abhishek-Varma: Oops. Remains of debugging.
Thanks for pointing this out - I've taken it off!

AmosLewis:
Looks like the amax led to a bug in distilgpt2. The gdb trace is in the comment. @vivekkhandelwal1 
https://gist.github.com/AmosLewis/465795bfa1a4004fb96c742666515027

Issue #1618, opened 2022-11-21
Vision Model tests failing in eager_mode with the latest PyTorch release


vivekkhandelwal1:
Hi, the three vision model tests namely, `MobilenetV3Module_basic, ResNet18Module_basic, ResNet18StaticModule_basic` are failing in eager_mode execution with the latest PyTorch release https://github.com/llvm/torch-mlir/actions/runs/3508295814/jobs/5876600953. As of now, I'm adding them to the xfail set (https://github.com/llvm/torch-mlir/pull/1617) so that the Roll PyTorch action runs successfully.

makslevental:
What is the failure mode? How do I repro?

PR #1639, opened 2022-11-23
[MLIR][TORCH] Refix differentiable view

makslevental:
[Re]fix https://github.com/llvm/torch-mlir/issues/1618 by stripping `requires_grad` from results of view ops (before the exception is thrown...)

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -19,7 +19,6 @@ EAGER_MODE_XFAIL_SET = COMMON_TORCH_MLIR_LOWERING_XFAILS | {
     # These tests pass in the regular RefBackend flow, so it's unclear
     # why they fail here.
     "Matmul_vecmat",
-    "BatchMlpLayerModule_basic",
     "UpSampleNearest2dDynamicFactor_basic",
 }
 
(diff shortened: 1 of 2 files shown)


ashay (approved):
Thanks for the quick fix!

~I imagine that the eager mode test is now passing because the PyTorch version update, but I am not sure.~  Sorry, I just realized that this PR doesn't update the PyTorch version yet.

vivekkhandelwal1:
> This failed on an `XPASS`:
> 
> ```
> ****** Unexpectedly Passed tests - 1 tests
>     XPASS - "BatchMlpLayerModule_basic"
> 
> Summary:
>     Passed: 735
>     Expectedly Failed: 3
>     Unexpectedly Passed: 1
> ```
> 
> I don't know why `BatchMlpLayerModule_basic` failed before so I'm not sure if this patch should actually fix, or if this is an unrelated fail.

Remove this from the eager mode xfail set and merge.

makslevental:
> Thanks for the quick fix!
> 
> ~I imagine that the eager mode test is now passing because the PyTorch version update, but I am not sure.~ Sorry, I just realized that this PR doesn't update the PyTorch version yet.

either way, I tested locally against `dev20221122` and got a pass so as soon as this lands (and you rebase) your other PR should pass as well.

PR #1651, opened 2022-11-28
Automatically strip overloads for functorch models.

dellis23:


dellis23:
Could also add as an option that defaults to on, but Ramiro's comment in https://github.com/iree-org/iree-torch/pull/54#discussion_r995191504 made me think we might not need an option at all.

python/torch_mlir/__init__.py
```diff
@@ -300,6 +301,10 @@ def compile(model: torch.nn.Module,
     else:
         backend_legal_ops = BACKEND_LEGAL_OPS.get(output_type, [])
 
+    # For functorch-based models, automatically strip overloads
```
silvasean, line 304: strictly speaking, this is "FX-based modules", not "functorch-based models"

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -3098,3 +3099,22 @@ def forward(self):
 @register_test_case(module_factory=lambda: SortIntListReverse())
 def SortIntListReverse_basic(module, tu: TestUtils):
     module.forward()
+
+
+# ==============================================================================
+
+
+def AutomaticallyStripOverloads():
```
silvasean, line 3107: An e2e test is not appropriate for this kind of test, because we don't actually care about the computation being performed. This would be appropriate as a new file make_fx.py in https://github.com/llvm/torch-mlir/blob/main/python/test/compile_api/

silvasean (changes requested):

python/torch_mlir_e2e_test/test_suite/basic.py
dellis23: Thanks, wasn't quite sure where to put that.

silvasean (approved):

python/torch_mlir/__init__.py
```diff
@@ -300,6 +301,10 @@ def compile(model: torch.nn.Module,
     else:
         backend_legal_ops = BACKEND_LEGAL_OPS.get(output_type, [])
 
+    # For FX-based models, automatically strip overloads
```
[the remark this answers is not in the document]
dellis23: Done.

PR #1653, opened 2022-11-29
Reduce memory usage of e2e tests by reducing input sizes

ramiro050:
There are a few e2e tests that take several very large tensors as input, which leads to the e2e test suite leaking too much memory. Running things locally resulted in a total memory usage of 12.5 GB when running the suite sequentially on the refbackend.

Many of the tests that take large tensors don't actually need such large tensors to pass, and some that take several large tensors as input are just doing the same thing multiple times. This commit reduces the size of some of the tensors and removes repetitive parts of tests to reduce the memory usage to a total of 3 GB.

### DIFF
--- a/python/torch_mlir_e2e_test/test_suite/rng.py
+++ b/python/torch_mlir_e2e_test/test_suite/rng.py
@@ -44,44 +44,6 @@ def UniformModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
-class UniformStaticModule(torch.nn.Module):
-
-    def __init__(self):
-        super().__init__()
-
-    @export
-    @annotate_args([
-        None,
-        ([256, 512, 8], torch.float64, True),
-        ([512, 1024, 4], torch.float64, True),
-        ([512, 256, 4], torch.float64, True),
-    ])
-    def forward(self, x, y, z):
-        a = torch.ops.aten.uniform_(x, 1.0, 10.0)
-        b = torch.ops.aten.uniform_(y, -20.0, -5.0)
-        c = torch.ops.aten.uniform_(z, -15.0, 3.0)
-        std = torch.cat([
-            torch.flatten(torch.std(a)),
-            torch.flatten(torch.std(b)),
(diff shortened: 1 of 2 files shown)


python/torch_mlir_e2e_test/test_suite/rng.py
```diff
@@ -44,44 +44,6 @@ def UniformModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
-class UniformStaticModule(torch.nn.Module):
```
ramiro050, line 47: This test is identical to the dynamic one above, so it is not needed

ramiro050:
Below is a plot of the memory usage before each test is run on the refbackend:

image

### First Jump at $x \approx 80$

- After running
    - `BatchNorm3DModule_basic`
    - `Bernoulli...` tests
    
### Second Jump at $x \approx 410$

- After running
    - `MobilenetV3Module`
    
### Third Jump at $x \approx 530$

- After running
    - `ResNet18...` testss
    
### Fourth Jump at $x \approx 650$

- After running
    - `Uniform...` tests

silvasean:
Awesome investigation and plot! How did you generate that? To avoid this becoming an issue in the future I'd like to look into plugging in the buffer deallocation pass into the refbackend, and that plot would be super useful.

silvasean (approved):

silvasean:
Oh I see -- since we don't have fusion, the Squares64 algorithm creates a lot more intermediate temporary values which blows up the memory usage of the random tests by a lot. Good intuition about the memory consumption!

ramiro050:
> Awesome investigation and plot! How did you generate that? To avoid this becoming an issue in the future I'd like to look into plugging in the buffer deallocation pass into the refbackend, and that plot would be super useful.

Thanks! It's currently a bit of a messy process, but I added a python line to the framework that printed out the memory usage. Then I parsed the output of running all the tests to plot the memory usage. The code changes and notebook are here: https://gist.github.com/ramiro050/5b5c2501f7389c008d9029210772c3a8

PR #1654, opened 2022-11-29
build: update llvm tag to 147fe9de

vivekkhandelwal1:
Summary of changes:
- Update call to `hasNoEffect` utility
- Make fix for the dynamic dims, since `kDynamicSize` value changed to `std::numeric_limits<int64_t>::min()` from `-1` in llvm
- `makeShapeLLVMCompatible` and `makeShapeTorchCompatible` utilities convert shapes in order to remain consistent with the Torch and MLIR semantics.
- Update tags llvm: 147fe9de29dc13c14835127b35280c4d95c8e8ba mhlo: 1944b5fa6062ec4c065d726c9c5d64f1487ee8c5

### DIFF
--- a/lib/Conversion/TorchToTosa/TosaLegalizeCommon.cpp
+++ b/lib/Conversion/TorchToTosa/TosaLegalizeCommon.cpp
@@ -64,8 +64,9 @@ llvm::Optional<Value> convertReduceOpCommon(
       auto axis_attr = rewriter.getI64IntegerAttr(axis_val);
 
       shape_vec[axis_val] = 1;
-      RankedTensorType reduce_type =
-          RankedTensorType::get(shape_vec, reduce_element_type);
+      RankedTensorType reduce_type = RankedTensorType::get(
+          shape_vec,
+          reduce_element_type);
 
       auto reduce_op = CreateOpAndInfer<T>(rewriter, op->getLoc(), reduce_type,
                                            val, axis_attr);
@@ -263,12 +264,16 @@ convertReduceMeanOp(PatternRewriter &rewriter, Operation *op,
   }
 
   int64_t input_rank = input_type.getRank();
+  ArrayRef<int64_t> inputShape = input_type.getShape();
   int64_t num_elems_on_reduced_axis = 1;
   for (int i = 0; i < axes_elems.getNumElements(); i++) {
     int64_t axis_val = axes_elems.getValues<IntegerAttr>()[i].getInt();
     if (axis_val < 0)
       axis_val += input_rank;
-    num_elems_on_reduced_axis *= input_type.getShape()[axis_val];
+    if (inputShape[axis_val] < 0)
+      op->emitOpError("Failed convertReduceMean: support for dynamic input "
+                      "shape not implemented");
+    num_elems_on_reduced_axis *= inputShape[axis_val];
   }
   double div_scale = 1.0 / static_cast<double>(num_elems_on_reduced_axis);
 
--- a/lib/Conversion/Utils/Utils.cpp
+++ b/lib/Conversion/Utils/Utils.cpp
@@ -321,11 +321,34 @@ Value convertScalarToDtype(OpBuilder &b, Location loc, Value scalar, Type dtype,
 int64_t getNumberOfElements(RankedTensorType inputType) {
   if (!inputType.hasStaticShape())
     return -1;
-  ArrayRef<int64_t> inputShape = inputType.getShape();
+  SmallVector<int64_t> inputShape =
+      makeShapeTorchCompatible(inputType.getShape());
   int64_t numel = 1;
   for (int64_t i = 0; i < inputType.getRank(); i++)
     numel *= inputShape[i];
   return numel;
+}
+
+SmallVector<int64_t> makeShapeLLVMCompatible(ArrayRef<int64_t> shape) {
+  SmallVector<int64_t> updatedShape(shape);
+  int64_t kDynamicSize = ShapedType::kDynamicSize;
+  for (unsigned i = 0; i < shape.size(); i++) {
+    assert(shape[i] >= 0 || shape[i] == kUnknownSize);
+    if (shape[i] == kUnknownSize)
+      updatedShape[i] = kDynamicSize;
+  }
+  return updatedShape;
+}
+
+SmallVector<int64_t> makeShapeTorchCompatible(ArrayRef<int64_t> shape) {
+  SmallVector<int64_t> updatedShape(shape);
+  int64_t kDynamicSize = ShapedType::kDynamicSize;
+  for (unsigned i = 0; i < shape.size(); i++) {
+    assert(shape[i] >= 0 || shape[i] == kDynamicSize);
+    if (shape[i] == kDynamicSize)
+      updatedShape[i] = kUnknownSize;
+  }
+  return updatedShape;
 }
 
 } // namespace Torch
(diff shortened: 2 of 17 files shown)


lib/Conversion/TorchToArith/TorchToArith.cpp
```diff
@@ -203,8 +203,8 @@ class ConvertTorchTensorLiteralOp
         if (auto intType = type.getElementType().dyn_cast<IntegerType>()) {
           Type builtinTensorElemTy =
               IntegerType::get(context, intType.getIntOrFloatBitWidth());
-          auto shapedType =
-              RankedTensorType::get(type.getShape(), builtinTensorElemTy);
+          auto shapedType = RankedTensorType::get(
+              makeShapeLLVMCompatible(type.getShape()), builtinTensorElemTy);
```
silvasean, line 207: This does not make sense, since `type` is already a RankedTensorType so it should already be LLVM compatible.

lib/Conversion/Utils/Utils.cpp
```diff
@@ -321,13 +321,34 @@ Value convertScalarToDtype(OpBuilder &b, Location loc, Value scalar, Type dtype,
...
   for (int64_t i = 0; i < inputType.getRank(); i++)
     numel *= inputShape[i];
   return numel;
 }
 
+SmallVector<int64_t> makeShapeLLVMCompatible(ArrayRef<int64_t> shape) {
+  SmallVector<int64_t> updatedShape(shape);
+  int64_t kDynamicSize = ShapedType::kDynamicSize;
+  for (unsigned i = 0; i < shape.size(); i++) {
+    if (shape[i] == kUnknownSize)
```
silvasean, line 336: Add `assert(shape[i] >= 0 || shape[i] == kUnknownSize)` to avoid calling this on an LLVM shape array.

lib/Conversion/Utils/Utils.cpp
```diff
@@ -321,13 +321,34 @@ Value convertScalarToDtype(OpBuilder &b, Location loc, Value scalar, Type dtype,
...
+      updatedShape[i] = kDynamicSize;
+  }
+  return updatedShape;
+}
+
+SmallVector<int64_t> makeShapeTorchCompatible(ArrayRef<int64_t> shape) {
+  SmallVector<int64_t> updatedShape(shape);
+  int64_t kDynamicSize = ShapedType::kDynamicSize;
+  for (unsigned i = 0; i < shape.size(); i++) {
+    if (shape[i] == kDynamicSize)
```
silvasean, line 346: Similarly, add `assert(shape[i] >= 0 || shape[i] == kDynamicSize)` to avoid misuse of this helper.

lib/Conversion/TorchToTosa/TosaLegalizeCommon.cpp
```diff
@@ -268,7 +273,8 @@ convertReduceMeanOp(PatternRewriter &rewriter, Operation *op,
     int64_t axis_val = axes_elems.getValues<IntegerAttr>()[i].getInt();
     if (axis_val < 0)
       axis_val += input_rank;
-    num_elems_on_reduced_axis *= input_type.getShape()[axis_val];
+    num_elems_on_reduced_axis *=
+        torch::Torch::makeShapeTorchCompatible(input_type.getShape())[axis_val];
```
silvasean, line 277: @eric-k256 -- TosaLegalizeCommon is [supposed to be upstream](https://discourse.llvm.org/t/rfc-moving-tosa-construction-lib-into-mlir-core/4961) -- so it doesn't make sense for it to call Torch-MLIR-specific helpers. Just noting that this is causing divergence and tech debt -- it would be nice if you folks could actually move this upstream as was intended.

lib/Dialect/TorchConversion/IR/TorchConversionOps.cpp
```diff
@@ -24,7 +25,12 @@ using namespace mlir::torch;
 static bool haveSameSizeAndElementType(TensorType lhs, TensorType rhs) {
   if (lhs.hasRank() != rhs.hasRank())
     return false;
-  bool sameSize = lhs.hasRank() ? lhs.getShape().equals(rhs.getShape()) : true;
+  bool sameSize =
+      lhs.hasRank() ? llvm::ArrayRef<int64_t>(
+                          Torch::makeShapeTorchCompatible(lhs.getShape()))
```
silvasean, line 30: Why can't we just compare them? They should both be LLVM compatible.

silvasean (changes requested):

vivekkhandelwal1:
Hi @silvasean, some e2e tests are failing for the config `torchdynamo`, should I xfail them?

silvasean:
> Hi @silvasean, some e2e tests are failing for the config `torchdynamo`, should I xfail them?

Those seem like legitimate regressions from this patch and should be fixed before submitting. This patch should not change the IR at the backend contract level at all so additional failures when lowering are regressions.

I see that after this patch some code in DecomposeComplexOps uses ShapedType::kDynamicSize -- there is no reason for DecomposeComplexOps to use that constant and it is probably a bug.

To help catch bugs I would recommend you add an assertion in verifyTensorType that asserts that all sizes are >=0 or kUnknownSize. https://github.com/llvm/torch-mlir/blob/a8cbfff95bf0407b598181a23083428d21dacf6b/lib/Dialect/Torch/IR/TorchTypes.cpp#L203

lib/Conversion/TorchToTosa/TosaLegalizeCommon.cpp
silvasean: Yes, a copy in here would be good.

eric-k256: Yes, this is meant to move upstream, but got a bit lost with Suraj leaving. I'll try to bring it back, but will take some work on the test frontend before we can accomplish it.
In the mean time, for this case, rather than multiplying here, if a dimension is unknown / dynamic, I would return immediately. Otherwise you are creating ReduceSum(t) and multiplying the result by a garbage value. The failure lets us know that we are failing due to an unknown number of elements to divide by, rather than silently outputting garbage.

vivekkhandelwal1:
> kDynamicSize

I have done the last 2 changes suggested by you. Everything is working other than these 7 tests from torchdynamo. I'm not able to find what's the issue that's leading to this bug.

vivekkhandelwal1:
> I have done the last 2 changes suggested by you. Everything is working other than these 7 tests from torchdynamo. I'm not able to find what's the issue that's leading to this bug. @silvasean

Hi @silvasean, could you help me in fixing these failing tests for the torchdynamo?

lib/Dialect/Torch/Transforms/InlineGlobalSlots.cpp
```diff
@@ -239,14 +239,15 @@ bool InlineGlobalSlotsAnalysis::isValueSafeTransferFunction(Value value) {
...
     // safe. This covers, for example, view-like ops that create aliases.
-    if ((op->hasTrait<Torch::OpTrait::ReadOnly>() ||
-         MemoryEffectOpInterface::hasNoEffect(op)) &&
-        llvm::all_of(op->getResults(), [&](Value result) {
-          auto *state =
-              getOrCreateFor<InlineGlobalSlotsAnalysisState>(value, result);
-          return state->isSafe;
-        }))
-      continue;
+    if (auto effects = dyn_cast<MemoryEffectOpInterface>(op)) {
```
ramiro050, line 242: The torchdynamo tests are failing because this is not correct. The only change that needs to happen in this file is replacing `MemoryEffectOpInterface::hasNoEffect` with `isMemoryEffectFree`.

See: https://reviews.llvm.org/D137857

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/InlineGlobalSlots.cpp
vivekkhandelwal1: Thanks, @ramiro050, I didn't think that this could be an issue.

powderluv (approved):

PR #1622, opened 2022-11-21
Decompose torch.slice_scatter

tanyokwok:

PR #1659, opened 2022-11-30
Revert "Decompose torch.slice_scatter"

tanyokwok:
Reverts llvm/torch-mlir#1622

Will resent the PR once one of the three backends e2e tests passed: linalg, tosa, mhlo.


ramiro050 (approved):
Thanks! LGTM

Issue #1573, opened 2022-11-11
[RFC] Implementing default tensor type


gpetters94:
Do we have a plan for how to implement [this?](https://pytorch.org/docs/stable/generated/torch.set_default_tensor_type.html#torch.set_default_tensor_type) Some ops we support (such as `AtenFullOp`) assume that this is set to the default, so that setting it to `torch.HalfTensor` or something like that causes issues in `RefineTypes.cpp`. Do we want to do some checking at compile time for the default value for type inference purposes? I think it'll require a new pass right before the refine types pass.

silvasean:
We could save it as a module attribute.

gpetters94:
> We could save it as a module attribute.

I think that will work. Only issue I'm seeing with it is that there doesn't seem to be a way to determine the default tensor type in PyTorch's C++ API. I can figure it out in python by just calling `torch.tensor([])` and checking the dtype. Do you think this warrants changing the signature of `ModuleBuilder.import_module` to add an optional arg for passing the value from the python side to the C++ side?

silvasean:
there should be the equivalent of `torch.tensor([])` in the PyTorch C++ API.

PR #1661, opened 2022-11-30
Add default dtype support and fix EmbeddingOp

gpetters94:
This adds support for PyTorch's default dtype by adding it as a module attribute, as per #1573. I also included a small change to EmbeddingOp's type support - I can break that one into its own PR if you'd prefer.


ramiro050:
The PR https://github.com/llvm/torch-mlir/pull/1662#pullrequestreview-1199259702 also has a fix for EmbeddingOp, but it is slightly different. Can you check which one is the correct one?

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -497,12 +498,25 @@ class TypeAnalysis : public dataflow::SparseDataFlowAnalysis<
...
+    auto default_dtype =
+        module->getAttr("torch_mlir.default_dtype").cast<StringAttr>().str();
+
+    Type dtype;
+    if (!default_dtype.compare("torch.float16")) {
+      dtype = Float16Type::get(context);
+    } else if (!default_dtype.compare("torch.float64")) {
+      dtype = Float64Type::get(context);
+    } else {
+      dtype = Float32Type::get(context);
```
ramiro050, line 516: nit: can you use an `else if ` for the `float32` case as well and have the `else` finish with an `llvm_unreachable`? This makes it very clear that every possible case is being handled appropriately

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/module_builder.cpp
```diff
@@ -12,6 +12,7 @@
 #include "function_importer.h"
 #include "ivalue_importer.h"
 #include "mlir_utils.h"
+#include <torch/csrc/tensor/python_tensor.h>
```
ramiro050, line 15: nit: can you place this include after all the other includes to match the style of the other files in this directory?

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/module_builder.cpp
```diff
@@ -174,6 +175,24 @@ void ModuleBuilder::importModule(torch::jit::Module jitModule,
...
+    break;
+  case at::ScalarType::Double:
+    globalDType = "torch.float64";
+    break;
+  default:
+    globalDType = "torch.float32";
+    break;
+  };
+  auto moduleDType = mlirStringAttrGet(context, toMlirStringRef(globalDType));
+  mlirOperationSetAttributeByName(mlirModuleGetOperation(module),
```
ramiro050, line 192: @silvasean, is there no way to set an MLIR Type as an attribute to avoid comparing strings in RefineTypes?

ramiro050 (changes requested):
Thanks for this!

gpetters94:
> The PR [#1662 (review)](https://github.com/llvm/torch-mlir/pull/1662#pullrequestreview-1199259702) also has a fix for EmbeddingOp, but it is slightly different. Can you check which one is the correct one?

They're both the same, it's just that this one adds it to the list of "take dtype from first operand" ops while #1662 explicitly sets it. I guess it comes down to stylistic choice.

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/module_builder.cpp
silvasean: You can use a TypeAttr -- construct via mlirTypeAttrGet.

gpetters94: Is there a way to get a Type here? I don't think there's an MLIR context at this point.

PR #1662, opened 2022-11-30
[RefineTypes] Fix knowledge dtype for `aten.embedding` op

Abhishek-Varma:
-- The dtype of the result of `aten.embedding` should match that of
   the `weight` operand's (operand[0]) instead of hardcoding to f32.
-- This commit aims to provide a fix for the same.

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -155,6 +155,7 @@ MHLO_PASS_SET = {
     "EmbeddingModuleI32Static_basic",
     "EmbeddingModuleI32_basic",
     "EmbeddingModuleI64_basic",
+    "EmbeddingModuleF16_basic",
     "ExpandAsIntModule_basic",
     "ExpandModule_basic",
     "FullLikeModuleDefaultDtype_basic",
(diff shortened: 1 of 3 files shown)


pashu123 (approved):
LGTM!

gpetters94:
It might be better practice to put the op in [this list](https://github.com/llvm/torch-mlir/blob/d57a16eb1085895d592f0daf2bcc87449bfbc168/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L680).

silvasean (changes requested):
This change requires an e2e test see https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md

pashu123:
> This change requires an e2e test see https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md

Let me enable refbackend for float16 types, I guess the conversion support is already there: https://github.com/llvm/llvm-project/blob/main/mlir/python/mlir/runtime/np_to_memref.py

pashu123:
@silvasean I have raised the patch for f16 support here: https://github.com/llvm/torch-mlir/pull/1668 This will require https://reviews.llvm.org/D139121 If you could review the patches that would be great.

pashu123:
@Abhishek-Varma The float16 support for refbackend is merged, now you can add an e2e test.

Abhishek-Varma:
Hi @silvasean - I've added one e2e test case. Can you approve now?

silvasean (approved):

PR #1671, opened 2022-12-02
Allow running DecomposeComplexOps more than once

ramiro050:
The current implementation of `DecomposeComplexOps` fails if an op expected to be decomposed does not get decomposed in the first iteration of the `createTorchSimplificationPipeline` in `LowerToBackendContractPass`. However, some graphs require multiple iterations of `createTorchSimplificationPipeline` to fully propagate all statically knowable information, such as dtypes and shapes, to the entire graph, sometimes resulting in the need to run `DecomposeComplexOps` more than once.

This commit changes `DecomposeComplexOps` to use a greedy algorithm for pattern application and moves the legalization check of ops to the `LowerToBackendContractPass` to allow for the `DecomposeComplexOps` to run more than once.

### DIFF
--- a/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
+++ b/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
@@ -11,10 +11,12 @@
 
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/Pass/PassManager.h"
+#include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/Passes.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchDialect.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Transforms/Passes.h"
+#include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/Support/Debug.h"
 
 #define DEBUG_TYPE "torch-lower-to-backend-contract"
@@ -26,6 +28,10 @@ using namespace mlir::torch::Torch;
 //===----------------------------------------------------------------------===//
 // Checking the backend contract.
 //===----------------------------------------------------------------------===//
+
+static void markDecomposedOpsAsIllegal(MLIRContext *context,
+                                       ConversionTarget &target,
+                                       ArrayRef<std::string> backendLegalOps);
 
 static LogicalResult checkType(Operation *op, Type type,
                                bool actuallyEmitDiagnostics) {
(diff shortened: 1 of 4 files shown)


lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -211,6 +217,12 @@ class LowerToBackendContractPass
   void runOnOperation() override {
     ModuleOp module = getOperation();
 
+    if (decompose) {
+      MLIRContext *context = &getContext();
+      ConversionTarget target(*context);
+      markDecomposedOpsAsIllegal(context, target, backendLegalOps);
```
silvasean, line 223: Is there something missing here? I would expect `target` to be used somewhere. In particular in satisfiesBackendContract.

(also is it possible to test this? this does sound like something that a .mlir test would be useful for)

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -265,3 +277,107 @@ std::unique_ptr<OperationPass<ModuleOp>>
 mlir::torch::Torch::createVerifyBackendContractPass() {
   return std::make_unique<VerifyBackendContractPass>();
 }
+
+static void markDecomposedOpsAsIllegal(MLIRContext *context,
```
silvasean, line 281: Is there an appropriate place where you can put a lightweight "architecture" comment describing the overall approach here (since it crosses multiple files)?

Since this change actually formalizes the backend contract better (now satisfiesBackendContract will check the legal set of ops, rather than having that "satisfaction" check hidden inside DecomposeComplexOps), we probably want a nice big comment in this file for when folks read "See the extensive comments in the function satisfiesBackendContract (and its callees) in the LowerToBackendContract pass for an extended rationale for these decisions, and a precise definition of the backend contract." in https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#the-backend-contract

silvasean (changes requested):

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
ramiro050: Whoops! Not sure why my brain thought all I had to do was set the target. Should be fixed now

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
ramiro050: I've added a comment explaining the rationale. Let me know if you want me to change anything

ramiro050:
Ignore the first commit. That's from this PR: https://github.com/llvm/torch-mlir/pull/1693

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -247,7 +280,10 @@ class VerifyBackendContractPass
     : public VerifyBackendContractBase<VerifyBackendContractPass> {
 public:
   void runOnOperation() override {
-    if (!satisfiesBackendContract(getOperation(), /*actuallyEmitDiagnostics=*/true)) {
+    MLIRContext *context = &getContext();
```
silvasean, line 283: Is there a way we can pass in the appropriate options here? It seems a bit weird for VerifyBackendContract to say "yep" but LowerToBackendContract to disagree.

silvasean (approved):

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
ramiro050: That's a good point. Also, do you know why we have this pass at all? I don't see it used anywhere in the codebase

silvasean: @GlebKazantaev added it with the goal of ensuring that the IR produced by LTC meets the backend contract. It looks like it never got enabled -- @GlebKazantaev can you please enable it?

PR #1673, opened 2022-12-02
[TORCH] Add Complex Number support

ataheridezfouli-groq:
Add Complex number dtype support to torch tensors. Add aten.fft_fft op to test complex numbers.


ataheridezfouli-groq:
@ramiro050 do you mind taking a look. Thanks in advance!

ramiro050:
Hey @ataheridezfouli-groq, the patch we were talking about has landed: https://github.com/llvm/torch-mlir/pull/1594. For an example of how to write a dtype function, see:

- https://github.com/llvm/torch-mlir/blob/main/docs/adding_abstract_interpretation_functions.md
- https://github.com/llvm/torch-mlir/blob/b1f68328497c53aa2a9aba5dfa079c69a0b72132/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py#L646-L655

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1157,6 +1157,19 @@ void TypeAnalysis::visitOperation(Operation *op,
     return;
   }
 
+  if (isa<AtenFftFftOp>(op)) {
```
ataheridezfouli-groq, line 1160: @ramiro050 if I remove this code from RefineTypes, it stops being able to infer the fft type. Not sure if I'm missing something.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/registry.py
```diff
@@ -183,8 +183,8 @@ def uppercase_first_letter(s):
         return op_name, cpp_class_name
 
     def _get_function_signature(self, function_kind: str,
-                                parameter_decl_builder: Callable["SIG_ATTR_TYPE", str],
-                                ret_decl_builder: Callable["SIG_ATTR_TYPE", str]) -> str:
+                                parameter_decl_builder: Callable[["SIG_ATTR_TYPE"], str],
+                                ret_decl_builder: Callable[["SIG_ATTR_TYPE"], str]) -> str:
```
ataheridezfouli-groq, line 187: @ramiro050 I was getting the following error so I made these changes:
```
TypeError: Callable[args, result]: args must be a list. Got SIG_ATTR_TYPE
```
Do you also see these errors, or do I have stale code?

ramiro050: Ah, this might be a Python version issue. Your version seems to be the right one.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
ramiro050: This probably means the dtype function is not being simplified correctly. Let me take a look

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/registry.py
ataheridezfouli-groq: Hmm interesting. I am on python version `3.8.10`.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
ramiro050: You need to add handling here: https://github.com/llvm/torch-mlir/blob/64f9a0e978d530a2c00d2134e6a01c41cb034b9a/lib/Dialect/Torch/Utils/Utils.cpp#L77

ataheridezfouli-groq: Ah perfect yeah that was the issue! Thanks for looking into it!

One final question. I removed my lit tests from `refine-types-ops.mlir` since I removed the code from `RefineTypes.cpp`. So where can I place lit tests now with this new flow? `reify-dtype-calculations.mlir`?

powderluv:
btw thanks for your contribution

lib/Dialect/Torch/Transforms/RefineTypes.cpp
ramiro050: I would just add one test in `simplify-dtype-calculations.mlir` showing that if the `dtypes` region returns a complex dtype, the right `dtype` gets assigned to the op. Something like this test:

https://github.com/llvm/torch-mlir/blob/64f9a0e978d530a2c00d2134e6a01c41cb034b9a/test/Dialect/Torch/simplify-dtype-calculations.mlir#L14

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/registry.py
ramiro050: There's probably something different about or Pythons. Mine never type checks when I execute a program, so that's why I didn't notice that error. Your implementation matches the docs for `Callable` so lets keep yours

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -970,6 +970,20 @@ def aten〇index〇Tensor_hacked_twin〡shape(self: List[int], indices: List[Lis
 def aten〇cat〡shape(tensors: List[List[int]], dim: int = 0) -> List[int]:
     return upstream_shape_functions.cat(tensors, dim)
 
+def aten〇fft_fft〡shape(self: List[int], n: Optional[int] = None, dim: int = -1, norm: Optional[str] = None) -> List[int]:
+    return self
+
+def aten〇fft_fft〡dtype(self_rank: int, self_dtype: int, n: Optional[int] = None, dim: int = -1, norm: Optional[str] = None) -> int:
```
ramiro050, line 976: Can you add checks thoroughly testing the logic of this implementation?

For reference, see:

https://github.com/llvm/torch-mlir/blob/64f9a0e978d530a2c00d2134e6a01c41cb034b9a/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py#L51-L65

https://github.com/llvm/torch-mlir/blob/64f9a0e978d530a2c00d2134e6a01c41cb034b9a/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/testing_framework.py#L15

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
```diff
@@ -54,6 +54,12 @@ static MlirType getMlirTypeForTorchScalarTypeRaw(MlirContext context,
     return torchMlirTorchQInt8TypeGet(context);
   case ScalarType::QUInt8:
     return torchMlirTorchQUInt8TypeGet(context);
+  case ScalarType::ComplexHalf:
```
ataheridezfouli-groq, line 57: Fix this

test/Dialect/Torch/simplify-dtype-calculations.mlir
```diff
@@ -283,3 +283,32 @@ func.func @refine_dtype$derefine_result_type(%arg0: !torch.int, %arg1: !torch.in
...
   return %0 : !torch.number
 }
+
+// CHECK-LABEL:   func.func @refine_dtype$complex_type(
+// CHECK-SAME:                     %[[ARG:.*]]: !torch.vtensor<*,f32>) -> !torch.vtensor {
+// CHECK:   %[[DTYPE_INT:.*]] = torch.constant.int 9
+// CHECK:   %[[NONE:.*]] = torch.constant.none
+// CHECK:   %[[INTM1:.*]] = torch.constant.int -1
+// CHECK:   %[[RESULT:.*]] = torch.dtype.calculate {
+// CHECK:     %[[FFT:.*]] = torch.aten.fft_fft %[[ARG]], %[[NONE]], %[[INTM1]], %[[NONE]] : !torch.vtensor<*,f32>, !torch.none, !torch.int, !torch.none -> !torch.vtensor<*,complex<f64>>
```
ramiro050, line 293: nit: I would just have a check for this line and remove the other lines to make it clear what is being tested, similar to the tests above this one

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
ramiro050: On Github I still don't see checks added here

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
ataheridezfouli-groq: Oh my bad I totally missed your comment. Will add.

test/Dialect/Torch/simplify-dtype-calculations.mlir
ataheridezfouli-groq: I absolutely agree with your nit! Will fix

ramiro050 (approved):
LGTM!

ataheridezfouli-groq:
Thanks so much for all the help @ramiro050! Do you mind merging for me as well please, I don't have write access.

ramiro050:
> Thanks so much for all the help @ramiro050! Do you mind merging for me as well please, I don't have write access.

Done! Thanks for your contribution!

PR #1681, opened 2022-12-06
[torchdynamo] Add "lockstep" numerical accuracy debugger.

silvasean:
Thanks to TorchDynamo's great layering and design, this is only about 100 lines of code for a basic lockstep debugger.

This should allow us to deprecate eager_mode, since AFAIK the only interesting use case that it was really supporting is for downstream users to write lockstep debuggers.

NOTE: The exact reporting and interface here is subject to change. Please try it out and provide feedback (or patches :) ).
- make_fx should not drop source locations: https://github.com/pytorch/pytorch/issues/90276
- Report tensors better (huge tensors should be summarized)
- Maybe don't abort, but just warn?
- Allow customizing atol/rtol.
- How best to print the failing node? And include surrounding graph context?

### DIFF
--- a/python/test/debug/lockstep_basic.py
+++ b/python/test/debug/lockstep_basic.py
@@ -1,0 +1,46 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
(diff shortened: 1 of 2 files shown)


stellaraccident (commented):
This is really neat!

powderluv (approved):
This is awesome. If we have this we can drop our downstream lockstep tracer. @makslevental @qedawkins FYI

PR #1684, opened 2022-12-06
build: update llvm tag to 798fa4b4

ramiro050:
- Support for non-prefixed accessors has been removed. See:
  https://reviews.llvm.org/D136727
- Rename `operands` to `methodOperands` in `prim.CallMethod` since the
  name `operands` overlaps with a builtin method name. See:
  https://reviews.llvm.org/D136727
- Add passes in refbackend to lower memref.subview. See:
  https://reviews.llvm.org/D136377
- Replace `CopyToValueTensorOps` first in `RewriteViewLikeSubgraph` in
  maximize-value-semantics.

  The current implementation of the `RewriteViewLikeSubgraph` pass in
  maximize-value-semantics creates temporarily invalid IR. In
  particular, given a forward slice starting from a
  `CopyToNonValueTensorOp` and ending in `CopyToValueTensorOp`s, the
  pass first replaces all uses of the `CopyToNonValueTensorOp` with
  its operand, which results in all the `CopyToValueTensorOp` users
  having their operand have type `!torch.vtensor`, which is invalid.

  The correct way to do things is to first replace all the
  `CopyToValueTensorOp`s with their operand, and then replace all uses
  of the `CopyToNonValueTensorOp` with its operand.

  This only started failing now because the generated accessor
  `getOperand` for the `CopyToValueTensorOp` now returns a
  `TypedValue<NonValueTensorType>`, which has an assert checking that
  the value returned is of the expected type.

### DIFF
--- a/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
+++ b/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
@@ -64,7 +64,7 @@ public:
     InterpretedOps result;
     result.copyLikeOps.push_back(copyToNonValueTensor);
     DenseSet<Value> availableAliases{
-        assertNonValueTensor(copyToNonValueTensor.result())};
+        assertNonValueTensor(copyToNonValueTensor.getResult())};
     for (Operation *user : nonValueTensorUsers) {
       for (Value operand : nonValueTensorsUsedByOp.lookup(user)) {
         if (!availableAliases.contains(operand)) {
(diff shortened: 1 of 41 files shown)


ashay:
Thanks @jpienaar, this is very useful!

lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
```diff
@@ -292,12 +292,12 @@ class RewriteViewLikeSubgraph
...
       return rewriter.notifyMatchFailure(copy, "no types to change");
 
-    // All uses of `copy` will be updated by the logic below.
-    copy.replaceAllUsesWith(copy.getOperand());
     // All CopyToValueTensorOp operands will be changed to the correct type
     // by the logic below.
     for (CopyToValueTensorOp op : copyToValueTensorOps)
       rewriter.replaceOp(op, op.getOperand());
+    // All uses of `copy` will be updated by the logic below.
+    copy.replaceAllUsesWith(copy.getOperand());
```
silvasean, line 300: LGTM as long as at least one of the the existing tests crashes with the new code

ramiro050: Some e2e tests crash. I'll add a lit test that also fails

ramiro050: Done: https://github.com/llvm/torch-mlir/pull/1684/files#diff-73baa2b9ed878e33c92f5cbe0fa758bdb08b93b1f7c0686feb033a7f48a3969a

PR #1693, opened 2022-12-07
Remove unnecessary decompose-complex-ops tests

ramiro050:
This commit removes lit tests from the `decompose-complex-ops` that are essentially testing a macro expansion, in accordance with https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing .


silvasean (approved):
Yeah, generally speaking .mlir tests are not very useful for this kind of code. We might want to remove even more but this is fine for now.

Issue #1460, opened 2022-10-04
Enable VerifyBackendContract in base lazy backend.


GlebKazantaev:
### Description
This issue is a remaining part of https://github.com/llvm/torch-mlir/pull/1412 (see discussions in this PR to understand the purpose of VerifyBackendContract pass in base lazy backend).

### Details
To enable `VerifyBackendContract` we have either to figure out how to detach MlirOperation from MlirModule after pass manager execution or update `TorchMlirComputation` and some other facilities to work with MlirModule. From my point of view the second option is more preferable as it looks more generic and detaching approach looks like workaround.

Assignee: @GlebKazantaev

GlebKazantaev:
Hi @silvasean, I have some thoughts regarding out discussion about backend contract and LTC violations. I did more investigation about `ScalarImplicit` and `!torch.number` type which is unsupported by `satisfiesBackendContract` https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L146

So the first concern is that original LTC graph we are trying to convert to torch-mlir doesn't consist of any `!torch.number` types. The reason why we get them later is because in [our torch-mlir code during](https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/csrc/base_lazy_backend/mlir_node_lowering.cpp#L278-L284) `Scalar::Lower` we convert scalar value to tensor `at::scalar_tensor(value, options)`. So in cases when consumer of `Scalar` requires `Tensor` type, we just insert `prim::Constant[value={1}]()` and no additional type alignment happens. But when the `Scalar` consumer requires scalar literal but we return `Tensor` type, so the type [alignment mechanism start working](https://github.com/GlebKazantaev/pytorch/blob/master/torch/csrc/jit/frontend/schema_matching.cpp#L137) and insert `ScalarImplicit` operation with `!torch.number` type.
From my point of view this is rather torch-mlir issue than a PyTorch LTC violation. To avoid this type alignment we need to know consumers types inside `TorchMlirNode::Lower`. But unfortunately `lazy::Node` stores only operands, so we don't have an access to consumer types during lowering.

I also wonder, is it a real problem with `!torch.number` type? Can we just extend the list of scalar types here https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L33 with `NumberType` ? From backends perspective this is a normal type like others, so they can handle it.

[...]

PR #1699, opened 2022-12-08
Extended TorchMLIRLoweringContext with virtual CreateComputation method

GlebKazantaev:
### Description

* Extended TorchMLIRLoweringContext with virtual `CreateComputation` method, so we can override it to create our custom computation.
* Fixed `device_data_cast` return value.

### DIFF
--- a/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
+++ b/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
@@ -142,7 +142,11 @@ ComputationPtr TorchMlirLoweringContext::Build() {
       /*getArgAttribute=*/[](int) -> MlirAttribute { return {nullptr}; },
       /*importOptions=*/{/*assumeTensorsHaveValueSemantics=*/true});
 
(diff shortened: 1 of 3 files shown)


antoniojkim (approved):

GlebKazantaev:
Hi @silvasean!
The only reason why VerifyBackendContract pass is not yet enabled in the LTC frontend is that PyTorch produces `ScalarImplicit` type which is prohibited by our backend contract, so enabling VerifyBackendContract pass will immediately fail conversion for all models. We had a discussion with you about this problem and enabling VerifyBackendContract pass  sometimes ago in this thread: https://github.com/llvm/torch-mlir/issues/1460 and the plan was try to fix issue with `ScalarImplicit` on PyTorch side first and in case if changes are not accepted, we should implement JIT pass to handle this type.
Unfortunately PR with fix in PyTorch https://github.com/pytorch/pytorch/pull/86692 was reviewed just couple a days ago with a suggestion to fix it on our side... So my plan now is to develop JIT pass and then enable VerifyBackendContract pass as I promised.

PR #1704, opened 2022-12-09
[cleanup] fix naming of private variable according to the style guide

ashay:


ashay:
Fixing the naming based on a comment in a previous PR: https://github.com/llvm/torch-mlir/pull/1555#discussion_r1015248466.

silvasean (approved):

PR #1705, opened 2022-12-09
Pass op legality info to verifyBackendContractPass

ramiro050:
In order to verify if a given IR satisfies the backend contract, the verifier needs to know if decompositions took place, and if so, which ops were decomposed and which were not.

This commit adds two arguments to `verifyBackendContractPass` to specify if decompositions took place and which ops to consider backend legal, similar to the arguments of `LowerToBackendContractPass`.


silvasean (approved):
Thanks!

nit: Remove WIP from final commit title.

PR #1707, opened 2022-12-10
Make `getTensorRank` safer by changing return to `Optional<unsigned>`

ramiro050:
Currently `getTensorRank` returns -1 if it was unable to get the rank of the tensor. However, not every use in the codebase was checking the return value, and in some cases, the return value was casted to unsigned leading to some infinte loops when an unranked tensor reached a decomposition.

This commit changes the return of `getTensorRank` to `Optional<unsigned>` to make it clear to the user that the function can fail.

This commit also changes a couple of for loops that iterate a vector in reverse order that can potentially become infinite loops into range-based for loops.


vivekkhandelwal1 (approved):
Thanks for the fix! LGTM

Issue #1709, opened 2022-12-12
run_audit_wheel causes ODR violation with torch


dellis23:
See: https://github.com/iree-org/iree-torch/actions/runs/3674670300/jobs/6213170178#step:5:8

`run_audit_wheel` appears to be necessary for publishing packages to pypi.  I'll be temporarily disabling pypi publishing until we can figure out a fix.

powderluv:
Now we have a non-jit version that should pass audit_wheel. I will try and hook it up to build the wheels.

PR #1715, opened 2022-12-13
Disable pypi publishing

dellis23:
See https://github.com/llvm/torch-mlir/issues/1709


dellis23:
Tested here: https://github.com/llvm/torch-mlir/actions/runs/3678892754

powderluv (approved):

PR #1717, opened 2022-12-14
build: update llvm tag to 3a020527

ashay:
Summary of changes:

 - Replace `llvm::None` with `std::nullopt`, since the former is deprecated
   (https://reviews.llvm.org/D139763)

 - Use setter for symbol visibility instead of passing string attribute when
   creating FuncOp

---

@powderluv @silvasean I am not sure if these changes are sufficient, since the ./build_tools/update_*.sh scripts fail at my end, but most of the lit and e2e tests pass.  If the build fails, feel free to take over.

### DIFF
--- a/include/torch-mlir/Conversion/Utils/Utils.h
+++ b/include/torch-mlir/Conversion/Utils/Utils.h
@@ -84,7 +84,7 @@ SmallVector<Value> getTypeConvertedValues(OpBuilder &b, Location loc,
 // should be converted builtin types.
 Value convertScalarToDtype(
     OpBuilder &b, Location loc, Value scalar, Type dtype,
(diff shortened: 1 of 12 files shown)


powderluv (approved):

powderluv:
Ok to land if tests pass and tree is green. But will let @silvasean take over if there is more work needed for the bump.

silvasean:
Hey thanks for the fixes! I was on the rotation this week but thank you for taking over :)

I will be on vacation over the weeks but I will need to do an extra rotation at some point to make up for this.

ashay:
No worries!  It's too bad that the green commit wasn't ready to be published early like y'all had requested, so I figured I'd go ahead and do the LLVM bump as soon as the commit hash was known.

PR #1718, opened 2022-12-14
[cleanup] Fix a few more llvm::None -> std::nullopt

silvasean:


ashay (approved):
LGTM, thanks for catching these!

PR #1722, opened 2022-12-14
Add `report_fatal_error` to `getTypeForScalarType`

ramiro050:
Functions like `getTypeForScalarType` that do a mapping from one set of types to another should not fail, and if they do it should be obvious to the developer that that function has an unhandled case.

Instead of silently failing when encountering an unsupported type, this commit adds a `report_fatal_error` at the end, similar to other type translation functions in this file.


ramiro050:
This was an issue here: https://github.com/llvm/torch-mlir/pull/1673#discussion_r1049051482

gpetters94 (approved):
LGTM

vivekkhandelwal1 (approved):

PR #1723, opened 2022-12-14
Update buildAndTest.yml

powderluv:


powderluv:
Switch CI to using GPU builders. Seems to work reasonably well and has ccache working too https://github.com/llvm/torch-mlir/actions/runs/3699601811/attempts/2 and https://github.com/llvm/torch-mlir/actions/runs/3699601811/attempts/3 .  

Will let it run a few days before moving the nightly release builder too. Then we can start enabling gpu tests and binary builds with CUDA support.

silvasean (approved):
Thanks!!

ashay:
This is great!

PR #1724, opened 2022-12-15
Replace asIntArrayRefSlow with macro

antoniojkim:
https://github.com/pytorch/pytorch/commit/8090cb5386dccf4cf341aea585c793dfbb6c6002 replaced usages of `asIntArrayRefSlow` with a macro. As such, we need to replace the usages in Torch-MLIR too


ashay:
Thanks for looking into this!  I think you might need to update the PyTorch version and commit hash as well.  Here are the relevant lines from the RollPyTorch action:

> Found torch release 2.0.0.dev20221215
> Found torch commit hash 1902ca9d533da6d24b1fb9d0c345349d51ae26ad

ashay (approved):
Thanks for the quick fix!

PR #1727, opened 2022-12-16
build: fix LTC code after upstream PyTorch change

ashay:
pytorch/pytorch@140a3139 reverted a change from yesterday, causing the
RollPyTorch action to break.  This patch reverts the corresponding
change in the torch-mlir LTC code.

This patch also re-enables tests that were previously marked as XFAIL.


ashay:
The [upstream PyTorch issue](https://github.com/pytorch/torchdynamo/issues/1891) isn't marked as fixed, but we don't seem to run into the assertion failure now.

antoniojkim (approved):

PR #1729, opened 2022-12-16
Switch to PyTorch CUDA nightly

powderluv:
Switch to PyTorch CUDA nightlies so we can run e2e tests on GPU too


pytorch-requirements.txt
```diff
@@ -1,4 +1,4 @@
--f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
+-f https://download.pytorch.org/whl/nightly/cu117
```
pashu123, line 1: cu117/torch_nightly.html.

powderluv:
No longer required

PR #1730, opened 2022-12-17
ci: post comment on RollPyTorch tracker issue upon build failure

ashay:
Now that the RollPyTorch tracker issue exists, we can automate the job
of notifying folks of failures instead of having to do it manually.
This patch adds a step to the workflow to post such a message.


ashay:
The comment would look like this:

> The RollPyTorch action has failed.  See [CI log](https://github.com/llvm/torch-mlir/actions/runs/3713343822) for details.
> 
> The following information may come handy when fixing the code.
> ```
> torch version: 2.0.0.dev20221216
> torch commit hash: 2bd3fbe2e6b42f3c8f7eea443c89b515cb1981f1
> torchvision version: 0.15.0.dev20221216
> ```

powderluv (approved):

PR #1734, opened 2022-12-18
[Torch][MHLO] Decompose aten.copy op. Lower aten.rsqrt & sigmoid to mhlo.

li-plus:
This PR enabled lowering `aten.copy`, `aten.rsqrt` and `aten.sigmoid` to MHLO dialect, making additional eight e2e tests pass with MHLO config. It also fixed the tanh dtype function for `float16`.


lib/Conversion/TorchToMhlo/Basic.cpp
```diff
@@ -769,6 +769,38 @@ LogicalResult ConvertAtenOp<AtenContiguousOp>::matchAndRewrite(
...
+
+  // The non_blocking should be a constant `False`.
+  bool nonBlocking;
+  if (!matchPattern(op.getNonBlocking(), m_TorchConstantBool(&nonBlocking))) {
+    return rewriter.notifyMatchFailure(
+        op, "unimplemented: non_blocking must be a constant");
+  } else if (nonBlocking) {
+    return rewriter.notifyMatchFailure(
+        op, "unimplemented: non_blocking is expected to be false");
+  }
```
tanyokwok, line 796: Nit: there is no need to have a check here. The memory layout and semantics can be discarded.

lib/Conversion/TorchToMhlo/Basic.cpp
```diff
@@ -769,6 +769,38 @@ LogicalResult ConvertAtenOp<AtenContiguousOp>::matchAndRewrite(
...
+LogicalResult ConvertAtenOp<AtenCopyOp>::matchAndRewrite(
+    AtenCopyOp op, OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+
+  Value self = adaptor.getSelf();
+  Value src = adaptor.getSrc();
+  auto selfType = self.getType().cast<RankedTensorType>();
+  auto srcType = src.getType().cast<RankedTensorType>();
+
+  if (!selfType.hasStaticShape() || !srcType.hasStaticShape()) {
```
tanyokwok, line 783: Can you use dynamic broadcast to support dynamic shape as well?

li-plus: Dynamic broadcast is handled automatically after decomposition

lib/Conversion/TorchToMhlo/Basic.cpp
li-plus: Okay, there is no check in the new version.

li-plus:
Pushed a new version. The `aten.copy` op will be decomposed into `aten.to.dtype` and `aten.expand_as` before lowering to backend dialects. The dynamic broadcasting will be handled when lowering `aten.broadcast_to` op. In this way I got three additional tests passed with MHLO config. Decomposition might be a better way to deal with `aten.copy`.

tanyokwok (approved):
LGTM

PR #1736, opened 2022-12-19
build: update llvm tag to 7ccbb4df

tanyokwok:
Summary of changes:
    
     - LLVM now includes <optional> instead of "llvm/ADT/Optional.h" in most
       (although not all) places
       (https://reviews.llvm.org/rG541ef3d61e9341cd38420c0dbca9250c4d0ea04c).
       This patch replaces the affected instances of `llvm::Optional` with
       `std::optional`.
    
     - In the usages of llvm::Optional that remain, llvm::Optional::value()
       is deprecated, so this patch replaces them with a dereference


silvasean (approved):
Thanks!

PR #1737, opened 2022-12-19
Unrevert #1724

antoniojkim:
https://github.com/llvm/torch-mlir/pull/1724 was reverted in https://github.com/llvm/torch-mlir/pull/1727 because the change in upstream PyTorch also reverted. Seems it was unreverted in upstream PyTorch so unreverting here


ashay (approved):

ashay:
I believe yesterday's WHL files (for both torch and torchvision) should work, since I think we ran into the same RollPyTorch error (undefined symbol `c10::asIntArrayRefSlow`) yesterday.

Issue #1076, opened 2022-07-18
Failed to legalize operation 'torch.aten.linear' that was explicitly marked illegal


jxhekang:
Try to lower torch model to LINALG_ON_TENSORS ,  but met exception below when nn.Linear's out_features is 1. The Exception will not appear if nn.Linear's out_features is greater than 1 !

Lowering Torch Backend IR -> Linalg-on-Tensors Backend IR failed with the following diagnostics:
error: failed to legalize operation 'torch.aten.linear' that was explicitly marked illegal
note: see current operation: %2 = "torch.aten.linear"(%arg0, %0, %1) : (!torch.vtensor<[1,32],f32>, !torch.vtensor<[1,32],f32>, !torch.vtensor<[1],f32>) -> !torch.vtensor<[1,1],f32>

The exception can be reproduced with the code below：

```
import torch
from torch import nn
import torch_mlir

class SimpleModel(nn.Module):
    def __init__(self):
        super(SimpleModel, self).__init__()
        self.linear = nn.Linear(32, 1)
    
    def forward(self, x):
        out = self.linear(x)
        return out


def main():
    model = SimpleModel()

    example_input = torch.rand((1, 32), dtype=torch.float32)

    traced = torch.jit.trace(model, example_input)
    linalg_on_tensors_mlir = torch_mlir.compile(traced, example_input,
                                                output_type=torch_mlir.OutputType.LINALG_ON_TENSORS)

    print(linalg_on_tensors_mlir)


if __name__ == "__main__":
    main()


```

ramiro050:
This is currently failing because there is no support for broadcasting the `bias` tensor:

https://github.com/llvm/torch-mlir/blob/df0b1e77a475cffcd758c99efea7f9ba64ae060a/lib/Conversion/TorchToLinalg/Linear.cpp#L493-L497

@Shukla-Gaurav, do you know if after your patch https://github.com/llvm/torch-mlir/pull/862, this will work?

jxhekang:
> @ramiro050 Yes #862 will solve this issue. I tested it also. Thanks!

Hi,  #862 has not been merged to the main branch, so I try to rebuild your personal repo :https://github.com/Shukla-Gaurav/torch-mlir/tree/gaurav/decompose_linear, but got the error like the pic below:
![image](https://user-images.githubusercontent.com/26559926/180392287-667a5979-8bb5-4905-9e23-d4197d0cabbb.png)

[...]

PR #1740, opened 2022-12-21
Fix decomposition of aten.linear wrt input rank

xgupta:
This fix Failed to legalize operation 'torch.aten.linear' that was explicitly marked illegal #1076. 

As per the comment of @ramiro050  here - https://github.com/llvm/torch-mlir/issues/1076#issuecomment-1255578687 -

To match the [op documentation](https://pytorch.org/docs/stable/generated/torch.nn.functional.linear.html?highlight=linear), which allows any tensor with rank >= 1, change the expected input tensor rank here - https://github.com/llvm/torch-mlir/blob/4ef6e69ed49f309aa70ea40f12bf7488cd8e3434/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L2078.

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -209,6 +209,7 @@ MHLO_PASS_SET = {
     "LayerNormModule_basic",
     "LayerNormNormalizeOverAllDimsModule_basic",
     "LeakyReluBackwardStaticModule_basic",
+    "Linear1DModule_basic",
     "MatmulBroadcastBatchDim_basic",
     "MatmulSingleDynamicBatchDim_basic",
     "Matmul_3d",
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -2490,14 +2490,14 @@ public:
     Value bias = op.getBias();
 
     BaseTensorType inputType = input.getType().cast<BaseTensorType>();
-    if (!inputType.hasSizes() || inputType.getSizes().size() < 2)
+    if (!inputType.hasSizes() || inputType.getSizes().size() < 1)
       return rewriter.notifyMatchFailure(
-          op, "expected input to be rank 2 or greater");
+          op, "expected input to be rank 1 or greater");
 
     BaseTensorType weightType = weight.getType().cast<BaseTensorType>();
-    // `weight` must be a rank 2 matrix.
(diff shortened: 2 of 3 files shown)


RamirezLucas:
I think you mentioned the wrong person :)

xgupta:
> I think you mentioned the wrong person :)

Er, sorry for that :)

ramiro050 (changes requested):
Thanks! Can you also add an e2e test that uses a rank-1 tensor?

xgupta:
> Thanks! Can you also add an e2e test that uses a rank-1 tensor?

Hi @ramiro050, 

I want to know why the weight rank must be 2 here  - https://github.com/llvm/torch-mlir/blob/1cf09a6a34a4b8b412da85f4fa44757ed0d3622f/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L2499 (one thing I can think of because it is decomposed to aten.matmul). because of this test case is failing with the message - 

```
        error: found an op that was marked as backend illegal
        note: see current operation: %1 = "torch.aten.linear"(%arg0, %arg1, %0) : (!torch.vtensor<[?],f32>, !torch.vtensor<[?,?],f32>, !torch.none) -> !torch.vtensor<[?],f32>
        note: this is likely due to DecomposeComplexOps being unable to decompose this op
```

ramiro050:
> > Thanks! Can you also add an e2e test that uses a rank-1 tensor?
> 
> Hi @ramiro050,
> 
> I want to know why the weight rank must be 2 here -
> 
> https://github.com/llvm/torch-mlir/blob/1cf09a6a34a4b8b412da85f4fa44757ed0d3622f/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L2499
> 
> (one thing I can think of because it is decomposed to aten.matmul). because of this test case is failing with the message -
> ```
>         torch_mlir.compiler_utils.TorchMlirCompilerError: Lowering TorchScript IR -> Torch Backend IR failed with the following diagnostics:
>         error: found an op that was marked as backend illegal
>         note: see current operation: %1 = "torch.aten.linear"(%arg0, %arg1, %0) : (!torch.vtensor<[?],f32>, !torch.vtensor<[?,?],f32>, !torch.none) -> !torch.vtensor<[?],f32>
>         note: this is likely due to DecomposeComplexOps being unable to decompose this op
> ```

Must be a similar bug. The documentation says that it can be rank 1 or 2: https://pytorch.org/docs/stable/generated/torch.nn.functional.linear.html?highlight=linear#torch.nn.functional.linear

Issue #1739, opened 2022-12-20
AttributeError: 'torch._C.ScriptMethod' object has no attribute '__func__'


zhping5678:
When I run the bert model example in iree-torch [https://github.com/iree-org/iree-torch/blob/main/examples/bert.py](https://github.com/iree-org/iree-torch/blob/main/examples/bert.py), it reports `AttributeError: 'torch._C.ScriptMethod' object has no attribute '__func__'`. 

I noticed a closed [issue](https://github.com/llvm/torch-mlir/issues/1582#issue-1448224019) which said this bug had been fixed.I have pulled code after that issue, but now it still reports errors.

Code is below.
``` print("Tracing model.")
    traced = torch.jit.trace(model, example_input)
    print("Compiling with Torch-MLIR")
    linalg_on_tensors_mlir = torch_mlir.compile(traced, example_input, output_type=torch_mlir.OutputType.LINALG_ON_TENSORS)
```
It reports below.
```
Traceback (most recent call last):
  File "examples/bert.py", line 115, in <module>
    main()
  File "examples/bert.py", line 96, in main
    linalg_on_tensors_mlir = torch_mlir.compile(traced, example_input,
  File "~/code/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/__init__.py", line 337, in compile
    torch.jit.export(getattr(model, method_name).__func__)

```

ramiro050:
Hi @zhping5678, this seems to be an issue with how `torch_mlir.compile` was handling models already in JIT IR. Can you check if this patch fixes your problem? https://github.com/llvm/torch-mlir/pull/1743

ramiro050:
@zhping5678, I've merged the patch in `main`. Feel free to reopen this if the patch does not fix your problem!

zhping5678:
Thank you very much. It's ok.

PR #1743, opened 2022-12-21
Allow passing traced `torch.nn.Module`s into `torch_mlir.compile`

ramiro050:
This commit adds support for passing to `torch_mlir.compile` the result of running `torch.jit.trace` on a model by relaxing the condition that checks if the model is already in JIT IR to allow any `torch.jit.ScriptModule`.

Fixes https://github.com/llvm/torch-mlir/issues/1739


vivekkhandelwal1 (approved):
LGTM

PR #1744, opened 2022-12-22
Revert "[MHLO] reimplement linear lowering torchToMhlo"

tanyokwok:
Reverts llvm/torch-mlir#1524


vivekkhandelwal1 (approved):
Thanks for reverting the patch.

PR #1745, opened 2022-12-22
[MLIR][TORCH] Add missing ops in RefineTypes.cpp

vivekkhandelwal1:


AmosLewis (approved):
LGTM

AmosLewis:
Related Issue: https://github.com/nod-ai/SHARK/issues/494

vivekkhandelwal1:
> This should not be needed. These ops were removed from here to make sure the custom op code works:
> 
> https://github.com/llvm/torch-mlir/blob/3260a1ea6e885a2b19dbcb391ea22c5cbf742047/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py#L65
> 
> From looking at the issue linked, it seems that the problem is that some places in the code base call `getDtype` without first checking `hasDtype`. The solution should be to add the right checks

Thanks for letting me know.

PR #1747, opened 2022-12-22
Support aten.stack op and decompose it into unsqueeze & cat

li-plus:
This PR added `aten.stack` op, which will be decomposed into multiple `aten.unsqueeze` ops to create a new dimension for input tensors, and then a `aten.cat` op to concatenate the tensor list along the given dimension.


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -973,6 +973,17 @@ def aten〇index〇Tensor_hacked_twin〡shape(self: List[int], indices: List[Lis
...
+def aten〇stack〡shape(tensors: List[List[int]], dim: int = 0) -> List[int]:
+    unsqueezed_tensors: List[List[int]] = []
+    for tensor in tensors:
+        tensor_rank = len(tensor) + 1
+        assert -tensor_rank <= dim < tensor_rank, "Dimension out of range"
+        insert_dim = dim if dim >= 0 else dim + tensor_rank
+        unsqueezed = upstream_shape_functions._copy(tensor)
+        unsqueezed.insert(insert_dim, 1)
+        unsqueezed_tensors.append(unsqueezed)
+    return upstream_shape_functions.cat(unsqueezed_tensors, dim)
```
ramiro050, line 985: Can you upstream this function first? See: https://github.com/pytorch/pytorch/pull/76889

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
li-plus: Okay, will need some time to do this.

ramiro050: That took longer than expected on the PyTorch side, but they finally merged it!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -811,6 +811,65 @@ class DecomposeAtenTOp : public OpRewritePattern<AtenTOp> {
...
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenStackOp op,
+                                PatternRewriter &rewriter) const override {
+    SmallVector<Value> tensors;
+    if (!getListConstructElements(op.getTensors(), tensors)) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: the tensor list is not from list construct");
+    } else if (tensors.empty()) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: the tensor list is empty");
```
ramiro050, line 827: nit: This case actually results in an error in PyTorch, so I would remove the `unimplemented: ` part.

```python
In [2]: torch.stack([], dim=0)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Input In [2], in <cell line: 1>()
----> 1 torch.stack([], dim=0)

RuntimeError: stack expects a non-empty TensorList
```

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -811,6 +811,65 @@ class DecomposeAtenTOp : public OpRewritePattern<AtenTOp> {
...
+    }
+
+    SmallVector<Value> unsqueezedTensors;
+    for (Value tensor : tensors) {
+      BaseTensorType tensorType = tensor.getType().cast<BaseTensorType>();
+
+      if (!tensorType.hasSizes()) {
+        return rewriter.notifyMatchFailure(
+            op, "unimplemented: one tensor does not have known sizes");
+      }
```
ramiro050, line 837: Lets make this check outside the `for` loop to avoid creating IR when we already know things are invalid.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -811,6 +811,65 @@ class DecomposeAtenTOp : public OpRewritePattern<AtenTOp> {
...
+
+      ArrayRef<int64_t> tensorShape = tensorType.getSizes();
+      int64_t dimInt = 0;
+      SmallVector<int64_t> unsqueezedShape;
+      if (matchPattern(op.getDim(), m_TorchConstantInt(&dimInt))) {
```
ramiro050, line 842: This is always the same value for every iteration of the `for` loop. I would extract this out

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -811,6 +811,65 @@ class DecomposeAtenTOp : public OpRewritePattern<AtenTOp> {
...
+          return rewriter.notifyMatchFailure(op, "dimension out of range");
+        }
+        unsqueezedShape.append(tensorShape.begin(), tensorShape.end());
+        unsqueezedShape.insert(unsqueezedShape.begin() + dimInt, 1);
+      } else {
+        unsqueezedShape.resize(tensorShape.size() + 1, kUnknownSize);
+      }
+
+      Type unsqueezedType = tensorType.getWithSizesAndDtype(
+          unsqueezedShape, tensorType.getDtype());
```
ramiro050, line 855: use `getOptionalDtype()`. There is no guarantee when `DecomposeComplexOps` that tensors will have dtypes

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -811,6 +811,65 @@ class DecomposeAtenTOp : public OpRewritePattern<AtenTOp> {
...
+                                PatternRewriter &rewriter) const override {
+    SmallVector<Value> tensors;
+    if (!getListConstructElements(op.getTensors(), tensors)) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: the tensor list is not from list construct");
+    } else if (tensors.empty()) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: the tensor list is empty");
+    }
+
```
ramiro050, line 829: The code towards the end makes the assumption that all the tensors have the same type. Can you add a check here to make sure this is true and return early if it's not?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3447,6 +3506,7 @@ class DecomposeComplexOpsPass
         DecomposeConstantTensorAllocLikeOp<AtenOnesLikeOp, 1>>(patterns);
     addPatternIfTargetOpIsIllegal<
         DecomposeConstantTensorAllocLikeOp<AtenZerosLikeOp, 0>>(patterns);
+    addPatternIfTargetOpIsIllegal<DecomposeAtenStackOp>(patterns);
```
ramiro050, line 3509: You also need to add a similar line here:

https://github.com/llvm/torch-mlir/blob/86b792520bb1331f106f7eae2f38fb18db70c397/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L335

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
li-plus: Done!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
li-plus: Switched `getDtype ` to `getOptionalDtype` in  `unsqueezeTensor`.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
li-plus: Do we need to support tensors of different dtypes in the input list? PyTorch actually supports type promotion on all input tensors of `cat` and `stack` ops.

ramiro050: This case can be left as `unimplemented: ...`. If you do implement this, you just need to be careful when creating the `!torch.list` type for the operand that will be passed to `cat`. If the dtypes are different for the tensors in the list, then the contained type in `!torch.list` should be a type with no `dtype`

li-plus: I do support this. Added a new e2e case `TensorsStackPromoteDTypeModule_basic` to test type promotion.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1055,6 +1055,50 @@ class DecomposeAtenTOp : public OpRewritePattern<AtenTOp> {
...
+      if (failed(unsqueezedInfo)) {
+        return rewriter.notifyMatchFailure(
+            op, "cannot generate unsqueeze tensor op");
+      }
+      unsqueezedTensors.push_back(*unsqueezedInfo);
+    }
+
+    Type listElemType =
+        op.getType().cast<BaseTensorType>().getWithSizesAndDtype(
+            /*optionalSizes=*/{}, /*optionalDtype=*/nullptr);
```
ramiro050, line 1091: `optionalSizes` should be `std::nullopt`. I think `{}` will result in a zero rank tensor

ramiro050 (approved):
LGTM, just one change request

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
li-plus: Done! `{}` is still correct but `std::nullopt` is more explicit.

li-plus: Switched to `std::nullopt` anyway.

ramiro050: Great, thanks!

Issue #1748, opened 2022-12-22
tanh + select.int causing compiler core dumped


li-plus:
In this PR #1594, `AtenTanhOp` has been removed from this type refinement list:

https://github.com/llvm/torch-mlir/blob/52669cbbd5973d537a7d23720822a3b45dd07495/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L675-L689

This is causing some subtle problems. For example, now we cannot compile the below simple model.

```python
import torch
import torch.nn as nn
import torch_mlir

class Model(nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        return torch.tanh(x + 1)[0]

model = Model().eval()
inputs = torch.randn(2, 3)
model = torch.jit.trace(model, inputs)
model.save('model.jit')

model = torch.jit.load('model.jit')
print(model(inputs))
module = torch_mlir.compile(model, inputs, output_type="torch")
print(module)
```

The error log:

```
(thmlir) ➜  ~/ada_local/torch-mlir git:(main) ✗ python3 examples/tanh.py
tensor([0.7073, 0.2370, 0.9675])
python3: /data00/home/lijiahao.plus/ada_local/torch-mlir/include/torch-mlir/Dialect/Torch/IR/TorchTypes.h:77: mlir::Type mlir::torch::Torch::BaseTensorType::getDtype() const: Assertion `hasDtype() && "must have a dtype"' failed.
[1]    2719954 abort (core dumped)  python3 examples/tanh.py
```

Backtrace:

```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f03a2b85535 in __GI_abort () at abort.c:79
#2  0x00007f03a2b8540f in __assert_fail_base (fmt=0x7f03a2ce7ee0 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x7f028ba9efd3 "hasDtype() && \"must have a dtype\"", 
    file=0x7f028ba9eee9 "/data00/home/lijiahao.plus/ada_local/torch-mlir/include/torch-mlir/Dialect/Torch/IR/TorchTypes.h", line=77, 
    function=<optimized out>) at assert.c:92
#3  0x00007f03a2b93102 in __GI___assert_fail (assertion=0x7f028ba9efd3 "hasDtype() && \"must have a dtype\"", 
[...]

PR #1750, opened 2022-12-22
Add `hasDtype` checks everywhere dtypes are used in decompositions

ramiro050:
There are several decompositions that assume the operands of the op have dtypes available; however, the only time dtypes are guaranteed to be present is when the graph has reached the backend contract. In general, every pass that happens before reaching the backend contract should not assume dtypes are available and should use `hasDtype` to check first.

This commit adds `hasDtype` checks to every decomposition that uses dtypes.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -71,7 +71,7 @@ static Type computeReductionType(PatternRewriter &rewriter, Operation *op,
   Type resultType = tensorType.getWithSizesAndDtype(
       sizes.size() == 0 ? std::optional<ArrayRef<int64_t>>()
                         : llvm::makeArrayRef(sizes),
-      tensorType.getDtype());
+      tensorType.getOptionalDtype());
   return resultType;
 }
 
@@ -407,6 +407,11 @@ public:
           op, "Expected a boolean value for half_to_float");
 
     BaseTensorType resultTensorType = op.getType().cast<BaseTensorType>();
+    if (!resultTensorType.hasDtype()) {
+      return rewriter.notifyMatchFailure(
+          op, "expected result type to have a dtype");
+    }
+    Type resultTensorDtype = resultTensorType.getDtype();
     // `torch.ops.aten._softmax`'s softmax with half to float conversion is not
     // supported on CPU, but we go ahead with the decomposing.
     // TODO: Add an e2e test once upstream support is added.
@@ -418,7 +423,7 @@ public:
       Value cstFalse = rewriter.create<ConstantBoolOp>(loc, false);
       self = rewriter.create<AtenToDtypeOp>(
           loc, resultTensorType, self,
-          getDtypeIntValueForType(rewriter, loc, resultTensorType.getDtype()),
+          getDtypeIntValueForType(rewriter, loc, resultTensorDtype),
           /*non_blocking=*/cstFalse, /*copy=*/cstFalse, /*memory_format=*/none);
     }
     Value result = getSoftmaxResult(op, self, resultTensorType, rewriter);
@@ -558,8 +563,8 @@ public:
       return failure();
     BaseTensorType valueTensorType =
         inputType
-            .getWithSizesAndDtype(indicesTensorType.getSizes(),
-                                  inputType.getDtype())
+            .getWithSizesAndDtype(indicesTensorType.getOptionalSizes(),
+                                  inputType.getOptionalDtype())
             .cast<BaseTensorType>();
 
     // If the dim type is `NoneType` i.e. reduce along all the dimensions.
@@ -568,7 +573,9 @@ public:
     // 0th dimension.
     if (dim.getType().isa<Torch::NoneType>()) {
       BaseTensorType flattenType =
-          inputType.getWithSizesAndDtype({kUnknownSize}, inputType.getDtype())
+          inputType
+              .getWithSizesAndDtype({kUnknownSize},
+                                    inputType.getOptionalDtype())
               .cast<BaseTensorType>();
       dim = rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(0));
       Value end = rewriter.create<ConstantIntOp>(
@@ -850,7 +857,7 @@ public:
       sizes.append(inputShape.begin(), inputShape.end());
       sizes[cstDim] = kUnknownSize;
       Type sliceTy = selfTy.getWithSizesAndDtype(llvm::makeArrayRef(sizes),
-                                                 selfTy.getDtype());
+                                                 selfTy.getOptionalDtype());
       Value slice0 = rewriter.create<AtenSliceTensorOp>(
           loc, sliceTy, input, dim, negShift, constNone, constOne);
       Value slice1 = rewriter.create<AtenSliceTensorOp>(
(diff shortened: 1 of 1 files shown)


AmosLewis (approved):
LGTM

vivekkhandelwal1:
Hi @ramiro050, I have a doubt regarding this patch. At some places you have directly used `getOptionalDtype`, while in some places you have used `getDtype`, and thrown an error if not present. What's the rationale behind doing this?

ramiro050:
> Hi @ramiro050, I have a doubt regarding this patch. In some places, you have directly used `getOptionalDtype`, while in some places you have used `getDtype`, and thrown an error if not present. What's the rationale behind doing this?

That's a good question. There are two scenarios that take place regarding the use of `getDtype`: 

1. The first scenario is the `dtype` of a tensor is used to create a new type for another tensor. In this case, it does not actually matter if the `dtype` exists or not. If the current `dtype` is `unk`, then the new tensor will get assigned that `dtype` too, and that is okay to do.
2. The second scenario is the `dtype` is used to determine what IR to generate in the decomposition. For example, if we need to cast a tensor from one type to another, we need to know the actual value of `dtype` to know how to perform the cast correctly.

For the first scenario, I used the `getOptionalDtype` since knowing the `dtype` is not truly necessary. This allows some decompositions that don't depend conditionally on the value of `dtype` to still succeed. For example, there are several decompositions where the result tensor has the same `dtype` as the input tensor, so `dtype`s only come about in the decomposition when creating the type for the result tensor. In this case, it does not matter if `dtype` exists or not for the decomposition to succeed.

The the second scenario, because it is important for `dtype` to have a value, I added error messages when the `dtype` does not exist.

Let me know if this makes sense to you and if the changes look good.

li-plus:
Hi, I have a few more questions.

Why can't we determine the dtype before the decomposition happens? If we can, is it a better way to handle this issue? (like finalizing the dtype  in `RefineTypes`)

In the second scenario, will there be unexpected decomposition errors in some combination of ops, when we can determine the dtype but we didn't? Like `tanh` + `select.dim` in this issue #1748.

ramiro050:
> Hi, I have a few more questions.
> 
> Why can't we determine the dtype before the decomposition happens? If we can, is it a better way to handle this issue? (like finalizing the dtype in `RefineTypes`)

Currently the `dtype` in your example is not being determined before decompose complex ops because we are in the process of migrating from using `RefineTypes` to using the `dtype` functions defined in `abstract_interp_lib_gen.py`. (See: https://github.com/llvm/torch-mlir/blob/main/docs/abstract_interp_lib.md). The order of the passes is:

https://github.com/llvm/torch-mlir/blob/f0998f92dfc0b783a178e2c0250c04d80848130d/lib/Dialect/Torch/Transforms/Passes.cpp#L123-L136

`createTorchDtypeRefinementPipeline` is the set of passes that use the `dtype` functions in `abstract_interp_lib_gen.py`.

Because your example does an `aten.add.Tensor` first, it needs to wait for `RefineTypes` to happen to get its result type. Then `aten.tanh` depends on `createTorchDtypeRefinementPipeline`, which already passed, so no result `dtype` is determined for `aten.tanh`, leaving `aten.squeeze.dim` with no dtypes in its inputs when `DecomposeComplexOps` is reached, leading to the error you're seeing. 

However, these set of passes are meant to be run over and over again to avoid the catch-22 issue outlined above between the two dtype propagation passes (see: https://github.com/llvm/torch-mlir/commit/57681f794764a34c34e2be7f07f7dfbcafa683c1). In your case, all that is needed is for the set of passes to be run one more time. What prevents this from happening is that `DecomposeComplexOps` currently assumes `dtypes` will be available when in reality there is no guarantee of that.

> In the second scenario, will there be unexpected decomposition errors in some combination of ops, when we can determine the dtype but we didn't? Like `tanh` + `select.dim` in this issue #1748.

My patch fixes the example in your issue. I used your example to make sure things worked locally.

AmosLewis:
I try cherry pick this patch for the [gpt2 to tosa](https://github.com/nod-ai/SHARK/issues/338). But it fails.
Here is the [torchscipt ir](https://gist.github.com/AmosLewis/d5ca2b2d4fe3aee29add60db32a8b987).
Please run torchscript IR to torch the backend pipeline over that, and it will fail.

AmosLewis (changes requested):
I try to cherry pick this patch for the https://github.com/nod-ai/SHARK/issues/338. But it fails.
Here is the [torchscipt ir](https://gist.github.com/AmosLewis/d5ca2b2d4fe3aee29add60db32a8b987).
Please run torchscript IR to torch the backend pipeline over that, and it will fail.
`torch-mlir-opt -pass-pipeline='builtin.module(torchscript-module-to-torch-backend-pipeline{backend-legal-ops=torch.aten.flatten.using_ints})' /tmp/gpt2_torch_raw_elide.mlir --mlir-print-ir-after-all > gpt2_tosa_ramiro.mlir `
You will got this after refine type
` %884 = torch.aten.tanh %883 : !torch.vtensor<[1,5,3072],f32> -> !torch.vtensor<[1,5,3072],unk>`

ramiro050:
> I try to cherry pick this patch for the [nod-ai/SHARK#338](https://github.com/nod-ai/SHARK/issues/338). But it fails. Here is the [torchscipt ir](https://gist.github.com/AmosLewis/d5ca2b2d4fe3aee29add60db32a8b987). Please run torchscript IR to torch the backend pipeline over that, and it will fail. `torch-mlir-opt -pass-pipeline='builtin.module(torchscript-module-to-torch-backend-pipeline{backend-legal-ops=torch.aten.flatten.using_ints})' /tmp/gpt2_torch_raw_elide.mlir --mlir-print-ir-after-all > gpt2_tosa_ramiro.mlir ` You will get this after refine type ` %884 = torch.aten.tanh %883 : !torch.vtensor<[1,5,3072],f32> -> !torch.vtensor<[1,5,3072],unk>`

Please take a look at https://github.com/llvm/torch-mlir/pull/1769, and let me know if that fixes it

ramiro050:
@AmosLewis, while https://github.com/llvm/torch-mlir/pull/1769 might fix the issue you're having with `tanh`, this PR also makes changes that are important to have to make decompositions more robust. If things look good to you here, I can go ahead and merge this.

Note: the commit message in https://github.com/llvm/torch-mlir/pull/1769 explains why things still fail in GPT-2 when using this patch.

AmosLewis (approved):

AmosLewis:
> @AmosLewis, while #1769 might fix the issue you're having with `tanh`, this PR also makes changes that are important to have to make decompositions more robust. If things look good to you here, I can go ahead and merge this.
> 
> Note: the commit message in #1769 explains why things still fail in GPT-2 when using this patch.

I think @vivekkhandelwal1  also mentioned there is a similar thing for **AtenRsubScalarOp**

ramiro050:
> The `tanh` works. I think @vivekkhandelwal1 also mentioned there is a similar refinetype bug for `AtenRsubScalarOp` he worked on.

@vivekkhandelwal1, let me know if this PR and https://github.com/llvm/torch-mlir/pull/1769 don't fix your issue.

vivekkhandelwal1:
> > The `tanh` works. I think @vivekkhandelwal1 also mentioned there is a similar refinetype bug for `AtenRsubScalarOp` he worked on.
> 
> @vivekkhandelwal1, let me know if this PR and #1769 don't fix your issue.

Hi @ramiro050, the issue with `AtenRsubScalarOp` still exists. I have to add it to the RefineTypes.cpp to work.

ramiro050:
> Hi @ramiro050, the issue with `AtenRsubScalarOp` still exists. I have to add it to the RefineTypes.cpp to work.

I think the best way forward is to finish the transition to using dtype functions written in Python. I have made an issue to track the progress: https://github.com/llvm/torch-mlir/issues/1807 This should once and for all fix these issues.

vivekkhandelwal1:
> > Hi @ramiro050, the issue with `AtenRsubScalarOp` still exists. I have to add it to the RefineTypes.cpp to work.
> 
> I think the best way forward is to finish the transition to using dtype functions written in Python. I have made an issue to track the progress: #1807 This should once and for all fix these issues.

Thanks!

PR #1756, opened 2022-12-30
Bump to Python 3.8

powderluv:


ashay:
Sadly, Python v3.8 is going to fail the Release build as well, since the nightly releases for torchvision for Linux are only available for Python 3.10.

ashay (approved):

Issue #1758, opened 2022-12-30
Failing e2e tests in Linalg backend


ashay:
The `ElementwiseAtenLogicalAndOpPromoteBroadcastModule_basic` and `ElementwiseAtenLogicalXorOpPromoteBroadcastModule_basic` seem to non-deterministically fail in the post-merge builds.

Sample log from [CI](https://github.com/llvm/torch-mlir/actions/runs/3801245116/jobs/6465503034#step:7:5793):

```
2022-12-29T14:44:56.9009809Z ****** Failed tests - 2 tests
2022-12-29T14:44:56.9010495Z     FAIL - "ElementwiseAtenLogicalAndOpPromoteBroadcastModule_basic"
2022-12-29T14:44:56.9011546Z         @ trace item #0 - call to "forward"
2022-12-29T14:44:56.9012123Z         @ output of call to "forward"
2022-12-29T14:44:56.9012904Z         ERROR: value (Tensor with shape=[4, 5], dtype=torch.bool, min=+0.0, max=+1.0, mean=+0.65) is not close to golden value (Tensor with shape=[4, 5], dtype=torch.bool, min=+0.0, max=+1.0, mean=+0.8)
2022-12-29T14:44:56.9013378Z 
2022-12-29T14:44:56.9013896Z     FAIL - "ElementwiseAtenLogicalXorOpPromoteBroadcastModule_basic"
2022-12-29T14:44:56.9014513Z         @ trace item #0 - call to "forward"
2022-12-29T14:44:56.9014931Z         @ output of call to "forward"
2022-12-29T14:44:56.9015648Z         ERROR: value (Tensor with shape=[4, 5], dtype=torch.bool, min=+0.0, max=+1.0, mean=+0.4) is not close to golden value (Tensor with shape=[4, 5], dtype=torch.bool, min=+0.0, max=+1.0, mean=+0.2)
```

PR https://github.com/llvm/torch-mlir/pull/1757 reverts the commit that introduced these tests.

ashay:
Resolved

PR #1752, opened 2022-12-23
Add e2e support for aten logical or/and/xor/not ops

li-plus:
This PR supported lowering `AtenLogicalOrOp` to MHLO, and added `AtenLogicalAndOp`, `AtenLogicalXorOp` and `AtenLogicalNotOp` for both LINALG and MHLO backends.

PR #1761, opened 2022-12-30
[Resubmit] Add e2e support for aten logical or/and/xor/not ops

li-plus:
This PR fixed the e2e testing failure mentioned in issue #1758 regarding to PR #1752.

The e2e test added in #1752 triggered the error because the model was compiled with `dtype=float64` but fed with a random tensor with `dtype=float32`. It might cause precision error if the random values are close to zero when converting to boolean.

In this PR the model is compiled with `dtype=float32` to be consistent with input data. The random float tensor shape also expanded from `[5]` to `[256]` to minimize non-determinacy. If this CI has passed, it is unlikely that the future CI will fail.

@ashay @ramiro050 Would you review this PR?


li-plus:
@ramiro050 I also added a static-shape e2e test for each new op that will pass in MHLO backend.

ashay (approved):
Thanks for resolving the problem quickly!  LGTM, but let's wait for Ramiro's review before merging this.

ramiro050 (approved):
Good catch on the `float64` issue, and thanks for adding the tests that check the MHLO lowering!

li-plus:
Would you mind merging this PR for me? I still don't have write access of this repo.

PR #1499, opened 2022-10-17
[TM_TENSOR] Add `aten.scatter.[src|value]` op

Shukla-Gaurav:
This commit adds support of `aten.scatter.src` and `aten.scatter.value`
ops.

Signed-Off-by: Gaurav Shukla <gaurav@nod-labs.com>

PR #1762, opened 2022-12-30
[MLIR][TORCH] Support 2-d index for aten.index_put_impl op

vivekkhandelwal1:
Signed-Off By: Vivek Khandelwa<vivek@nod-labs.com>

### DIFF
--- a/lib/Conversion/TorchToArith/TorchToArith.cpp
+++ b/lib/Conversion/TorchToArith/TorchToArith.cpp
@@ -296,7 +296,7 @@ public:
 } // namespace
 
 namespace {
-template <typename OpTy>
+template <typename OpTy, typename BinOp>
 class ConvertAtenAnyOrAllBoolOp : public OpConversionPattern<OpTy> {
 public:
   using OpConversionPattern<OpTy>::OpConversionPattern;
@@ -305,38 +305,37 @@ public:
   LogicalResult
   matchAndRewrite(OpTy op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-
+    Location loc = op.getLoc();
+    Value result;
     SmallVector<Value> inputListTorchBool;
     if (!getListConstructElements(op.getSelf(), inputListTorchBool)) {
       return rewriter.notifyMatchFailure(
-          op, "Unimplemented input list not constructed from ListConstruct");
+          op, "unimplemented: input list not constructed from ListConstruct");
     }
-    SmallVector<bool> inputListBool;
-    for (Value v : inputListTorchBool) {
-      bool cst;
-      if (!matchPattern(v, m_TorchConstantBool(&cst)))
-        return rewriter.notifyMatchFailure(
-            op, "only support constant bool input list elements");
-      inputListBool.push_back(cst);
-    }
-    bool result = reductionFunction(inputListBool);
-
-    rewriter.replaceOpWithNewOp<arith::ConstantOp>(
-        op, rewriter.getBoolAttr(result));
+    SmallVector<Value> inputList = getTypeConvertedValues(
+        rewriter, loc, this->getTypeConverter(), inputListTorchBool);
+    result = inputList[0];
+    for (unsigned i = 1; i < inputList.size(); i++)
+      result = rewriter.create<BinOp>(loc, result, inputList[i]);
+    rewriter.replaceOp(op, result);
     return success();
   }
 };
 
-class ConvertAtenAnyOp : public ConvertAtenAnyOrAllBoolOp<AtenAnyBoolOp> {
-  using ConvertAtenAnyOrAllBoolOp<AtenAnyBoolOp>::ConvertAtenAnyOrAllBoolOp;
+class ConvertAtenAnyOp
+    : public ConvertAtenAnyOrAllBoolOp<AtenAnyBoolOp, arith::OrIOp> {
+  using ConvertAtenAnyOrAllBoolOp<AtenAnyBoolOp,
+                                  arith::OrIOp>::ConvertAtenAnyOrAllBoolOp;
   bool reductionFunction(ArrayRef<bool> inputArray) const override {
     return llvm::any_of(inputArray,
                         [](bool inputListElem) { return inputListElem; });
   }
 };
 
-class ConvertAtenAllOp : public ConvertAtenAnyOrAllBoolOp<AtenAllBoolOp> {
-  using ConvertAtenAnyOrAllBoolOp<AtenAllBoolOp>::ConvertAtenAnyOrAllBoolOp;
+class ConvertAtenAllOp
+    : public ConvertAtenAnyOrAllBoolOp<AtenAllBoolOp, arith::AndIOp> {
+  using ConvertAtenAnyOrAllBoolOp<AtenAllBoolOp,
+                                  arith::AndIOp>::ConvertAtenAnyOrAllBoolOp;
   bool reductionFunction(ArrayRef<bool> inputArray) const override {
     return llvm::all_of(inputArray,
                         [](bool inputListElem) { return inputListElem; });
@@ -467,6 +466,9 @@ public:
         typeConverter, context);
     target.addIllegalOp<AtenFloordivIntOp>();
     patterns.add<ConvertAtenBinaryOp<AtenFloordivIntOp, arith::FloorDivSIOp>>(
+        typeConverter, context);
+    target.addIllegalOp<PrimMaxIntOp>();
+    patterns.add<ConvertAtenBinaryOp<PrimMaxIntOp, arith::MaxSIOp>>(
         typeConverter, context);
     target.addIllegalOp<AtenCeilFloatOp>();
     patterns
--- a/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
+++ b/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
@@ -360,6 +360,234 @@ public:
 };
 } // namespace
 
+//     """Create a map from each dimension of the input tensor to the
+//     subspace that dimension corresponds to in the result shape one gets
+//     from indexing the tensor with the optional index tensors.
+//
+//     Note: Index tensors are first broadcasted to a common shape before
+//     creating the mapping. So the index of every index tensor will map to
+//     the same dimensions in the result shape.
+//
+//     For example:
+//     indices = [None, None, torch.randint(4, (6, 1)), torch.randint(5, (7,))]
+//     indexBroadcastShapeValue = [6, 7]
+//     map = {0: [0], 1: [1], 2: [2, 3], 3: [2, 3]}
+static SmallVector<SmallVector<int64_t>>
+getInputShapeToOutputShapeMap(SmallVector<Value> optionalIndices,
+                              SmallVector<Value> indexBroadcastShapeValue) {
+  SmallVector<Value> indices;
+  for (Value index : optionalIndices) {
+    if (!index.getType().isa<Torch::NoneType>())
+      indices.push_back(index);
+  }
+
+  unsigned broadcastRank = indexBroadcastShapeValue.size();
+  unsigned numIndexTensors = indices.size();
+  int64_t indexOfFirstIndexTensor = -1;
+  SmallVector<SmallVector<int64_t>> result;
+
+  for (unsigned i = 0; i < optionalIndices.size(); i++) {
+    if (optionalIndices[i].getType().isa<Torch::NoneType>()) {
+      unsigned val = i;
+      if (indexOfFirstIndexTensor >= 0)
+        val += broadcastRank - numIndexTensors;
+      result.push_back({val});
+    } else {
+      if (indexOfFirstIndexTensor < 0)
+        indexOfFirstIndexTensor = i;
+      SmallVector<int64_t> outputIndices;
+      for (unsigned j = indexOfFirstIndexTensor;
+           j < (indexOfFirstIndexTensor + broadcastRank); j++)
+        outputIndices.push_back(j);
+      result.push_back(outputIndices);
+    }
+  }
+  return result;
+}
+
+static std::tuple<SmallVector<Value>, SmallVector<int64_t>>
+getIndicesFinalShape(ConversionPatternRewriter &rewriter, Location loc,
+                     Value input, SmallVector<Value> optionalIndices,
+                     SmallVector<int64_t> inputShapeInt,
+                     SmallVector<Value> inputShapeValue,
+                     SmallVector<int64_t> indexBroadcastShapeInt,
+                     SmallVector<Value> indexBroadcastShapeValue) {
+  SmallVector<Value> result;
+  SmallVector<int64_t> resultInt;
+  bool handledIndexTensorSpace = false;
+
+  for (unsigned i = 0; i < inputShapeValue.size(); i++) {
+    if (optionalIndices[i].getType().isa<Torch::NoneType>()) {
+      result.push_back(inputShapeValue[i]);
+      resultInt.push_back(inputShapeInt[i]);
+    } else {
+      if (!handledIndexTensorSpace) {
+        handledIndexTensorSpace = true;
+        for (unsigned j = 0; j < indexBroadcastShapeValue.size(); j++) {
+          result.push_back(indexBroadcastShapeValue[j]);
+          resultInt.push_back(indexBroadcastShapeInt[j]);
+        }
+      }
+    }
+  }
+  return std::make_tuple(result, resultInt);
+}
+
+static FailureOr<Value>
+getScatterIndices(Aten_IndexPutImplOp op, ConversionPatternRewriter &rewriter,
+                  Type indicesDtype, SmallVector<Value> optionalIndices,
+                  SmallVector<int64_t> indexBroadcastShapeInt,
+                  SmallVector<Value> indexBroadcastShapeValue) {
+  Location loc = op.getLoc();
+  MLIRContext *context = op->getContext();
+  Value input = op.getSelf();
+
+  SmallVector<SmallVector<int64_t>> shapeMap =
+      getInputShapeToOutputShapeMap(optionalIndices, indexBroadcastShapeValue);
+
+  SmallVector<int64_t> inputShapeInt{
+      input.getType().cast<BaseTensorType>().getSizes()};
+  int64_t inputRank = inputShapeInt.size();
+  SmallVector<Value> inputShapeValue;
+  for (unsigned i = 0; i < inputShapeInt.size(); i++) {
+    Value dim = rewriter.create<Torch::ConstantIntOp>(
+        loc, rewriter.getI64IntegerAttr(i));
+    inputShapeValue.push_back(
+        rewriter.createOrFold<AtenSizeIntOp>(loc, input, dim));
+  }
+
+  auto finalShapeResult = getIndicesFinalShape(
+      rewriter, loc, input, optionalIndices, inputShapeInt, inputShapeValue,
+      indexBroadcastShapeInt, indexBroadcastShapeValue);
+  SmallVector<Value> finalShapeValue = std::get<0>(finalShapeResult);
+  SmallVector<int64_t> finalShapeInt = std::get<1>(finalShapeResult);
+
+  Value torchCstNone = rewriter.create<Torch::ConstantNoneOp>(loc);
+  Value torchCstZero =
+      rewriter.create<Torch::ConstantIntOp>(loc, rewriter.getI64IntegerAttr(0));
+  Value torchCstOne =
+      rewriter.create<Torch::ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
+
+  Value indexBroadcastShapeTorchList = rewriter.create<PrimListConstructOp>(
+      loc, Torch::ListType::get(Torch::IntType::get(context)),
+      indexBroadcastShapeValue);
+
+  // Calculating index count.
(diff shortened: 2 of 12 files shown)


lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -275,14 +277,63 @@ class ConvertAten_IndexPutImplOp
...
+    if (*indexTensorRank == 2) {
+      auto indexTensorType = indexTensor.getType().cast<BaseTensorType>();
+      SmallVector<int64_t> indexTensorShape(indexTensorType.getSizes());
+      if (indexTensorShape[0] == 1) {
+        ValueTensorType collapsedIndexTensorType = ValueTensorType::get(
+            context, llvm::makeArrayRef({indexTensorShape[1]}),
+            indexTensorType.getDtype());
+        Value torchCstZero = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(0));
+        indexTensor = rewriter.create<AtenSqueezeDimOp>(
```
ramiro050, line 299: I think we can just lower to `AtenFlattenUsingIntsOp`, and this way the implementation works for all 2-d index inputs, as well as higher ranked tensors

ramiro050 (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -275,14 +277,66 @@ class ConvertAten_IndexPutImplOp
...
+      }
+    } else if (*indexTensorRank > 2) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: index tensor with rank > 2 is not supported");
+    }
+
+    if (inputRank != valuesRank) {
+      auto valueTensorType = op.getValues().getType().cast<BaseTensorType>();
+      SmallVector<int64_t> valueTensorShape(valueTensorType.getSizes());
+      if (valuesRank == (inputRank + 1) && valueTensorShape[0] == 1) {
```
ramiro050, line 317: This is also very restrictive. I would restrict things to the following case:

```python
if index_tensor.rank() > 1:
    if values.rank() != index_tensor.rank() + input.rank() - 1:
        return matchFailure("...")
    collapsed_index = torch.flatten(index_tensor)
    collapsed_values = torch.flatten(values, end_dim=index_tensor.rank() - 1)
```

This should handle any `index_tensor` and `values` tensor as long as `values.rank() == index_tensor.rank() + input.rank() - 1` is satisfied

python/torch_mlir_e2e_test/test_suite/index_put.py
```diff
@@ -818,3 +818,28 @@ def forward(self, input, index, value):
...
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([4, 7], torch.float32, True),
+        ([1, 3], torch.int64, True),
+        ([1, 3, 7], torch.float32, True),
```
ramiro050, line 833: If you use the less restrictive approach, you should no longer need to have static shapes here

ramiro050 (changes requested):

vivekkhandelwal1:
Hi @ramiro050, I need to support the following case for `aten._index_put_impl` op:
```
>>> a = torch.randn((1, 1, 2, 2))
>>> a
tensor([[[[ 1.4184, -1.8768],
          [-1.9334,  0.6857]]]])
>>> updates = torch.randn((1,1,4,4))
>>> updates
tensor([[[[ 1.2037,  1.7812, -0.2039,  0.3809],
          [-0.6564, -0.4894,  0.8012,  0.9277],
          [-1.2519, -1.1837,  1.0080, -0.3857],
          [ 1.8249, -2.4080,  0.0611,  0.7709]]]])
>>> index1 = torch.randint(high=2, size=(4, 1))
>>> index1
tensor([[0],
        [1],
        [1],
        [0]])
>>> index2 = torch.randint(high=2, size=(4,))
>>> index2
tensor([0, 0, 1, 1])
>>> result = torch.ops.aten._index_put_impl(a, (None, None, index1, index2), updates, accumulate=True, unsafe=False)
>>> result
tensor([[[[ 3.8203, -0.8679],
          [-5.5148,  3.0369]]]])

```
Do you think with the `tm_tensor.scatter` op, it would be possible? Since the `tm_tensor.scatter` op has many constraints of its own.

ramiro050:
This should be doable with the `tm_tensor.scatter` op, although it is going to be a very tricky lowering. You need to transform the index tensors into the single index tensor that `tm_tensor.scatter` expects. An example of how this might be done is:

```python
import torch

input = torch.rand((2, 3, 4, 5))
index_dim2 = torch.randint(4, (6, 1))
index_dim3 = torch.randint(5, (7,))

# For the case of `torch.index_put(input, (None, None, index1, index2), ...)`
# TODO: compute this using numpy rules
# See: https://numpy.org/doc/stable/user/basics.indexing.html#combining-advanced-and-basic-indexing
final_shape = (input.size(0), input.size(1), 6, 7)

index_dim0_broadcast = torch.arange(0, input.size(0))\
                            .reshape(input.size(0), 1, 1, 1)\
                            .broadcast_to(final_shape)\
                            .flatten()
index_dim1_broadcast = torch.arange(0, input.size(1))\
                            .reshape(input.size(1), 1, 1)\
                            .broadcast_to(final_shape)\
                            .flatten()
index_dim2_broadcast = index_dim2.broadcast_to(final_shape).flatten()
index_dim3_broadcast = index_dim3.broadcast_to(final_shape).flatten()

indices = torch.stack([index_dim0_broadcast, index_dim1_broadcast,
                       index_dim2_broadcast, index_dim3_broadcast], dim=0).t()

print(indices.shape)
print(index_dim2)
print(index_dim3)
for index in indices:
    print(index)
```

This will create a tensor of size `(2 * 3 * 6 * 7, 4)`. You can think of it as `2 * 3 * 6 * 7 = 252` indices. Each index is `4` elements long because our input tensor is rank-4. @Shukla-Gaurav is a good person to consult on this since he had to do almost the same thing for `torch.scatter` (https://github.com/llvm/torch-mlir/pull/1499). I think we have support for most of the torch ops used in my example above, and in the `TTorchToMTensor` pass you're allowed to create more PyTorch ops, so hopefully this approach is not too bad.

Once you have the index tensor, you just need to flatten the `updates` tensor, and pass that to `tm_tensor.scatter`.

To determine the `final_shape` value in the example above, you need to be familiar with the indexing rules of numpy. [This link](https://numpy.org/doc/stable/user/basics.indexing.html#combining-advanced-and-basic-indexing) has information on it. In particular, this section:

> Let x.shape be (10, 20, 30, 40, 50) and suppose ind_1 and ind_2 can be broadcast to the shape (2, 3, 4). Then x[:, ind_1, ind_2] has shape (10, 2, 3, 4, 40, 50) because the (20, 30)-shaped subspace from X has been replaced with the (2, 3, 4) subspace from the indices. However, x[:, ind_1, :, ind_2] has shape (2, 3, 4, 10, 30, 50) because there is no unambiguous place to drop in the indexing subspace, thus it is tacked-on to the beginning.

I would reduce the problem to only handling the case where the index tensors are along consecutive dimensions, since this is the case you need for your example.

vivekkhandelwal1:
Thanks @ramiro050 for such a detailed response!

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -265,42 +268,256 @@ class ConvertAten_IndexPutImplOp
...
+      assert(indexTensorType.getSizes().size() == 1 &&
+             "expected index tensor to be of rank 1");
+      int64_t indexTensorSize = indexTensorType.getSizes()[0];
+      SmallVector<int64_t> expandedIndexTensorSizes{indexTensorSize, 1};
+      ValueTensorType expandedIndexTensorType = ValueTensorType::get(
+          context, llvm::ArrayRef(expandedIndexTensorSizes),
+          indexTensorType.getDtype());
+      indexTensor = rewriter.create<AtenUnsqueezeOp>(
+          loc, expandedIndexTensorType, indexTensor, torchCstOne);
+    } else if (indicesList.size() > 1 && indicesList.size() == 4) {
```
ramiro050, line 320: This is too restrictive. I don't think it would require much more to have full support  of 2 index tensors. We should have a helper function that computes the broadcasted size of the index tensors (not sure if this already exists in torch-mlir) to then compute `final_shape`. Then it should be a matter of turning every `None_i` in the index list into a `torch.arange(0, input.size(i)).reshape(1, ..., input.size(i), ..., 1, 1).flatten()`.

ramiro050 (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
vivekkhandelwal1: Hi @ramiro050, could you please review this patch now?

ramiro050: Sure! I'm a bit busy right now, but I will for sure review it tomorrow.

vivekkhandelwal1: Thanks!

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -410,42 +413,306 @@ class ConvertAten_IndexPutImplOp
...
     Value torchCstOne = rewriter.create<Torch::ConstantIntOp>(
         loc, rewriter.getI64IntegerAttr(1));
-    Value expandedIndexTensor = rewriter.create<AtenUnsqueezeOp>(
-        loc, expandedIndexTensorType, indexTensor, torchCstOne);
+    Value torchCstNone = rewriter.create<Torch::ConstantNoneOp>(loc);
+    Value indexTensor, flattenedValuesTensorLastDim;
+    BaseTensorType indexTensorType;
+    bool needToFlattenValuesTensor = false;
+    SmallVector<int64_t> flattenedValuesTensorSize;
+    if (indicesList.size() == 1) {
```
ramiro050, line 425: Why do we need to have a special case for 1 index tensor? I would've expected the logic for 2 index tensors to work for 1. Moreover, using the code for 2 index tensors would allow removing the restrictions on `indexTensorRank`

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -410,42 +413,306 @@ class ConvertAten_IndexPutImplOp
...
+             "expected index tensor to be of rank 1");
+      int64_t indexTensorSize = indexTensorType.getSizes()[0];
+      SmallVector<int64_t> expandedIndexTensorSizes{indexTensorSize, 1};
+      ValueTensorType expandedIndexTensorType = ValueTensorType::get(
+          context, llvm::ArrayRef(expandedIndexTensorSizes),
+          indexTensorType.getDtype());
+      indexTensor = rewriter.create<AtenUnsqueezeOp>(
+          loc, expandedIndexTensorType, indexTensor, torchCstOne);
+    } else if (indicesList.size() > 1 &&
+               (int64_t)indicesList.size() == inputRank) {
```
ramiro050, line 466: This limitation is not needed. You just need to pad the `indicesList` with `None`s

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -410,42 +413,306 @@ class ConvertAten_IndexPutImplOp
...
+      // Support the cases where index list size is equal to the rank of the
+      // input tensor. Also, the index list should have atmost two non none
+      // index tensors along the consecutive dimension.
+
+      // This implementation assumes input and index tensors to be statically
+      // shaped.
```
ramiro050, line 472: We should be able to determine the broadcasted shape using dynamic dimensions. We just need to create runtime asserts for dynamic dimensions and create `aten.size.int` ops to get the size of the dynamic dimension.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -410,42 +413,306 @@ class ConvertAten_IndexPutImplOp
...
+      //                       .flatten()
+      // index_dim2_broadcast = index_dim2.broadcast_to(final_shape).flatten()
+      // index_dim3_broadcast = index_dim3.broadcast_to(final_shape).flatten()
+
+      // indices = torch.stack([index_dim0_broadcast, index_dim1_broadcast,
+      //                        index_dim2_broadcast, index_dim3_broadcast],
+      //                        dim=0).t()
+
+      Type indicesDtype =
+          indicesList[2].getType().cast<BaseTensorType>().getDtype();
```
ramiro050, line 578: Why 2? I don't think `indicesList` is guaranteed to have size 3 or larger

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -410,42 +413,306 @@ class ConvertAten_IndexPutImplOp
...
+          loc, transposedIndicesType, indexTensor, torchCstZero, torchCstOne);
+
+      // Flatten values tensor.
+      flattenedValuesTensorSize.push_back(indexCount);
+      needToFlattenValuesTensor = true;
+      flattenedValuesTensorLastDim = inputLastDim;
+    } else {
+      return rewriter.notifyMatchFailure(op,
+                                         "Indices list must not be empty and "
+                                         "its size could be either 1 or 4.");
```
ramiro050, line 667: This is not right

ramiro050 (changes requested):
A few more comments

vivekkhandelwal1:
Hi @ramiro050, I have made the changes as per your comments. Right now, the changes work only when the shapes are static. I am working on adding support when the shapes are dynamic. That might require changes in some other op' lowering as well since the implementation creates view-like ops. But, there's one more issue, if I remove the code for handling cases with only 1 index tensor differently, the implementation is failing for the cases like this:
```
class IndexPutImpl2DIndexModule(torch.nn.Module):

    def __init__(self):
        super().__init__()

    @export
    @annotate_args([
        None,
        ([4, 7], torch.float32, True),
        ([2, 3], torch.int64, True),
        ([2, 3, 7], torch.float32, True),
    ])
    def forward(self, input, index, value):
        return torch.ops.aten._index_put_impl_(input, (index, ),
                                               value,
                                               accumulate=True,
                                               unsafe=False)


@register_test_case(
    module_factory=lambda: IndexPutImpl2DIndexModule())
def IndexPutImpl2DIndexModule_basic(module, tu: TestUtils):
    module.forward(tu.rand(4, 7), tu.randint(2, 3, high=3), tu.rand(2, 3, 7))
```
And, I need this case to be supported. Do you think the current generalized implementation can somehow handle this case as well?

vivekkhandelwal1:
> Hi @ramiro050, I have made the changes as per your comments. Right now, the changes work only when the shapes are static. I am working on adding support when the shapes are dynamic. That might require changes in some other op' lowering as well since the implementation creates view-like ops. But, there's one more issue, if I remove the code for handling cases with only 1 index tensor differently, the implementation is failing for the cases like this:
> 
> ```
> class IndexPutImpl2DIndexModule(torch.nn.Module):
> 
>     def __init__(self):
>         super().__init__()
> 
>     @export
>     @annotate_args([
>         None,
>         ([4, 7], torch.float32, True),
>         ([2, 3], torch.int64, True),
>         ([2, 3, 7], torch.float32, True),
>     ])
>     def forward(self, input, index, value):
>         return torch.ops.aten._index_put_impl_(input, (index, ),
>                                                value,
>                                                accumulate=True,
>                                                unsafe=False)
> 
> 
> @register_test_case(
>     module_factory=lambda: IndexPutImpl2DIndexModule())
> def IndexPutImpl2DIndexModule_basic(module, tu: TestUtils):
>     module.forward(tu.rand(4, 7), tu.randint(2, 3, high=3), tu.rand(2, 3, 7))
> ```
> 
> And, I need this case to be supported. Do you think the current generalized implementation can somehow handle this case as well?

Hi @ramiro050, I have addressed all the issues. But in the process of making it generalized it has started unsupporting this particular case, as I have mentioned in the above comment. Do you have any idea regarding how this can be handled? Since we don't want to handle this as a separate case.

ramiro050:
> Do you have any idea regarding how this can be handled? Since we don't want to handle this as a separate case.

Let me take a look.

ramiro050 (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -409,43 +412,327 @@ class ConvertAten_IndexPutImplOp
...
+            indexBroadcastShapeValue.push_back(
+                rewriter.create<Torch::ConstantIntOp>(
+                    loc, rewriter.getI64IntegerAttr(indexBroadcastShape[i])));
+          }
+        }
+      }
+
+      // Check for broadcast shape size equal to number of non-none index
+      // tensors.
+      if (indexBroadcastShape.size() != nonNoneIndexTensorDim.size()) {
```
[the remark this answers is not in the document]
vivekkhandelwal1: Thanks for the response! This implementation covers all the required cases.

lib/Conversion/Utils/Utils.cpp
```diff
@@ -347,6 +347,100 @@ Value toPositiveValidDim(ConversionPatternRewriter &rewriter, Location loc,
...
+  // Check whether the shapes of the tensors are broadcastable or not.
+  // Two tensors are “broadcastable” if the following rules hold:
+  // 1.) Each tensor has at least one dimension.
+  // 2.) When iterating over the dimension sizes, starting at the trailing
+  // dimension, the dimension sizes must either be equal, one of them is 1, or
+  // one of them does not exist.
+  if (rankA == 0 && rankB == 0) {
+    // Do nothing.
+    return success();
+  } else if (rankA && rankB) {
```
ramiro050, line 373: I think the code in this section works regardless of whether some of the inputs are zero rank or not. No need for the `if-else` statements

lib/Conversion/TorchToArith/TorchToArith.cpp
```diff
@@ -305,38 +305,38 @@ class ConvertAtenAnyOrAllBoolOp : public OpConversionPattern<OpTy> {
   LogicalResult
   matchAndRewrite(OpTy op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-
+    Location loc = op.getLoc();
+    SmallVector<bool> inputListBool;
```
ramiro050, line 309: nit: unused

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -360,6 +360,222 @@ class ConvertAtenBincountOp : public OpConversionPattern<AtenBincountOp> {
 };
 } // namespace
 
+static SmallVector<SmallVector<int64_t>>
```
ramiro050, line 363: Nit: can you add a small documentation string to this function? Without the context of the Python script, it's not very obvious what this function does

lib/Conversion/Utils/Utils.cpp
```diff
@@ -347,6 +347,100 @@ Value toPositiveValidDim(ConversionPatternRewriter &rewriter, Location loc,
 
   return castIntToIndex(rewriter, loc, boundedByDimSize);
 }
+
+// Checks whether the `shapeA` and `shapeB` are broadcast compatible or not. If
+// yes, then computes the final broadcast shape.
+LogicalResult computeBroadcastShape(ConversionPatternRewriter &rewriter,
```
ramiro050, line 353: LogicalResult return type no longer needed. All paths lead to `success`

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -401,51 +623,211 @@ class ConvertAten_IndexPutImplOp
...
+    Value torchCstNone = rewriter.create<Torch::ConstantNoneOp>(loc);
+    unsigned sizeOptionalIndicesList = optionalIndicesList.size();
+    SmallVector<int64_t> nonNoneIndexTensorDim;
+    unsigned numNonNoneIndices;
 
-    if (indexTensor.getType().isa<Torch::NoneType>())
-      return rewriter.notifyMatchFailure(op, "Index tensor must not be None.");
+    if (sizeOptionalIndicesList == 0) {
+      return rewriter.notifyMatchFailure(op, "Indices list must not be empty.");
+    } else {
```
ramiro050, line 641: The previous `if` returns early. No need to have `else` branch (decreases indentation)

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -401,51 +623,211 @@ class ConvertAten_IndexPutImplOp
...
+      for (unsigned i = 0; i < optionalIndicesList.size(); i++) {
+        if (!optionalIndicesList[i].getType().isa<Torch::NoneType>()) {
+          // indicesList.push_back(optionalIndicesList[i]);
```
ramiro050, line 644: not needed

ramiro050 (approved):
Great work on this! I just have a few small comments

vivekkhandelwal1:
> Great work on this! I just have a few small comments

Thanks, @ramiro050, for your continuous feedback and help, without which it would have been hard to get this working.

PR #1765, opened 2023-01-01
Revert "rebase llvm: 5f24f893cac7aaea292c70f8aa83b021499114be"

powderluv:
Reverts llvm/torch-mlir#1760

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -3553,7 +3553,7 @@ public:
 
     GreedyRewriteConfig config;
     config.useTopDownTraversal = true;
-    config.maxIterations = GreedyRewriteConfig::kNoLimit;
+    config.maxIterations = GreedyRewriteConfig::kNoIterationLimit;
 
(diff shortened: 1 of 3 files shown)


powderluv:
Nightly builds broke with this bump so running a fully nightly with this revert here: https://github.com/llvm/torch-mlir/actions/runs/3817648354 .

We had to delete releases between 25-30th due to a PyTorch security issue.

powderluv:
Windows build failure is here: https://github.com/llvm/torch-mlir/actions/runs/3816546937/jobs/6492247476

ashay (approved):
LGTM.  It seems `ssize_t` for MSVC/Windows requires the following lines, but it's not clear what changed that caused the errors to surface.

```c++
#ifdef _MSC_VER
#include <BaseTsd.h>
typedef SSIZE_T ssize_t;
#else
#include <unistd.h>
#endif
```

Since the error isn't reproducible in our internal builds nor in pre- and post-merge builds, I suspect that our Release build invocation could be changed to work around the error, but I don't have any concrete evidence yet.

powderluv:
Thanks . We can investigate and reland. Will land this once all CIs are green and nightly build works (windows nightly just passed).

PR #1767, opened 2023-01-03
build: match windows CI and Release build steps

ashay:
After a recent LLVM tag update, the Windows Release build failed even
though pre- and post-merge Windows CI builds passed.  This patch makes
the Windows Release build use the same flags as the Window CI build to
try and address the Release build failures in the future.

---

@powderluv This was the only difference I found between Windows CI and Windows Release builds.  I'm not sure if it will resolve the Release build issue, but it's worth a try.

The Release builds for all platforms completed successfully with this patch [https://github.com/llvm/torch-mlir/actions/runs/3823707248], but of course, the true impact of this patch will only be visible once we update the LLVM tag.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -77,6 +77,7 @@ class CMakeBuild(build_py):
             cmake_args = [
                 f"-DCMAKE_BUILD_TYPE=Release",
                 f"-DPython3_EXECUTABLE={sys.executable}",
+                f"-DPython3_FIND_VIRTUALENV=ONLY",
                 f"-DLLVM_TARGETS_TO_BUILD=host",
                 f"-DMLIR_ENABLE_BINDINGS_PYTHON=ON",
                 f"-DLLVM_ENABLE_PROJECTS=mlir",


Shukla-Gaurav (approved):

Shukla-Gaurav:
@ashay Looks like there is more to debug: https://github.com/llvm/torch-mlir/actions/runs/3838374096/jobs/6534756674#step:5:5000
same `ssize` error.

ashay:
I'm still trying to resolve it.  Tried a couple of venv-related changes to make sure that pybind11 and other modules were the same between CI and Release, but that didn't help.  I'll let you know what I find.

powderluv:
btw we use python pip wheels for releases and cmake in the CI.

ashay:
Indeed, but interestingly, I was able to reproduce the error on a Windows Server 2022 VM using cmake alone (although setup.py also ran into the error).  I'm still puzzled why CI doesn't run into the same problem.  In any case, I did check that including that DataTypes.h resolves the problem on the Windows Server VM, so Release builds should be fixed in the future.

PR #1769, opened 2023-01-03
[custom op] Replace `tanh` dtype function with `expm1`

ramiro050:
This commit replaces the `tanh` dtype function, which was being used to test the implementation of dtype functions in
a710237437337a59cf067accab2c24a90fbb1d48, with a dtype function for `expm1`. The dtype function for `expm1` is identical to the `tanh` one, so the same level of testing is maintained.

Currently, there are ops getting dtype information from the `RefineTypes` pass and ops getting dtype information from the `TorchDtypeRefinementPipeline`. Since each pass can only propagete dtype information for the ops it knows how to handle, some models with many ops handled in both passes require the two dtype propagation passes to execute many times, reaching the iteration limit set in the `LowerToBackendContractPass`. To temporarily avoid this issue while the migration to `TorchDtypeRefinementPipeline` is finished, this commit switches `tanh` to `expm1`, since the latter is used a lot less in large models.


AmosLewis (approved):
LGTM. I cherry-pick this patch for distilgpt2, it works just fine.

Issue #1772, opened 2023-01-04
Transpose error when decomposing the convolution backward


ZihengJiang:
FX Graph:
```
def forward(self, primals_1, primals_8, tangents_1):                                                                                                                                                                                                                              convolution_backward = torch.ops.aten.convolution_backward.default(tangents_1, primals_8, primals_1, [16], [1, 1], [1, 1], [1, 1], False, [0, 0], 1, [True, True, True]);  tangents_1 = primals_8 = primals_1 = None
    getitem = convolution_backward[0]
    getitem_1 = convolution_backward[1]
    getitem_2 = convolution_backward[2];  convolution_backward = None
    return [getitem_1, getitem_2, None, None, None, None, None, getitem]
```

Converted torch dialect
```
module attributes {torch.debug_module_name = "GraphModule"} {
  func.func @forward(%arg0: !torch.vtensor<[16,3,5,5],f32>, %arg1: !torch.vtensor<[2,3,200,200],f32>, %arg2: !torch.vtensor<[2,16,198,198],f32>) -> (!torch.vtensor<[16,3,5,5],f32>, !torch.vtensor<[16],f32>, !torch.vtensor<[2,3,200,200],f32>) {
    %int0 = torch.constant.int 0
    %int1 = torch.constant.int 1
    %none = torch.constant.none
    %false = torch.constant.bool false
    %true = torch.constant.bool true
    %int2 = torch.constant.int 2
    %int3 = torch.constant.int 3
    %0 = torch.prim.ListConstruct %int1, %int1 : (!torch.int, !torch.int) -> !torch.list<int>
    %1 = torch.prim.ListConstruct %int0, %int0 : (!torch.int, !torch.int) -> !torch.list<int>
    torch.runtime.assert %true, "unimplemented: only strides of 1 supported."
    torch.runtime.assert %true, "unimplemented: only strides of 1 supported."
    torch.runtime.assert %true, "unimplemented: only dilations of 1 supported."
    torch.runtime.assert %true, "unimplemented: only dilations of 1 supported."
    %2 = torch.prim.ListConstruct %int2, %int3 : (!torch.int, !torch.int) -> !torch.list<int>
    %3 = torch.aten.flip %arg0, %2 : !torch.vtensor<[16,3,5,5],f32>, !torch.list<int> -> !torch.vtensor<[16,3,5,5],f32>
[...]

PR #1773, opened 2023-01-04
Add support for non-unit stride for conv backward

vivekkhandelwal1:


vivekkhandelwal1:
The following three cases are failing: ConvolutionBackwardModule2D_basic, ConvolutionBackwardModule2DPadded_basic, ConvolutionBackwardModule2DPaddedStrided_basic. The reason is that for these cases the conv op created during decomposition is a grouped convolution.

vivekkhandelwal1:
> The following three cases are failing: ConvolutionBackwardModule2D_basic, ConvolutionBackwardModule2DPadded_basic. The reason is that for these cases the conv op created during decomposition is a grouped convolution.

Working now.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -169,6 +169,35 @@ static Value createSoftmaxBackwardCommonKernel(PatternRewriter &rewriter,
...
+  for (unsigned i = 0; i < inputRank; i++) {
+    if (srcDims[i] != -1)
+      sourceDims.push_back(srcDims[i]);
+    if (dstDims[i] != -1)
+      destinationDims.push_back(dstDims[i]);
+  }
+  unsigned restDim = inputRank - srcDimInt.size();
+  for (unsigned i = 0; i < restDim; i++)
+    dimsOrder[destinationDims[i]] = sourceDims[i];
+  return dimsOrder;
```
ramiro050, line 198: Copying my comment over:

> I think you can do instead

> ```c++
>   SmallVector<int> dimsOrder(llvm::iota_range<int>(0, inputRank, /*inclusive=*/false));
>   dimsOrder.erase(dimsOrder.begin() + srcDimInt);
>   dimsOrder.insert(dimsOrder.begin() + dstDimInt, srcDimInt);
> ```

It might make it clearer to explain by example. Suppose you have

```python
input = torch.rand((2,3,4,5))
torch.movedim(input, 3, 1)
```

The result will be a tensor of shape `(2, 5, 3, 4)`. This is the same as doing

```python
torch.permute(input, (0, 3, 1, 2))
```

In other words, move the last dimension to the second dimension slot and shift all the other dimensions. This can be achieved by `pop`ing the last element of `(0, 1, 2, 3)` and inserting it in the index position `1`. This is what the C++ code I wrote above is doing: remove the `src` dimension from the list and re-insert it in the right location.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -169,6 +169,35 @@ static Value createSoftmaxBackwardCommonKernel(PatternRewriter &rewriter,
   return sub;
 }
 
+// This algorithm is adopted from:
+// https://github.com/pytorch/pytorch/blob/c47bdd7522c7146c091c95ff5b4ce354d7a8870c/aten/src/ATen/native/TensorShape.cpp#L3660
+static SmallVector<int64_t>
+computeDimsOrderForMoveDim(ArrayRef<int64_t> srcDimInt,
+                           ArrayRef<int64_t> dstDimInt, unsigned inputRank) {
```
ramiro050, line 176: I don't see a need for `srcDimInt` and `dstDimInt` to be arrays. They can just be `int`s

ramiro050 (changes requested):
Initial comments. Will add more comments tomorrow

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ZihengJiang: Agree. Introducing new op will increase the burden of backends

python/torch_mlir_e2e_test/test_suite/backprop.py
```diff
@@ -66,9 +66,9 @@ def __init__(self):
     @export
     @annotate_args([
         None,
-        ([-1, -1, -1, -1], torch.float32, True),
-        ([-1, -1, -1, -1], torch.float32, True),
-        ([-1, -1, -1, -1], torch.float32, True),
```
ramiro050, line 71: Why are these changed to static?

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
vivekkhandelwal1: Thanks for the suggested change!

python/torch_mlir_e2e_test/test_suite/backprop.py
vivekkhandelwal1: Because the updated decomposition now involves `aten.view` op https://github.com/llvm/torch-mlir/blob/3c954a9137cecf253ed7080b4aad3d6af94a8397/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L1580-L1581, and the lowering for view op doesn't support dynamic shapes being passed as an operand.

ramiro050: This is a step back for this op. In general, changes should not remove existing support for features that are e2e tested. Is there no way of supporting non-unit padding without removing support for dynamic sizes?

vivekkhandelwal1: TBH I don't see any other way. Because this is how the convolution backwards is implemented in PyTorch. Also, we have a trade-off here, either support this op for all args being non-unit or extend it for all but with static shapes.

silvasean: We should fix the view op to support dynamic shapes. I think that might be a bit of a yak shave but this is probably the final straw and we need to buckle down and do it. Regressing something as basic as dynamic shapes for convolution due to a missing feature in the view lowering seems like a good signal we need to dig in and fix the view op lowering.

vivekkhandelwal1: @silvasean The `tensor::CollapsShape` and `tensor::ExpandShape` op, do not work with the dynamic shapes. That's why we have this restriction.

silvasean: We can fix that or collaborate with the folks that can.

vivekkhandelwal1: Can you fix the upstream tensor dialect ops? Or who would be the appropriate person to fix this?

silvasean: I think at the moment Ramiro would be the best to scope this and make a judgment call about how to move forward.

ZihengJiang:
@ramiro050 Can we merge this?

ZihengJiang:
@vivekkhandelwal1 @ramiro050 any update for this one? it should fix https://github.com/llvm/torch-mlir/issues/1772

python/torch_mlir_e2e_test/test_suite/backprop.py
vivekkhandelwal1: Hi @ramiro050, what are your views on Sean's comments?

ramiro050: Hi, sorry I've been a bit busy. I will have an answer for you by Friday

ramiro050: I ran the e2e test locally with dynamic sizes and the error I get does not seem to be related to `ViewOp`. I get an error when the `ConvolutionOp` is being converted to linalg because the `groups` argument is not a constant int anymore:

https://github.com/llvm/torch-mlir/blob/9f927a7aa17ba8af79a4c9f054bd87914af35a64/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L1610-L1615

Do you see a similar thing?

ramiro050: I don't think that is an issue with `view`. The problem is that in the decomposition, the code that creates `gradWeightViewShapeValue` does not handle dynamic sizes correctly, so the `view` op gets a list of `-1`s for the output size argument.

https://github.com/llvm/torch-mlir/blob/9f927a7aa17ba8af79a4c9f054bd87914af35a64/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L1654-L1656

vivekkhandelwal1: That's true because the code was supposed to handle static cases. Even after fixing that it throws the following error:
```
** Failure : found ambiguous expand of dynamic sizes (e.g. [-1, -1] -> [-1, -1, -1])
```
which is related to view op.

vivekkhandelwal1: I have updated the patch with the changes, you can try the test `ConvolutionBackwardModule2D_basic`, and it will fail with this error. And particularly, this is the case that I initially said is not supported in Torch-MLIR, as well as by the `expand`/`collapse` op of the tensor dialect. Where the expansion of one or more dynamic dims is done to another set of dynamic dims.

ramiro050: Thanks, I'll take a look

ramiro050: Quick update: There is already some work being done to add dynamic shape support to `tensor.expand_shape` (https://discourse.llvm.org/t/rfc-add-explicit-shape-inputs-to-tensor-expand-shape/65952/16). I will be helping out to get the patch landed. This should fix the dynamic shape limitations.

ZihengJiang: @ramiro050 any update for this one?

ramiro050: The patch for `expand_shape` is almost complete. I made an update to it today. All that is missing is review + updating lit tests. Patch: https://reviews.llvm.org/D140821

vivekkhandelwal1:
Hey @ramiro050, can we get this patch merged which supports all static-shaped conv backwards as that's what I require right now? And @ZihengJiang can extend the support for dynamic cases as needed by them. Since this patch is stalled for a long time because of dependency from the MLIR side.

ramiro050:
> Hey @ramiro050, can we get this patch merged which supports all static-shaped conv backwards as that's what I require right now?

I'm a bit reluctant to merge this, since it is making a trade in functionality for the `convolutionBackward` op. We are going from an op having some support with full dynamic shapes to no support with dynamic shapes, which is to me a big step back.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1501,137 +1576,295 @@ class DecomposeAtenConvolutionBackwardOp
...
+
+    BaseTensorType gradWeightTy =
+        inputTransposedTy
+            .getWithSizesAndDtype(llvm::ArrayRef(gradWeightSizesInt),
+                                  inputTransposedTy.getOptionalDtype())
+            .cast<BaseTensorType>();
+
+    // Value numGroup = rewriter.create<AtenSizeIntOp>(loc, input, cstZero);
+    // Setting this to value two for `ConvolutionBackwardModule2D_basic` test case.
+    Value numGroup = cstTwo;
```
ramiro050, line 1767: In addition, this issue was never addressed

ramiro050:
> Hi @ramiro050, can we merge this patch now? Because it doesn't downgrade the existing support.

Interesting! What changed?

vivekkhandelwal1:
> > Hi @ramiro050, can we merge this patch now? Because it doesn't downgrade the existing support.
> 
> Interesting! What changed?

In the decomposition of convolution backwards, we need to compute three results: `gradInput`, `gradWeight`, and `gradBias`. Earlier, the decomposition was not able to handle the cases where the stride values were non-unit. In this patch, I have made a distinction in the decomposition of the value `gradWeight` for the unit and non-unit stride cases. So, this way I get the support for non-unit stride (and my use case right now is limited only to static shapes), while all the remaining cases that were earlier supported are still supported without any down-gradation to the functionality.

I hope we can merge this now since it has been stalling for a long time. And the changes only result in more cases being supported without removing any existing support.

@ramiro050

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1581,120 +1662,273 @@ class DecomposeAtenConvolutionBackwardOp
...
       return rewriter.notifyMatchFailure(
-          op, "unimplemented: only true values for output_mask supported.");
+          op, "unimplemented: only output padding of 0 supported.");
+
+    // TODO: Add a check for output mask. Removed it now since we don't support
+    // returning none values as of now in Torch-MLIR. So, right now we just
+    // assume that all the values of output mask are `True`, even if they turn
+    // out to be `False`, there is no harm in this assumption, since the
+    // resultant tensors corresponding to the `False` mask value would never be
+    // used.
```
ramiro050, line 1699: Can you explain this to me? The op returns `None`'s? From the ODS for this op, `None` values should never be returned. If it does, a bug should be created in PyTorch.

Moreover, torch-mlir does not support returning `None`s out of the entire `forward` function, but `None` values can be created along the way without a problem, so this should not be a reason for removing a check.

Lastly, ops should never knowingly produce outputs that are inconsistent with PyTorch. Even if the normal user behavior is to ignore the `False` mask when `output_mask` is `False`, the fact that one can still access the `False` mask and inspect it means that the value should match what one would get from PyTorch.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1581,120 +1662,273 @@ class DecomposeAtenConvolutionBackwardOp
...
+          loc, gradWeightTy, inputTransposed, gradOutputView, cstNone,
+          /*stride=*/op.getDilation(), op.getPadding(),
+          /*dilation=*/op.getStride(), op.getTransposed(),
+          op.getOutputPadding(), numGroup);
+
+      BaseTensorType weightTy = weight.getType().cast<BaseTensorType>();
+      if (!weightTy.hasSizes())
+        return failure();
+      SmallVector<int64_t> weightSizes(weightTy.getSizes());
+      for (unsigned i = 0; i < 2; i++) {
```
ramiro050, line 1842: Can you express this `2` in terms of existing variables? It seems to be the rank of `gradWeight` with the first two dimensions removed

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1581,120 +1662,273 @@ class DecomposeAtenConvolutionBackwardOp
...
+      gradWeight = rewriter.create<AtenMovedimIntOp>(
+          loc, gradWeightTypeForMoveDim, gradWeight, /*source=*/cstZero,
+          /*destination=*/cstTwo);
+
+      Value gradIntList = rewriter.create<Torch::PrimListConstructOp>(
+          loc, Torch::ListType::get(Torch::IntType::get(op.getContext())),
+          llvm::ArrayRef{cstZero});
+      gradWeight = rewriter.create<Torch::AtenSumDimIntListOp>(
+          loc, op.getResultTypes()[1], gradWeight, gradIntList, cstFalse,
+          cstNone);
```
ramiro050, line 1908: nit: add `/*var=*/` to false and nones

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
vivekkhandelwal1: If you run this test
```
class ConvolutionBackwardModule2DStatic(torch.nn.Module):

    def __init__(self):
        super().__init__()

    @export
    @annotate_args([
        None,
        ([1, 4, 64, 64], torch.float32, True),
        ([1, 320, 64, 64], torch.float32, True),
        ([4, 320, 3, 3], torch.float32, True),
    ])
    def forward(self, grad_out, input_vec, weight):
        return torch.ops.aten.convolution_backward(
            grad_out,
            input_vec,
            weight,
            bias_sizes=[4],
            stride=[1, 1],
            padding=[1, 1],
            dilation=[1, 1],
            transposed=False,
            output_padding=[0, 0],
            groups=1,
            output_mask=[True, True, False])


@register_test_case(module_factory=lambda: ConvolutionBackwardModule2DStatic())
def ConvolutionBackwardModule2DStatic_basic(module, tu: TestUtils):
    with torch.backends.mkldnn.flags(enabled=False):
        module.forward(tu.rand(1, 4, 64, 64), tu.rand(1, 320, 64, 64),
                       tu.rand(4, 320, 3, 3))
```
in PyTorch, it would give you a tuple with first two values tensor, and the third value None. On the other hand, if you run it in Torch-MLIR. It generates IR with the op returning three tensors. Don't you think this is wrong? Also, the test won't run completely, it will fail with the following error message:
```
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/e2e_testing/main.py", line 137, in <module>
    main()
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/e2e_testing/main.py", line 122, in main
    results = run_tests(tests, config, args.sequential, args.verbose)
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 338, in run_tests
    return [compile_and_run_test(test, config, verbose) for test in tests]
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 338, in <listcomp>
    return [compile_and_run_test(test, config, verbose) for test in tests]
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 315, in compile_and_run_test
    golden_trace=clone_trace(golden_trace))
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 85, in clone_trace
    return [
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 88, in <listcomp>
    output=clone_torch_script_value(item.output))
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 64, in clone_torch_script_value
    return tuple(clone_torch_script_value(field) for field in v)
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 64, in <genexpr>
    return tuple(clone_torch_script_value(field) for field in v)
  File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 74, in clone_torch_script_value
    assert False, "unhandled cloning of TorchScriptValue value type"
AssertionError: unhandled cloning of TorchScriptValue value type
```
I wanted to support this case. And since in the IR it was always returning three values irrespective of the output mask, that's why I removed the check.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -3470,3 +3470,41 @@ def forward(self, x):
...
+# ==============================================================================
+
+
+class MoveDimIntModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([None, ([3, 4, 2, 1], torch.float32, True)])
```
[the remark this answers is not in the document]
vivekkhandelwal1: The implementation works with dynamic dims. I'll change the tests.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ramiro050: The `AssertionError` is likely due to the fact that a `None` is being returned by a test. However, the fact that the op returns a `None` when the schema for it says it always returns tensors sounds like a bug in PyTorch. Can you open a bug in PyTorch for this?

python/torch_mlir_e2e_test/test_suite/backprop.py
```diff
@@ -148,6 +181,40 @@ def ConvolutionBackwardModule2DPadded_basic(module, tu: TestUtils):
...
+        ([2, 2, 3, 3], torch.float32, True),
+    ])
+    def forward(self, grad_out, input_vec, weight):
+        return torch.ops.aten.convolution_backward(
+            grad_out,
+            input_vec,
+            weight,
+            bias_sizes=[4],
+            stride=[2, 2],
+            padding=[1, 1],
```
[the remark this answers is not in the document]
vivekkhandelwal1: Actually, this PR adds support for non-unit strides primarily. Also, a test case `ConvolutionBackwardModule2DPadded_basic` for non-unit padding is already supported.

python/torch_mlir_e2e_test/test_suite/backprop.py
```diff
@@ -148,6 +181,40 @@ def ConvolutionBackwardModule2DPadded_basic(module, tu: TestUtils):
...
+            dilation=[1, 1],
+            transposed=False,
+            output_padding=[0, 0],
```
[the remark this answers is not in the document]
vivekkhandelwal1: This implementation only supports non-zero output padding. Here's the existing check: https://github.com/llvm/torch-mlir/pull/1773/files#diff-bec402e56f7c88a9b22018c90df7e9ca721bdd097d4ffdd7c13d9d3f4c2e0688R1685-R1692

python/torch_mlir_e2e_test/test_suite/backprop.py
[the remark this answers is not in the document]
ramiro050: Ah, we should update the PR title then :)

python/torch_mlir_e2e_test/test_suite/backprop.py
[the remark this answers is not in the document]
ramiro050: I think my comment was a bit confusing. I meant adding a test for `AtenConvolutionOp`, since the changes to `Linear.cpp` have expanded the ops support to handle non-zero `output_padding`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ramiro050: I'm afraid I can't allow this change to land in its current state. By removing the check for `output_mask`, you're essentially extending the set of valid values this op can take, and all such extensions need to be e2e tested in torch-mlir, as mentioned in the [testing guidelines](https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing).

vivekkhandelwal1: What changes do you want me to make, so that we can land this patch? It should support the case of False output mask values as well because that's the support I require. Also, is there a way to get this resolved faster? https://github.com/pytorch/pytorch/issues/97524

ramiro050: It seems that the issue has been triaged, so hopefully there will be a response soon. I have an idea for how we could go around this issue. I'll push a patch by today that should let us modify the signature of the op in torch-mlir to return optional tensors.

ramiro050: Turns out that changing the signature to return optional tensors requires some complex modifications to the way the shape library works. I gotta think a bit about how to approach this. Also, I commented on your issue with some examples. Hopefully we'll get a response soon.

vivekkhandelwal1: Okay, thanks!

ramiro050: Okay, so the response on the issue is basically "this is a bug, but it won't get fixed." I can think of two approaches we can take in Torch-MLIR:

1. Manually fix the ODS for the `convolution_backward` op to return optional tensors
2. Only allow `False` in the `output_mask` when the `None` result is not used anywhere in the graph

Solution 1 would allow us to e2e test the return of `None` values, so it seems like the most complete solution. However, the fact that we would need to have a manual change to the ODS for this op indefinitely (since the signature will not get fixed upstream), is not ideal. Moreover, changing the signature to return optional tensors would require *significant* changes to the abstract interpretation library code to support the handling of optional tensor returns. Given that the `None` values are meant to be ignored, investing this much time in modifying the abstract interpretation library code to add support for returning values people are unlikely to ever use seems like a waste of time.

If possible I would prefer going with option 2. It requires minimal changes, but it does restrict the use of `False` to graphs that ignore the `None` values. Does the graph you're trying to lower ignore the `None` values? If so, let's go with option 2.

vivekkhandelwal1: I'm assuming that when you say `Does the graph you're trying to lower ignore the None values?`, you mean the None value from the result of the `convolution_backward` op is not used anywhere. Yes, that's true the value is not being used anywhere in the graph, because ideally, it expects that there would only be 2 tensors as a result of convolution backward, not three.

vivekkhandelwal1: Btw what changes do we need for supporting the 2nd approach?

ramiro050: For the 2nd, you need to go over each value of `output_mask`. If the value is false, you need to check that the corresponding result does not have users. If it does, then we need to return a `notifyMatchFailure`.

vivekkhandelwal1:
Hi @ramiro050, I have addressed all the comments related to this PR. Could you please review it now?

ramiro050 (approved):
Awesome! LGTM

PR #1774, opened 2023-01-04
Added support for aten::norm.ScalarOpt_dim

GlebKazantaev:
Added support for `aten::norm.ScalarOpt_dim` which is required by `torch.nn.functional.normalize`.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -4502,6 +4502,32 @@ def Torch_AtenLayerNormOp : Torch_Op<"aten.layer_norm", [
   }];
 }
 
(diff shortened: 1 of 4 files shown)


ke1337 (approved):

antoniojkim (approved):

GlebKazantaev:
@silvasean is it okay if I create github issue to cover https://github.com/llvm/torch-mlir/pull/1774/files#diff-882c555d6fd90bc6e098e9d278f623fc33be59f4514eb7d4fd2c294c59a9dca3R11 ? 
I tried to handle torch to linalg conversion for `AtenNormScalarOptDimOp` but looks like it require smth more than just shape inference and type propagation.

GlebKazantaev:
Created an issue: https://github.com/llvm/torch-mlir/issues/1788

PR #1776, opened 2023-01-05
[MLIR][TORCH] Add canonicalizer for `aten.new_empty_strided` op

vivekkhandelwal1:


vivekkhandelwal1:
hi, @ramiro050 could you please review this PR?

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1393,6 +1393,48 @@ void AtenSortIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+        isDefaultStride = false;
+        break;
+      }
+    }
+
+    if (!isDefaultStride)
+      return rewriter.notifyMatchFailure(
+          op, "only default strides supported for new_empty_strided op");
+
+    rewriter.replaceOpWithNewOp<AtenNewEmptyOp>(
```
ramiro050, line 1431: This is more a decomposition than a canonicalization. I would move it to decompose complex ops

ramiro050 (approved):
Other than the one comment below, LGTM

lib/Dialect/Torch/IR/TorchOps.cpp
vivekkhandelwal1: I was also confused between placing it as a decomposition or canonicalization. Thanks for the clarification! Btw, how do we primarily distinguish them?

ramiro050: This is a pretty good article explaining canonicalizations: https://sunfishcode.github.io/blog/2018/10/22/Canonicalization.html. The main point being that a canonicalization makes "subsequent optimizations more effective." In this particular case, replacing `AtenNewEmptyStrided` with `AtenNewEmpty` does not really unlock new optimizations in torch-mlir, so it is more of a decomposition. Of course, the boundary between the canonicalizations and decompositions can often be very blurry. But in general in torch-mlir, if all the code does is replace one op with another, I would handle it as a decomposition.

Issue #1708, opened 2022-12-12
[LTC] Tests failing for the latest PyTorch release 2022-12-11


vivekkhandelwal1:
Hi, about 30 tests are failing for LTC with the latest Pytorch release (https://github.com/llvm/torch-mlir/pull/1703) version 2022-12-11. The error is similar for all the tests and is given below:
```
    FAIL - "Conv2dNoPaddingModule_basic"
        Runtime error: Traceback (most recent call last):
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 302, in compile_and_run_test
            trace = config.run(compiled, golden_trace)
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/configs/lazy_tensor_core.py", line 35, in run
            cpu_outputs = tree_map(to_device('cpu'), output)
          File "/home/vivek/work/02_07/mlir_venv/lib/python3.8/site-packages/torch/utils/_pytree.py", line 195, in tree_map
            return tree_unflatten([fn(i) for i in flat_args], spec)
          File "/home/vivek/work/02_07/mlir_venv/lib/python3.8/site-packages/torch/utils/_pytree.py", line 195, in <listcomp>
            return tree_unflatten([fn(i) for i in flat_args], spec)
          File "/home/vivek/work/02_07/vivekkhandelwal1-torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/configs/lazy_tensor_core.py", line 15, in <lambda>
            return lambda e: e.to(device) if isinstance(e, torch.Tensor) else e
        RuntimeError: vector::_M_range_check: __n (which is 2) >= this->size() (which is 2)
```
The list of the failing tests is: https://github.com/llvm/torch-mlir/blob/1715371fe2e3030dcdddaee5597ce43aa620719c/e2e_testing/xfail_sets.py#L784-L815

antoniojkim:
I am unable to reproduce this with the latest `PyTorch` and latest `Torch-MLIR`. In my local testing:

Lazy tests:
```
Summary:
    Passed: 576
    Expectedly Failed: 170
```

Python tests:
```
Testing Time: 7.92s
  Passed: 17
```

GlebKazantaev:
[...]

PR #1779, opened 2023-01-06
Fix OptionalCType class name

GlebKazantaev:
Due to [recent changes in torchgen types](https://github.com/pytorch/pytorch/pull/90589/files#diff-67f65b3bf1f3df18d7688de4f4fc0cb276fab638da426299e337439eeb938083R1) we started to observe problems with optional arguments in generated lazy operations.

Issue: https://github.com/llvm/torch-mlir/issues/1708


antoniojkim (approved):

build_tools/autogen_ltc_backend.py
```diff
@@ -38,7 +38,7 @@ def reindent(text, prefix=""):
 @dataclass(frozen=True)
 class GenMlirLazyIr(torchgen.dest.GenLazyIR):
     def isOptionalCType(self, arg):
-        return str(type(arg)) == "<class 'torchgen.api.types.OptionalCType'>"
+        return str(type(arg)) == "<class 'torchgen.api.types.types.OptionalCType'>"
```
ke1337, line 41: Is there a test to catch changes like this? It would be the best if we can have CI to detect any breaking changes.

GlebKazantaev: @ke1337 we have such tests and they catched the problem. See https://github.com/llvm/torch-mlir/issues/1708 for more details.

ke1337 (approved):

PR #1780, opened 2023-01-06
Fix unsqueeze in Torch to Tosa conversion

navahgar:
This PR fixes a couple of issues with the conversion of `torch.aten.unsqueeze` to `tosa.reshape`.

* When input dim < 0, the conversion to positive is off by 1. 
* When input dim == rank, 1 does not get appended to the output shape.

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -2510,7 +2510,10 @@ LogicalResult ConvertAtenOp<AtenUnsqueezeOp>::matchAndRewrite(
   if (!matchPattern(op.getDim(), m_TorchConstantInt(&dim)))
     return rewriter.notifyMatchFailure(op, "dim must be a Scalar constant");
 
-  dim = toPositiveDim(dim, selfRank);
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -2510,7 +2510,7 @@ LogicalResult ConvertAtenOp<AtenUnsqueezeOp>::matchAndRewrite(
   if (!matchPattern(op.getDim(), m_TorchConstantInt(&dim)))
     return rewriter.notifyMatchFailure(op, "dim must be a Scalar constant");
 
-  dim = toPositiveDim(dim, selfRank);
+  dim = toPositiveDim(dim, selfRank + 1);
```
sjarus, line 2513: I recommend adding a comment stating how unsqueeze cannot use the native toPositiveDim semantics and has special needs.

sjarus (approved):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
navahgar: Done.

asaadaldien (approved):

PR #1785, opened 2023-01-09
Expose metadata of torch-mlir types

makslevental:
I would like to use some of the torch-mlir types/ops in a downstream project, through the C API. It would be nice to have access to some of the metadata in order to do some light type checking (sizes, ranks, dtypes, etc).

I tried to match the C API but maybe there are some C conventions that I've flouted...

I also didn't see any tests for this API but happy to add some (somewhere).

cc @powderluv


silvasean:
Since these are not exercised by our load-bearing test flow, you need to add tests like this: https://github.com/llvm/llvm-project/blob/main/mlir/test/CAPI/quant.c

include/torch-mlir-c/TorchTypes.h
```diff
@@ -208,6 +224,16 @@ torchMlirTorchValueTensorTypeGetWithLeastStaticInformation(MlirContext context);
 MLIR_CAPI_EXPORTED MlirType
 torchMlirTorchValueTensorTypeGetFromAttribute(MlirAttribute attr);
 
+/// Gets the the rank (number of dimensions) of a !torch.vtensor
+MLIR_CAPI_EXPORTED size_t torchMlirTorchValueTensorTypeGetRank(MlirType t);
```
silvasean, line 228: expose the "hasSizes/hasDtype" functions as well

include/torch-mlir-c/TorchTypes.h
```diff
@@ -52,6 +52,9 @@ MLIR_CAPI_EXPORTED MlirType
 torchMlirTorchTupleTypeGet(MlirContext context, intptr_t numContainedTypes,
                            MlirType const *containedTypes);
 
+/// Gets the length of the !torch.tuple type (i.e. number of contained types).
+MLIR_CAPI_EXPORTED size_t torchMlirTorchTupleGetArity(MlirType t);
```
silvasean, line 56: also expose a way to get the contained types.

silvasean (changes requested):

makslevental:
Seems `out-of-tree` isn't picking up the `torch-mlir-capi-torch-test` executable. Will investigate.

test/lit.cfg.py
```diff
@@ -56,6 +56,8 @@
 
 # Tweak the PATH to include the tools dir.
 llvm_config.with_environment('PATH', config.llvm_tools_dir, append_path=True)
+# Tweak the PATH to include the binary build dir, in order to pick up CAPI tests during out-of-tree.
+llvm_config.with_environment('PATH', os.path.join(config.llvm_build_dir, 'bin'), append_path=True)
```
makslevental, line 60: I'm not sure if this is the right thing to do but this (and `config.llvm_build_dir = "@CMAKE_BINARY_DIR@"` below) is the only way I could get lit to find the `torch-mlir-capi-torch-test` executable (which gets deposited there). 

Note that

```
config.torch_mlir_obj_root = "@TORCH_MLIR_BINARY_DIR@"
...
llvm_config.with_environment('PATH', config.torch_mlir_obj_root, append_path=True)
```

doesn't work because `TORCH_MLIR_BINARY_DIR==CMAKE_CURRENT_BINARY_DIR==torch-mlir/build/tools/torch-mlir` for out-of-tree.

Finally 

```
add_lit_testsuite(check-torch-mlir-capi "Running the torch-mlir CAPI tests"
        ${CMAKE_CURRENT_BINARY_DIR}
        --->EXCLUDE_FROM_CHECK_ALL<---
        DEPENDS torch-mlir-capi-torch-test
        )
```

doesn't work either (with or without 

```
add_dependencies(check-torch-mlir-all
  check-torch-mlir
  check-torch-mlir-dialects
  check-torch-mlir-capi
  )
```

include/torch-mlir-c/TorchTypes.h
```diff
@@ -182,6 +201,22 @@ torchMlirTorchNonValueTensorTypeGetWithLeastStaticInformation(
...
+
+/// Return true if this type has a list of sizes.
+MLIR_CAPI_EXPORTED bool torchMlirTorchNonValueTensorTypeHasSizes(MlirType t);
+
+/// Return true if this type has a dtype.
+MLIR_CAPI_EXPORTED bool torchMlirTorchNonValueTensorTypeHasDtype(MlirType t);
+
+/// Gets the the sizes of the dimensions of a !torch.tensor; note -1 size
+/// indicates an unrefined/unknown size dimension.
+MLIR_CAPI_EXPORTED const int64_t *torchMlirTorchNonValueTensorTypeGetSizes(MlirType t);
```
silvasean, line 215: please have this write into a user-provided buffer.

include/torch-mlir-c/TorchTypes.h
```diff
@@ -208,6 +243,22 @@ torchMlirTorchValueTensorTypeGetWithLeastStaticInformation(MlirContext context);
 MLIR_CAPI_EXPORTED MlirType
 torchMlirTorchValueTensorTypeGetFromAttribute(MlirAttribute attr);
 
+/// Gets the the rank (number of dimensions) of a !torch.vtensor
+MLIR_CAPI_EXPORTED size_t torchMlirTorchValueTensorTypeGetRank(MlirType t);
```
silvasean, line 247: int64_t to be consistent with mlirShapedTypeGetRank

test/CAPI/torch.c
```diff
@@ -0,0 +1,163 @@
...
+  // CHECK: ValueTensorType has-sizes-no-dtype pos 2 size: 3
+  testTensor(ctx, -1, sizes, f32, "no-sizes-has-dtype");
+  // CHECK: NonValueTensorType no-sizes-has-dtype hasSizes: 0
+  // CHECK: NonValueTensorType no-sizes-has-dtype hasDtype: 1
+  // CHECK: NonValueTensorType no-sizes-has-dtype dtype: f32
+  // CHECK: ValueTensorType no-sizes-has-dtype hasSizes: 0
+  // CHECK: ValueTensorType no-sizes-has-dtype hasDtype: 1
+  // CHECK: ValueTensorType no-sizes-has-dtype dtype: f32
+
+  MlirType dict = torchMlirTorchDictTypeGet(f16, f32);
```
silvasean, line 144: This is actually not a valid dict type. The key and value type need to be torch dialect types, but we don't seem to verify that. Can you maybe add that check too? It should be just adding `let genVerifyDecl = 1;` to the td and implementing the method.

test/lit.cfg.py
silvasean: Hmmm... I'm not an expert but this seems reasonable.

silvasean (changes requested):

test/CAPI/torch.c
makslevental: Man this was an "excursion". 

Okay so firstly, indeed adding `genVerifyDecl = 1` is just a matter of implementing `LogicalResult DictType::verify` that checks key types and value types. But there are already [type predicates](https://github.com/llvm/torch-mlir/blob/7f95f4429795d80923a3a9c291ff5a4c77fdd180/include/torch-mlir/Dialect/Torch/IR/TorchTypes.td#L408) for this so I thought hmm I should try to use them somehow (save bytes in the final binary size) but it turns out that you can't do this ([upstream issue](https://github.com/llvm/llvm-project/issues/60015)).

Okay fine, code repetition it is, but then you gotta test it right? But adding lit/FileCheck tests, such as 

```mlir
// expected-error @+1 {{invalid torch.none for !torch.dict key type}}
func.func private @fail_dict() -> !torch.dict<none, tensor>
```

because the parser goes through the ["unchecked" factory](https://github.com/llvm/llvm-project/blob/main/mlir/docs/DefiningDialects/AttributesAndTypes.md#verification):

```
<unknown>:0: error: unexpected error: invalid '!torch.none' for !torch.dict key type
torch-mlir-opt:
PLEASE submit a bug report to https://github.com/llvm/llvm-project/issues/ and include the crash backtrace.
Stack dump:
0.      Program arguments: ./bin/torch-mlir-opt -split-input-file -verify-diagnostics testdict.mlir
1.      MLIR Parser: custom op parser 'func.func'
2.      MLIR Parser: custom op parser 'torch.prim.DictConstruct'
 #0 0x0000564f22b5bf74 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int)
 #1 0x0000564f22b5c34a PrintStackTraceSignalHandler(void*)
 #2 0x0000564f22b59b68 llvm::sys::RunSignalHandlers()
 ...
#12 0x0000564f1ddd30db mlir::torch::Torch::DictType mlir::detail::StorageUserBase
#13 0x0000564f1dd498e1 mlir::torch::Torch::DictType::get(mlir::MLIRContext*, mlir::Type, mlir::Type)
#14 0x0000564f1dd4ad04 mlir::torch::Torch::DictType::parse(mlir::AsmParser&)
#15 0x0000564f1dd4889f generatedTypeParser(mlir::AsmParser&, ...
#16 0x0000564f1dd4b396
```

So therefore this has to be tested in driver/user code but I'm writing such a test anyway right? So I added/exposed 

```C
MLIR_CAPI_EXPORTED MlirType torchMlirTorchDictTypeGetChecked(
    MlirContext context, MlirType keyType, MlirType valueType);
```

which seems useful/reasonable enough but the implementation is annoyingly verbose:

```cpp
MlirType torchMlirTorchDictTypeGetChecked(MlirContext context, MlirType keyType,
                                          MlirType valueType) {
  auto unknownLoc = unwrap(mlirLocationUnknownGet(context));
  return wrap(Torch::DictType::getChecked(unknownLoc, unwrap(context),
                                          unwrap(keyType), unwrap(valueType)));
}
```

The reason you have to pass both the location *and* the context is because otherwise (for :shrug: knows why) you still hit the unchecked factory

```
// llvm-project/mlir/include/mlir/IR/StorageUniquerSupport.h

class StorageUserBase : public BaseT, public Traits<ConcreteT>... {
  ...
  static ConcreteT getChecked(const Location &loc, Args... args) {
    return ConcreteT::getChecked(getDefaultDiagnosticEmitFn(loc), args...);
  }
}

// build/tools/torch-mlir/include/torch-mlir/Dialect/Torch/IR/TorchTypes.cpp.inc

DictType DictType::get(::mlir::MLIRContext *context, ::mlir::Type keyType, ::mlir::Type valueType) {
  return Base::get(context, keyType, valueType);
}

DictType DictType::getChecked(::llvm::function_ref<::mlir::InFlightDiagnostic()> emitError, ::mlir::MLIRContext *context, ::mlir::Type keyType, ::mlir::Type valueType) {
  return Base::getChecked(emitError, context, keyType, valueType);
}

DictType DictType::get(::mlir::Type keyType, ::mlir::Type valueType) {
  return Base::get(keyType.getContext(), keyType, valueType);
}

DictType DictType::getChecked(::llvm::function_ref<::mlir::InFlightDiagnostic()> emitError, ::mlir::Type keyType, ::mlir::Type valueType) {
  return Base::get(keyType.getContext(), keyType, valueType);
}
```

silvasean (approved):
Thanks for diving in and fixing this!

Issue #1764, opened 2023-01-01
Empty tensor in LINALG when select index equals -1


li-plus:
Compile the following simple model to LINALG backend.

```python
import torch
import torch_mlir

class Model(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self, x):
        return x[-1]

model = Model().eval()
inputs = torch.randn(2, 3)
model = torch.jit.trace(model, inputs)
model.save('model.jit')

model = torch.jit.load('model.jit')
print(model(inputs))
module = torch_mlir.compile(model, inputs, output_type="linalg_on_tensors")
print(module)
```

It prints out the below text, where `%extracted_slice` is an empty tensor. This should be a bug.

```
tensor([-1.5684,  3.1238,  0.7494])
module attributes {torch.debug_module_name = "Model"} {
  ml_program.global private mutable @global_seed(dense<0> : tensor<i64>) : tensor<i64>
  func.func @forward(%arg0: tensor<2x3xf32>) -> tensor<3xf32> {
    %extracted_slice = tensor.extract_slice %arg0[1, 0] [0, 3] [1, 1] : tensor<2x3xf32> to tensor<0x3xf32>
    %cast = tensor.cast %extracted_slice : tensor<0x3xf32> to tensor<?x?xf32>
    %cast_0 = tensor.cast %cast : tensor<?x?xf32> to tensor<1x3xf32>
    %collapsed = tensor.collapse_shape %cast_0 [[0, 1]] : tensor<1x3xf32> into tensor<3xf32>
    return %collapsed : tensor<3xf32>
  }
}
```

One way to fix this is to convert the start to positive index before adding one in the decomposition of `aten.select.int`. This can avoid the `end` being 0 when the `start` being -1.

https://github.com/llvm/torch-mlir/blob/9e71d271f67951cdfb1bca335a665bf10e672472/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L251-L278

Another way might be to handle the case of `end=0` when lowering to LINALG.

I don't know which way is preferred.

li-plus:
One way to fix this is to convert the `start` to positive index before adding one in the decomposition of `aten.select.int`. This can avoid the `end` being 0 when the `start` being -1.

https://github.com/llvm/torch-mlir/blob/9e71d271f67951cdfb1bca335a665bf10e672472/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L251-L278

[...]

PR #1787, opened 2023-01-10
Fix empty tensor when select -1

li-plus:
This PR fixed issue #1764 by first converting the `start` index to non-negative value before calculating the `end` index. The conversion is: `start += int(start < 0) * dimSize`. This can avoid `end` being 0 when `start` is -1.

I also added a test case `SelectIntNegativeDimAndIndexStaticModule_basic` that did not pass in LINALG backend without this PR to prove the correctness of my implementation.


li-plus:
@ramiro050 Would you review this PR? Thanks.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1347,6 +1347,18 @@ OpFoldResult AtenIntScalarOp::fold(ArrayRef<Attribute> operands) {
   return nullptr;
 }
 
+//===----------------------------------------------------------------------===//
+// AtenIntBoolOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenIntBoolOp::fold(ArrayRef<Attribute> operands) {
```
ramiro050, line 1354: Can you add e2e tests for this op too?

li-plus: Done! Added three e2e tests for this op.

li-plus:
If everything is ok, would you please merge it for me? And I'll close the related issue.

ZihengJiang:
Merged. Thanks @li-plus and @ramiro050

PR #1791, opened 2023-01-11
[MHLO] Add new conversion pattern for aten.stack

ZihengJiang:


li-plus:
Stack op will be supported once #1747 is merged. It might be better to support it by decomposition.

Issue #1677, opened 2022-12-05
quantize_per_tensor causes "error: unsupported by backend contract: tensor with unknown rank"


mgehre-amd:
With
```
import torch
import torch_mlir

class TestModule(torch.nn.Module):
  def __init__(self):
    super().__init__()

  def forward(self, x):
    y = torch.quantize_per_tensor(x, 1.0, 0, torch.qint8)
    return y

args = (torch.tensor([1., 3., 4.], dtype=torch.float),)
print(TestModule().forward(*args))

print("Compiling with Torch-MLIR")
linalg_on_tensors_mlir = torch_mlir.compile(
    TestModule(),
    args,
    output_type=torch_mlir.OutputType.LINALG_ON_TENSORS,
    use_tracing=True)
```
I see 
```
tensor([1., 3., 4.], size=(3,), dtype=torch.qint8,
       quantization_scheme=torch.per_tensor_affine, scale=1.0, zero_point=0)
Compiling with Torch-MLIR
Traceback (most recent call last):
  File "/scratch/torch-mlir/examples/test.py", line 16, in <module>
    linalg_on_tensors_mlir = torch_mlir.compile(
  File "/scratch/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/__init__.py", line 371, in compile
    run_pipeline_with_repro_report(
  File "/scratch/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/compiler_utils.py", line 73, in run_pipeline_with_repro_report
    raise TorchMlirCompilerError(trimmed_message) from None
torch_mlir.compiler_utils.TorchMlirCompilerError: Lowering TorchScript IR -> Torch Backend IR failed with the following diagnostics:
error: unsupported by backend contract: tensor with unknown rank
note: see current operation: %3 = "torch.tensor_static_info_cast"(%arg0) : (!torch.vtensor<[3],f32>) -> !torch.vtensor<*,f32>
note: this is likely due to a missing shape transfer function in shape_lib_gen.py


For Torch-MLIR developers, the error can be reproduced with:
$ torch-mlir-opt -pass-pipeline='builtin.module(torchscript-module-to-torch-backend-pipeline{backend-legal-ops=torch.aten.flatten.using_ints})' /tmp/TestModule.mlir
Add '-mlir-print-ir-after-all -mlir-disable-threading' to get the IR dump for debugging purpose.
```
with
```
[...]

PR #1795, opened 2023-01-12
Implement quantize_per_tensor and int_repr with their lowering to linalg

mgehre-amd:
This PR adds support for `quantize_per_tensor` and `int_repr` and their lowering to Linalg.
The use case is to support the pattern `int_repr(quantize_per_tensor(x, scale, zero_point, type))`, which is used by the brevitas quantizer.

While working on this PR, I discovered two issues that make lowering `quantize_per_tensor` fail for `quint8` and `qint32`. `qint8` works and is the value of this PR.
(1) For uint8, I get `ERROR: dtype (torch.int8) is not equal to golden dtype (torch.uint8)` in the e2e test. It's unclear to me why this happens and how to fix. After learning more about it, I will fix it in a follow-up PR.
(2) qint32 doesn't even have a type representation in `TorchTypes.td` yet.

The strategy to lower `quantize_per_tensor` is to do `cast(clamp(round(input/scale + zero_point))` into an integer type.
Doing that, `int_repr` becomes a no-op. When we want to implement other operators that work on quantized types (like qint8),
we will also need to carry the side-channel information scale and zero_point to them. For `int_repr`, this is not needed because it doesn't use that information.

*Details*:
With this PR from
```
  def forward(self, x):
   return torch.int_repr(torch.quantize_per_tensor(x, 0.5, 0, torch.qint8))
```
we first get
```
module attributes {torch.debug_module_name = "TestModule"} {
  func.func @forward(%arg0: !torch.vtensor<[3],f32>, %arg1: !torch.vtensor<[3],f32>) -> !torch.vtensor<[3],si8> {
    %float5.000000e-01 = torch.constant.float 5.000000e-01
    %int0 = torch.constant.int 0
    %int12 = torch.constant.int 12
    %0 = torch.aten.quantize_per_tensor %arg0, %float5.000000e-01, %int0, %int12 : !torch.vtensor<[3],f32>, !torch.float, !torch.int, !torch.int -> !torch.vtensor<[3],!torch.qint8>
    %1 = torch.aten.int_repr %0 : !torch.vtensor<[3],!torch.qint8> -> !torch.vtensor<[3],si8>
    return %1 : !torch.vtensor<[3],si8>
  }
}
```
and then lower that into
```
 #map = affine_map<(d0) -> (d0)>
module attributes {torch.debug_module_name = "TestModule"} {
  ml_program.global private mutable @global_seed(dense<0> : tensor<i64>) : tensor<i64>
  func.func @forward(%arg0: tensor<3xf32>, %arg1: tensor<3xf32>) -> tensor<3xi8> {
    %cst = arith.constant 0.000000e+00 : f32
    %cst_0 = arith.constant 2.550000e+02 : f32
    %0 = tensor.empty() : tensor<3xi8>
    %1 = linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel"]} ins(%arg0 : tensor<3xf32>) outs(%0 : tensor<3xi8>) {
    ^bb0(%in: f32, %out: i8):
      %2 = arith.addf %in, %cst : f32
      %3 = math.roundeven %2 : f32
      %4 = arith.maxf %3, %cst : f32
      %5 = arith.minf %4, %cst_0 : f32
      %6 = arith.fptoui %5 : f32 to i8
      linalg.yield %6 : i8
    } -> tensor<3xi8>
    return %1 : tensor<3xi8>
  }
}
```

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -12,7 +12,10 @@
 
 from torch_mlir_e2e_test.test_suite import COMMON_TORCH_MLIR_LOWERING_XFAILS
 
-LINALG_XFAIL_SET = COMMON_TORCH_MLIR_LOWERING_XFAILS
+LINALG_XFAIL_SET = COMMON_TORCH_MLIR_LOWERING_XFAILS.union({
+    # TODO ERROR: dtype (torch.int8) is not equal to golden dtype (torch.uint8)
+    "Quantize_uint8"
+})
 
 TORCHDYNAMO_XFAIL_SET = {
     #### General TorchDynamo/PyTorch errors
@@ -43,6 +46,10 @@ TORCHDYNAMO_XFAIL_SET = {
     # TODO: This is due to returning a scalar float as output from the test.
     # We should probably just standardize all tests to return tensors.
     "DivIntModule_basic",
+
+    # NotImplementedError: could not find kernel for aten.quantize_per_tensor.default at dispatch key DispatchKey.Meta
+    "Quantize_int8",
+    "Quantize_uint8",
 
     #### Torch-MLIR internal compiler errors
 
@@ -735,6 +742,8 @@ LTC_XFAIL_SET = {
     "DivIntModule_basic",
     "NeFloatIntModule_basic",
     "NeIntModule_basic",
+    "Quantize_int8",
+    "Quantize_uint8",
     "QuantizedMLP_basic",
     "RandLikeDtypeModule_basic",
     "RandLikeModule_basic",
--- a/lib/Dialect/Torch/IR/TorchTypes.cpp
+++ b/lib/Dialect/Torch/IR/TorchTypes.cpp
@@ -372,6 +372,10 @@ static Type convertDtypeToBuiltinElementType(MLIRContext *context, Type dtype) {
   } else if (auto integerType = dtype.dyn_cast<IntegerType>()) {
     return IntegerType::get(context, integerType.getWidth(),
                             IntegerType::Signless);
+  } else if (auto integerType = dtype.dyn_cast<QInt8Type>()) {
+    return IntegerType::get(context, 8, IntegerType::Signless);
+  } else if (auto integerType = dtype.dyn_cast<QUInt8Type>()) {
+    return IntegerType::get(context, 8, IntegerType::Signless);
   }
   emitError(UnknownLoc::get(context))
       << "unimplemented: conversion of dtype " << dtype
--- a/test/Conversion/TorchToLinalg/elementwise.mlir
+++ b/test/Conversion/TorchToLinalg/elementwise.mlir
@@ -94,3 +94,89 @@ func.func @insufficient_dims_for_triu(%arg0: !torch.vtensor<[?],f32>) -> !torch.
   %0 = torch.aten.triu %arg0, %int0 : !torch.vtensor<[?],f32>, !torch.int -> !torch.vtensor<[?],f32>
   return %0 : !torch.vtensor<[?],f32>
 }
+
+// -----
+
+// CHECK-LABEL:   func.func @torch.aten.quantize_per_tensor(
+// CHECK-SAME:                                %[[ARG:.*]]: !torch.vtensor<[],f32>) -> !torch.vtensor<[],!torch.qint8> {
+// CHECK:           %[[BUILTIN_TENSOR:.*]] = torch_c.to_builtin_tensor %[[ARG]] : !torch.vtensor<[],f32> -> tensor<f32>
+// CHECK:           %[[F3:.*]] = torch.constant.float 3.000000e+00
+// CHECK:           %[[C3:.*]] = torch_c.to_f64 %[[F3]]
+// CHECK:           %[[I7:.*]] = torch.constant.int 7
+// CHECK:           %[[C7:.*]] = torch_c.to_i64 %[[I7]]
+// CHECK:           %[[C4:.*]] = torch.constant.int 12
+// CHECK:           %[[C1:.*]] = arith.constant 1 : index
+// CHECK:           %[[INIT_TENSOR:.*]] = tensor.empty() : tensor<i8>
+// CHECK:           %[[GENERIC:.*]] = linalg.generic {indexing_maps = [affine_map<() -> ()>, affine_map<() -> ()>], iterator_types = []} ins(%[[BUILTIN_TENSOR]] : tensor<f32>) outs(%[[INIT_TENSOR]] : tensor<i8>) {
+// CHECK:           ^bb0(%[[BBARG0:.*]]: f32, %{{.*}}: i8):
+// CHECK:             %[[TRUNCF:.*]] = arith.truncf %[[C3]] : f64 to f32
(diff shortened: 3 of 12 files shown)


lib/Dialect/Torch/IR/TorchTypes.cpp
```diff
@@ -372,6 +372,10 @@ static Type convertDtypeToBuiltinElementType(MLIRContext *context, Type dtype) {
   } else if (auto integerType = dtype.dyn_cast<IntegerType>()) {
     return IntegerType::get(context, integerType.getWidth(),
                             IntegerType::Signless);
+  } else if (auto integerType = dtype.dyn_cast<QInt8Type>()) {
+    return IntegerType::get(context, 8, IntegerType::Signless);
```
silvasean, line 376: This is not correct. As discussed in #1677, a tensor with !torch.qint8 dtype needs to expand to 3 separate values -- the tensor, scale, and zero point. We want to get this right from the very beginning to allow incremental progress. I think that it would make sense to include a quantized add in this PR as well since that exercises this.

e2e_testing/xfail_sets.py
```diff
@@ -12,7 +12,10 @@
 
 from torch_mlir_e2e_test.test_suite import COMMON_TORCH_MLIR_LOWERING_XFAILS
 
-LINALG_XFAIL_SET = COMMON_TORCH_MLIR_LOWERING_XFAILS
+LINALG_XFAIL_SET = COMMON_TORCH_MLIR_LOWERING_XFAILS.union({
+    # TODO ERROR: dtype (torch.int8) is not equal to golden dtype (torch.uint8)
```
silvasean, line 16: XFAIL's are for unimplemented features. But this is a miscompile (the compiler succeeds, but the code runs and produces an incorrect result). It is absolutely never okay to knowingly commit a miscompile into the tree, since they are extremely difficult to debug if they ever reach production.

You will need to either implement this feature or guard it so that compilation fails with a useful diagnostic.

test/Conversion/TorchToLinalg/elementwise.mlir
```diff
@@ -94,3 +94,89 @@ func.func @insufficient_dims_for_triu(%arg0: !torch.vtensor<[?],f32>) -> !torch.
   %0 = torch.aten.triu %arg0, %int0 : !torch.vtensor<[?],f32>, !torch.int -> !torch.vtensor<[?],f32>
   return %0 : !torch.vtensor<[?],f32>
 }
+
+// -----
+
+// CHECK-LABEL:   func.func @torch.aten.quantize_per_tensor(
```
silvasean, line 100: per our testing policy do not include these tests: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

silvasean (changes requested):

e2e_testing/xfail_sets.py
mgehre-amd: Sure, how about I add an explicit error "unimplemented feature" into this PR, and attempt to provide a proper solution in the next PR?

test/Conversion/TorchToLinalg/elementwise.mlir
mgehre-amd: Understood, thanks for the link! I will remove those tests.

lib/Dialect/Torch/IR/TorchTypes.cpp
mgehre-amd: I was torn on this question. Just having quantize_per_tensor + int_repr is useful to the networks that we are looking at right now. I didn't want to delay this PR by making it bigger with the quantized add.
My initial plan was to only add scale and zero point together with quantized add in another PR.

Would you agree to changing the type to (tensor, scale, zero point) in this PR, but delay quantized add to another PR?

silvasean: I think we need to implement quantized add end-to-end in the first PR. There are other things besides just converting the types that we want to get right from the start (materializations, file structure for quantized operators, etc.). I don't think we will have full visibility of the consequences of changing the type to (tensor, scale, zero point) without implementing quantized add end-to-end.

e2e_testing/xfail_sets.py
silvasean: That sounds good to me.

lib/Dialect/Torch/IR/TorchTypes.cpp
mgehre-amd: Hm, then I will need to figure out how this works in MLIR. I first tried to go replace `!torch.vtensor<[3],!torch.qint8>`
by an mlir::Tuple `(tensor<[3],int8>, int8, f32, int8)` in `ValueTensorType::toBuiltinTensor`, but there seem to be no operations on an mlir::Tuple like insert_element or extract_element, see [1].
Then I tried replacing the type in a 1-n fashion using `typeConverter.addConversion([](Torch::ValueTensorType type, SmallVectorImpl<Type> &Tys)`, but there is no support for mapping then 1-n values in the rewriter infra [2]; the workaround being tuple unpacking.

I'm afraid that implementing this would take a while.

[1] https://mlir.llvm.org/docs/Rationale/Rationale/#tuple-types
[2] https://discourse.llvm.org/t/replacing-a-result-value-with-multiple-values-error/3221/2

silvasean: Yes, the dialect conversion framework is notoriously difficult to use. I even gave a talk about it at one point: [slides](https://drive.google.com/file/d/1FVbzCXxZzS9LBLuvpPNLWJD-XDkt54ky/view?usp=sharing) - [recording](https://drive.google.com/file/d/1VfVajitgf8ZPnd-HRkJvaJiFLhBsluXN/view?usp=sharing)

This use case is supported by the dialect conversion framework though. Doing this with a tuple type is not the right way. Registering a direct 1-n conversion is the correct way. You can find guidance in the SCF dialect's StructuralTypeConversions.cpp. In particular the manual calling of ` typeConverter->materializeSourceConversion` for 1-n conversions here: https://github.com/llvm/llvm-project/blob/fefe655baafb9aa11ae3e2a34b19aef1f47e2b8d/mlir/lib/Dialect/SCF/Transforms/StructuralTypeConversions.cpp#L85 You can ask on #mlir on the llvm discord for further guidance.

mgehre-amd: It seems that `quantized.add` is not supported on CPU backends. I don't know how I can make end2end tests work with that.
```
class TestModule(torch.nn.Module):
  def forward(self, x, y):
    a = torch.quantize_per_tensor(x, 1.0, 0, torch.qint8)
    b = torch.quantize_per_tensor(y, 1.0, 0, torch.qint8)
    c = torch.ops.quantized.add(a, y, 1.0, 0)
    return c

args = (torch.FloatTensor([0, 1]),torch.FloatTensor([0, 1]))

model = TestModule()
print(model(*args))
```
gives
```
NotImplementedError: Could not run 'aten::qscheme' with arguments from the 'CPU' backend. This could be because the operator doesn't exist for this backend, ...
```

PR #1797, opened 2023-01-12
Only use dynamic padding when necessary in convolution op

gpetters94:
This adds back the old static logic when the padding is static for optimization purposes.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -506,12 +506,21 @@ public:
       return rewriter.create<arith::IndexCastOp>(loc, intType, v);
     };
 
+    bool staticPadding = false;
     SmallVector<Value> paddingIntValues;
-    if (!getListConstructElements(op.getPadding(), paddingIntValues))
-      return rewriter.notifyMatchFailure(
-          op, "only support padding from a list construct");
-    paddingIntValues = getTypeConvertedValues(rewriter, loc, getTypeConverter(),
-                                              paddingIntValues);
+    SmallVector<int64_t> paddingInts;
+    if (matchPattern(op.getPadding(), m_TorchListOfConstantInts(paddingInts))) {
+      staticPadding = true;
+      paddingIntValues = getAsConstantIntValues(rewriter, loc, paddingInts);
+    } else {
+      staticPadding = false;
+      if (!getListConstructElements(op.getPadding(), paddingIntValues))
+        return rewriter.notifyMatchFailure(
+            op, "only support padding from a list construct");
+      paddingIntValues = getTypeConvertedValues(
+          rewriter, loc, getTypeConverter(), paddingIntValues);
+    }
+
     SmallVector<int64_t> strideInts;
     if (!matchPattern(op.getStride(), m_TorchListOfConstantInts(strideInts)))
       return rewriter.notifyMatchFailure(op,
@@ -651,8 +660,16 @@ public:
 
     } else {
       // Pad input
-      paddedInput = torch_to_linalg::getDynamicZeroPaddedTensor(
-          op, rewriter, input, paddingIntValues, /*unpaddedDims=*/2);
+      if (staticPadding) {
+        SmallVector<int64_t, 4> paddingIncludingNC = {0, 0};
+        paddingIncludingNC.insert(paddingIncludingNC.end(), paddingInts.begin(),
+                                  paddingInts.end());
+        paddedInput = torch_to_linalg::getZeroPaddedTensor(op, rewriter, input,
+                                                           paddingIncludingNC);
+      } else {
+        paddedInput = torch_to_linalg::getDynamicZeroPaddedTensor(
+            op, rewriter, input, paddingIntValues, /*unpaddedDims=*/2);
+      }
 
       // Calculate output dims
       for (size_t i = 0; i < numSpacialDims; i++)


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -506,12 +506,21 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
-      return rewriter.notifyMatchFailure(
-          op, "only support padding from a list construct");
-    paddingIntValues = getTypeConvertedValues(rewriter, loc, getTypeConverter(),
-                                              paddingIntValues);
+    SmallVector<int64_t> paddingInts;
+    if (matchPattern(op.getPadding(), m_TorchListOfConstantInts(paddingInts))) {
+      staticPadding = true;
+      paddingIntValues = getAsConstantIntValues(rewriter, loc, paddingInts);
+    } else {
+      staticPadding = false;
```
pashu123, line 516: unnecessary.

pashu123 (approved):
LGTM!

silvasean (changes requested):
This should be canonicalized later in the stack. Why do we need to do it here?

pashu123:
> This should be canonicalized later in the stack. Why do we need to do it here?

The shapes are still dynamic in the linalg-on-tensors backend. Somehow the shape-canonicalization/calculation is not happening.

pashu123:
> This should be canonicalized later in the stack. Why do we need to do it here?

@silvasean Any suggestions?

gpetters94:
> > This should be canonicalized later in the stack. Why do we need to do it here?
> 
> The shapes are still dynamic in the linalg-on-tensors backend. Somehow the shape-canonicalization/calculation is not happening.

Yep, I've confirmed that it's happening for both convolution and convolution_backward - it's producing dynamic shapes in the linalg output.

silvasean:
> Somehow the shape-canonicalization/calculation is not happening.

That seems like a bug in those downstream projects. It is nothing for Torch-MLIR to worry about.

gpetters94:
> > Somehow the shape-canonicalization/calculation is not happening.
> 
> That seems like a bug in those downstream projects. It is nothing for Torch-MLIR to worry about.

I've tested and it's happening in torch-mlir, so unless this is expected behavior then this looks like it's not a downstream issue.

powderluv:
FYI. We just enabled torch-mlir by default for SHARK users and looks like this bug causes a perf degradation.

gpetters94:
Here you go @ramiro050 https://gist.github.com/gpetters94/d8e48803a79794bd785df773a3c4c672. Let me know if you need the torch IR, it's very big.

The issue is 

```
    %padded_1176 = tensor.pad %1631 low[%c0, %c0, %c1, %c1] high[%c0, %c0, %c1, %c1] {
    ^bb0(%arg3: index, %arg4: index, %arg5: index, %arg6: index):
      tensor.yield %cst_695 : f32
    } : tensor<2x2560x16x16xf32> to tensor<?x?x?x?xf32>
    %1632 = linalg.generic {indexing_maps = [#map7, #map8], iterator_types = ["parallel", "parallel", "parallel", "parallel"]} ins(%cst_345 : tensor<1280xf32>) outs(%792 : tensor<2x1280x16x16xf32>) {
    ^bb0(%in: f32, %out: f32):
      linalg.yield %in : f32
    } -> tensor<2x1280x16x16xf32>
    %1633 = linalg.conv_2d_nchw_fchw {dilations = dense<1> : vector<2xi64>, strides = dense<1> : vector<2xi64>} ins(%padded_1176, %cst_346 : tensor<?x?x?x?xf32>, tensor<1280x2560x3x3xf32>) outs(%1632 : tensor<2x1280x16x16xf32>) -> tensor<2x1280x16x16xf32>
```

The intermediate `tensor<?x?x?x?xf32>` is what is causing the perf issue.

powderluv:
@ramiro050 @silvasean any guidance on this issue would be much appreciated. It is blocking down stream general deployment of torch-mlir. Thanks

ramiro050:
After playing around with this, it seems part of the problem has to do with the fact that dynamic shapes for the result type are being hardcoded

https://github.com/llvm/torch-mlir/blob/27a31157b9ee4154dd2846e8991395ec3a697ecd/lib/Conversion/TorchToLinalg/Utils.cpp#L92-L95

The other issue is a lack of folding for the sake of calculating the result type. In particular, here

https://github.com/llvm/torch-mlir/blob/27a31157b9ee4154dd2846e8991395ec3a697ecd/lib/Conversion/TorchToLinalg/Linear.cpp#L513-L514

you can do `rewriter.createOrFold<torch::TorchConversion::ToI64Op>(...)` instead to perform the type conversion, which makes sure the padding ints are folded when possible (converting to i64 is safe to do since the convolution op only takes int values for padding).

Similarly, here

https://github.com/llvm/torch-mlir/blob/27a31157b9ee4154dd2846e8991395ec3a697ecd/lib/Conversion/Utils/Utils.cpp#L143

you can do `b.createOrFold<arith::IndexCastOp>(...)` instead.

Lastly, if you change the result type of the `PadOp`

https://github.com/llvm/torch-mlir/blob/27a31157b9ee4154dd2846e8991395ec3a697ecd/lib/Conversion/TorchToLinalg/Utils.cpp#L101

to `Type()`, this will make the `PadOp` [infer the result type](https://github.com/llvm/llvm-project/blob/9acc2f37bdfce08ca0c2faec03392db10d1bb7a9/mlir/lib/Dialect/Tensor/IR/TensorOps.cpp#L2560-L2562). If the padding can be folded into constants, this should make the result type have static sizes.

@silvasean, what do you think of this approach?

powderluv:
@ramiro050 thanks for the review. @gpetters94 lets see if we can address these comments quickly

silvasean:
> @silvasean, what do you think of this approach?

Those seem like reasonable improvements and could work, but it still seems suboptimal (doesn't seem ideal to test, for example). I don't believe that TorchToLinalgPass guarantees that if the input is statically shaped that the output IR is statically shaped (and indeed sometimes it can be a development burden to maintain that invariant, as in this case). We run ResolveShapedTypeResultDimsPass and canonicalization afterward which I believe re-establishes all the static shapes for most cases, and some of the IR in this thread (like pad with constants) seems like missing resolution at that level. Looking at the IR posted by @gpetters94 suggests that making those later cleanup passes kick in for `tensor.pad` would give the desired statically shaped IR.

Note that the linalg-on-tensors backend contract does not guarantee static shapes, nor do we guarantee that statically shaped torch backend contract IR will produce statically shaped IR at the linalg-on-tensors backend contract. If we want to change that to a guarantee I think we would need to change our testing strategy. For example, to structure our e2e tests to verify that invariant, and always ensure that for every dynamically shaped test we have an analogous statically shaped test (this might be nicely automatable once we are 100% on the Dynamo path by having all e2e tests statically shaped and optionally erasing shapes during FX import to create dynamically shaped tests).

powderluv:
I think it would be worthwhile changing it to a guarantee that static backendIR generates static linalg-on-tensors. We can help make the improvements to testing to enable that.

gpetters94:
> > @silvasean, what do you think of this approach?
> 
> Those seem like reasonable improvements and could work, but it still seems fragile, so I would be opposed to solving the problem that way unless my suggestion below does not work.
> 
> I don't believe that TorchToLinalgPass guarantees that if the input is statically shaped that the output IR is statically shaped (and indeed sometimes it can be a development burden to maintain that invariant, as in this case). We run ResolveShapedTypeResultDimsPass and canonicalization afterward which I believe re-establishes all the static shapes for most cases, and some of the IR in this thread (like pad with constants) seems like missing resolution at that level. Looking at the IR posted by @gpetters94 suggests that making those later cleanup passes kick in for `tensor.pad` would give the desired statically shaped IR.
> 
> Note that the linalg-on-tensors backend contract does not guarantee static shapes, nor do we guarantee that statically shaped torch backend contract IR will produce statically shaped IR at the linalg-on-tensors backend contract. If we want to change that to a guarantee I think we would need to change our testing strategy. For example, to structure our e2e tests to verify that invariant, and always ensure that for every dynamically shaped test we have an analogous statically shaped test (this might be nicely automatable once we are 100% on the Dynamo path by having all e2e tests statically shaped and optionally erasing shapes during FX import to create dynamically shaped tests).

If I'm reading the code correctly, isn't the `ResolveShapedTypeResultDimsPass` run on the memref level? Doesn't that happen after the TorchToLinalg pipeline?

silvasean:
> If I'm reading the code correctly, isn't the `ResolveShapedTypeResultDimsPass` run on the memref level? Doesn't that happen after the TorchToLinalg pipeline?

It is run at the tensor level. I still don't understand why it is in the memref namespace.

> I think it would be worthwhile changing it to a guarantee that static backendIR generates static linalg-on-tensors. We can help make the improvements to testing to enable that.

Okay, but it is going to be a significant amount of work to change our testing infra to cover this (on the order of auditing and adjusting every single e2e test combined with API design work, CI changes, and implementing the actual verification in a nicely layered way) .

raikonenfnu:
> > If I'm reading the code correctly, isn't the `ResolveShapedTypeResultDimsPass` run on the memref level? Doesn't that happen after the TorchToLinalg pipeline?
> 
> It is run at the tensor level. I still don't understand why it is in the memref namespace.
> 
> > I think it would be worthwhile changing it to a guarantee that static backendIR generates static linalg-on-tensors. We can help make the improvements to testing to enable that.
> 
> Okay, but it is going to be a significant amount of work to change our testing infra to cover this (on the order of auditing and adjusting every single e2e test combined with API design work, CI changes, and implementing the actual verification in a nicely layered way) .

Hey @silvasean and @ramiro050 , I had a chat with @gpetters94 about this issue. I think `ResolveShapedTypeResultDimsPass` is used for dynamic ops where memref.dim/tensor.dim is used to extract the dimension size. This may not work for our current case because we are feeding the dynamic tensor output of tensor.pad directly into linalg.conv2d. 

But I do agree on your point of canonicalizing later on, as this output is technically legal, and adding a canonicalization/cleanup upstream would allow fixes for other MLIR stacks as well. 

I am thinking a good candidate solution is for @gpetters94 to add a new canonicalization pattern into `PadOp:: getCanonicalizationPatterns()` [here](https://github.com/llvm/llvm-project/blob/cb05c2ffc79eefe74c569263e27fcb5fad167ba3/mlir/lib/Dialect/Tensor/IR/TensorOps.cpp#L2844-L2848). The proposed pattern to add will analyze that it the output doesn't need to be dynamic and replace with the same tensor.pad Op but with static output. Any thoughts/opinions on this solution? :)

ramiro050:
> I am thinking a good candidate solution is for @gpetters94 to add a new canonicalization pattern into PadOp:: getCanonicalizationPatterns() [here](https://github.com/llvm/llvm-project/blob/cb05c2ffc79eefe74c569263e27fcb5fad167ba3/mlir/lib/Dialect/Tensor/IR/TensorOps.cpp#L2844-L2848). The proposed pattern will analyze whether the output needs to be dynamic. If not, it will replace the padOp with an equivalent padOp but with static output. Any thoughts/opinions on this solution? :)

Yeah, that sounds like the right solution to me. A similar thing is already being done for the [`tensor.empty` op](https://github.com/llvm/llvm-project/blob/cb05c2ffc79eefe74c569263e27fcb5fad167ba3/mlir/lib/Dialect/Tensor/IR/TensorOps.cpp#L576-L587), where the inputs are analyzed to see if they come from constants, and if so the result type is updated with the new static information.

gpetters94:
Okay, LLVM patch is up here: https://reviews.llvm.org/D143135 and I've verified that it fixes the issue. Closing this as unnecessary.

powderluv:
Just want to say thank you to everyone for resolving this issue. It just went live with our users and here is the before/after:

Before 1.6s/it

image



After 3.36 it/s 
image

ramiro050:
That's a pretty big improvement! Awesome work

PR #1798, opened 2023-01-12
Enable VerifyBackendContract in LTC backend

GlebKazantaev:
### Description
In this PR I enable `VerifyBackendContract` pass as a part of LTC backend contract to check that MLIR satisfies backend requirements. See https://github.com/llvm/torch-mlir/issues/1460 for more details.

### Changes
* Update `TorchMlirLoweringContext::Build` with `VerifyBackendContract` pass execution.
* Updated `TorchMlirComputation` to accept `MlirModule`.
* Removed useless `VerifyBackendContract` pass parametrization.

This PR depends on changes in PyTorch https://github.com/pytorch/pytorch/pull/92095 so once they land I will mark this PR as ready for review.

### DIFF
--- a/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
+++ b/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
@@ -289,15 +289,12 @@ class VerifyBackendContractPass
     : public VerifyBackendContractBase<VerifyBackendContractPass> {
 public:
   VerifyBackendContractPass() = default;
-  VerifyBackendContractPass(bool decompose,
-                            ArrayRef<std::string> backendLegalOps) {
-    this->decompose = decompose;
-    this->backendLegalOps = backendLegalOps;
-  }
+
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     ConversionTarget target =
-        getBackendContractTarget(context, decompose, backendLegalOps);
+        getBackendContractTarget(context, /*decompose*/false,
+                                 /*backendLegalOps*/{});
 
     if (!satisfiesBackendContract(getOperation(), target,
                                   /*actuallyEmitDiagnostics=*/true)) {
@@ -315,10 +312,8 @@ mlir::torch::Torch::createLowerToBackendContractPass(
 }
 
 std::unique_ptr<OperationPass<ModuleOp>>
-mlir::torch::Torch::createVerifyBackendContractPass(
-    bool decompose, ArrayRef<std::string> backendLegalOps) {
-  return std::make_unique<VerifyBackendContractPass>(decompose,
-                                                     backendLegalOps);
+mlir::torch::Torch::createVerifyBackendContractPass() {
+  return std::make_unique<VerifyBackendContractPass>();
 }
 
 // The backend contract guarantees that ops with decompositions available will
--- a/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
+++ b/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
@@ -285,10 +285,10 @@ public:
   }
 };
 
-class VerifyBackendContractPass
-    : public VerifyBackendContractBase<VerifyBackendContractPass> {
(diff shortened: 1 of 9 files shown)


GlebKazantaev:
@ramiro050 I just realized that you've added parameterization for `VerifyBackendContract` pass here: https://github.com/llvm/torch-mlir/pull/1705
Just curios, in which case we need to check that IR was decomposed? In my understanding `decompose` is an option of `LowerToBackendContract` pass which take care about all the checks after lowering and `VerifyBackendContract` pass should check the types consistency, shape existence, value semantics (at least by default).

I also found that in generated CAPI there is no way to specify arguments. All arguments are just embedded to the cpp file. For example in `./torch-mlir/build/cmake_build/tools/torch-mlir/include/torch-mlir/Dialect/Torch/Transforms/Transforms.capi.cpp.inc` :
```cpp
MlirPass mlirCreateLowerToBackendContract(void) {
  return wrap(
    mlir::torch::Torch::createLowerToBackendContractPass(
      /*maxIterations=*/10, /*decompose=*/true, /*backendLegalOps=*/{})
  .release());
}
```
all parameters were set, and `mlirCreateLowerToBackendContract` doesn't take any arguments. The same thing I see for all other MLIR passes. So probably it smth not yet supported... Do you have any ideas?

As a workaround we can have two separate passes for `decompose=true` and `decompose=false`.

python/torch_mlir/csrc/base_lazy_backend/utils/jit_utils.cpp
```diff
@@ -0,0 +1,45 @@
...
+      output_type = IntType::get();
+    } else if (c10::isFloatingType(*scalar_type)) {
+      node_type = c10::aten::FloatImplicit;
+      output_type = FloatType::get();
+    } else {
+      throw std::runtime_error(
+        "Expected isIntegralType or isFloatingType");
+    }
+
+    Value * output = graph
```
silvasean, line 34: Is this spacing correct?

silvasean (approved):

silvasean:
LGTM pending Ramiro's response on the options.

ramiro050:
> @ramiro050 I just realized that you've added parameterization for VerifyBackendContract pass here: https://github.com/llvm/torch-mlir/pull/1705
Just curios, in which case we need to check that IR was decomposed? In my understanding decompose is an option of LowerToBackendContract pass which take care about all the checks after lowering and VerifyBackendContract pass should check the types consistency, shape existence, value semantics (at least by default).

The parameters were added to be consistent with [the definition of the backend contract](https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#the-backend-contract), which includes decompositions. However, the `verifyBackendContract` pass is only used by LTC, so I think we should just rename the pass to `verifyBackendContractAssumingNoDecompositions`.

GlebKazantaev:
@ramiro050 I renamed the pass. Could you please take a look?

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -285,19 +285,16 @@ class LowerToBackendContractPass
...
-    this->decompose = decompose;
-    this->backendLegalOps = backendLegalOps;
-  }
+  VerifyBackendContractNoDecompositionsPass() = default;
+
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     ConversionTarget target =
-        getBackendContractTarget(context, decompose, backendLegalOps);
+        getBackendContractTarget(context, /*decompose*/false,
```
ramiro050, line 296: nit: these comments should have an `=`, to match the style. In other words, it should be `/*decompose=*/`. Same for `backendLegalOps`

ramiro050 (approved):
Just a small nit

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
GlebKazantaev: Im working on another PR now to cleanup some changes. So let me fix this to in this PR.

PR #1799, opened 2023-01-15
infer aten.topk's dtypes

qingyunqu:


lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -891,6 +891,16 @@ void TypeAnalysis::visitOperation(Operation *op,
     return;
   }
 
+  if (isa<AtenTopkOp>(op)) {
```
ramiro050, line 894: There is currently no support of `AtenTopKOp` in any of the backends of torch-mlir, so this code is not being tested properly.

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/RefineTypes.cpp
qingyunqu: Yes, we have custom backend for `AtenTopKOp`. Is there any suggestion to test it?

ramiro050: At the moment there is no way to add this. However, once the [transition to dtype functions](https://github.com/llvm/torch-mlir/issues/1807) is complete, you will be able to add this and test it without needing to have a lowering for the op in torch-mlir

qingyunqu: how about only add the dtype infer and remove the E2E test？

ramiro050: In general, we expect every addition to torch-MLIR to be e2e tested, since this is the only way right now that we can ensure that the behavior of torch-mlir is in accordance with PyTorch's behavior.

ramiro050: Thinking about this a bit more, since `AtenTopKOp` does tend to occur many times in a model, you can add it's dtype function into `main`. This will allow you to add tests that ensure your dtype logic agrees with pytorch. See: https://github.com/llvm/torch-mlir/blob/main/docs/adding_abstract_interpretation_functions.md#step-by-step-guide

li-plus: @qingyunqu maybe you could decompose topk into sort+slice?

Issue #1615, opened 2022-11-18
Find a better solution for backends that don't support i64/f64


silvasean:
Some backends, such as IREE and TOSA (by today's spec) don't support i64 or f64 (or it is very expensive). This bug is to start collecting use cases and motivation to better support reducing bit width for these targets in Torch-MLIR.

The general theme is that PyTorch's semantics are the 64-bit semantics, but sometimes it is "common sense" to want to drop those down to smaller bit widths. E.g. for a 100kB model, a user might be quite comfortable to use a 32-bit integer to hold the tensor sizes, even if technically it is supposed to be a 64-bit integer.

As we think about this, there are 4 basic cases that probably need separate treatment:

1. Scalar floats (`!torch.float`). This represents the PyTorch/TorchScript float type which is 64-bit.
    - Thoughts: I don't have specific insight here, but there are likely many scenarios where 64-bit is overkill (e.g. when deploying to a microcontroller). Some scientific codes might care about some factor being very precise though.
3. Scalar ints (`!torch.int`). This represents the PyTorch/TorchScript int type which is specced as 64-bit in TorchScript (though the Python type is arbitrary precision)
    - Thoughts: Often these are used to calculate view sizes and such.  With large language models in the 100's of GB these days, we cannot arbitrarily use 32-bit indexing though (though perhaps individual tensor dimensions remain in 32-bit range?).
4. Tensors with 64-bit floating point numbers (`!torch.tensor<[10,12],f64>`). This represents tensor computations on f64.
    - Thoughts: PyTorch defaults to f32, so if a user asks for f64 they probably actually want the extra precision (?).
5. Tensors with 64-bit integers (e.g. `!torch.tensor<[10,12],si64>`). This is probably most common for embedding indices.
    - Thoughts: Most embeddings are likely OK to index with 32-bit indices, but they seem to be getting larger and larger, and it is not out of the question to need 64-bit indices there.

We need to discuss with the PyTorch devs and see their thinking on this and align on a solution.

silvasean:
I've created a PyTorch dev discussion: https://dev-discuss.pytorch.org/t/how-to-approach-targets-that-dont-support-i64-f64/867

AmosLewis:
https://github.com/llvm/torch-mlir/pull/1802

PR #1802, opened 2023-01-16
[TOSA] Add torch.prim.NumToTensor.Scalar float support

AmosLewis:
Find this in OPT model : https://github.com/nod-ai/SHARK/issues/865
> Also find this f64 support issue in GPT2 and distilGPT2 https://github.com/nod-ai/SHARK/issues/494 with transformer 4.25.1

This patch needs the f64 support of tosa https://reviews.llvm.org/D142599

For f64 support test file and debug output: https://gist.github.com/AmosLewis/fca6b0d16ee325fcf7ee400459f4fd40

Related issues: https://github.com/llvm/torch-mlir/issues/1615


AmosLewis:
@eric-k256 @ramiro050 is there any possibility to add f64 in tosa. I got it from GPT2/distilGPT2 model torchscript with transformer package 4.25.2. I have no idea how to walk around this unless I downgrade transformer version to 4.21.

ramiro050 (commented):
I would've expected the e2e test for `NumToTensor` to pass on TOSA now

eric-k256:
For TOSA, we're looking into adding something like a 'compiler profile' that would allow a wider set of data types, with a corresponding validation pass that would fail networks that are using f64/i64 with devices that don't expect it. If we added this to TOSA, and expanded the types in the dialect, would this be enough for you? Also, are your passes that consume TOSA capable of handling f64? I'd still recommend against f64 unless it is requested by the network developer, as it's likely to be slower on almost every device.

AmosLewis:
> For TOSA, we're looking into adding something like a 'compiler profile' that would allow a wider set of data types, with a corresponding validation pass that would fail networks that are using f64/i64 with devices that don't expect it. If we added this to TOSA, and expanded the types in the dialect, would this be enough for you? Also, are your passes that consume TOSA capable of handling f64? I'd still recommend against f64 unless it is requested by the network developer, as it's likely to be slower on almost every device.

What I need is at least if we use cast, it can walk around this issue.Fot i64 I use to use tosa::cast i64 to i32, then at the end of the work, I use tosa::cast i32 back to i64. It work. But for f64, cast fail. As you can see in the following:
With tosa::cast f32 to f64 code(has been deleted in this patch):
```
  auto outElemTy = resultType.getElementType();
  if (outElemTy.isF64()) {
    auto resultF32 = tosa::getConstTensor<float>(rewriter, op, floatValue, {}).value();
    rewriter.replaceOpWithNewOp<tosa::CastOp>(op, resultType, resultF32);
  }
```
Bug Output: 
```
// *** IR Dump After Pattern Application ***
mlir-asm-printer: Verifying operation: func.func
ImplicitTypeIDRegistry::lookupOrInsert(mlir::InferShapedTypeOpInterface::Trait<Empty>)
'tosa.cast' op result #0 must be tensor of number values, but got 'tensor<f64>'
mlir-asm-printer: 'func.func' failed to verify and will be printed in generic form
"func.func"() ({
  %0 = "torch.constant.float"() {value = 8.000000e+00 : f64} : () -> !torch.float
  %1 = "builtin.unrealized_conversion_cast"(%0) : (!torch.float) -> f64
  %2 = "tosa.const"() {value = dense<8.000000e+00> : tensor<f32>} : () -> tensor<f32>
  %3 = "tosa.cast"(%2) : (tensor<f32>) -> tensor<f64>
  %4 = "torch.prim.NumToTensor.Scalar"(%0) : (!torch.float) -> !torch.vtensor<[],f64>
  "func.return"(%4) : (!torch.vtensor<[],f64>) -> ()
}) {function_type = () -> !torch.vtensor<[],f64>, sym_name = "torch.prim.NumToTensor.Scalar"} : () -> ()


} -> SUCCESS
//===-------------------------------------------===//

//===-------------------------------------------===//
Legalizing operation : 'func.return'(0x55a53fab0c30) {
  "func.return"(%4) : (!torch.vtensor<[],f64>) -> ()

  * Fold {
  } -> FAILURE : unable to fold
} -> FAILURE : no matched legalization pattern
//===-------------------------------------------===//
** Insert  : 'torch_c.from_builtin_tensor'(0x55a53fb17c60)
ImplicitTypeIDRegistry::lookupOrInsert(mlir::detail::PreservedAnalyses::AllAnalysesType)
/tmp/NumToTensor.mlir:3:8: error:   'tosa.cast' op result #0 must be tensor of number values, but got 'tensor<f64>'
  %1 = "torch.prim.NumToTensor.Scalar"(%float8.000000e00) : (!torch.float) -> !torch.vtensor<[],f64>
       ^
/tmp/NumToTensor.mlir:3:8: note: see current operation: %2 = "tosa.cast"(%1) : (tensor<f32>) -> tensor<f64>

```

AmosLewis:
> I would've expected the e2e test for `NumToTensor` to pass on TOSA now

I guess they pass is because by default it generated f32. So no f64 issue.

AmosLewis:
Let's at least enable the f32 for now.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3350,13 +3350,25 @@ LogicalResult ConvertAtenOp<PrimNumToTensorScalarOp>::matchAndRewrite(
...
+  auto isFloat = matchPattern(op.getA(), m_TorchConstantFloat(&doubleValue));
+  int64_t intValue;
+  auto isInt = matchPattern(op.getA(), m_TorchConstantInt(&intValue));
+  if (!isFloat && !isInt)
+    return rewriter.notifyMatchFailure(op,
+                                       "Unable to extract the scalar constant");
+  float floatValue = static_cast<float>(doubleValue);
+
+  auto outElemTy = resultType.getElementType();
+  if (outElemTy.isa<mlir::IntegerType>() || outElemTy.isF32()) {
```
ramiro050, line 3363: `PrimNumToTensorScalarOp` does not produce `f32` tensors.

```python
In [5]: torch.ops.prim.NumToTensor(0.0).dtype
Out[5]: torch.float64
```

so this is not the right lowering for this op. Also, e2e tests are needed anytime a new op is extended. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

ramiro050 (changes requested):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: Oh, it's XFAIL - "NumToTensorFloatModule_basic"

AmosLewis:
> For TOSA, we're looking into adding something like a 'compiler profile' that would allow a wider set of data types, with a corresponding validation pass that would fail networks that are using f64/i64 with devices that don't expect it. If we added this to TOSA, and expanded the types in the dialect, would this be enough for you? Also, are your passes that consume TOSA capable of handling f64? I'd still recommend against f64 unless it is requested by the network developer, as it's likely to be slower on almost every device.

@eric-k256  Is there any possibility to add at least f64 in` tosa.cast`? For the new model [facebook/opt-125m](https://huggingface.co/facebook/opt-125m) I am working on. I definitely have no way to walk around the f64 support. Something like: 
%2 = "tosa.cast"(%1) : (tensor<f32>) -> tensor<f64>
%3 = "tosa.cast"(%2) : (tensor<f64>) -> tensor<f32>

eric-k256:
I would prefer to start with the smallest possible scope for f64. If we only want to support it for cast, we could create a new type that is used for the inputs and outputs of tosa.cast instead of Tosa_Tensor, it could be something like Tosa_CastableTensor, which included all types from Tosa_Tensor and f64. That would allow us to manage the use of f64, avoiding it for most cases.

AmosLewis:
the f64 support of tosa https://reviews.llvm.org/D142599

AmosLewis:
> I would prefer to start with the smallest possible scope for f64. If we only want to support it for cast, we could create a new type that is used for the inputs and outputs of tosa.cast instead of Tosa_Tensor, it could be something like Tosa_CastableTensor, which included all types from Tosa_Tensor and f64. That would allow us to manage the use of f64, avoiding it for most cases.

Done. Please review in llvm

AmosLewis:
```
module attributes {torch.debug_module_name = "NumToTensorFloatModule"} {
  func.func @forward() -> !torch.vtensor<[],f64> {
    %float1.000000e00 = torch.constant.float 1.000000e+00 loc(#loc1)
    %0 = torch.prim.NumToTensor.Scalar %float1.000000e00 : !torch.float -> !torch.vtensor<[],f64> loc(#loc2)
    return %0 : !torch.vtensor<[],f64> loc(#loc)
  } loc(#loc)
} loc(#loc)
```


```
module attributes {torch.debug_module_name = "NumToTensorFloatModule"} {
  func.func @forward() -> !torch.vtensor<[],f64> {
    %float1.000000e00 = torch.constant.float 1.000000e+00
    %0 = "tosa.const"() {value = dense<1.000000e+00> : tensor<f32>} : () -> tensor<f32>
    %1 = "tosa.cast"(%0) : (tensor<f32>) -> tensor<f64>
    %2 = torch_c.from_builtin_tensor %1 : tensor<f64> -> !torch.vtensor<[],f64>
    return %2 : !torch.vtensor<[],f64>
  }
}
```

AmosLewis:
Since this https://reviews.llvm.org/rGa2dcd994a7f8cc33640f58105276b78acf3483e5 has been merged and update in torch-mlir. This patch should be ok to merged now.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
ramiro050: I don't see this as having been addressed

sjw36:
I am trying to solve a similar problem with f64 vtensor.literals in the dynamo flow. The rewrite already has to convert signed integers to unsigned, pretty simple to also convert f64 to f32. 
And in general if TOSA is not going to support F64, would it be valid to make the typeConverter convert to f32?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis:
```
module attributes {torch.debug_module_name = "NumToTensorFloatModule"} {
  func.func @forward() -> !torch.vtensor<[],f64> {
    %float1.000000e00 = torch.constant.float 1.000000e+00 loc(#loc1)
    %0 = torch.prim.NumToTensor.Scalar %float1.000000e00 : !torch.float -> !torch.vtensor<[],f64> loc(#loc2)
    return %0 : !torch.vtensor<[],f64> loc(#loc)
  } loc(#loc)
} loc(#loc)
```
`➜  torch-mlir git:(indexput) ✗ torch-mlir-opt -convert-torch-to-tosa /tmp/NumToTensorFloatModule.mlir       `
```
 module attributes {torch.debug_module_name = "NumToTensorFloatModule"} {
  func.func @forward() -> !torch.vtensor<[],f64> {
    %float1.000000e00 = torch.constant.float 1.000000e+00
    %0 = "tosa.const"() {value = dense<1.000000e+00> : tensor<f32>} : () -> tensor<f32>
    %1 = "tosa.cast"(%0) : (tensor<f32>) -> tensor<f64>
    %2 = torch_c.from_builtin_tensor %1 : tensor<f64> -> !torch.vtensor<[],f64>
    return %2 : !torch.vtensor<[],f64>
  }
}
```

AmosLewis: Got it, Need to change the tosa::constOp to support Tosa_Tensor_Cast f64.
def Tosa_ConstOp : Tosa_Op<"const", ...
  let results = (outs
    Tosa_Tensor_Cast:$output
  );

AmosLewis:
> I am trying to solve a similar problem with f64 vtensor.literals in the dynamo flow. The rewrite already has to convert signed integers to unsigned, pretty simple to also convert f64 to f32. And in general if TOSA is not going to support F64, would it be valid to make the typeConverter convert to f32?

Just one line change the tosa::constOp definition to support Tosa_Tensor_Cast f64 should be enough.
def Tosa_ConstOp : Tosa_Op<"const", ...
let results = (outs
**Tosa_Tensor_Cast**:$output
);

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: Done

AmosLewis: Done

AmosLewis:
https://reviews.llvm.org/D145336 this has been merged and update in torch-mlir. Please review @ramiro050

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3718,13 +3718,27 @@ LogicalResult ConvertAtenOp<PrimNumToTensorScalarOp>::matchAndRewrite(
...
-    return rewriter.notifyMatchFailure(
-        op, "unimplemented: input should be a torch constant int");
+  double doubleValue;
+  auto isFloat = matchPattern(op.getA(), m_TorchConstantFloat(&doubleValue));
+  int64_t intValue;
+  auto isInt = matchPattern(op.getA(), m_TorchConstantInt(&intValue));
+  if (!isFloat && !isInt)
+    return rewriter.notifyMatchFailure(op,
+                                       "Unable to extract the scalar constant");
+  float floatValue = static_cast<float>(doubleValue);
```
ramiro050, line 3728: This should not be casted to lower precision. This leads to a loss of information

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3718,13 +3718,27 @@ LogicalResult ConvertAtenOp<PrimNumToTensorScalarOp>::matchAndRewrite(
...
+
+  auto outElemTy = resultType.getElementType();
+  if (outElemTy.isa<mlir::IntegerType>() || outElemTy.isF32()) {
+    DenseElementsAttr constAttr =
+        isInt ? DenseElementsAttr::get(resultType, {intValue})
+              : DenseElementsAttr::get(resultType, {floatValue});
+    rewriter.replaceOpWithNewOp<tosa::ConstOp>(op, resultType, constAttr);
+  } else if (outElemTy.isF64()) {
+    auto resultF32 =
+        tosa::getConstTensor<float>(rewriter, op, floatValue, {}).value();
```
ramiro050, line 3738: this should be an f64 constant

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3718,13 +3718,27 @@ LogicalResult ConvertAtenOp<PrimNumToTensorScalarOp>::matchAndRewrite(
...
+  auto isFloat = matchPattern(op.getA(), m_TorchConstantFloat(&doubleValue));
+  int64_t intValue;
+  auto isInt = matchPattern(op.getA(), m_TorchConstantInt(&intValue));
+  if (!isFloat && !isInt)
+    return rewriter.notifyMatchFailure(op,
+                                       "Unable to extract the scalar constant");
+  float floatValue = static_cast<float>(doubleValue);
+
+  auto outElemTy = resultType.getElementType();
+  if (outElemTy.isa<mlir::IntegerType>() || outElemTy.isF32()) {
```
ramiro050, line 3731: outElemTy is never f32 for this op

ramiro050 (changes requested):

ramiro050 (approved):

PR #1804, opened 2023-01-16
[MLIR][TORCH] Add decomposition for aten.randn_like op

vivekkhandelwal1:
This commit decomposes aten.randn_like op into aten.randn.generator op.


python/torch_mlir_e2e_test/test_suite/rng.py
```diff
@@ -364,3 +364,46 @@ def forward(self):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float64, True),
+    ])
+    def forward(self, x):
+        a = torch.ops.aten.randn_like(x, dtype=torch.float32)
```
pashu123, line 402: Any reason we can pass x as type `torch.float32`?

vivekkhandelwal1: `torch.float64` is passed as the default value for numerical stability of the result of the mean/std op for large-sized tensors.

pashu123 (approved):
LGTM!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3526,6 +3526,26 @@ class DecomposeAtenRandnOp : public OpRewritePattern<AtenRandnOp> {
...
 } // namespace
 
+namespace {
+// Decompose `aten.randn_like` op into `aten.randn.generator` op.
+class DecomposeAtenRandnLikeOp : public OpRewritePattern<AtenRandnLikeOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenRandnLikeOp op,
+                                PatternRewriter &rewriter) const override {
+    Value none = rewriter.create<Torch::ConstantNoneOp>(op.getLoc());
```
ramiro050, line 3536: handling for the `memory_format` argument of `randn_like` is missing

ramiro050 (approved):
Small nit, but other than that it LGTM

PR #1812, opened 2023-01-18
update llvm to d23516e9ad477527a9db4d06b1fa9566680ac67c

eric-k256:
Rename BlockAndValueMapping to IRMapping
*TODO* Dropped TypesMatchedWith Trait from Torch_PrimTupleConstructOp.
  Upstream change https://reviews.llvm.org/D141231 creates a new
inferReturnTypes method that fails to compile for this Op. It appears to not support a mix of Variadic operands/non-variadic return type.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
@@ -376,9 +376,6 @@ def Torch_PrimListUnpackOp: Torch_Op<"prim.ListUnpack", [
 
 def Torch_PrimTupleConstructOp: Torch_Op<"prim.TupleConstruct", [
     Pure,
-    TypesMatchWith<"contained types correspond to operand types",
-    "elements", "result", "Torch::TupleType::get($_ctxt, llvm::to_vector<6>($_self))",
-    "isValidSubtype">,
     AllowedInModuleInitializer,
   ]> {
   let summary = "TorchScript prim::TupleConstruct op";
@@ -397,6 +394,8 @@ def Torch_PrimTupleConstructOp: Torch_Op<"prim.TupleConstruct", [
   let assemblyFormat = [{
     $elements attr-dict `:` qualified(type($elements)) `->` qualified(type($result))
   }];
+
+  let hasVerifier = 1;
 }
 
 def Torch_PrimListConstructOp: Torch_Op<"prim.ListConstruct", [
--- a/externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorInterfaces.h
+++ b/externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorInterfaces.h
@@ -10,7 +10,7 @@
 #ifndef TORCH_MLIR_DIALECTS_DIALECT_TMTENSOR_IR_TMTENSORINTERFACES_H_
 #define TORCH_MLIR_DIALECTS_DIALECT_TMTENSOR_IR_TMTENSORINTERFACES_H_
 
-#include "mlir/IR/BlockAndValueMapping.h"
+#include "mlir/IR/IRMapping.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/OpDefinition.h"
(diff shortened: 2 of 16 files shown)


include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -376,9 +376,6 @@ def Torch_PrimListUnpackOp: Torch_Op<"prim.ListUnpack", [
 
 def Torch_PrimTupleConstructOp: Torch_Op<"prim.TupleConstruct", [
     Pure,
-    TypesMatchWith<"contained types correspond to operand types",
-    "elements", "result", "Torch::TupleType::get($_ctxt, llvm::to_vector<6>($_self))",
-    "isValidSubtype">,
```
eric-k256, line 381: I'm not sure what the right replacement is for this trait. With this trait, this is the build error:
```
/staging/torch-mlir/build/tools/torch-mlir/include/torch-mlir/Dialect/Torch/IR/TorchOps.cpp.inc:79294:66: error: no matching function for call to 'to_vector'
  ::mlir::Type odsInferredType0 = Torch::TupleType::get(context, llvm::to_vector<6>(operands[0].getType()));
                                                                 ^~~~~~~~~~~~~~~~~~
/staging/torch-mlir/externals/llvm-project/llvm/include/llvm/ADT/SmallVector.h:1298:46: note: candidate template ignored: substitution failure [with Size = 6, R = mlir::Type]: no matching function for call to 'begin'
SmallVector<ValueTypeFromRangeType<R>, Size> to_vector(R &&Range) {
                                             ^
/staging/torch-mlir/externals/llvm-project/llvm/include/llvm/ADT/SmallVector.h:1302:40: note: candidate template ignored: invalid explicitly-specified argument for template parameter 'R'
SmallVector<ValueTypeFromRangeType<R>> to_vector(R &&Range) {
                                       ^
```

ramiro050: Hm, the use of `to_vector` here is a bit confusing to me, since `$_self` is a single type, not a range of types. 

What if you remove the `llvm::to_vector<6>` part and just do `{$_self}`? `TupleType::get` takes an `ArrayRef` as the second argument, so I'm assuming the list will get converted to the right type. Alternatively you can also use the `ArrayRef` constructor: `ArrayRef<...>{$_self}`

eric-k256: No luck. Fixing the problem on the currently indicated line (either solution), breaks a previous line in verifyInvariantsImpl(). That is using getODSOperands(0), which returns a ::mlir::Operation::operand_range which has no cast to ArrayRef<::mlir::Type>.

 verifyInvariantsImpl has $_self = getODSOperands(0).getType() and inferReturnTypes uses $_self = operands[0].getType().

```
/staging/torch-mlir/build/tools/torch-mlir/include/torch-mlir/Dialect/Torch/IR/TorchOps.cpp.inc:79282:83: error: no matching conversion for functional-style cast from 'mlir::OperandRange::type_range' (aka 'ValueTypeRange<mlir::OperandRange>') to 'ArrayRef< ::mlir::Type>'
  if (!((isValidSubtype(Torch::TupleType::get(this->getOperation()->getContext(), ArrayRef<::mlir::Type>(this->getODSOperands(0).getType())), (*this->getODSResults(0).begin()).getType()))))
                                                                                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ramiro050: After talking to @jpienaar, he mentioned that a solution for now could be to implement the infer type interface for the `PrimTupleConstructOp`, which would prevent the autogenerated version that is causing the error. The long term solution is to add support for variadic operand + non-variadic return type in `TypesMatchedWith` upstream

eric-k256: I tried adding the infer type interface to the op: `DeclareOpInterfaceMethods<InferTypeOpInterface>`, but that doesn't seem to have any effect on the creation of the inferReturnType method. I'll try to find why this is still happening and possibly we can discuss at the meeting on Monday.
I agree on the long term solution.

jpienaar: elements is `Variadic<AnyTorchType>:$elements`, so I'd expect $_self to be a range here, but it actually becomes operands[0] instead of getODSOperands(0) with the former a value and the latter a range. Do you need anything other than the verification for this? If not, I'd just move this to a custom verify for the purposes of the integrate and follow up upstream.

eric-k256: Yes, I think this is the right answer. I wasn't sure what the trait was giving us, but it appears to only be used for verification based on the tablegened files.

powderluv:
Was this PR reviewed by anyone ? Please get a review before merge, thanks

eric-k256:
Sorry about merging prior to review. I wouldn't have done it for a change of mine, but wanted to get the llvm update caught up. 
It was the wrong choice, I should have waited for review.

powderluv:
no worries. thank you for taking on this bump.

PR #1814, opened 2023-01-19
Make `getTypeForScalarType` safer by returning `FailureOr<Type>`

ramiro050:
One of the potential values for a `torch_upstream::ScalarType` is `Undefined`. This means that conversion of a `ScalarType` to another type is a computation that can fail. To enforce handling of the failure case, this commit makes the two helper functions that convert `ScalarType`s into other types return `failure()` when the `ScalarType` is `Undefined`.


mgehre-amd:
Looks great!

mgehre-amd (approved):

Issue #1819, opened 2023-01-21
Windows DLL import error 


powderluv:
On a Windows system follow https://github.com/llvm/torch-mlir#install-torch-mlir-snapshot to install the latest torch-mlir windows snapshot in a venv and activate it. 

```
mkdir tm_test
cd .\tm_test\
python -m venv tm.venv
.\tm.venv\Scripts\activate
pip install --pre torch-mlir torchvision -f https://llvm.github.io/torch-mlir/package-index/ --extra-index-url https://download.pytorch.org/whl/nightly/cpu

```

Download https://raw.githubusercontent.com/llvm/torch-mlir/main/examples/torchscript_resnet18.py and run it 

```
(tm.venv) PS C:\g\tm_test> python .\torchscript_resnet18.py
Traceback (most recent call last):
  File "C:\g\tm_test\torchscript_resnet18.py", line 15, in <module>
    import torch_mlir
  File "C:\g\tm_test\tm.venv\lib\site-packages\torch_mlir\__init__.py", line 15, in <module>
    from torch_mlir.passmanager import PassManager
  File "C:\g\tm_test\tm.venv\lib\site-packages\torch_mlir\passmanager.py", line 5, in <module>
    from ._mlir_libs._mlir.passmanager import *
  File "C:\g\tm_test\tm.venv\lib\site-packages\torch_mlir\_mlir_libs\__init__.py", line 107, in <module>
    _site_initialize()
  File "C:\g\tm_test\tm.venv\lib\site-packages\torch_mlir\_mlir_libs\__init__.py", line 56, in _site_initialize
    from ._mlir import ir
ImportError: DLL load failed while importing _mlir: The specified module could not be found.

```

powderluv:
@ashay fyi

PR #1820, opened 2023-01-22
Use delvewheel to package windows DLLs

powderluv:
Use the delvewheel to package our windows whl. It is similar to auditwheel in Linux.

Test: Import and run the samples via torch-mlir on windows

Fixes https://github.com/llvm/torch-mlir/issues/1819


pashu123 (approved):

stellaraccident:
Cool. Such a tool had been discussed for a long time but I didn't know someone has done it.

PR #1822, opened 2023-01-23

vivekkhandelwal1:
Set PyTorch and TorchVision version to nightly release 2023-01-22.


pashu123 (approved):
LGTM!

pashu123:
@vivekkhandelwal1 Please add another commit with `Add default decompositions for .... ops in the dynamo backend`.

pashu123 (changes requested):
.

pashu123:
Also, I think the PyTorch update would be taken care by `Roll PyTorch Action`, you could just add a commit with default decompositions and that would be enough.

vivekkhandelwal1:
> Also, I think the PyTorch update would be taken care by `Roll PyTorch Action`, you could just add a commit with default decompositions and that would be enough.

Actually, one of the decompositions is needed for the PyTorch update. https://github.com/llvm/torch-mlir/actions/runs/3979807311

PR #1825, opened 2023-01-23
[MHLO] Support cumsum op in mhlo backend

li-plus:
This PR supports lowering `aten.cumsum` op to the MHLO backend. The static-shape cumsum e2e test has passed in MHLO config.

### DIFF
--- a/python/torch_mlir_e2e_test/test_suite/basic.py
+++ b/python/torch_mlir_e2e_test/test_suite/basic.py
@@ -3035,6 +3035,23 @@ class CumsumStaticModule(torch.nn.Module):
 def CumsumStaticModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(2, 7, 4))
 
+class CumsumStaticNegativeDimModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([2, 7, 4], torch.float32, True),
+    ])
+    def forward(self, val):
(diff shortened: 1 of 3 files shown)


li-plus:
@ZihengJiang Would you please take a look at this PR? Thanks.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -3029,10 +3027,10 @@ def __init__(self):
         ([2, 7, 4], torch.float32, True),
     ])
     def forward(self, val):
-        return torch.ops.aten.cumsum(val, 1)
+        return torch.ops.aten.cumsum(val, dim=-2)
```
ramiro050, line 3030: drive by comment: In general, existing tests should not be modified. E2E tests are relatively cheap, so if you want to test an edge case, you can just add a new test. Having more coverage of edge cases is always welcomed.

Also, I don't think we need to rename to `AtenCumsum...` the convention so far has been to not include the `Aten` part, although `Aten_EmbeddingBag...` above did slip through the cracks in this regard.

ZihengJiang:
LGTM. Could you fix @ramiro050 's comment?

python/torch_mlir_e2e_test/test_suite/basic.py
li-plus: Okay I get it. Now I keep the existing tests unchanged and have added a new test.

ZihengJiang (approved):

PR #1827, opened 2023-01-23
CI: build torch-mlir python package for Python v3.8

ashay:
Previously, torchvision had not released WHL files for Python v3.8,
causing failures in torch-mlir python package builds, so we had disabled
building for Python v3.8.

Now that the WHL files are back, this patch re-enables v3.8 builds.

### DIFF
--- a/build_tools/python_deploy/build_linux_packages.sh
+++ b/build_tools/python_deploy/build_linux_packages.sh
@@ -44,7 +44,7 @@ TM_RELEASE_DOCKER_IMAGE="${TM_RELEASE_DOCKER_IMAGE:-stellaraccident/manylinux201
 # ./build_tools/docker/Dockerfile
 TM_CI_DOCKER_IMAGE="${TM_CI_DOCKER_IMAGE:-powderluv/torch-mlir-ci:latest}"
 # Version of Python to use in Release builds. Ignored in CIs.
-TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp310-cp310}"
+TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp38-cp38 cp310-cp310}"
 # Location to store Release wheels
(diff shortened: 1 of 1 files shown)


powderluv (approved):

build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -44,7 +44,7 @@ TM_RELEASE_DOCKER_IMAGE="${TM_RELEASE_DOCKER_IMAGE:-stellaraccident/manylinux201
 # ./build_tools/docker/Dockerfile
 TM_CI_DOCKER_IMAGE="${TM_CI_DOCKER_IMAGE:-powderluv/torch-mlir-ci:latest}"
 # Version of Python to use in Release builds. Ignored in CIs.
-TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp310-cp310}"
+TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp38-cp38 cp310-cp310}"
```
wangkuiyi, line 47: Does this script only build for Linux? Could we release for macOS too?

wangkuiyi: Could we release for macOS and Python 3.9? When we install Xcode or the Command Line Tools on macOS Ventura, the default version is 3.9 (https://discussions.apple.com/thread/254320722).

ashay: I'm not sure if there was some other reason for not building v3.9 on macOS.

@powderluv, I see that you bumped the version in commit 2f0b1d0b0.  Is it the slow build that prevents us from building for multiple versions?

powderluv: Yeah we can build just one version on macOS GHA builders without getting killed so we chose 3.10 (which we hope can bump up 3.11) in the near future.

ashay: I see.  Thanks!

Issue #1807, opened 2023-01-16
Push for finishing transition to Python dtype functions


ramiro050:
Hi everyone,

I would like to get some help from the community to finish up the transition from using [`RefineTypes.cpp`](https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/RefineTypes.cpp) for encoding how to calculate the resulting dtypes of the tensors of all the ops to using dtype functions that are written in Python. For more information about the rationale behind this transition see the [Custom Op RFC](https://github.com/llvm/torch-mlir/issues/1462).

The transition will take place on the branch [`dtype-functions-staging`](https://github.com/llvm/torch-mlir/tree/dtype-functions-staging), so that it happens in a single go in `main`, otherwise many large workloads can fail to lower through Torch-MLIR. The reason things can fail is that if you have some ops being handled by `RefineTypes` and other ops using the Python dtype functions that get applied in the [`TorchDtypeRefinementPipeline`](https://github.com/llvm/torch-mlir/blob/4f94831fed2aa016f8d3f03b63f8841a064c4152/lib/Dialect/Torch/Transforms/Passes.cpp#L175), then the `RefineTypes` pass and the `TorchDtypeRefinementPipeline` will cycle over and over, each pass propagating dtype information until encountering an op that is not handled by the current pass. For workloads with many ops, the cycling of passes can easily reach the [iteration limit](https://github.com/llvm/torch-mlir/blob/4f94831fed2aa016f8d3f03b63f8841a064c4152/include/torch-mlir/Dialect/Torch/Transforms/Passes.td#L294) in the `LowerToBackendContract` pass, resulting in a failure when lowering through Torch-MLIR. Currently, there are only a few ops using dtype functions in `main`, and it is already [affecting some user workloads](https://github.com/llvm/torch-mlir/pull/1750#issuecomment-1375370220).

To avoid breaking workloads for users, or forcing users to increase the iteration limit affecting performance of Torch-MLIR, we will instead do the entire transition on the branch [`dtype-functions-staging`](https://github.com/llvm/torch-mlir/tree/dtype-functions-staging), where we can `XFAIL` complex tests temporarily as needed.

### The contribution process

1. Choose an item from the work list below and edit this issue by adding your name alongside the checklist item you're working on to avoid duplicate work
   - If you don't have edit access, just mention it as a comment, and I will update the issue
[...]

Issue #1669, opened 2022-12-01
Some ops that take f16 tensor inputs require GPU to run in e2e tests


ramiro050:
There are some ops in PyTorch that don't have a CPU implementation for `f16` inputs. For example:

```python
In [15]: a = torch.rand((2,4,3))

In [16]: torch.ops.aten.native_layer_norm(a.to(torch.float16), [2,4,3], a.to(torch.float64), a
    ...: .to(torch.float64), 0.0)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Input In [16], in <cell line: 1>()
----> 1 torch.ops.aten.native_layer_norm(a.to(torch.float16), [2,4,3], a.to(torch.float64), a.to(torch.float64), 0.0)

File ~/torch-mlir/venv-torch-mlir/lib/python3.10/site-packages/torch/_ops.py:446, in OpOverloadPacket.__call__(self, *args, **kwargs)
    441 def __call__(self, *args, **kwargs):
    442     # overloading __call__ to ensure torch.ops.foo.bar()
    443     # is still callable from JIT
    444     # We save the function ptr as the `op` attribute on
    445     # OpOverloadPacket to access it here.
--> 446     return self._op(*args, **kwargs or {})

RuntimeError: "LayerNormKernelImpl" not implemented for 'Half'
```

Because currently the CIs only use CPUs, there is no way of testing `f16` support of these ops e2e. We should have a CI that has access to a GPU and add support to the e2e testing library for specifying device in order to ensure correctness of the `f16` implementations.

ramiro050:
Copying over @powderluv's [message](https://github.com/llvm/torch-mlir/pull/1623#issuecomment-1334260825) for reference:

> yes easy to add a GPU builder/tester. However there is logistics / cost with standing it up since it has to be done by someone with admin access to the LLVM project. We can find a gpu VM for the purpose.
> 
> However -- another option would be to mark the GPU tests experimental and not run as part of the CI but available for anyone to try locally.

powderluv:
I think we can get an VM with GPUs on Google Cloud, but the LLVM org has to add a self-hosted runner.

powderluv:
The gpu builder is online as `a100`. I will try to move the current builds / tests to it on a test branch. @ashay fyi

PR #1831, opened 2023-01-27
Add dtype functions for two tensor promotion ops

ramiro050:
This commit adds dtype functions for ops in RefineTypes under the category of "Promote the two dtypes". The only ops not added here are convolution ops, since they take an optional tensor argument, and the dtype pipeline currently does not correctly handle that case. I will add a follow up patch fixing this.

This commit also adds two helper functions that perform a very thorough testing of dtype functions. The helper function `_check_two_tensor_op` is able to independently test invalid input dtypes and invalid output dtypes.

Lastly, this commit also XFAILs "MobilenetV3Module_basic".


ramiro050:
This is for https://github.com/llvm/torch-mlir/issues/1807

ramiro050:
@li-plus, this is the patch with the new helper functions to test the dtype functions.

ramiro050:
@li-plus can you review this patch, since you're familiar with most of the changes?

li-plus:
@ramiro050 LGTM, except for the float16 dtype. I think the shape function shouldn't raise an error for float16 since it is implemented on cuda.

ramiro050:
> @ramiro050 LGTM, except for the float16 dtype. I think the shape function shouldn't raise an error for float16 since it is implemented on cuda.

Because our testing framework in the CI currently runs only on CPU, we need to raise an error for f16. However, there are discussions about moving the CI testing to GPU instead. See: https://github.com/llvm/torch-mlir/issues/1669. If that happens in the future, we can remove the f16 assert checks

li-plus:
Ok I see, but the problem is that we have lots of gpu models to convert using torch-mlir. Most of them are `half()`ed and full of float16 ops. Currently we are running torch-mlir with cuda-version pytorch to convert these models, and float16 can be correctly inferred by `RefineTypes.cpp`.

If shape functions raise at float16, all our pipelines will fail. So I was wondering if we could skip float16 checking on cpu machine in CI, and we don't need to raise an error for float16. In fact, float16 always fail on cpu, so maybe there is no point testing it.

qingyunqu:
Hi, we strongly recommend to support float16 in shape infer and dtype infer. At least not to raise an error.

ramiro050:
It turns out that if we perform the dtype checks using `meta` tensors, we can avoid this issue of `cpu` not supporting `f16`.

```python
In [6]: torch.tanh(a)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Input In [6], in <cell line: 1>()
----> 1 torch.tanh(a)

RuntimeError: "tanh_vml_cpu" not implemented for 'Half'

In [7]: torch.tanh(a.to("meta"))
Out[7]: tensor(..., device='meta', size=(3, 4, 5), dtype=torch.float16)
```

I will create a PR soon adding this change.

qingyunqu:
How about the progress of pytorch's migration to `meta` tensor? I have commited some ops's `meta` migration to pytorch two years ago.
BTW, is there any discussion about implementing dynamic shape in torch-mlir? Including dynamic inputs and dynamic ops(like non_zero)?

ramiro050:
> How about the progress of pytorch's migration to meta tensor? I have commited some ops's meta migration to pytorch two years ago.

Are `meta` tensors not finished in PyTorch?

> BTW, is there any discussion about implementing dynamic shape in torch-mlir? Including dynamic inputs and dynamic ops(like non_zero)?

What do you mean by dynamic inputs and dynamic ops? We currently support at the torch dialect level dynamic sizes for the tensors. In the case of `torch.nonzero`, the op should work without a problem in the case where `as_tuple=False`. It's just a matter of the backends to torch-mlir being able to handle the behavior.

The case of `as_tuple=True` is not yet supported in torch-mlir, since currently need to have a known size at compile time. As far as I know, there is currently no discussion for adding support for tuples that don't have a known size at compile time. However, if this is something that is needed, we can start such a discussion

PR #1832, opened 2023-01-27
[TOSA] Add aten.cumsum tosa support

AmosLewis:
Find this issue when lower opt model: https://github.com/nod-ai/SHARK/issues/865

[opt_torchbackend.mlir](https://gist.github.com/AmosLewis/3faccdf32c91d30f21daed12bf7197b4#file-opt_torchbackend-mlir)

https://pytorch.org/docs/stable/generated/torch.cumsum.html

Related patch cumsum in TMTensor https://github.com/llvm/torch-mlir/pull/1320

```
func.func @torch.aten.to.dtype(%arg0: !torch.vtensor<[1,6],si64>) -> !torch.vtensor<[1,6],si64> {
  %int1 = torch.constant.int 1
  %none = torch.constant.none
  %0 =  torch.aten.cumsum %arg0, %int1, %none : !torch.vtensor<[1,6],si64>, !torch.int, !torch.none -> !torch.vtensor<[1,6],si64>
  return %0 : !torch.vtensor<[1,6],si64>
}
```

`torch-mlir-opt -convert-torch-to-tosa /tmp/cumsum.mlir`

```
module {
  func.func @torch.aten.to.dtype(%arg0: !torch.vtensor<[1,6],si64>) -> !torch.vtensor<[1,6],si64> {
    %0 = torch_c.to_builtin_tensor %arg0 : !torch.vtensor<[1,6],si64> -> tensor<1x6xi64>
    %int1 = torch.constant.int 1
    %none = torch.constant.none
    %1 = "tosa.slice"(%0) {size = array<i64: 1, 1>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x1xi64>
    %2 = "tosa.reduce_sum"(%1) {axis = 1 : i64} : (tensor<1x1xi64>) -> tensor<1x1xi64>
    %3 = "tosa.slice"(%0) {size = array<i64: 1, 2>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x2xi64>
    %4 = "tosa.reduce_sum"(%3) {axis = 1 : i64} : (tensor<1x2xi64>) -> tensor<1x1xi64>
    %5 = "tosa.slice"(%0) {size = array<i64: 1, 3>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x3xi64>
    %6 = "tosa.reduce_sum"(%5) {axis = 1 : i64} : (tensor<1x3xi64>) -> tensor<1x1xi64>
    %7 = "tosa.slice"(%0) {size = array<i64: 1, 4>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x4xi64>
    %8 = "tosa.reduce_sum"(%7) {axis = 1 : i64} : (tensor<1x4xi64>) -> tensor<1x1xi64>
    %9 = "tosa.slice"(%0) {size = array<i64: 1, 5>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x5xi64>
    %10 = "tosa.reduce_sum"(%9) {axis = 1 : i64} : (tensor<1x5xi64>) -> tensor<1x1xi64>
    %11 = "tosa.slice"(%0) {size = array<i64: 1, 6>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x6xi64>
    %12 = "tosa.reduce_sum"(%11) {axis = 1 : i64} : (tensor<1x6xi64>) -> tensor<1x1xi64>
    %13 = "tosa.concat"(%2, %4, %6, %8, %10, %12) {axis = 1 : i64} : (tensor<1x1xi64>, tensor<1x1xi64>, tensor<1x1xi64>, tensor<1x1xi64>, tensor<1x1xi64>, tensor<1x1xi64>) -> tensor<1x6xi64>
    %14 = torch_c.from_builtin_tensor %13 : tensor<1x6xi64> -> !torch.vtensor<[1,6],si64>
    return %14 : !torch.vtensor<[1,6],si64>
  }
}
```

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -3491,6 +3491,68 @@ LogicalResult ConvertAtenOp<AtenWhereSelfOp>::matchAndRewrite(
 }
 
 template <>
+LogicalResult ConvertAtenOp<AtenCumsumOp>::matchAndRewrite(
+    AtenCumsumOp op, OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+
+  // Not a tensor type.
+  auto selfType = adaptor.getSelf().getType().dyn_cast<TensorType>();
+  auto inputShape = selfType.getShape();
+  if (!selfType)
+    return rewriter.notifyMatchFailure(
+        op, "Only tensor types input are currently supported");
+  int64_t dim;
+  if (!matchPattern(op.getDim(), m_TorchConstantInt(&dim)))
+    return rewriter.notifyMatchFailure(op,
+                                       "non-const dim parameter unsupported");
+
+  dim = toPositiveDim(dim, selfType.getRank());
+
+  if (!isValidDim(dim, selfType.getRank()))
+    return rewriter.notifyMatchFailure(op, "dim must be less than tensor rank");
+
+  if (!op.getDtype().getType().template isa<Torch::NoneType>())
+    return rewriter.notifyMatchFailure(op,
+                                       "Only default dtype none is supported");
+
+  auto outType = getTypeConverter()->convertType(op.getType());
+
+  // cumsum = concat(reduceSum(slice(0:1)) , ... , reduceSum(slice(0:n))))
+  size_t cumsumDimShape = inputShape[dim];
+  SmallVector<Value> sliceReduceSumValues; // used as concat inputs
+  SmallVector<int64_t> startSlice(selfType.getRank(), 0);
+  SmallVector<int64_t> sizeSlice =
+      llvm::to_vector(makeShapeTorchCompatible(inputShape));
+  SmallVector<int64_t> reduceSumShape =
+      llvm::to_vector(makeShapeTorchCompatible(inputShape));
+  reduceSumShape[dim] = 1;
+  startSlice[dim] = 0;
+
+  for (size_t step = 1; step <= cumsumDimShape; step++) {
+    sizeSlice[dim] = step; // 1:cumsumDimShape
+    Value sliceValue = rewriter.create<tosa::SliceOp>(
+        op->getLoc(),
+        RankedTensorType::get(makeShapeLLVMCompatible(sizeSlice),
+                              selfType.getElementType()),
+        adaptor.getSelf(), rewriter.getDenseI64ArrayAttr(startSlice),
+        rewriter.getDenseI64ArrayAttr(sizeSlice));
+    Value reduceSumValue = rewriter.create<tosa::ReduceSumOp>(
+        op->getLoc(),
+        RankedTensorType::get(makeShapeLLVMCompatible(reduceSumShape),
+                              selfType.getElementType()),
+        sliceValue, rewriter.getI64IntegerAttr(dim));
+
+    sliceReduceSumValues.push_back(reduceSumValue);
+  }
+
+  // tosa::concat sliceValues, outputshape
+  rewriter.replaceOpWithNewOp<tosa::ConcatOp>(op, outType, sliceReduceSumValues,
+                                              rewriter.getI64IntegerAttr(dim));
+
+  return success();
+}
+
+template <>
 LogicalResult ConvertAtenOp<AtenClampOp>::matchAndRewrite(
     AtenClampOp op, OpAdaptor adaptor,
     ConversionPatternRewriter &rewriter) const {
@@ -4435,6 +4497,7 @@ public:
     INSERT_ATENOP_PATTERN(AtenGatherOp);
     INSERT_ATENOP_PATTERN(AtenIndexTensorOp);
     INSERT_ATENOP_PATTERN(AtenWhereSelfOp);
+    INSERT_ATENOP_PATTERN(AtenCumsumOp);
     INSERT_ATENOP_PATTERN(AtenClampOp);
     INSERT_ATENOP_PATTERN(AtenArangeStartStepOp);
     INSERT_ATENOP_PATTERN(PrimNumToTensorScalarOp);
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -632,6 +632,8 @@ TOSA_PASS_SET = {
     "ArgmaxModule_with_dim",
     "_LogSoftmaxModuleStable_basic",
     "ElementwiseAtenWhereSelfModule_basic",
+    "CumsumStaticModule_basic",
+    "CumsumStaticNegativeDimModule_basic",
     "ElementwiseUnsqueezeBroadcastModule_basic",
     "MaskedFillScalarIntValueModule_basic",
     "MaskedFillScalarIntValueStaticModule_basic",


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -4307,6 +4333,7 @@ class ConvertTorchToTosa : public ConvertTorchToTosaBase<ConvertTorchToTosa> {
     INSERT_ATENOP_PATTERN(AtenGatherOp);
     INSERT_ATENOP_PATTERN(AtenIndexTensorOp);
     INSERT_ATENOP_PATTERN(AtenWhereSelfOp);
+    INSERT_ATENOP_PATTERN(AtenCumsumOp);
```
ramiro050, line 4336: Drive by comment: these types of changes should *always* be accompanied by at least one passing e2e test (if the op is complex, more than one is often expected). See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3364,6 +3364,32 @@ LogicalResult ConvertAtenOp<AtenWhereSelfOp>::matchAndRewrite(
...
+    return rewriter.notifyMatchFailure(op,
+                                       "non-const dim parameter unsupported");
+
+  if (!op.getDtype().getType().template isa<Torch::NoneType>())
+    return rewriter.notifyMatchFailure(op,
+                                       "Only default dtype none is supported");
+
+  auto outType = getTypeConverter()->convertType(op.getType());
+  rewriter.replaceOpWithNewOp<tosa::ReduceSumOp>(
+      op, outType, adaptor.getSelf(), rewriter.getI64IntegerAttr(cumsumDim));
```
eric-k256, line 3388: This doesn't look right. TOSA ReduceSum is a different operator than cumsum. The TOSA operator is a reduction, so the output tensor along the reduced axis has size 1, where cumsum is still size N, with each value in the summation axis has the partial sum to that point. I don't have an immediate trivial sequence of TOSA operators that can replace cumsum. 

An end to end test with this legalization should fail due to this problem. I'll try to think of a sequence that would work.

AmosLewis: Yes, this implementation is wrong. So I change this patch to a draft. Still working on this.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
AmosLewis: Yes, I will add one.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
eric-k256: Someone on my team pointed out that the best legalization is to construct a constant matrix with appropriate weights of 0 or 1, and then you can use TOSA MATMUL. For >2d, you'd have to flatten the original tensor, but if you only need to handle 1 or 2d tensors, it's not that bad.

AmosLewis: I am planning to so tosa::Slice(start=0, step=1:dimShape) + tosa::ReduceSum(sliced tensor, dim) + tosa::Concat(ReduceSum tensors) to solve it once for all.

AmosLewis:
```
(mlir_venv) nod% torch-mlir-opt -convert-torch-to-tosa /tmp/CumsumIntStaticModule.mlir
module attributes {torch.debug_module_name = "CumsumIntStaticModule"} {
  func.func @forward(%arg0: !torch.vtensor<[2,7,4],si64>) -> !torch.vtensor<[2,7,4],si64> {
    %0 = torch_c.to_builtin_tensor %arg0 : !torch.vtensor<[2,7,4],si64> -> tensor<2x7x4xi64>
    %int1 = torch.constant.int 1
    %none = torch.constant.none
    %1 = "tosa.slice"(%0) {size = array<i64: 2, 1, 4>, start = array<i64: 0, 0, 0>} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %2 = "tosa.reduce_sum"(%1) {axis = 1 : i64} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %3 = "tosa.slice"(%0) {size = array<i64: 2, 2, 4>, start = array<i64: 0, 0, 0>} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %4 = "tosa.reduce_sum"(%3) {axis = 1 : i64} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %5 = "tosa.slice"(%0) {size = array<i64: 2, 3, 4>, start = array<i64: 0, 0, 0>} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %6 = "tosa.reduce_sum"(%5) {axis = 1 : i64} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %7 = "tosa.slice"(%0) {size = array<i64: 2, 4, 4>, start = array<i64: 0, 0, 0>} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %8 = "tosa.reduce_sum"(%7) {axis = 1 : i64} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %9 = "tosa.slice"(%0) {size = array<i64: 2, 5, 4>, start = array<i64: 0, 0, 0>} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %10 = "tosa.reduce_sum"(%9) {axis = 1 : i64} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %11 = "tosa.slice"(%0) {size = array<i64: 2, 6, 4>, start = array<i64: 0, 0, 0>} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %12 = "tosa.reduce_sum"(%11) {axis = 1 : i64} : (tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %13 = "tosa.concat"(%2, %4, %6, %8, %10, %12) {axis = 1 : i64} : (tensor<2x7x4xi64>, tensor<2x7x4xi64>, tensor<2x7x4xi64>, tensor<2x7x4xi64>, tensor<2x7x4xi64>, tensor<2x7x4xi64>) -> tensor<2x7x4xi64>
    %14 = torch_c.from_builtin_tensor %13 : tensor<2x7x4xi64> -> !torch.vtensor<[2,7,4],si64>
    return %14 : !torch.vtensor<[2,7,4],si64>
  }
}

(mlir_venv) nod% python -m e2e_testing.main -c tosa -f "CumsumIntStaticModule"                  
XFAIL - "CumsumIntStaticModule_basic"

Summary:
    Expectedly Failed: 1

```

AmosLewis:
```
func.func @torch.aten.to.dtype(%arg0: !torch.vtensor<[1,6],si64>) -> !torch.vtensor<[1,6],si64> {
  %int1 = torch.constant.int 1
  %none = torch.constant.none
  %0 =  torch.aten.cumsum %arg0, %int1, %none : !torch.vtensor<[1,6],si64>, !torch.int, !torch.none -> !torch.vtensor<[1,6],si64>
  return %0 : !torch.vtensor<[1,6],si64>
}
```

`torch-mlir-opt -convert-torch-to-tosa /tmp/cumsum.mlir`

```
module {
  func.func @torch.aten.to.dtype(%arg0: !torch.vtensor<[1,6],si64>) -> !torch.vtensor<[1,6],si64> {
    %0 = torch_c.to_builtin_tensor %arg0 : !torch.vtensor<[1,6],si64> -> tensor<1x6xi64>
    %int1 = torch.constant.int 1
    %none = torch.constant.none
    %1 = "tosa.slice"(%0) {size = array<i64: 1, 1>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x1xi64>
    %2 = "tosa.reduce_sum"(%1) {axis = 1 : i64} : (tensor<1x1xi64>) -> tensor<1x1xi64>
    %3 = "tosa.slice"(%0) {size = array<i64: 1, 2>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x2xi64>
    %4 = "tosa.reduce_sum"(%3) {axis = 1 : i64} : (tensor<1x2xi64>) -> tensor<1x1xi64>
    %5 = "tosa.slice"(%0) {size = array<i64: 1, 3>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x3xi64>
    %6 = "tosa.reduce_sum"(%5) {axis = 1 : i64} : (tensor<1x3xi64>) -> tensor<1x1xi64>
    %7 = "tosa.slice"(%0) {size = array<i64: 1, 4>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x4xi64>
    %8 = "tosa.reduce_sum"(%7) {axis = 1 : i64} : (tensor<1x4xi64>) -> tensor<1x1xi64>
    %9 = "tosa.slice"(%0) {size = array<i64: 1, 5>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x5xi64>
    %10 = "tosa.reduce_sum"(%9) {axis = 1 : i64} : (tensor<1x5xi64>) -> tensor<1x1xi64>
    %11 = "tosa.slice"(%0) {size = array<i64: 1, 6>, start = array<i64: 0, 0>} : (tensor<1x6xi64>) -> tensor<1x6xi64>
    %12 = "tosa.reduce_sum"(%11) {axis = 1 : i64} : (tensor<1x6xi64>) -> tensor<1x1xi64>
    %13 = "tosa.concat"(%2, %4, %6, %8, %10, %12) {axis = 1 : i64} : (tensor<1x1xi64>, tensor<1x1xi64>, tensor<1x1xi64>, tensor<1x1xi64>, tensor<1x1xi64>, tensor<1x1xi64>) -> tensor<1x6xi64>
    %14 = torch_c.from_builtin_tensor %13 : tensor<1x6xi64> -> !torch.vtensor<[1,6],si64>
    return %14 : !torch.vtensor<[1,6],si64>
  }
}
```

lib/Conversion/TorchToTosa/TorchToTosa.cpp
eric-k256: That should work. If you are calculating cumsum across a big tensor, you will have a big instruction stream, but the solution with MATMUL would have a big constant tensor, so they may not be that different.

silvasean: TOSA should really have a cumsum operator though since this decomposition is an algorithmic lowering decision that we should be baking in here at the frontend level.

If we cannot add that op, we should do it with the algorithm shown below with log2 reduction steps:
Run with `DEBUG = True` for more intuition for the algorithm.


```
import numpy as np

DEBUG = False

def cumsum_log2(x):
    """Compute cumsum with log2 reduction steps."""
    orig_size = x.shape[0]
    log2 = int(np.ceil(np.log2(orig_size)))
    padded_size = 2**log2
    x_padded = np.pad(x, (0, padded_size - orig_size), constant_values=0)
    cumsum = np.copy(x_padded)
    if DEBUG:
        print(f"input: {x_padded}")
    for i in range(log2):
        summand = np.reshape(np.copy(x_padded), (2**i, 2, 2**(log2 - i - 1)))
        summand[:, 1, :] = np.sum(summand[:, 0, :], axis=1, keepdims=True)
        summand[:, 0, :] = 0
        if DEBUG:
            print(f"summing: {summand.flatten()}")
        cumsum += summand.flatten()
    if DEBUG:
        print(f"result: {cumsum}")
    return cumsum[:orig_size]


def check(x):
    ours = cumsum_log2(x)
    reference = np.cumsum(x)
    if not np.allclose(ours, reference):
        print(f"FAILED: {x}")
        print(f"OURS:      {ours}")
        print(f"REFERENCE: {reference}")
    else:
        print("PASSED")

check(np.ones(7))
check(1. + np.linspace(0.0, 10.0, 6))
check(np.random.randn(600))
```

AmosLewis:
```
log2 = int(np.ceil(np.log2(orig_size))) # tosa.ceil (mlir::tosa::CeilOp)  tosa.log (mlir::tosa::LogOp) 
padded_size = 2**log2 # tosa.exp (mlir::tosa::ExpOp) 
x_padded = np.pad(x, (0, padded_size - orig_size), constant_values=0) # tosa.pad (mlir::tosa::PadOp)
summand = np.reshape(np.copy(x_padded), (2**i, 2, 2**(log2 - i - 1))) # tosa.reshape (mlir::tosa::ReshapeOp) 
summand[:, 1, :] = np.sum(summand[:, 0, :], axis=1, keepdims=True) # tosa.slice (mlir::tosa::SliceOp)
```
@silvasean  Here is the corresponding tosa ops that I can find and translate from your python code.
But the log2 is the return mlir::Value of tosa.ceil, then if I want to use it in the `for i in range(log2)`, I will have to get the exact int value of log2, so I can write a for loop in c++. **How can I get this int value from mlir::value?** 
In previous code, whenever I need to get the int  value of mlir::Value, I use something like this: 
```
  int64_t int_min;
  if (!matchPattern(op.getMin(), m_TorchConstantInt(&int_min)))
        op, "unimplemented: value `int_min` should be a torch constant int");
```
But it is not the same case in this issue. Should I try something like ` int log2 =  tosaCeilReturnValue.getValue().getSExtValue();`?
Otherwise, should we manually implement the c++ ceil and log helper function so we can get rid of using tosa::log and tosa::ceil?

eric-k256: I'll look at adding CUMSUM to TOSA. It isn't ruled out, although it does trend against the principle of "If the operator can be broken down, then we should look at the component operators." It's not an absolute rule, so it's a matter of defining the rationale for the exception.

silvasean: @AmosLewis you can manually unroll the computation in the compiler since it is only log2 steps. So you would calculate log2 inside the compiler in C++ based on the static shape (since TOSA only supports static shapes anyways).

@eric-k256 There is a strong argument here for providing the CUMSUM op, since there are many different ways to lower it. In particular, the logarithmic reduction I showed here is known to be suboptimal for many hardware, since the smaller reduction steps will likely want to be done with the equivalent of a warp-level cumsum operation, which is very difficult to reconstruct from the code we lower here. Other approaches could be to do some number of logarithmic reduction steps and then a batch matmul for the smaller ones. And if there is a bunch of parallelism in the non-summed dimensions then a more straightforward serialized cumsum algorithm would be very efficient (which is not efficiently representable in TOSA for current compilers, since it would require turning a TOSA WHILE operator into a within-dispatch loop rather than a host-side loop).

AmosLewis: @eric-k256  How long do you think it will take for adding tosa::cumsum? I am currently waiting for this op. Before the tosa::cumsum merged, I am planning to use my concat(reducesum(slice)) implementation locally and continue on fix the following up other ops(broadcast_to/....) in the OPT model.

PR #1836, opened 2023-01-30
Add in-place variant of torch.scatter_add

cetiniz:
This PR adds the in-place version of the aten.scatter_add op.


ramiro050 (approved):
Thanks!

PR #1838, opened 2023-01-31
[Tcp] Add sub, mul, and clamp ops

navahgar:
This PR adds `tcp.sub`, `tcp.mul`, and `tcp.clamp` ops.

This also adds lowering from Torch to Tcp for all these ops (including `torch.aten.relu`) and from Tcp to Linalg.


externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -61,6 +93,48 @@ def Tcp_AddOp : Tcp_BinaryElementwiseOp<"add", [SameOperandsAndResultElementType
   let assemblyFormat = "$in1 `,` $in2 attr-dict `:` type($in1) `,` type($in2) `->` type($out)";
 }
 
+def Tcp_SubOp : Tcp_BinaryElementwiseOp<"sub", [SameOperandsAndResultElementType]> {
+  let summary = "Computes elementwise addition";
+
+  let description = [{
+    Computes the elementwise subtraction of `in2` from `in1`.
+
+    Same constraints as tcp.add applies to tcp.sub as well.
```
jpienaar, line 102: Are the type constraints captured in BinaryElementwiseOp? If so should this refer to that instead of add

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
```diff
@@ -59,15 +59,56 @@ createLinalgPayloadForElementwiseOp(Operation *op,
...
   if (isa<TanhOp>(op))
     return {b.create<math::TanhOp>(loc, payloadArgs[0])};
 
+  if (auto clampOp = dyn_cast<ClampOp>(op)) {
+    Value result = payloadArgs[0];
+    if (elemType.isa<mlir::FloatType>()) {
+      auto minFloat = clampOp.getMinFloat();
+      auto maxFloat = clampOp.getMaxFloat();
+      if (minFloat)
+        result = b.create<arith::MaxFOp>(
```
jpienaar, line 73: I seem to recall that for degenerative floating point numbers the order of min and max matter (but can't recall details). What is expectation around denormals if any?

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
```diff
@@ -115,5 +156,8 @@ void mlir::TcpToLinalg::populateElementwisePatternsAndLegality(
 
   target.addIllegalDialect<TcpDialect>();
   patterns.add<ConvertElementwiseOp<TanhOp>>(typeConverter, context);
+  patterns.add<ConvertElementwiseOp<ClampOp>>(typeConverter, context);
   patterns.add<ConvertElementwiseOp<AddOp>>(typeConverter, context);
+  patterns.add<ConvertElementwiseOp<SubOp>>(typeConverter, context);
```
jpienaar, line 161: Should these be in alphabetical order?

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -19,6 +19,33 @@
 #include "torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.cpp.inc"
 
 namespace mlir::tcp {
+
+LogicalResult ClampOp::verify() {
+  auto inputType = getIn().getType().dyn_cast<RankedTensorType>();
```
jpienaar, line 24: Isn't this already verified by the type constraint you use ODS side?

jpienaar (commented):
First scan

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -107,6 +127,126 @@ class ConvertAtenTanhOp : public OpConversionPattern<AtenTanhOp> {
...
+  using OpConversionPattern<AtenOpT>::OpConversionPattern;
+  using OpAdaptor = typename AtenOpT::Adaptor;
+
+  LogicalResult
+  matchAndRewrite(AtenOpT op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Value input = adaptor.getSelf();
+    RankedTensorType inputType = input.getType().dyn_cast<RankedTensorType>();
+    auto elementType = inputType.getElementType();
+    if (!inputType)
```
ramiro050, line 217: This should happen before `elementType` is defined

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -107,6 +127,126 @@ class ConvertAtenTanhOp : public OpConversionPattern<AtenTanhOp> {
...
+      return rewriter.notifyMatchFailure(
+          op, "Only Ranked Tensor types are supported in TCP");
+    if (!inputType.getElementType().isIntOrFloat())
+      return rewriter.notifyMatchFailure(
+          op,
+          "Clamp input tensor must have integer or floating-point datatype");
```
ramiro050, line 223: Should be `"Relu ...`

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -116,7 +256,20 @@ void torch_to_tcp::populateElementwisePatternsAndLegality(
 
   target.addIllegalOp<AtenTanhOp>();
   patterns.add<ConvertAtenTanhOp>(typeConverter, context);
+  target.addIllegalOp<AtenClampOp>();
+  patterns.add<ConvertAtenClampOp>(typeConverter, context);
+  target.addIllegalOp<AtenReluOp>();
+  target.addIllegalOp<AtenRelu6Op>();
+  patterns.add<ConvertAtenReluOp<AtenReluOp>>(typeConverter, context);
+  patterns.add<ConvertAtenReluOp<AtenRelu6Op>>(typeConverter, context);
```
ramiro050, line 264: Is there a reason the e2e test for `Relu6` is not passing? Every time support for a new op is added, it should be accompanied by at least one passing e2e test. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -56,3 +56,23 @@ Value torch_to_tcp::broadcastShapeInLeadingDims(
...
                                            input, dimSizes, axesAttr);
 }
+
+Value torch_to_tcp::broadcastInLeadingDimsToMatchShape(
+    ConversionPatternRewriter &rewriter, Value input, Value target) {
+  RankedTensorType targetType = target.getType().dyn_cast<RankedTensorType>();
+  assert(targetType);
+
+  RankedTensorType inputType = input.getType().dyn_cast<RankedTensorType>();
+  assert(inputType);
```
ramiro050, line 66: doing `cast` instead of `dyn_cast` performs the `assert` automatically. I would also add a comment above this function mentioning that inputs are expected to be `RankedTensorType`

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -107,6 +127,126 @@ class ConvertAtenTanhOp : public OpConversionPattern<AtenTanhOp> {
...
+          op, "Only Ranked Tensor types are supported in TCP");
+    if (!inputType.getElementType().isIntOrFloat())
+      return rewriter.notifyMatchFailure(
+          op,
+          "Clamp input tensor must have integer or floating-point datatype");
+
+    FloatAttr minFloatAttr, maxFloatAttr;
+    IntegerAttr minIntAttr, maxIntAttr;
+
+    if (isa<AtenReluOp>(op)) {
```
ramiro050, line 228: I'm not a big fan of doing this inside of a template because templates and `isa`s/`dyn_cast`s are each designed to be used under very different circumstances. Templates are for when your code looks exactly the same for all the types involved (with a few exceptions now that C++17 allows things like `if constexpr`) *and* you know all the types involved at compile time. On the other hand, `isa`s/`dyn_cast`s are for when your code does not look the same for all types *and* you don't know all the types involved at compile time.

For this particular case, you do know all the types involved at compile time, so there should be no need for `isa`s in the code. One way to fix this is doing `if constexpr (std::is_same_v<AtenOpT, AtenReluOp>()) ...`. Another way would be to not use templates and just have two classes, one for each `Relu` variant, and extract the code in common (the error checking at the beginning) into a helper function that both `Relu` classes and the `Clamp` class can use.

ramiro050 (changes requested):

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
navahgar: Thats a good question. I haven't looked into those cases. The definition of `clamp` seems to be different in different frameworks. 
* PyTorch - max followed by min (https://pytorch.org/docs/stable/generated/torch.clamp.html)
* TensorFlow - min followed by max (https://github.com/tensorflow/tensorflow/blob/v2.11.0/tensorflow/python/ops/clip_ops.py#L113)

I just left a TODO for now to handle those cases.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
```diff
@@ -59,15 +59,56 @@ createLinalgPayloadForElementwiseOp(Operation *op,
...
+      auto maxInt = clampOp.getMaxInt();
+      if (minInt)
+        result = b.create<arith::MaxSIOp>(
+            loc, result,
+            b.create<arith::ConstantIntOp>(loc, *minInt, b.getIntegerType(64)));
+      if (maxInt)
+        result = b.create<arith::MinSIOp>(
+            loc, result,
+            b.create<arith::ConstantIntOp>(loc, *maxInt, b.getIntegerType(64)));
+    }
```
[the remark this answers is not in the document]
navahgar: Sure, done.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
```diff
@@ -59,15 +59,56 @@ createLinalgPayloadForElementwiseOp(Operation *op,
...
+    if (elemType.isa<mlir::FloatType>())
+      return {b.create<arith::SubFOp>(loc, payloadArgs[0], payloadArgs[1])};
+    return {b.create<arith::SubIOp>(loc, payloadArgs[0], payloadArgs[1])};
+  }
+
+  if (isa<MulOp>(op)) {
+    if (elemType.isa<mlir::FloatType>())
+      return {b.create<arith::MulFOp>(loc, payloadArgs[0], payloadArgs[1])};
+    return {b.create<arith::MulIOp>(loc, payloadArgs[0], payloadArgs[1])};
+  }
```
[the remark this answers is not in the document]
navahgar: Good point. The ODS supports complex numbers as well.  Added an else to error out in those cases for now.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
navahgar: Done.

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
navahgar: Yes, good point. Removed this check.

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -107,6 +127,126 @@ class ConvertAtenTanhOp : public OpConversionPattern<AtenTanhOp> {
...
+  using OpConversionPattern<AtenClampOp>::OpConversionPattern;
+  using OpAdaptor = typename AtenClampOp::Adaptor;
+
+  LogicalResult
+  matchAndRewrite(AtenClampOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Value input = adaptor.getSelf();
+    RankedTensorType inputType = input.getType().dyn_cast<RankedTensorType>();
+    auto elementType = inputType.getElementType();
+    if (!inputType)
```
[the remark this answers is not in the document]
navahgar: Fixed

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -107,6 +127,126 @@ class ConvertAtenTanhOp : public OpConversionPattern<AtenTanhOp> {
...
+      } else if (matchPattern(maxValue, m_TorchConstantInt(&maxInt))) {
+        if (elementType.isa<mlir::FloatType>())
+          maxFloatAttr = rewriter.getF32FloatAttr(static_cast<float>(maxInt));
+        else if (elementType.isa<mlir::IntegerType>())
+          maxIntAttr = rewriter.getI64IntegerAttr(maxInt);
+      } else {
+        return rewriter.notifyMatchFailure(
+            op, "expected to find a float or integer constant as max");
+      }
+    }
```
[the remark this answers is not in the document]
navahgar: Done

lib/Conversion/TorchToTcp/Elementwise.cpp
navahgar: Fixed

lib/Conversion/TorchToTcp/Elementwise.cpp
navahgar: Fixed

lib/Conversion/TorchToTcp/Elementwise.cpp
navahgar: Thanks for those points. I replaced this with different classes for `relu` and `relu6`.

lib/Conversion/TorchToTcp/Elementwise.cpp
navahgar: Thats a good question. It looks like the `relu6` e2e test ends up doing a `relu` followed by a `min` (which is not supported in Tcp yet). Here is the module I see before `TorchToTcp` conversion:
```
  func.func @forward(%arg0: !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[?,?],f32> {
    %int6 = torch.constant.int 6
    %none = torch.constant.none
    %0 = torch.aten.relu %arg0 : !torch.vtensor<[?,?],f32> -> !torch.vtensor<[?,?],f32>
    %1 = torch.prim.ListConstruct  : () -> !torch.list<int>
    %2 = torch.prim.NumToTensor.Scalar %int6 : !torch.int -> !torch.vtensor<[],si64>
    %3 = torch.aten.to.dtype %2, %int6, %false, %false, %none : !torch.vtensor<[],si64>, !torch.int, !torch.bool, !torch.bool, !torch.none -> !torch.vtensor<[],f32>
    %4 = torch.aten.broadcast_to %3, %1 : !torch.vtensor<[],f32>, !torch.list<int> -> !torch.vtensor<[],f32>
    %5 = torch.aten.minimum %0, %4 : !torch.vtensor<[?,?],f32>, !torch.vtensor<[],f32> -> !torch.vtensor<[?,?],f32>
    return %5 : !torch.vtensor<[?,?],f32>
  }
```

Is this a known issue?

lib/Conversion/TorchToTcp/Utils.cpp
navahgar: Good point. Fixed.

navahgar (commented):
Thanks a lot for the review @jpienaar and @ramiro050. 

Addressed your comments. PTAL.

lib/Conversion/TorchToTcp/Elementwise.cpp
navahgar: Reg extracting the error checking code into a helper function, I'm not sure how that is going to help. I will still have to catch that error message and return it in this function conditionally. Is there a more compact way to do that with error messages in mlir?

lib/Conversion/TorchToTcp/Elementwise.cpp
sjain-stanford: `relu6` can be represented as `min(relu(x), 6)` so the module above is sort of expected. Perhaps we can hold off on `relu6` until we have `tcp.min`, or capture a TODO to re-enable the e2e test once `tcp.min` is added.

navahgar: @sjain-stanford While that representation is correct, the fact that it is represented that way when torch has a `torch.aten.relu6` op seems like a bug to me. 

Both those options seem reasonable to me.

sjain-stanford: Ah I see. In that case I agree that I'm not sure why it would lower to these primitives, or if it is somehow configurable to disable this.

ramiro050: Ah right, this is because we have a decomposition for `Relu6`: https://github.com/llvm/torch-mlir/blob/65846010cbfd121c8666ec3a44827f975600e081/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L745

You should be able to turn off the decomposition by adding your ops to your backend here:

https://github.com/llvm/torch-mlir/blob/65846010cbfd121c8666ec3a44827f975600e081/python/torch_mlir/__init__.py#L250

ramiro050: Or if it's not urgent to have support for `Relu6`, you can just remove the `Relu6` pattern and wait until support for `min` is added. Either approach is fine

navahgar: Ah okay. Good to know that is configurable. Thanks for the pointer.

For now, I've removed support for `Relu6`.

lib/Conversion/TorchToTcp/Elementwise.cpp
ramiro050: I was thinking of having something like

```c++
static LogicalResult isIntOrFloatRankedTensor(Operation *op, Value input, PatternRewriter &rewriter);

// In matchAndRewrite
if (failed(isIntOrFloatRankedTensor(op, input, rewriter))
```

However, this is not really necessary right now. This can be done later if this check becomes a very common one

ramiro050 (approved):
LGTM!

lib/Conversion/TorchToTcp/Elementwise.cpp
navahgar: Okay, sounds good. 

I wish there was something like `RETURN_IF_ERROR`.

jpienaar (approved):
Works for me thanks

PR #1839, opened 2023-01-31
[MHLO] Support AtenMaskedFillScalar

ZihengJiang:


lib/Conversion/TorchToMhlo/Basic.cpp
```diff
@@ -1292,6 +1292,51 @@ LogicalResult ConvertAtenOp<AtenArangeStartStepOp>::matchAndRewrite(
   return success();
 }
 
+// AtenMaskedFillScalarOp
+template <>
+LogicalResult ConvertAtenOp<AtenMaskedFillScalarOp>::matchAndRewrite(
```
tanyokwok, line 1297: I think we can decompose AtenMaskedFillScalarOp into `AtenWhereScalarSelfOp`, which is more straightforward

tanyokwok: And the `AtenWhereScalarSelfOp` can be convert into `mhlo::BroadcastSelectOp`

ZihengJiang:
The CI is failed due to TOSA failed the `MaskedFillScalarIntValueModule_basic` test, but this test is actually not added into the TOSA_PASS_SET in xfail_sets. Any idea? @ramiro050

ramiro050:
The test seems to be passing

```
Unexpected outcome summary:

****** Unexpectedly Passed tests - 1 tests
    XPASS - "MaskedFillScalarIntValueModule_basic"
```

Should just be a matter of adding it to the tosa_pass_set

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1390,6 +1390,25 @@ class DecomposeAtenWhereScalarSelfOp
 };
 } // namespace
 
+// Decompose aten.masked_fill.Scalar into aten.where.self op.
+namespace {
+class DecomposeAtenMaskedFillScalarOp
```
ramiro050, line 1395: I think we can remove the linalg lowering for this op now that we have the decomposition

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3821,6 +3840,7 @@ class DecomposeComplexOpsPass
     addPatternIfTargetOpIsIllegal<DecomposeAtenWhereScalarOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenWhereScalarOtherOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenWhereScalarSelfOp>(patterns);
+    addPatternIfTargetOpIsIllegal<DecomposeAtenMaskedFillScalarOp>(patterns);
```
ramiro050, line 3843: You also need to add the op here: https://github.com/llvm/torch-mlir/blob/eb7bac7ca43ef76581f025fb1c0417308a29582a/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L335

ramiro050 (changes requested):

ramiro050 (approved):
Thanks!

PR #1840, opened 2023-02-01
mhlo: migrate conversion to stablehlo

ashay:
This patch replaces all MHLO operations with their StableHLO
counterparts and adds a validation pass to ensure that no MHLO operations
remain before translating all Stablehlo operations to the MHLO dialect
for further lowering to the Linalg dialect.

This patch also updates all lit tests so that they refer to the
`convert-torch-to-stablehlo` pass and so that they check for StableHLO
operations.


burmako:
@ashay Oh that's a good catch! When we forked StableHLO from MHLO, we did a survey of frontends and looked into which MHLO ops are used and which ops aren't. Only the former ops made the cut, and `copy` was in the latter category. Sounds like we have missed the usage of `copy` in Torch-MLIR!

`mhlo.copy` is an interesting op. It was added a while ago - I imagine it was done for the reasons of parity with kCopy in HLO - but it didn't model the fullness of HLO's semantics, and became kind of a no-op (MHLO semantics involves immutable tensors, so a copy is just a no-op). Then, at a later point, this was recognized, and [a folder](http://google3/third_party/tensorflow/compiler/xla/mlir_hlo/mhlo/IR/hlo_ops.cc;l=3202;rcl=504910725) was added which unconditionally replaced `mhlo.copy` with its operand.

I see that at the moment `mhlo.copy` is only used in one place in Torch-MLIR. I wonder if it's possible to just use its operand instead of creating this op in the first place: https://github.com/llvm/torch-mlir/blob/5057e8871592d6c52c2e2c9e646c07360ee07517/lib/Conversion/TorchToMhlo/Basic.cpp#L1382.

lib/Conversion/TorchToMhlo/Basic.cpp
```diff
@@ -1376,23 +1391,24 @@ void mlir::torch::torch_to_mhlo::populateBasicOpPatternsAndLegality(
   target.addIllegalOp<RuntimeAssertOp>();
   patterns.add<ConvertRuntimeAssertOp>(typeConverter, context);
 
-#define INSERT_UNARY_PATTERN(AtenOp, MhloOp)                                   \
+#define INSERT_UNARY_PATTERN(AtenOp, StablehloOp)                              \
   target.addIllegalOp<AtenOp>();                                               \
-  patterns.add<ConvertAtenUnaryOp<AtenOp, MhloOp>>(typeConverter, context)
+  patterns.add<ConvertAtenUnaryOp<AtenOp, StablehloOp>>(typeConverter, context)
   INSERT_UNARY_PATTERN(AtenCloneOp, mhlo::CopyOp);
```
tanyokwok, line 1397: Nit: how about changing `mhlo::CopyOp` to `stablehlo::ConvertOp` or `tensor::CastOp`, since `AtenCloneOp` is a no-op exactly in HLO? So we can remove all dependencies on MHLO in the TorchToMhlo pass.

lib/Conversion/TorchToMhlo/TorchToMhlo.cpp
```diff
@@ -49,7 +51,8 @@ class ConvertTorchToMhlo : public ConvertTorchToMhloBase<ConvertTorchToMhlo> {
     MLIRContext *context = &getContext();
     ConversionTarget target(*context);
     target.addLegalDialect<chlo::ChloDialect, mhlo::MhloDialect,
-                           tensor::TensorDialect, arith::ArithDialect>();
+                           stablehlo::StablehloDialect, tensor::TensorDialect,
```
tanyokwok, line 54: Nit: The naming MHLO now could be renamed to StableHLO.

silvasean:
Nit: possibly in a separate PR, we can rename all the `Mhlo` in file paths / directories to `Stablehlo` (note `h` is lowercase officially in the dialect name).

Also when possible, updating the VerifyStablehloBackendContract pass to reject Mhlo would be good (in some sense that is the true milestone of completion of this workstream).

And if we can get sign-off from @burmako that it is indeed checking the correct set of ops: https://github.com/llvm/torch-mlir/pull/1840/files#diff-d8eb5bebe72fb7e5c0d49bd9c2234d6ac74e44f250d38e7f0d7af0fee52530d7 -- two questions:
- Is CHLO part of the StableHLO set?
- Is shape::ShapeOfOp part of StableHLO set?

lib/Conversion/TorchToMhlo/Basic.cpp
ashay: Thanks!  Replaced with `stablehlo::ConvertOp`.

lib/Conversion/TorchToMhlo/TorchToMhlo.cpp
ashay: Fixed.

ashay:
> Nit: possibly in a separate PR, we can rename all the `Mhlo` in file paths / directories to `Stablehlo` (note `h` is lowercase officially in the dialect name).

Fixed in this PR.

> Also when possible, updating the VerifyStablehloBackendContract pass to reject Mhlo would be good (in some sense that is the true milestone of completion of this workstream).

Since the TorchToMhlo pipeline ultimately lowers to Mhlo, I don't think we can get rid of the VerifyMhloBackendContract pass.  So I added the VerifyStablehloBackendContract pass after the torch-to-stablehlo conversion while retaining the existing VerifyMhloBackendContract pass.,

silvasean:
> Since the TorchToMhlo pipeline ultimately lowers to Mhlo, I don't think we can get rid of the VerifyMhloBackendContract pass. So I added the VerifyStablehloBackendContract pass after the torch-to-stablehlo conversion while retaining the existing VerifyMhloBackendContract pass.,

It should lower to StableHLO and not Mhlo. The later lowering to MHLO is really just used as part of e2e testing since only mhlo has a path to linalg.

Concretely we would put lowering stablehlo->mhlo here: https://github.com/llvm/torch-mlir/blob/95361747c2c350886d192261667bfc57d870971d/python/torch_mlir_e2e_test/mhlo_backends/linalg_on_tensors.py#L39

(and that whole directory structure also needs to be renamed to stablehlo, as well as the e2e test config)

burmako:
As far as VerifyStablehloBackendContract goes, how about we populate it with the contents of the current VerifyMhloBackendContract with the only change being `mhlo::MhloDialect` replaced with `stablehlo::StablehloDialect`, and then over time work on shrinking this contract?

Here's the rationale for this proposal: StableHLO is ready to serve as a drop-in replacement for MHLO in the role of an interface between ML frameworks and ML compilers. It supports everything that MHLO supports at the interface level, and then some, but some future work still remains, namely:
  1. We are planning to replace `func.func` and `func.return` in StableHLO programs with StableHLO-specific ops (https://github.com/openxla/stablehlo/issues/425).
  2. There is no standard way of writing dynamically-shaped programs at the moment. The spec doesn't take a position on this at the moment of writing, because it's ongoing work (https://github.com/openxla/stablehlo/issues/8).
  3. CHLO is not part of the spec at the moment of writing, and I'm starting to think that maybe it shouldn't be (https://github.com/openxla/stablehlo/issues/602).

Given that, the existing pieces of VerifyMhloBackendContract make sense:
  * `target.addDynamicallyLegalOp<ModuleOp, func::FuncOp, func::ReturnOp>(opHasLegalTypes)` corresponds to 1).
  * `target.addDynamicallyLegalOp<shape::ShapeOfOp>(opHasLegalTypes)`, `target.addLegalDialect<tensor::TensorDialect>()`, `target.addLegalDialect<arith::ArithDialect>()` correspond to 2).
  * `target.addLegalDialect<chlo::ChloDialect>()` correspond to 3).

As we're working through these open questions, I expect that we'll be able to eventually shrink the backend contract. For example, one potential future could involve: a) keeping `builtin.module`, b) introducing `stablehlo.func` and using `stablehlo.return` for all returns, c) vendoring all shape computations into the StableHLO dialect, obviating the need for ops from arith, shape and tensor dialects in StableHLO programs, d) deconstructing CHLO to StableHLO. In that future, the contract would shrink to allowing just the ModuleOp and the StableHLO dialect.

What do you think?

silvasean:
Thanks Eugene, this makes sense to me!

silvasean (approved):
LGTM, but let's wait for one of the owners of StableHLO support in Torch-MLIR approve as well.

tanyokwok (approved):
LGTM!

PR #1845, opened 2023-02-02
[TOSA] Add todtype f64 support

AmosLewis:
Find this issue when lower opt model: https://github.com/nod-ai/SHARK/issues/865

This will need the llvm patch tosa cast op f64 support merged https://reviews.llvm.org/D142599


vivekkhandelwal1:
> For this patch work, it will need the llvm patch tosa cast op f64 support merged https://reviews.llvm.org/D142599

In that case, we should wait for that patch to get merged in LLVM and then include it in Torch-MLIR through bump before merging this patch.

PR #1848, opened 2023-02-04
Python 3.11 support

powderluv:


ashay (approved):
LGTM!  Just out of curiosity, won't we hit against the GitHub Actions Limit now that we've added another Python version for our macOS builds?

ashay:
Sorry, checked again and noticed that macOS builds refer to 3.11 only now.

powderluv:
We added it to the default list but we override it from the GHA to only build 3.11. Here is a sample run https://github.com/llvm/torch-mlir/releases/tag/oneshot-20230210.82 with 3.11.

PR #1849, opened 2023-02-05
[Torch] decompose aten.norm.ScalarOpt_dim to aten.linalg_vector_norm

qingyunqu:


python/torch_mlir_e2e_test/test_suite/__init__.py
```diff
@@ -8,7 +8,6 @@
 # to the backend contract.
 COMMON_TORCH_MLIR_LOWERING_XFAILS = {
     "QuantizedMLP_basic",
-    "NormalizeModule_basic",
```
ramiro050, line 11: Can you also add e2e tests testing the other arguments that `norm.ScalarOpt_dim` takes?

ramiro050 (changes requested):
Thanks! A few initial comments

qingyunqu:
I have add the E2E test. But there is something wrong.
In my local test, I must use `torch.jit.trace` to trace the test Module:

```
class ScalarOptDimNormModule(torch.nn.Module):
    def __init__(self) -> None:
        super().__init__()
    
    def forward(self, a):
        return torch.ops.aten.norm.ScalarOpt_dim(a, 2.0, dim=[0, 1], keepdim=False)

model = ScalarOptDimNormModule().eval()
inputs = torch.randn(3, 4, 5)
model = torch.jit.trace(model, inputs)   // trace the Module
module = torch_mlir.compile(model, inputs, output_type="torch")
```

Otherwise it will failed as:
image

qingyunqu:
@ramiro050 Do you have any advice?

ramiro050:
> @ramiro050 Do you have any advice?

In the e2e tests, the ops must be specified without overloads, otherwise torchscript fails to generate IR as you've shown. The write overload is chosen based on the inputs to the op. If you do

```python
torch.ops.aten.norm(grad_out, 2.0, dim=[0, 1], keepdim=False)
```

This generates the MLIR

```mlir
%1 = torch.prim.ListConstruct %int0, %int1 : (!torch.int, !torch.int) -> !torch.list<int>
%2 = torch.aten.norm.ScalarOpt_dim %arg1, %float2.000000e00, %1, %false : !torch.tensor, !torch.float, !torch.list<int>, !torch.bool -> !torch.tensor
```

qingyunqu:
@ramiro050 Please review this PR again, thanks!

python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -678,6 +678,46 @@ def forward(self, a):
...
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+    ])  
+    def forward(self, a):
+        return torch.ops.aten.norm(a, self.p, dim=[0, 1], keepdim=True)
+
+@register_test_case(module_factory=lambda: NormScalarOptDimKeepDimModule())
+def NormScalarOptDimKeepDimModule_basic(module, tu: TestUtils):
+    module.forward(torch.rand(3, 4, 5))
```
ramiro050, line 718: nit: use `tu.rand` to generate the random tensors. (I know there are some tests in this file that do not follow this convention. I'll be fixing this soon). See:

https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#test-case-function

ramiro050 (approved):
Just a small change request, other than that LGTM

python/torch_mlir_e2e_test/test_suite/reduction.py
qingyunqu: done

PR #1854, opened 2023-02-06
BUILD: No PyTorch dep

makslevental:
```shell
$ rm -rf venv
$ python -m venv venv
$ source venv/bin/activate
(venv) $ python test/python/smoketest.py
  File "test/python/smoketest.py", line 3, in <module>
    import torch_mlir.ir
ModuleNotFoundError: No module named 'torch_mlir'
$ pip install -r build-requirements.txt
$ TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS=1 CMAKE_GENERATOR=Ninja python setup.py bdist_wheel -d wheelhouse
<STUFF HAPPENS>
(venv) $ pip install wheelhouse/torch-mlir.whl
<STUFF HAPPENS>
(venv) $ python test/python/smoketest.py
module {
  %none = torch.constant.none
}
(venv) $ pip freeze | grep torch
torch-mlir==0.0.1
```

Note the absense of `torch==2.0.0.dev20230205`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,12 +41,14 @@ from setuptools import setup, Extension
 from setuptools.command.build_ext import build_ext
 from setuptools.command.build_py import build_py
 
-import torch
 
 PACKAGE_VERSION = os.environ.get("TORCH_MLIR_PYTHON_PACKAGE_VERSION") or "0.0.1"
 
 # If true, enable LTC build by default
 TORCH_MLIR_ENABLE_LTC_DEFAULT = True
+TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS = int(os.environ.get('TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS', False))
+if not TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS:
+    import torch
 
 # Build phase discovery is unreliable. Just tell it what phases to run.
 class CustomBuild(_build):
@@ -90,6 +92,7 @@ class CMakeBuild(build_py):
                 f"-DCMAKE_C_VISIBILITY_PRESET=hidden",
                 f"-DCMAKE_CXX_VISIBILITY_PRESET=hidden",
                 f"-DTORCH_MLIR_ENABLE_LTC={'ON' if enable_ltc else 'OFF'}",
+                f"-DTORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS={'ON' if TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS else 'OFF'}",
             ]
 
             os.makedirs(cmake_build_dir, exist_ok=True)
@@ -158,10 +161,8 @@ setup(
     },
     ext_modules=[
         CMakeExtension("torch_mlir._mlir_libs._jit_ir_importer"),
-    ],
-    install_requires=[
-        "numpy",
-        f"torch=={torch.__version__}".split("+", 1)[0],
-    ],
+    ] if not TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS else [CMakeExtension("_")],
+    install_requires=["numpy", ] + (
+        [f"torch=={torch.__version__}".split("+", 1)[0], ] if not TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS else []),
     zip_safe=False,
 )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,7 +41,13 @@ if(TORCH_MLIR_ENABLE_STABLEHLO)
   add_definitions(-DTORCH_MLIR_ENABLE_STABLEHLO)
 endif()
 
+option(TORCH_MLIR_ENABLE_JIT_IR_IMPORTER "Enables JIT IR Importer" ON)
 option(TORCH_MLIR_ENABLE_LTC "Enables LTC backend" OFF)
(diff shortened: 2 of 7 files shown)


python/torch_mlir/__init__.py
```diff
@@ -2,19 +2,26 @@
...
+try:
+    from torch._functorch.compile_utils import strip_overloads
+    import torch
+except ImportError:
+    warnings.warn("PyTorch not installed")
 
 from torch_mlir.passmanager import PassManager
 from .compiler_utils import run_pipeline_with_repro_report
-from torch_mlir.dialects.torch.importer.jit_ir import ClassAnnotator, ImportOptions, ModuleBuilder
+from torch_mlir.dialects import torch as torch_dialect
```
makslevental, line 20: This artificial split is necessary so that import machinery loads the modules in the `dialects` directory (since it's not a module itself - no `__init__.py`) even if JIT IR importer isn't installed.

powderluv (approved):

powderluv:
We should also try a "Oneshot nightly build"

silvasean (changes requested):
Hey folks, we really need to think this through a bit more. The current torch_mlir python module is very fundamentally tied to PyTorch. I think if we want to do this properly, we should basically just expose the vanilla `torch_mlir` python module which is basically just a wrapper around our MLIR infra. This would mostly be cmake work, such as bypassing things like https://github.com/llvm/torch-mlir/blob/cc819e73dd5a48162dcfc099bbdf1b54ca6831da/python/CMakeLists.txt#L48

We might want to go even further -- having two python modules -- one which is just the vanilla MLIR infra, and then a wrapper around those which provides the integration into PyTorch.

For context, the `torch_mlir` Python module is currently the MLIR wrapper one (things like torch_mlir.ir.Module, torch_mlir.passmanager, etc.) and then we decorate some stuff inside it (all of which depends on PyTorch). We can either use cmake to eliminate all the additional stuff we put inside it, or pivot to having two separate Python modules. I think using cmake here is probably the right choice. The flag would be called `TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS` which implies `TORCH_MLIR_ENABLE_JIT_IR_IMPORTER=NO` and `DTORCH_MLIR_ENABLE_LTC=NO` and some other stuff.

powderluv:
I like having two. python modules / whl packages. This way we can actually pass auditwheel and publish to pypi

makslevental:
> I like having two. python modules / whl packages. This way we can actually pass auditwheel and publish to pypi

I'm not 100% sure what it means to "pass auditwheel" but after building the wheel with `TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS=1` I can do this:

```shell
(venv) $ auditwheel repair --plat manylinux_2_35_x86_64 torch_mlir-0.0.1-cp310-cp310-linux_x86_64.whl 
INFO:auditwheel.main_repair:Repairing torch_mlir-0.0.1-cp310-cp310-linux_x86_64.whl
INFO:auditwheel.wheeltools:Previous filename tags: linux_x86_64
INFO:auditwheel.wheeltools:New filename tags: manylinux_2_35_x86_64
INFO:auditwheel.wheeltools:Previous WHEEL info tags: cp310-cp310-linux_x86_64
INFO:auditwheel.wheeltools:New WHEEL info tags: cp310-cp310-manylinux_2_35_x86_64
INFO:auditwheel.main_repair:
Fixed-up wheel written to /home/mlevental/dev_projects/torch-mlir/wheelhouse/wheelhouse/torch_mlir-0.0.1-cp310-cp310-manylinux_2_35_x86_64.whl
```

setup.py
```diff
@@ -158,10 +161,8 @@ def build_extension(self, ext):
     },
     ext_modules=[
         CMakeExtension("torch_mlir._mlir_libs._jit_ir_importer"),
-    ],
-    install_requires=[
-        "numpy",
-        f"torch=={torch.__version__}".split("+", 1)[0],
-    ],
+    ] if not TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS else [CMakeExtension("_")],
```
makslevental, line 164: you need something here for the built libs to actually be copied into the wheel; leaving `CMakeExtension("torch_mlir._mlir_libs._jit_ir_importer")` as is produces a spurious lib (basically nothing in it):

```shell
(venv) $ nm -gDC torch-mlir/_mlir_libs/_jit_ir_importer.cpython-311-x86_64-linux-gnu.so 
                 w __cxa_finalize
                 w __gmon_start__
                 w _ITM_deregisterTMCloneTable
                 w _ITM_registerTMCloneTable
```

I believe the alternative (correct-ish way) is to use a [MANIFEST.in](https://packaging.python.org/en/latest/guides/using-manifest-in/) but I haven't had much luck in the past getting that to work smoothly with MLIR Python bindings.

makslevental:
I believe this looks reasonable (but I've been wrong before 😅). Adjusting the actual build scripts should be a separate PR as they are ...dense...

silvasean:
Do we want to frontload splitting the modules? Or do that later? This seems reasonable to me if it works. @powderluv ?

makslevental:
> Do we want to frontload splitting the modules? Or do that later? This seems reasonable to me if it works. @powderluv ?

I'm always happy to kick the can down the road until there's a particular need (especially since this works perfectly as is) but if you want the "frontloading" I'm fine with that too.

makslevental:
> Do we want to frontload splitting the modules? Or do that later? This seems reasonable to me if it works. @powderluv ?

@silvasean 

I believe this meets all of our current needs completely so I'm in support of deferring the split until there's a particular need.

powderluv:
Agreed. This is good to land and iterate.

powderluv (approved):
Please change commit message to remove WIP

powderluv:
Also FYI for @ashay since he is cleaning up the whl and pytorch requirements etc.

PR #1862, opened 2023-02-08
[Torch Dialect] Add support for aten::glu

Groverkss:
This patch adds support for aten::glu

The implementation is based on: https://github.com/pytorch/pytorch/blob/73bf32cb5709b85ab7df4f986d259612073eb0fc/aten/src/ATen/native/GatedLinearUnit.cpp#L23


Groverkss:
I was not sure where to put this operation in test_suite, so I put it in basic.py for now. Suggestions on where I should put this instead would be helpful.

I did not add this operation to the abstract interp lib because I was not sure if that is required for decompositions. If it is needed, please let me know, I will add that too.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -3249,3 +3251,23 @@ def forward(self, input, boundaries):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True)
+    ])
+    def forward(self, x):
+        return torch.ops.aten.glu(x, dim=-1)
```
makslevental, line 3269: as discussed, can you please add a test that exercises this lowering through the public [torch api](https://pytorch.org/docs/stable/generated/torch.nn.GLU.html)

makslevental (changes requested):
LGTM (modulo that one test) but should have another set of eyes @vivekkhandelwal1 can you PTAL

makslevental:
@Groverkss @vivekkhandelwal1 also I believe that indeed it is the case that you need to update the shape library for decomps? https://github.com/llvm/torch-mlir/pull/1804/files

vivekkhandelwal1:
@Groverkss Please add the shape and dtype info for the op as specified here: https://github.com/llvm/torch-mlir/blob/main/docs/abstract_interp_lib.md

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3924,6 +3976,7 @@ class DecomposeComplexOpsPass
     addPatternIfTargetOpIsIllegal<DecomposeAtenLeakyReluBackwardOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenNewEmptyStridedOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenBucketizeTensorOp>(patterns);
+    addPatternIfTargetOpIsIllegal<DecomposeAtenGluOp>(patterns);
```
vivekkhandelwal1, line 3979: The op decomposed also needs to be added here: https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L443 like other ops.

lib/Conversion/TorchToArith/TorchToArith.cpp
```diff
@@ -140,6 +140,25 @@ class ConvertAtenDivIntOp : public OpConversionPattern<AtenDivIntOp> {
 };
 } // namespace
 
+namespace {
+class ConvertAtenRemainderIntOp : public OpConversionPattern<AtenRemainderIntOp> {
```
ramiro050, line 144: Drive by comment: this op also needs an e2e test. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

python/torch_mlir_e2e_test/test_suite/scalar.py
```diff
@@ -428,3 +428,27 @@ def forward(self, val):
...
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([], torch.int64, True),
+        ([], torch.int64, True),
+    ])
+    def forward(self, lhs, rhs):
+        # Cast the result to float to make e2e test baseline result to be a float.
```
ramiro050, line 448: nit: replace `float` with `int`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3769,6 +3769,58 @@ class DecomposeAtenNewEmptyStridedOp
...
+    Value a = rewriter.create<AtenNarrowOp>(loc, op.getResult().getType(), x,
+                                            dim, zeroVal, selfSize);
+    // b = x.narrow(dim, selfSize, selfSize)
+    Value b = rewriter.create<AtenNarrowOp>(loc, op.getResult().getType(), x,
+                                            dim, selfSize, selfSize);
+    // sigmoid = sigmoid(b).
+    Value sigmoid = rewriter.create<AtenSigmoidOp>(loc, b.getType(), b);
+    // return a * sigmoid.
+    Value aMulSigmoid =
+        rewriter.create<AtenMulTensorOp>(loc, sigmoid.getType(), a, sigmoid);
```
ramiro050, line 3816: nit: I believe all these ops have the same result type, so I would create a variable `Value resultType = op.getResult().getType()`, and use that for all of them to make things clear

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -579,6 +579,10 @@ def aten〇new_empty〡shape(self: List[int], size: List[int], dtype: Optional[i
 def aten〇new_empty_strided〡shape(self: List[int], size: List[int], stride: List[int], dtype: Optional[int] = None, layout: Optional[int] = None, device: Optional[device] = None, pin_memory: Optional[bool] = None) -> List[int]:
     return size
 
+def aten〇glu〡shape(self: List[int], dim: int = -1) -> List[int]:
+    self[dim] = self[dim] // 2
```
ramiro050, line 583: This needs to be upstreamed first. See: https://github.com/llvm/torch-mlir/blob/main/docs/adding_abstract_interpretation_functions.md#step-by-step-guide

> New shape functions should be added upstream following the example of https://github.com/pytorch/pytorch/pull/76889, though it can be useful to iterate locally in abstract_interp_lib_gen.py first.

ramiro050 (changes requested):

PR #1691, opened 2022-12-07
Add aten.slice.Tensor & aten.cat folders

asaadaldien:
- Add folders for single input concatenation and full range slices.

PR #1257, opened 2022-08-21
Add folder for ToF64Op and FromF64Op

Vremold:
This PR adds folder for `ToF64Op` and `FromF64Op` in `TorchConversion` dialect, and helps to canonicalize IR pattern like this:
```mlir
%float1 = torch.constant.float 1.000000e+00
%0 = torch_c.to_f64 %float1
```
or:
```mlir
%c5_f64 = arith.constant 5.000000e+00 : f64
%0 = torch_c.from_f64 %c5_f64
```

PR #1863, opened 2023-02-08
[Torch Dialect] add folder for aten.Int.float

qingyunqu:


test/Dialect/Torch/canonicalize.mlir
```diff
@@ -1253,6 +1253,15 @@ func.func @torch.aten.Int.Tensor(%arg0: !torch.int) -> !torch.int {
   return %scalar : !torch.int
 }
 
+// CHECK-LABEL:   func.func @torch.aten.Int.float() -> !torch.int {
+// CHECK:             %[[NUM:.*]] = torch.constant.int 1
+// CHECK:             return %[[NUM]] : !torch.int
+func.func @torch.aten.Int.float() -> !torch.int {
```
ramiro050, line 1259: Can you also add an e2e test?

ramiro050 (changes requested):
Thanks!

test/Dialect/Torch/canonicalize.mlir
ZihengJiang: Hi Ramiro, it seems hard to add a MHLO conversion for a scalar integer to enable a e2e test. Also, the previous folder PRs all only do canonicalize test: https://github.com/llvm/torch-mlir/pull/1691, https://github.com/llvm/torch-mlir/pull/1257

qingyunqu: I add the e2e test. The torch.constant.int‘s value attribute is Singed Integer, while the  arith.constant only accept Signless Integer. So I add the type conversion when lowering torch.constant.int to arith.constant.

qingyunqu: The CI failed in strange way. Could you rerun it? Thanks.

ramiro050 (approved):
Thanks for adding the right conversion to arith!

Issue #1462, opened 2022-10-05
[RFC] Custom ops support


silvasean:
Hey folks, custom ops has been a common feature request, and I'd like to share a design that we've been thinking about.

Let's talk first about the functionality we want to unlock for users. Generally speaking, custom ops are a way to "tunnel" information from the frontend to the backend. This can be something like a custom fusion, or something more advanced, like a representation of quantization or sparsity. Note that ops can also carry various different types of information, such as `!torch.int`, `!torch.string`, etc. (which we would generally need to be constrained to statically known values to be useful in the compiler). So an op like `my.custom_quantize %foo, %quantization_info, ...` etc. could actually carry a fairly elaborate representation of quantization/sparsity/etc. in the `%quantization_info` string.

The key requirement here is the desire to support new ops without recompiling Torch-MLIR. We've had some [initial efforts](https://github.com/llvm/torch-mlir/pull/1355) in the custom op direction, but they always required recompiling Torch-MLIR. Put another way, the goal is to generalize the current way of handling ops so that there is no "hardcoded C++" for each op. Here by **"support new ops" we really mean lowering to the backend contract** (see [architecture.md](https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md)). Lowering to Linalg/etc. generally is out of scope of this effort, since a different set of technology is involved.

In our longer-term roadmap (see [slide](https://docs.google.com/presentation/d/1YXRi0B9sGYjs7m4SS4TZAZeZiN4kOOUPttC_5rFQzr4/edit?resourcekey=0-RXKEOmWsvLoEiS8uMwLFJw#slide=id.gfee1e8ed26_0_14)) of aligning with TorchDynamo and upstream PyTorch, we will directly import into the backend contract, so all of this work on custom ops will be obsoleted. But the indication is that this longer-term roadmap will take 1-2 years to fully pan out, while we want to unlock this user requirement in a shorter timeframe (ideally months).

------------------------------------

When lowering to the backend contract, there are 4 main responsibilities (see [architecture.md](https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#the-backend-contract)):

- All tensors have been converted to value semantics.
- All tensors have at least a known number of dimensions (i.e. rank), and ideally also have a precise size for each dimension.
- All tensors have a known dtype.
- Certain ops have been decomposed to make them easier to handle (this is configurable).

Let's go over these one by one:

[...]

PR #1865, opened 2023-02-09
Change dtype functions interface to take ints tuple for each tensor

ramiro050:
The original design for the dtype functions outlined in https://github.com/llvm/torch-mlir/issues/1462 was unable to properly handle ops that take optional tensors as an input when the optional tensor has a value of None. By the time the op gets imported into torch-mlir, if an optional value is None, all information about the original type is lost from the op type signature, preventing torch-mlir from knowing if a value of None was from an optional tensor or not, which was crucial in the original design since each tensor argument must be turned into two separate arguments for the dtype function.

This commit changes the interface to dtype functions such that each tensor turns into a tuple of two ints, the first representing the rank of the tensor and the second the dtype of the tensor. Since now there is a one-to-one correspondence between the operands of an op and the operands of its dtype function, there is no ambiguity about which operand of the op corresponds with which operand of the dtype function.

To test the implementation, this commit defines dtype functions for the convolution ops, all of which take one optional tensor as an argument.


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1216,33 +1233,38 @@ def aten〇frobenius_norm〇dim〡dtype(self_rank: int, self_dtype: int, dim: Li
...
 @check_dtype_function(_check_tensors_with_the_same_dtype(num_of_tensors=1, other=0.0) +
                       _check_tensors_with_the_same_dtype(num_of_tensors=1, other=0))
-def aten〇eq〇Scalar〡dtype(self_rank: int, self_dtype: int, other: Union[int, float]) -> int:
+def aten〇eq〇Scalar〡dtype(self_rank_dtype: Tuple[int, int], other: Union[int, float]) -> int:
     return torch.bool
 
 @check_dtype_function(_check_two_tensor_op())
-def aten〇eq〇Tensor〡dtype(self_rank: int, self_dtype: int, other_rank: int, other_dtype: int) -> int:
+def aten〇eq〇Tensor〡dtype(self_rank_dtype: Tuple[int, int], other_rank_dtype: Tuple[int, int]) -> int:
+    other_rank, other_dtype = other_rank_dtype
```
li-plus, line 1258: This line can be removed.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1267,46 +1292,53 @@ def aten〇gt〇Tensor〡dtype(self_rank: int, self_dtype: int, other_rank: int,
...
-def aten〇le〇Scalar〡dtype(self_rank: int, self_dtype: int, other: Union[int, float]) -> int:
+def aten〇le〇Scalar〡dtype(self_rank_dtype: Tuple[int, int], other: Union[int, float]) -> int:
+    self_rank, self_dtype = self_rank_dtype
     assert not is_complex_dtype(self_dtype), "`self` cannot be complex"
     return torch.bool
 
 @check_dtype_function(_check_two_tensor_op())
-def aten〇logical_and〡dtype(self_rank: int, self_dtype: int, other_rank: int, other_dtype: int) -> int:
+def aten〇logical_and〡dtype(self_rank_dtype: Tuple[int, int], other_rank_dtype: Tuple[int, int]) -> int:
+    other_rank, other_dtype = other_rank_dtype
```
li-plus, line 1302: This line can be removed.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1267,46 +1292,53 @@ def aten〇gt〇Tensor〡dtype(self_rank: int, self_dtype: int, other_rank: int,
...
 
 @check_dtype_function(_check_tensors_with_the_same_dtype(num_of_tensors=1))
-def aten〇logical_not〡dtype(self_rank: int, self_dtype: int) -> int:
+def aten〇logical_not〡dtype(self_rank_dtype: Tuple[int, int]) -> int:
     return torch.bool
 
 @check_dtype_function(_check_two_tensor_op())
-def aten〇logical_or〡dtype(self_rank: int, self_dtype: int, other_rank: int, other_dtype: int) -> int:
+def aten〇logical_or〡dtype(self_rank_dtype: Tuple[int, int], other_rank_dtype: Tuple[int, int]) -> int:
+    other_rank, other_dtype = other_rank_dtype
```
li-plus, line 1311: This line can be removed.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1267,46 +1292,53 @@ def aten〇gt〇Tensor〡dtype(self_rank: int, self_dtype: int, other_rank: int,
...
     return torch.bool
 
 @check_dtype_function(_check_two_tensor_op())
-def aten〇logical_xor〡dtype(self_rank: int, self_dtype: int, other_rank: int, other_dtype: int) -> int:
+def aten〇logical_xor〡dtype(self_rank_dtype: Tuple[int, int], other_rank_dtype: Tuple[int, int]) -> int:
+    other_rank, other_dtype = other_rank_dtype
```
li-plus, line 1316: And this

li-plus (changes requested):

li-plus:
LGTM!

PR #1866, opened 2023-02-09
[TCP] tcp.sigmoid, tcp.divf ops and lowerings

sjain-stanford:
### Description

Adds definitions for `tcp.sigmoid` and `tcp.divf` elementwise ops, and lowerings from `TorchToTcp` and `TcpToLinalg` along with `Dialect` and `Conversion` lit tests. Also includes the e2e integration tests enabled in the TCP XPASS list. 

Only supports floating-point division. Integer division support is future work (TBD).

The PR is broken down by commits per op in case it is easier to review per op.

### Testing
```
TM_PACKAGES="in-tree" ./build_tools/python_deploy/build_linux_packages.sh
```


externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -121,6 +139,25 @@ def Tcp_MulOp : Tcp_BinaryElementwiseOp<"mul", [SameOperandsAndResultElementType
   let assemblyFormat = "$in1 `,` $in2 attr-dict `:` type($in1) `,` type($in2) `->` type($out)";
 }
 
+def Tcp_DivOp : Tcp_BinaryElementwiseOp<"div", [SameOperandsAndResultElementType]> {
```
sanjoy, line 142: What is the behavior on integer division by zero?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -121,6 +139,25 @@ def Tcp_MulOp : Tcp_BinaryElementwiseOp<"mul", [SameOperandsAndResultElementType
...
+  let summary = "Computes elementwise division";
```
sanjoy, line 143: You probably need two division operations: signed & unsigned.

sanjoy (changes requested):

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Should we just make it TBD, just like `arith` (https://mlir.llvm.org/docs/Dialects/ArithOps/#arithdivsi-mlirarithdivsiop) since we are anyways lowering it down to the corresponding arith op.

sanjoy: Or just say it is undefined behavior?

sjain-stanford: Thanks. Will update the description to denote division by zero results in UB.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sjain-stanford: That's a good point. Do we have a preference between separate ops for float, signed int, unsigned int, versus using a `bool` attribute indicating signed/unsigned if integer type?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sanjoy: For signed division dividing `INT_MIN` by `-1` should also invoke UB.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -121,6 +139,25 @@ def Tcp_MulOp : Tcp_BinaryElementwiseOp<"mul", [SameOperandsAndResultElementType
...
+
+  let description = [{
+    Computes the elementwise division of `in1` and `in2`.
+  }];
+
+  let arguments = (ins
+    Tcp_Tensor:$in1,
```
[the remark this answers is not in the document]
sjain-stanford: We do support both integer and float division (using the same `tcp.div` op at the moment), and lower to `arith::DivFOp` if float and `arith::DivSIOp` if integer, however as @sanjoy pointed out above, I will update the integer lowering to differentiate the signed and unsigned usecases.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sanjoy: My preference is to have three variants: `fdiv`, `sidiv` and `uidiv`.  Unclear how quantization fits in here though.

sjain-stanford: With real (not-fake) quantization, tensors from a quantized op will be no different than regular integers, so IIUC this construct would still hold. LMK if I'm missing something (cc: @sjarus ). 

Also, hyper minor point on the naming for multiple ops - we may want to use `divf`, `divsi`, and `divui` to keep the same prefix across all div (like in `arith` dialect) (cc: @navahgar ).

navahgar: I was thinking of a `bool` attribute, only because it will keep the opset smaller. But thats not a strong preference though.

If we are going with multiple ops, `divf`, `divsi`, and `divui` sounds good to me.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -64,6 +64,24 @@ def Tcp_ClampOp : Tcp_UnaryElementwiseOp<"clamp"> {
...
 
+def Tcp_SigmoidOp : Tcp_UnaryElementwiseOp<"sigmoid"> {
+  let summary = "Computes sigmoid of input, elementwise";
+
+  let description = [{
+    Computes elementwise sigmoid of the input tensor.
+  }];
+
+  let arguments = (ins
+    Tcp_FloatOrComplexTensor:$in
```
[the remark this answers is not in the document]
navahgar: The datatype restriction is baked into the op in `tcp.div`'s case as well, only that the `Tcp_Tensor` type includes all data types supported in Tcp, whereas here we are restricting it.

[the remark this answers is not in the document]
sjain-stanford: Quantized integer type support for sigmoid is interesting, although I'd like to defer it until we have an actual use-case. LMK if you feel strongly about it.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sjain-stanford: Noticed that using separate ops allows easier definition of type constraints for inputs and outputs (e.g. `Tcp_FloatTensor` or `Tcp_IntTensor`), whereas in the combined op we'd use `Tcp_FloatOrIntTensor` type, requiring a separate verifier to do the type legalization.

navahgar: Not sure I see why we need separate verifier even with `Tcp_FloatOrIntTensor` type. 

Are you suggesting that to verify that both inputs have the same element type? We could use `SameOperandsElementType` trait for that.

sjain-stanford: Error when a `tcp.divf` op receives an `i32` operand (without a separate verifier):
```
unexpected error: 'tcp.divf' op operand #0 must be ranked tensor of floating-point values, but got 'tensor<?x?xi32>'
```

sjain-stanford: Moreover, the integer divide requires additional inputs (e.g. rounding modes), and targets a separate `torch.aten` op to lower from (`Torch_AtenDivTensorMode`). So I'm leaning towards separate ops for cleaner separation of concerns. Will also defer the integer divide case for later, and revise this PR to include `tcp.divf` alone.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sjain-stanford: Only supports floating-point division. Integer division support is future work (TBD).

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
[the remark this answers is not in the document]
sjain-stanford: Only supports floating-point division. Integer division support is future work (TBD).

PR #1867, opened 2023-02-09
Add compile_commands.json to gitignore

Groverkss:


Groverkss:
For neovim, generally clangd looks at the root of the directory for the compilation database.

Please let me know if there is an alternative workflow preferred!

ramiro050:
I'm personally fine with these changes, but I'm not sure if there are any rules regarding adding things to the .gitignore. @silvasean PTAL

powderluv (approved):

powderluv:
should be ok to add to .gitignore.

silvasean:
seems fine

PR #1871, opened 2023-02-12
[Torch Dialect] add folder for aten.sub.float

qingyunqu:
The e2e test failed. The error is similar with #1849.  
image

### DIFF
--- a/python/torch_mlir_e2e_test/test_suite/basic.py
+++ b/python/torch_mlir_e2e_test/test_suite/basic.py
@@ -2707,6 +2707,27 @@ def IntFloatModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
+class AtenSubFloatModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+        self.value1 = 1.0
+        self.value2 = 2.0
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return float(torch.ops.aten.sub(self.value1, self.value2))
+
+@register_test_case(module_factory=lambda: AtenSubFloatModule())
+def AtenSubFloatModule_basic(module, tu: TestUtils):
+    module.forward()
+
+
+# ==============================================================================
+
 class ScalarImplicitFloatModule(torch.nn.Module):
 
     def __init__(self):
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -44,10 +44,6 @@ TORCHDYNAMO_XFAIL_SET = {
     # RuntimeError: Failed running call_function aten.uniform(...
     # https://github.com/pytorch/torchdynamo/issues/1954
     "UniformNoCorrelationModule_basic",
-    # TypeError: expected np.ndarray (got float)
-    # TODO: This is due to returning a scalar float as output from the test.
-    # We should probably just standardize all tests to return tensors.
-    "DivIntModule_basic",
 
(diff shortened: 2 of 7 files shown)


python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -2684,6 +2684,27 @@ def IntFloatModule_basic(module, tu: TestUtils):
...
+        super().__init__()
+        self.value1 = 1.0
+        self.value2 = 2.0
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.ops.aten.sub.float(self.value1, self.value2)
```
ramiro050, line 2701: You need to remove the overload, i.e. do `torch.ops.aten.sub(...)`. See: https://github.com/llvm/torch-mlir/pull/1849#issuecomment-1428564242

ramiro050 (approved):
LGTM!

python/torch_mlir_e2e_test/test_suite/basic.py
qingyunqu: But I found that when I set `raw` output type, it is already folded：（maybe the folding happened in python level？）
image

ramiro050: If you use `self.value` instead of passing constants directly, does it work?

qingyunqu: Yes, you are right, thanks. 
Another problem, I found the E2E test failed locally:
image

ramiro050: You might need to add a cast `float(...)` to the result of `torch.ops.aten.sub` before returning. Some ops that return scalars will return a tensor when run on regular Python, so this extra cast is needed to make types match.

```python
In [2]: torch.ops.aten.sub(1.0, 1.0)
Out[2]: tensor(0.)
```

See for example:

https://github.com/llvm/torch-mlir/blob/3bed35c0ee5ed3cef4ccc00264051dea5b2983a3/python/torch_mlir_e2e_test/test_suite/scalar.py#L168-L173

qingyunqu: Thanks. BTW, there is something wrong in CI?

ramiro050: Yeah, scalars were not properly handled by the torchdynamo backend. The following should fix it:

```diff
--- a/python/torch_mlir_e2e_test/configs/torchdynamo.py
+++ b/python/torch_mlir_e2e_test/configs/torchdynamo.py
@@ -59,10 +59,14 @@ def _refbackend_torchdynamo_backend(fx_graph: torch.fx.GraphModule,
     def compiled_callable(*inputs):
         inputs = [x.numpy() for x in inputs]
         result = loaded.forward(*inputs)
-        if not isinstance(result, tuple):
+        if isinstance(result, tuple):
+            result = tuple(torch.from_numpy(x) for x in result)
+        elif isinstance(result, torch.Tensor):
             result = torch.from_numpy(result)
+        elif isinstance(result, (bool, int, float)):
+            pass
         else:
-            result = tuple(torch.from_numpy(x) for x in result)
+            raise ValueError(f"Unhandled return type {type(result)}")
         return result
     return compiled_callable
```

qingyunqu:
@ramiro050 Please review this PR again. Thanks!

ramiro050 (approved):
Great, thanks!

PR #1873, opened 2023-02-12
[Torch Dialect] Add support for aten::linspace

Groverkss:


python/torch_mlir_e2e_test/test_suite/arange.py
```diff
@@ -248,3 +248,66 @@ def forward(self):
...
+class LinspaceModuleFloat32(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.linspace(3, 10, 5, dtype=torch.float32)
```
vivekkhandelwal1, line 264: Use `torch.ops.aten.` in place of `torch.`. Do this for all the tests.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -1304,6 +1310,16 @@ void TypeAnalysis::visitAtenArangeOp(AtenArangeOp op) {
...
+void TypeAnalysis::visitAtenLinspaceOp(AtenLinspaceOp op) {
+  auto knowledge =
+      ValueKnowledge::getTensorPessimisticValueState(op->getContext());
+  // TODO: Should get the dtype from torch.get_default_dtype().
+  // For now, use float32 which is the initial default dtype.
+  knowledge.dtype = getDtypeOrDefault(op->getContext(), op.getDtype(),
+                                      Float32Type::get(op->getContext()));
+  incorporateKnowledge(op->getResult(0), knowledge);
+}
+
```
vivekkhandelwal1, line 1322: Can't you add this op to the list here: https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/RefineTypes.cpp#L680

vivekkhandelwal1 (changes requested):
Also, I think by mistake you have also pushed the llvm and mhlo bump. That you need to remove from this PR.

lib/Dialect/Torch/Transforms/RefineTypes.cpp
Groverkss: I don't think so. Those operations always have the default type. This operation uses the default type when dtype is None. This operation's signature is similar to aten::arange behavior and has a similar visit function.

python/torch_mlir_e2e_test/test_suite/arange.py
Groverkss: Fixed

python/torch_mlir_e2e_test/test_suite/arange.py
```diff
@@ -248,3 +248,66 @@ def forward(self):
...
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.ops.aten.linspace(-100, 100, 8, dtype=torch.int64)
+
+@register_test_case(module_factory=lambda: LinspaceModuleNegativeInt64())
+def LinspaceModuleNegativeInt64_basic(module, tu: TestUtils):
```
ramiro050, line 312: nit: can you also add a test with float values for `start` and `end`?

Groverkss:
Taken up by https://github.com/llvm/torch-mlir/pull/2430

PR #1874, opened 2023-02-12
python: separate build- and test-related pip dependencies

ashay:
We want to ensure that pip packages required for building torch-mlir
should be included in the dependencies of torch-mlir, but we don't want
the pip packages required for _testing_ of torch-mlir to be included
among the dependencies.  To be able to specify and install one set of
dependencies and not the other, this patch separates the pip packages
into two files: build-requirements.txt and test-requirements.txt.

This patch also updates references to the requirements.txt file so that
CI builds that run end-to-end tests install test-related pip
dependencies while everything else (including WHL builds) sticks to just
the build-related pip dependencies.

Despite this change, this patch should not affect a torch-mlir
developer's workflow.  More precisely, since this patch makes the
top-level requirements.txt file refer to both build-requirements.txt and
test-requirements.txt files, a torch-mlir developer should be able to
continue referring to the requirements.txt file without any impact.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,3 @@
--r build-requirements.txt
 -r pytorch-requirements.txt
-
-# Test Requirements
-pillow
-dill
-multiprocess
+-r build-requirements.txt
+-r test-requirements.txt
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -26,7 +26,8 @@ runs:
 
   - name: Install PyTorch nightly depends
     run: |
-      python -m pip install -r requirements.txt
+      python -m pip install -r pytorch-requirements.txt
+      python -m pip install -r build-requirements.txt
     shell: bash
 
   - name: Install prerequisites (Linux)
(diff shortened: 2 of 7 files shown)


requirements.txt
```diff
@@ -1,6 +1,2 @@
 -r build-requirements.txt
-
-# Test Requirements
-pillow
-dill
-multiprocess
+-r test-requirements.txt
```
makslevental, line 2: this is gonna need a rebase as i moved these around

makslevental:
@ashay I haven't been tracking this work stream but I might've thrown a small monkey wrench in your works since I refactored `requirements.txt`: https://github.com/llvm/torch-mlir/blob/main/requirements.txt#L1-L2

ashay:
> @ashay I haven't been tracking this work stream but I might've thrown a small monkey wrench in your works since I refactored `requirements.txt`: https://github.com/llvm/torch-mlir/blob/main/requirements.txt#L1-L2

Oh, no worries at all!  I saw your PR and it looks like it will not be too hard to merge.  I'm currently waiting for the GitHub builder issue to be resolved, after which I'll rebase.

Thanks for the heads up, though!

makslevental (approved):
LGTM!

powderluv (approved):
Lets land it and I can help watch for any breakages. Maybe worth a oneshot build too.

ashay:
I now have doubts about whether this drops the torchvision dependency from the WHL files.  It probably does for the macOS and Windows WHL files but I am not sure about Linux.

I am merging it now, and I'll start a one shot release build, but I'll come back to this tomorrow and see if the dependencies are updated as we expect.

PR #1875, opened 2023-02-12
[Torch Dialect] Add aten.alias and it's folder

Groverkss:


vivekkhandelwal1 (changes requested):
Can you please add an e2e test case for this?

PR #1877, opened 2023-02-13
Set torch / torchvision versions in sync. 

powderluv:
rollback to yesterday's pytorch since today's torchvision depends on yesterday's torch.


powderluv:
This seems to fix the temporary version conflict with torch and torchvision. Hopefully this is just required until we get @ashay patch in.

ashay (approved):

PR #1884, opened 2023-02-16
[Torch Dialect] Scatter reduce lowering

cetiniz:
- Lowers the torch.scatter_reduce to linalg_on_tensors dialect.
- Includes support for "sum", "prod", "amax", "amin" and "mean".
- Adds E2E tests for "sum", "prod", "amax", "amin" and "mean" for support with both 'include_self' and without 'include_self'.


cetiniz:
I will update the PR with E2E tests for integer types

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -26,6 +26,9 @@
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
+#include "llvm/ADT/APFloat.h"
+#include "llvm/Support/ErrorHandling.h"
+#include <limits>
```
cetiniz, line 31: Remove

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -552,6 +569,272 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+    indSliceExpr.push_back(rewriter.getAffineConstantExpr(0));
+    SmallVector<AffineMap> mapping(
+        indexType.getRank(), AffineMap::get(/*dimCount=*/1, /*symbolCount=*/0,
+                                  indSliceExpr, context));
+    // mapping for updates
+    mapping.push_back(rewriter.getDimIdentityMap());
+    SmallVector<utils::IteratorType> iteratorTypes({utils::IteratorType::parallel});
+
+    // Need to flatten the scatter argument
+    auto tmTensorScatterInputs = rewriter.create<linalg::GenericOp>(
```
cetiniz, line 639: Potentially refactor into general function so other lowerings can use

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -552,6 +569,272 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+    }
+    // Cast to static shapes
+    SmallVector<int64_t> staticShape({ShapedType::kDynamic, indexType.getRank()});
+
+    auto staticType = RankedTensorType::get(staticShape, rewriter.getI32Type());
+    indices = rewriter.create<tensor::CastOp>(loc, staticType, indices);
+
+    Value updates = scatterInputsVector[indexType.getRank()];
+
+    // @TODO: Create special count tensor to calculate the 'mean'
```
cetiniz, line 702: Remove

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Done

lib/Dialect/Torch/Transforms/RefineTypes.cpp
```diff
@@ -664,7 +664,7 @@ void TypeAnalysis::visitOperation(Operation *op,
           AtenUnsqueezeOp, AtenViewOp, Aten_UnsafeViewOp, AtenReshapeOp,
           Aten_ReshapeAliasOp, AtenResize_Op, AtenTransposeIntOp, AtenTOp,
           AtenPermuteOp, AtenIndexSelectOp, AtenSelectIntOp,
-          AtenSelectScatterOp, AtenNarrowOp, AtenSliceTensorOp,
+          AtenSelectScatterOp, AtenNarrowOp, AtenSliceTensorOp,AtenScatterReduceTwoOp,
```
ataheridezfouli-groq, line 667: Do you still need to add this here?

cetiniz: For now yes, I will have to add this to the dtype branch at some point as well

python/torch_mlir_e2e_test/test_suite/indirect_memory.py
```diff
@@ -0,0 +1,158 @@
...
+
+import torch
+
+from torch_mlir_e2e_test.framework import TestUtils
+from torch_mlir_e2e_test.registry import register_test_case
+from torch_mlir_e2e_test.annotations import annotate_args, export
+
+# ==============================================================================
+
+class ScatterReduceModule(torch.nn.Module):
```
cetiniz, line 14: Should be called ScatterReduceFloatModule

cetiniz: Done

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Done

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -552,6 +568,272 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+    Value counts = nullptr;
+    if (reduceType == "mean") {
```
ramiro050, line 701: Can you copy

https://github.com/pytorch/pytorch/blob/0a9c608461f53a6c7dc019199f6c93860efb4c0b/aten/src/ATen/native/ReductionType.h#L7-L23

into

https://github.com/llvm/torch-mlir/blob/main/include/torch-mlir/Dialect/Torch/Utils/TorchUpstream.h

and use the enum instead of the strings?

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
ramiro050: Yeah, I think extracting the logic that transforms a PyTorch scatter indices tensor to the indices tensor tm_tensor expects can be refactored out into a helper function to make the lowering for this op easier to digest.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -552,6 +568,270 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+        auto minAttr = getNumericLimit(rewriter, srcType.getElementType(), /*getMin=*/true);
+        normalizationValue = rewriter.create<arith::ConstantOp>(loc, minAttr);
+      } else if (reduceType == "amin") {
+        // Set the values in the input tensor to the largest element of that type
+        auto maxAttr = getNumericLimit(rewriter, srcType.getElementType(), /*getMin=*/false);
+        normalizationValue = rewriter.create<arith::ConstantOp>(loc, maxAttr);
+      }
+      Value normalizations = createInitTensor(
+          rewriter, loc, SmallVector<Value>({indexSize}),
+          srcType.getElementType(), /*init_element*/ normalizationValue);
```
ramiro050, line 742: nit: the format for these comments is `/*init_element=*/normalizationValue`

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -552,6 +568,270 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+      self = createTMTensorScatterOp(
+          rewriter, loc, normalizations, indices, self,
+          /*uniqueIndices=*/false,
+          [&](OpBuilder &b, Location loc, Value update, Value current) {
+            b.create<TMTensor::YieldOp>(loc, update);
+      });
+      if (reduceType == "mean") {
+          counts = createTMTensorScatterOp(
```
ramiro050, line 750: Is `counts` meant to be the same as `self`? If so, is this scatter needed?

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -552,6 +568,270 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+            } else {
+              result = b.create<arith::MaxFOp>(loc, update, current);
+            }
+          } else if (reduceType == "amin") {
+            if (update.getType().isa<mlir::IntegerType>()) {
+              result = b.create<arith::MinSIOp>(loc, update, current);
+            } else {
+              result = b.create<arith::MinFOp>(loc, update, current);
+            }
+          }
```
ramiro050, line 789: nit: add an else with an `llvm_unreachable` to make it explicit that every case is being handled

python/torch_mlir_e2e_test/test_suite/__init__.py
```diff
@@ -47,3 +47,4 @@ def register_all_tests():
     from . import return_types
     from . import control_flow
     from . import stats
+    from . import indirect_memory
```
ramiro050, line 50: I think we can put these with the `index_put` tests. We can rename the file to `scatter`

python/torch_mlir_e2e_test/test_suite/indirect_memory.py
```diff
@@ -0,0 +1,158 @@
...
+        super().__init__()
+        self.include_self = include_self
+        self.reduce_type = reduce_type
+
+    @export
+    @annotate_args([
+        None,
+        ([10, 8, 6], torch.float32, True),
+        ([2, 4, 3], torch.int64, True),
+        ([5, 8, 6], torch.float32, True),
```
ramiro050, line 28: Does the code not work with dynamic shapes?

python/torch_mlir_e2e_test/test_suite/indirect_memory.py
```diff
@@ -0,0 +1,158 @@
...
+        ([10, 8, 6], torch.int32, True),
+        ([2, 4, 3], torch.int64, True),
+        ([5, 8, 6], torch.int32, True),
+    ])
+    def forward(self, input, index, src):
+        return torch.ops.aten.scatter_reduce(input, 0, index, src, self.reduce_type, include_self=self.include_self)
+
+
+@register_test_case(
+    module_factory=lambda: ScatterReduceIntModule("sum", False))
```
ramiro050, line 108: Thanks for the thorough testing :)

ramiro050 (changes requested):
Great job! This is a very complicated lowering. I have a few initial comments

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Counts stores the number of times an index is reduced into. It is initially set to one so that dividing elements that are not scattered into remain the same. The initial scatter into counts (if 'include_self' is false), is because we will be using counts to do the division later and need to zero out the indices that should not be included to calculate the mean

ramiro050: Ah, I see. Missed the fact that the tensor used in the scatter is `counts` itself haha. Makes sense

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Done: let me know if there should be any other simplifications

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -52,6 +54,20 @@ using namespace mlir::torch::TMTensor;
 // that these patterns become mostly mechanical associations of
 // "aten.foo -> linalg.foo".
 
+static Attribute getNumericLimit(PatternRewriter& rewriter, Type elementType, bool getMin = true) {
+    auto bitWidth = elementType.getIntOrFloatBitWidth();
```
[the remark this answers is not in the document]
cetiniz: Should be done

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Done

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Added this to all cases

python/torch_mlir_e2e_test/test_suite/__init__.py
cetiniz: Refactored into `scatter.py`

python/torch_mlir_e2e_test/test_suite/indirect_memory.py
cetiniz: No problemo :)

python/torch_mlir_e2e_test/test_suite/indirect_memory.py
cetiniz: Switched the annotations to make them dynamic

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Fixed annotation

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -552,6 +568,272 @@ class ConvertAtenMaxPool2dWithIndicesBackwardOp
...
+      SmallVector<Value> sizes = getTensorSizes(rewriter, loc, slice);
+      indices = rewriter.createOrFold<tensor::InsertSliceOp>(
+          loc, slice, indices,
+          llvm::to_vector(llvm::map_range(offsets, toOpFoldResult)),
+          llvm::to_vector(llvm::map_range(sizes, toOpFoldResult)),
+          llvm::to_vector(llvm::map_range(strides, toOpFoldResult)));
+      offsets[1] =
+          rewriter.createOrFold<arith::AddIOp>(loc, offsets[1], cstOne);
+    }
+    // Cast to static shapes
```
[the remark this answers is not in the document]
cetiniz: Removed, you are right

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -52,6 +54,20 @@ using namespace mlir::torch::TMTensor;
...
+            return rewriter.getIntegerAttr(elementType, APInt::getSignedMinValue(bitWidth));
+        } else {
+            return rewriter.getIntegerAttr(elementType, APInt::getSignedMaxValue(bitWidth));
+        }
+    } else if (mlir::FloatType floatType = llvm::dyn_cast<mlir::FloatType>(elementType)) {
+        return rewriter.getFloatAttr(elementType, APFloat::getLargest(floatType.getFloatSemantics(), getMin));
+    } else {
+      llvm_unreachable("Only float/integer types are supported!");
+    }
+}
```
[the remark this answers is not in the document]
cetiniz: Done

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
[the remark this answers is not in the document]
ramiro050: This isn't right. The formatting should make the indentation two spaces rather than four. Depending on how you have your changes, you have to run `clang-format` differently. For example, if you have changes staged, `git clang-format` will only format the staged changes, and this will result as new unstaged changes. However, if your changes are already in a commit, then you need to do `git clang-format HEAD~1` to format the latest commit. This results in new changes that you then need to stage and commit. (You can also do multiple commits by changing `HEAD~1` to whatever range you need).

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -52,6 +56,133 @@ using namespace mlir::torch::TMTensor;
...
+
+// This function will reformat the `index` and `src` from torch operations 
+// like `torch.scatter` or `torch.scatter_reduce` to match the expected 
+// input for the TMScatterOp.
```
ramiro050, line 76: Nit: add info on what the return type of this function means

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -52,6 +56,133 @@ using namespace mlir::torch::TMTensor;
...
+static std::pair<Value,Value> formatTMTensorScatterOpInputs(
+        PatternRewriter& rewriter, Value self, Value indices, Value src, int64_t dim) {
```
ramiro050, line 78: I don't think `self` is needed in this function

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -52,6 +56,133 @@ using namespace mlir::torch::TMTensor;
...
+        return rewriter.getFloatAttr(elementType, APFloat::getLargest(floatType.getFloatSemantics(), getMin));
+    } else {
+      llvm_unreachable("Only float/integer types are supported!");
+    }
+}
+
+// This function will reformat the `index` and `src` from torch operations 
+// like `torch.scatter` or `torch.scatter_reduce` to match the expected 
+// input for the TMScatterOp.
+static std::pair<Value,Value> formatTMTensorScatterOpInputs(
```
ramiro050, line 77: nit: I think we can have a more descriptive name. Maybe something like `convertTorchScatterIndexAndSrcToTMScatterIndexAndSrc`. A bit long, but gets the point across

ramiro050 (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Changed function name

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Refactored to remove self from function parameters

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
cetiniz: Added additional information on function definition including return type and what it should be used for

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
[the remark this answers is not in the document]
cetiniz: Ah I think this is what I was missing, I will change the range and check it over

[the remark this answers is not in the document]
cetiniz: I changed the range of commits and reformatted using clang-format which worked (sorry, my fault here). It should now all be formatted properly and two-spaced instead of four

ramiro050 (approved):
Thanks! LGTM

PR #1885, opened 2023-02-16
update llvm & mlir

qiuxiafei:


ramiro050:
@ashay @powderluv, any idea what could be causing the MacOS CI failure?

powderluv:
I'm without proper connectivity till Tuesday to take a look. @stellaraccident  in case she has any idea. Looks like an upstream LLVM dependency ?

ashay:
Like Anush said, this could be because of an upstream bug (it seems like a patch to mlir/test/CMakeLists.txt was reverted the same day, but our chosen commit doesn't include the revert).

In picking the LLVM green commit each week, we run Windows and Linux builds but not macOS builds, which probably caused a bad commit to slip.  I'll add macOS to our green commit builds so that we don't run into problems like this in the future.

ramiro050:
> I'll add macOS to our green commit builds so that we don't run into problems like this in the future.

Thanks!

ashay:
Closing because we have merged LLVM after this PR was created.

PR #1889, opened 2023-02-16
[Tcp] Add group, isolated_group, and yield ops

navahgar:
This PR adds grouping ops `tcp.group`, `tcp.isolated_group` and a `tcp.yield` op.

### DIFF
--- a/externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
+++ b/externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
@@ -191,4 +191,95 @@ def Tcp_BroadcastOp : Tcp_Op<"broadcast", [
   let hasVerifier = 1;
 }
 
+def Tcp_YieldOp : Tcp_Op<"yield", [Terminator, Pure]> {
+  let summary = "Yields values from regions in Tcp";
+
+  let description = [{
+    Terminates and yields values from regions in Tcp.
+
+    The values that are yielded should correspond to the parent operation's
+    results.
+  }];
+
+  let arguments = (ins
+    Variadic<Tcp_TensorOrScalar>:$ins
+  );
+
+  let builders = [
+    OpBuilder<(ins), [{ }]>,
+  ];
+
+  let assemblyFormat = "$ins attr-dict `:` type($ins)";
+}
+
+def Tcp_GroupOp : Tcp_Op<"group", [
+          AffineScope,
+          ParentOneOf<["func::FuncOp", "tcp::GroupOp", "tcp::IsolatedGroupOp"]>,
+          SingleBlockImplicitTerminator<"tcp::YieldOp">]> {
+  let summary = "Groups ops into a region that is not isolated from above";
+
+  let description = [{
+    Provides a way to group ops into a region that is not isolated from above.
+
+    Since this is not isolated from above, ops inside this region can refer to
+    values in the outer scopes. So, this op does not have any input parameters.
+
+    This op must not include control flow ops in its region. However, it can
+    be present within regions that are inside control flow ops.
+
+    The enclosed region will terminate with a `tcp.yield` op. This ops results
+    should correspond to the values that are yielded by the terminating op.
+  }];
+
+  let results = (outs
+    Variadic<Tcp_TensorOrScalar>:$outs
+  );
+
+  let regions = (region
+    SizedRegion<1>:$body
+  );
+
+  let assemblyFormat = "attr-dict-with-keyword $body `:` type($outs)";
+
+  let hasVerifier = 1;
+}
+
+def Tcp_IsolatedGroupOp : Tcp_Op<"isolated_group", [
+          AffineScope,
+          IsolatedFromAbove,
+          ParentOneOf<["func::FuncOp", "tcp::GroupOp", "tcp::IsolatedGroupOp"]>,
+          SingleBlockImplicitTerminator<"tcp::YieldOp">]> {
+  let summary = "Groups ops into a region that is isolated from above";
+
+  let description = [{
+    Provides a way to group ops into a region that is isolated from above.
+
+    Since the enclosed region is isolated from above, any use of values from
+    outer scopes requires passing that value as an input parameter to this op
+    and using that parameter instead.
+
+    This op must not include control flow ops in its region. However, it can
+    be present within regions that are inside control flow ops.
+
+    The enclosed region will terminate with a `tcp.yield` op. This ops results
+    should correspond to the values that are yielded by the terminating op.
+  }];
+
+  let arguments = (ins
+    Variadic<Tcp_TensorOrScalar>:$ins
+  );
+
+  let results = (outs
+    Variadic<Tcp_TensorOrScalar>:$outs
+  );
+
+  let regions = (region
+    SizedRegion<1>:$body
+  );
+
+  let assemblyFormat = "$ins attr-dict-with-keyword $body `:` type($ins) `->` type($outs)";
+
+  let hasVerifier = 1;
+}
+
 #endif // TORCH_MLIR_DIALECT_TCP_OPS
--- a/externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/misc.mlir
+++ b/externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/misc.mlir
@@ -62,3 +62,122 @@ func.func @test_broadcast_axes_w_duplicates(%arg0 : tensor<?x1x?x1xf32>, %arg1 :
   %0 = "tcp.broadcast"(%arg0, %arg1, %arg2) {axes = [1, 1]} : (tensor<?x1x?x1xf32>, index, index) -> tensor<?x?x?x?xf32>
   return %0 : tensor<?x?x?x?xf32>
 }
+
+// -----
+
+// CHECK-LABEL: func.func @test_group(
+// CHECK-SAME:          %[[ARG0:.*]]: tensor<?x?xf32>,
+// CHECK-SAME:          %[[ARG1:.*]]: tensor<?x?xf32>) -> tensor<?x?xf32>
+// CHECK:         %[[GROUP:.*]] = tcp.group {
+// CHECK:            %[[ADD:.*]] = tcp.add %[[ARG0]], %[[ARG1]] : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:            %[[TANH:.*]] = tcp.tanh %[[ADD]] : tensor<?x?xf32> -> tensor<?x?xf32>
+// CHECK:            tcp.yield %[[TANH]] : tensor<?x?xf32>
+// CHECK:         } : tensor<?x?xf32>
+// CHECK:         return %[[GROUP]] : tensor<?x?xf32>
+func.func @test_group(%arg0 : tensor<?x?xf32>, %arg1 : tensor<?x?xf32>) -> tensor<?x?xf32> {
+  %10 = "tcp.group" () ({
+    ^bb0() :
+      %2 = tcp.add %arg0, %arg1 : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+      %3 = tcp.tanh %2 : tensor<?x?xf32> -> tensor<?x?xf32>
+      tcp.yield %3 : tensor<?x?xf32>
+  }) : () -> tensor<?x?xf32>
+  return %10 : tensor<?x?xf32>
+}
+
+// -----
+
+func.func @test_group_no_yield(%arg0 : tensor<?x?xf32>, %arg1 : tensor<?x?xf32>) -> tensor<?x?xf32> {
+  // expected-error @+1 {{'tcp.group' op failed to verify that op region ends with a terminator}}
+  %10 = "tcp.group" () ({
+    ^bb0() :
+      %2 = tcp.add %arg0, %arg1 : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+      %3 = tcp.tanh %2 : tensor<?x?xf32> -> tensor<?x?xf32>
+  }) : () -> tensor<?x?xf32>
+  return %10 : tensor<?x?xf32>
+}
+
+// -----
+
+func.func @test_group_incorrect_yield_args(%arg0 : tensor<?x?xf32>, %arg1 : tensor<?x?xf32>) -> tensor<?x?xf32> {
+  // expected-error @+1 {{'tcp.group' op failed to verify that the number of yielded values is same as the number of results}}
+  %10 = "tcp.group" () ({
+    ^bb0() :
+      %2 = tcp.add %arg0, %arg1 : tensor<?x?xf32>, tensor<?x?xf32> -> tensor<?x?xf32>
+      %3 = tcp.tanh %2 : tensor<?x?xf32> -> tensor<?x?xf32>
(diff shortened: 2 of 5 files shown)


externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -191,4 +191,92 @@ def Tcp_BroadcastOp : Tcp_Op<"broadcast", [
   let hasVerifier = 1;
 }
 
+def Tcp_YieldOp : Tcp_Op<"yield", [Terminator, Pure]> {
+  let summary = "Yields values from regions in Tcp";
+
+  let description = [{
+    Terminates and yields values from regions in Tcp.
```
jpienaar, line 198: Will this be a generic one or per group?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -191,4 +191,92 @@ def Tcp_BroadcastOp : Tcp_Op<"broadcast", [
...
+    The values that are yielded should correspond to the parent operation's
+    results.
+  }];
+
+  let arguments = (ins
+    Variadic<Tcp_TensorOrScalar>:$ins
+  );
+
+  let builders = [
+    OpBuilder<(ins), [{ }]>,
```
jpienaar, line 209: If you include InferReturnType.td (or whatever it is called) do these get generated?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -191,4 +191,92 @@ def Tcp_BroadcastOp : Tcp_Op<"broadcast", [
...
+
+    Since this is not isolated from above, ops inside this region can refer to
+    values in the outer scopes. So, this op does not have any input parameters.
+
+    The enclosed region will terminate with a `tcp.yield` op. This ops results
+    should correspond to the values that are yielded by the terminating op.
+  }];
+
+  let arguments = (ins
+    StrAttr:$group_type
```
jpienaar, line 233: Could you perhaps add document for argument?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -191,4 +191,92 @@ def Tcp_BroadcastOp : Tcp_Op<"broadcast", [
...
+    OpBuilder<(ins), [{ }]>,
+  ];
+
+  let assemblyFormat = "$ins attr-dict `:` type($ins)";
+}
+
+def Tcp_GroupOp : Tcp_Op<"group", [
+          AffineScope,
+          AutomaticAllocationScope,
+          ParentOneOf<["func::FuncOp", "tcp::GroupOp", "tcp::IsolatedGroupOp"]>,
```
ThomasRaoux, line 218: Groups can be nested? Also they can't be in control flow?

ThomasRaoux (approved):
Looks good to me overall

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Every group should have a region that terminates with `tcp.yield` op. Is that what you are asking about?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: I tried including a few different `.td` files and none of them helped generate this. 

@sjarus Do you know what to include for this?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Sure, added that.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Yes, they can be nested. 

Good question regarding control flow. Do you see any problems with groups inside control flow? I didn't add them here because we don't have them yet in Tcp.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sjarus: I assume @jpienaar means the inferReturnTypeComponents interface. You can check out include/mlir/Dialect/Tosa/IR/TosaOps.td for usage of that.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sjarus: My view of cflow and groups is that they should be ok because groups are not a runtime artifact but a compile-time aggregation for the purposes of aiding codegen.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
jpienaar: Ah this one I figured as variadic should allow empty there.

externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/misc.mlir
```diff
@@ -62,3 +62,48 @@ func.func @test_broadcast_axes_w_duplicates(%arg0 : tensor<?x1x?x1xf32>, %arg1 :
   %0 = "tcp.broadcast"(%arg0, %arg1, %arg2) {axes = [1, 1]} : (tensor<?x1x?x1xf32>, index, index) -> tensor<?x?x?x?xf32>
   return %0 : tensor<?x?x?x?xf32>
 }
+
+// -----
+
+// CHECK-LABEL: func.func @test_group(
```
sanjoy, line 68: Do we need tests that check the verifier fails as intended (e.g. when there is no `yield` op)?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -191,4 +191,100 @@ def Tcp_BroadcastOp : Tcp_Op<"broadcast", [
...
+
+  let description = [{
+    Provides a way to group ops into a region that is not isolated from above.
+
+    Since this is not isolated from above, ops inside this region can refer to
+    values in the outer scopes. So, this op does not have any input parameters.
+
+    The group_type attribute is a free-form string that can be used to denote
+    any aspect of the group. For example, it could be set to "elementwise" for
+    a group that only contains elementwise ops.
```
sanjoy, line 230: Please note here what semantic effect the `group_type` string has, if any.  E.g. if it is `"elementwise"` then does that mean the group body cannot have non-elementwise ops?

My suggestion: I suggest we say that `group_type` is purely a hint and has no semantic meaning.  So even if it is `"elementwise"`, a downstream pass cannot rely on this information for correctness and must explicitly check the body of the group op.  Maybe it should be called `group_hint` actually.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -191,4 +191,100 @@ def Tcp_BroadcastOp : Tcp_Op<"broadcast", [
...
+    Provides a way to group ops into a region that is isolated from above.
+
+    Since the enclosed region is isolated from above, any use of values from
+    outer scopes requires passing that value as an input parameter to this op
+    and using that parameter instead.
+
+    The group_type attribute is a free-form string that can be used to denote
+    any aspect of the group. For example, it could be set to "elementwise" for
+    a group that only contains elementwise ops.
+
```
sanjoy, line 269: Same comment about the semantics of `group_type`.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sanjoy: How about control flow within groups?

sanjoy (changes requested):
Mostly LG, but I have some questions inline about `group_type` and control flow.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
MaheshRavishankar: Can you explain more why groups are not runtime artifact? I am guessing they correspond to some unit of work that is somehow "launched".

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
MaheshRavishankar: I would completely drop it from the op definition. It seems too vague and free form and is just a StrAttr. Maybe with the properties proposal, such things won't be necessary. Clients can add properties to ops to "classify" groups

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: > How about control flow within groups?

Both grouping ops will not have control flow within them. The reasoning behind that is just to keep the transformations to move ops into and out of groups simple. Just clarified this in the description. 

Do you think we need to support control flow in groups?

navahgar: > Can you explain more why groups are not runtime artifact?

Grouping ops in Tcp are meant to facilitate lowering efficiently to different backends. We do not foresee these grouping ops to be present once we lower from Tcp to a backend dialect.

externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/misc.mlir
navahgar: Typically, I don't add tests for the cases that are controlled by the traits, which is the case for the presence of a `yield` op. But I realized that the traits don't verify other constraints: the number and type of `yield` args matching the result.

Add verifier to check for all of those and also added appropriate tests.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sanjoy: > Do you think we need to support control flow in groups?

No explicit control flow.  I don't see how we can prevent "implicit" control flow though -- a functional `while` loop for instance -- as long as we have allow non-TCP operations in groups.

We should have a verifier check for this.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Good points regarding `group_type`. As outlined in the RFC doc, we are not clear how this is going to be used. In this situation I agree that having such a free-from string attribute may make things more confusing / erroneous. So, removed the attribute now.

If properties are meant to handle such cases, that would be great. We can look to use that once it is available.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Removed that attribute.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Your point regarding implicit control flow is valid. But how do you propose checking that the group region does not include any such ops? I don't see them implementing any trait / interface.

MaheshRavishankar: > > Can you explain more why groups are not runtime artifact?
> 
> Grouping ops in Tcp are meant to facilitate lowering efficiently to different backends. We do not foresee these grouping ops to be present once we lower from Tcp to a backend dialect.

Well, the grouping gets lowered somehow to the backend. I think traits like `AutomaticAllocationScope` etc. are related to runtime (someone needs to do the allocation). So my sense is that there is some bearing on runtime hand-shake here that isnt accounted for.

navahgar: @MaheshRavishankar Just replied to your comment in the doc: https://docs.google.com/document/d/1jHXI-ZSyt-3nFlO0ZtQZX0E41k4mCmuO7djmrvpyHms/edit?disco=AAAAq40iWww

navahgar: Regarding verifying for control flow ops, clarified with @sanjoy offline. By construction these ops only allow one basic block (because their body is a `SizedRegion<1>`). There is no other feasible way to catch all control flow ops AFAICT.

sanjoy (approved):

navahgar:
Discussed with @MaheshRavishankar offline. He agreed to land this PR as is now.

We have an ongoing discussion regarding the runtime aspects of grouping ops [here](https://docs.google.com/document/d/1jHXI-ZSyt-3nFlO0ZtQZX0E41k4mCmuO7djmrvpyHms/edit?disco=AAAAq40iWww). Once we have a conclusion on that discussion, we can update these ops to reflect any changes we need.

PR #1892, opened 2023-02-19
Add dtype functions for ops that take dtype from 1st operand

li-plus:
This PR finished the task 1 of #1807.

The three crashing cases (`ElementwiseClampModule_basic`, `IouOfModule_basic`, `NormalizeModule_basic`) are related to `clamp` op whose dtype function accepts optional non-tensor arguments.
```python
def aten〇clamp〡dtype(self_rank_dtype: Tuple[int, int], min: Optional[Union[int, float]] = None, max: Optional[Union[int, float]] = None) -> int:
    ...
```

They are crashing on the `assert` below. @ramiro050 do you have any idea for this?
https://github.com/llvm/torch-mlir/blob/63945a2fd411e768186dcc82a54c56a2a9470cbb/lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp#L180-L186


li-plus:
The CI kept failing at strange point. Trying to close and reopen it.

PR #1895, opened 2023-02-22
Add dtype functions for ops that take dtype from 1st operand

li-plus:
This PR finished the task 1 of #1807.

The three crashing cases (`ElementwiseClampModule_basic`, `IouOfModule_basic`, `NormalizeModule_basic`) are related to `clamp` op whose dtype function accepts optional non-tensor arguments.
```python
def aten〇clamp〡dtype(self_rank_dtype: Tuple[int, int], min: Optional[Union[int, float]] = None, max: Optional[Union[int, float]] = None) -> int:
    ...
```

They are crashing on the `assert` below. @ramiro050 do you have any idea for this?
https://github.com/llvm/torch-mlir/blob/63945a2fd411e768186dcc82a54c56a2a9470cbb/lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp#L180-L186


ramiro050:
This seems to be an issue with our importer. If you look at the MLIR snippet generated for `clamp`, the type `Optional[Union[int, float]]` is being converted to `!torch.union<int, float, none>` instead of `!torch.optional<union<int, float>>`.

ramiro050:
Actually, the type `Union[int, float, None]` is being created by `torch.jit.script`. I think it should simply be a matter of widening the types allowed by the assert to include union types that have `None` as one of the contained types

li-plus:
> Actually, the type `Union[int, float, None]` is being created by `torch.jit.script`. I think it should simply be a matter of widening the types allowed by the assert to include union types that have `None` as one of the contained types

Thanks! This is exactly the case. I fixed it by derefining these optional scalars to type `!torch.union<int, float, none>`.

Once https://github.com/llvm/torch-mlir/pull/1990 is merged, I will switch to meta tensors and resolve the WIP status.

li-plus:
@ramiro050 This PR is ready to merge. Would you review it? Thanks.

ramiro050 (approved):
Awesome!

PR #1899, opened 2023-02-24
[Torch Dialect] add RSub, ScalarImplicit canonicalize

zhekunz2:


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -877,8 +877,11 @@ LogicalResult rewrite0DBinaryTensorOp(Operation *op,
     return rewriter.notifyMatchFailure(
         op, "only int scalar lhs or rhs is supported");
   }
-  if (isa<AtenSubTensorOp, AtenSubScalarOp, AtenAddTensorOp, AtenAddScalarOp>(
-          op)) {
+  if (isa<AtenRsubScalarOp>(op)) {
+    std::swap(lhs, rhs);
```
ramiro050, line 881: nit: I would prefer being more explicit here to avoid confusion as to what `lhs` and `rhs` represent. 

Here I would instead insert an `if` inside the `if` with all the `sub` and `add` ops:

```c++
if (isa<AtenSubTensorOp, ...>(op)) {
  if (isa<AtenRsubScalarOp>(op))
    rhs = rewriter.create<AtenMulIntOp>(loc, lhs, alpha);
  else
    rhs = rewriter.create<AtenMulIntOp>(loc, rhs, alpha);
}
```

And below where `AtenSubIntOp` gets created, I would handle `AtenRsubScalarOp` on its own.

This is a bit more verbose, but it makes sure someone reading the code does not need context from different places to understand how things work.

ramiro050 (changes requested):
Thanks for this! I just have a small change request

ramiro050:
The CI is failing on the lit tests. You can run these tests locally by doing `cmake --build build --target check-torch-mlir`.

ZihengJiang:
@ramiro050 Hi Ramiro, has your concern been addressed? How to rerun the CI?

ramiro050 (changes requested):

ramiro050:
> How to rerun the CI?

I've never seen this happen to the CI... @powderluv any ideas?

powderluv:
yeah haven't seen this - maybe something on the github side

powderluv:
can you try to update your commit message and do a push again ?

ramiro050 (approved):
LGTM!

e2e_testing/xfail_sets.py
```diff
@@ -316,6 +316,7 @@
     "RsubFloatModule_noalpha_basic",
     "RsubIntModule_basic",
     "RsubIntModule_noalpha_basic",
+    "RsubInt0d_NumToTensor_Module",
```
ramiro050, line 319: This should be the name of the Python function that gets registered as the test. In this case, it should be: `RsubInt0d_NumToTensor_Module_basic`

ZihengJiang:
@zhekunz2 Thanks for the hard work!

PR #1906, opened 2023-02-28

vivekkhandelwal1:
Set PyTorch and TorchVision version to nightly release 2023-02-27.
This commit also adds the lowering for aten.add and aten.Float.Scalar op.


powderluv (approved):

lib/Conversion/TorchToArith/TorchToArith.cpp
```diff
@@ -251,6 +251,50 @@ class ConvertTorchConstantIntOp
...
+    Value result =
+        convertScalarToDtype(rewriter, op.getLoc(), adaptor.getA(), resultType);
+    rewriter.replaceOp(op, result);
+    return success();
+  }
+};
+} // namespace
+
+namespace {
+class ConvertAtenAddOp : public OpConversionPattern<AtenAddOp> {
```
ramiro050, line 272: Can you also add an e2e test for this op?

ramiro050: Hm, yeah I can see that, since the inputs are supposed to be `Scalar`. I'll see if I can think of something

PR #1910, opened 2023-03-02
[TCP] Implements folder for tcp.const op (+ unit test)

sjain-stanford:
As titled. This is needed to unblock CI on `mlir-tcp` (results in failing assertion on `ConstantLike` ops without a folder).

Will need to refactor to use the new `FoldAdaptor` convention after `mlir-tcp` rebases to `main`.


navahgar (approved):
LGTM

Issue #1896, opened 2023-02-23
Failed when compiling RNN model to mlir 


Alaskra:
this is the input file tmp1.txt:
```
func.func @forward(%arg0: !torch.vtensor<[5,1,10],f32>) -> (!torch.vtensor<*,f32>) {
  %1 = torch.tensor_static_info_cast %arg0 : !torch.vtensor<[5,1,10],f32> to !torch.vtensor<*,f32>
  %2 = torch.copy.to_tensor %1 : !torch.tensor<*,f32>
  %3 = torch.tensor_static_info_cast %2 : !torch.tensor<*,f32> to !torch.tensor
  %false = torch.constant.bool false
  %16 = torch.prim.If %false -> (!torch.tensor) {
    torch.prim.If.yield %3: !torch.tensor
  } else {
    torch.prim.If.yield %3: !torch.tensor
  }
  %17 = torch.tensor_static_info_cast %16 : !torch.tensor to !torch.tensor<*,f32>
  // %17 = torch.tensor_static_info_cast %3 : !torch.tensor to !torch.tensor<*,f32>
  %22 = torch.copy.to_vtensor %17 : !torch.vtensor<*,f32>
  return %22: !torch.vtensor<*,f32>
}
```

when I run `torch-mlir-opt tmp1.mlir -mlir-print-ir-after-all -mlir-disable-threading --torch-refine-types`, an error happen:
```
tmp1.mlir:13:9: error: 'torch.copy.to_vtensor' op failed to verify that operand is corresponding !torch.tensor
  %22 = torch.copy.to_vtensor %17 : !torch.vtensor<*,f32>
        ^
tmp1.mlir:13:9: note: see current operation: %8 = "torch.copy.to_vtensor"(%7) : (!torch.tensor<*,f32>) -> !torch.vtensor
// -----// IR Dump After RefineTypes Failed (torch-refine-types) //----- //
"func.func"() ({
^bb0(%arg0: !torch.vtensor<[5,1,10],f32>):
  %0 = "torch.tensor_static_info_cast"(%arg0) : (!torch.vtensor<[5,1,10],f32>) -> !torch.vtensor<*,f32>
  %1 = "torch.copy.to_tensor"(%0) : (!torch.vtensor<*,f32>) -> !torch.tensor<*,f32>
  %2 = "torch.tensor_static_info_cast"(%1) : (!torch.tensor<*,f32>) -> !torch.tensor<*,f32>
[...]

PR #1911, opened 2023-03-02
Handle uninitialized lattice elements in RefineTypes

ramiro050:
The data-flow analysis does not always propagate information to the entire graph. This results in some lattice elements being uninitialized. Currently the lattice elements are not checked to see if they are uninitialized before rewriting the graph, potentially resulting in invalid IR (see
https://github.com/llvm/torch-mlir/issues/1896).

This commit adds handling for uninitialized lattice elements.


silvasean (approved):
Wow that's tricky.

PR #1913, opened 2023-03-03
[Torch Dialect] Add slice tensor canonicalize pattern

zhekunz2:
- Rewrite `AtenSliceTensorOp` to infer result shape when all inputs(start, end, etc.) are known and result shape is unknown. 
- Update ValueTensorType's subtype rules to consider unknown shapes.


ramiro050:
I would've expected the shape library to be able to propagete the shape information to the result of the `AtenSliceTensorOp`. If this is not the case, we should figure out why the shape library is not doing this.

PR #1914, opened 2023-03-03
Fix handling of non-int tensors in `getScalarValue`

ramiro050:
The current implementation of `getScalarValue` does not check that the input to a `ValueTensorLiteralOp` is an i64 before extracting the value, and it does not check that the result type of the `PrimNumToTensorScalarOp` is also an i64. This leads to crashes or invalid IR generated when the `input` is something other than an i64 tensor or `!torch.int`.

This commit addresses those issues. In addition, the function `getScalarValue` is renamed to `getScalarIntValue` to make it clear that it *only* extracts scalar integers.


vivekkhandelwal1 (approved):
LGTM!

PR #1916, opened 2023-03-05
add jit,no-jit release matrix

makslevental:


powderluv (approved):

powderluv:
since no-jit is a new feature we can enable only Python 3.11 whls to reduce the build matrix.

powderluv (approved):

silvasean:
In the end what is the plan here? Are we going to publish a separate wheel for just the mlir bits and then have a separate one with the torch_mlir.compile API and JIT importer?

powderluv:
yes. The core MLIR bits will be a torch-mlir-core package and a separate one with the torch_mlir.compile() API and JIT importer.

PR #1922, opened 2023-03-08
LTC->MLIR Debug Info support

GlebKazantaev:
### Description
Added support for LTC stack trace propagation up to torch MLIR locations. This option is turned off by default. To enable debug info you need to specify these options:
* `LTC_IR_DEBUG=1` - enables debug info.
* `LTC_IR_DEBUG_ROOT_PATH="/home/gleb/torch-mlir/:/home/gleb/pytorch/"` - list of paths which are used to trim absolute paths for readability _(optional)._

### Example
To check how it works you can run LTC MNIST test with reference backend:
```
LTC_IR_DEBUG=1 python ./torch-mlir/examples/ltc_backend_mnist.py -d MLIR_EXAMPLE
```

This is an example of `addmm` operation locations which consist of `function_name/file_name/line_number` per `callsite`:
```
%4 = torch.aten.addmm %arg4, %arg0, %1, %2, %3 : !torch.vtensor<[10],f32>, !torch.vtensor<[1,5],f32>, !torch.vtensor<[5,10],f32>, !torch.int, !torch.int -> !torch.vtensor<[1,10],f32> loc(callsite("forward"("pytorch/torch/nn/modules/linear.py":114:0) at callsite("_call_impl"("pytorch/torch/nn/modules/module.py":1488:0) at callsite("forward"("./torch-mlir/examples/ltc_backend_mnist.py":38:0) at callsite("_call_impl"("pytorch/torch/nn/modules/module.py":1488:0) at callsite("main"("./torch-mlir/examples/ltc_backend_mnist.py":56:0) at "<module>"("./torch-mlir/examples/ltc_backend_mnist.py":105:0)))))))
```

### DIFF
--- a/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
+++ b/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
@@ -211,7 +211,8 @@ void TorchMlirLoweringContext::AssignOutputOp(
 
   std::vector<std::string> source_files, functions;
   std::vector<int64_t> line_numbers;
-  const auto& frames = torch_mlir_node->metadata().frame_info;
+  const auto& metadata = torch_mlir_node->metadata();
+  const auto& frames = metadata.frame_info;
   if (!frames.empty()) {
     static std::vector<std::string> g_roots =
       string_split(sys_util::GetEnvString("LTC_IR_DEBUG_ROOT_PATH", ""), ":");
@@ -241,6 +242,10 @@ void TorchMlirLoweringContext::AssignOutputOp(
           c10::Symbol::attr("line_numbers"), line_numbers);
     }
   }
+  auto scope = ::c10::Symbol::scope(metadata.scope);
+  op->node()->setScope(
+    c10::make_intrusive<torch::jit::Scope>()->push(scope));
+
   emitted_outputs_[output] = std::move(op);
 }
 
--- a/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
+++ b/python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
@@ -15,6 +15,7 @@
 #include <torch/csrc/jit/api/compilation_unit.h>
 #include <torch/csrc/jit/passes/refine_tuple_types.h>
 #include <torch/csrc/lazy/core/lazy_graph_executor.h>
+#include <torch/csrc/lazy/core/config.h>
 #include "torch-mlir-c/Registration.h"
 #include "torch-mlir-c/Transforms.h"
 #include "mlir-c/IR.h"
@@ -205,13 +206,41 @@ void TorchMlirLoweringContext::AssignOutputOp(
     const Output& output, torch::jit::Value* op) {
   PRINT_FUNCTION();
 
-  // TODO (antoniojkim): Do we need this?
-  // auto torch_mlir_node =
-  //     NodeCast<TorchMlirNode>(output.node, output.node->op());
-  // if (!torch_mlir_node->getPythonStacktrace().empty()) {
-  //   op->node()->s_(
-  //       c10::Symbol::attr("source"), torch_mlir_node->getPythonStacktrace());
-  // }
+  auto torch_mlir_node =
+      NodeCast<TorchMlirNode>(output.node, output.node->op());
+
+  std::vector<std::string> source_files, functions;
+  std::vector<int64_t> line_numbers;
+  const auto& frames = torch_mlir_node->metadata().frame_info;
+  if (!frames.empty()) {
+    static std::vector<std::string> g_roots =
+      string_split(sys_util::GetEnvString("LTC_IR_DEBUG_ROOT_PATH", ""), ":");
+
+    std::for_each(frames.rbegin(), frames.rend(),
+      [&](const torch::lazy::SourceLocation& location) {
+        functions.push_back(location.function);
+        line_numbers.push_back(location.line);
+
+        std::string file_name = location.file;
+        for (const std::string& root : g_roots) {
+          if (startswith(file_name, root)) {
+            // location.file starts with root, strip it off
+            file_name = file_name.substr(root.size());
+            break;
+          }
+        }
+        source_files.push_back(file_name);
+    });
+
+    if (!source_files.empty()) {
+      op->node()->ss_(
+          c10::Symbol::attr("source_files"), source_files);
+      op->node()->ss_(
+          c10::Symbol::attr("functions"), functions);
+      op->node()->is_(
+          c10::Symbol::attr("line_numbers"), line_numbers);
+    }
+  }
   emitted_outputs_[output] = std::move(op);
 }
 
@@ -424,7 +453,11 @@ const std::string TorchMlirComputation::to_string() const {
     *ss_ptr << std::string(part.data, part.length);
   };
   std::stringstream ss;
-  mlirOperationPrint(mlirModuleGetOperation(module_op_), print_callback, &ss);
+
+  // Setup flags for MLIR serialization.
+  MlirOpPrintingFlags flags = mlirOpPrintingFlagsCreate();
+  mlirOpPrintingFlagsEnableDebugInfo(flags, FLAGS_torch_lazy_ir_debug, false);
+  mlirOperationPrintWithFlags(mlirModuleGetOperation(module_op_), flags, print_callback, &ss);
   return ss.str();
 }
 
--- a/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
+++ b/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
@@ -11,6 +11,8 @@
 #include "function_importer.h"
 #include "ivalue_importer.h"
 
+#include <c10/util/irange.h>
+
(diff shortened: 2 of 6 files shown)


python/torch_mlir/csrc/base_lazy_backend/mlir_lowering_context.cpp
```diff
@@ -424,7 +453,12 @@ const std::string TorchMlirComputation::to_string() const {
     *ss_ptr << std::string(part.data, part.length);
   };
   std::stringstream ss;
-  mlirOperationPrint(mlirModuleGetOperation(module_op_), print_callback, &ss);
+
+  // Setup flags for MLIR serialization.
+  MlirOpPrintingFlags flags = mlirOpPrintingFlagsCreate();
+  mlirOpPrintingFlagsEnableDebugInfo(flags, FLAGS_torch_lazy_ir_debug, false);
+  mlirOpPrintingFlagsUseLocalScope(flags);
+  mlirOperationPrintWithFlags(mlirModuleGetOperation(module_op_), flags, print_callback, &ss);
```
mabrowning, line 461: I don't think we should upstream this part. Printing using local scope by default is a huge cost.

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
```diff
@@ -403,8 +405,27 @@ MlirAttribute torch_mlir::importAttribute(MlirLocation loc,
...
+        context,
+        toMlirStringRef(functions[i]),
+        mlirLocationFileLineColGet(
+          context,
+          toMlirStringRef(source_files[i]),
+          line_numbers[i],
+          0 /* column is not available */
+        )
+      );
+      loc = (i == 0 ? new_loc : mlirLocationCallSiteGet(new_loc, loc));
```
ramiro050, line 425: Can you explain what is happening here? (Adding a comment would be useful). Is the idea that multiple `source_files` represents a call stack, so you have to chain the call site locations?

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
```diff
@@ -403,8 +405,27 @@ MlirAttribute torch_mlir::importAttribute(MlirLocation loc,
 
 MlirLocation torch_mlir::getMlirLocationFromNode(MlirContext context,
                                                  torch::jit::Node *node) {
-  auto flc = node->sourceRange().file_line_col();
-  if (flc) {
+  if (node->hasAttribute(c10::Symbol::attr("source_files"))) {
+    const auto& source_files = node->ss(c10::Symbol::attr("source_files"));
+    const auto& line_numbers = node->is(c10::Symbol::attr("line_numbers"));
+    const auto& functions = node->ss(c10::Symbol::attr("functions"));
```
ramiro050, line 411: nit: in torch-mlir variable names should be camelCase

ramiro050 (commented):
I'm not the most familiar with the changes made here. However, nothing really stood out to me as strange. @silvasean can you also review this?

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
GlebKazantaev: Yeah, I'll add some comment here.

> Is the idea that multiple source_files represents a call stack, so you have to chain the call site locations?

Right, LTC gives us full stack trace per operation, so we need to chain it to get a sequence of `callsite` like that:
```
loc(callsite("forward"("pytorch/torch/nn/modules/linear.py":114:0) at callsite("_call_impl"("pytorch/torch/nn/modules/module.py":1488:0) at callsite("forward"("./torch-mlir/examples/ltc_backend_mnist.py":38:0) at callsite("_call_impl"("pytorch/torch/nn/modules/module.py":1488:0) at callsite("main"("./torch-mlir/examples/ltc_backend_mnist.py":56:0) at "<module>"("./torch-mlir/examples/ltc_backend_mnist.py":105:0)))))))
```

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
```diff
@@ -407,15 +409,52 @@ MlirAttribute torch_mlir::importAttribute(MlirLocation loc,
...
-  if (flc) {
+  MlirLocation loc = mlirLocationUnknownGet(context);
+
+  auto getLocOpName = [&](const c10::FunctionSchema * schema)->MlirLocation {
+    return mlirLocationNameGet(context,
+                               toMlirStringRef(schema->operator_name().name),
+                               /*childLoc=*/{nullptr});
+  };
+
+  if (node->hasAttribute(c10::Symbol::attr("source_files"))) {
```
mabrowning, line 420: So, the current logic is:
 1. Try to get the full stack trace. If found, the CallSite is the only loc and the op name is skipped.
 2. Else, try to get the single flie/line/col, and add a `FusedLoc[operator_name, FileLineCol]`
 3. Otherwise, just use the op name.
Finally, always name the location with the scope if it exists.

That's slightly inconsistent and a creative interpretation of `FusedLoc`. Also, it might involve a `NameLoc(NameLoc)` if no source code location is present which is technically legal in newer MLIR versions, but a bit strange and also won't be parsable by older MLIR versions I happen to know you're interested in.

Instead, I'd suggest a 2 stage approach, with the idea that if we have name scopes, the simpler op name location may not provide any additional info:

First:
1. Try to get `loc = CallsiteLoc(...)`
2. Else try get just `loc = FileLineColLoc(...)`
3. Else, `loc = UnknownLoc`

Next, add the relevant "name", with the source code loc as the child:
1. Try to get the scope, and add `loc = Name(scope, loc)`
2. Else, fall back on the op name and add `loc = Name(operator_name, loc)`

Alternatively, we can do what pytorch_xla does and combine them into one "name", something like:
```
name = operator_name
if scope:
   name = scope + "/" + name
```

test/python/importer/jit_ir/node_import/debug-info.py
```diff
@@ -19,9 +19,14 @@
 def add3(t0, t1, t2):
   # TODO: Checks for debug info are quite hard with the new trailing debug
   # attribute print. See if this can be improved.
```
mabrowning, line 21: I know this predates your change, but "this can be improved" with `mb.module.operation.print(enable_debug_info=True, use_local_scope=True)`
```python
import mlir.ir
with mlir.ir.Context() as c:
    c.allow_unregistered_dialects = True
    m = mlir.ir.Module.parse("\"test.op\"() {} : () -> () loc(\"loc\")")
    m.operation.print(enable_debug_info=True)
    print("// -----------------")
    m.operation.print(enable_debug_info=True, use_local_scope=True)
```
```mlir 
module  {
  "test.op"() : () -> () loc(#loc1)
} loc(#loc0)
#loc0 = loc("-":0:0)
#loc1 = loc("loc")
// -----------------
module  {
  "test.op"() : () -> () loc("loc")
} loc("-":0:0)
```

mabrowning (commented):
Thanks! I think how we combine the 3 pieces of information either needs options controlling it or a bit more cohesion.

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
GlebKazantaev: > Try to get the scope, and add loc = Name(scope, loc)
Else, fall back on the op name and add loc = Name(operator_name, loc)

That looks reasonable! Just pushed a commit to match new logic. Please take a look.

test/python/importer/jit_ir/node_import/debug-info.py
GlebKazantaev: Thanks! Test was updated.

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp
GlebKazantaev: CC: @ramiro050

mabrowning: So we actually just discovered a use case where it would be helpful to have the oeprator name _and_ the other location info (scope + stack trace), to handle performance profiilng per-op-basis. 

If the user has pytorch code like
```python
a = torch.add(b,c).mul(d,e)
```

Then both ops will have the same location.

If we go with the `{scope}/{op_name}` route, they'll be distinct.

ramiro050: Thanks @mabrowning for the help on this!

ramiro050:
Removing myself from reviewers and letting @mabrowning continue the review.

GlebKazantaev:
> Removing myself from reviewers and letting @mabrowning continue the review.

Could you please add @mabrowning as a collaborator and grant merge rights?

antoniojkim (approved):

mabrowning (approved):

ramiro050:
> Could you please add @mabrowning as a collaborator and grant merge rights?

Done! @mabrowning, you should receive an email to join.

PR #1923, opened 2023-03-08
[TCP] Introduce tcp.concat operation & aten.cat -> tcp.concat conversion

asaadaldien:
As titled.


externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -320,4 +320,25 @@ def Tcp_SqrtOp : Tcp_UnaryElementwiseOp<"sqrt", [SameOperandsAndResultElementTyp
   let assemblyFormat = "$in attr-dict `:` type($in) `->` type($out)";
 }
 
+def Tcp_ConcatOp : Tcp_Op<"concat", [SameOperandsAndResultElementType]> {
+  let summary = "Concatenation of input tensors";
```
sanjoy, line 324: This should be written as an action: "Concatenates two tensors."

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -320,4 +320,25 @@ def Tcp_SqrtOp : Tcp_UnaryElementwiseOp<"sqrt", [SameOperandsAndResultElementTyp
...
+
+  let description = [{
+    Concatenates tensors of same datatype and rank into a single tensor on specific single axis.
```
sanjoy, line 327: Maybe also add the precondition that the dims on the other axes must match.  (Undefined behavior if they don't match.)

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -111,4 +111,52 @@ LogicalResult IsolatedGroupOp::verify() {
...
+  // the same.
+  for (auto type : getInputs().getTypes()) {
+    auto inputTensorType = type.cast<TensorType>();
+    if (inputTensorType.getElementType() != outputTensorType.getElementType())
+      return emitOpError() << "failed to verify tcp.concat with input and "
+                              "output operands rank mismatch";
+    if (outputTensorType.getRank() != inputTensorType.getRank())
+      return emitOpError() << "failed to verify tcp.concat with input and "
+                              "output operands rank mismatch";
+    for (int64_t dim = 0; dim < inputTensorType.getRank(); ++dim) {
```
sanjoy, line 130: Nit: I think these messages should be "failed to verify that ... rank matches".

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -111,4 +111,52 @@ LogicalResult IsolatedGroupOp::verify() {
...
+      if (dim == axis) {
+        concatDimAcc += inputTensorType.getShape()[dim];
+      } else {
+        if (inputTensorType.getShape()[dim] != outputTensorType.getShape()[dim])
```
sanjoy, line 134: What if an input is dynamic but other inputs are known?  I.e. we're concatenating `<4x3>` and `<5x?>` along axis 0 to get `<9x3>`?  Is that allowed?

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -111,4 +111,52 @@ LogicalResult IsolatedGroupOp::verify() {
 
 OpFoldResult ConstOp::fold(FoldAdaptor) { return getValueAttr(); }
 
+LogicalResult ConcatOp::verify() {
+  auto outputTensorType = getResult().getType().cast<TensorType>();
+  int64_t axis = getAxis();
+  int64_t concatDimVal = outputTensorType.getShape()[axis];
+  // Accumulate concat dim, ignored if any of the inputs is dynamic.
+  int64_t concatDimAcc = 0;
+  // All tensors must have the smae dtype, rank and all non-concat dims must be
```
sanjoy, line 120: Nit s/smae/same/

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -111,4 +111,52 @@ LogicalResult IsolatedGroupOp::verify() {
...
+                              "output operands rank mismatch";
+    for (int64_t dim = 0; dim < inputTensorType.getRank(); ++dim) {
+      if (dim == axis) {
+        concatDimAcc += inputTensorType.getShape()[dim];
+      } else {
+        if (inputTensorType.getShape()[dim] != outputTensorType.getShape()[dim])
+          emitOpError() << "failed to verify tcp.concat with non concat dim "
+                        << dim << ", having different values "
+                        << inputTensorType.getShape()[dim] << " "
+                        << outputTensorType.getShape()[dim];
```
sanjoy, line 138: Will this error message be legible if one of this is unknown?

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -111,4 +111,52 @@ LogicalResult IsolatedGroupOp::verify() {
...
+    if (!llvm::any_of(getInputs().getTypes(), [axis](Type type) {
+          return ShapedType::isDynamic(
+              type.cast<TensorType>().getShape()[axis]);
+        }))
+      return emitOpError() << "failed to verify tcp.concat with dynamic concat "
+                              "axis and static inputs";
+    else
+      return success();
+  }
+
```
sanjoy, line 154: Very very minor nit (so optional to fix): IMO it will be cleaner to track this in `concatDimAcc` itself.  I.e. above instead of `concatDimAcc += inputTensorType.getShape()[dim];` do

```
if (ShapedType::isDynamic(inputTensorType.getShape()[dim]) || concatDimAcc == ShapedType::kDynamicSize)
  concatDimAcc = ShapedType::kDynamicSize;
else
  concatDimAcc += inputTensorType.getShape()[dim]
```

Then you just need to check `concatDimAcc != concatDimVal` here.

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -111,4 +111,52 @@ LogicalResult IsolatedGroupOp::verify() {
 
 OpFoldResult ConstOp::fold(FoldAdaptor) { return getValueAttr(); }
 
+LogicalResult ConcatOp::verify() {
+  auto outputTensorType = getResult().getType().cast<TensorType>();
+  int64_t axis = getAxis();
+  int64_t concatDimVal = outputTensorType.getShape()[axis];
+  // Accumulate concat dim, ignored if any of the inputs is dynamic.
```
sanjoy, line 118: Nit: if any of the inputs are dynamic

sanjoy (changes requested):

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -111,4 +111,52 @@ LogicalResult IsolatedGroupOp::verify() {
...
+  int64_t concatDimAcc = 0;
+  // All tensors must have the smae dtype, rank and all non-concat dims must be
+  // the same.
+  for (auto type : getInputs().getTypes()) {
+    auto inputTensorType = type.cast<TensorType>();
+    if (inputTensorType.getElementType() != outputTensorType.getElementType())
+      return emitOpError() << "failed to verify tcp.concat with input and "
+                              "output operands rank mismatch";
```
navahgar, line 126: This check is for same element type on inputs and output. The error message is for rank mismatch.

Actually, I think this check is not necessary here, since the op is implementing the trait `SameOperandsAndResultElementType`, right?

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -111,4 +111,52 @@ LogicalResult IsolatedGroupOp::verify() {
...
+    if (outputTensorType.getRank() != inputTensorType.getRank())
+      return emitOpError() << "failed to verify tcp.concat with input and "
+                              "output operands rank mismatch";
```
navahgar, line 129: Can we replace this check with the `AllRanksMatch` trait? https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/IR/OpBase.td#L2451

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -111,4 +111,52 @@ LogicalResult IsolatedGroupOp::verify() {
...
+        if (inputTensorType.getShape()[dim] != outputTensorType.getShape()[dim])
+          emitOpError() << "failed to verify tcp.concat with non concat dim "
+                        << dim << ", having different values "
+                        << inputTensorType.getShape()[dim] << " "
+                        << outputTensorType.getShape()[dim];
+      }
+    }
+  }
+
+  // If concat dim is dynamic, at least one input must be dynamic.
```
navahgar, line 143: This check is assuming that shape propagation has already happened before converting to TCP. Is that always guaranteed? 

I understand Torch does shape propagation. But what if there are other frontend dialects that convert to TCP but do not perform shape propagation completely?

In the case, would we end up with this scenario where the inputs are static and the output is dynamic (because its shape hasn't be propagated fully)?

navahgar (changes requested):
Thanks for doing this @asaadaldien. Some comments inline.

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
asaadaldien: If we want to support type inference at the TCP level, we have to not use `?` or dynamic shape  to carry the semantics of both `dynamic` & `unknown` dim. I can see many issues including doing that.

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -101,6 +102,37 @@ class ConvertValueTensorLiteralOp
...
+public:
+  using OpConversionPattern<AtenCatOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(AtenCatOp catOp, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    SmallVector<Value> inputs;
+    if (!getListConstructElements(adaptor.getTensors(), inputs))
+      return rewriter.notifyMatchFailure(
+          catOp, "aten.cat operands be a list of tensors");
```
navahgar, line 115: typo: "must" be a list of tensors

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -101,6 +102,37 @@ class ConvertValueTensorLiteralOp
...
+    RankedTensorType resultType = getTypeConverter()
+                                      ->convertType(catOp.getType())
+                                      .cast<RankedTensorType>();
+
+    rewriter.replaceOpWithNewOp<tcp::ConcatOp>(
+        catOp, resultType, tensorInputs,
+        rewriter.getIntegerAttr(rewriter.getI64Type(), dim));
+    return success();
+  }
+}; // namespace
```
navahgar, line 134: Incorrect comment `// namespace` here.

test/Conversion/TorchToTcp/data_movment.mlir
```diff
@@ -0,0 +1,13 @@
+// RUN: torch-mlir-opt <%s -convert-torch-to-tcp -split-input-file -verify-diagnostics | FileCheck %s
```
navahgar, line 1: typo on the file name: "data_movement.mlir"

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -101,6 +102,37 @@ class ConvertValueTensorLiteralOp
   }
 };
 
+class ConvertAtenCatOp : public OpConversionPattern<AtenCatOp> {
```
navahgar, line 105: It might be better to create a new file for the lowerings of all data-movement ops, and put this there. That would also match with what you have already done with the unit test.

navahgar (commented):
Mostly LGTM. A few more minor comments.

navahgar (commented):
LGTM. I don't have any other comments.

I will let @sanjoy approve this if he doesn't have any other comments.

navahgar (approved):

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sanjoy: But `::verify` will check this only if the shapes are static.  My point is that we should document that the op has undefined behavior if the dims are dynamic (`?`) and don't match at runtime.

sanjoy (approved):
Minor documentation nit, otherwise LGTM.

navahgar (approved):

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
sanjoy: Yes and you're right that it isn't documented for these other binary ops either.  I think it should be documented in `Tcp_BinaryElementwiseOp`.

So feel free to skip this comment for now, I'll try fix this documentation issue in a later PR.

sanjoy (approved):

Issue #1926, opened 2023-03-09
Unable to Lower to Backend Contract List Unsqueeze


cetiniz:
Below is a minimally reproducible example:
```python
def stack_decomposition(tensors: List[torch.Tensor], dim: int=0) -> torch.Tensor:
    combined = torch.unsqueeze(tensors[0], dim=dim)
    for i in range(1, len(tensors)):
        combined = torch.cat([torch.unsqueeze(tensors[i], dim=dim),combined], dim=dim)
    return combined
```

The issue is inside the `torch.tensor_static_info_cast` operator (I believe). Inside of the comparison for folding, it will check if there is a valid subtype:
https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/IR/TorchOps.cpp#L1531

Inside the check in this case the subtype for the operand is a ValueTensorType which performs the following check:
https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/IR/TorchTypes.cpp#L78

It will compare the subtype with a ValueTensorType without dtype or shape information. This will fail because the ValueTensorType being compared has dtype information at this point but does not have shape information.

```
%567 = torch.tensor_static_info_cast %566 : !torch.vtensor<[206,5],f32> to !torch.vtensor<*,f32>
```

PR #1929, opened 2023-03-10
Expand definition of tensor subtype to include shape/dtype info

ramiro050:
Currently, the op `torch.tensor_static_info_cast` will not get canonicalized away if the result type has any shape or dtype information. This is because `isValidSubtype` only returns true when the tensor types being compared are exactly the same or the supertype has no shape and dtype information. Being unable to canonicalize away the `torch.tensor_static_info_cast` gets in the way of further optimizations, such as shape propagation.

This commit improves `isValidSubtype` by adding logic that compares the shapes and dtypes of the two tensor types to determine of one type is indeed a valid subtype of the other.

Fixes https://github.com/llvm/torch-mlir/issues/1926

### DIFF
--- a/lib/Dialect/Torch/IR/TorchTypes.cpp
+++ b/lib/Dialect/Torch/IR/TorchTypes.cpp
@@ -68,16 +68,32 @@ bool Torch::isValidSubtype(Type subtype, Type type) {
     return true;
   }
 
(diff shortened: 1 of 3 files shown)


li-plus (approved):
LGTM!

PR #1931, opened 2023-03-11
[e2e] fix stack e2e test typo

qingyunqu:


ramiro050 (approved):
Good catch! 😄 Did not notice this was wrong when I reviewed the patch.

PR #1933, opened 2023-03-13
Constant pad nd to tosa

lisaliu1:
This PR contains the implementation of lowering torch.aten.constant_pad_nd to tosa.


lisaliu1:
@ramiro050 @silvasean could you please take a look?

test/Conversion/TorchToTosa/constant_pad_nd.mlir
```diff
@@ -0,0 +1,37 @@
+// RUN: torch-mlir-opt <%s -convert-torch-to-tosa -split-input-file -verify-diagnostics | FileCheck %s
```
eric-k256, line 1: The preference in torch-mlir is to not add unit test cases for straightforward expansions like this, we rely on the e2e testing. You have enabled e2e testing above, so you don't need this file.
`DON'T use a unit test if your lowering pattern could be described as a trivial "macro expansion" of one op into another op or set of ops. That is, if you feel like your unit test is just rewriting b.create<...>(...) into CHECK: ... then it is probably not a useful unit test.`

eric-k256 (changes requested):
The code itself looks good. All that is needed is to remove the unit tests based on the project policy.

test/Conversion/TorchToTosa/constant_pad_nd.mlir
lisaliu1: @eric-k256 Thanks for your review. I've removed the unit test test/Conversion/TorchToTosa/constant_pad_nd.mlir.

PR #1934, opened 2023-03-13
build: update llvm tag to b884f4ef

ramiro050:
This commit updates submodules to:
llvm: b884f4ef0a2de3d0f24111411dff663fd68c2eb0
mlir-hlo: d89b2ade0cc7b42c6da05c4ed92dfa3241c4f697

This commit also:

- replaces the use of `mergeBlockBefore` with `inlineBlockBefore` (see
https://reviews.llvm.org/D144969)
- swaps order of passes `tosa-to-tensor` and `tosa-to-linalg` in
`tosa` backend for e2e tests (see https://reviews.llvm.org/D145119)

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -409,7 +409,7 @@ static void replaceOpWithRegion(PatternRewriter &rewriter, Operation *op,
   Block *block = &region.front();
   Operation *terminator = block->getTerminator();
   ValueRange results = terminator->getOperands();
-  rewriter.mergeBlockBefore(block, op, blockArgs);
+  rewriter.inlineBlockBefore(block, op, blockArgs);
   rewriter.replaceOp(op, results);
   rewriter.eraseOp(terminator);
 }
--- a/lib/Dialect/Torch/Transforms/DropAbstractInterpCalculations.cpp
+++ b/lib/Dialect/Torch/Transforms/DropAbstractInterpCalculations.cpp
@@ -29,7 +29,7 @@ public:
     Block *block = &op.getBody().front();
     Operation *terminator = block->getTerminator();
     ValueRange results = terminator->getOperands();
-    rewriter.mergeBlockBefore(block, op);
+    rewriter.inlineBlockBefore(block, op);
     rewriter.replaceOp(op, results);
     rewriter.eraseOp(terminator);
     return success();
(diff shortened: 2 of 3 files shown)


ashay (approved):

vivekkhandelwal1 (approved):
LGTM!

eric-k256 (approved):
LGTM

ramiro050:
There was a change made in `TosaToLinalg` that breaks some of the tosa e2e tests. I've made a patch fixing things https://reviews.llvm.org/D145986, but this means we will not be able to bump llvm until the fix lands. Maybe we can use a different green commit this week after the fix lands, @ashay?

ashay:
Thanks for the upstream fix!  Once your patch lands, could you drop a line here?  I'll then send another LLVM and MHLO green commit that includes your change.

ramiro050:
> Thanks for the upstream fix! Once your patch lands, could you drop a line here? I'll then send another LLVM and MHLO green commit that includes your change.

@ashay, the fix has landed: https://github.com/llvm/llvm-project/commit/2d61628c1f49963921b9ac1995218191dc5e3091

ashay:
Thanks @ramiro050!  Here are the new green commits from today.

Green LLVM commit: `d99d765fc07e8338ee2b54337dbf4a3dbc8c310a`
Green MHLO commit: `8a80eb6b313c4f7aa392eee82d4168f10c1ef215` (branch `greencommit/2023-03-15-d99d765f`)

ramiro050:
@ashay, I'm getting the following error with the new green commit for mlir-hlo:

```
CMake Error at /usr/local/google/home/ramiroleal/torch-mlir/externals/llvm-project/mlir/cmake/modules/AddMLIR.cmake:298 (message):
  MhloRngUtils specifies LINK_LIBS LLVMSupport, but LINK_LIBS cannot be used
  for LLVM libraries.  Please use LINK_COMPONENTS instead.
Call Stack (most recent call first):
  /usr/local/google/home/ramiroleal/torch-mlir/externals/mlir-hlo/mhlo/utils/CMakeLists.txt:21 (add_mlir_library)
```

Have you seen something like this?

ashay:
Sorry that took a while to get to.  I've updated the MHLO green commit (`6fc921d7ca04fb1508e9bdb64fab08a173a7cf66`) to include the fix.  Let me know if you see other errors after picking that hash.

ramiro050:
There was another change on the tosa side that breaks two of the e2e tests. I've made a fix for it: https://reviews.llvm.org/D146174. Let's see how many llvm commits we have to go through before we land on a good one 😆

ramiro050:
@ashay, fix landed https://github.com/llvm/llvm-project/commit/fbf719b86846fa35b6dd0e658e107afc9973fc86

ashay:
Hey @ramiro050, it looks like the creating an MHLO green commit will require some non-trivial changes.  Unfortunately, I have a few other things I need to finish at work today, so how about we defer this to Monday, when I need to push the new green commits?

ramiro050:
> Hey @ramiro050, it looks like the creating an MHLO green commit will require some non-trivial changes. Unfortunately, I have a few other things I need to finish at work today, so how about we defer this to Monday, when I need to push the new green commits?

Yeah, no worries. Thanks for all the help with this! I can close this PR and let the next person next week take over.

PR #1935, opened 2023-03-13
Merge main into dtype-functions-staging

ramiro050:


ramiro050:
The only changes needed, apart from fixing some minor merge conflicts, where adding dtype functions for `aten.ge.Tensor` and `aten.le.Tensor`, and marking `RandnLikeModule_basic`. I figured out why to `Randn` tests are crashing, but I will fix this first in `main` then in a future update we can merge the fix into this branch.

I'll try to make merge updates more frequently to avoid such crazy PRs.

li-plus (approved):
There are too many changes for a careful review. I'll approve this as long as we can fix the crashing cases later.
Is it a better option if we rebase dtype staging branch onto the main branch, instead of merging?

ramiro050:
> There are too many changes for a careful review. 

Yeah, I agree. If it helps, you can see what things look like relative to main: https://github.com/llvm/torch-mlir/compare/main...ramiro050:torch-mlir:merge-main. 

> I'll approve this as long as we can fix the crashing cases later.

Yeah, I'll fix the crashing cases. Although the CI failure seems to be due to something else. I'm still debugging this.

> Is it a better option if we rebase dtype staging branch onto the main branch, instead of merging?

Rebasing would make things much simpler, but it makes it harder for contributors because they have to delete their local branch and get the branch again. Moreover, rebasing would not go through the PR process, which would prevent me from running the CI.

ramiro050:
Figured out the CI issue. The `Matmul_dot` test was crashing because of a bug in `RefineTypes`. @li-plus, if it's fine with you, I will go ahead and merge this and fix the issues in `main`.

li-plus:
LGTM! Please go ahead. BTW how about we finishing and merging dtype functions to main as soon as possible, so we won't need to sync from main again.

ramiro050:
> BTW how about we finishing and merging dtype functions to main as soon as possible, so we won't need to sync from main again.

Sure!

PR #1939, opened 2023-03-15
[MLIR][TORCH] Add E2E support for cross entropy lowering

vivekkhandelwal1:


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -903,6 +903,11 @@ def aten〇mse_loss〡shape(self: List[int], target: List[int], reduction: int =
         return upstream_shape_functions.unary(self)
     return []
 
+def aten〇cross_entropy_loss〡shape(self: List[int], target: List[int], weight: Optional[List[int]] = None, reduction: int = 1, ignore_index: int = -100, label_smoothing: float = 0.) -> List[int]:
+    if reduction == 0:
+        return [self[0]]
```
ramiro050, line 908: This does not seem intuitive to me. I would expect no reduction to return the full tensor. Can you upstream this to make sure it's properly tested?

ramiro050 (changes requested):
Thanks, I just have one change request

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
vivekkhandelwal1: This is similar to this: https://github.com/pytorch/pytorch/blob/master/torch/jit/_shape_functions.py#L999-L1003

ramiro050: Some shape functions had not been updated to use the upstream one. I've updated the library: https://github.com/llvm/torch-mlir/pull/1952. There is only one instance remaining that performs a similar calculation (`mse_loss`) and this one should also be upstreamed. Upstreaming is something that is expected out of every new shape function to ensure proper testing: https://github.com/llvm/torch-mlir/pull/1952

> New shape functions should be added upstream

The testing from PyTorch gives me more confidence that the shape function is correct. For example, in this particular case, it is unclear to me why this shape function would be different from `nll_loss_forward` that you linked given that they have a similar signature. Moreover, the shape function in this PR was recently modified from `if reduction == 0` to a `if reduction == 0 and len(self) == 2:` without a change in test outcomes, so it's hard for me to know which is correct.

ramiro050 (approved):
LGTM!

PR #1944, opened 2023-03-15
RefineTypes aten.sum: check dtype exists before accessing methods

ramiro050:
This commit adds a check that `defaultDtype` exists in the RefineTypes handling of `AtenSumOp` before accessing the method `isInteger`, which crashes the program is `defaultDtype` is null.

The handling of `defaultDtype` is the same as the one used for the `AtenSumDimIntListOp`.


vivekkhandelwal1 (approved):
LGTM!

PR #1946, opened 2023-03-16
[e2e_test] fix typo in pow_int e2e test

qingyunqu:
There is a different behavior between torch and torch-mlir:
img_v2_5ade5ba8-5b2c-4668-857c-b31555e9e0cg
  
It's a bug or feature of pytorch?

### DIFF
--- a/python/torch_mlir_e2e_test/test_suite/basic.py
+++ b/python/torch_mlir_e2e_test/test_suite/basic.py
@@ -2868,7 +2868,7 @@ def ScalarImplicitIntModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
-class PowIntFloat(torch.nn.Module):
+class PowIntFloatModule(torch.nn.Module):
 
     def __init__(self):
         super().__init__()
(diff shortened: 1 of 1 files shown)


ramiro050 (approved):
Good catch! Ideally we should have a way to catch these types of errors when running the test suite. I have opened up an issue to track this: https://github.com/llvm/torch-mlir/issues/1948

ramiro050:
> It's a bug or feature of pytorch?

Wow, that's unexpected. This definitely looks like a bug to me. Below are the valid schemas according to PyTorch:

```
aten::pow.Tensor_Tensor(Tensor self, Tensor exponent) -> Tensor
aten::pow.Tensor_Scalar(Tensor self, Scalar exponent) -> Tensor
aten::pow.Scalar(Scalar self, Tensor exponent) -> Tensor
aten::pow.Scalar_out(Scalar self, Tensor exponent, *, Tensor(a!) out) -> Tensor(a!)
aten::pow.Tensor_Scalar_out(Tensor self, Scalar exponent, *, Tensor(a!) out) -> Tensor(a!)
aten::pow.Tensor_Tensor_out(Tensor self, Tensor exponent, *, Tensor(a!) out) -> Tensor(a!)
aten::pow.int(int a, int b) -> float
aten::pow.complex(complex a, complex b) -> complex
aten::pow.float(float a, float b) -> float
aten::pow.int_float(int a, float b) -> float
aten::pow.float_int(float a, int b) -> float
aten::pow.float_complex(float a, complex b) -> complex
aten::pow.complex_float(complex a, float b) -> complex
aten::pow.Scalar_Scalar(Scalar a, Scalar b) -> float
aten::pow.int_to_int(int a, int b) -> int
```

The only ones that should result in a complex value are the ones that take at least one complex value as an input. Can you open an issue in PyTorch about this?

PR #1947, opened 2023-03-16
LowerToBackendContract: Explicitly error out on unimplemented operator

mgehre-amd:
As discussed on discord [0], this PR will emit more direct errors like `unsupported by backend contract: Unimplemented operator 'aten.clamp.Tensor'` instead of diagnosing unknown ranks or missing value conversion.

[0] https://discordapp.com/channels/636084430946959380/742573221882364009/1083826291301679225

### DIFF
--- a/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
+++ b/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
@@ -197,6 +197,17 @@ static bool satisfiesBackendContract(ModuleOp module,
   if (walkResult0.wasInterrupted())
     return false;
 
+  // Check for unimplemented operators first to give more direct diagnostics.
+  walkResult0 = module.walk([&](Torch::OperatorOp op) {
+    if (actuallyEmitDiagnostics) {
+      op->emitError("unsupported by backend contract: Unimplemented operator '"
+        + op.getName() + "'");
+    }
+    return WalkResult::interrupt();
+  });
+  if (walkResult0.wasInterrupted())
+    return false;
+
   // Check all the types of all Value's in the program and the legality of all
   // the ops.
   //
--- a/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
+++ b/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
@@ -199,6 +199,13 @@ static bool satisfiesBackendContract(ModuleOp module,
 
   // Check for unimplemented operators first to give more direct diagnostics.
   walkResult0 = module.walk([&](Torch::OperatorOp op) {
+    if (llvm::all_of(op.getResults(), [&op](auto res) {
+          return succeeded(
+              checkType(op.getOperation(), res.getType(), /*actuallyEmitDiagnostics=*/false));
+        })) {
+      return WalkResult::advance();
+    }
+
     if (actuallyEmitDiagnostics) {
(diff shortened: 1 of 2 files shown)


mgehre-amd:
I'm a bit surprised by the CI failure. Turn out that 
`./tools/e2e_test.sh -f NormalizeModule_basic -c lazy_tensor_core` fails on this PR with `error: unsupported by backend contract: Unimplemented operator 'aten.clamp_min.Tensor'` but it passes without my PR.

I don't quite understand how it could be passing when the operator is not implemented?

ramiro050:
That's very strange. Let me take a look and see what I find

ramiro050:
> I don't quite understand how it could be passing before when the operator is not implemented?

I think I know what's happening. The LTC reference backend in Torch-MLIR creates an MLIR graph that is expected to satisfy the backend contract, but then executes things using the TorchScript LTC backend. In other words, the MLIR graph is not compiled further down and executed.

https://github.com/llvm/torch-mlir/blob/3f3833d7700cd976bcf929e2a175f2af10490baf/python/torch_mlir/csrc/reference_lazy_backend/backend_impl.cpp#L102-L103

This is why the tests were passing even though there is no ODS for `aten.clamp_max.Tensor`.

In reality, the `torch.operator` is not the erroneous part. We expect to use `torch.operator`s to represent [custom ops](https://github.com/llvm/torch-mlir/issues/1462) in the future. The cases we want to error out on are when `torch.operator`s don't satisfy the guarantees of the backend contract: value semantics, dtype, and shapes for all tensors.

In the case of LTC, the clamp tests create the op

```
%4 = torch.operator "aten.clamp_min.Tensor"(%3, %arg0) : (!torch.vtensor<[3,1],f32>, !torch.vtensor<[],f64>) -> !torch.vtensor<[3,1],f32>
```

which does satisfy the backend contract. The check should be modified to use the `checkType` helper to make sure the outputs of `torch.operator` satisfy the backend contract.

mgehre-amd:
> > I don't quite understand how it could be passing before when the operator is not implemented?
> 
> I think I know what's happening. The LTC reference backend in Torch-MLIR creates an MLIR graph that is expected to satisfy the backend contract, but then executes things using the TorchScript LTC backend. In other words, the MLIR graph is not compiled further down and executed.
> 
> https://github.com/llvm/torch-mlir/blob/3f3833d7700cd976bcf929e2a175f2af10490baf/python/torch_mlir/csrc/reference_lazy_backend/backend_impl.cpp#L102-L103
> 
> This is why the tests were passing even though there is no ODS for `aten.clamp_max.Tensor`.
> 
> In reality, the `torch.operator` is not the erroneous part. We expect to use `torch.operator`s to represent [custom ops](https://github.com/llvm/torch-mlir/issues/1462) in the future. The cases we want to error out on are when `torch.operator`s don't satisfy the guarantees of the backend contract: value semantics, dtype, and shapes for all tensors.
> 
> In the case of LTC, the clamp tests create the op
> 
> ```
> %4 = torch.operator "aten.clamp_min.Tensor"(%3, %arg0) : (!torch.vtensor<[3,1],f32>, !torch.vtensor<[],f64>) -> !torch.vtensor<[3,1],f32>
> ```
> 
> which does satisfy the backend contract. The check should be modified to use the `checkType` helper to make sure the outputs of `torch.operator` satisfy the backend contract.

Thanks, that's it. I updated the PR and now it passes :-)

PR #1949, opened 2023-03-16
[TCP] Check and notifyMatchFailure for integer cases of `AtenDivTensorOp` -> TCP

sjain-stanford:
As titled.

Also fixes `s/llvm::makeArrayRef/llvm::ArrayRef/` per deprecation notice.


navahgar (approved):
LGTM

PR #1950, opened 2023-03-16
Add `RecomposeComplexOps` declaration + fix typos in pass name

ramiro050:
The `RecomposeComplexOps` pass currently does not have a TableGen declaration and it is using the base class of `DecomposeComplexOps`, which causes `--mlir-print-ir-after-all` to create wrong pass labels. This commit fixes that as well as some minor typos in the name of the pass.


ramiro050:
Not sure how I missed this when I reviewed your patch, @gpetters94

ramiro050:
@gpetters94, PTAL

AmosLewis (approved):
LGTM

PR #1952, opened 2023-03-17
Use upstream shape functions when available

ramiro050:
There are several ops that have their shape function upstream and had not been updated in Torch-MLIR to use the upstream version. This commit updates those shape function. In addition, TODOs have been added for shape functions that should be upstream but are not.


vivekkhandelwal1 (approved):
LGTM!

PR #1954, opened 2023-03-19
[TCP] Lower torch integer constants -> tcp.const

asaadaldien:
As titled

### DIFF
--- a/lib/Conversion/TorchToTcp/Misc.cpp
+++ b/lib/Conversion/TorchToTcp/Misc.cpp
@@ -90,10 +90,14 @@ public:
             ->convertType(op.getType())
             .cast<RankedTensorType>();
 
-    // TODO: Implement integer constant lowering to tcp::ConstOp
-    if (auto elements = op.getValueAttr().dyn_cast<DenseIntElementsAttr>())
-      return rewriter.notifyMatchFailure(
-          op, "Integer constants lowering to TCP is unimplemented");
+    if (auto elements = op.getValueAttr().dyn_cast<DenseIntElementsAttr>()) {
+      Type elementType = resultType.getElementType();
+      auto denseIntAttr = elements.mapValues(elementType, [&](const APInt &v) {
+        return APInt(elementType.getIntOrFloatBitWidth(), v.getSExtValue());
+      });
+      rewriter.replaceOpWithNewOp<tcp::ConstOp>(op, resultType, denseIntAttr);
+      return success();
+    }
 
     rewriter.replaceOpWithNewOp<tcp::ConstOp>(op, resultType,
                                               adaptor.getValue());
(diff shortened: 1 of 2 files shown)


navahgar (approved):
LGTM

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -90,10 +90,14 @@ class ConvertValueTensorLiteralOp
...
-    // TODO: Implement integer constant lowering to tcp::ConstOp
-    if (auto elements = op.getValueAttr().dyn_cast<DenseIntElementsAttr>())
-      return rewriter.notifyMatchFailure(
-          op, "Integer constants lowering to TCP is unimplemented");
+    if (auto elements = op.getValueAttr().dyn_cast<DenseIntElementsAttr>()) {
+      Type elementType = resultType.getElementType();
+      auto denseIntAttr = elements.mapValues(elementType, [&](const APInt &v) {
+        return APInt(elementType.getIntOrFloatBitWidth(), v.getSExtValue());
+      });
+      rewriter.replaceOpWithNewOp<tcp::ConstOp>(op, resultType, denseIntAttr);
```
sjain-stanford, line 98: I'm assuming this handles both signed and unsigned integer cases? If so, could you add a lit test for `ui32` as well?

asaadaldien: Only `ui8` is the legal unsigned integer type via torch dialect.

silvasean:
Hey folks, now that development has started picking up here, the inboxes of Torch-MLIR developers are getting quite a few TCP-related emails. In the end what is the plan for the final home of TCP? Is it going to become an LLVM incubator? What is the timeline? Or do you at least know how to turn off emails/notifications on the branch?

sjain-stanford:
Hey Sean. Thanks for bringing this up. Let us discuss internally and get back to you on this by end of next week?

sjain-stanford:
Regarding the email spam, apologies for the same - I couldn't find a way to disable notifications on GH on a per-branch basis. One option is to use email filters on `[TCP]` in the PR title - would that work for you for the time-being, while we discuss the longer term solution?

sjain-stanford:
Hi @silvasean. After discussing with the team, we're going to make a decision (on the home for TCP) by the end of Q2, likely an incubator in `llvm-project` or `openxla`. We'll actually make the move in Q3/Q4. Does that timeframe work?
cc: @sanjoy, @sjarus, @navahgar, @stellaraccident

silvasean:
Thanks for looking into this! Yes, that timeline seems fine.

sjain-stanford:
Hi @silvasean! To close the loop on this, and as promised above, we have found a new home for TCP (https://github.com/cruise-automation/mlir-tcp) and are in the last few stages of completing the integrations to our internal monorepo. We have already paused contributions to the `mlir-tcp` branch of Torch-MLIR, and we should be good to archive this branch once the integrations are complete (hopefully in a week or two - I'll let you know). Thanks for your support all along.

cc: @stellaraccident , @ramiro050 , @zezhang , @navahgar , @sanjoy

PR #1957, opened 2023-03-20
Max pool2d ceil mode to tosa

lisaliu1:
This PR implements the support for lowering torch.aten.max_pool2d to tosa with ceil_mode==true.

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -3794,23 +3794,14 @@ public:
   static int64_t getOutputDim(int64_t inputDim, int64_t kernelDim,
                               int64_t stride, int64_t padBefore,
                               int64_t padAfter, int64_t dilation,
-                              bool ceilMode=false) {
+                              bool ceilMode = false) {
     if (inputDim == kUnknownSize) {
       return kUnknownSize;
     } else {
       int64_t dimSize = inputDim + padBefore + padAfter - dilation * (kernelDim - 1) - 1;
-      if (!ceilMode) {
-        dimSize = (dimSize / stride + 1);
-      }
-      else {
-        if (dimSize % stride == 0) {
-          dimSize = (dimSize / stride + 1);
-        }
-        else {
-          dimSize = (dimSize / stride + 2);
-        }
-      }
-      return dimSize;
+      if (ceilMode && (dimSize % stride != 0))
+        return dimSize / stride + 2;
+      return dimSize / stride + 1;
     }
   }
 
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -4030,23 +4047,22 @@ static LogicalResult getOutputTypeAndPoolingParameters(
...
+  
   bool ceilMode;
   if (!matchPattern(op.getCeilMode(), m_TorchConstantBool(&ceilMode)))
     return rewriter.notifyMatchFailure(
         op, "only support constant bool ceil_mode for pooling op");
-  if (ceilMode)
-    return rewriter.notifyMatchFailure(
-        op, "only support ceil_mode equals to False for pooling op");
 
+  // add ceil_mode support.
```
eric-k256, line 4059: nit: this comment only makes sense for this change, I would delete it.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3977,17 +3987,24 @@ template <typename AtenOpT, typename tosaOp>
...
 
   int64_t outputHDim = ConvertAtenPoolingBaseOp<AtenOpT, tosaOp>::getOutputDim(
       inputShape[inputRank - 2], kernelSize[0], strideArray[0], padArray[0],
-      padArray[0], dilationArray[0]);
+      padArray[0], dilationArray[0], ceilMode);
   int64_t outputWDim = ConvertAtenPoolingBaseOp<AtenOpT, tosaOp>::getOutputDim(
       inputShape[inputRank - 1], kernelSize[1], strideArray[1], padArray[1],
-      padArray[1], dilationArray[1]);
+      padArray[1], dilationArray[1], ceilMode);
+  padArray[0] = (outputHDim - 1) * strideArray[0] +
```
eric-k256, line 4002: Code formatting looks a bit inconsistent here (sometimes a*b and others a * b). Can you verify you've run clang-format? `git clang-format HEAD~1` should only modify lines you have already changed.
https://github.com/llvm/torch-mlir/wiki/Coding-Style

eric-k256 (commented):
Your design looks right, just two minor points to check.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
lisaliu1: Thanks Eric for your reviews. I've addressed your comments.

eric-k256 (approved):
Thanks Lisa! Looks good to me.

PR #1958, opened 2023-03-20
[TCP] unary ops support for floor and ceil operations

zezhang:
- Add the tcp op definitions for floor and ceil operations
- Add the torch->tcp and tcp->linarg conversion for floor and ceil operations.
- Updated test cases to include floor and ceil operations

To test:

`./build/bin/llvm-lit externals/llvm-external-projects/torch-mlir-dialects/test/Conversion/TcpToLinalg/ -v`
`./build/bin/llvm-lit externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/ -v`
`./build/bin/llvm-lit test/Conversion/TorchToTcp/ -v`
`python -m e2e_testing.main --config=tcp -v -s`

### DIFF
--- a/lib/Conversion/TorchToTcp/Elementwise.cpp
+++ b/lib/Conversion/TorchToTcp/Elementwise.cpp
@@ -304,6 +304,48 @@ public:
   }
 };
 
+class ConvertAtenCeilOp : public OpConversionPattern<AtenCeilOp> {
+public:
+  using OpConversionPattern<AtenCeilOp>::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(AtenCeilOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Value input = adaptor.getSelf();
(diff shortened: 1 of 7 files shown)


sjain-stanford (approved):
LGTM, thanks. If not done already, please run clang-format on the modified files.

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -304,6 +304,48 @@ class ConvertAtenSqrtOp : public OpConversionPattern<AtenSqrtOp> {
   }
 };
 
+class ConvertAtenCeilOp : public OpConversionPattern<AtenCeilOp> {
```
navahgar, line 307: These conversions are very similar to `tcp::SqrtOp` and `tcp::TanhOp`. We should refactor all of these into a single function that is templatized on the aten and tcp ops. So, we don't have to duplicate the code.

zezhang: I can create a followup PR to make a templated function for this.

PR #1959, opened 2023-03-20
RFC: Support pushing custom ops through backend-contract using `torch.operator`

makslevental:
# Pitch

I think lots of people want to be able to push opaque kernels through backend contract (https://github.com/llvm/torch-mlir/issues/1519, https://github.com/llvm/torch-mlir/pull/1947, https://github.com/llvm/torch-mlir/pull/1514). Indeed, there is also a similarly themed [proposal](https://github.com/llvm/torch-mlir/pull/1563) that goes through `backend-legal-ops` (and is specific to TOSA). I believe that there's also a need for supporting **custom ops**, i.e., ops that are effectively placeholders for some possible lowering/implementation on the otherside of backend contract (posterchild here is quantized ops).

# Approach

The approach here is an adaptation of the other approach - we go through `backend-legal-ops` using `torch.operator`. The catch is that because `OperandOp` doesn't possess the `HasValueSemantics` trait, `ReduceOpVariants` and `MaximizeValueSemantics` will stumble and ultimately backend contract won't be satisfied. Thus, we add `HasValueSemantics` to `torch.operator`. In addition, we extend `wrapWithCalculateOpIfLibraryFunctionAvailable` to support user-provided shape and dtype functions; in particular we provide these two (trivial) refinement functions:

```mlir
func.func @__torch_mlir_shape_fn.operator.goofy.identity(%arg0: !torch.list<int>) -> !torch.list<int> {
  return %arg0 : !torch.list<int>
}
func.func @__torch_mlir_dtype_fn.operator.goofy.identity(%arg0: !torch.int, %arg1: !torch.int) -> !torch.int {
  return %arg1 : !torch.int
}
```

 With the example/demo here I am able fully lower to

```
module attributes {torch.debug_module_name = "CustomOpExampleModule"} {
  func.func @forward(%arg0: !torch.vtensor<[3,4],f32>) -> !torch.vtensor<[3,4],f32> {
    %int2 = torch.constant.int 2
    %0 = torch.aten.mul.Scalar %arg0, %int2 : !torch.vtensor<[3,4],f32>, !torch.int -> !torch.vtensor<[3,4],f32>
    %1 = torch.operator "goofy.identity"(%0) {has_value_semantics = true} : (!torch.vtensor<[3,4],f32>) -> !torch.vtensor<[3,4],f32>
    return %1 : !torch.vtensor<[3,4],f32>
  }
}
```

Certainly there are other ways to do this, so I'm open to suggestions/advice.

cc @powderluv @qedawkins @AmosLewis

### DIFF
--- a/examples/custom_op_demo.py
+++ b/examples/custom_op_demo.py
@@ -1,0 +1,95 @@
+import torch
+import torch.utils.cpp_extension
+import torch_mlir
+from torch_mlir import run_pipeline_with_repro_report
+from torch_mlir.ir import BoolAttr, Context, Module, InsertionPoint, Location
+from torch_mlir_e2e_test.annotations import export, annotate_args
+
+
+def identity(_5: torch.Tensor):
+    return _5
+
+
+goofy_lib = torch.library.Library("goofy", "DEF")
+goofy_lib.define("identity(Tensor t) -> Tensor")
+goofy_lib.impl("identity", identity)
+
+
+class CustomOpExampleModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args(
+        [
+            None,
+            ([-1, -1], torch.float32, True),
+        ]
+    )
+    def forward(self, a):
+        b = 2 * a
+        return torch.ops.goofy.identity(b)
+
+
+mod = CustomOpExampleModule()
+mod.eval()
+
+module = torch_mlir.compile(mod, torch.ones(3, 4), output_type="raw")
+
+pipeline = (
+    "symbol-dce,"
+    "torch-prepare-for-globalize-object-graph,"
+    "torch-globalize-object-graph,"
+    "symbol-dce,"
+    "inline{default-pipeline= max-iterations=4 },"
+    "torch-adjust-calling-conventions"
+)
+
+run_pipeline_with_repro_report(
+    module, pipeline=f"builtin.module({pipeline})", description=""
+)
+print(module)
+
+forward = module.operation.regions[0].blocks[0].operations[1]
+goofy_op = forward.operation.regions[0].blocks[0].operations[4]
+goofy_op.attributes["has_value_semantics"] = BoolAttr.get(True, context=module.context)
+
+print(module)
+
+abstract_interp_src = """\
+func.func @__torch_mlir_shape_fn.operator.goofy.identity(%arg0: !torch.list<int>) -> !torch.list<int> {
+  return %arg0 : !torch.list<int>
+}
+func.func @__torch_mlir_dtype_fn.operator.goofy.identity(%arg0: !torch.int, %arg1: !torch.int) -> !torch.int {
+  return %arg1 : !torch.int
+}
+"""
+
+with Location.unknown(module.context) as loc:
+    abstract_interp_module = Module.parse(abstract_interp_src)
+    with InsertionPoint.at_block_begin(module.body) as ip:
+        shape_fn = abstract_interp_module.body.operations[0]
+        dtype_fn = abstract_interp_module.body.operations[1]
+        InsertionPoint.insert(ip, shape_fn.detach_from_parent())
+        InsertionPoint.insert(ip, dtype_fn.detach_from_parent())
+
+print(module)
+
+run_pipeline_with_repro_report(
+    module,
+    pipeline="builtin.module(func.func(torch-reduce-op-variants,torch-maximize-value-semantics))",
+    description="",
+)
+
+print(module)
+
+run_pipeline_with_repro_report(
+    module,
+    pipeline="builtin.module(torch-lower-to-backend-contract{backend-legal-ops=torch.operator decompose=true max-iterations=10})",
+    description="",
+)
+
+shape_fn.detach_from_parent()
+dtype_fn.detach_from_parent()
+
+print(module)
--- a/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
+++ b/lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
@@ -191,7 +191,10 @@ static bool isValidNonContainerResultType(Type resultType) {
          resultType.isa<Torch::FloatType>() ||
          resultType.isa<Torch::IntType>() ||
          resultType.isa<Torch::BoolType>() ||
-         resultType.isa<Torch::NoneType>();
+         resultType.isa<Torch::NoneType>() ||
+         (resultType.isa<Torch::ListType>() && cast<Torch::ListType>(resultType)
+                                                   .getContainedType()
+                                                   .isa<Torch::IntType>());
 }
 
 static LogicalResult validateReturns(func::FuncOp func) {
--- a/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
+++ b/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
@@ -52,6 +52,14 @@ static Type getContainerOrTensorTypeWithValueSemantics(Type type) {
   }
 }
 
+static bool operatorOpHasValueSemantics(OperatorOp opOp) {
+  if (!opOp->hasAttr("has_value_semantics"))
+    return false;
+  auto hasValueSemantics =
+      opOp->getAttr("has_value_semantics").cast<BoolAttr>().getValue();
+  return hasValueSemantics;
+}
+
 namespace {
 // Convert value semantic ops operating on mutable arrays to instead operate on
 // immutable tensors.
@@ -61,8 +69,13 @@ public:
       : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {}
   LogicalResult matchAndRewrite(Operation *op,
                                 PatternRewriter &rewriter) const override {
-    if (!op->hasTrait<Torch::OpTrait::HasValueSemantics>())
+    if (isa<OperatorOp>(op)) {
+      if (!operatorOpHasValueSemantics(cast<OperatorOp>(op))) {
+        return rewriter.notifyMatchFailure(op, "does not have value semantics");
+      }
+    } else if (!op->hasTrait<Torch::OpTrait::HasValueSemantics>()) {
       return rewriter.notifyMatchFailure(op, "does not have value semantics");
+    }
 
     rewriter.startRootUpdate(op);
     // Convert all operands.
@@ -254,7 +267,9 @@ class ReduceOpVariantsPass : public ReduceOpVariantsBase<ReduceOpVariantsPass> {
     target.addIllegalOp<NonValueTensorLiteralOp>();
     target.addIllegalOp<AtenBernoulli_FloatOp>();
     target.markUnknownOpDynamicallyLegal([](Operation *op) {
-      if (op->hasTrait<Torch::OpTrait::HasValueSemantics>()) {
+      if (op->hasTrait<Torch::OpTrait::HasValueSemantics>() ||
+          (isa<OperatorOp>(op) &&
+           operatorOpHasValueSemantics(cast<OperatorOp>(op)))) {
         auto hasValueSemantics = [](Type t) {
           // TODO: Make this an allowlist based on a closed torch dialect
           // type system.
--- a/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
+++ b/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
@@ -52,14 +52,6 @@ static Type getContainerOrTensorTypeWithValueSemantics(Type type) {
   }
 }
 
-static bool operatorOpHasValueSemantics(OperatorOp opOp) {
-  if (!opOp->hasAttr("has_value_semantics"))
-    return false;
-  auto hasValueSemantics =
-      opOp->getAttr("has_value_semantics").cast<BoolAttr>().getValue();
-  return hasValueSemantics;
-}
-
 namespace {
 // Convert value semantic ops operating on mutable arrays to instead operate on
 // immutable tensors.
@@ -69,13 +61,8 @@ public:
       : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {}
   LogicalResult matchAndRewrite(Operation *op,
                                 PatternRewriter &rewriter) const override {
-    if (isa<OperatorOp>(op)) {
-      if (!operatorOpHasValueSemantics(cast<OperatorOp>(op))) {
-        return rewriter.notifyMatchFailure(op, "does not have value semantics");
-      }
-    } else if (!op->hasTrait<Torch::OpTrait::HasValueSemantics>()) {
+    if (!op->hasTrait<Torch::OpTrait::HasValueSemantics>())
       return rewriter.notifyMatchFailure(op, "does not have value semantics");
-    }
 
     rewriter.startRootUpdate(op);
     // Convert all operands.
@@ -267,9 +254,7 @@ class ReduceOpVariantsPass : public ReduceOpVariantsBase<ReduceOpVariantsPass> {
     target.addIllegalOp<NonValueTensorLiteralOp>();
     target.addIllegalOp<AtenBernoulli_FloatOp>();
     target.markUnknownOpDynamicallyLegal([](Operation *op) {
-      if (op->hasTrait<Torch::OpTrait::HasValueSemantics>() ||
-          (isa<OperatorOp>(op) &&
-           operatorOpHasValueSemantics(cast<OperatorOp>(op)))) {
+      if (op->hasTrait<Torch::OpTrait::HasValueSemantics>()) {
         auto hasValueSemantics = [](Type t) {
           // TODO: Make this an allowlist based on a closed torch dialect
           // type system.
--- a/lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp
+++ b/lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp
@@ -66,7 +66,16 @@ LogicalResult Torch::wrapWithCalculateOpIfLibraryFunctionAvailable(
         libFuncArgsBuilder) {
   Location loc = op->getLoc();
   MLIRContext *context = op->getContext();
-  auto name = op->getName().stripDialect();
+
+  std::string name_;
+  if (isa<OperatorOp>(op)) {
+    auto opOp = cast<OperatorOp>(op);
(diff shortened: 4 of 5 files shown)


examples/custom_op_demo.py
```diff
@@ -0,0 +1,95 @@
...
+from torch_mlir_e2e_test.annotations import export, annotate_args
+
+
+def identity(_5: torch.Tensor):
+    return _5
+
+
+goofy_lib = torch.library.Library("goofy", "DEF")
+goofy_lib.define("identity(Tensor t) -> Tensor")
+goofy_lib.impl("identity", identity)
```
makslevental, line 15: Same thing as the "classical" torch custom op registration.

lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp
```diff
@@ -76,9 +87,17 @@ LogicalResult Torch::wrapWithCalculateOpIfLibraryFunctionAvailable(
...
-  libFuncNamesUsed.push_back(libFuncName);
+  if (!libFunc) {
+    auto parentModule = op->getParentOfType<ModuleOp>();
+    if (parentModule)
+      libFunc =
+          op->getParentOfType<ModuleOp>().lookupSymbol<func::FuncOp>(libFuncName);
+    if (!libFunc)
+      return success();
+  } else {
+    libFuncNamesUsed.push_back(libFuncName);
```
makslevental, line 98: `libFuncNamesUsed` is ultimately used to "import" shape functions - since the user-provided shape functions are already in the user module, this isn't necessary (and causes a segfault somewhere around [ReifyAbstractInterpCalculationsUtils.cpp#L159](https://github.com/llvm/torch-mlir/blob/cc5e2c230c576eca7210a4e7f475e1e9bcd965b3/lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp#L159).

lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp
```diff
@@ -66,7 +68,16 @@ LogicalResult Torch::wrapWithCalculateOpIfLibraryFunctionAvailable(
         libFuncArgsBuilder) {
   Location loc = op->getLoc();
   MLIRContext *context = op->getContext();
-  auto name = op->getName().stripDialect();
+
+  std::string name_;
+  if (isa<OperatorOp>(op)) {
+    auto opOp = cast<OperatorOp>(op);
+    auto opName = opOp->getAttr("name").cast<StringAttr>().getValue();
+    name_ = "operator." + opName.str();
```
makslevental, line 76: shape/dtype functions for operator ops should be further namespaced.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -191,7 +191,10 @@ static bool isValidNonContainerResultType(Type resultType) {
          resultType.isa<Torch::FloatType>() ||
          resultType.isa<Torch::IntType>() ||
          resultType.isa<Torch::BoolType>() ||
-         resultType.isa<Torch::NoneType>();
+         resultType.isa<Torch::NoneType>() ||
+         (resultType.isa<Torch::ListType>() && cast<Torch::ListType>(resultType)
+                                                   .getContainedType()
+                                                   .isa<Torch::IntType>());
```
makslevental, line 197: shape functions return `list<int>`.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -61,8 +69,13 @@ class ConvertHasValueSemanticsOpsToValueTensors : public RewritePattern {
       : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {}
   LogicalResult matchAndRewrite(Operation *op,
                                 PatternRewriter &rewriter) const override {
-    if (!op->hasTrait<Torch::OpTrait::HasValueSemantics>())
+    if (isa<OperatorOp>(op)) {
+      if (!operatorOpHasValueSemantics(cast<OperatorOp>(op))) {
```
makslevental, line 73: you need a nesting here because you want the outer branch to "succeed" for appropriate instances - if `isa<OperatorOp>(op)` and `operatorOpHasValueSemantics(cast<OperatorOp>(op))` then succeed.

powderluv:
This is potentially a way to support LLaMA 4bit quant via Torch-mlir.

makslevental:
> We already have a plan for this (#1462) and it is in progress. The next step is to migrate to the dtype functions (see #1807). Any help on that would be appreciated.

Okay but the approach proposed here works today with minimal expansion of the API surface. Is it possible to merge this in before the other roadmap is complete and then remove it afterwards?

silvasean:
> > We already have a plan for this (#1462) and it is in progress. The next step is to migrate to the dtype functions (see #1807). Any help on that would be appreciated.
> 
> Okay but the approach proposed here works today with minimal expansion of the API surface. Is it possible to merge this in before the other roadmap is complete and then remove it afterwards?

No, I don't think that's a wise engineering decision. Removing workarounds is usually 10x the work of adding them. It is better to push forward and complete the plan as originally specified. I think we will probably end up with something kind of similar to this in some aspects, but until we have migrated to the dtype functions there isn't much point since we won't be able to compile real models (see explanation in https://github.com/llvm/torch-mlir/issues/1807 for why it happens on a branch). An equally big problem is how this feature exposed to users -- the approach in this patch "happens to work", but does not rely on a supported API surface area.

makslevental:
> > > We already have a plan for this (#1462) and it is in progress. The next step is to migrate to the dtype functions (see #1807). Any help on that would be appreciated.
> > 
> > 
> > Okay but the approach proposed here works today with minimal expansion of the API surface. Is it possible to merge this in before the other roadmap is complete and then remove it afterwards?
> 
> No, I don't think that's a wise engineering decision. Removing workarounds is usually 10x the work of adding them. It is better to push forward and complete the plan as originally specified. I think we will probably end up with something kind of similar to this in some aspects, but until we have migrated to the dtype functions there isn't much point since we won't be able to compile real models (see explanation in #1807 for why it happens on a branch). An equally big problem is how this feature exposed to users -- the approach in this patch "happens to work", but does not rely on a supported API surface area.

And how about with the change I just pushed?

silvasean:
> And how about with the change I just pushed?

It doesn't fundamentally address any of the issues I mentioned.

makslevental:
> It doesn't fundamentally address any of the issues I mentioned.

It does; 

1. from your proposal

     >From a design perspective, I don't see any option other than to require all custom ops be value-semantic. This seems like it will be enough for the use cases that the community has presented so far.
    
     so you will need to mark `torch.operator` as `HasValueSemantics` inevitably as well.
2. >An equally big problem is how this feature exposed to users -- the approach in this patch "happens to work", but does not rely on a supported API surface area.

    With the explicit trait, it becomes explicit API surface.
3. >we won't be able to compile real models 

    I can compile, today, real models, by lowering with an opaque `torch.operator "quantized.matmul"` and then replacing it with a custom implementation (in terms of `linalg`) on the otherside of the backend contract.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -191,7 +191,10 @@ static bool isValidNonContainerResultType(Type resultType) {
          resultType.isa<Torch::FloatType>() ||
          resultType.isa<Torch::IntType>() ||
          resultType.isa<Torch::BoolType>() ||
-         resultType.isa<Torch::NoneType>();
+         resultType.isa<Torch::NoneType>() ||
+         (resultType.isa<Torch::ListType>() && cast<Torch::ListType>(resultType)
```
ramiro050, line 195: We shouldn't be doing this. This function was created with the goal of preventing something like a `ListType` return to reach the backend contract. This would lead to invalid IR being generated.

makslevental:
@ramiro050 

> We will likely require all custom ops to be value-semantics. However, this does not mean we want all `torch.operator`s generated to have value semantics. This would generate invalid IR in cases where the op in the `torch.operator` is an actual torch op that does not have value semantics, which could lead to errors that are very difficult to debug. What we should do is have a method of identifying ops that have a shape and dtype function in the abstract interpretation library, and mark those ops as having value semantics.

If you check the first commit in this stack, this literally exactly what I had

```cpp
static bool operatorOpHasValueSemantics(OperatorOp opOp) {
  if (!opOp->hasAttr("has_value_semantics"))
    return false;
  auto hasValueSemantics =
      opOp->getAttr("has_value_semantics").cast<BoolAttr>().getValue();
  return hasValueSemantics;
}
```

and then special case checks in `ConvertHasValueSemanticsOpsToValueTensors`

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
makslevental: That's fine but if you want to enable user-provided shape and dtype functions in the same parent module then there needs to be special casing for them. The alternative, is to provide some mechanism for passing handles to a `ModuleOp` all the way down into `wrapWithCalculateOpIfLibraryFunctionAvailable`.

powderluv:
Today torch-mlir lacks the ability to lower state of art quantized models (llama 4bit, stable diffusion 8bit, and other custom models).

IIUC this PR allows for a pragmatic but short term way to support 4bit llama and 8bit SD today for our customers. It is forward progress. We can continue to invest in bringing the best approach and can get dedicated resources to help with the right approach but we do have to ship something and can't wait O(months) since customers will just bypass torch-mlir.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
ramiro050: User provided shape and dtype functions will be handled exactly the same way that current shape and dtype functions are handled. The plan is to load them from a `.mlir` file when the shape or dtype pass is taking place, then remove them once the pass is done. This is all outlined in the RFC: https://github.com/llvm/torch-mlir/issues/1462. There shouldn't be any coupling between the shape and dtype pipelines and the rest of the passes in torch-mlir.

ramiro050:
> @ramiro050
> 
> > We will likely require all custom ops to be value-semantics. However, this does not mean we want all `torch.operator`s generated to have value semantics. This would generate invalid IR in cases where the op in the `torch.operator` is an actual torch op that does not have value semantics, which could lead to errors that are very difficult to debug. What we should do is have a method of identifying ops that have a shape and dtype function in the abstract interpretation library, and mark those ops as having value semantics.
> 
> If you check the first commit in this stack, this literally exactly what I had
> 
> ```c++
> static bool operatorOpHasValueSemantics(OperatorOp opOp) {
>   if (!opOp->hasAttr("has_value_semantics"))
>     return false;
>   auto hasValueSemantics =
>       opOp->getAttr("has_value_semantics").cast<BoolAttr>().getValue();
>   return hasValueSemantics;
> }
> ```
> 
> and then special case checks in `ConvertHasValueSemanticsOpsToValueTensors`. I removed this because Sean believes "the approach in this [commit] 'happens to work', but does not rely on a supported API surface area."
> 
> I'm happy to develop more safety and docs in order to get it to cross the threshold from "happens to work" to "supported API surface area."

Sorry, this isn't what I meant. Users shouldn't have to modify the IR. We should have a very simple pass that checks if the `torch.operator` has a dtype+shape function in the library, and if so, turn the operator into a value semantics one.

makslevental:
> Sorry, this isn't what I meant. Users shouldn't have to modify the IR. We should have a very simple pass that checks if the `torch.operator` has a dtype+shape function in the library, and if so, turn the operator into a value semantics one.

I'm sorry, I'm not trying to be argumentative but I'm not sure why 

```
hasDtypeFn && hasShapeFn => HasValueSemantics
```

is an appropriate contract/model/invariant and isn't "hidden API surface" but

```
def Torch_OperatorOp : Torch_Op<"operator", [
    AllowsTypeRefinement
  ]> {
  let arguments = (ins StrAttr:$name,  
    DefaultValuedOptionalAttr<BoolAttr, "false">:$has_value_semantics,
    Variadic<AnyTorchType>:$operands
  );
}
```

is hidden API surface? Or onerous? Or bad pratice? There are ample examples of just this pattern in `mlir-hlo`; in particular
https://github.com/tensorflow/mlir-hlo/blob/cad6055ad2e9779ba01d318ca2c0a9ce5dc968e7/lhlo/IR/lhlo_ops.td#L642.

ramiro050:
> I'm sorry, I'm not trying to be argumentative but I'm not sure why
> 
> ```
> hasDtypeFn && hasShapeFn => HasValueSemantics
> ```

Not exactly. It's more like

```
hasDtypeFn && hasShapeFn && isTorchOperator => isCustomOp
```

The RFC then makes the proposal of assuming `isCustomOp => hasValueSemantics`.

> can't (conceptually) non-valsem ops support dtype and shape refinement? Like I don't see what type refinement even has to do (conceptually) with value semantics.

Indeed, type refinement on its own has nothing to do with value semantics. Once again, this is a fundamental assumption of the RFC. Since it was posted several months ago, no member of the community has disagreed with this assumption in the design, but if you have concerns, we should definitely discuss them in the RFC.

> And having to implement those functions isn't "hidden API surface" (combined a with a whole pass) but
> 
> ```
> def Torch_OperatorOp : Torch_Op<"operator", [
>     AllowsTypeRefinement
>   ]> {
>   let arguments = (ins StrAttr:$name,  
>     DefaultValuedOptionalAttr<BoolAttr, "false">:$has_value_semantics,
>     Variadic<AnyTorchType>:$operands
>   );
> }
> ```
> 
> is hidden API surface? Or onerous? Or bad practice? There are ample examples of just this pattern in `mlir-hlo`; in particular https://github.com/tensorflow/mlir-hlo/blob/cad6055ad2e9779ba01d318ca2c0a9ce5dc968e7/lhlo/IR/lhlo_ops.td#L642.

Not sure I understand what your point is here. The implementing of shape+dtype functions is the design proposed in the RFC that so far no one has had issues with. It uses an approach people are already familiar with thanks to the shape library. Again, if you have design concerns, do bring them up on the RFC.

makslevental:
>I wanna quickly comment on this since I am one of the people most familiar with this issue and I think it is important for guiding future decisions on this.

Okay it's true that wasn't clear to me so thank you for the clarification but

>If the graph has many custom ops, it will require many iterations of the LowerToBackendContract pipeline. If the number of iterations reaches the iteration limit, this will cause LowerToBackendContract to fail.

This is a perfectly conservative and explicit graceful fail for anyone attempting to use this PR (if it were merged) - users who are willing to pay the price will bump the `maxIterations` until things go through. In addition, `torch.operator` is literally the only op for which this partial implemenation is satisfactory even from the `maxIterations` perspective; no user will ever have more than a few of these so the iterations will look like 

```
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
dtype-functions block // for the lone torch.operator
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
RefineTypes block
...
```

silvasean:
@powderluv what is your specific timeline that you would like to have this done by? If 2-3 weeks is okay with your timeline, then let's work together, prioritize this, and implement it as we already had specced and planned. If this is needed sooner, then I really can't think of a way to move forward with this on the main branch while sticking to our engineering principles and overarching LLVM community practices.

> That sounds like "yeah the community can do all the grunt / mechanical work let Sean / Ramiro handle the architectural work". Besides that is what we have been doing since September 2022 when we created this branch https://github.com/llvm/torch-mlir/tree/custom-op-example for existing custom op users and the RFC was floated for the new way forward. We are six months in and so we see this PR as a pragmatic stopgap. We are happy to hash out what the compile api should look like.

Sorry, I didn't intend it to come off like "grunt work", but there is a separation between the more mechanical changes here and the parts that need a deeper analysis of the architectural implications, and Ramiro and I happen to be the ones with more extensive knowledge here. I am happy to help other folks who want to work on this, but since delivery time is the primary criterion here, having Ramiro and I work on it seems most practical.

> As the largest contributor in terms of commits to torch-mlir, including the grunt work of keeping up to date with RollPyTorch the CI, LLVM rotations we would like to know how we can do better. Besides we don't think this is a short term workaround - it accelerates the custom op support (happy to discuss any api surface changes) see three below.

I think that pinging PR's or RFC's that aren't moving but that are important would be my specific request, and giving expected timelines for things. This PR as-is is definitely a short-term workaround. Undoing these things is really, really difficult. Once you start punching layering/API holes things fall apart very quickly and repairing that is always 10x more work.

> All that said, like I offered in late January for dtype porting we are happy to get someone to help on the mechanical dtype work but we shouldn't conflate it as a mandatory requirement for custom op support as proven by this PR (modulo API refinements).

I agree, but the "modulo API refinements" is probably 2-3 weeks end-to-end time to implement. And as Ramiro said you are likely to hit a significant compilation time issue if we don't port the dtype functions too so it is kind of risky to try to ignore the dtype functions too (could result in 100x+ compilation time blowup for a model like llama with a lot of layers).

powderluv:
2-3 weeks is great. Let's split up the tasks and prioritize them. @gpetters94  is back on Friday and can help with dtype porting.

makslevental:
> And as Ramiro said you are likely to hit a significant compilation time issue if we don't port the dtype functions too so it is kind of risky to try to ignore the dtype functions too (could result in 100x+ compilation time blowup for a model like llama with a lot of layers)

That's only 100x (at worst) blowup in compile from PyTorch to `backend-contract`, not the entire compilation to LLVM target. Let's say it's a 10s compile time (I would hope it doesn't take that long to go from PyTorch to `backend-contract`). Wouldn't it be okay if some people are fine with waiting 16 minutes to lower llama/SD today? And again, it's a completely opt-in slow-down; no one is going to hit this accidentally for custom ops/unimplemented ops because they won't provide shape/dtype functions for their custom `torch.operator` and thus `TorchDtypeRefinementPipeline` will never run at all.

I don't have SD or llama on hand but I timed `EfficientNet_B7`, which has ~1100 ops (491 convs) and it took `~8.043s`. So again, if someone is willing to spend 13 minutes waiting for a large model to compile, what's the harm in enabling them?

silvasean:
Hey folks I've created another issue to shepherd this feature to completion at a level of concrete action items: https://github.com/llvm/torch-mlir/issues/1963

I know some folks are on a deadline and hopefully this will help show remaining work and how on track we are for transparency.

PR #1961, opened 2023-03-21
[e2e] remove COMMON_TORCH_MLIR_LOWERING_XFAILS

qingyunqu:
Since it only represent linalg's XFAILS.


silvasean (commented):
It should be OR'ed with the other sets. It likely is being duplicated inadvertently.

silvasean:
I would suggest changing the patch to OR it with other sets and remove the redundant specification of those tests in the other sets.

PR #1962, opened 2023-03-22
Add more dtype funcs

ramiro050:
This PR adds dtype functions for:

- `aten.bincount`
- "Promote three dtypes"
- "Promote LHS with scalar RHS"
- "Promote 2nd and 3rd Operands"

Each addition is made in its own commit to make review easier.


li-plus (approved):
LGTM!

Issue #1963, opened 2023-03-22
Custom Ops Sprint


silvasean:
This issue is meant to track and help collaboration across the community to land the Custom Ops RFC in the next 2-3 weeks ideally: https://github.com/llvm/torch-mlir/issues/1462. This is a continuation of the discussion in https://github.com/llvm/torch-mlir/pull/1959

The main outstanding work items are the following:
1. [15 engineer-days] Porting the dtype functions (see https://github.com/llvm/torch-mlir/issues/1807). This is parallel to the other efforts.
2. [3 days] Implement splicing an MLIR file into the library: as in `torch-mlir-opt -pass-pipeline='torchscript-module-to-torch-backend-pipeline{extra-library=my_custom_ops.mlir}'` mentioned in the RFC. (unlike https://github.com/llvm/torch-mlir/pull/1959 we won't "namespace" torch.operator specially, since it comes from the same namespace source of truth -- the PyTorch op registry)
3. [2 days] Implement the ability to tag value-semantic ops in the library. This essentially augments the existing code in ReduceOpVariants so it would be good to put there. This would look something like `torch-mlir-opt -pass-pipeline='reduce-op-variants{extra-library=my_custom_ops.mlir}'`. The tentative design is a dummy function `__torch_mlir_has_value_semantics.foo.bar` which, if present in the library, indicates value semantics for the `foo.bar` op.
4. [5 days] implement `torch_mlir.compile` API.

The critical path here is 3+2+5 = 10 working days for the API design work (we could in theory parallelize this a bit more). We will need to port the dtype functions within that time in parallel (this work is parallelizable so we can have two people helping ideally).

The current open action items (this is a living list) are

- [ ] Figure out who is available to help with the dtype functions (1. above)
- [ ] 2. above
- [ ] 3. above
- [ ] 4. above

ramiro050:
Thanks Sean for compiling the tasks. I'm happy to take on 2, 3, and 4. These tasks can be parallelized a little bit (3 can to some degree happen independently from 2, 4, but 2, 4 are quite dependent on one another), so if anyone is interested in taking 3 we can do that too.

For the sake of transparency and to help gauge expectations, my plan is to take these tasks as my highest priority items, meaning that I plan on spending most of my time getting them done. This should allow me to finish the tasks in accordance to the timeline outlined.

powderluv:
[...]

PR #1964, opened 2023-03-22
Implementation for user-provided shape and dtype functions for custom ops

makslevental:
handles 2,3,4 from https://github.com/llvm/torch-mlir/issues/1963

```python
class CustomOpExampleModule(torch.nn.Module):
    def __init__(self):
        super().__init__()

    @export
    @annotate_args(
        [
            None,
        ]
    )
    def forward(self, a):
        b = 2 * a
        return torch.ops.goofy.identity(b)


mod = CustomOpExampleModule()
mod.eval()

abstract_interp_src = """\
func.func @__torch_mlir_shape_fn.goofy.identity(%arg0: !torch.list<int>) -> !torch.list<int> {
  return %arg0 : !torch.list<int>
}
func.func @__torch_mlir_dtype_fn.goofy.identity(%arg0 : !torch.tuple<int, int>) -> !torch.int {
  %0:2 = torch.prim.TupleUnpack %arg0 : !torch.tuple<int, int> -> !torch.int, !torch.int
  return %0#1 : !torch.int
}
func.func @__torch_mlir_has_value_semantics_fn.goofy.identity() { return }
"""
tmp = open("shapes.mlir", "w")
tmp.write(abstract_interp_src)
tmp.flush()
```

passes through

```python
module = torch_mlir.compile(
    mod,
    torch.ones(3, 4),
    output_type="torch",
    backend_legal_ops=["goofy.identity"],
    _completely_unsupported_in_progress_extra_library="/tmp/custom_op_shape_dtype_fn.mlir",
)
```
all the way to

```
module attributes {torch.debug_module_name = "CustomOpExampleModule"} {
  func.func @forward(%arg0: !torch.vtensor<[3,4],f32>) -> !torch.vtensor<[3,4],f32> {
    %int2 = torch.constant.int 2
    %0 = torch.aten.mul.Scalar %arg0, %int2 : !torch.vtensor<[3,4],f32>, !torch.int -> !torch.vtensor<[3,4],f32>
    %1 = torch.operator "goofy.identity"(%0) : (!torch.vtensor<[3,4],f32>) -> !torch.vtensor<[3,4],f32>
    return %1 : !torch.vtensor<[3,4],f32>
  }
}
```

Test is checked in.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/Transforms/Passes.h
+++ b/include/torch-mlir/Dialect/Torch/Transforms/Passes.h
@@ -56,7 +56,12 @@ struct TorchLoweringPipelineOptions
   // to check for a specific set of legal ops to stop its iteration.
   ListOption<std::string> backendLegalOps{
       *this, "backend-legal-ops",
-      llvm::cl::desc("List of ops to be considered legal for the backend.")};
+      llvm::cl::desc("List of ops to be considered legal for the backend, such "
+                     "as 'aten.foo'.")};
+
+  Option<std::string> extraLibrary{
+      *this, "extra-library",
+      llvm::cl::desc("Filename of MLIR module for splicing into the abstract interpretation library.")};
 };
 
 /// Creates a pipeline that lowers the object graph IR that is produced by
@@ -78,10 +83,12 @@ void createTorchSimplificationPipeline(
     OpPassManager &pm, const TorchLoweringPipelineOptions &options);
 
 /// Creates a pipeline that refines shapes of tensor operations in the program.
-void createTorchShapeRefinementPipeline(OpPassManager &pm);
+void createTorchShapeRefinementPipeline(
+    OpPassManager &pm, const TorchLoweringPipelineOptions &options);
 
 /// Creates a pipeline that refines dtype of tensor operations in the program.
-void createTorchDtypeRefinementPipeline(OpPassManager &pm);
+void createTorchDtypeRefinementPipeline(
+    OpPassManager &pm, const TorchLoweringPipelineOptions &options);
 
 std::unique_ptr<OperationPass<ModuleOp>> createAdjustCallingConventionsPass();
 
@@ -89,7 +96,8 @@ std::unique_ptr<OperationPass<func::FuncOp>> createRefineTypesPass();
 
 std::unique_ptr<OperationPass<ModuleOp>> createInlineGlobalSlotsPass();
 
-std::unique_ptr<OperationPass<func::FuncOp>> createReduceOpVariantsPass();
+std::unique_ptr<OperationPass<func::FuncOp>>
+createReduceOpVariantsPass(StringRef extraLibrary);
 
 std::unique_ptr<OperationPass<func::FuncOp>> createMaximizeValueSemanticsPass();
 
@@ -100,14 +108,14 @@ createDecomposeComplexOpsPass(ArrayRef<std::string> legalOps);
 
 std::unique_ptr<OperationPass<func::FuncOp>> createRecomposeComplexOps();
 
-std::unique_ptr<OperationPass<ModuleOp>> createPreprocessShapeLibraryPass();
-
-std::unique_ptr<OperationPass<ModuleOp>> createReifyShapeCalculationsPass();
+std::unique_ptr<OperationPass<ModuleOp>>
+createReifyShapeCalculationsPass(StringRef extraLibrary);
 
 std::unique_ptr<OperationPass<func::FuncOp>>
 createSimplifyShapeCalculationsPass();
 
-std::unique_ptr<OperationPass<ModuleOp>> createReifyDtypeCalculationsPass();
+std::unique_ptr<OperationPass<ModuleOp>>
+createReifyDtypeCalculationsPass(StringRef extraLibrary);
 
 std::unique_ptr<OperationPass<func::FuncOp>>
 createSimplifyDtypeCalculationsPass();
@@ -120,12 +128,15 @@ createEraseModuleInitializerPass();
 
 std::unique_ptr<OperationPass<ModuleOp>>
 createLowerToBackendContractPass(int maxIterations, bool decompose,
-                                 ArrayRef<std::string> backendLegalOps);
+                                 ArrayRef<std::string> backendLegalOps,
+                                 StringRef extraLibrary);
 
 std::unique_ptr<OperationPass<ModuleOp>>
 createVerifyBackendContractNoDecompositionsPass();
 
 StringRef getAbstractInterpLibrary();
+
+static const char kTorchOpPrefix[] = R"(torch.)";
 
 } // namespace Torch
 
--- a/include/torch-mlir/Dialect/Torch/Transforms/Passes.td
+++ b/include/torch-mlir/Dialect/Torch/Transforms/Passes.td
@@ -151,7 +151,13 @@ def InlineGlobalSlots : Pass<"torch-inline-global-slots", "ModuleOp"> {
 
 def ReduceOpVariants : Pass<"torch-reduce-op-variants", "func::FuncOp"> {
   let summary = "Reduces variants of ops to a smaller set of ops.";
-  let constructor = "mlir::torch::Torch::createReduceOpVariantsPass()";
+  let constructor = [{
+    mlir::torch::Torch::createReduceOpVariantsPass(/*extraLibrary=*/"")
+  }];
+  let options = [
+    Option<"extraLibrary", "extra-library", "std::string", /*default=*/"",
+           "MLIR module for verifying custom op value semantics">,
+  ];
   let description = [{
     Replaces ops with other ops to reduce the number of variants that
     need to be handled elsewhere in the code.
@@ -240,7 +246,13 @@ def DecomposeComplexOps : Pass<"torch-decompose-complex-ops", "func::FuncOp"> {
 
 def ReifyShapeCalculations : Pass<"torch-reify-shape-calculations", "ModuleOp"> {
   let summary = "Reify shape calculations.";
-  let constructor = "mlir::torch::Torch::createReifyShapeCalculationsPass()";
+  let constructor = [{
+    mlir::torch::Torch::createReifyShapeCalculationsPass(/*extraLibrary=*/"")
+  }];
+  let options = [
+    Option<"extraLibrary", "extra-library", "std::string", /*default=*/"",
+           "MLIR module for splicing into the shape library">,
+  ];
   let description = [{
   }];
 }
@@ -255,7 +267,13 @@ def SimplifyShapeCalculations : Pass<"torch-simplify-shape-calculations", "func:
 
 def ReifyDtypeCalculations : Pass<"torch-reify-dtype-calculations", "ModuleOp"> {
   let summary = "Reify dtype calculations.";
-  let constructor = "mlir::torch::Torch::createReifyDtypeCalculationsPass()";
+  let constructor = [{
+    mlir::torch::Torch::createReifyDtypeCalculationsPass(/*extraLibrary=*/"")
+  }];
+  let options = [
+    Option<"extraLibrary", "extra-library", "std::string", /*default=*/"",
+           "MLIR module for splicing into the dtype library">,
+  ];
   let description = [{
   }];
 }
@@ -291,7 +309,7 @@ def LowerToBackendContract
   let summary = "Perform simplifications until the backend contract is satisfied.";
   let constructor = [{
     mlir::torch::Torch::createLowerToBackendContractPass(
-      /*maxIterations=*/10, /*decompose=*/true, /*backendLegalOps=*/{})
+      /*maxIterations=*/10, /*decompose=*/true, /*backendLegalOps=*/{}, /*extraLibrary=*/"")
   }];
   let description = [{
     This pass performs the bulk of the lowering of the program's computations
@@ -335,7 +353,9 @@ def LowerToBackendContract
     Option<"decompose", "decompose", "bool", /*default=*/"true",
            "Decompose ops.">,
     ListOption<"backendLegalOps", "backend-legal-ops", "std::string",
-               "List of ops to be considered legal for the backend.">
+               "List of ops to be considered legal for the backend, such as 'aten.foo'.">,
+    Option<"extraLibrary", "extra-library", "std::string", /*default=*/"",
+           "MLIR module for splicing into the abstract interpretation library">,
 
   ];
   // TODO: Debug why this is needed, even though the input program has func.func
--- a/lib/Dialect/Torch/Transforms/Passes.cpp
+++ b/lib/Dialect/Torch/Transforms/Passes.cpp
@@ -25,7 +25,7 @@ void mlir::torch::registerTorchPasses() {
       "torch-simplification-pipeline",
       "Pipeline simplifying computations in the program.",
       mlir::torch::Torch::createTorchSimplificationPipeline);
-  mlir::PassPipelineRegistration<>(
+  mlir::PassPipelineRegistration<Torch::TorchLoweringPipelineOptions>(
       "torch-shape-refinement-pipeline", "Pipeline refining shapes of tensors.",
       mlir::torch::Torch::createTorchShapeRefinementPipeline);
 }
@@ -66,7 +66,8 @@ void mlir::torch::Torch::createTorchFunctionToTorchBackendPipeline(
   // Perform the bulk of lowering to the backend contract.
   // See the pass documentation for more information.
   pm.addPass(createLowerToBackendContractPass(
-      options.maxIterations, options.decompose, options.backendLegalOps));
+      options.maxIterations, options.decompose, options.backendLegalOps,
+      options.extraLibrary));
 }
 
 // A simplification pipeline to establish the invariants of the backend
@@ -108,7 +109,8 @@ void mlir::torch::Torch::createTorchSimplificationPipeline(
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   pm.addNestedPass<func::FuncOp>(createRecomposeComplexOps());
   // Reduce variants of ops to a smaller set of primitives.
-  pm.addNestedPass<func::FuncOp>(createReduceOpVariantsPass());
+  pm.addNestedPass<func::FuncOp>(
+      createReduceOpVariantsPass(options.extraLibrary));
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   // Remove dead global slots.
   pm.addPass(createSymbolDCEPass());
@@ -121,8 +123,8 @@ void mlir::torch::Torch::createTorchSimplificationPipeline(
   // This should be run before RefineTypes (which primarily does dtype
   // inference), because Torch type promotion rules actually depend on the shape
   // of the operand.
-  createTorchShapeRefinementPipeline(pm);
-  createTorchDtypeRefinementPipeline(pm);
+  createTorchShapeRefinementPipeline(pm, options);
+  createTorchDtypeRefinementPipeline(pm, options);
   // Refine types in the program, which mainly means inferring dtypes of ops.
   pm.addNestedPass<func::FuncOp>(Torch::createRefineTypesPass());
   // Propagate to ABI return types the shape/dtype information discovered by
@@ -141,13 +143,15 @@ void mlir::torch::Torch::createTorchSimplificationPipeline(
 
 static void createRefinementPipeline(
     mlir::OpPassManager &pm,
-    llvm::function_ref<std::unique_ptr<mlir::OperationPass<mlir::ModuleOp>>()>
+    llvm::function_ref<
+        std::unique_ptr<mlir::OperationPass<mlir::ModuleOp>>(llvm::StringRef)>
         reifyCalculationsPass,
     llvm::function_ref<
         std::unique_ptr<mlir::OperationPass<mlir::func::FuncOp>>()>
-        simplifyCalculationsPass) {
+        simplifyCalculationsPass,
+    const mlir::torch::Torch::TorchLoweringPipelineOptions &options) {
   // Reify the library functions for each op that is present in the library.
-  pm.addPass(reifyCalculationsPass());
+  pm.addPass(reifyCalculationsPass(options.extraLibrary));
 
   // Inline the library functions to enable analysis and transformation.
   // TODO: Only inline library functions (this will currently inline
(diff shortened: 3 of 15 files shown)


lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -52,17 +53,39 @@ static Type getContainerOrTensorTypeWithValueSemantics(Type type) {
...
 // immutable tensors.
 class ConvertHasValueSemanticsOpsToValueTensors : public RewritePattern {
 public:
-  ConvertHasValueSemanticsOpsToValueTensors(MLIRContext *context)
-      : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {}
+  ConvertHasValueSemanticsOpsToValueTensors(MLIRContext *context,
+                                            StringRef extraLibrary)
+      : RewritePattern(MatchAnyOpTypeTag(), /*benefit=*/1, context) {
+    if (!extraLibrary.empty()) {
+      loadExtraLibrary(extraLibrary.str(), context, this->extraLibrary);
```
makslevental, line 77: I'm parsing both here in the pattern and in the pass because `loadExtraLibrary` writes into a `mlir::OwningOpRef<mlir::ModuleOp>` and i'm guessing i shouldn't be passing pointers to the underlying `ModuleOp` around willy-nilly. but you guys can tell me what's the right/better way to do this.

examples/custom_op_demo.py
```diff
@@ -0,0 +1,81 @@
...
+    tmp.write(abstract_interp_src)
+    tmp.flush()
+
+    run_pipeline_with_repro_report(
+        module,
+        pipeline=f"builtin.module(func.func(torch-reduce-op-variants{{extra-library={tmp.name}}},torch-maximize-value-semantics))",
+        description="",
+    )
+
+    # run_pipeline_with_repro_report(
```
makslevental, line 75: not working yet (segfault)

lib/Dialect/Torch/Transforms/ReifyDtypeCalculations.cpp
```diff
@@ -106,13 +106,24 @@ dtypeFunctionArgsBuilder(OpBuilder &b, Location loc,
...
+  ReifyDtypeCalculationsPass(StringRef extraLibrary) {
+    this->extraLibrary = extraLibrary.str();
+  }
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     ModuleOp module = getOperation();
     OwningOpRef<ModuleOp> library =
         parseSourceString<ModuleOp>(getAbstractInterpLibrary(), context);
+    if (!extraLibrary.empty()) {
+      OwningOpRef<ModuleOp> extraLibraryModule;
```
makslevental, line 121: duplicated in reifyshapecalulcations so it could be factor but it's only a few lines. let me know if you want it factored.

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
```diff
@@ -127,6 +141,28 @@ createVerifyBackendContractNoDecompositionsPass();
...
+  llvm::SourceMgr sourceMgr;
+  sourceMgr.AddNewSourceBuffer(std::move(*fileOrErr), llvm::SMLoc());
+  module = mlir::parseSourceFile<mlir::ModuleOp>(sourceMgr, context);
+  if (!module) {
+    llvm::errs() << "Error can't load file " << extraLibraryPath << "\n";
+    return 3;
+  }
+  return 0;
+}
+
```
makslevental, line 165: this needs to go somewhere else - let me know where.

lib/Dialect/Torch/Transforms/Passes.cpp
```diff
@@ -168,12 +171,36 @@ static void createRefinementPipeline(
...
 
-void mlir::torch::Torch::createTorchDtypeRefinementPipeline(OpPassManager &pm) {
+void mlir::torch::Torch::createTorchDtypeRefinementPipeline(
+    OpPassManager &pm, const TorchLoweringPipelineOptions &options) {
   createRefinementPipeline(pm, Torch::createReifyDtypeCalculationsPass,
-                           Torch::createSimplifyDtypeCalculationsPass);
+                           Torch::createSimplifyDtypeCalculationsPass, options);
 }
+
+int mlir::torch::Torch::loadExtraLibrary(const std::string &extraLibraryPath,
```
makslevental, line 186: this needs to go somewhere else - let me know where.

examples/custom_op_demo.py
```diff
@@ -0,0 +1,81 @@
...
+
+    run_pipeline_with_repro_report(
+        module,
+        pipeline=f"builtin.module(func.func(torch-reduce-op-variants{{extra-library={tmp.name}}},torch-maximize-value-semantics))",
+        description="",
+    )
+
+    run_pipeline_with_repro_report(
+        module,
+        pipeline=f"builtin.module(torch-lower-to-backend-contract{{extra-library={tmp.name} backend-legal-ops=torch.operator decompose=true max-iterations=10}})",
```
makslevental, line 77: right now this fails with

```
error: unsupported by backend contract: Unimplemented operator 'goofy.identity'
note: see current operation: %2 = "torch.operator"(%1) {name = "goofy.identity"} : (!torch.vtensor<[3,4],f32>) -> !torch.vtensor
```

not sure why

test/python/custom_op_demo.py
```diff
@@ -0,0 +1,96 @@
...
+        module,
+        pipeline=f"builtin.module(func.func(torch-reduce-op-variants{{extra-library={tmp.name}}},torch-maximize-value-semantics))",
+        description="",
+    )
+
+    # run_pipeline_with_repro_report(
+    #     module,
+    #     pipeline=f"builtin.module(torch-lower-to-backend-contract{{extra-library={tmp.name} backend-legal-ops=torch.operator decompose=true max-iterations=10}})",
+    #     description="",
+    # )
```
makslevental, line 79: right now this fails with

```
error: unsupported by backend contract: Unimplemented operator 'goofy.identity'
note: see current operation: %2 = "torch.operator"(%1) {name = "goofy.identity"} : (!torch.vtensor<[3,4],f32>) -> !torch.vtensor
```

not sure why

test/python/custom_op_demo.py
```diff
@@ -0,0 +1,96 @@
...
+)
+
+abstract_interp_src = """\
+func.func @__torch_mlir_shape_fn.operator.goofy.identity(%arg0: !torch.list<int>) -> !torch.list<int> {
+  return %arg0 : !torch.list<int>
+}
+func.func @__torch_mlir_dtype_fn.operator.goofy.identity(%arg0: !torch.int, %arg1: !torch.int) -> !torch.int {
+  return %arg1 : !torch.int
+}
+func.func @__torch_mlir_has_value_semantics.goofy.identity() { return }
```
makslevental, line 62: you need either `{ return }` or `func.func private`. whichever you want is fine with me.

test/python/custom_op_shape_dtype_fn.py
```diff
@@ -0,0 +1,71 @@
...
+mod.eval()
+
+abstract_interp_src = """\
+func.func @__torch_mlir_shape_fn.goofy.identity(%arg0: !torch.list<int>) -> !torch.list<int> {
+  return %arg0 : !torch.list<int>
+}
+func.func @__torch_mlir_dtype_fn.goofy.identity(%arg0: !torch.int, %arg1: !torch.int) -> !torch.int {
+  return %arg1 : !torch.int
+}
+func.func @__torch_mlir_has_value_semantics.goofy.identity() { return }
```
makslevental, line 48: you need either `{ return }` or `func.func private`. whichever you want is fine with me.

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
```diff
@@ -100,14 +112,14 @@ createDecomposeComplexOpsPass(ArrayRef<std::string> legalOps);
 
 std::unique_ptr<OperationPass<func::FuncOp>> createRecomposeComplexOps();
 
-std::unique_ptr<OperationPass<ModuleOp>> createPreprocessShapeLibraryPass();
```
makslevental, line 103: dead code

test/python/custom_op_shape_dtype_fn.py
```diff
@@ -0,0 +1,71 @@
...
+"""
+
+with tempfile.NamedTemporaryFile(mode="w") as tmp:
+    tmp.write(abstract_interp_src)
+    tmp.flush()
+    module = torch_mlir.compile(
+        mod,
+        torch.ones(3, 4),
+        output_type="torch",
+        backend_legal_ops=["torch.operator"],
```
silvasean, line 58: We need to change the handling of backend_legal_ops to treat torch.operator specially, since otherwise this just opens up a weird can of worms suddenly allowing all torch.operator's through.

I think we should cahnge it so that the names passed in backend_legal_ops should always be in the sense of the PyTorch operator registry namespace (currently they are prefixed with `torch.`). I think the best path forward here is to change the handling of backend_legal_ops to not need the `torch.` prefix.

That is, `backend_legal_ops=["aten.foo", "goofy.identity"]` would be how we use it and it wouldn't matter if it was an op we know about or a torch.operator. If it is torch.operator then we check the name attribute, and if it is a different `torch` dialect op then we check that the op name is `torch.goofy.identity` etc.

That will require stripping the torch. prefix from a couple places like here: https://github.com/llvm/torch-mlir/blob/f2a05f2dc0dade5cf03f23b317835b36391e0c2d/python/torch_mlir/__init__.py#L243 but I think this puts us in a much better place.

lib/Dialect/Torch/Transforms/ReifyDtypeCalculations.cpp
silvasean: I think the small amount of duplication is fine if you change the signature to loadExtraLibrary as I suggest below.

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
```diff
@@ -12,6 +12,11 @@
 
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/Pass/Pass.h"
+#include "mlir/IR/BuiltinOps.h"
+#include "llvm/Support/ErrorOr.h"
+#include "llvm/Support/MemoryBuffer.h"
+#include "llvm/Support/SourceMgr.h"
```
silvasean, line 18: nit: sort includes.

include/torch-mlir/Dialect/Torch/Transforms/Passes.h
```diff
@@ -57,6 +62,10 @@ struct TorchLoweringPipelineOptions
   ListOption<std::string> backendLegalOps{
       *this, "backend-legal-ops",
       llvm::cl::desc("List of ops to be considered legal for the backend.")};
```
silvasean, line 64: With the change I suggested below, we should update this description to be "List of ops to be considered legal for the backend, such as 'aten.foo'"). (I think the description might be duplicated elsewhere too).

silvasean (changes requested):
Wow, thanks for the quick turnaround on this! I just have a couple suggestions.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -241,20 +268,32 @@ reduceNonValueTensorLiteralOpToValueTensorLiteralOp(NonValueTensorLiteralOp op,
...
+    this->extraLibrary = extraLibrary.str();
+  }
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     RewritePatternSet patterns(context);
-    patterns.add<ConvertHasValueSemanticsOpsToValueTensors>(context);
+    OwningOpRef<ModuleOp> extraLibraryModule;
+    if (!extraLibrary.empty())
+      loadExtraLibrary(extraLibrary, context, extraLibraryModule);
+    patterns.add<ConvertHasValueSemanticsOpsToValueTensors>(context,
```
silvasean, line 283: I suggest creating a SymbolTable for the  extra library module and passing the symbol table to extraLibrary -- that will avoid a linear scan for each lookup.

silvasean (changes requested):

lib/Dialect/Torch/Transforms/Passes.cpp
silvasean: ReifyAbstractInterpCalculationsUtils.h would make sense.

include/torch-mlir/Dialect/Torch/Transforms/Passes.td
```diff
@@ -151,7 +151,11 @@ def InlineGlobalSlots : Pass<"torch-inline-global-slots", "ModuleOp"> {
 
 def ReduceOpVariants : Pass<"torch-reduce-op-variants", "func::FuncOp"> {
   let summary = "Reduces variants of ops to a smaller set of ops.";
-  let constructor = "mlir::torch::Torch::createReduceOpVariantsPass()";
+  let constructor = "mlir::torch::Torch::createReduceOpVariantsPass(\"\")";
```
ramiro050, line 154: nit: can you add `/*extraLibrary=*/` to the constructor argument to be consistent with other constructors in this file?

Same comment applies to changes below

ramiro050 (changes requested):
Great work on this, both on speed and execution. I just have some minor comments

lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp
```diff
@@ -15,11 +15,14 @@ using namespace mlir;
...
 
-static std::string getLibraryFunctionPrefix(LibraryFunctionKind libFuncKind) {
+std::string
+mlir::torch::Torch::getLibraryFunctionPrefix(LibraryFunctionKind libFuncKind) {
   if (libFuncKind == LibraryFunctionKind::ShapeFunction)
     return "__torch_mlir_shape_fn.";
   else if (libFuncKind == LibraryFunctionKind::DtypeFunction)
     return "__torch_mlir_dtype_fn.";
+  else if (libFuncKind == LibraryFunctionKind::HasValueSemantics)
+    return "__torch_mlir_has_value_semantics.";
```
ramiro050, line 25: nit: can you add an `_fn` after `semantics`? This makes it consistent with the shape and dtype ones, which means we don't have to special case when generating the MLIR snippets from Python

ramiro050 (changes requested):

ramiro050 (changes requested):

test/python/custom_op_shape_dtype_fn.py
```diff
@@ -0,0 +1,71 @@
...
+func.func @__torch_mlir_shape_fn.goofy.identity(%arg0: !torch.list<int>) -> !torch.list<int> {
+  return %arg0 : !torch.list<int>
+}
+func.func @__torch_mlir_dtype_fn.goofy.identity(%arg0: !torch.int, %arg1: !torch.int) -> !torch.int {
+  return %arg1 : !torch.int
+}
+func.func @__torch_mlir_has_value_semantics.goofy.identity() { return }
+"""
+
+with tempfile.NamedTemporaryFile(mode="w") as tmp:
```
ramiro050, line 51: Do you know if this test works on Windows? According to the docs, [opening a temporary file a second time](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) while the file is still open does not work on Windows.

makslevental: good catch

lib/Dialect/Torch/Transforms/ReifyDtypeCalculations.cpp
```diff
@@ -106,13 +106,24 @@ dtypeFunctionArgsBuilder(OpBuilder &b, Location loc,
...
+      loadExtraLibrary(extraLibrary.getValue(), context, extraLibraryModule);
```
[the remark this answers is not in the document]
makslevental: Might've misinterpreted the intended API but changed it to

```cpp
LogicalResult mlir::torch::Torch::loadExtraLibrary(
    const std::string &filename, OptionalOwningOpRefModuleRef module,
    OptionalOwningOpRefModuleRef moduleToAppendTo,
    std::optional<MLIRContext *> context)
```

where you can provide either an unintialized `OwningOpRef<ModuleOp> module` and a `context` or a `moduleToAppendTo`. Let me know if you had something different in mind.

lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp
```diff
@@ -288,3 +297,43 @@ FailureOr<Value> Torch::adjustFunctionArg(
   // Pass the operand as-is.
   return operand;
 }
+
+#include <iostream>
```
ramiro050, line 301: `iostream` should not be included

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -241,20 +267,38 @@ reduceNonValueTensorLiteralOpToValueTensorLiteralOp(NonValueTensorLiteralOp op,
...
+  }
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     RewritePatternSet patterns(context);
-    patterns.add<ConvertHasValueSemanticsOpsToValueTensors>(context);
+    OwningOpRef<ModuleOp> extraLibraryModule;
+    std::optional<SymbolTable> extraLibraryModuleSymTable = std::nullopt;
+    if (!extraLibrary.empty() &&
+        succeeded(loadExtraLibrary(extraLibrary, extraLibraryModule,
+                                   std::nullopt, context))) {
```
ramiro050, line 283: nit: can you add `/*moduleToAppendTo=*/` to make it clear what `nullopt` represents?

lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp
```diff
@@ -288,3 +297,43 @@ FailureOr<Value> Torch::adjustFunctionArg(
...
+
+LogicalResult mlir::torch::Torch::loadExtraLibrary(
+    const std::string &filename, OptionalOwningOpRefModuleRef module,
+    OptionalOwningOpRefModuleRef moduleToAppendTo,
+    std::optional<MLIRContext *> context) {
+  assert(((module.has_value() && context.has_value()) ||
+          moduleToAppendTo.has_value()) &&
+         "Need to provide either moduleToAppendTo or context");
```
ramiro050, line 309: I think we can simplify the interface to this helper function by removing the `module` argument. This allows us to get rid of all the `optional`s in the signature, making it very clear to the user what arguments are expected when using this helper from just looking at the signature .

Then, when setting the value of `moduleToAppendTo`, we can just check how many blocks the module has, and handle accordingly (we should assert that the module has zero or one blocks in it). The handling of the empty module is still in a way an append, so the name `moduleToAppendTo` being used for both cases makes sense to me.

lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.h
```diff
@@ -60,6 +65,17 @@ FailureOr<Value> adjustFunctionArg(
     OpBuilder &b, Location loc, Value operand, Type desiredType,
     function_ref<Value(OpBuilder &, Location, Value, Type)> baseTransformation =
         [](OpBuilder &, Location, Value operand, Type) { return operand; });
+
+std::string getLibraryFunctionPrefix(LibraryFunctionKind libFuncKind);
+
+using OptionalOwningOpRefModuleRef =
+    std::optional<std::reference_wrapper<OwningOpRef<ModuleOp>>>;
+LogicalResult
+loadExtraLibrary(const std::string &filename,
```
ramiro050, line 74: nit: add small comments to both of these describing what they do

lib/Dialect/Torch/Transforms/ReifyDtypeCalculations.cpp
```diff
@@ -61,13 +61,21 @@ dtypeFunctionArgsBuilder(OpBuilder &b, Location loc,
...
+  }
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     ModuleOp module = getOperation();
     OwningOpRef<ModuleOp> library =
         parseSourceString<ModuleOp>(getAbstractInterpLibrary(), context);
+    if (!extraLibrary.empty())
+      if (failed(mlir::torch::Torch::loadExtraLibrary(
+          extraLibrary, std::nullopt, library)))
+        return signalPassFailure();
```
ramiro050, line 78: We should `emitError` here with a nice error message

lib/Dialect/Torch/Transforms/ReifyShapeCalculations.cpp
```diff
@@ -66,6 +71,10 @@ class ReifyShapeCalculationsPass
     // O(#ops in the program) ideally.
     OwningOpRef<ModuleOp> library =
         parseSourceString<ModuleOp>(getAbstractInterpLibrary(), context);
+    if (!extraLibrary.empty())
+      if (failed(mlir::torch::Torch::loadExtraLibrary(
```
ramiro050, line 75: same comment: emitError

ramiro050 (changes requested):
Looking good. Just a few more comments

include/torch-mlir/Dialect/Torch/Transforms/Passes.td
```diff
@@ -151,7 +151,13 @@ def InlineGlobalSlots : Pass<"torch-inline-global-slots", "ModuleOp"> {
 
 def ReduceOpVariants : Pass<"torch-reduce-op-variants", "func::FuncOp"> {
   let summary = "Reduces variants of ops to a smaller set of ops.";
-  let constructor = "mlir::torch::Torch::createReduceOpVariantsPass()";
+  let constructor = [{
+    mlir::torch::Torch::createReduceOpVariantsPass(/*extraLibrary=*/"")
+  }];
+  let options = [
+    Option<"extraLibrary", "extra-library", "std::string", /*default*/"",
```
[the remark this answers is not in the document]
makslevental: sorry - missed these

lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.cpp
ramiro050: From looking at Sean's comment about `LogicalResult` above, I'm not sure if this is the interface you already had previously. We might be going around in circles a bit.

lib/Dialect/Torch/Transforms/ReifyAbstractInterpCalculationsUtils.h
```diff
@@ -60,6 +65,16 @@ FailureOr<Value> adjustFunctionArg(
...
         [](OpBuilder &, Location, Value operand, Type) { return operand; });
+
+std::string getLibraryFunctionPrefix(LibraryFunctionKind libFuncKind);
+
+// Parse MLIR module at `filename` into a ModuleOp that will then
+// be appended to an existing, fully hydrated, ModuleOp; note the module
+// should have been instantiated with an associated context like so:
+// `OwningOpRef<ModuleOp> module = ModuleOp::create(UnknownLoc::get(&context));`
+LogicalResult loadExtraLibrary(const std::string &filename,
+                               OwningOpRef<ModuleOp> &moduleToAppendTo);
```
ramiro050, line 76: Ah, even better :)

ramiro050 (approved):
LGTM! Thanks for all the work on this

PR #1972, opened 2023-03-25
Proof of concept TorchDialectExtension

makslevental:


silvasean:
Can you explain a bit more what your goal is here? The original idea was that we would tunnel custom ops through torch.operator. If we have this mechanism, then it is somewhat redundant with that.

I'm just curious why it is needed given the custom ops RFC we are implementing.

silvasean:
Maybe we can chat about it at the developer hour

powderluv:
Can we reopen this ? We are hitting the need for this again where we have some custom ops that are defined that need some custom passes

makslevental:
> Can we reopen this ? We are hitting the need for this again where we have some custom ops that are defined that need some custom passes

Okay let me rebase and fix up

PR #1982, opened 2023-03-28
[MLIR][TORCH] Add decomposition for prims::squeeze op

vivekkhandelwal1:
This commit adds the decomposition for the prims.squeeze op for the cases where the dimensions list contains atmost one dim value. This restriction is there because right now, we only need support for this limited case.


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -736,6 +736,13 @@ def aten〇squeeze〡shape(self: List[int]) -> List[int]:
 def aten〇squeeze〇dim〡shape(self: List[int], dim: int) -> List[int]:
     return upstream_shape_functions.squeeze(self, dim)
 
+def prims〇squeeze〡shape(a: List[int], dimensions: List[int]) -> List[int]:
```
ramiro050, line 739: Can you upstream this? This op does not have the same behavior as `aten.squeeze`. In particular, if the dimension specified to squeeze is not equal to 1, the `prims` version results in an error. Also, we should be able to have the full version of this shape function, since it does not require that much more work. You just need to sort `dimensions` in decreasing order, then check the corresponding values in `a` is equal to `1` and call `upstream_shape_functions.squeeze` if it is.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3872,6 +3872,38 @@ class DecomposeAtenNewEmptyStridedOp
...
+    SmallVector<int64_t> dimensions;
+    if (!matchPattern(op.getDimensions(),
+                      m_TorchListOfConstantInts(dimensions)))
+      return rewriter.notifyMatchFailure(
+          op, "all dimensions must be constant ints");
+
+    if (dimensions.size() > 1)
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: prims.squeeze op supported only for dimensions "
+              "list containing single value");
```
ramiro050, line 3893: Having full support does not require that much more work, so we should just add it. You just need to sort `dimensions in decreasing order and call `AtenSqueezeDimOp` for each dimension. (Do remember to check that `a.size(dim) == 1` for each `dim`.

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
vivekkhandelwal1: I think we can use this shape function: https://github.com/pytorch/pytorch/blob/master/torch/jit/_shape_functions.py#L429-L436, it's doing the same behaviour. It's also called the `upstream_shape_functions.squeeze` and here(https://github.com/pytorch/pytorch/blob/master/torch/jit/_shape_functions.py#L423) it checks for the shape equal to 1.

ramiro050: Thanks!

ramiro050 (approved):
LGTM!

PR #1983, opened 2023-03-28
[TOSA] Add support for AtenZerosOp 0 layout

AmosLewis:
Find this issue after the slice and copy https://github.com/llvm/torch-mlir/pull/1970 patch and lower to tosa for t5 model.
`shifted_input_ids = decoder_input_ids.new_zeros(decoder_input_ids.shape)`
`%2 = torch.aten.new_zeros %arg2, %1, %int4, %int0, %cpu, %false : !torch.tensor, !torch.list<int>, !torch.int, !torch.int, !torch.Device, !torch.bool -> !torch.tensor
`


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -4197,9 +4197,13 @@ class ConvertAtenConstPatternOp : public OpConversionPattern<AtenOpT> {
 
     // FIXME: Handle layout, device and pin_memory. Assume dtype has been
     // processed to set output type correctly?
-    if (!op.getLayout().getType().template isa<Torch::NoneType>())
+    int64_t layout;
+    if (!op.getLayout().getType().template isa<Torch::NoneType>() &&
+        (!matchPattern(op.getLayout(), m_TorchConstantInt(&layout)) ||
+         layout != 0)) {
```
eric-k256, line 4203: Looking at similar code in TorchOps.cpp, can we check against torch_upstream::Layout::Strided instead of 0? It makes it easier to read and keeps consistent in both locations, which could save us from unexpected failures if torch changes.

eric-k256 (commented):
In addition to the one comment, is there something we should add to the e2e tests to help us catch problems?

eric-k256 (approved):

AmosLewis:
> In addition to the one comment, is there something we should add to the e2e tests to help us catch problems?

Done

PR #1984, opened 2023-03-28
Remove convolution_overrideable, convolution_backward_overrideable

ramiro050:
The ops `aten.convolution_overrideable` and
`aten.convolution_backward_overrideable` are currently not e2e tested in Torch-MLIR. Moreover, there is no way to add e2e tests for them because the ops cannot be called using the CPU backend (this also prevents adding tested dtype functions for these ops). Since these two ops are not expected to ever appear in PyTorch traces obtained through standard means (https://github.com/pytorch/pytorch/issues/97481), Torch-MLIR should not have to worry about them.


li-plus (approved):
LGTM, btw do we have gpu machine for ci by any chance?

ramiro050:
> btw do we have gpu machine for ci by any chance?

One had been added https://github.com/llvm/torch-mlir/issues/1669, but then I think it got removed since people no longer needed it. @powderluv, can you confirm? 

What is the use case you need a GPU in CI for?

powderluv:
yup we dropped it since we now moved to tracing the F16 graphs on the cpu.

PR #1990, opened 2023-03-31
Use meta tensors for shape and dtype function tests

ramiro050:
There are some restrictions on the CPU device that prevent the current dtype functions from working with graphs extracted using other devices. For example, many ops don't support float16 on CPU, but they do on CUDA. To avoid these restrictions, this commit switches the abstract interpretation library testing framework to use the META device. The META device is a lot more lenient in terms of what dtypes are supported, runs computations on cpu, and because META tensors don't carry any data, the actual checks are performed much faster. This removes all the float16 restrictions.

I did notice some inconsistencies in the expected results for one op when using CPU vs META, and some ops simply don't work on the META device. I will open bugs upstream for the inconsistencies, and in a follow up patch I will add a way for `@check`s to specify device to test the ops that don't run on META. For now, I have commented out the tests of those ops.

Inconsistent ops:
- aten.threshold_backward

Cannot run on META:
- aten._convolution
- aten._convolution.deprecated
- aten.bincount


li-plus (approved):
LGTM!

PR #1991, opened 2023-03-31
Minor fixes for `ConvertTorchConversionToMLProgram`.

arames:
* Only create the global seed variable if it does not exist already.
* Make the pass a module pass. A func pass may not modify its parent op.

### DIFF
--- a/test/Conversion/TorchConversionToMLProgram/multiple_functions.mlir
+++ b/test/Conversion/TorchConversionToMLProgram/multiple_functions.mlir
@@ -1,0 +1,15 @@
+// RUN: torch-mlir-opt %s -convert-torch-conversion-to-mlprogram -split-input-file | FileCheck %s
+
+module {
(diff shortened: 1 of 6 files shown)


arames:
The current implementation works when there is  only a single function in the input module.

test/Conversion/TorchConversionToMLProgram/multiple_functions.mlir
```diff
@@ -0,0 +1,15 @@
+// RUN: torch-mlir-opt %s -convert-torch-conversion-to-mlprogram -split-input-file | FileCheck %s
+
+module {
+  func.func private @f0() -> i64
```
arames, line 4: A higher number of functions helps trigger the race when multiple threads process them concurrently.

vivekkhandelwal1 (approved):
LGTM!

silvasean:
Good catch! Thanks!

PR #1996, opened 2023-04-03
[WIP] Pure-Python FX importer.

makslevental:
Currently passes

```
Summary:
    Passed: 706
    Failed: 104
    Expectedly Failed: 28
```

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -56,20 +56,7 @@ TORCHDYNAMO_XFAIL_SET = {
     "ElementwiseWhereScalarSelfModule_basic",
     "ElementwiseWhereScalarOtherStaticModule_basic",
     "ElementwiseWhereScalarSelfStaticModule_basic",
-    # %7 = torch.operator "aten._index_put_impl_.hacked_twin"(%1, %6, %5, %true, %false) : (!torch.tensor<*,f32>, !torch.list<tensor>, !torch.tensor<*,f32>, !torch.bool, !torch.bool) -> !torch.tensor
-    "IndexPutImpl1DFloatAccumulateModule_basic",
-    "IndexPutImpl1DFloatNonAccumulateModule_basic",
-    "IndexPutImpl1DIntAccumulateModule_basic",
-    "IndexPutImpl1DIntNonAccumulateModule_basic",
-    "IndexPutImpl2DFloatAccumulateModule_basic",
-    "IndexPutImpl2DFloatNonAccumulateModule_basic",
-    "IndexPutImpl2DIndexModule_basic",
-    "IndexPutImpl3DFloatAccumulateModule_basic",
-    "IndexPutImpl3DFloatNonAccumulateModule_basic",
 
-    # https://github.com/llvm/torch-mlir/issues/1611
-    # error: 'tensor.cast' op operand type 'tensor<0xi64>' and result type 'tensor<18xi64>' are cast incompatible
-    "Aten_EmbeddingBagExample_basic",
     # error: failed to legalize operation 'torch.valsem.aten.bernoulli.float' that was explicitly marked illegal
     "BernoulliFloatModule_basic",
     "BernoulliPModule_basic",
@@ -77,8 +64,6 @@ TORCHDYNAMO_XFAIL_SET = {
     "ElementwiseFlattenBroadcastModule_basic",
     "FlattenRank0Module_basic",
     "UniformModule_basic",
-    # error: failed to materialize conversion for result #0 of operation 'torch.aten.t' that remained live after conversion
-    "TModuleRank1_basic",
     # error: unsupported by backend contract: tensor with unknown rank
     # note: see current operation: %1 = "torch.tensor_static_info_cast"(%arg0) : (!torch.vtensor<[5,4,3,2,1],f32>) -> !torch.vtensor<*,f32>
     "ElementwisePreluModule_basic",
@@ -201,10 +186,87 @@ TORCHDYNAMO_XFAIL_SET = {
     'IsFloatingPointInt_False',
     'TorchPrimLoopForLikeModule_basic',
     'TorchPrimLoopWhileLikeModule_basic',
+
+    # backend never runs because of empty frame
+    'ConstantBoolParameterModule_basic',
+
+    # 'torch.aten.mul.Tensor' op operand #1 must be Any Torch tensor type, but got '!torch.float'
+    "AddCDivModule_basic",
+    "ElementwiseMulScalarModule_basic",
+    "ElementwiseMulScalarModule_float",
+    "UpSampleNearest2dDynamicSize_basic",
+    "UpSampleNearest2dStaticFactor_basic",
+    "UpSampleNearest2dStaticSize_basic",
+    "UpSampleNearest2d_basic",
+
+    # 'torch.aten.add.Tensor' op operand #1 must be Any Torch tensor type, but got '!torch.float'
+    "BatchNorm1DModule_basic",
+    "BatchNorm1DWith2DInputModule_basic",
+    "BatchNorm2DModule_basic",
+    "BatchNorm3DModule_basic",
+    "ElementwiseAddScalarFloatModule_basic",
+    "ElementwiseAddScalarInt64Module_basic",
+    "ElementwiseAddScalarIntModule_basic",
+    "MobilenetV3Module_basic",
+    "NativeBatchNorm1DModule_basic",
+    "NativeBatchNorm2DModule_basic",
+    "NativeBatchNorm3DModule_basic",
+    "NativeBatchNormNoneWeightModule_basic",
+    "ResNet18Module_basic",
+    "ResNet18StaticModule_basic",
+
+    # 'torch.aten.add.Tensor' op operand #1 must be Any Torch tensor type, but got '!torch.int'
+    "ElementwiseAddScalar_TensorLiteralInt32_Module_basic",
+    "HBC_basic",
+
+    # 'torch.aten.div.Tensor' op operand #1 must be Any Torch tensor type, but got '!torch.float'
+    "ElementwiseDivScalarModule_basic",
+
+    # 'torch.aten.mul.Tensor' op operand #1 must be Any Torch tensor type, but got '!torch.int'
+    "ElementwiseMulScalarModule_int",
+
+    # 'torch.aten.sub.Tensor' op operand #1 must be Any Torch tensor type, but got '!torch.float'
+    "ElementwiseSubScalarFloatModule_basic",
+    "ElementwiseSubScalarIntModule_basic",
+
+    # Exception: Unsupported: missing default value for argument 0 in schema for aten.div.Tensor_mode
+    "ElementwiseDivRoundingModeFloorModule_basic",
+    "ElementwiseDivRoundingModeTruncModule_basic",
+
+    # Exception: Unsupported op: get_attr
+    "NumToTensorFloatModule_basic",
+    "NumToTensorIntModule_basic",
+    "TensorFloatModule_basic",
+    "TensorIntModule_basic",
+
+    # Exception: Unsupported: missing default value for argument 0 in schema for aten.randn.generator
+    "RandnGeneratorF64Module_basic",
+    "RandnGeneratorModule_basic",
+
+    # AssertionError: None (python constant/literal being passed to get_op_results_or_values through ods_gen)
+    "SliceEndSleStartModule_basic",
+    "SliceOutOfUpperBoundIndexModule_basic",
+    "SliceStartEqEndModule_basic",
+    "TensorsConcatModule_basic",
+    "TensorsConcatNegativeDimModule_basic",
+    "TensorsConcatPromoteDTypeModule_basic",
+    "TensorsStackModule_basic",
+    "TensorsStackNegativeDimModule_basic",
+    "TensorsStackPromoteDTypeModule_basic",
 }
 
-# See https://github.com/llvm/torch-mlir/issues/2050
 TORCHDYNAMO_CRASHING_SET = {
+    # No upstream decompositions.
+    # %6:4 = torch.operator "aten._embedding_bag_forward_only"(%1, %3, %5, %false, %int0, %false, %none, %false, %int-1) : (!torch.tensor<*,f32>, !torch.tensor<*,si64>, !torch.tensor<*,si64>, !torch.bool, !torch.int, !torch.bool, !torch.none, !torch.bool, !torch.int) -> (!torch.tensor, !torch.tensor, !torch.tensor, !torch.tensor)
+    # See also: https://github.com/pytorch/torchdynamo/issues/327
+    "Aten_EmbeddingBagExample_basic",
+    # https://github.com/pytorch/pytorch/issues/100838
+    "BaddbmmDifferentDtypesModule_basic",
+    "FullModuleInt3D_basic",
+    "ThresholdBackward1dIntModule_basic",
+    "ThresholdBackward2dIntModule_basic",
+    "ThresholdBackward3dIntModule_basic",
+    # See https://github.com/llvm/torch-mlir/issues/2050
     "ElementwiseCloneChannelsLastMemoryFormatModule_basic",
     "ElementwiseCloneContiguousModule_basic",
     "ElementwiseCloneModule_basic",
@@ -226,6 +288,7 @@ TORCHDYNAMO_CRASHING_SET = {
     "SliceOutOfLowerBoundStartIndexModule_basic",
     "SliceSizeTwoStepModule_basic",
     "SliceStaticModule_basic",
+    "TestMultipleTensorAndPrimitiveTypesReturn_basic",
     "TModuleRank2_basic",
     "ToCopyModule_basic",
     "TransposeIntModule_basic",
--- a/python/torch_mlir/__init__.py
+++ b/python/torch_mlir/__init__.py
@@ -12,8 +12,8 @@ import tempfile
 
 from torch._functorch.compile_utils import strip_overloads
 import torch
+import torch.fx
 
-from torch_mlir.passmanager import PassManager
 from .compiler_utils import run_pipeline_with_repro_report
 from torch_mlir.dialects.torch.importer.jit_ir import ClassAnnotator, ImportOptions, ModuleBuilder
 from torch_mlir.dialects.torch.importer.jit_ir.build_tools.library_generator import generate_library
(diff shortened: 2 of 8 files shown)


python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -2946,7 +2946,7 @@ def forward(self, input, batch1, batch2):
         return torch.ops.aten.baddbmm(input, batch1, batch2)
 
 
-@register_test_case(module_factory=lambda: BaddbmmDifferentDtypesModule())
+# @register_test_case(module_factory=lambda: BaddbmmDifferentDtypesModule())
```
makslevental, line 2949: segfault

python/torch_mlir_e2e_test/test_suite/threshold.py
```diff
@@ -152,7 +152,7 @@ def __init__(self):
     def forward(self, grad, input):
         return torch.ops.aten.threshold_backward(grad, input, 1)
 
-@register_test_case(module_factory=lambda: ThresholdBackward1dIntModule())
+# @register_test_case(module_factory=lambda: ThresholdBackward1dIntModule())
```
makslevental, line 155: segfault

python/torch_mlir_e2e_test/test_suite/threshold.py
```diff
@@ -171,7 +171,7 @@ def __init__(self):
     def forward(self, grad, input):
         return torch.ops.aten.threshold_backward(grad, input, 0.5)
 
-@register_test_case(module_factory=lambda: ThresholdBackward2dIntModule())
+# @register_test_case(module_factory=lambda: ThresholdBackward2dIntModule())
```
makslevental, line 174: segfault

python/torch_mlir_e2e_test/test_suite/threshold.py
```diff
@@ -190,7 +190,7 @@ def __init__(self):
     def forward(self, grad, input):
         return torch.ops.aten.threshold_backward(grad, input, 1)
 
-@register_test_case(module_factory=lambda: ThresholdBackward3dIntModule())
+# @register_test_case(module_factory=lambda: ThresholdBackward3dIntModule())
```
makslevental, line 193: segfault

python/torch_mlir_e2e_test/configs/torchdynamo.py
```diff
@@ -76,24 +77,30 @@ def refine_result_type(_result):
...
+            module = torch_mlir.compile(
+                artifact, item.inputs, output_type="linalg-on-tensors")
+            module = self.backend.compile(module)
+            backend_module = self.backend.load(module)
+            params = {
+                **dict(artifact.named_parameters(remove_duplicate=False)),
+                **dict(artifact.named_buffers(remove_duplicate=False)),
+            }
+            params_flat, params_spec = pytree.tree_flatten(params)
+            params_flat = list(params_flat)
```
makslevental, line 100: aot functionalizes modules

makslevental:
@ramiro050 @silvasean @powderluv 

There's a blocker here (I believe); take for example

```python
class ZeroInt32Module(torch.nn.Module):

    def __init__(self):
        super().__init__()

    def forward(self, tensor):
        return torch.ops.aten.zero_(tensor)
```

Note the underscore, i.e. mutation/non-value semantics. Currently (through the TS path), this gets decomposed to 

```mlir
module attributes {torch.debug_module_name = "ZeroInt32Module"} {
  func.func @forward(%arg0: !torch.vtensor<[10,4],f32>) -> !torch.vtensor<[10,4],f32> {
    %int0 = torch.constant.int 0
    %0 = torch.aten.fill.Scalar %arg0, %int0 : !torch.vtensor<[10,4],f32>, !torch.int -> !torch.vtensor<[10,4],f32>
    return %0 : !torch.vtensor<[10,4],f32>
  }
}
```

(somewhere in `DecomposeComplexOps` or `ReduceOpVariants`?)

If you pass the same model through the current version of the fx importer, you get

```mlir
module attributes {torch.debug_module_name = "ZeroInt32Module"} {
  func.func @ZeroInt32Module(%arg0: !torch.vtensor<[10,4],f32>) -> (!torch.vtensor<[10,4],f32>, !torch.vtensor<[10,4],f32>) {
    %0 = torch.aten.zero %arg0 : !torch.vtensor<[10,4],f32> -> !torch.vtensor<[10,4],f32> loc(#loc1)
    return %0, %0 : !torch.vtensor<[10,4],f32>, !torch.vtensor<[10,4],f32> loc(#loc)
  }
} 
```

Note the repetition of the result (`%0, %0`); correctness aside, this is happening because `aot_autograd` functionalizes and for mutating ops that means keeping mutated inputs; see [here](https://github.com/pytorch/pytorch/blob/1c226f5aad4230caa36b28ebf69482a9398c6d08/torch/_functorch/aot_autograd.py#L923-L928). Again, this is being lowered incorrectly by the current fx importer - it should be the input itself in a tuple with the zeroed tensor (or something like that), but that's not the issue.

If you [click to the aot_autograd source](https://github.com/pytorch/pytorch/blob/1c226f5aad4230caa36b28ebf69482a9398c6d08/torch/_functorch/aot_autograd.py#L923-L928) you'll see a flag `keep_data_input_mutations` - this flag allows input mutations:

```
# This function returns a new function that returns mutated inputs as outputs.
# if keep_data_input_mutations is set, then we assume that data-only mutations
# will be left in the graph, and we only return metadata-mutated inputs as outputs.
def fn_input_mutations_to_outputs(
...
```
and is handled appropriately elsewhere in `aot_autograd`. It can be passed to 

```python
my_backend = aot_autograd(fw_compiler=my_aot_autograd_backend, 
                          keep_inference_input_mutations=False,
                          decompositions=_get_decomposition_table)
```

But then we hit the real problem; importing again we get:

```mlir
module attributes {torch.debug_module_name = "ZeroInt32Module"} {
  func.func @ZeroInt32Module(%arg0: !torch.vtensor<[10,4],f32>) -> !torch.vtensor<[10,4],f32> {
    %0 = torch.aten.zero %arg0 : !torch.vtensor<[10,4],f32> -> !torch.vtensor<[10,4],f32>
    %1 = torch.aten.copy_ %arg0, %0, %false : !torch.vtensor<[10,4],f32>, !torch.vtensor<[10,4],f32>, !torch.bool -> !torch.vtensor<[10,4],f32>
    return %0 : !torch.vtensor<[10,4],f32>
  }
}
```

Ignoring that the types are wrong (because, as I'm implying here, fx importer wrongly assumes everything is always value semantics), we see that we now have to contend with a `torch.aten.copy_`. 

So if we're to use the `keep_inference_input_mutations` backdoor to handle "mutated inputs as outputs", then fx importer needs to be aware of such operators and furthermore we will still have a dependency on `ReduceOpVariants`. And if we're not going to use the backdoor then I don't know how to handle such cases; possibly this should be handled by an fx decomposition that reproduces what `ReduceOpVariants` currently does? But from what I can tell you guys are relying on only upstream decompositions and I don't know if upstream will take such PRs.

powderluv:
Adding @Chillee and @ezyang in case they have guidance on what can be upstreamed and the FX importer in general.

ezyang:
Not sure what you're proposing to upstream.

python/torch_mlir/__init__.py
```diff
@@ -13,11 +12,28 @@
...
 from .compiler_utils import run_pipeline_with_repro_report
 from torch_mlir.dialects.torch.importer.jit_ir import ClassAnnotator, ImportOptions, ModuleBuilder
 from torch_mlir.dialects.torch.importer.jit_ir.build_tools.library_generator import generate_library
 
+import torch.fx
+from torch._dynamo.backends.common import aot_autograd
+from torch._functorch.aot_autograd import make_boxed_compiler, set_model_name, get_aot_compilation_context
+from torch_mlir._dynamo_fx_importer import import_fx_graph_as_func
+
+mlir_module = None
```
silvasean, line 25: we should not be using a global here. The right solution uses dynamo.export or similar API's. We can also expose helpers in torch_mlir.dynamo for helping users to plug in their MLIR backends

python/torch_mlir/__init__.py
```diff
@@ -333,105 +349,125 @@ def compile(model: torch.nn.Module,
...
-                    f"Model does not have exported method '{method_name}', "
-                    f"requested in `example_args`. Consider adding "
-                    f"`@torch.jit.export` to the method definition.")
-        scripted = model
-    elif use_tracing:
-        scripted = torch.jit.trace_module(
-            model,
-            example_args._get_for_tracing(use_tracing, ignore_traced_shapes)
+    global mlir_module
+    if not isinstance(model, torch.fx.GraphModule):
```
silvasean, line 353: we should not need this branch. We need a flag like use_tracing, except it will be `use_dynamo_export` or something like that (not sure if dynamo.export has been promoted to a more real/supported API yet. may need an "experimental" in the name somewhere). Dynamo can trace the through any torch.nn.Module, not just FX modules.

python/torch_mlir/__init__.py
silvasean: torch_mlir.compile is fundamentally an "AoT" compilation path. If we want an online/JIT Dynamo/torch.compile backend then we will want to expose that as a different API.

makslevental:
>Can you replace the current dynamo e2e test config with using this path and show the xfail diffs? I think even if this causes a few regressions it is an overall step forward.

That's what the quoted numbers in the PR description are - do you want the exact fails?

makslevental:
> Can you replace the current dynamo e2e test config with using this path and show the xfail diffs? I think even if this causes a few regressions it is an overall step forward.

The numbers in the description at the top are this. Do you want the names of the tests?

> Can you replace the current dynamo e2e test config with using this path and show the xfail diffs? I think even if this causes a few regressions it is an overall step forward.

The numbers in the description reflect this; here are the specifics:

[New fails](https://gist.github.com/makslevental/f37f675ccf8c2ea95216146fde6616ab#file-new_fails) (`all_fails - TORCHDYNAMO_XFAIL_SET`): 282

[Still fails](https://gist.github.com/makslevental/f37f675ccf8c2ea95216146fde6616ab#file-still_fails) (`TORCHDYNAMO_XFAIL_SET - all_fails`): 37

silvasean:
> > Can you replace the current dynamo e2e test config with using this path and show the xfail diffs? I think even if this causes a few regressions it is an overall step forward.
> 
> The numbers in the description reflect this; here are the specifics:
> 
> [New fails](https://gist.github.com/makslevental/f37f675ccf8c2ea95216146fde6616ab#file-new_fails) (`all_fails - TORCHDYNAMO_XFAIL_SET`): 282
> 
> [Still fails](https://gist.github.com/makslevental/f37f675ccf8c2ea95216146fde6616ab#file-still_fails) (`TORCHDYNAMO_XFAIL_SET - all_fails`): 37

Yes please update the PR with the xfail_sets.py that we would be submitting along with the new path.

I think we should analyze and diagnose the new fails. I suspect that 99% of them are covered by a few further improvements to the importer which we should do before submitting this.

python/torch_mlir/__init__.py
makslevental: >we should not be using a global here. The right solution uses dynamo.export or similar API's. We can also expose helpers in torch_mlir.dynamo for helping users to plug in their MLIR backends

I don't understand what this means; `dynamo.export`:

> Export an input function f to a format that can be executed outside of PyTorch using the FX graph.

It uses a trivial compiler `dynamo_normalization_capturing_compiler` to capture the FX graph; there is no support for plugging in your compiler; furthermore `dynamo_normalization_capturing_compiler` is **using nonlocals**:

```python
    def dynamo_normalization_capturing_compiler(
        gm: torch.fx.GraphModule, example_inputs
    ):
        nonlocal graph
        assert (
            graph is None
        ), "Tried to emit a second graph during export. Tracing through 'f' must produce a single graph."
        graph = gm

        def result_capturing_wrapper(*graph_inputs):
            nonlocal graph_captured_result
            nonlocal graph_captured_input

            graph_captured_input = graph_inputs
            assert graph is not None
            graph_captured_result = graph(*graph_inputs)
            return graph_captured_result

        return result_capturing_wrapper
```

So I can change to `nonlocal` but otherwise, I don't see a way.

silvasean: We don't plug in any compiler in the dynamo.export path. It gives us an FX graph (I don't care how it is implemented internally), and we import that. (we might need to file feature requests on their end for functionalization/decomposition and stuff).

I think the best path for the initial integration with dynamo/2.0 is via the JIT mode, i.e. not through torch_mlir.compile.

e2e_testing/xfail_sets.py
```diff
@@ -97,6 +93,13 @@
     # ERROR: RuntimeError: Found a custom (non-ATen) operator that either mutates or its inputs: prims::squeeze.. Getting these operators to work with functionalization requires some extra work. For mutable ops you need to register a corresponding out-of-place variant of the op, and you also need to register a Functionalization kernel that performs some boilerplate, telling functionalization to map from the mutable op to the out-of-place op. See a more complete example of how to do this at https://gist.github.com/bdhirsh/7dadbf6296f8f7d1abcf4c482f438aaa.
     "PrimsSqueezeModule_basic",
     "PrimsSqueezeEmptyDimensionsModule_basic",
+
+    # https://github.com/pytorch/pytorch/blob/master/torch/_dynamo/symbolic_convert.py#L1972
```
silvasean, line 97: can you file an upstream bug for this?

Also this looks like a dynamo issue -- I don't understand how this worked with the previous code but fails with the new code.

e2e_testing/xfail_sets.py
```diff
@@ -97,6 +93,13 @@
...
+
+    # https://github.com/pytorch/pytorch/blob/1c226f5aad4230caa36b28ebf69482a9398c6d08/torch/_functorch/aot_autograd.py#L923-L928
```
silvasean, line 100: can you file an upstream bug for this?

sherry-1001:
hi, I am a mlir community developer. could I ask a question: When is this PR expected to merge?

makslevental:
> hi, I am a mlir community developer. could I ask a question: When is this PR expected to merge?

Soonish - I was hoping to get it done today but it'll have to wait until Monday.

e2e_testing/xfail_sets.py
makslevental: These aren't upstream bugs, these are false negative tests in torch-mlir; see [false_negatives.py](https://github.com/llvm/torch-mlir/pull/1996/files#diff-4f5914874fa7aa4c05ae28dbbf8b6468f612a1885ce226cd3fab03cdd3b84ffa) where you have two examples (`ViewCollapseDynamicWithAtenSizeIntModule` and `NeFloatIntModule`) of tests that should fail because torch-mlir returns the wrong thing but nonetheless pass because there's a graph break and dynamo punts to runtime eval:

```
m = ViewCollapseDynamicWithAtenSizeIntModule()
o = run(m, (torch.rand(2, 3, 5, 4, 12, 32), torch.tensor(3), torch.tensor(5)))
assert isinstance(o, torch.Tensor)
assert o.shape == (2, 3, 5, 4, 384)
```

but `print(type(r))` inside of `_refbackend_torchdynamo_backend` prints `<class 'tuple'>` corresponding to an fx graph like this

```
graph():
    %arg0_1 : [#users=1] = placeholder[target=arg0_1]
    %arg1_1 : [#users=1] = placeholder[target=arg1_1]
    %arg2_1 : [#users=1] = placeholder[target=arg2_1]
    %clone : [#users=1] = call_function[target=torch.ops.aten.clone](args = (%arg0_1,), kwargs = {})
    %clone_1 : [#users=1] = call_function[target=torch.ops.aten.clone](args = (%arg1_1,), kwargs = {})
    %clone_2 : [#users=1] = call_function[target=torch.ops.aten.clone](args = (%arg2_1,), kwargs = {})
    return (clone, clone_1, clone_2)
```

Dynamo break the graph for an unrelated reason:

```
[2023-04-14 17:23:35,811] torch._dynamo.symbolic_convert: [WARNING] Graph break: call_function BuiltinVariable(int) [TensorVariable()] {} from user code at   File "false_negative.py", line 93, in <lambda>
    f = lambda method, *inputs: method(*inputs)
  File "false_negative.py", line 90, in item_symbol_that_clones_inputs
    result = model.forward(*cloned_inputs)
  File "false_negative.py", line 104, in forward
    return a.view(a.size(0), int(b), int(c), a.size(3), 384)
```

cc @powderluv

makslevental: putting 

```python
                if True:
                    raise Exception("raising break")
```

[here](https://github.com/pytorch/pytorch/blob/e2923b521bd7fc67fa1e8f404397b20d1ec81668/torch/_dynamo/symbolic_convert.py#L358-L359) in dynamo gives you the real number of fails:

```
Summary:
    Passed: 750
    Failed: 57
    Expectedly Failed: 38
```

python/test/dynamo_fx_importer/false_negative.py
```diff
@@ -0,0 +1,139 @@
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+# RUN: %PYTHON %s | FileCheck %s
+
```
silvasean, line 6: Why are we effectively running e2e tests here in the unittest directory?

e2e_testing/xfail_sets.py
```diff
@@ -59,27 +59,14 @@
     "ElementwiseWhereScalarOtherModule_basic",
     "ElementwiseWhereScalarSelfModule_basic",
     # %7 = torch.operator "aten._index_put_impl_.hacked_twin"(%1, %6, %5, %true, %false) : (!torch.tensor<*,f32>, !torch.list<tensor>, !torch.tensor<*,f32>, !torch.bool, !torch.bool) -> !torch.tensor
```
silvasean, line 61: I think you need to remove this comment too

python/torch_mlir/__init__.py
```diff
@@ -13,11 +12,12 @@
 from torch._functorch.compile_utils import strip_overloads
 import torch
 
-from torch_mlir.passmanager import PassManager
 from .compiler_utils import run_pipeline_with_repro_report
 from torch_mlir.dialects.torch.importer.jit_ir import ClassAnnotator, ImportOptions, ModuleBuilder
 from torch_mlir.dialects.torch.importer.jit_ir.build_tools.library_generator import generate_library
 
+import torch.fx
```
silvasean, line 19: Can you move this up next to the other `import torch`?

python/test/dynamo_fx_importer/false_negative.py
makslevental: Because this isn't a test - this a demo to show**you** that current tests are broken

e2e_testing/xfail_sets.py
silvasean: Yes, this seems like a Dynamo bug -- Dynamo should trace a graph but bails out, so it is missing feature work in Dynamo. I would suggest that we run the tests with nopython=True and file any failures as upstream Dynamo bugs.

python/test/dynamo_fx_importer/false_negative.py
silvasean: Hey sorry looks like I was going through my email notifications in a bad order and saw this before your other comment. Moved discussion to there.

.github/workflows/buildAndTest.yml
```diff
@@ -23,7 +23,7 @@ concurrency:
 jobs:
   build-test:
     strategy:
-      fail-fast: true
```
makslevental, line 26: TODO: undo

sherry-1001:
Hi, I'm also doing fx graph lowering.  Could I ask a question: why not we lower fx graph based on `fx.Interpreter` which can help analysis `fx.GraphModule`, be like `inductor`.

makslevental:
> Hi, I'm also doing fx graph lowering. Could I ask a question: why not we lower fx graph based on `fx.Interpreter` which can help analysis `fx.GraphModule`, be like `inductor`.

I think this is a fair question/criticism. I can't speak for @silvasean (who is the original author of this PR) but I wasn't aware of `fx.Interpreter` until your comment. Nonetheless, I think for such a simple thing (an interpreter that only handles ~6 ops) it doesn't matter in one direction or the other (reuse their code to save some on our side vs. be married to the `fx.Interpreter` API).

silvasean:
IIRC the fx.Interpreter just made the code uglier.

Issue #1989, opened 2023-03-31
Add support for ToMe [dbolya/tomesd]


Abhishek-Varma:
[ToMe](https://github.com/dbolya/tomesd) updates forward functions of certain blocks from Unet due to which we need to have support for the following ops :-
1. `aten.randint`.
2. `aten.scatter_.src`.
3. `aten.sort`.

Shukla-Gaurav:
There is a PR for the second one. It's been long since the path is raised, I will try to wrap this up quickly https://github.com/llvm/torch-mlir/pull/1499. Thanks!

ramiro050:
@Shukla-Gaurav, you may want to look at the changes in https://github.com/llvm/torch-mlir/pull/1884. It adds a helper that I think will be useful for `scatter_`

https://github.com/llvm/torch-mlir/blob/b2af43b17cb2276f0b795d6ebf6feb5ccd41c6cf/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp#L85

Abhishek-Varma:
`aten.randint` PR for review : https://github.com/llvm/torch-mlir/pull/2001

@powderluv @ramiro050 @vivekkhandelwal1

PR #2003, opened 2023-04-04
[MLIR][TORCH] Add e2e support for aten.sort op

Abhishek-Varma:
-- This commit adds e2e support for `atend.sort` op.

This is a RFC PR.

I’d appreciate some help on how we can go about implementing this cleanly in Linalg or TMTensor (or if some canonicalization earlier in the pipeline can be implemented to deal with this).

1. There’s an `aten.sort.int` op support added by @vivekkhandelwal1 which deals with sorting a list of constant integers using `std::sort`.
2. Whereas I’m trying to add `aten.sort` op which would sort the N-Dimensional input tensor and return both :-
     (a) the sorted input tensor as well as 
     (b) the input tensor’s indices in the sorted tensor.

`aten.sort` op's support is needed to support https://github.com/llvm/torch-mlir/issues/1989 and on close inspection it seems that aforementioned point `2.b` is of particular interest to the code.

I looked up Linalg, Arith, Math and Tensor dialect ops but couldn’t find a 1:1 support for this op.
Looked at [Linalg_Ext's SortOp](https://github.com/openxla/iree/blob/366edbfd7ea1116b7465224ccd7ae61a470bb78e/llvm-external-projects/iree-dialects/include/iree-dialects/Dialect/LinalgExt/IR/LinalgExtOps.td#L178) for the TMTensor path but that's basically same as [MHLO SortOp](https://www.tensorflow.org/mlir/hlo_ops#mhlosort_mlirmhlosortop) and doesn't fit well, at least not 1:1, with the requirement of `aten.sort`.

@ramiro050 @powderluv @vivekkhandelwal1


Abhishek-Varma:
> `linalg_ext.sort` should work. The inputs to `linalg_ext.sort` will be two tensors: the input tensor and an indices tensor. The comparator will then use the input tensor values for the comparison, sorting both tensors in the same manner.
> 
> What was your concern with `linalg_ext.sort`?

Thanks for your inputs @ramiro050 !

Must've been an error on my part in understanding the op's constraints.
So, I misunderstood the comparator's relation w.r.t the two tensors involved :-
```
Sorts a variadic number of tensors in inputs together, according to a custom comparator, along the given dimension and produces a variadic number of tensors as results
```
By the sound of it, it seemed like the `comparator` being used will be applied on both the input tensor and indices tensor.
Typically the C++ bias of using custom comparator functions on homogenous containers played in here.

I hope I was able communicate the err in my understanding.

I'd start off with the TMTensors' implementation in that case and keep you posted on this thread.

ramiro050:
Yeah, the description for `lianlg_ext.sort` is not very descriptive. There is more info here: https://www.tensorflow.org/xla/operation_semantics#sort. In particular, the `n > 1` section describes how arguments are passed to the comparator when there is more than one tensor.

There is also an example of the actual MLIR where `linalg_ext.sort` takes multiple tensor arguments here:

https://github.com/openxla/iree/blob/d474394200ee504c394818b9661db5656c2472b5/llvm-external-projects/iree-dialects/test/Dialect/iree_linalg_ext/tiling.mlir#L413

Happy to help with any issues that arise along the way.

Abhishek-Varma:
> Yeah, the description for `lianlg_ext.sort` is not very descriptive. There is more info here: https://www.tensorflow.org/xla/operation_semantics#sort. In particular, the `n > 1` section describes how arguments are passed to the comparator when there is more than one tensor.
> 
> There is also an example of the actual MLIR where `linalg_ext.sort` takes multiple tensor arguments here:
> 
> https://github.com/openxla/iree/blob/d474394200ee504c394818b9661db5656c2472b5/llvm-external-projects/iree-dialects/test/Dialect/iree_linalg_ext/tiling.mlir#L413
> 
> Happy to help with any issues that arise along the way.

Thanks @ramiro050 !

I've started off with the implementation already and made some speedy progress in this PR with a recent `push`.

Next up is "forming" the indices tensor since `aten.sort` will provide us only with the input.

Just looked up `tensor.generate`, which I believe might be a potential candidate to form such a tensor (off of input tensor's shape).

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -242,6 +242,62 @@ static Value createTMTensorScanOp(
   return scanOp->getResult(0);
 }
 
+// Utility function to create a TMTensor::SortOp. Returns `true` if success,
+// `failure` otherwise.
+bool createTMTensorSortOp(PatternRewriter *rewriter, const Location &sortOploc,
+                          const llvm::ArrayRef<Value> &operands,
+                          const llvm::ArrayRef<Type> &elementTypes,
+                          int64_t dimension, bool isStable, bool isDescending,
+                          TMTensor::SortOp &sortOp) {
```
ramiro050, line 251: The convention here is to pass `PatternRewriter` as a reference, and all the others by value.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -3385,6 +3385,82 @@ def SortIntListReverse_basic(module, tu: TestUtils):
...
+
+class SortTensorNegativeDimension(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([3, 4], torch.float32, True)
```
ramiro050, line 3452: Does it not work with dynamic sizes?

ramiro050 (changes requested):
Some initial comments. I will finish the review later today

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
```diff
@@ -476,6 +476,172 @@ LogicalResult ScatterOp::generateScalarImplementation(OpBuilder &b,
   return success();
 }
 
+//===----------------------------------------------------------------------===//
```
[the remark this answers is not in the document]
Abhishek-Varma: No, they're same other than `payloadUsesValueFromOperand` like rest of the Ops from what I observed in the file.

test/Conversion/TorchToTMTensor/basic.mlir
```diff
@@ -0,0 +1,29 @@
...
+// CHECK:           %[[SORT:.*]]:2 = tm_tensor.sort dimension(1) outs(%[[INPUT_TENSOR]], %[[INDICES_TENSOR]] : tensor<3x4xf32>, tensor<3x4xi64>) {
+// CHECK:                                   ^bb0(%[[INP1:.*]]: f32, %[[INP2:.*]]: f32, %[[IND1:.*]]: i64, %[[IND2:.*]]: i64):
+// CHECK:                                      %[[PREDICATE:.*]] = arith.cmpf ole, %[[INP1]], %[[INP2]] : f32
+// CHECK:                                      tm_tensor.yield %[[PREDICATE]] : i1
+// CHECK:                            } -> tensor<3x4xf32>, tensor<3x4xi64>
+// CHECK:           %[[OUTPUT_SORTED_TENSOR:.*]] = torch_c.from_builtin_tensor %[[SORT]]#0 : tensor<3x4xf32> -> !torch.vtensor<[3,4],f32>
+// CHECK:           %[[OUTPUT_INDICES_TENSOR:.*]] = torch_c.from_builtin_tensor %[[SORT]]#1 : tensor<3x4xi64> -> !torch.vtensor<[3,4],si64>
+// CHECK:           return %[[OUTPUT_SORTED_TENSOR]], %[[OUTPUT_INDICES_TENSOR]] : !torch.vtensor<[3,4],f32>, !torch.vtensor<[3,4],si64>
+// CHECK:        }
+func.func @torch.aten.sort(%arg0: !torch.vtensor<[3,4],f32>) -> (!torch.vtensor<[3,4],f32>, !torch.vtensor<[3,4],si64>) {
```
ramiro050, line 24: No need to have this test. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

test/Dialect/Torch/refine-types-ops.mlir
```diff
@@ -289,6 +289,21 @@ func.func @torch.aten.Matmul.Broadcast.Matrix(%arg0: !torch.vtensor<*,f32>, %arg
...
+// CHECK-LABEL:   func.func @torch.aten.sort(
+// CHECK-SAME:                               %[[INPUT:.*]]: !torch.vtensor<*,f32>,
+// CHECK-SAME:                               %[[DIM:.*]]: !torch.int,
+// CHECK-SAME:                               %[[DESCENDING:.*]]: !torch.bool) -> (!torch.vtensor, !torch.vtensor) {
+// CHECK:           %[[VALUES:.*]], %[[INDICES:.*]] = torch.aten.sort %[[INPUT]], %[[DIM]], %[[DESCENDING]] : !torch.vtensor<*,f32>, !torch.int, !torch.bool -> !torch.vtensor<*,f32>, !torch.vtensor<*,si64>
+// CHECK:           %[[VALUES_CAST:.*]] = torch.tensor_static_info_cast %[[VALUES]] : !torch.vtensor<*,f32> to !torch.vtensor
+// CHECK:           %[[INDICES_CAST:.*]] = torch.tensor_static_info_cast %[[INDICES]] : !torch.vtensor<*,si64> to !torch.vtensor
+// CHECK:           return %[[VALUES_CAST]], %[[INDICES_CAST]] : !torch.vtensor, !torch.vtensor
+// CHECK:         }
+func.func @torch.aten.sort(%input: !torch.vtensor<*,f32>, %dim: !torch.int, %descending: !torch.bool) -> (!torch.vtensor, !torch.vtensor) {
```
ramiro050, line 302: nit: no need to have this test

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -242,6 +242,64 @@ static Value createTMTensorScanOp(
   return scanOp->getResult(0);
 }
 
+// Utility function to create a TMTensor::SortOp.
+LogicalResult createTMTensorSortOp(PatternRewriter &rewriter,
+                                   Location sortOploc,
```
ramiro050, line 247: nit: use camel-case: `sortOpLoc`

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1297,6 +1355,95 @@ class ConvertAtenScatterReduceTwoOp
...
+        dynDims.push_back(rewriter.create<tensor::DimOp>(loc, inputTensor, i));
+      }
+    }
+    Value initEmptyTensor = rewriter.create<tensor::EmptyOp>(
+        loc, inputType.getShape(), rewriter.getI64Type(), dynDims);
+    Value zeroConst = rewriter.create<arith::ConstantOp>(
+        loc, rewriter.getZeroAttr(rewriter.getI64Type()));
+    Value outVector =
+        rewriter.create<linalg::FillOp>(loc, zeroConst, initEmptyTensor)
+            .getResult(0);
```
ramiro050, line 1407: No need to zero out this tensor, since the values are not used in the `linalg.generic` computation

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1297,6 +1355,95 @@ class ConvertAtenScatterReduceTwoOp
...
+    operands.push_back(inputTensor);
+    operands.push_back(indicesTensor);
+    SmallVector<Type> elementTypes;
+    elementTypes.push_back(inputType.getElementType());
+    elementTypes.push_back(indicesType.getElementType());
+
+    TMTensor::SortOp sortOp = NULL;
+    LogicalResult logicalResult =
+        createTMTensorSortOp(rewriter, loc, operands, elementTypes,
+                             /*dimension=*/dim, /*isStable=*/true,
```
ramiro050, line 1436: The documentation says that the default value for `isStable` is `False`: https://pytorch.org/docs/stable/generated/torch.sort.html

Can you add the last example in the documentation as an e2e test to make sure we are setting the right default value here?

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1297,6 +1355,95 @@ class ConvertAtenScatterReduceTwoOp
...
+
+    // Step 6. Create TMTensor::SortOp.
+    SmallVector<Value> operands;
+    operands.push_back(inputTensor);
+    operands.push_back(indicesTensor);
+    SmallVector<Type> elementTypes;
+    elementTypes.push_back(inputType.getElementType());
+    elementTypes.push_back(indicesType.getElementType());
+
+    TMTensor::SortOp sortOp = NULL;
```
ramiro050, line 1433: nit: It is safer to have `createTMTensorSortOp` create the entire op and return the result rather than populate the result into a reference. This also makes the behavior consistent with what `rewriter.create<...>(...)` does.

I would remove the `sortOp` argument from the helper function and change the result type to `FailureOr<Value>` to allow you to return the result from the sortOp.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -242,6 +242,64 @@ static Value createTMTensorScanOp(
...
+  SmallVector<Type> sortResultTypes;
+  for (Value val : operands) {
+    sortResultTypes.push_back(val.getType());
+  }
+  ValueRange inputs;
+  sortOp = rewriter.create<TMTensor::SortOp>(
+      sortOploc, sortResultTypes, inputs, operands,
+      rewriter.getI64IntegerAttr(dimension));
+
+  OpBuilder::InsertionGuard insertionPointGurad(rewriter);
```
ramiro050, line 263: not used

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -242,6 +242,64 @@ static Value createTMTensorScanOp(
...
+  Block *block = rewriter.createBlock(body);
+  Location loc = body->getLoc();
+  for (Type elementType : elementTypes) {
+    block->addArguments({elementType, elementType},
+                        SmallVector<Location, 2>(2, loc));
+  }
+
+  // Step 3. Create comparison op which will be used as the sorting predicate.
+  Value compareOp = NULL;
+  if (auto intType = elementTypes[0].dyn_cast<mlir::IntegerType>()) {
```
ramiro050, line 276: can you also add e2e test for integer input?

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -242,6 +242,64 @@ static Value createTMTensorScanOp(
...
+  Region *body = &sortOp.getRegion();
+  Block *block = rewriter.createBlock(body);
+  Location loc = body->getLoc();
+  for (Type elementType : elementTypes) {
+    block->addArguments({elementType, elementType},
+                        SmallVector<Location, 2>(2, loc));
+  }
+
+  // Step 3. Create comparison op which will be used as the sorting predicate.
+  Value compareOp = NULL;
```
ramiro050, line 275: nit: I would either use `nullptr` or simply `Value compareOp;`

ramiro050 (changes requested):
Looking good, just a few comments

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
Abhishek-Varma: That's correct but the `aten.sort` op's definition generated within `GeneratedTorchOps.td` doesn't seem to bear the `isStable` argument.

Let me know what needs to be done here. I used the `JITOperatorRegistryDump.txt` to find the op's definition.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -242,6 +242,63 @@ static Value createTMTensorScanOp(
...
+  } else {
+    return rewriter.notifyMatchFailure(
+        sortOpLoc, "Only Integer and Floating element type expected.");
+  }
+
+  // Step 4. Create yield op for yielding the sorting predicate.
+  rewriter.create<TMTensor::YieldOp>(loc, compareOp);
+  SmallVector<Value, 2> sortOpValues;
+  sortOpValues.push_back(sortOp.getResult(0));
+  sortOpValues.push_back(sortOp.getResult(1));
```
ramiro050, line 298: The function `createTMTensorSortOp` does not limit the number of inputs, so the number of outputs should also not be limited. The result type should change to `FailureOr<SmallVector<Value>>` and here I would just do `return sortOp.getResults()` (you might need to add a `SmallVector` constructor)

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -3385,6 +3385,100 @@ def SortIntListReverse_basic(module, tu: TestUtils):
...
+        None,
+        ([-1, -1], torch.int64, True)
+    ])
+    def forward(self, input):
+        return torch.sort(input)
+
+
+@register_test_case(module_factory=lambda: SortTensorInteger())
+def SortTensorInteger_basic(module, tu: TestUtils):
+    module.forward(torch.tensor([[1, 3, 2], [3, 1, 2]]))
```
ramiro050, line 3423: nit: int tensors can also be random: `tu.randint(2, 3)`

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
ramiro050: Yeah, the op you're adding does not take `stable` as an argument and uses the default value. If you do specify `stable` in PyTorch, the op generated is `aten.sort.stable` rather than `aten.sort`. Given that the default value of PyTorch is `stable=False`, we should set it to `false` here too (I would add a comment mentioning that this is the default value and link to the docs).

For example, when I run:

```python
In [8]: x = torch.tensor([0, 1] * 9)

In [9]: torch.ops.aten.sort(x)
Out[9]:
(tensor([0, 0, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1]),
 tensor([ 2, 16,  4,  6, 14,  8,  0, 10, 12,  9, 17, 15, 13, 11,  7,  5,  3,  1]))
```

I get the unstable result.

ramiro050 (changes requested):

ramiro050 (approved):
Great!

PR #2004, opened 2023-04-05
[TorchDynamo] Add aten.squeeze op to the decomposition list

vivekkhandelwal1:

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -97,6 +97,10 @@ TORCHDYNAMO_XFAIL_SET = {
     # ERROR: RuntimeError: Found a custom (non-ATen) operator that either mutates or its inputs: prims::squeeze.. Getting these operators to work with functionalization requires some extra work. For mutable ops you need to register a corresponding out-of-place variant of the op, and you also need to register a Functionalization kernel that performs some boilerplate, telling functionalization to map from the mutable op to the out-of-place op. See a more complete example of how to do this at https://gist.github.com/bdhirsh/7dadbf6296f8f7d1abcf4c482f438aaa.
     "PrimsSqueezeModule_basic",
     "PrimsSqueezeEmptyDimensionsModule_basic",
+
+    # ERROR: Lowering for op `prims.view_of` missing.
+    "SqueezeDimModule_identity",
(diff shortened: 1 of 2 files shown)


pashu123 (approved):

e2e_testing/xfail_sets.py
```diff
@@ -97,6 +97,10 @@
     # ERROR: RuntimeError: Found a custom (non-ATen) operator that either mutates or its inputs: prims::squeeze.. Getting these operators to work with functionalization requires some extra work. For mutable ops you need to register a corresponding out-of-place variant of the op, and you also need to register a Functionalization kernel that performs some boilerplate, telling functionalization to map from the mutable op to the out-of-place op. See a more complete example of how to do this at https://gist.github.com/bdhirsh/7dadbf6296f8f7d1abcf4c482f438aaa.
     "PrimsSqueezeModule_basic",
     "PrimsSqueezeEmptyDimensionsModule_basic",
+
+    # ERROR: Lowering for op `prims.view_of` missing.
+    "SqueezeDimModule_identity",
```
ramiro050, line 102: @vivekkhandelwal1, can you revert this PR? Changes that make previously passing tests fail should not get merged.

vivekkhandelwal1: They were failing because of changes in upstream PyTorch decomposition not any local changes in the Torch-MLIR. It requires an op lowering to be added which wasn't a trivial task. Anyways, I will add that op lowering.

ramiro050: If `aten.squeeze` is not added to the decomposition table, does that lead to more failures? I'm trying to understand the purpose of adding it in this PR rather than just XFAILing the broken tests.

To make this types of changes more clear in the future, the PR message should contain a nice description of the issue. Moreover, if something has to be marked as failing, an issue should also be opened explaining the problem, and the PR should reference the issue.

vivekkhandelwal1: Actually, I needed that decomp for a model. And as a result of adding that decomp, and upstream changes in squeeze op decomposition those tests failed. Anyways, I have raised a PR for that missing op' lowering. Thanks for your valuable piece of advice. I will surely take care of this in future PRs.

PR #2005, opened 2023-04-06
Add support for negative and out-of-bound indices for the slice+copy_ pattern

gpetters94:


gpetters94:
This is currently failing due to some `torch.constant.int` ops floating around longer than they should. I think it might be due to the placement of the Recompose pass in the pipeline but I'd like a second opinion.

ramiro050:
The issue is that there is currently no support for

```
  %20 = torch.prim.min.int %int2, %18 : !torch.int, !torch.int -> !torch.int
  %21 = torch.aten.neg.int %18 : !torch.int -> !torch.int
  %22 = torch.prim.max.int %20, %21 : !torch.int, !torch.int -> !torch.int
```

You would need to add those ops.

AmosLewis:
> The issue is that there is currently no support for
> 
> ```
>   %20 = torch.prim.min.int %int2, %18 : !torch.int, !torch.int -> !torch.int
>   %21 = torch.aten.neg.int %18 : !torch.int -> !torch.int
>   %22 = torch.prim.max.int %20, %21 : !torch.int, !torch.int -> !torch.int
> ```
> 
> You would need to add those ops.

That would be plenty of work.
I suggest we just add the INTMAX instead and add a TODO there for later fixing. Otherwise, otherwise I will just be stuck here.

PR #2006, opened 2023-04-06
Add Shape inference for CopyOp for lazy tensor core backend

rahuls-cerebras:
This PR adds the shape inference loging for CopyOp for lazy tensor core backend


rahuls-cerebras:
Gentle Reminder @ramiro050

GlebKazantaev:
Just to add more context. `aten.copy` and `aten.copy_` ops are not supported by reference TS LTC backend, so they doesn't work out-of-the box. But as they have CPU fallback they successfully compute these ops which is not the case for out backend. In order to support full codegen for these ops we need this extra shape inference function `compute_shape_copy` ([which is missing in PyTorch](https://github.com/pytorch/pytorch/blob/master/torch/csrc/lazy/core/shape_inference.h)). Ideally all these shape inference functions from `python/torch_mlir/csrc/base_lazy_backend/shape_inference.cpp` should be upstreamed to PyTorch, but as LTC flow is still under development (for example, in XLA LTC backend they [use partial codegen](https://github.com/pytorch/xla/blob/f763d0d2a4dac0872e2e1670332fd490f524b8be/xla_native_functions.yaml#L162-L163) for `aten.copy` ops and  manually lower them https://github.com/pytorch/xla/blob/f763d0d2a4dac0872e2e1670332fd490f524b8be/torch_xla/csrc/tensor_methods.cpp#L2172-L2180) so we decided to wait sometime until things stabilize.

GlebKazantaev (approved):

GlebKazantaev:
@rahuls-cerebras it seems like we fixed some of the tests:
image
could you please update them?

rahuls-cerebras:
> @rahuls-cerebras it seems like we fixed some of the tests: image could you please update them?

fixed

rahuls-cerebras:
@ramiro050 can you please give me appropriate permission to merge the PR

ramiro050:
> rahuls-cerebras

Done! You should've gotten an email

PR #2007, opened 2023-04-06
Revert "CI: drop deletion of workspace and limit submodule fetch concurrency"

powderluv:
Reverts llvm/torch-mlir#1921 since we still need the sudo rm to clean up after nightly builds.


ashay (approved):

powderluv:
btw we can use fetch-depth for the checkout.

PR #2009, opened 2023-04-06
Set fetch-depth 0

powderluv:
This is to potentially workaround the index.lock issue in git when we checkout new depth 1 submodules of recently updated mhlo.


ashay (approved):
LGTM!  My only concern is whether this will extend the checkout time significantly, but if this is only the Release build (which runs once every day), then the impact doesn't seem so bad.

powderluv:
yeah we can time it and see what it takes too.

PR #2013, opened 2023-04-07
[MLIR][TORCH] Patch up Ops and their lowerings to deal with +ve `dim`

Abhishek-Varma:
-- In Python we have the concept of negative dimension indexing.
-- We would want to normalize such dimensions to be +ve and within the expected range instead.
-- This commit takes care of a few remaining set of Ops and their lowerings by applying `toPositiveDim` and `isValidDim` to the extracted integer `dim` value.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -213,12 +213,16 @@ class DecomposeAtenAmaxOp : public OpRewritePattern<AtenAmaxOp> {
       return rewriter.notifyMatchFailure(
           op, "Expected a constant boolean value for keepDim");
 
-    Value input = op.getSelf();
+    Value input = op.getSelf();
```
ramiro050, line 216: nit: remove extra whitespace at end of line

ramiro050 (approved):
This is great, thanks for fixing these!

PR #2014, opened 2023-04-07
Fix a few missing dependencies.

arames:
`TorchToTMTensor` depends on `TorchMLIRTorchUtils` for `mlir::torch::torch_upstream::get_reduction_enum`.

`TorchMLIRTorchConversionPasses` depends on multiple libs for both tblgen'd headers and definitions. Test with `ninja TorchMLIRTorchConversionPasses` from a clean build.


arames:
It looks like there are a few more missing dependencies to fix. Update incoming.

ramiro050 (approved):
Thanks!

PR #2016, opened 2023-04-10
[MLIR][TORCH] Add support for prims::view_of op

vivekkhandelwal1:
This op does nothing and just returns the input operand as the result of the op.


ramiro050 (approved):
Thanks for the quick follow up

PR #2021, opened 2023-04-10
[Torch Dialect] fold aten.detach

qingyunqu:
fold aten.detach, it seems a useless op for torch-mlir's backends.


ramiro050 (approved):
The CI is failing because a test that was previously failing is now passing, so you need to change the `xfail_sets.py` file

qingyunqu:
@ramiro050 I just have one thing concerned. Is there any possible that `torch-mlir` will do `torch.autograd` on Torch Dialect? If it will, folding `aten.detach` will cause something wrong.  
BTW, in our training practice, we will use `torch._dynamo` + `torch.fx` to capture a `forward` or `backward` graph.

ramiro050:
> @ramiro050 I just have one thing concerned. Is there any possible that `torch-mlir` will do `torch.autograd` on Torch Dialect? If it will, folding `aten.detach` will cause something wrong. BTW, in our training practice, we will use `torch._dynamo` + `torch.fx` to capture a `forward` or `backward` graph. So in our practice, `aten.detach` looks like a useless op.

Torch-MLIR currently does not keep track of gradients of tensors, so `aten.detach` is a no-op. In terms of long term plans for Torch-MLIR, the plan is to align with torchdynamo and have that be the main form of input into Torch-MLIR. This means that the expectation will be for users to do what you're doing: use torch.fx to capture the forward and backward passes as separate graphs.

qingyunqu:
> > @ramiro050 I just have one thing concerned. Is there any possible that `torch-mlir` will do `torch.autograd` on Torch Dialect? If it will, folding `aten.detach` will cause something wrong. BTW, in our training practice, we will use `torch._dynamo` + `torch.fx` to capture a `forward` or `backward` graph. So in our practice, `aten.detach` looks like a useless op.
> 
> Torch-MLIR currently does not keep track of gradients of tensors, so `aten.detach` is a no-op. In terms of long term plans for Torch-MLIR, the plan is to align with torchdynamo and have that be the main form of input into Torch-MLIR. This means that the expectation will be for users to do what you're doing: use torch.fx to capture the forward and backward passes as separate graphs.

Yeah, I agree with you. Could you merge this PR to `main` branch? Thanks!

PR #2026, opened 2023-04-11
[StableHlo] Fix transposed convolution conversion

zhekunz2:
Convolution transpose with output padding case was padding zeros, which is not the correct behavior (compared to pytorch). 

The upstream shape function for ConvTranspose with output padding is incorrect for now: https://github.com/pytorch/pytorch/issues/98129

Once that solved, ConvTranspose test cases with static shape and output padding should be able to pass


ZihengJiang (approved):
Nice catch! Thanks for the work

CaoE:
https://github.com/pytorch/pytorch/pull/102139

PR #2027, opened 2023-04-12
Add TMTensor::Attention and lower scaled_dot_product_attention to it

gpetters94:
This adds an op in TMTensor, `AttentionOp` which is designed to be lowered one-to-one to backend ups (like IREE's [attention op](https://github.com/openxla/iree/blob/9466b5ed0f329df70a14c9242613fde226ae8c4e/llvm-external-projects/iree-dialects/test/Dialect/iree_linalg_ext/tile_and_decompose_attention.mlir#L3)) in the name of supporting Torch's `scaled_dot_product_attention`. This is my first time adding a new TMTensor op so it's pretty messy, but it's almost there. I'm making this a draft until I fix this last bug and write the lowering.

The bug in question is that `AttentionOp::getNumInputs()` and `AttentionOp::getNumOutputs()` aren't generating, despite them generating just fine for `ScanOp`. Could I get some guidance on that?


gpetters94:
Figured the bug out, I somehow deleted `generateScalarImplementation`.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorOps.td
```diff
@@ -192,6 +192,34 @@ def TMTensor_ScatterOp : TMTensor_Op<"scatter",
   }];
 }
 
+def TMTensor_AttentionOp : TMTensor_Op<"attention",
```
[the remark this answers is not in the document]
gpetters94: Not sure if I understand you. Do you just mean that it should match the signature of IREE's op? I don't think we'd be able to support anything but default values for the keyword arguments then. Pytorch's implementation of FlashAttention is given as 

```
scale_factor = 1 / math.sqrt(Q.size(-1)) if scale is None else scale
attn_mask = torch.ones(L, S, dtype=torch.bool).tril(diagonal=0) if is_causal else attn_mask
attn_mask = attn_mask.masked_fill(not attn_mask, -float('inf')) if attn_mask.dtype==torch.bool else attn_mask
attn_weight = torch.softmax((Q @ K.transpose(-2, -1) * scale_factor) + attn_mask, dim=-1)
attn_weight = torch.dropout(attn_weight, dropout_p)
return attn_weight @ V
```

but we just call the backend up with the key, query, and value then it'd be something like 

```
scale_factor = 1 / math.sqrt(Q.size(-1)) if scale is None else scale
attn_mask = torch.ones(L, S, dtype=torch.bool).tril(diagonal=0) if is_causal else attn_mask
attn_mask = attn_mask.masked_fill(not attn_mask, -float('inf')) if attn_mask.dtype==torch.bool else attn_mask
return iree_op(query, key, mask)
```

right? I'd think that the backend ops would have to support attention_masks and scaling as well unless we want to only support the default case.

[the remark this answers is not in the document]
pashu123: Yes, the op will have to evolve at the IREE side too. Currently, we can just bail out if we encounter the other parameters.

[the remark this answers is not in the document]
pashu123: Did you take a look at: https://github.com/openxla/iree/blob/0eeae4f712cf4df4ccfd29c671413f0b486020ea/llvm-external-projects/iree-dialects/lib/Dialect/LinalgExt/IR/LinalgExtOps.cpp#L2521 ?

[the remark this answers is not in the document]
gpetters94: Okay, I've copied over the IREE implementation and adjusted it so it'd compile. I'm getting the following error now though:

```
/home/gap/work/mlir-npcomp/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/basic.py:3548:15: error: 'tm_tensor.attention' op operand #0 must be shaped of any type values, but got '!torch.vtensor<[?,?,?,?],f32>'
        return torch.ops.aten.scaled_dot_product_attention(query, key, value)
              ^
/home/gap/work/mlir-npcomp/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/basic.py:3548:15: note: see current operation: %6 = "tm_tensor.attention"(%arg0, %arg1, %arg2) {operand_segment_sizes = array<i32: 2, 1>} : (!torch.vtensor<[?,?,?,?],f32>, !torch.vtensor<[?,?,?,?],f32>, !torch.vtensor<[?,?,?,?],f32>) -> !torch.vtensor<[?,?,?,?],f32>
// -----// IR Dump After ConvertTorchToTMTensor Failed (convert-torch-to-tmtensor) //----- //
"func.func"() ({
^bb0(%arg0: !torch.vtensor<[?,?,?,?],f32>, %arg1: !torch.vtensor<[?,?,?,?],f32>, %arg2: !torch.vtensor<[?,?,?,?],f32>):
  %0 = "torch_c.to_builtin_tensor"(%arg0) : (!torch.vtensor<[?,?,?,?],f32>) -> tensor<?x?x?x?xf32>
  %1 = "torch_c.to_builtin_tensor"(%arg1) : (!torch.vtensor<[?,?,?,?],f32>) -> tensor<?x?x?x?xf32>
  %2 = "torch_c.to_builtin_tensor"(%arg2) : (!torch.vtensor<[?,?,?,?],f32>) -> tensor<?x?x?x?xf32>
  %3 = "torch.constant.none"() : () -> !torch.none
  %4 = "torch.constant.float"() {value = 0.000000e+00 : f64} : () -> !torch.float
  %5 = "torch.constant.bool"() {value = false} : () -> !torch.bool
  %6 = "tm_tensor.attention"(%arg0, %arg1, %arg2) {operand_segment_sizes = array<i32: 2, 1>} : (!torch.vtensor<[?,?,?,?],f32>, !torch.vtensor<[?,?,?,?],f32>, !torch.vtensor<[?,?,?,?],f32>) -> !torch.vtensor<[?,?,?,?],f32>
  "func.return"(%6) : (!torch.vtensor<[?,?,?,?],f32>) -> ()
}) {function_type = (!torch.vtensor<[?,?,?,?],f32>, !torch.vtensor<[?,?,?,?],f32>, !torch.vtensor<[?,?,?,?],f32>) -> !torch.vtensor<[?,?,?,?],f32>, sym_name = "forward"} : () -> ()
```

Could this be an issue in `MaximizeValueSemantics`? I can't think of another reason it'd be complaining about a vtensor not being a shaped type.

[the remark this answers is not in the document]
gpetters94: Okay, figured that out. I was calling `op.getQuery` instead of `adaptor.getQuery`

gpetters94:
@ramiro050 Do we need an e2e test for the new TMTensor op if it's intended to be implemented directly by backends? I was discussing with @pashu123 and he suggested we add a decomposition for `scaled_dot_product_attention` and then people can choose to disable that decomposition if they want to use the TMTensor op for performance reasons.

gpetters94:
Okay, I've copied over the IREE loops implementation and rewritten it so it compiles, but now I'm getting a segfault here:

```
llvm::detail::PunnedPointer<llvm::ilist_node_base<true>*>::asInt (this=0x24d6dc61f7965e00) at /home/gap/work/mlir-npcomp/externals/llvm-project/llvm/include/llvm/ADT/PointerIntPair.h:41
41          std::memcpy(&R, Data, sizeof(R));
```

The backtrace is:

```
#0  llvm::detail::PunnedPointer<llvm::ilist_node_base<true>*>::asInt (this=0x688132af49e4b700) at /home/gap/work/mlir-npcomp/externals/llvm-project/llvm/include/llvm/ADT/PointerIntPair.h:41
#1  0x00007fffc39c8735 in llvm::detail::PunnedPointer<llvm::ilist_node_base<true>*>::operator long (this=0x688132af49e4b700)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/llvm/include/llvm/ADT/PointerIntPair.h:45
#2  0x00007fffc39c8b85 in llvm::PointerIntPair<llvm::ilist_node_base<true>*, 1u, unsigned int, llvm::PointerLikeTypeTraits<llvm::ilist_node_base<true>*>, llvm::PointerIntPairInfo<llvm::ilist_node_base<true>*, 1u, llvm::PointerLikeTypeTraits<llvm::ilist_node_base<true>*> > >::getInt (this=0x688132af49e4b700)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/llvm/include/llvm/ADT/PointerIntPair.h:96
#3  0x00007fffc39c8b55 in llvm::ilist_node_base<true>::isSentinel (this=0x688132af49e4b700) at /home/gap/work/mlir-npcomp/externals/llvm-project/llvm/include/llvm/ADT/ilist_node_base.h:45
#4  0x00007fffc39c8b15 in llvm::ilist_node_base<true>::isKnownSentinel (this=0x688132af49e4b700)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/llvm/include/llvm/ADT/ilist_node_base.h:46
#5  0x00007fffc39c929c in llvm::ilist_iterator<llvm::ilist_detail::node_options<mlir::Operation, true, false, void>, false, false>::operator* (this=0x7fffb5ff4d90)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/llvm/include/llvm/ADT/ilist_iterator.h:138
#6  0x00007fffccad264b in mlir::eraseUnreachableBlocks (rewriter=..., regions=...) at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/lib/Transforms/Utils/RegionUtils.cpp:97
#7  0x00007fffccad33e5 in mlir::simplifyRegions (rewriter=..., regions=...) at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/lib/Transforms/Utils/RegionUtils.cpp:732
#8  0x00007fffccac4ab7 in (anonymous namespace)::RegionPatternRewriteDriver::simplify() && (this=0x7fffb5ff5060)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/lib/Transforms/Utils/GreedyPatternRewriteDriver.cpp:463
#9  0x00007fffccac4711 in mlir::applyPatternsAndFoldGreedily (region=..., patterns=..., config=...)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/lib/Transforms/Utils/GreedyPatternRewriteDriver.cpp:487
#10 0x00007fffc3b105ec in mlir::applyPatternsAndFoldGreedily (op=0x55555accfd30, patterns=..., config=...)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/include/mlir/Transforms/GreedyPatternRewriteDriver.h:115
#11 0x00007fffca9116a1 in (anonymous namespace)::TMTensorToLoopsPass::runOnOperation (this=0x55555acced10)
    at /home/gap/work/mlir-npcomp/externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/Transforms/ConvertToLoops.cpp:113
#12 0x00007fffc3b8806b in mlir::detail::OpToOpPassAdaptor::run(mlir::Pass*, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int)::$_7::operator()() const (this=0x7fffb5ff5648)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/lib/Pass/Pass.cpp:482
#13 0x00007fffc3b88005 in llvm::function_ref<void ()>::callback_fn<mlir::detail::OpToOpPassAdaptor::run(mlir::Pass*, mlir::Operation*, mlir::AnalysisManager, bool, unsigned int)::$_7>(long)
    (callable=140736246797896) at /home/gap/work/mlir-npcomp/externals/llvm-project/llvm/include/llvm/ADT/STLFunctionalExtras.h:45
#14 0x00007fffc3b96d59 in llvm::function_ref<void ()>::operator()() const (this=0x7fffb5ff55c8)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/llvm/include/llvm/ADT/STLFunctionalExtras.h:68
#15 0x00007fffc3b8b7e2 in mlir::MLIRContext::executeAction<mlir::PassExecutionAction, mlir::Pass&>(llvm::function_ref<void ()>, llvm::ArrayRef<mlir::IRUnit>, mlir::Pass&) (
    this=0x555559f7f370, actionFn=..., irUnits=..., args=...) at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/include/mlir/IR/MLIRContext.h:275
#16 0x00007fffc3b8297d in mlir::detail::OpToOpPassAdaptor::run (pass=0x55555acced10, op=0x55555accfd30, am=..., verifyPasses=true, parentInitGeneration=1)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/lib/Pass/Pass.cpp:476
#17 0x00007fffc3b82fe5 in mlir::detail::OpToOpPassAdaptor::runPipeline (pm=..., op=0x55555accfd30, am=..., verifyPasses=true, parentInitGeneration=1, instrumentor=0x0, 
    parentInfo=0x7fffffffb318) at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/lib/Pass/Pass.cpp:548
#18 0x00007fffc3b895ce in mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::$_15::operator()(mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo&) const (
    this=0x7fffffffb2c0, opInfo=...) at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/lib/Pass/Pass.cpp:768
#19 0x00007fffc3b8a016 in mlir::failableParallelForEach<__gnu_cxx::__normal_iterator<mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo*, std::vector<mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo, std::allocator<mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo> > >, mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::$_15&>(mlir::MLIRContext*, __gnu_cxx::__normal_iterator<mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo*, std::vector<mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo, std::allocator<mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo> > >, __gnu_cxx::__normal_iterator<mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo*, std::vector<mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo, std::allocator<mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::OpPMInfo> > >, mlir::detail::OpToOpPassAdaptor::runOnOperationAsyncImpl(bool)::$_15&)::{lambda()#1}::operator()() const (this=0x55555ad59650)
    at /home/gap/work/mlir-npcomp/externals/llvm-project/mlir/include/mlir/IR/Threading.h:62
```

I don't have much experience with TMTensor, so I'm not really sure how to proceed from here. @MaheshRavishankar @vivekkhandelwal1 could you take a look and see if you can spot what's going wrong? You can reproduce the error by running `build/bin/torch-mlir-opt -pass-pipeline='builtin.module(func.func(refback-generalize-tensor-pad),func.func(linalg-fuse-elementwise-ops),func.func(scf-bufferize),func.func(tm-tensor-bufferize),func.func(empty-tensor-to-alloc-tensor),func.func(linalg-bufferize),func-bufferize,arith-bufferize,refback-mlprogram-bufferize,func.func(tensor-bufferize),func.func(finalizing-bufferize),func.func(buffer-deallocation),refback-munge-calling-conventions,func.func(tm-tensor-to-loops),func.func(refback-munge-memref-copy),func.func(convert-linalg-to-loops),func.func(lower-affine),convert-scf-to-cf,func.func(refback-expand-ops-for-llvm),func.func(arith-expand),func.func(convert-math-to-llvm),convert-math-to-libm,convert-linalg-to-llvm,expand-strided-metadata,finalize-memref-to-llvm,lower-affine,func.func(convert-arith-to-llvm),convert-func-to-llvm,convert-cf-to-llvm,reconcile-unrealized-casts)' ScaledDotProductAttentionModule_2.mlir` on [this](https://gist.github.com/gpetters94/ec199d1f85f879c66e652e3b051bc29c) mlir file.

ramiro050:
I'll take a look at the crash later today/tomorrow and let you know if I find anything. Is this patch the most up to date?

gpetters94:
> I'll take a look at the crash later today/tomorrow and let you know if I find anything. Is this patch the most up to date?

Yep, it should be. Thanks for taking a look.

ramiro050:
After playing around with things for a bit, I've come to the conclusion that the easiest way forward is to write our own loops implementation. Reusing the decomposition in IREE requires copying over more code and creating a new pass in `TMTensor`. And, even after doing this, things don't work in torch-mlir. There seems to be an issue with the decomposition that causes a crash when all the operands are dynamically sized. Moreover, when using static sizes, calls to `bufferization.clone` are inserted and these are not handled properly in the `RefBackend`, so we would need to make non-trivial changes to the `RefBackend` too.

Since the attention op is composed of a few simple ops, the loops implementation should require a lot less work than getting the IREE decomposition to work. I'm happy to help out with any loops dialect questions you might have.

ramiro050:
I've opened an issue for the crash: https://github.com/openxla/iree/issues/13368

gpetters94:
> After playing around with things for a bit, I've come to the conclusion that the easiest way forward is to write our own loops implementation. Reusing the decomposition in IREE requires copying over more code and creating a new pass in `TMTensor`. And, even after doing this, things don't work in torch-mlir. There seems to be an issue with the decomposition that causes a crash when all the operands are dynamically sized. Moreover, when using static sizes, calls to `bufferization.clone` are inserted and these are not handled properly in the `RefBackend`, so we would need to make non-trivial changes to the `RefBackend` too.
> 
> Since the attention op is composed of a few simple ops, the loops implementation should require a lot less work than getting the IREE decomposition to work. I'm happy to help out with any loops dialect questions you might have.

Alright, I'm most of the way through with writing my own loops implementation anyway, so that works for me. I'll let you know if I run into anything with it but it looks like it'll be simple enough.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -3695,6 +3695,30 @@ def forward(self, x):
 def MoveDimIntNegativeIndexModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(3, 4, 2))
 
+# ==============================================================================
+
+class ScaledDotProductAttentionModule(torch.nn.Module):
```
pashu123, line 3700: Please add a test with different shapes for query, key and value.

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
```diff
@@ -84,6 +84,254 @@ OpFoldResult TMTensor::getDim(OpBuilder &builder, Location loc, Value v,
...
+            loc, min, max, steps,
+            [&](OpBuilder &b, Location loc, ValueRange localIVs) {
+              SmallVector<Value> coords(localIVs);
+              coords.push_back(outerDim);
+              Value x = b.create<memref::LoadOp>(loc, weight, coords);
+              x = b.create<math::ExpOp>(loc, x);
+              b.create<memref::StoreOp>(loc, x, weight, coords);
+              b.create<scf::YieldOp>(loc);
+            });
+      });
```
ramiro050, line 274: What is the reason for using two `ParallelOp` here? I would've expected `exp(weight)` to be a single `ParallelOp` over the `weight` tensor.

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
```diff
@@ -84,6 +84,254 @@ OpFoldResult TMTensor::getDim(OpBuilder &builder, Location loc, Value v,
...
+               loc, elementType,
+               b.create<arith::IndexCastUIOp>(loc, b.getI32Type(),
+                                              queryDynSizes[queryRank - 1])));
+  // calculate max(weight)
+  Value init = b.create<memref::LoadOp>(
+      loc, weight, SmallVector<Value>(weightDynSizes.size(), zero));
+  Value globalMax =
+      b.create<scf::ParallelOp>(
+           loc, SmallVector<Value>(queryRank, zero), weightDynSizes,
+           SmallVector<Value>(queryRank, one), init,
```
ramiro050, line 233: There are a few instances in this pattern where the `weight` tensor is being iterated over, but the bounds are specified in terms of `queryRank`. I understand that `queryRank == weightRank`, but I think it would be easier to read if a variable `weightRank` was used instead. This makes it very clear that it is just an iteration over the `weight`

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
```diff
@@ -84,6 +84,254 @@ OpFoldResult TMTensor::getDim(OpBuilder &builder, Location loc, Value v,
...
+              Value x = b.create<memref::LoadOp>(loc, weight, coords);
+              x = b.create<arith::DivFOp>(loc, x, sum);
+              b.create<memref::StoreOp>(loc, x, weight, coords);
+              b.create<scf::YieldOp>(loc);
+            });
+      });
+
+  // output = weight @ value
+  matmul(b, loc, weight, weightDynSizes, value, valueDynSizes, output,
+         outputDynSizes, false);
```
ramiro050, line 330: nit: add `/*transpose=*/false`

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
```diff
@@ -84,6 +84,254 @@ OpFoldResult TMTensor::getDim(OpBuilder &builder, Location loc, Value v,
...
+            loc, zero, dim, one,
+            [&](OpBuilder &b, Location loc, ValueRange outerDim) {
+              SmallVector<Value> coords(localIVs);
+              coords.push_back(outerDim[0]);
+              Value x = b.create<memref::LoadOp>(loc, weight, coords);
+              x = b.create<arith::DivFOp>(loc, x, sum);
+              b.create<memref::StoreOp>(loc, x, weight, coords);
+              b.create<scf::YieldOp>(loc);
+            });
+      });
```
ramiro050, line 326: Similar comment here: this is an elementwise operation on `weight` so a single `ParallelOp` should be enough.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1494,6 +1497,70 @@ class ConvertAtenCumsumOp : public OpConversionPattern<AtenCumsumOp> {
...
+                                        elementType);
+
+    // Overwrite with tm_tensor::attention
+    auto attention = rewriter.create<AttentionOp>(
+        op.getLoc(), outType,
+        SmallVector<Value>{adaptor.getQuery(), adaptor.getKey(),
+                           adaptor.getValue()},
+        SmallVector<Value>{output});
+    auto segmentLengthAttr = rewriter.getDenseI32ArrayAttr({3, 1});
+    attention->setAttr("operand_segment_sizes", segmentLengthAttr);
```
ramiro050, line 1555: Why is this needed?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -551,6 +551,11 @@ def aten〇flatten〇using_ints〡shape(self: List[int], start_dim: int = 0, end
 def aten〇linear〡shape(input: List[int], weight: List[int], bias: Optional[List[int]] = None) -> List[int]:
     return upstream_shape_functions.linear(input, weight, bias)
 
+def aten〇scaled_dot_product_attention〡shape(query: List[int], key: List[int], value: List[int], attn_mask: Optional[List[int]] = None, dropout_p: float = 0., is_causal: bool = False, scale: Optional[float] = None) -> List[int]:
```
ramiro050, line 554: Since we are no longer upstreaming the shape functions, can you add `@check_shape_function` tests to it?

ramiro050 (changes requested):
Great work on this!

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR/TMTensorOps.cpp
gpetters94: I wrote it that way to avoid loading `sum` multiple times along the last dim, but if it's clearer this way then I'll rewrite it.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -691,6 +700,10 @@ def aten〇logical_xor〡shape(self: List[int], other: List[int]) -> List[int]:
 def aten〇logical_not〡shape(self: List[int]) -> List[int]:
     return upstream_shape_functions.unary(self)
 
+def aten〇scaled_dot_product_attention〡dtype(query_rank_dtype: Tuple[int, int], key_rank_dtype: Tuple[int, int], value_rank_dtype: Tuple[int, int], attn_mask_rank_dtype: Optional[Tuple[int, int]] = None, dropout_p: float = 0., is_causal: bool = False, scale: Optional[float] = None) -> int:
```
ramiro050, line 703: Now that all the ops are using dtype functions, all new dtype functions will need testing as well. I would do something like

```python
@check_dtype_function(_check_tensors_with_the_same_dtype(num_of_tensors=3) + [
Invocation(NonZeroDTensorWithDtype(torch.float32), NonZeroDTensorWithDtype(torch.bfloat16), NonZeroDTensorWithDtype(torch.float64)), # different width
Invocation(NonZeroDTensorWithDtype(torch.float32), NonZeroDTensorWithDtype(torch.int32), NonZeroDTensorWithDtype(torch.int64)) # different type
]) 
```

ramiro050 (approved):
Looks good, just one more small change

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
gpetters94: Got it, does this work?

ramiro050 (approved):

PR #2032, opened 2023-04-13
[TOSA] Add aten.abs support

AmosLewis:
https://github.com/nod-ai/SHARK/issues/986
[test_abs.mlir](https://gist.github.com/AmosLewis/e637465b23f6f96965487ad6e7e8f7b1#file-test_abs-mlir)
%149 = torch.aten.abs %137 : !torch.vtensor<[15,15],si64> -> !torch.vtensor<[15,15],si64>

### DIFF
--- a/test/Conversion/TorchToTosa/basic.mlir
+++ b/test/Conversion/TorchToTosa/basic.mlir
@@ -1085,6 +1085,19 @@ func.func @torch.aten.masked_fill.Tensor(%arg0: !torch.vtensor<[1,12,128,128],f3
 }
 
 // -----
+// CHECK-LABEL:   func.func @torch.aten.abs(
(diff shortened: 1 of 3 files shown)


test/Conversion/TorchToTosa/basic.mlir
```diff
@@ -1084,6 +1084,19 @@ func.func @torch.aten.masked_fill.Tensor(%arg0: !torch.vtensor<[1,12,128,128],f3
   return %0 : !torch.vtensor<[1,12,128,128],f32>
 }
 
+// -----
```
eric-k256, line 1087: Shouldn't we use an e2e test here instead of a lit test?

eric-k256 (commented):
You should be able to add the existing Abs test to the passing list for TOSA in xfail_sets.py and avoid needing to add a lit test.

eric-k256 (approved):
I'm not sure what's happening with the build failures, but looks good to me.

Issue #2025, opened 2023-04-11
 Test against the most recent stable version of PyTorch


ljfitz:
In the weekly meeting of 27 March we discussed adding a test of torch-mlir built and run with the most recent stable PyTorch version. The expectation was that the torchscript path is likely to work, while newer paths may not.
We're investigating adding this to testing by:

- adding `torch-version` to the [buildAndTest.yml matrix](https://github.com/llvm/torch-mlir/blob/72c3326097140726f7b31a75e62f40d5e3df86d2/.github/workflows/buildAndTest.yml#L27)
- renaming [pytorch-requirements.txt](https://github.com/llvm/torch-mlir/blob/72c3326097140726f7b31a75e62f40d5e3df86d2/https://github.com/llvm/torch-mlir/blob/72c3326097140726f7b31a75e62f40d5e3df86d2/pytorch-requirements.txt#L3#L3) to `pytorch-nightly-requirements.txt` and add `pytorch-stable-requirements.txt`
- passing the version by env. variable to scripts like [build_linux_packages.sh](https://github.com/llvm/torch-mlir/blob/72c3326097140726f7b31a75e62f40d5e3df86d2/.github/workflows/buildAndTest.yml#L101) and adjust to use the appropriate `pytorch-requirements.txt` file.

So far this seems to be successful at building and running tests, and we're starting to see what is failing - see https://github.com/llvm/torch-mlir/issues/2025#issuecomment-1517673883:
- :::: Run Lazy Tensor Core e2e integration tests:
  - Passed: 699
  - Expectedly Failed: 143
  - **Unexpectedly Passed: 24**
- :::: Run TorchDynamo e2e integration tests
  - Passed: 17
  - **Failed: 809**
  - Expectedly Failed: 40
- torch-mlir-dialects regression:
  - Failed Tests (1): TORCH_MLIR_PYTHON :: debug/lockstep_basic.py (TorchDynamo related)
- `update_abstract_interp_lib.sh` and `update_torch_ods.sh` mismatches

(Note: I see somewhat related discussion in https://github.com/llvm/torch-mlir/issues/1009, but the suggestion from the meeting to add an item to the test matrix seems new enough to make a separate issue.)

silvasean:
Hey Liam, thanks for doing this. Is there an actionable next step here? Do we want to add a non-blocking presubmit for this already and declare a burndown on the remaining issues?

maxbartel:
Hi!
Here is a summary of the failures:
```
[...]

PR #2038, opened 2023-04-17
[DRAFT] Add Stable PyTorch CI Pipeline

maxbartel:
As discussed in https://github.com/llvm/torch-mlir/issues/2025, this PR introduces CI runs that depend on the latest stable PyTorch release, not the nightly release. The goal of this PR is to see which tests fail and to gather feedback and opinions from the community. 

All tests are marked `true` to make sure all are executed, and the errors can be seen.

I disabled some pipelines to make development easier.


.github/workflows/buildAndTest.yml
```diff
@@ -38,8 +39,16 @@ jobs:
...
+          - os-arch: macos-arm64
+            torch-version: stable
           - os-arch: windows-x86_64
             llvm-build: out-of-tree
+          - os-arch: windows-x86_64
+            torch-version: stable
+          - os-arch: ubuntu-x86_64
+            llvm-build: out-of-tree
+          - os-arch: ubuntu-x86_64
+            torch-version: nightly
```
powderluv, line 51: Is this intended ? We will only build on stable if we exclude nightly right ?

build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -264,43 +267,57 @@ function _check_file_not_changed_by() {
 
 function test_in_tree() {
   echo ":::: Test in-tree"
-  cmake --build /main_checkout/torch-mlir/build --target check-torch-mlir-all
+  cmake --build /main_checkout/torch-mlir/build --target check-torch-mlir-all || true # TODO remove - here to see all potential failures
```
powderluv, line 270: We can just mark the top level CI Stable torch build as experimental so it will fail with these but still not hold up the builds.

build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -264,43 +267,57 @@ function _check_file_not_changed_by() {
...
-  python -m e2e_testing.main --config=tosa -v
+  python -m e2e_testing.main --config=tosa -v || true # TODO remove - here to see all potential failures
 
   echo ":::: Run Lazy Tensor Core e2e integration tests"
-  python -m e2e_testing.main --config=lazy_tensor_core -v
+  python -m e2e_testing.main --config=lazy_tensor_core -v || true # TODO remove - here to see all potential failures
 
   echo ":::: Run TorchDynamo e2e integration tests"
-  python -m e2e_testing.main --config=torchdynamo -v
+  python -m e2e_testing.main --config=torchdynamo -v || true # TODO remove - here to see all potential failures
```
powderluv, line 294: same for all || true.

build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -264,43 +267,57 @@ function _check_file_not_changed_by() {
...
   echo ":::: Setting up VENV with Python: $python_version"
   python3 -m venv /main_checkout/torch-mlir/docker_venv
   source /main_checkout/torch-mlir/docker_venv/bin/activate
 
   echo ":::: pip installing dependencies"
   python3 -m pip install --no-cache-dir -r /main_checkout/torch-mlir/externals/llvm-project/mlir/python/requirements.txt
-  python3 -m pip install --no-cache-dir -r /main_checkout/torch-mlir/requirements.txt
-
+  case $torch_version in
+    nightly)
```
powderluv, line 307: echo ":::: Using nightly dependencies"

utils/bazel/docker/Dockerfile
```diff
@@ -29,9 +29,10 @@ RUN wget -q https://github.com/bazelbuild/bazel/releases/download/${BAZEL_VERSIO
 # Install torch-mlir requirements
 COPY requirements.txt /opt/app/requirements.txt
 COPY build-requirements.txt /opt/app/build-requirements.txt
-COPY test-requirements.txt  /opt/app/test-requirements.txt
-COPY torchvision-requirements.txt  /opt/app/torchvision-requirements.txt
-COPY pytorch-requirements.txt /opt/app/pytorch-requirements.txt
+COPY test-nightly-requirements.txt  /opt/app/test-nightly-requirements.txt
+COPY torchvision-nightly-requirements.txt  /opt/app/torchvision-nightly-requirements.txt
+COPY pytorch-nightly-requirements.txt /opt/app/pytorch-nightly-requirements.txt
+COPY pytorch-stable-requirements.txt /opt/app/pytorch-stable-requirements.txt
```
powderluv, line 35: I wouldn't change these for now.

.github/workflows/buildAndTest.yml
maxbartel: By default, every combination of parameters is activated. We have to actively disable some of them. Here I removed some of the Linux runs to save build time. I will reactivate them to avoid confusion.

build_tools/python_deploy/build_linux_packages.sh
maxbartel: I would rather mark some of the pipelines as experimental then the whole CI. Things like the `linalg` or the `TOSA `backend should stay green even with a stable PyTorch version. What do you think?

utils/bazel/docker/Dockerfile
maxbartel: I will revert this change

build_tools/python_deploy/build_linux_packages.sh
maxbartel: I have added the option to mark tests as experimental in `e2e_testing.main`

powderluv:
@maxbartel  lmk when it ready to review. We can land and iterate on it.

maxbartel:
@powderluv I think it is ready for review now. Thank you!

mgehre-amd (approved):
LGTM!

.github/actions/setup-build/action.yml
```diff
@@ -9,6 +9,12 @@ inputs:
       but the content is irrelevant.
     required: false
     default: ''
+  torch-version:
+    description: |
+      Additional string to determine wether to test against a stable
+      torch release or against the nightly build
+    required: true
```
powderluv, line 16: nit: you can leave this as false so it is optional and defaults to nightly

.github/actions/setup-build/action.yml
```diff
@@ -26,13 +32,15 @@ runs:
...
-      python -m pip install -r pytorch-requirements.txt
+      python -m pip install -r pytorch-${{ inputs.torch-version }}-requirements.txt
       python -m pip install -r build-requirements.txt
     shell: bash
 
   - name: Install prerequisites (Linux)
     if: ${{ runner.os == 'Linux' }}
-    run: sudo apt-get install --yes ccache ninja-build
+    run: |
+      sudo apt-get update
```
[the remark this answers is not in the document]
powderluv: Can we remove this ? Doing a system update may not be ideal.

.github/actions/setup-build/action.yml
```diff
@@ -26,13 +32,15 @@ runs:
...
+      sudo apt-get install --yes ccache ninja-build
```
powderluv, line 43: these should already be in the builder lets remove it.

.github/workflows/buildAndTest.yml
```diff
@@ -98,6 +104,7 @@ jobs:
         TM_PACKAGES="${{ matrix.llvm-build }}" \
         TM_USE_PYTORCH_BINARY="${{ matrix.torch-binary }}" \
         TM_PYTORCH_INSTALL_WITHOUT_REBUILD="${{ steps.cache-pytorch.outputs.cache-hit }}" \
+        TORCH_VERSION="${{ matrix.torch-version }}" \
```
powderluv, line 107: nit: we should probably keep the variables with TM_ and TM_TORCH_VERSION for the variables passed into the .sh script

build-requirements.txt
```diff
@@ -4,6 +4,7 @@ wheel
 setuptools
 cmake
 ninja
+packaging
```
powderluv, line 7: I hit the same on the arm64 build PR, incase it lands before you can drop this.

build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -55,6 +55,8 @@ TM_USE_PYTORCH_BINARY="${TM_USE_PYTORCH_BINARY:-ON}"
 TM_SKIP_TESTS="${TM_SKIP_TESTS:-OFF}"
 # Update ODS and abstract interpretation library files
 TM_UPDATE_ODS_AND_ABSTRACT_INTERP_LIB="${TM_UPDATE_ODS_AND_ABSTRACT_INTERP_LIB:-OFF}"
+# Determine wether to use a stable or a nightly torch build
+TORCH_VERSION="${TORCH_VERSION:-nightly}"
```
powderluv, line 59: TM_TORCH_VERSION to be consistent.

python/torch_mlir_e2e_test/test_suite/__init__.py
```diff
@@ -13,6 +13,16 @@
...
 
+# check for torch version and disable tests
+TORCH_2_1_REQUIRED = {
+    "ScaledDotProductAttentionDifferentModule_basic",
+    "ScaledDotProductAttentionSameModule_basic"
+}
+import torch
+from packaging import version
+if version.parse(torch.__version__) < version.parse("2.1.0"):
+    COMMON_TORCH_MLIR_LOWERING_XFAILS.update(TORCH_2_1_REQUIRED)
```
powderluv, line 24: We should add a TODO: Delete once PyTorch 2.1.0 is marked stable.

test-requirements.txt
```diff
@@ -1,5 +1,3 @@
--r torchvision-requirements.txt
-
```
powderluv, line 2: Dont we still need it for tests ?

powderluv (changes requested):
Added some comments and feedback.

powderluv:
Also once the comments are approved lets push it a branch in llvm/torch-mlir and then we can run the "Release oneshot candidate" to make sure nightly builds are ok. 

@ashay fyi on the requirements.txt stuff

.github/workflows/buildAndTest.yml
```diff
@@ -75,6 +80,7 @@ jobs:
       uses: ./.github/actions/setup-build
       with:
         cache-suffix: 'build-${{ matrix.llvm-build }}'
```
ashay, line 82: I suggest adding `-${{ matrix.torch-version }}` to the cache suffix, because otherwise, we might start seeing build failures when a workflow tries to create a cache entry that already exists.

e2e_testing/main.py
```diff
@@ -72,6 +72,10 @@ def _get_argparse():
     parser.add_argument("--crashing_tests_to_not_attempt_to_run_and_a_bug_is_filed",
                         metavar="TEST", type=str, nargs="+",
                         help="A set of tests to not attempt to run, since they crash and cannot be XFAILed.")
+    parser.add_argument("--ignore_failures", 
+                        default=False,
+                        action="store_true",
+                        help="return exit code 0 even if the test fails to unblock pipeline")
```
ashay, line 78: I understand the utility of this flag to figure out which tests fail, but I wonder if we should drop it once this PR is ready to be merged.

requirements.txt
```diff
@@ -1,3 +1,4 @@
--r pytorch-requirements.txt
 -r build-requirements.txt
+-r pytorch-nightly-requirements.txt
+-r torchvision-nightly-requirements.txt
 -r test-requirements.txt
```
ashay, line 4: Nit: What do you think about (a separate PR for) moving all of the *-requirements.txt files to a directory?  I confess I added most of those files, but I am beginning to think that we have too many in the top-level directory.

ashay (commented):
My apologies for the much delayed response.  Thanks for making the necessary changes; I suggested a few small changes, but otherwise, this looks good.  Nice work!

.github/actions/setup-build/action.yml
maxbartel: ccache is still needed, so I will not fully remove it

test-requirements.txt
maxbartel: I moved it to the top level requirements file, because we also need these other packages here for the stable build. Do you have another idea how to do it?

.github/actions/setup-build/action.yml
powderluv: The CI runner already has it.

e2e_testing/main.py
maxbartel: Currently it is only used because we have some unexpected passes with the stable torch version. I think this should be replaced to something with finer granularity. If you are ok with it, I would like to do this in a follow up PR

maxbartel:
@powderluv @ashay I think it is ready for another round of reviews. https://github.com/llvm/torch-mlir/commit/9b7909b599b6f1cd3f45751e38d6a92d41c94bc2 did reduce the amount of code this PR has by quite a bit

build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -315,7 +332,7 @@ function setup_venv() {
       ;;
     stable)
       echo ":::: Using stable dependencies"
-      python3 -m pip install --no-cache-dir torch torchvision
+      python3 -m pip install --no-cache-dir torch==2.0.1 torchvision==0.15.2 --index-url https://download.pytorch.org/whl/cpu
```
mgehre-amd, line 335: Why did you need to add version numbers here?
If we drop the version numbers here, would it automatically test with the latest stable version?

powderluv: I would leave it unpinned testing on latest stable. (Future planned work may take a while so meanwhile let it test on latest stable than being pinned to 2.0.1).

python/torch_mlir/dynamo.py
```diff
@@ -62,7 +62,6 @@ def _get_decomposition_table():
         aten.native_group_norm_backward,
         aten.sigmoid_backward,
         aten._native_batch_norm_legit,
-        aten._native_batch_norm_legit_no_training,
```
powderluv, line 65: Hmm so we are disabling a nightly feature since it is broken on stable? Shouldn't this be guarded for stable ?

e2e_testing/main.py
ashay: Sounds good.

.github/actions/setup-build/action.yml
```diff
@@ -37,11 +43,6 @@ runs:
       python -m pip install -r build-requirements.txt
     shell: bash
 
-  - name: Install prerequisites (Linux)
```
mgehre-amd, line 40: Can we keep this in? With this change, the build fails on our fork of torch-mlir which uses standard github runners.

powderluv: Sounds good

test-requirements.txt
maxbartel: Ok, I will just call the file directly from` build_linux_packages.sh`, that should remove the dependency

powderluv: This sounds good but let's also make sure the regular CI doesn't require a similar direct install

python/torch_mlir/dynamo.py
```diff
@@ -62,9 +63,12 @@ def _get_decomposition_table():
...
         aten.squeeze,
-    ])
+    ]
+    # TODO: enable test once 2.1.0 is stable
+    if version.parse(torch.__version__) > version.parse("2.0.1+cpu"):
```
powderluv, line 69: Nit: I think we can avoid the +CPU here.

powderluv (approved):
Looking great. Minor Nit. 

Should be good once CI is green.

python/torch_mlir/dynamo.py
maxbartel: I tried it and it behaves differently from what one would expect. For example 
```
>>> version.parse("2.0.1") >= version.parse("2.0.1+cpu")
False
```
I would leave it there for now, if you are ok with it

powderluv: Sounds good.

maxbartel:
@powderluv The required pipelines are now different because of the nightly and stable inclusion. I think this is something that someone with more rights to the repository compared to me has to change.

maxbartel:
@powderluv @ashay @mgehre-amd The pipeline seems to be good. Thank you for all the reviews and help!

powderluv:
Ok this looks ok to land and iterate. I think we can possibly turn off some redundant CI runs in future (like out of tree stable ?)

PR #2039, opened 2023-04-17
[TOSA] lowering support for aten cat

yifei410:
Add support for lowering torch.aten.cat to tosa.concat


eric-k256 (commented):
Sorry for the delay, I was out of the office the last few days. TOSA-wise, it looks good to me. Thanks for contributing @yifei410!

yifei410:
Not sure why the test was cancelled, could not reproduce on my side.

eric-k256:
I've retriggered the failing job to see if it was a test glitch.

eric-k256:
To run the TOSA tests on their own, you will want to run `tools/e2e_test.sh -c tosa`. Before you run that, you need to run `build_tools/write_env_file.sh` to generate the .env file used by the e2e testing.
If you can reproduce locally, you can add a filter argument to the `e2e_test.sh` command line to run single tests.

lisaliu1:
@eric-k256 
We didn't change anything with regards to stable-hlo. But the e2e-test here shows stable-hlo e2e test error. See:
https://github.com/llvm/torch-mlir/actions/runs/4823054313/jobs/8591062630?pr=2039#step:8:6130
Can you help to check why this is happening?

eric-k256:
I'm not sure why the most recent run failed in torchdynamo. I posted in the torch-mlir discord to see if anyone else is seeing test failures like this.

lisaliu1:
Thanks @eric-k256 . When I ran the failed test locally, it passes, see below.

./tools/e2e_test.sh -c torchdynamo -f SelectIntNegativeDimAndIndexStaticModule_basic  -vv

Compiling SelectIntNegativeDimAndIndexStaticModule_basic...
Running SelectIntNegativeDimAndIndexStaticModule_basic...
PASS - "SelectIntNegativeDimAndIndexStaticModule_basic"

Summary:
    Passed: 1

eric-k256:
The dynamo tests fail randomly when run as a whole. If we retrigger the job, it may pass, but it may fail on a different test. A change is being added to the checks to try to avoid these failures. Once that is in, we might need to include that fix and then the tests should pass. 
There is an issue to track the problem: https://github.com/llvm/torch-mlir/issues/2050

lisaliu1:
That's great. Thanks a lot @eric-k256 for sharing this information. Let's wait until this fix is in and check again.

eric-k256:
I think the fixes are in for the testing. If you rebase your change onto the latest, the testing should pass. It depends on a newer version of the testlist, so I can't rerun the existing change to test.

yifei410:
@eric-k256 Hi Eric, could you help to start building the test?

yifei410:
the previous hanging issue was caused by missing the check if tensor type shape is static or not. It should be fixed in the latest commit.

eric-k256 (approved):
Looks good. I accidentally cancelled the workflow thinking that it did not have the latest commits. Then I realized my mistake and restarted the jobs.
Thanks for your patience, you hit a spot where things out of your control were happening.

PR #2040, opened 2023-04-17
[Dynamo] turn on `no_python=True` for dynamo tests

makslevental:
Per https://github.com/llvm/torch-mlir/pull/1996#discussion_r1169203933:

```
Summary:
    Passed: 762
    Failed: 59
    Expectedly Failed: 40
```

cc @powderluv

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -98,6 +98,147 @@ TORCHDYNAMO_XFAIL_SET = {
     # ERROR: RuntimeError: Found a custom (non-ATen) operator that either mutates or its inputs: prims::view_of.. Getting these operators to work with functionalization requires some extra work. For mutable ops you need to register a corresponding out-of-place variant of the op, and you also need to register a Functionalization kernel that performs some boilerplate, telling functionalization to map from the mutable op to the out-of-place op. See a more complete example of how to do this at https://gist.github.com/bdhirsh/7dadbf6296f8f7d1abcf4c482f438aaa.
     "PrimsViewOfModule_basic",
     "PrimsViewOfZeroRankModule_basic",
+
+    # See https://github.com/llvm/torch-mlir/pull/2040 and corresponding upstream issue
+    # https://github.com/pytorch/pytorch/issues/99752.
+    # torch._dynamo.exc.Unsupported: call_function BuiltinVariable(bool) [TensorVariable()] {}
+    'TensorToBoolZeroRank_basic',
+    'TensorToBool_basic',
+
+    # torch._dynamo.exc.Unsupported: call_function BuiltinVariable(float) [TensorVariable()] {}
+    'AtenSubFloatModule_basic',
+    'BoolFloatFalseModule_basic',
+    'BoolFloatTrueModule_basic',
+    'CeilFloatModule_basic',
+    'DivFloatModule_basic',
+    'GeFloatIntModule_basic',
+    'GeFloatModule_basic',
+    'GtFloatIntModule_basic',
+    'NeFloatIntModule_basic',
+    'SubFloatModule_basic',
+    'TensorToFloatZeroRank_basic',
+    'TensorToFloat_basic',
+
+    # torch._dynamo.exc.Unsupported: call_function BuiltinVariable(int) [TensorVariable()] {}
+    'AddIntModule_basic',
+    'AtenIntTensorCharDtypeModule_basic',
+    'BoolIntFalseModule_basic',
+    'BoolIntTrueModule_basic',
+    'DivIntModule_basic',
+    'EqIntModule_basic',
+    'GeIntModule_basic',
+    'GtIntModule_basic',
+    'MulIntModule_basic',
+    'NeIntModule_basic',
+    'SqrtIntModule_basic',
+    'SubIntModule_basic',
+    'TensorToIntZeroRank_basic',
+    'TensorToInt_basic',
+    'UnsafeViewCollapseDynamicWithAtenSizeIntModule_basic',
+    'ViewCollapseDynamicWithAtenSizeIntModule_basic',
+
+    # torch._dynamo.exc.Unsupported: call_method ListVariable() sort [] {'reverse': ConstantVariable(bool)}
+    'SortIntListReverse_basic',
+
+    # torch._dynamo.exc.Unsupported: call_method ListVariable() sort [] {}
+    'SortIntList_basic',
+
+    # torch._dynamo.exc.Unsupported: data dependent operator: aten._local_scalar_dense.default
+    'AtenFloatScalarModule_basic',
+    'AtenIntBoolOpModule_basic',
+    'OneHotModule_basic',
+    'QuantizedMLP_basic',
+    'ScalarImplicitFloatModule_basic',
+    'ScalarImplicitIntModule_basic',
+
+    # torch._dynamo.exc.Unsupported: dynamic shape operator: aten.bincount.default
+    'BincountMinlengthModule_basic',
+    'BincountModule_basic',
+    'BincountStaticSizeModule_basic',
+
+    # torch._dynamo.exc.Unsupported: torch.* op returned non-Tensor bool call_function aten.Bool
+    'BoolFloatConstantModule_basic',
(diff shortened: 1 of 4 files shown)


ramiro050:
This is interesting. It seems like the main source of errors is torchdynamo not knowing how to handle the builtin python functions `int, float, bool`, which is odd since each of those functions has an `aten` equivalent.

The second large set of errors has to do with torchdynamo currently not supporting graphs returning scalars.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -759,7 +759,7 @@ def forward(self, tensor, indices):
         return torch.gather(tensor, -1, indices)
 
 
-@register_test_case(module_factory=lambda: GatherNegativeDimModule())
+# @register_test_case(module_factory=lambda: GatherNegativeDimModule())
```
makslevental, line 762: segfault

python/torch_mlir_e2e_test/test_suite/index_select.py
```diff
@@ -45,7 +45,7 @@ def __init__(self):
     def forward(self, input, indices):
         return torch.index_select(input, -1, indices)
 
-@register_test_case(module_factory=lambda: IndexSelectNegativeDimModule())
+# @register_test_case(module_factory=lambda: IndexSelectNegativeDimModule())
```
makslevental, line 48: segfault

python/torch_mlir_e2e_test/test_suite/rng.py
```diff
@@ -343,7 +343,7 @@ def forward(self):
         return mean
 
 
-@register_test_case(module_factory=lambda: RandIntModule())
+# @register_test_case(module_factory=lambda: RandIntModule())
```
makslevental, line 346: segfault

python/torch_mlir_e2e_test/test_suite/rng.py
```diff
@@ -363,7 +363,7 @@ def forward(self):
         return mean
 
 
-@register_test_case(module_factory=lambda: RandIntDtypeModule())
+# @register_test_case(module_factory=lambda: RandIntDtypeModule())
```
makslevental, line 366: segfault

python/torch_mlir_e2e_test/test_suite/rng.py
```diff
@@ -383,7 +383,7 @@ def forward(self):
         return mean
 
 
-@register_test_case(module_factory=lambda: RandIntPinMemoryModule())
+# @register_test_case(module_factory=lambda: RandIntPinMemoryModule())
```
makslevental, line 386: segfault

ramiro050:
You can avoid running tests that segfault by adding them here 

https://github.com/llvm/torch-mlir/blob/6973abbfaa8a8faa50b9b883e8bad85cce6f2ff9/build_tools/python_deploy/build_linux_packages.sh#L291

using the flag

https://github.com/llvm/torch-mlir/blob/6973abbfaa8a8faa50b9b883e8bad85cce6f2ff9/e2e_testing/main.py#L65

As the flag suggests, we should have an issue tracking the reason for the failure.

python/torch_mlir_e2e_test/configs/torchdynamo.py
```diff
@@ -8,6 +8,10 @@
 import numpy
 import torch
 import torch._dynamo as dynamo
+import torch._dynamo.config
+
+torch._dynamo.config.cache_size_limit = 6400
```
ramiro050, line 13: Why is this needed?

makslevental: I don't know what the connection between `nopython=True` and their cache is but with this new settings that cache limit (of 64) is hit.

silvasean: I think the cache is always hit, even currently. We have one compilation per test case approximately which always overflows the cache.

makslevental: my issue must be because I ran with `-s` to identify the segfaults - will remove.

e2e_testing/xfail_sets.py
```diff
@@ -98,6 +98,106 @@
...
+
+    # torch._dynamo.exc.Unsupported: torch.* op returned non-Tensor int call_function prim.min
+    'PrimMinIntModule_basic',
+}
+
+TORCHDYNAMO_CRASHING_SET = {
+    "GatherNegativeDimModule_basic",
+    "IndexSelectNegativeDimModule_basic",
+    "RandIntModule_basic",
+    "RandIntDtypeModule_basic",
```
makslevental, line 200: still need to be investigated (i.e., TODO for this PR)

makslevental:
> This is interesting. It seems like the main source of errors is torchdynamo not knowing how to handle the builtin python functions `int, float, bool`, which is odd since each of those functions has an `aten` equivalent.
> 
> The second large set of errors has to do with torchdynamo currently not supporting graphs returning scalars.

see https://github.com/pytorch/pytorch/issues/99752

python/torch_mlir_e2e_test/configs/torchdynamo.py
```diff
@@ -45,12 +49,13 @@ def _refbackend_torchdynamo_backend(fx_graph: torch.fx.GraphModule,
     # Torch-MLIR does not support returning an empty tuple. The reason is
     # that both returning an empty tuple and returning `None` results in MLIR
     # functions that have as a return type `()`. In other words, there is no
-    # way of differentiating between the two. Moreover, since Torch-MLIR treats
-    # inputs as having value semantics, graphs that return nothing are no-ops to
-    # Torch-MLIR.
+    # way of differentiating between the two.
     if _returns_empty_tuple(fx_graph):
         return fx_graph
```
ramiro050, line 54: I think we should just remove this `if`. Currently it will result in an `assert` error saying that `_refbackend_torchdynamo_backend` did not run, but it would be more accurate to let `torch_mlir.compile` run and have `torch_mlir` create the "no empty returns allowed error"

ramiro050 (approved):
Just one small comment, but other than that LGTM

python/torch_mlir_e2e_test/configs/torchdynamo.py
makslevental: I put this back after our offline discussion because what actually happens is you get ~134 new fails that are all like

```
        torch._dynamo.exc.BackendCompilerFailed: backend='compiler_fn' raised:
        Exception: 
        PyTorch TorchScript module -> torch-mlir Object Graph IR import failed with:
        ### Importer C++ Exception:
        see diagnostics
        ### Importer Diagnostics:
        error: unhandled prim::Constant node: %12 : () = prim::Constant[value=()](
```

But failing on just empty return isn't right either because (possibly) dynamo broke the graph and this is just pathological/empty/useless side of the break (i.e., the next pass through, for the same test, will have useful work and will compile). 

Ultimately I don't know what the answer is here - you have to check whether the substantive part of the test actually ran/got compiled but you can't predict when dynamo will break and you can't know after the fact except in some cases (wherein it decides to percolate up by some means, raising an Exception of otherwise).

ramiro050: I see, that makes sense. In that case, I'm fine with living it the way it is

ramiro050: I tried this locally, and the reason that more failures are caused is that the importer error makes either torchdynamo or our testing framework trip up, and many of the tests that execute after end up failing with the same error. When I ran things, I got around 20 new fails, but when those tests are run individually, they pass.

What we can do is turn the empty graph check into an assert instead:

```python
assert not _returns_empty_tuple(fx_graph), "encountered graph that does not return anything"
```

When I did this locally, I did not get any new failing tests. I don't think we have to worry anymore about an empty graph followed by a non-empty graph, because `nopython` is supposed to guarantee just one graph per test. If this is not the case, we should open a bug upstream.

makslevental: TEAMWORK

ramiro050 (approved):

PR #2043, opened 2023-04-18
Fix creation of empty tensor in decomposition for randn ops

ramiro050:
The current decomposition for `aten.randn.generator` does not specify the `dtype` argument of the empty tensors created to store the random values. This leads to invalid IR when the output type of the `randn` op is not the default PyTorch dtype.


vivekkhandelwal1 (approved):
Thanks for fixing this!

PR #2044, opened 2023-04-19
Add complex types support with basic complex ops.

pashu123:
Add complex types support with basic complex types. Add aten.imag and aten.real op lowering via linalg_backend.


lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1330,6 +1331,86 @@ class ConvertAtenSliceScatterOp
...
+    Value constantOne =
+        getConstant(rewriter, loc, 1, mlir::IndexType::get(context));
+
+    AffineMap outputMap =
+        AffineMap::get(resultType.getRank(), 0, outputExpr, op->getContext());
+
+    SmallVector<AffineMap> indexingMaps{outputMap};
+    SmallVector<utils::IteratorType> iteratorTypes(
+        resultType.getRank(), utils::IteratorType::parallel);
+    auto transpose =
```
ramiro050, line 1382: nit: name of variable does not match operation

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -644,6 +645,32 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+  if (auto imag = dyn_cast<AtenImagOp>(op)) {
+    Type dtype = converter->convertType(imag.getType())
+                     .cast<RankedTensorType>()
+                     .getElementType();
+    if (!dtype.isa<mlir::FloatType>()) {
+      imag.emitError("unimplemented: non-floating point dtype");
+      return nullptr;
+    }
+    Value imagVal = b.create<complex::ImOp>(loc, payloadArgs[0]);
+    Value result = convertScalarToDtype(b, loc, imagVal, dtype);
```
ramiro050, line 657: Is this cast necessary?

lib/Dialect/Torch/IR/TorchTypes.cpp
```diff
@@ -404,6 +404,26 @@ static Type convertDtypeToBuiltinElementType(MLIRContext *context, Type dtype) {
...
+      return ComplexType::get(mlir::FloatType::getF16(context));
+    else if (floatType.getWidth() == 64)
+      return ComplexType::get(mlir::FloatType::getF32(context));
+    else if (floatType.getWidth() == 128)
+      return ComplexType::get(mlir::FloatType::getF64(context));
+    else {
+      emitError(UnknownLoc::get(context))
+          << "unimplemented: conversion of complex dtype" << dtype
+          << " to builtin tensor element type";
+      return nullptr;
```
ramiro050, line 425: nit: `else` branch not needed. The `emitError` below would catch these

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -3700,3 +3700,45 @@ def forward(self, x):
...
+
+class AtenImagModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1,-1], torch.float32, True),
```
ramiro050, line 3735: Does passing a `torch.complex` directly not work?

ramiro050 (changes requested):
Some initial comments

lib/Conversion/TorchToLinalg/DataMovement.cpp
pashu123: Done.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
pashu123: No, this is not necessary. Thanks for pointing out.

python/torch_mlir_e2e_test/test_suite/basic.py
ramiro050: What is the error?

ramiro050:
> @ramiro050 Error:
> 
> ```
>  <unknown>:0: error: failed to legalize operation 'builtin.unrealized_conversion_cast' that was explicitly marked illegal
> <unknown>:0: note: see current operation: %41 = "builtin.unrealized_conversion_cast"(%40) : (!llvm.struct<(f32, f32)>) -> complex<f32>
> ```
> 
> Here's the IR: https://gist.github.com/pashu123/8e2fce24870e3e896708b10c0e93c016
> 
> Pass:` torch-mlir-opt -pass-pipeline='builtin.module(func.func(refback-generalize-tensor-pad),func.func(linalg-fuse-elementwise-ops),func.func(scf-bufferize),func.func(tm-tensor-bufferize),func.func(empty-tensor-to-alloc-tensor),func.func(linalg-bufferize),func-bufferize,arith-bufferize,refback-mlprogram-bufferize,func.func(tensor-bufferize),func.func(finalizing-bufferize),func.func(buffer-deallocation),refback-munge-calling-conventions,func.func(tm-tensor-to-loops),func.func(refback-munge-memref-copy),func.func(convert-linalg-to-loops),func.func(lower-affine),convert-scf-to-cf,func.func(refback-expand-ops-for-llvm),func.func(arith-expand),func.func(convert-math-to-llvm),convert-math-to-libm,convert-linalg-to-llvm,expand-strided-metadata,finalize-memref-to-llvm,lower-affine,func.func(convert-arith-to-llvm),convert-func-to-llvm,convert-cf-to-llvm,reconcile-unrealized-casts)' /tmp/AtenComplexRealModule.mlir`

I tried using your patch to recreate the error, but I get 

```
error: argument must be a memref of f32, f64, i32, i64, i8, i1 but got 'memref<?xcomplex<f32>>'
```

instead. Do you have a patch with the refbackend work?

pashu123:
> > @ramiro050 Error:
> > ```
> >  <unknown>:0: error: failed to legalize operation 'builtin.unrealized_conversion_cast' that was explicitly marked illegal
> > <unknown>:0: note: see current operation: %41 = "builtin.unrealized_conversion_cast"(%40) : (!llvm.struct<(f32, f32)>) -> complex<f32>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Here's the IR: https://gist.github.com/pashu123/8e2fce24870e3e896708b10c0e93c016
> > Pass:` torch-mlir-opt -pass-pipeline='builtin.module(func.func(refback-generalize-tensor-pad),func.func(linalg-fuse-elementwise-ops),func.func(scf-bufferize),func.func(tm-tensor-bufferize),func.func(empty-tensor-to-alloc-tensor),func.func(linalg-bufferize),func-bufferize,arith-bufferize,refback-mlprogram-bufferize,func.func(tensor-bufferize),func.func(finalizing-bufferize),func.func(buffer-deallocation),refback-munge-calling-conventions,func.func(tm-tensor-to-loops),func.func(refback-munge-memref-copy),func.func(convert-linalg-to-loops),func.func(lower-affine),convert-scf-to-cf,func.func(refback-expand-ops-for-llvm),func.func(arith-expand),func.func(convert-math-to-llvm),convert-math-to-libm,convert-linalg-to-llvm,expand-strided-metadata,finalize-memref-to-llvm,lower-affine,func.func(convert-arith-to-llvm),convert-func-to-llvm,convert-cf-to-llvm,reconcile-unrealized-casts)' /tmp/AtenComplexRealModule.mlir`
> 
> I tried using your patch to recreate the error, but I get
> 
> ```
> error: argument must be a memref of f32, f64, i32, i64, i8, i1 but got 'memref<?xcomplex<f32>>'
> ```
> 
> instead. Do you have a patch with the refbackend work?

I just added https://github.com/llvm/torch-mlir/blob/f9c2b46e621d21d374e4f167f0ee173c3592115e/lib/RefBackend/RefBackend.cpp#L71  ```if (elemTy.isa<ComplexType>()) {
      return true; ```

pashu123:
@ramiro050 Here's the branch with the test: https://github.com/pashu123/torch-mlir/commit/43d92efcd451788daa98126704007555e09cf108

pashu123:
@ramiro Could you PTAL. Thanks.

lib/Dialect/Torch/IR/TorchTypes.cpp
```diff
@@ -404,6 +404,20 @@ static Type convertDtypeToBuiltinElementType(MLIRContext *context, Type dtype) {
   } else if (auto integerType = dtype.dyn_cast<IntegerType>()) {
     return IntegerType::get(context, integerType.getWidth(),
                             IntegerType::Signless);
+  } else if (auto complexType = dtype.dyn_cast<mlir::ComplexType>()) {
```
ramiro050, line 407: I think we should have the importer create the right `complex<...>` type rather than changing things here. Since we are using the complex dialect from the moment we import, we should use the `complex` type the way it is defined in the dialect.

ramiro050 (changes requested):
The issue with the refbackend was that we needed to add an extra pass to turn complex types to llvm. The changes in [this commit](https://github.com/ramiro050/torch-mlir/commit/da3975d47bf4d7e6831eceaaf26209a5335ac59e) allow having complex inputs and returns

lib/Dialect/Torch/IR/TorchTypes.cpp
pashu123: Do you mean adding it the way you added it in the commit - https://github.com/ramiro050/torch-mlir/blob/da3975d47bf4d7e6831eceaaf26209a5335ac59e/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp#L57 here? I tried running the other tests, and they fail because of the type conversion there.

pashu123:
```
<unknown>:0: error: 'linalg.yield' op type of yield operand 1 ('f32') doesn't match the element type of the enclosing linalg.generic op ('f16')
<unknown>:0: note: see current operation: "linalg.yield"(%8) : (f32) -> ()
(shark.venv) ➜  torch-mlir git:(new_complex) ✗ torch-mlir-opt -pass-pipeline='builtin.module(torch-backend-to-linalg-on-tensors-backend-pipeline)' /tmp/AtenComplexRealModule.mlir --mlir-print-ir-after-all
// -----// IR Dump After ConvertTorchToTMTensor (convert-torch-to-tmtensor) //----- //
func.func @forward(%arg0: !torch.vtensor<[?],complex<f32>>) -> !torch.vtensor<[?],f16> {
  %0 = torch.aten.real %arg0 : !torch.vtensor<[?],complex<f32>> -> !torch.vtensor<[?],f16>
  return %0 : !torch.vtensor<[?],f16>
}

<unknown>:0: error: 'linalg.yield' op type of yield operand 1 ('f32') doesn't match the element type of the enclosing linalg.generic op ('f16')
<unknown>:0: note: see current operation: "linalg.yield"(%8) : (f32) -> ()
// -----// IR Dump After ConvertTorchToLinalg Failed (convert-torch-to-linalg) //----- //
"func.func"() ({
^bb0(%arg0: !torch.vtensor<[?],complex<f32>>):
  %0 = "torch_c.to_builtin_tensor"(%arg0) : (!torch.vtensor<[?],complex<f32>>) -> tensor<?xcomplex<f32>>
  %1 = "arith.constant"() {value = 1 : index} : () -> index
  %2 = "arith.constant"() {value = 0 : index} : () -> index
  %3 = "tensor.dim"(%0, %2) : (tensor<?xcomplex<f32>>, index) -> index
  %4 = "tensor.empty"(%3) : (index) -> tensor<?xf16>
  %5 = "linalg.generic"(%0, %4) ({
  ^bb0(%arg1: complex<f32>, %arg2: f16):
    %8 = "complex.re"(%arg1) : (complex<f32>) -> f32
    "linalg.yield"(%8) : (f32) -> ()
  }) {indexing_maps = [affine_map<(d0) -> (d0)>, affine_map<(d0) -> (d0)>], iterator_types = [#linalg.iterator_type<parallel>], operand_segment_sizes = array<i32: 1, 1>} : (tensor<?xcomplex<f32>>, tensor<?xf16>) -> tensor<?xf16>
  %6 = "tensor.cast"(%5) : (tensor<?xf16>) -> tensor<?xf16>
  %7 = "torch_c.from_builtin_tensor"(%6) : (tensor<?xf16>) -> !torch.vtensor<[?],f16>
  "func.return"(%7) : (!torch.vtensor<[?],f16>) -> ()
}) {fun
```ction_type = (!torch.vtensor<[?],complex<f32>>) -> !torch.vtensor<[?],f16>, sym_name = "forward"} : () -> ()

ramiro050: I see. Yeah, in that case, let's leave the importer as is and keep your changes to `convertDtypeToBuiltinElementType`. The right approach would be to create our own `!torch.complex` type and use that during import than have a pass that converts to the MLIR `complex` types.

pashu123: Sounds good! I will push my changes in convertDtypeToBuiltinElementType and raise an issue for the right approach.

ramiro050 (approved):
Great! Thanks for working on this

PR #2047, opened 2023-04-20
Fix import of constant bool tensor parameters

ramiro050:
Bool tensors are represented in TorchScript as an array of `int8_t`s. However, when importing them into Torch-MLIR, the importer was assuming the array had `int32_t` elements, leading to the importer reading into memory that was out of bounds. This commit fixes the casting of the bool tensor.

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -101,6 +101,7 @@ TORCHDYNAMO_XFAIL_SET = {
 }
 
 STABLEHLO_PASS_SET = {
+    "ConstantBoolParameterModule_basic",
     "MaskedFillScalarIntValueStaticModule_basic",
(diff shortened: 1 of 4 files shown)


dan-garvey (approved):

silvasean (commented):
should also update this test: https://github.com/llvm/torch-mlir/blob/97fbb6334b821bf9bafb648be5b45a5a99535d9a/test/python/importer/jit_ir/ivalue_import/tensors.py#L23

ramiro050:
> should also update this test:
> 
> https://github.com/llvm/torch-mlir/blob/97fbb6334b821bf9bafb648be5b45a5a99535d9a/test/python/importer/jit_ir/ivalue_import/tensors.py#L23

Done!

PR #2048, opened 2023-04-21
[TCP] support for zeros and ones lowering

zezhang:
As titled.

Lower `torch.aten.zeros` and `torch.aten.ones` to `tcp.const`

Update: also lower `torch.aten.zeros_like` and `torch.aten.ones_like` to `tcp.const` (with dynamic shape)

Note:

For zeros_like and ones_like, torch-mlir will automatically apply `DecomposeComplexOpsPass` to convert them into const + broadcast. TCP doesn't support all decomposed ops yet (e.g. cast, NumToTensor, size).

One can skip zeros_like and ones_like in `DecomposeComplexOpsPass` and use TCP to directly lower them.

Testing:

`./build/bin/llvm-lit test/Conversion/TorchToTcp/ -v`
`python -m e2e_testing.main --config=tcp -v -s`


lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -415,6 +415,86 @@ class ConvertAtenAtan2Op : public OpConversionPattern<AtenAtan2Op> {
...
+    bool pinMemory;
+    if (!op.getPinMemory().getType().template isa<Torch::NoneType>() &&
+        (!matchPattern(op.getPinMemory(), m_TorchConstantBool(&pinMemory)) ||
+         pinMemory)) {
+      return rewriter.notifyMatchFailure(
+          op, "Unsupported pin_memory, should be either None or false");
+    }
+
+    SmallVector<int64_t> shape;
+    if (!matchPattern(op.getSize(), m_TorchListOfConstantInts(shape))) {
```
silvasean, line 453: Ouch -- isn't TCP aiming to be more dynamic-shape aware?

zezhang: It looks like aten.zeros and aten.ones is used with static shape. Dynamic shape is supported in aten.zeros_like and aten.ones_like

sanjoy: @zezhang I could be missing something here, but why can't we do what we're doing for zeros_like here (create a scalar constant & broadcast with dynamic dims)?  That should allow us to lift the restriction that `op.getSize()` is a constant.

navahgar: This must support dynamic shapes. I just had a discussion with @zezhang. He will be updating this and handling it the same way like `zeros_like`.

zezhang: Just updated.

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -105,6 +105,143 @@ class ConvertValueTensorLiteralOp
...
+          rewriter, op, llvm::ArrayRef(static_cast<int8_t>(fillVal)), {});
+    } else if (outElemTy.isF32()) {
+      constOp = *torch_to_tcp::getConstTensor<float>(
+          rewriter, op, llvm::ArrayRef(static_cast<float>(fillVal)), {});
+    } else if (outElemTy.isF64()) {
+      constOp = *torch_to_tcp::getConstTensor<double>(
+          rewriter, op, llvm::ArrayRef(static_cast<double>(fillVal)), {});
+    } else {
+      return rewriter.notifyMatchFailure(op, "Unsupported output type");
+    }
```
navahgar, line 234: The code until this point is exactly the same for Zeros and ZerosLike ops. Can we refactor this into a function and reuse it? We should avoid duplicating these whenever possible.

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -133,6 +139,72 @@ Value torch_to_tcp::broadcast0DOr1DToNDAndMatchShape(
...
+      resultShape.push_back(num);
+    else
+      resultShape.push_back(ShapedType::kDynamic);
+  }
+  return resultShape;
+}
+
+Value torch_to_tcp::broadcast0DOr1DFromPrimList(
+    ConversionPatternRewriter &rewriter, Value input, Value target,
+    int64_t axisInOutput) {
```
navahgar, line 158: Instead of having this utility take in a `prim::ListConstruct` op, it would be more useful if it takes in a `ArrayRef<Value>`, which are already extracted from the list construct op. 

Same for the `getShapeFromPrimList` function above as well.

You could extract the input values into a vector in the caller function `matchAndRewrite`. This will also avoid having to check for `prim::ListConstruct` op in two places.

test/Conversion/TorchToTcp/misc.mlir
```diff
@@ -30,3 +30,303 @@ func.func @torch.vtensor.literal() -> !torch.vtensor<[4],ui8> {
...
+// CHECK:         %[[T8:.*]] = torch_c.from_builtin_tensor %[[T7]] : tensor<2x3xf32> -> !torch.vtensor<[2,3],f32>
+// CHECK:         return %[[T8]] : !torch.vtensor<[2,3],f32>
+func.func @torch.aten.ones_f32() -> !torch.vtensor<[2,3],f32> {
+  %false = torch.constant.bool false
+  %none = torch.constant.none
+  %int2 = torch.constant.int 2
+  %int3 = torch.constant.int 3
+  %0 = torch.prim.ListConstruct %int2, %int3 : (!torch.int, !torch.int) -> !torch.list<int>
+  %cpu = torch.constant.device "cpu"
+  %1 = torch.aten.ones %0, %none, %none, %cpu, %false : !torch.list<int>, !torch.none, !torch.none, !torch.Device, !torch.bool -> !torch.vtensor<[2,3],f32>
```
navahgar, line 56: Can we have these tests with dynamic shapes instead?

You could have 1 static shape test, but otherwise we should have all dynamic shape tests.

In this case, you could take `int`s as inputs to the function, and use them in `ListConstruct`. Then the shape will be dynamic.

navahgar (changes requested):
Thanks for doing this . Left some comments..

lib/Conversion/TorchToTcp/Misc.cpp
```diff
@@ -105,6 +105,143 @@ class ConvertValueTensorLiteralOp
...
+
+    Type outElemTy = outType.getElementType();
+    if (!outElemTy.isIntOrFloat()) {
+      return rewriter.notifyMatchFailure(
+          op, "Output tensors must have integer or floating-point datatype");
+    }
+
+    // TODO: Make sure this is a valid condition for layout parameter
+    int64_t memoryLayout;
+    if (!op.getLayout().getType().template isa<Torch::NoneType>() &&
```
navahgar, line 197: When any of these params are `None`, the op is supposed to use the corresponding attribute of the input tensor.

See https://pytorch.org/docs/stable/generated/torch.zeros_like.html

This needs to be handled in the lowering as well.

e2e_testing/xfail_sets.py
```diff
@@ -528,6 +545,12 @@
     "TypeAsSameModule_basic",
     "UnsafeView1DFoldModule_basic",
     "View1DFoldModule_basic",
+    "ZerosModuleDefaultDtype_basic",
+    "ZerosModuleFalsePinMemory_basic",
+    "ZerosModuleFloat2D_basic",
+    "ZerosModuleFloat3D_basic",
+    "ZerosModuleInt2D_basic",
+    "ZerosModuleInt3D_basic",
```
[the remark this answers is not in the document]
zezhang: As we discussed, ZerosLike and OnesLike are handled by the `DecomposeComplexOpsPass` first, which involves a few ops that are not supported by TCP yet.

lib/Conversion/TorchToTcp/Misc.cpp
navahgar: Can we refactor all these checks into one function that is called from both these ops?

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -79,7 +85,7 @@ Value torch_to_tcp::broadcastInLeadingDimsToMatchShape(
 
 Value torch_to_tcp::broadcast0DOr1DToNDAndMatchShape(
     ConversionPatternRewriter &rewriter, Value input, Value target,
-    int64_t axisInOutput) {
+    int64_t axisInOutput, bool useInputAsResultType) {
```
navahgar, line 88: I'm not sure I understand why we need the `useInputAsResultType` bool. Instead, can we just create the target type in the callsite and send that in as a parameter?

lib/Conversion/TorchToTcp/Utils.cpp
navahgar: I meant that in the callsite of the utility, we could do something like:
```
   auto resultShape = getShapeFromValues(primtListVals);
   auto resultOp = broadcast0DOr1D(rewriter, input, values, resultShape, axisInOutput);
```

This way, both the util funcs are not tied to a primList and can be used in other places as well.

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -210,3 +295,27 @@ torch_to_tcp::getConstTensor<APInt>(PatternRewriter &rewriter, Operation *op,
...
+    constOp = *torch_to_tcp::getConstTensor<float>(
+        rewriter, op, llvm::ArrayRef(static_cast<float>(fillVal)), {});
+  } else if (resultType.isF64()) {
+    constOp = *torch_to_tcp::getConstTensor<double>(
+        rewriter, op, llvm::ArrayRef(static_cast<double>(fillVal)), {});
+  } else {
+    return false;
+  }
+  return true;
+}
```
navahgar, line 321: nit: newline

navahgar (approved):
LGTM. Thanks for adding this.

PR #2054, opened 2023-04-24
Remove Randn* tests from crashing tests list in torchdynamo

ramiro050:
This PR cherry-picks f85f5799 from `main`, which fixes the crashes with the `Randn*` tests. A subsequent commit removes the list of crashing tests from the CI config.


li-plus (approved):
Awesome! So we don't have crashing cases anymore.

PR #2055, opened 2023-04-24
Cherry-pick 4f6de745 "Cast `number` to `float` when shape function takes Scalar arg"

ramiro050:
To keep things simple in shape functions, `Scalar` inputs are considered `float`s. This means that when inserting the shape functions into the IR, we must cast any `!torch.number`s into `float`s so that the operand type matches the expected type in the shape function. This commit adds the cast from `Scalar` to `float`.


ramiro050:
This is needed for my next set of dtype functions.

li-plus (approved):
LGTM!

PR #2058, opened 2023-04-24
[Torch Op] Add unbind.int support with ListUnpack

zhekunz2:
As title.

Decompose `unbind.int` + `ListUnpack` to several `slice.Tensor` ops.


zhekunz2:
Update: looks like torch-mlir runtime doesn't support unbind 
https://github.com/llvm/torch-mlir/issues/1896#issuecomment-1459096288

Postpone this PR when it's supported.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2025,6 +2026,36 @@ void PrimListUnpackOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
     rewriter.replaceOp(op, listConstruct.getElements());
     return success();
   });
+
+  patterns.add(+[](PrimListUnpackOp op, PatternRewriter &rewriter) {
+    // decompose AtenUnbindOp + PrimListUnpackOp to slice.tensor ops
+    if (!isa<AtenUnbindIntOp>(op.getOperand().getDefiningOp()))
+      return failure();
+    AtenUnbindIntOp unbind = cast<AtenUnbindIntOp>(op.getOperand().getDefiningOp());
```
ramiro050, line 2034: nit: you can avoid the double cast by using `dyn_cast` instead

```c++
auto unbind = dyn_cast<AtenUnbindIntOp>(op.getOperand().getDefiningOp());
if (!unbind)
  return failure();
```

ramiro050 (changes requested):
Thanks!

ramiro050:
> Update: looks like torch-mlir runtime doesn't support unbind [#1896 (comment)](https://github.com/llvm/torch-mlir/issues/1896#issuecomment-1459096288)

Yeah, torch-mlir cannot support `unbind` in a normal way because it returns a list of tensors, and there is no runtime support for lists. However, having this type of pattern where the op gets rewritten into other ops torch-mlir *can* handle is completely fine.

zhekunz2:
Looks like build failed because it returns vector of `tensor`s. 
```
/main_checkout/torch-mlir/build/tools/torch-mlir/python/torch_mlir/csrc/base_lazy_backend/generated/LazyNativeFunctions.cpp:5383:16: error: no viable conversion from returned value of type 'at::Tensor' to function return type '::std::vector<at::Tensor>'
        return result;
```

python/torch_mlir_e2e_test/test_suite/slice_like.py
```diff
@@ -542,3 +542,23 @@ def forward(self, x, y):
 @register_test_case(module_factory=lambda: SliceCopyNegative_Module())
 def SliceCopyNegative_Module_basic(module, tu: TestUtils):
     module.forward(tu.rand(10, 4, 4), tu.rand(4, 4, 4))
+
+
+# ==============================================================================
+
+class UnbindIntListUnpack_Module(torch.nn.Module):
```
ramiro050, line 549: Can you add an e2e test for the `__getitem__` pattern?

lib/Dialect/Torch/Transforms/RecomposeComplexOps.cpp
```diff
@@ -121,6 +121,64 @@ class RecomposeSelectFill_ : public OpRewritePattern<AtenFill_TensorOp> {
...
+
+    Location loc = op.getLoc();
+    Value dim = unbind.getDim();
+    Value input = unbind.getSelf();
+    // rewrite to slice op
+    auto resultTy = op.getResult().getType();
+    auto newSelect = rewriter.create<AtenSelectIntOp>(loc, resultTy, input, dim,
+                                                      op.getIdx());
+    Value squeeze =
+        rewriter.create<AtenSqueezeDimOp>(loc, resultTy, newSelect, dim);
```
ramiro050, line 177: Why is the squeezing necessary?

lib/Dialect/Torch/Transforms/RecomposeComplexOps.cpp
```diff
@@ -121,6 +121,64 @@ class RecomposeSelectFill_ : public OpRewritePattern<AtenFill_TensorOp> {
...
+class RecomposeUnbindGetItem : public OpRewritePattern<Aten__Getitem__TOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(Aten__Getitem__TOp op,
+                                PatternRewriter &rewriter) const override {
+    // recompose AtenUnbindIntOp + __getitem__t to select.int + squeeze op
+    auto unbind = dyn_cast<AtenUnbindIntOp>(op.getList().getDefiningOp());
+    if (!unbind)
+      return failure();
+    if (isListPotentiallyMutated(op.getList()))
```
ramiro050, line 162: The function `isListPotentiallyMutated` looks at the users of the passed list. Since we are trying to prevent the case of

```python
result_list = torch.unbind(...)
mutate(result_list)
first_result = torch.__getitem__(result_list, 0)
```

then the list we actually want to pass is the result of `AtenUnbindIntOp`. This check should also be performed in the `RecomposeUnbindListUnpack` pattern.

ramiro050 (commented):
I think you also need to XFAIL some LTC tests

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -588,6 +588,7 @@ def emit_with_mutating_variants(key, **kwargs):
     emit("aten::any.bool : (bool[]) -> (bool)")
     emit("aten::sort.int : (int[], bool) -> ()", has_canonicalizer=True)
     emit("aten::sort : (Tensor, int, bool) -> (Tensor, Tensor)")
+    emit("aten::unbind.int : (Tensor, int) -> (Tensor[])", traits=["Pure"])
```
ramiro050, line 591: This is not a Pure op, since it can raise an exception when the inputs are invalid

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
zhekunz2: I wanted to use `Pure` trait to DCE the op if there's no use. Sounds like I need a pattern for it then.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1502,6 +1502,20 @@ void AtenSortIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
   });
 }
 
+//===----------------------------------------------------------------------===//
+// AtenUnbindIntOp
+//===----------------------------------------------------------------------===//
+
+void AtenUnbindIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                                  MLIRContext *context) {
+  patterns.add(+[](AtenUnbindIntOp op, PatternRewriter &rewriter) {
```
ramiro050, line 1511: Same comment as in the other PR: https://github.com/llvm/torch-mlir/pull/2102#discussion_r1194437856

ramiro050 (changes requested):

PR #2060, opened 2023-04-25
Update to LLVM 3157f03a349cfc852cdd994675eaa9652caa2e3a

eric-k256:
New requirement to explicitly cast for interfaces https://reviews.llvm.org/D148493


ashay (approved):
LGTM. That was fast! :)

PR #2065, opened 2023-04-26
Fix default value of `stride` in 2d pooling ops in linalg and tosa

ramiro050:
When the user does not specify the `stride` value in 2d pooling ops, `stride` is given the value of an empty list. However, the current lowerings for pooling ops assumed that the `stride` operand would always be a list of two ints, leading to crashes when that was not the case. This commit fixes the crashes by setting the value of `stride` to `kernel_size` when `stride` is the empty list, since this is the default `stride` value specified in PyTorch docs. See: https://pytorch.org/docs/stable/generated/torch.nn.MaxPool2d.html#torch.nn.MaxPool2d

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -500,6 +500,7 @@ STABLEHLO_PASS_SET = {
 # Write the TOSA set as a "passing" set as it is very early in development
 # and very few tests work yet.
 TOSA_PASS_SET = {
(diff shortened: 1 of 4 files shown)


eric-k256 (approved):
Looks good to me. 

Before getting the testcase, i was trying to reproduce the tosa backend failure with a single value stride, but the e2e tests wouldn't fail with a single value.  Thanks for finding the root cause.

vivekkhandelwal1 (approved):
LGTM!

PR #2070, opened 2023-04-27
Add e2e linalg support for aten.atan

zezhang:
As titled.

To test:

`python -m e2e_testing.main -v`


ramiro050 (approved):
LGTM! Can you just update the PR title to something like "Add e2e linalg support for aten.atan", since this PR is doing more than just adding the ODS.

Issue #2067, opened 2023-04-27
Missing ";" bewtween includes in MHLOTargets.cmake


tebartsch:
A `;`  is missing in `<build-dir>/lib/cmake/mlir-hlo/MHLOTargets.cmake`:

```cmake
set_target_properties(TorchMLIRTMTensorDialect PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "/test_project/third_party/torch-mlir/externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/TMTensor/IR//test_project/third_party/llvm-project/mlir/include;/test_project/build/llvm/tools/mlir/include;/test_project/third_party/torch-mlir/externals/llvm-external-projects/torch-mlir-dialects/include;/test_project/build/torch-mlir/externals/llvm-external-projects/torch-mlir-dialects/include"
  INTERFACE_LINK_LIBRARIES "MLIRAffineDialect;MLIRDialectUtils;MLIRIR;MLIRLinalgDialect;MLIRMathDialect;MLIRMemRefDialect;MLIRPass;MLIRSideEffectInterfaces;MLIRSupport;MLIRSCFDialect;MLIRFuncDialect;MLIRTensorDialect;MLIRViewLikeInterface;LLVMSupport"
)
```
(no `;` at `.../TMTensor/IR//test_project...`).

Adding quotations marks to `_dirs` in `torch-mlir/externals/llvm-external-projects/torch-mlir-dialects/CMakeLists.txt` seems to fix the issue.

```diff
diff --git a/externals/llvm-external-projects/torch-mlir-dialects/CMakeLists.txt b/externals/llvm-external-projects/torch-mlir-dialects/CMakeLists.txt
index 2de2d4eb..3756bc1b 100644
--- a/externals/llvm-external-projects/torch-mlir-dialects/CMakeLists.txt
+++ b/externals/llvm-external-projects/torch-mlir-dialects/CMakeLists.txt
@@ -47,9 +47,9 @@ function(torch_mlir_dialects_target_includes target)
   # target, when present, is just used for compilation and does not
   # contribute to the interface properties.
   # TODO: Normalize this upstream.
-  target_include_directories(${target} PUBLIC ${_dirs})
+  target_include_directories(${target} PUBLIC "${_dirs}") 
   if(TARGET obj.${target})
-    target_include_directories(obj.${target} PRIVATE ${_dirs})
+    target_include_directories(obj.${target} PRIVATE "${_dirs}")
[...]

PR #2074, opened 2023-04-29
Missing ";" bewtween includes in MHLOTargets.cmake

makslevental:
Fixes https://github.com/llvm/torch-mlir/issues/2067

FYI you can land commits on behalf of people and still attribute like so:

```
git commit --amend --author="John Doe <jdoe@llvm.org>"
```

cc @tebartsch


ashay (approved):
Thanks!

PR #2078, opened 2023-05-01
[Torch Dialect] Enforce signless attribute for ConstantIntOp

zhekunz2:
as title.


lib/Dialect/TorchConversion/IR/TorchConversionOps.cpp
```diff
@@ -91,7 +91,20 @@ OpFoldResult FromI64Op::fold(FoldAdaptor adaptor) {
...
-    return attr;
+    Attribute foldResult = attr;
+    if (attr.getType().isSignedInteger()) {
+      // convert sint type to i64 type
+      foldResult =
+          IntegerAttr::get(IntegerType::get(getContext(), 64), attr.getSInt());
+    } else if (attr.getType().isUnsignedInteger()) {
+      // convert uint to i64 type
+      foldResult = IntegerAttr::get(IntegerType::get(getContext(), 64),
+                                    static_cast<int64_t>(attr.getUInt()));
```
makslevental, line 102: This one is fishy - you need something like https://stackoverflow.com/a/17251989 to make sure that it fits - but I don't trust that particular snippet and I don't know have a hardened implementation on-hand. 

If you want to land without this, I think the rest is good.

makslevental (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1713,7 +1713,10 @@ ParseResult ConstantIntOp::parse(OpAsmParser &parser, OperationState &result) {
 
 void ConstantIntOp::print(OpAsmPrinter &p) {
   p << " ";
-  p << getValue().getSExtValue();
+  if (getValueAttr().getType().isUnsignedInteger())
```
silvasean, line 1716: The intention is that this attribute is always `i64`. Can you update the verifier to reject any attribute that is not `i64` in type? We shouldn't be hitting this case at all -- how are you running into it?

zhekunz2: Sure, I can update the verifier. 
I actually didn't run into `uint` case. Mine case was a folding a `torch_c.toi64` with `si64` attribute `torch.constant.int`. My intention is to remove the signedness when folding. I think torch integer types by default are signed integers according to https://github.com/llvm/torch-mlir/blob/main/include/torch-mlir/Dialect/Torch/IR/TorchTypes.td#L127

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1721,6 +1721,14 @@ OpFoldResult Torch::ConstantIntOp::fold(FoldAdaptor adaptor) {
   return getValueAttr();
 }
 
+LogicalResult ConstantIntOp::verify() {
+  Type valTy = getValueAttr().getType();
+  if (!valTy.isSignedInteger() && !valTy.isSignlessInteger())
```
zhekunz2, line 1726: @silvasean @ramiro050 I've noticed that many places uses `torch.constant.int` with signless integer attributes, eg `i64/i32`, instead of signed integers like `si64/si32`. This isn't causing any issue right now because backends convert that to signless integers, like `arith.constant`. 

I will leave the verifier to check if it's either signed or signless integer for now, but it would be great to be consistent about the attribute type.

lib/Dialect/Torch/IR/TorchOps.cpp
silvasean: I think `torch.constant.int` should always have a signless integer attribute -- https://github.com/llvm/torch-mlir/blob/eaaaeb6ff15d676f901035f86af55a17dc252781/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp#L391

For now we want the verifier to enforce that I think. Where are you seeing si64? The case for tensor dtype is a bit different so TorchTypes.td#L127 doesn't apply here. Here we are talking about the modeling of Python scalars.

lib/Dialect/Torch/IR/TorchOps.cpp
silvasean: Let's verify that it is always signless for now.

silvasean:
Thanks for digging into this! This is very subtle :)

zhekunz2:
> I think `torch.constant.int` should always have a signless integer attribute --
> 
> https://github.com/llvm/torch-mlir/blob/eaaaeb6ff15d676f901035f86af55a17dc252781/python/torch_mlir/dialects/torch/importer/jit_ir/csrc/torch_to_mlir_utils.cpp#L391
> 
> For now we want the verifier to enforce that I think. Where are you seeing si64? The case for tensor dtype is a bit different so TorchTypes.td#L127 doesn't apply here. Here we are talking about the modeling of Python scalars.


Thank you for your clarifications! I'm seeing `si64` because some fold rules give `si64` attributes. I fixed those in the latest commits, and my case was resolved. @silvasean Can you help review this PR again to adds the verifier for `ConstantIntOp`?

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -728,6 +728,7 @@ def Torch_ConstantIntOp : Torch_Op<"constant.int", [
   // Indicate that the operation has a custom parser and printer method.
   let hasCustomAssemblyFormat = 1;
   let hasFolder = 1;
+  let hasVerifier = 1;
```
silvasean, line 731: I think if you change `AnyI64Attr` to `I64Attr` above then you won't need the custom verifier.

silvasean (approved):
Thanks!

zhekunz2:
@makslevental Did I resolve your concerns?

zhekunz2:
@makslevental Can you take another look at this PR?

makslevental (approved):
LGTM!

zhekunz2:
> LGTM! @zhekunz2 let me know if you need me to merge.

@makslevental Yes please help merge this as well. Thank you!

PR #2080, opened 2023-05-01
Revert "update PyTorch version to 2.1.0.dev20230430"

ramiro050:
This reverts commit eab1930c491d0a59cc0467f0091f5a05f9d2bd60.

The PyTorch version does not yet exist for MacOS.


makslevental:
Can we just disable this thing? Or at least gate it by whether things successfully compile after the roll?

makslevental (approved):
Can we just disable this thing? Or at least gate it by whether things successfully compile after the roll?

powderluv (approved):

powderluv:
yeah we should check for mac builds also before roll. @ashay should we just add the normal macos test ?

ramiro050:
The MacOS build will take (from looking at the current pace) around 2 more hours to finish. However, this PR should pass all the builds, since this is reverting the latest commit on main and the previous commit is green. I will go ahead and merge early to unblock people.

ashay:
> should we just add the normal macos test ?

Yes, this seems to be a recurring issue, so I'll add macOS to the list of tested platforms.  I'll try to get this out this week.

PR #2081, opened 2023-05-01
Increase ccache max_size

makslevental:
Are we "cache" poor that we can't afford to increase this? I'm seeing recompiles after a sequence of test builds that make no change to llvm. [`6G`](https://github.com/iree-org/iree-llvm-sandbox/blob/main/.github/workflows/buildAndTestStructured.yml#L45-L47) seems to be a good number.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -53,7 +53,7 @@ runs:
       mkdir -p ${{ github.workspace }}/.ccache
       ccache --set-config "cache_dir=${{ github.workspace }}/.ccache"
       ccache --set-config "compression=true"
-      ccache --set-config "max_size=300M"
+      ccache --set-config "max_size=6G"
       ccache --zero-stats
     shell: bash
 


powderluv:
@ashay tested / tuned this number based on how much we can save IIRC

ashay:
Howdy!  My concerns with larger cache sizes are that (a) it takes longer to download the cache artifact and (b) we end up with fewer available cache artifacts (since they all have to fit within GitHub's 10GB limit).  So when cache artifacts reach 6GB (for each platform), we will start seeing invalidations very quickly.

The 300M limit sounds low but it was based on sizes of newly created cache artifacts, which happen to be around 100MB to 200MB.  The ccache documentation says that once caches hit the max size, ccache will start pruning old entries, but I'm not sure if it works (the [cache artifacts page](https://github.com/llvm/torch-mlir/actions/caches) shows that caches go beyond 300MB).  You might ask, why even set the 300MB limit?  Because I am hoping that the cache size limit does _something_ and I fear that setting it to a large value might let it go hog wild.

All that said, my workaround for this issue so far has been rather dumb.  I occasionally delete all cache artifacts.  This makes the first run slow (cache miss), but it brings the cache artifacts back to the <300MB size and it also appears to bring back the cache hits.

If you point me to the cache misses, I'll spend some time debugging them tomorrow.  I hope you're referring to the Linux or macOS builds, because ccache on Windows has some other issues that are harder to debug.

PR #2083, opened 2023-05-02
[MLIR][TORCH] Add E2E support for `aten.topk` op

vivekkhandelwal1:
This commit adds the decomposition for the aten.topk op.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4320,6 +4320,49 @@ class DecomposeAtenVarMeanDimOp : public OpRewritePattern<AtenVarMeanDimOp> {
...
+      return rewriter.notifyMatchFailure(
+          op, "Expected a constant boolean value for sorted");
+    if (!sorted)
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: sorted value arg must be set to True");
+
+    Value self = op.getSelf();
+    Value dim = op.getDim();
+    auto selfType = self.getType().cast<ValueTensorType>();
+    auto sortIndicesType = ValueTensorType::get(
```
ramiro050, line 4344: nit: I would cast to `BaseTensorType` and use `selfType.getWithSizesAndDtype` to define `sortIndicesType`, since there is no guarantee that things will have value semantics when this pattern is applied.

ramiro050 (approved):
LGTM, just one change request

PR #2091, opened 2023-05-04
[Torch Dialect] fix aten.native_layer_norm's dtype infer and decompos…

qingyunqu:
…ition.
This PR is special for `aten.native_layer_norm` with fp16.
But it's hard to add a fp16 testcase, https://github.com/llvm/torch-mlir/issues/2059.


qingyunqu:
@ramiro050 Could you please review this PR?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2532,29 +2524,38 @@ class DecomposeAtenNativeLayerNormOp
     Value cstTrue = rewriter.create<Torch::ConstantBoolOp>(loc, true);
     Value none = rewriter.create<Torch::ConstantNoneOp>(loc);
     // mean(x)
+    auto convertedInput = convertTensorToDtype(rewriter, loc, op.getInput(),
```
ramiro050, line 2527: The changes to this pattern seem a bit hard to follow. Can we instead keep the original decomposition, but then before the `rewriter.replaceOp` line at the end add casts to the three outputs to their respective result dtypes?

Also, can you add an e2e test for this? I think you mentioned f16 does not work, but [support for complex values was recently added](https://github.com/llvm/torch-mlir/pull/2044), so we should be able to check complex inputs.

ramiro050 (changes requested):
Thanks for the reminder, this PR slipped through the cracks

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
qingyunqu: Undecomposed NativeLayerNorm:
```
module attributes {torch.debug_module_name = "NativeLayerNormModule"} {
  func.func @forward(%arg0: !torch.vtensor<[2,5,2,2,3],f16>, %arg1: !torch.vtensor<[2,2,3],f16>, %arg2: !torch.vtensor<[2,2,3],f16>) -> (!torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,5,1,1,1],f32>, !torch.vtensor<[2,5,1,1,1],f32>) {
    %float5.000000e-01 = torch.constant.float 5.000000e-01
    %int3 = torch.constant.int 3
    %int2 = torch.constant.int 2
    %0 = torch.prim.ListConstruct %int2, %int2, %int3 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %result0, %result1, %result2 = torch.aten.native_layer_norm %arg0, %0, %arg1, %arg2, %float5.000000e-01 : !torch.vtensor<[2,5,2,2,3],f16>, !torch.list<int>, !torch.vtensor<[2,2,3],f16>, !torch.vtensor<[2,2,3],f16>, !torch.float -> !torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,5,1,1,1],f32>, !torch.vtensor<[2,5,1,1,1],f32>
    return %result0, %result1, %result2 : !torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,5,1,1,1],f32>, !torch.vtensor<[2,5,1,1,1],f32>
  }
}
```

Decomposed NativeLayerNorm:
```
module attributes {torch.debug_module_name = "NativeLayerNormModule"} {
  func.func @forward(%arg0: !torch.vtensor<[2,5,2,2,3],f16>, %arg1: !torch.vtensor<[2,2,3],f16>, %arg2: !torch.vtensor<[2,2,3],f16>) -> (!torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,5,1,1,1],f32>, !torch.vtensor<[2,5,1,1,1],f32>) {
    %int2 = torch.constant.int 2
    %int3 = torch.constant.int 3
    %int4 = torch.constant.int 4
    %int1 = torch.constant.int 1
    %none = torch.constant.none
    %int12 = torch.constant.int 12
    %int5 = torch.constant.int 5
    %float5.000000e-01 = torch.constant.float 5.000000e-01
    %0 = torch.prim.ListConstruct %int2, %int3, %int4 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %1 = torch.aten.sum.dim_IntList %arg0, %0, %true, %none : !torch.vtensor<[2,5,2,2,3],f16>, !torch.list<int>, !torch.bool, !torch.none -> !torch.vtensor<[2,5,1,1,1],f32>
    %2 = torch.aten.div.Scalar %1, %int12 : !torch.vtensor<[2,5,1,1,1],f32>, !torch.int -> !torch.vtensor<[2,5,1,1,1],f32>
    %3 = torch.prim.ListConstruct %int2, %int5, %int2, %int2, %int3 : (!torch.int, !torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %4 = torch.aten.broadcast_to %2, %3 : !torch.vtensor<[2,5,1,1,1],f32>, !torch.list<int> -> !torch.vtensor<[2,5,2,2,3],f16>
    %5 = torch.aten.sub.Tensor %arg0, %4, %int1 : !torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,5,2,2,3],f16>, !torch.int -> !torch.vtensor<[2,5,2,2,3],f16>
    %6 = torch.aten.mul.Tensor %5, %5 : !torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,5,2,2,3],f16> -> !torch.vtensor<[2,5,2,2,3],f16>
    %7 = torch.aten.sum.dim_IntList %6, %0, %true, %none : !torch.vtensor<[2,5,2,2,3],f16>, !torch.list<int>, !torch.bool, !torch.none -> !torch.vtensor<[2,5,1,1,1],f32>
    %8 = torch.aten.div.Scalar %7, %int12 : !torch.vtensor<[2,5,1,1,1],f32>, !torch.int -> !torch.vtensor<[2,5,1,1,1],f32>
    %9 = torch.aten.add.Scalar %8, %float5.000000e-01, %int1 : !torch.vtensor<[2,5,1,1,1],f32>, !torch.float, !torch.int -> !torch.vtensor<[2,5,1,1,1],f32>
    %10 = torch.aten.rsqrt %9 : !torch.vtensor<[2,5,1,1,1],f32> -> !torch.vtensor<[2,5,1,1,1],f32>
    %11 = torch.prim.ListConstruct %int2, %int5, %int2, %int2, %int3 : (!torch.int, !torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
    %12 = torch.aten.broadcast_to %10, %11 : !torch.vtensor<[2,5,1,1,1],f32>, !torch.list<int> -> !torch.vtensor<[2,5,2,2,3],f16>
    %13 = torch.aten.mul.Tensor %5, %12 : !torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,5,2,2,3],f16> -> !torch.vtensor<[2,5,2,2,3],f16>
    %14 = torch.aten.mul.Tensor %13, %arg1 : !torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,2,3],f16> -> !torch.vtensor<[2,5,2,2,3],f16>
    %15 = torch.aten.add.Tensor %14, %arg2, %int1 : !torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,2,3],f16>, !torch.int -> !torch.vtensor<[2,5,2,2,3],f16>
    return %15, %2, %10 : !torch.vtensor<[2,5,2,2,3],f16>, !torch.vtensor<[2,5,1,1,1],f32>, !torch.vtensor<[2,5,1,1,1],f32>
  }
}
```

qingyunqu: The problem this that `%4` and `%12` is broadcasted from `f32` to `f16` which is invalid.

qingyunqu: As for e2e test, I think it's also hard to add test of complex type. Because torch doesn't have LayerNormKernel of ComplexFloat:
```
RuntimeError: "LayerNormKernelImpl" not implemented for 'ComplexFloat'
```

ramiro050: Regarding the broadcast type error, the issue is that the decompositions is not using the right result dtypes for some of the ops. Sometimes `reducedTy` is used, as seen here,

https://github.com/llvm/torch-mlir/blob/ff810b04aac0c96261c4ed89b8df30c2333cafb3/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L2530

and sometimes `inputTy` is used, leading to incorrect IR.

I suggest we take the following approach:
1. Determine the type with the largest width from the three input tensors
2. cast all input tensors to that largest-width dtype
3. Perform the decomposition using the largest-width dtype as the result type for all the ops
4. At the end, cast the three result tensors to their corresponding output types

Regarding testing, this is a bit annoying that we cannot actually check that the op works properly. Let me think a little bit about how to approach this, and I'll get back to you.

ramiro050: @silvasean, the fixes in this PR cannot be tested e2e because there is no f16 version of native-layer-norm on CPU. However, the numerical correctness of the decomposition is already tested with other input types and the dtype function is also tested. I think it should be okay to not have an e2e test in this PR. What do you think?

PR #2094, opened 2023-05-05
[Torch Dialect] require dtype exists when decompose to aten.where.self

qingyunqu:
If dtype doesn't exist in intermediate state, it will decompose to something like: 
```
%745 = "torch.aten.to.dtype"(%744, %0, %1, %1, %2) : (!torch.vtensor<[],f64>, !torch.int, !torch.bool, !torch.bool, !torch.none) -> !torch.vtensor<[],f32>
%746 = "torch.aten.broadcast_to"(%745, %743) : (!torch.vtensor<[],f32>, !torch.list<int>) -> !torch.vtensor<[],f32>
%747 = "torch.aten.where.self"(%742, %746, %741) : (!torch.vtensor<[1,1,64,64],i1>, !torch.vtensor<[],f32>, !torch.vtensor<[1,1,64,64],f16>) -> !torch.vtensor<[1,1,64,64],f16>
```

Another fixing method is to support type converting when conversion to backend, like https://github.com/llvm/torch-mlir/pull/2093 do.

I think we need **both** of correct decomposition and type converting when to stablehlo.


qingyunqu:
BTW, I think it's hard to write a real model test. The problem always occur after the first iteration of `shape infer`, `dtype infer` and `decomposition`.

ramiro050:
The issue seems to be that `createInitTensor` does not specify the `dtype` argument when creating the `AtenFullOp`.

https://github.com/llvm/torch-mlir/blob/20e7d52d366e32d94b3d46bd38d155546d2a91d1/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L135

This should solve the decomposition error you're seeing

qingyunqu:
> The issue seems to be that `createInitTensor` does not specify the `dtype` argument when creating the `AtenFullOp`.
> 
> https://github.com/llvm/torch-mlir/blob/20e7d52d366e32d94b3d46bd38d155546d2a91d1/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L135
> 
> Specifying the `dtype` arg should solve the decomposition error you're seeing.

I think the problem is that `createRank0Tensor` doesn't check the `inputType`'s dtype existing.

ramiro050:
> I think the problem is createRank0Tensor. createRank0Tensor will convert the element type to inputType but it doesn't check the inputType has a dtype.

Right, we need to do both. At some point we need to check if the `dtype` exists. But the `dtype ` argument of `AtenFullOp` should also be specified since leaving it as `None` and having `resultType` of `AtenFullOp` be something other than `f32` is invalid IR.

ramiro050 (approved):
Thanks!

PR #2095, opened 2023-05-05
[Tcp] Clone constants into the IsolatedGroup region

navahgar:
This PR fixes the handling of constants in the IsolateGroupOps pass. Specifically, any constant that is used within the isolated group op will now be cloned into the start of the block.

This also fixes another issue with duplicate input arguments in the isolated blocks, when there are multiple uses of that value.


sjarus (approved):
Nice! Thanks for the quick turnaround!

PR #2097, opened 2023-05-07
[Linalg] fix lowering reduce max with -inf

qingyunqu:
This PR fix reduce max's init values to `-inf`, which is compatible with torch.
In torch:
```python
In [7]: b
Out[7]: tensor([-inf, -inf])
In [8]: torch.amax(b, 0)
Out[8]: tensor(-inf)
```


qingyunqu:
@vivekkhandelwal1 , could you please review this PR?

vivekkhandelwal1 (approved):
LGTM

PR #2098, opened 2023-05-07
Pure-Python FX importer.

makslevental:
This PR adds a pure fx graph importer based on dynamo.

cc @powderluv

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -56,20 +56,7 @@ TORCHDYNAMO_XFAIL_SET = {
     "ElementwiseWhereScalarSelfModule_basic",
     "ElementwiseWhereScalarOtherStaticModule_basic",
     "ElementwiseWhereScalarSelfStaticModule_basic",
-    # %7 = torch.operator "aten._index_put_impl_.hacked_twin"(%1, %6, %5, %true, %false) : (!torch.tensor<*,f32>, !torch.list<tensor>, !torch.tensor<*,f32>, !torch.bool, !torch.bool) -> !torch.tensor
-    "IndexPutImpl1DFloatAccumulateModule_basic",
-    "IndexPutImpl1DFloatNonAccumulateModule_basic",
-    "IndexPutImpl1DIntAccumulateModule_basic",
-    "IndexPutImpl1DIntNonAccumulateModule_basic",
-    "IndexPutImpl2DFloatAccumulateModule_basic",
-    "IndexPutImpl2DFloatNonAccumulateModule_basic",
-    "IndexPutImpl2DIndexModule_basic",
-    "IndexPutImpl3DFloatAccumulateModule_basic",
-    "IndexPutImpl3DFloatNonAccumulateModule_basic",
 
-    # https://github.com/llvm/torch-mlir/issues/1611
-    # error: 'tensor.cast' op operand type 'tensor<0xi64>' and result type 'tensor<18xi64>' are cast incompatible
-    "Aten_EmbeddingBagExample_basic",
     # error: failed to legalize operation 'torch.valsem.aten.bernoulli.float' that was explicitly marked illegal
     "BernoulliFloatModule_basic",
     "BernoulliPModule_basic",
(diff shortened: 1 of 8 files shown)


python/torch_mlir_e2e_test/configs/torchdynamo.py
```diff
@@ -2,96 +2,149 @@
...
-            return result
-        # TODO: Deepcopy the torch.nn.Module, so that if the program is
-        # stateful then it does not mutate the original compiled program.
         result: Trace = []
         for item in trace:
-            f = lambda method, *inputs: method(*inputs)
-            torch._dynamo.reset()
-            dynamo_f = dynamo.optimize(_refbackend_torchdynamo_backend, nopython=True)(f)
-            output = dynamo_f(item_symbol_that_clones_inputs, *item.inputs)
+            module = jit(artifact,
```
silvasean, line 131: You can do the self.backend.compile/self.backend.load inside the dynamo callback itself and have the thing you return ffrom the dynamo callback be the result of self.backend.load. Then you don't need to do this "nonlocal" thing.

e2e_testing/xfail_sets.py
```diff
@@ -201,10 +186,87 @@
     'IsFloatingPointInt_False',
     'TorchPrimLoopForLikeModule_basic',
     'TorchPrimLoopWhileLikeModule_basic',
+
+    # backend never runs because of empty frame
```
silvasean, line 190: One of the main purpose of the XFAIL set is to serve as a living and up-to-date-by-construction set of actionable work items needed for the project to reach a desired state. For many of these, it's not clear what is actionable -- are these upstream bugs? (Please file them), or are they deliberately not working functionality upstream that won't be fixed?

python/torch_mlir/__init__.py
```diff
@@ -248,6 +248,64 @@ def _get_for_tracing(
 }
 
 
+def canon_extra_library(extra_library):
```
silvasean, line 251: We shouldn't need to expose this.

python/torch_mlir/__init__.py
```diff
@@ -248,6 +248,64 @@ def _get_for_tracing(
...
+        mlir_library = generate_library(extra_library_dict)
+
+        extra_library_file_name = \
+            tempfile.gettempdir() + "/custom_op_extra_library.mlir"
+        with open(extra_library_file_name, "w") as f:
+            f.write(mlir_library)
+    return extra_library_file_name
+
+
+def lower_mlir_module(verbose, output_type, module):
```
silvasean, line 266: Why does this need to be exposed?

silvasean (changes requested):

e2e_testing/xfail_sets.py
makslevental: All of these have already been discussed and adjudicated here https://github.com/llvm/torch-mlir/pull/2040#issuecomment-1518303632.

python/torch_mlir_e2e_test/configs/torchdynamo.py
makslevental: That's extremely obscure - to pass a closure with all of these things all the way down the dynamo callstack. As I pointed out in the previous PR, the `nonlocal` "thing" is the upstream [recommended approach](https://github.com/pytorch/pytorch/blob/97bb4c2538bd1a49e633488a691f0b0a017c6dcf/torch/_dynamo/eval_frame.py#L844).

python/torch_mlir/__init__.py
makslevental: There's often a lot of talk of "exposure" in these review and I have to finally concede I have no idea what it means to expose/not-expose things. I'm sure you're well aware that you can't make private absolutely anything in Python. So what's the material recommendation here? That I should prefix these functions with the `_` fig leaf?

silvasean: The "exposure" is about managing user expectations -- without clear guidance (or mechanical enforcement, which as you pointed out is hard in Python) users can end up relying on things that we don't intend to support, which can cause grief (either on their side, via us breaking them, or our side, by us getting stuck supporting something that we didn't intend to). It is also a useful signal to us for users to come and say that they want particular API's to be exposed so that they can officially rely on them.

Indeed, using `__all__` would be great here. So far we have been using the leading `_` convention.

python/torch_mlir_e2e_test/configs/torchdynamo.py
silvasean: Hey Maks, I really appreciate all the work you've done on this and I sense your frustration and the pressure you are under. If you are completely unwilling to make the changes being suggested here then, considering all the factors and being respectful of your time, I think it would be best for you to go ahead to land this, and I'll make the changes I'm asking for in a follow-up.

makslevental: Thank you for the consideration. Yes, I'd prefer to land as is so can you please approve (or @ramiro050)?

PR #2102, opened 2023-05-08
[Torch Dialect] Add split.tensor support + recompose rules

zhekunz2:


zhekunz2:
Similar to https://github.com/llvm/torch-mlir/pull/2058, LazyTensorCore failed to build because it returns list of tensors.

lib/Dialect/Torch/Transforms/RecomposeComplexOps.cpp
```diff
@@ -121,6 +121,44 @@ class RecomposeSelectFill_ : public OpRewritePattern<AtenFill_TensorOp> {
     return success();
   }
 };
+
+class RecomposeSplitTensorGetItemOp
```
ramiro050, line 125: Can you add an e2e test that used to fail and now passes because of this pattern?

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/RecomposeComplexOps.cpp
```diff
@@ -121,6 +121,44 @@ class RecomposeSelectFill_ : public OpRewritePattern<AtenFill_TensorOp> {
...
+    : public OpRewritePattern<Aten__Getitem__TOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(Aten__Getitem__TOp op,
+                                PatternRewriter &rewriter) const override {
+    // recompose AtenSplitTensorOp + __getitem__t to AtenSliceTensorOp
+    auto splitTensorOp =
+        dyn_cast<AtenSplitTensorOp>(op.getList().getDefiningOp());
```
mgehre-amd, line 133: Do you need 
```
    if (isListPotentiallyMutated(op.getList())) {
      return failure();
    }
```
here?

mgehre-amd (approved):

lib/Dialect/Torch/Transforms/RecomposeComplexOps.cpp
```diff
@@ -121,6 +121,46 @@ class RecomposeSelectFill_ : public OpRewritePattern<AtenFill_TensorOp> {
...
+    if (!splitTensorOp)
+      return failure();
+    if (isListPotentiallyMutated(op.getList()))
```
ramiro050, line 136: Same comment here as in your other PR: https://github.com/llvm/torch-mlir/pull/2058#discussion_r1192694123

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -588,6 +588,7 @@ def emit_with_mutating_variants(key, **kwargs):
     emit("aten::any.bool : (bool[]) -> (bool)")
     emit("aten::sort.int : (int[], bool) -> ()", has_canonicalizer=True)
     emit("aten::sort : (Tensor, int, bool) -> (Tensor, Tensor)")
+    emit("aten::split.Tensor : (Tensor, int, int) -> (Tensor[])", traits=["Pure"])
```
ramiro050, line 591: This is not a `pure` op, since it can raise an exception when the inputs are invalid.

ramiro050 (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1502,6 +1502,20 @@ void AtenSortIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+// AtenSplitTensorOp
+//===----------------------------------------------------------------------===//
+
+void AtenSplitTensorOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                                    MLIRContext *context) {
+  patterns.add(+[](AtenSplitTensorOp op, PatternRewriter &rewriter) {
+    if (!op.use_empty())
+      return failure();
+    rewriter.eraseOp(op);
+    return success();
```
ramiro050, line 1515: I can think of two main approaches that we can take to achieve this:

1. Insert `RuntimeAssertOp`s in this canonicalization pattern that verify the inputs before deleting the op: this is to ensure that a runtime error is still generated when inputs are invalid)
2. Remove the op in the `RecomposeComplexOps` pattern: since the pattern replaces the op with another op that also produces runtime errors when the inputs are invalid, there is no need to verify things in the `RecomposeComplexOps` before deleting

I personally think the second option is the simplest

ramiro050 (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
zhekunz2: The reason why I didn't remove `SplitTensorOp` in recompose pattern is that right now recompose pattern is trying to match on `__GetItem__Op`, removing `SplitTensorOp` would result in erasing unmatched op. It's probably okay though, because there's no other patterns trying to match on `SplitTensorOp`.

ramiro050: It should be fine to delete the `SplitTensorOp`. The order should be:

1. Replace `__getitem__Op`
2. Check that `SplitTensorOp.use_empty()`
3. `rewriter.eraseOp`

ramiro050 (approved):
LGTM!

PR #2106, opened 2023-05-09
CI: create PR for RollPyTorch updates

ashay:
Currently, we run just the Linux in-tree tests when the RollPyTorch
workflow runs, but this is insufficient since WHL files for macOS or
Windows are sometimes not uploaded by PyTorch, causing the RollPyTorch
action to pass but all subsequent torch-mlir CI tests to fail because of
the broken build.

The easiest way to validate the RollPyTorch action on all platforms is
to run the standard set of tests that we run for each submitted PR, so
this patch makes the RollPyTorch action submit a PR instead of
committing the changes to the main branch directly.  The PR is assigned
to a handful of folks for review, although this can be changed in the
future.

---

@powderluv @vivekkhandelwal1 I included y'all among the reviewers for the ongoing PRs, but let me know if you'd prefer not to be on the list (also feel free to create another PR to remove yourself any time).


.github/workflows/RollPyTorch.yml
```diff
@@ -106,8 +106,28 @@ jobs:
...
+          torchvision version: ${{ env.PTVISION_RELEASE }}
+        commit-message: |
+          update PyTorch version to ${{ env.PT_RELEASE }}
+
+           - torch version: ${{ env.PT_RELEASE }}
+           - torch commit hash: ${{ env.PT_HASH }}
+           - torchvision version: ${{ env.PTVISION_RELEASE }}
+        committer: Roll PyTorch Action <torch-mlir@users.noreply.github.com>
+        title: update PyTorch version to ${{ env.PT_RELEASE }}
+        reviewers: ashay, powderluv, vivekkhandelwal1
```
powderluv, line 130: this is awesome. Do you think we can mark as reviewed to auto-merge if all tests pass ?

ashay: That's a nice idea!  I believe we can have another workflow that listens to newly created PRs; then if the PR is from the RollPyTorch action updating a restricted set of files, then the workflow could use the GitHub CLI to approve ([`gh pr review --aprove`](https://cli.github.com/manual/gh_pr_review)) and auto-merge ([`gh pr merge --auto`](https://cli.github.com/manual/gh_pr_merge)) the PR pending status checks.

I'd prefer to do that in stages, though.  I'm afraid of making too many changes in one PR, so if you're okay with it, I'd like to land this first before adding the auto-approving / auto-merging logic.

powderluv (approved):

ramiro050:
Thanks for looking into this, @ashay!

PR #2118, opened 2023-05-12
TorchToTosa: Support casts from and to bf16

mgehre-amd:
This PR adds all legal casts (according to TOSA spec) from/to bf16 into the TorchToTosa lowering pass,
and it also adds the cast from i64 to i32 (which was missing).

I did not add end2end tests as per the discussion on discord [0] and because the torch-mlir end2end testing
framework is not ready for bf16:
- It doesn't have the compiler-rt builtins for f32 to bf16 conversions
- Numpy doesn't support bf16, so it cannot tell them apart from f16.

[0] https://discordapp.com/channels/636084430946959380/742573221882364009/1104136103314849844


eric-k256 (approved):
Looks good. From the Discord thread, I agree with the reasoning on no e2e tests. 
I'll take a note to get F16 added to the casts so the function will be fully aligned with the TOSA spec, but that can be done separately.

PR #2122, opened 2023-05-12
[tosa] support lowering basic torch binary ops with mixed dtypes

tatwaichong:
Lowering torch operations that allow different compatible data types in its operands to tosa end up generating invalid tosa IR with mixed data types. In tosa spec, certain operations (generally elementwise operations) require all operands to have the same data type.
Add wrapper functions for those elementwise tosa ops to perform op creation with type conversion if necessary.

please note that I used lit tests because of the limited e2e support for bf16.

### DIFF
--- a/include/torch-mlir/Conversion/TorchToTosa/TosaLegalizeCommon.h
+++ b/include/torch-mlir/Conversion/TorchToTosa/TosaLegalizeCommon.h
@@ -10,8 +10,11 @@
 #ifndef TORCHMLIR_CONVERSION_TORCHTOTOSA_TOSALEGALIZECOMMON_H
 #define TORCHMLIR_CONVERSION_TORCHTOTOSA_TOSALEGALIZECOMMON_H
 
-#include "mlir/IR/PatternMatch.h" // from @llvm-project
-#include "mlir/Support/LLVM.h"    // from @llvm-project
+#include "torch-mlir/Conversion/TorchToTosa/TosaLegalizeUtils.h"
+
+#include "mlir/Dialect/Tosa/IR/TosaOps.h" // from @llvm-project
+#include "mlir/IR/PatternMatch.h"         // from @llvm-project
+#include "mlir/Support/LLVM.h"            // from @llvm-project
 
 namespace mlir {
 namespace tosa {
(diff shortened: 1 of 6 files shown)


tatwaichong:
rebase to top-of-tree to solve the format conflict in lit tests.

lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
```diff
@@ -149,6 +149,27 @@ Value getTosaConstTensorSingleF32(PatternRewriter &rewriter, Operation *op,
   return const_op.getResult();
 }
 
+// Create a zero constant tensor of the desired type and shape.
+std::optional<Value> getZerosLikeTensor(PatternRewriter &rewriter,
+                                        Operation *op, Type type) {
+  RankedTensorType result_type = type.dyn_cast<RankedTensorType>();
```
mgehre-amd, line 155: nit: use camelCase
same on variables below

include/torch-mlir/Conversion/TorchToTosa/TosaLegalizeCommon.h
```diff
@@ -21,6 +23,21 @@ createOneDimTfIndices(PatternRewriter &rewriter, Operation *op,
                       SmallVector<int64_t> indiceOneDimShape, int32_t dim,
                       ArrayRef<int64_t> indexShape);
 
+std::optional<Value> createMulOpAndCast(PatternRewriter &rewriter,
+                                        Operation *op, TensorType outType,
+                                        Value lhs, Value rhs, int32_t shift);
+
+// Create TOSA elementwise binary op with type conversion if necessary.
+template <typename TosaOpT>
+std::optional<Value> createBinaryOpAndCast(PatternRewriter &rewriter,
```
mgehre-amd, line 32: why does this return `std::optional<Value>` instead of a stronger `TosaOpT`? `CreateOpAndInfer` also returns `TosaOpT`

include/torch-mlir/Conversion/TorchToTosa/TosaLegalizeCommon.h
```diff
@@ -21,6 +23,21 @@ createOneDimTfIndices(PatternRewriter &rewriter, Operation *op,
                       SmallVector<int64_t> indiceOneDimShape, int32_t dim,
                       ArrayRef<int64_t> indexShape);
 
+std::optional<Value> createMulOpAndCast(PatternRewriter &rewriter,
```
mgehre-amd, line 26: I think the return type can be a stronger `tosa::MulOp` here

mgehre-amd (approved):

lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
tatwaichong: switch to use camelCase style.
though I personally like to use `_` more :), it appears LLVM recommends camelCase that mentioned in their document.

include/torch-mlir/Conversion/TorchToTosa/TosaLegalizeCommon.h
tatwaichong: Agree with you. These are createOp functions that intend to return an operation but not conversion-kind functions that transfer a complicated operation into a sequence of tosa ops. I think the latter focus more on the result (value).

Change to return tosa operation.

PR #2123, opened 2023-05-12
Fix canonicalization of `torch.prim.TupleUnpack`.

arames:


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2043,7 +2043,13 @@ void PrimTupleUnpackOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
     if (!tupleConstruct)
       return failure();
 
-    rewriter.replaceOp(op, tupleConstruct.getElements());
+    llvm::SmallVector<Value> derefinedElements;
+    for (auto [type, element] :
+         llvm::zip(op.getResultTypes(), tupleConstruct.getElements())) {
+      derefinedElements.push_back(
+          rewriter.createOrFold<DerefineOp>(op.getLoc(), type, element));
```
arames, line 2050: We need to maintain the correct type for the value.

IIRC, there is a TODO somewhere in the MLIR pattern rewriter to re-enable a check forcing rewrites to maintain exact types. That would have caught this.

Let me know if `DerefineOp` is not the right op for the type conversion. I took it from what I was seeing in our model.

test/CMakeLists.txt
ramiro050, line 1: Why are these needed?

test/Dialect/Torch/canonicalize.mlir
ramiro050, line 1: Can you also add an e2e test that used to fail and now passes with these changes?

ramiro050 (changes requested):

test/CMakeLists.txt
arames: This is just a fix on the way.

I added `torch-mlir-capi-torch-test`, and sorted the dependencies alphabetically.

Without this, `ninja check-torch-mlir` from a clean build directory fails because it is missing the executable:
```
******************** TEST 'TORCH_MLIR :: CAPI/torch.c' FAILED ********************
Script:
--
: 'RUN: at line 10';   torch-mlir-capi-torch-test 2>&1 | FileCheck /home/arames/work/nn-mlir/torch-mlir/test/CAPI/torch.c
--
Exit Code: 1

Command Output (stderr):
--
/home/arames/work/nn-mlir/torch-mlir/test/CAPI/torch.c:57:17: error: CHECK-LABEL: expected string not found in input
// CHECK-LABEL: testTypeMetaDataAccessors
                ^
<stdin>:1:1: note: scanning from here
/home/arames/work/nn-mlir/build/torch-mlir/tools/torch-mlir/test/CAPI/Output/torch.c.script: line 1: torch-mlir-capi-torch-test: command not found
```

While testing to reply to this comment, I also noticed `FileCheck` was missing as dependency in `CAPI/CMakeLists.txt`.

test/Dialect/Torch/canonicalize.mlir
arames: @ramiro050 : I'm having trouble creating a standalone test case.

The following does involve the issue, but fails on another error with the fix (`tensor.optional` is not allowed as a func return type).
```
class TupleModule(torch.nn.Module):
    def __init__(self):
        super().__init__()

    @export
    @annotate_args([
        None,
    ])

    def forward(self, a, b):
        cond = True
        if cond:
            tuple = a, b
        else:
            tuple = a + b, None
        _, y = tuple
        return y


@register_test_case(module_factory=lambda: TupleModule())
def TupleModule_basic(module, tu: TestUtils):
    module.forward(tu.rand(2, 2), tu.rand(2, 2))
```

We observed this deep within some Pytorch function, and also hitting the not-allowed optional return type after. So not too sure how to proceed.

lib/Dialect/Torch/IR/TorchOps.cpp
ramiro050, line 2050: Can you add a small comment here explaining that TorchScript will sometimes have the result type be a supertype of the tuple element, so a derefine is needed

ramiro050 (changes requested):

test/Dialect/Torch/canonicalize.mlir
ramiro050: Thanks for the small reproducer. I think we should be able to handle this e2e test in torch-mlir. Can you try adding handling for `DerefineOp` similar to `TensorStaticInfoCastOp` here:

https://github.com/llvm/torch-mlir/blob/c76a48308e17dc6a91c9b07f0adc96053e411c44/lib/Dialect/Torch/Transforms/RefinePublicReturn.cpp#L67-L68

(there are some comments near that line that will need updating too)

This should update the return type to be the refined type rather than the derefined one.

ramiro050: I agree that there is no need for a numerical correctness check. My interest in the e2e test was more for documenting how such a case for `TupleUnpack` is encountered in PyTorch. Just from looking at the lit test, it is not clear to me why we would ever end up with

```
%optional_tensor = torch.prim.TupleUnpack %tuple : !torch.tuple<tensor> -> !torch.optional<tensor>
```

arames: Thanks for the suggestion. I'm busy with something else right now, but will try it soon.

arames: > Thanks for the small reproducer. I think we should be able to handle this e2e test in torch-mlir. Can you try adding handling for `DerefineOp` similar to `TensorStaticInfoCastOp` here:
> 
> https://github.com/llvm/torch-mlir/blob/c76a48308e17dc6a91c9b07f0adc96053e411c44/lib/Dialect/Torch/Transforms/RefinePublicReturn.cpp#L67-L68
> 
> (there are some comments near that line that will need updating too)
> 
> This should update the return type to be the refined type rather than the derefined one.

This works nicely for a unit-test, but hits the issue that `AdjustCallingConventions` runs before `RefinePublicReturn`.
I have not spent much time looking yet.
Should we postpone the signature check to a separate later pass?

ramiro050: Yeah, postponing the return type fix works for me

arames: This seems to be working nicely!

test/Dialect/Torch/refine-public-return.mlir
```diff
@@ -3,10 +3,11 @@
 // CHECK-LABEL:   func.func @basic(
 // CHECK-SAME:                %[[ARG:.*]]: !torch.vtensor<[2,3,?],f32>) -> !torch.vtensor<[2,3,?],f32> {
 // CHECK:           return %[[ARG]] : !torch.vtensor<[2,3,?],f32>
-func.func @basic(%arg0: !torch.vtensor<[2,3,?],f32>) -> !torch.tensor {
+func.func @basic(%arg0: !torch.vtensor<[2,3,?],f32>) -> !torch.optional<tensor> {
```
ramiro050, line 6: Can you add new tests instead of changing these? Lit tests should ideally have a single responsibility regarding what they are testing. This makes debugging test failures easier and also makes it clear from looking at the test what is being checked.

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
```diff
@@ -186,53 +187,11 @@ class AdjustCallingConventionForReturn
...
-
-  return func->emitError(
-    "Functions must return a single tensor-like value, multiple tensor-like "
-    "values, or a tuple of more than one tensor-like value. Tensor-like "
-    "values: tensors, scalars, bools, and Nones.");
-}
-
 static LogicalResult adjustCallingConventions(func::FuncOp func,
                                               TypeBoundMap &typeBoundMap) {
-  if (failed(validateReturns(func)))
```
ramiro050, line 234: This code is a remnant of when we used to only run once the pipeline 

https://github.com/llvm/torch-mlir/blob/5223f990dfff28a1927ef61d70f58f6da2fc9138/lib/Dialect/Torch/Transforms/Passes.cpp#L33

When that was the case, it was important for passes to verify success before finishing and moving on to the next pass. However, we now run the lowering pipeline over and over until a valid IR satisfying the backend contract is reached. This means that we should remove the verification from these passes and move it to 

https://github.com/llvm/torch-mlir/blob/5223f990dfff28a1927ef61d70f58f6da2fc9138/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L176

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/AdjustCallingConventions.cpp
arames: Ok, done.
Looking at the code with your comment as context, it looks like that code can actually be removed.
The current backend contract already allows what we care about, and has clear explanations on the different expectations.

I added a unit test in particular for what we care about in this PR, plus a few extra on the way.

ramiro050 (approved):
Thanks! LGTM

include/torch-mlir/Dialect/Torch/Transforms/Passes.td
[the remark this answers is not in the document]
arames: Reverted.

PR #2129, opened 2023-05-16
[Stablehlo] add e2e test for aten.batch_norm

qingyunqu:


qingyunqu:
@ramiro050 Could you please review this PR? This PR will change code more than Stablehlo backend.

lib/Conversion/TorchToStablehlo/Basic.cpp
```diff
@@ -931,10 +931,10 @@ LogicalResult ConvertAtenOp<AtenBatchNormOp>::matchAndRewrite(
   Value momentum = adaptor.getMomentum();
   (void)momentum;
 
-  if (inputTy.getRank() <= 2) {
-    return rewriter.notifyMatchFailure(op,
-                                       "input should have rank larger than 2");
-  }
+  // handle feature index, see torch's BatchNorm1d, BatchNorm2d, BatchNorm3d,
+  // all of NC, NCL, NCHW, NCDHW's feature index is 1.
+  int64_t feature_index = 1;
```
ashay, line 936: Pardon my ignorance, but could you elaborate why the feature index is always 1?

qingyunqu: You could see torch's documents:
https://pytorch.org/docs/stable/generated/torch.nn.functional.batch_norm.html?highlight=batch_norm#torch.nn.functional.batch_norm
https://pytorch.org/docs/stable/generated/torch.nn.BatchNorm1d.html#torch.nn.BatchNorm1d

PyTorch has BatchNorm1d/2d/3d. BatchNorm1d has layout NC and NCL, BatchNorm2d has layout NCHW, BatchNorm3d has layout NCDHW. All of them have the C dim at index 1. So the feature index is always 1.

qingyunqu: BTW, I didn't see any description about Channel Last format in batch norm.

tools/torch-mlir-opt/CMakeLists.txt
```diff
@@ -16,3 +16,7 @@ target_link_libraries(torch-mlir-opt PRIVATE
   ${dialect_libs}
   ${conversion_libs}
 )
+
+if(TORCH_MLIR_ENABLE_STABLEHLO)
+  target_link_libraries(torch-mlir-opt PRIVATE MhloDialect AllMhloPasses)
```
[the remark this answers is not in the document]
ashay: Let's keep the dependencies to a minimum instead of linking `AllMhloPasses`.  If I understand correctly, the new pass you added (`TestUnfuseBatchNormPass`) is defined in mhlo/transforms/unfuse_batch_norm)/unfuse_batch_norm_pass.cc, which should be included in the `MhloPasses` library, which we link against in [`lib/CMakeLists.txt`](https://github.com/llvm/torch-mlir/blob/main/lib/CMakeLists.txt#L25-L29).

Which symbols do we miss if we don't link `MhloDialect` and `AllMhloPasses`?

lib/Conversion/TorchToStablehlo/Basic.cpp
ashay: Ah, I see.  Thanks for explaining!  It makes sense now.

tools/torch-mlir-opt/CMakeLists.txt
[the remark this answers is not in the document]
qingyunqu: `mhlo::registerAllMhloPasses()` has passes more than `MhloPasses` library. So if we register it in `torch-mlir-opt`, we must link `AllMhloPasses`. 
Another method is only register `TestUnfuseBatchNormPass` in `torch-mlir-opt`. BTW, we need register all mhlo passes registered in `InitAll.cpp`, including `registerSymbolicShapeOptimizationPass` and so on. Otherwise we couldn't reproduce the error from `torch-mlir-opt`.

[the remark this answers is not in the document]
qingyunqu: I choose the second way. Please review again, thanks! @ashay

ashay (approved):
LGTM

ramiro050 (approved):

Issue #2084, opened 2023-05-02
Primitive constant of tuple type support request


dvlp-r:
## Description

This issue is about the unsupported primitive constant of tuple type, @makslevental,  which leads to the following error:

```
Traceback (most recent call last):
  File "/Users/dvlpr/torch-mlir/examples/gin.py", line 107, in <module>
    module = torch_mlir.compile(gin, (x, edge_index, edge_attr, batch_f), output_type="linalg-on-tensors")
  File "/Users/dvlpr/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/__init__.py", line 377, in compile
    raise Exception(f"""
Exception: 
PyTorch TorchScript module -> torch-mlir Object Graph IR import failed with:
### Importer C++ Exception:
see diagnostics
### Importer Diagnostics:
error: unhandled prim::Constant node: %37 : (int, int) = prim::Constant[value=(0, 0)]()
```

## To reproduce the error

In order to reproduce the error, simply follow the installation of torch-mlir from source provided on the GitHub repository - https://github.com/llvm/torch-mlir/blob/main/docs/development.md, installation from source - plus manual installation in the mlir_venv of torch geometric and ogb through pip (`pip install ogb`, `pip install torch_geometric`).

The final environment should look like this

```
Package            Version
------------------ ------------------
certifi            2022.12.7
charset-normalizer 3.1.0
cmake              3.26.3
dill               0.3.6
filelock           3.12.0
fsspec             2023.4.0
idna               3.4
Jinja2             3.1.2
joblib             1.2.0
littleutils        0.2.2
MarkupSafe         2.1.2
mpmath             1.3.0
multiprocess       0.70.14
networkx           3.1
ninja              1.11.1
numpy              1.24.3
ogb                1.3.6
outdated           0.2.2
pandas             2.0.1
Pillow             9.5.0
pip                23.1.2
psutil             5.9.5
pybind11           2.10.4
pyparsing          3.0.9
python-dateutil    2.8.2
pytz               2023.3
PyYAML             6.0
requests           2.29.0
scikit-learn       1.2.2
scipy              1.10.1
setuptools         65.5.0
six                1.16.0
sympy              1.12rc1
threadpoolctl      3.1.0
torch              2.1.0.dev20230427
torch-geometric    2.3.1
torch-scatter      2.1.1
torch-sparse       0.6.17
torch-spline-conv  1.2.2
torchvision        0.16.0.dev20230427
[...]

PR #2132, opened 2023-05-17
[Importer] import constant tuple

qingyunqu:
This PR will fix https://github.com/llvm/torch-mlir/issues/2084. But there is another thing failed when I add e2e test.

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -540,6 +540,7 @@ STABLEHLO_PASS_SET = {
     "NormScalarOptDimKeepDimModule_basic",
     "NormScalarOptDimModule_basic",
     "NormalizeModule_basic",
+    "ScalarConstantTupleModule_basic",
     "SelectIntModule_basic",
     "SelectIntNegativeDimAndIndexStaticModule_basic",
     "SliceSingleIdxModule_basic",
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -268,6 +268,9 @@ TORCHDYNAMO_XFAIL_SET = {
     "ScatterValueFloatModule_basic",
     # ERROR: torch._dynamo.exc.Unsupported: call_function BuiltinVariable(int) [TensorVariable()] {}
     "ScatterValueIntModule_basic",
+
+    # ERROR: encountered graph that does not return anything
+    "ScalarConstantTupleModule_basic",
 }
 
 TORCHDYNAMO_CRASHING_SET = {
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -192,6 +192,7 @@ TORCHDYNAMO_XFAIL_SET = {
     'IsFloatingPointInt_False',
     'TorchPrimLoopForLikeModule_basic',
     'TorchPrimLoopWhileLikeModule_basic',
+    "ScalarConstantTupleModule_basic",
     # END tests failing due to: empty graph in dynamo
 
     # ERROR due to: backend never runs because of empty frame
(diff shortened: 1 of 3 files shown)


ramiro050:
What e2e test are you using and what is the error?

qingyunqu:
> What e2e test are you using and what is the error?

I will show the case tomorrow.

qingyunqu:
> What e2e test are you using and what is the error?

```python
import torch
import torch_mlir

class TupleModule(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.value = (1, 1)

    def forward(self, x):
        self.value = (1, 2)
        return torch.amax(x, self.value[0])

model = TupleModule().eval()
module = torch_mlir.compile(model, torch.randn(3, 4), output_type="torch")
print(module.operation.get_asm())
```

Error message:
```
error: unknown: unsupported by backend contract: module initializers
note: unknown: see current operation: "torch.initialize.global_slots"(%3) {slotSymNames = [@value]} : (!torch.tuple<int, int>) -> ()
note: unknown: this is likely due to InlineGlobalSlots being unable to inline a global slot

For Torch-MLIR developers, the error can be reproduced with:
$ torch-mlir-opt -pass-pipeline='builtin.module(torchscript-module-to-torch-backend-pipeline{backend-legal-ops= extra-library=})' /tmp/TupleModule.mlir
Add '-mlir-print-ir-after-all -mlir-disable-threading' to get the IR dump for debugging purpose.
```

ramiro050:
The `InlineGlobalSlots` pass fails because mutation of module parameters is not supported:

https://github.com/llvm/torch-mlir/blob/aa97c8383e96db645b579f58599255f6f41fa1b9/lib/Dialect/Torch/Transforms/InlineGlobalSlots.cpp#L85-L94

However, you should be able to just do

```
class TupleModule(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self):
        return (1, 2)
```

to test the functionality of importing a tuple constant.

python/torch_mlir/dialects/torch/importer/jit_ir/csrc/node_importer.cpp
```diff
@@ -233,6 +233,13 @@ void NodeImporter::importNode(Node *node, MlirBlock appendToBlock,
                        dummyAnnotator, importOptions);
       mapResults(node, mlirOpResultGetOwner(listValue));
       return; // Early return, since `importIValue` already added op to block.
+    } else if (output->type()->cast<c10::TupleType>()) {
```
ramiro050, line 236: This is the same code as the `ListType` case above. We should just change the `if` condition and reuse the logic.

ramiro050 (changes requested):

dvlp-r:
Hi, there is any news about it? @qingyunqu are you still working on it?

qingyunqu:
> Hi, there is any news about it? @qingyunqu are you still working on it?

This PR is ready now.

e2e_testing/xfail_sets.py
```diff
@@ -540,6 +543,7 @@
     "NormScalarOptDimKeepDimModule_basic",
     "NormScalarOptDimModule_basic",
     "NormalizeModule_basic",
+    "ScalarConstantTupleModule_basic",
```
ramiro050, line 546: The dynamo tests should either be `xfail` or `crashing`. From the error message, it seems like it is not crashing.

ramiro050 (approved):
LGTM, just a small change request

e2e_testing/xfail_sets.py
qingyunqu: I think this change is adding pass case on `STABLEHLO_PASS_SET`. It's not about dynamo.

ramiro050: Ah, sorry about that. The way github displays it confused me

PR #2138, opened 2023-05-20
update PyTorch version to 2.1.0.dev20230519

silvasean:
torch version: 2.1.0.dev20230519
torch commit hash: 61239df555df02e8c60a2ad63363878a2a57c161
torchvision version: 0.16.0.dev20230520

### DIFF
--- a/torchvision-requirements.txt
+++ b/torchvision-requirements.txt
@@ -1,3 +1,3 @@
 -f https://download.pytorch.org/whl/nightly/cpu/torch_nightly.html
 --pre
-torchvision==0.16.0.dev20230519
+torchvision==0.16.0.dev20230520


ashay:
Closed this PR because the latest torchvision WHL (v0.16.0.dev20230520) incorrectly depends on torch from yesterday:

```
> grep "Requires-Dist: torch" torchvision-0.16.0.dev20230520+cpu.dist-info/METADATA
Requires-Dist: torch (==2.1.0.dev20230519)
```

However, the torchvision WHL file for Python v3.10 correctly depends on today's torch package, and RollPyTorch uses Python 3.11 whereas the build tests use Python 3.10.  This version mismatch caused the build tests to fail but RollPyTorch action to succeed.  I will send a PR to fix these issues later today.

PR #2140, opened 2023-05-20
CI: Align Python version between RollPyTorch and Build Tests

ashay:
The RollPyTorch action uses Python v3.11 (so does the Release build),
but the workflow that runs the CI tests uses Python v3.10.  This can
cause subtle failures, so this patch makes all three of these workflows
use Python v3.11.

This patch also fixes a bug in the RollPyTorch action so that it skips
creating a PR if the pytorch commit hash has not changed.

### DIFF
--- a/.github/workflows/RollPyTorch.yml
+++ b/.github/workflows/RollPyTorch.yml
@@ -98,7 +98,7 @@ jobs:
           ```
 
     - name: Update PyTorch Build Cache (if running on main branch)
-      if: github.ref_name == 'main'
+      if: github.ref_name == 'main' && env.PT_HASH_CHANGED != '0'
       id: cache-pytorch
       uses: actions/cache@v3
       with:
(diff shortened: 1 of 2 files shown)


ashay:
@powderluv It looks like the Docker image has Python 3.10.  There's no urgency, but on Monday or later, do you think you could update it to use Python 3.11 so that we use the same Python version for all workflows?  The alternative would be to make the Release and RollPyTorch actions switch to Python 3.10.

From the CI logs for this PR:
```
******************** BUILDING PACKAGE out-of-tree (docker) ************
Could not find python: /opt/python/cp311-cp311 (using system default Python3)
Defaulting to /usr/bin/python3 (expected for CI builds)
:::: Python version Python 3.10.4
```

powderluv:
Yes we should update to 3.11. I'll try and get to it this weekend or early next week

ashay:
> I'll try and get to it this weekend or early next week

Sure, thanks!  No rush.

PR #2143, opened 2023-05-22
[Torch Dialect] add missing one_hot dtype function

qingyunqu:
There is an e2e test for OneHot, it passed because there is a decomposition which doesn't rely on dtype.


qingyunqu:
@ramiro050 Cloud you please review this PR?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -2641,6 +2641,10 @@ def aten〇native_layer_norm〡dtype(input_rank_dtype: Tuple[int, int], normaliz
         result_dtype = torch.float64
     return input_dtype, input_dtype, result_dtype
 
+@check_dtype_function([Invocation(TensorOfShape(5, dtype=torch.int64, device="cpu"), 2)])
```
ramiro050, line 2644: Can you do

```python
@check_dtype_function(_check_tensors_with_the_same_dtype(num_of_tensors=1, num_classes=2))
```

This performs a much more thorough check

ramiro050 (changes requested):

ramiro050:
> @ramiro050 Cloud you please review this PR? (BTW, how can I request a reviewer on the web page directly?

The permissions were set wrong. You should be able to request reviewers now. Sorry about that

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
qingyunqu: I think it's no need to add this check, torch defines that the input/output tensor must be int64 dtype. See:
https://pytorch.org/docs/stable/generated/torch.nn.functional.one_hot.html

I will add `assert` check in dtype function.

ramiro050: Sometimes docs can become out of date and inconsistent with actual behavior of functions. The only way to know that things work properly is by actually running them. This also makes it easy to know when API changes happen in PyTorch that affect dtype calculations. I would prefer to use the comprehensive check.

qingyunqu: Done, could you please reivew this PR again? Thanks!

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -2699,6 +2699,12 @@ def aten〇native_layer_norm〡dtype(input_rank_dtype: Tuple[int, int], normaliz
         result_dtype = torch.float64
     return input_dtype, input_dtype, result_dtype
 
+@check_dtype_function(_check_tensors_with_the_same_dtype(num_of_tensors=1, num_classes=2, tensor_device="cpu", error_types={torch.complex128, torch.complex64, torch.float64, torch.float32, torch.float16, torch.bfloat16, torch.int32, torch.int16, torch.int8, torch.uint8, torch.bool}))
```
ramiro050, line 2702: nit: can you add a comment here about why `tensor_device="cpu"` is needed?

ramiro050 (approved):
Thanks!

PR #2144, opened 2023-05-22
[TORCH-MLIR] Add ODS for aten.sign op.

prateekgu-cerebras:
This commit adds ODS for the aten.sign op.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -340,6 +340,51 @@ def Torch_AtenSigmoid_Op : Torch_Op<"aten.sigmoid_", [
   }];
 }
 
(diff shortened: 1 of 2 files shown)


pashu123 (approved):
NIT: If you could add the result dtype and shape in https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py that would be good. Rest LGTM!

rahuls-cerebras (approved):

prateekgu-cerebras:
> NIT: If you could add the result dtype and shape in https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py that would be good. Rest LGTM!

Hi @pashu123 we can take this in another PR. This one is only meant to add ODS in the torch-mlir. Thanks!

PR #2154, opened 2023-05-24
[Stablehlo] enhance e2e test framework with hlo-legalize-to-memref pass

qingyunqu:
Note: this is a way with a little hack. I think we should use `one-shot-bufferization` on e2e test in the future.


python/torch_mlir_e2e_test/stablehlo_backends/linalg_on_tensors.py
```diff
@@ -40,7 +40,19 @@ def compile(self, imported_module: Module):
...
+                "canonicalize",
+                "cse",
+                "func.func(symbolic-shape-optimization)",
+                "func.func(mhlo-test-unfuse-batch-norm)",
+                "canonicalize",
+                "func.func(hlo-legalize-to-linalg)",
+                "func.func(remove-shape-constraints)",
+                "canonicalize",
+                # Note: This will generate memref dialect. Suggest to replace with one-shot-bufferization.
+                "hlo-legalize-to-memref"]) + ")",
```
ramiro050, line 55: @silvasean, can you help me review this? Are there any constraints to what the e2e test backends should output? (memref vs tensor)

silvasean: We don't have a verifier for the "linalg on tensors backend contract" which is accepted by the RefBackend (and is a proxy for other systems like IREE), but if we did, it would fail with this change. So I would prefer we not merge this since it breaks the layering of the system (introduces memrefs where there should not be any).

It seems like we are going to be churning this soon based on the StableHLO/MHLO situation (see [here](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/KsRp9euuuB0/m/w2h74Ag_AgAJ?utm_medium=email&utm_source=footer)). So I would prefer for us to re-evaluate it once that lands (which I think should fix this, or at least cause somebody working on that to fix it on their end, since this layering violation is an issue in StableHLO/MHLO really).

@qingyunqu -- is this blocking you currently? We may be able to find a shorter-term solution if you need it sooner.

qingyunqu: Thanks! It doesn't blocking me. I also think it's better to fix it on StableHLO.

I think one possible way is to use `one-shot-bufferization`.

PR #2157, opened 2023-05-24
[arm64] Fix release builds for ARM64

powderluv:
Tested on Ubuntu 23.04 on Ampere Altra instance.

### DIFF
--- a/build-requirements.txt
+++ b/build-requirements.txt
@@ -4,6 +4,7 @@ wheel
 setuptools
 cmake
 ninja
+packaging
 
 # Workaround for what should be a torch dep
 # See discussion in #1174
--- a/build_tools/python_deploy/build_linux_packages.sh
+++ b/build_tools/python_deploy/build_linux_packages.sh
@@ -38,8 +38,10 @@ set -eu -o errtrace
 
 this_dir="$(cd "$(dirname "$0")" && pwd)"
 repo_root="$(cd "$this_dir"/../../ && pwd)"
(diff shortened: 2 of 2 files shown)


powderluv:
There are no nightlies for arm64 from PyTorch so we have to use the stable builds (or build from source)

makslevental:
> This looks good to me, but is this part of a larger change? We don't have ARM64 Linux release builds, do we?

I don't know the answer to the first question and the answer to the second is no but I'm just here to say that we could  [could](https://github.com/nod-ai/PI/releases/tag/torch-mlir-96327f59) using this [workflow](https://github.com/nod-ai/PI/blob/main/.github/workflows/build_torch_mlir.yml). And beyond that (as I've mentioned before) we could streamline the entire build/test/release process by moving to only have only out-of-tree builds and using a triggered build of LLVM like I'm doing in that [workflow for torch-mlir](https://github.com/nod-ai/PI/blob/main/.github/workflows/build_torch_mlir.yml#L8-L14).

powderluv:
> but is this part of a larger change? We don't have ARM64 Linux release builds, do we?

Sorry should have mentioned. Yeah I am slowly working towards enabling ARM64 builds, we don't have it in the GHA yet. Once this lands I will bring up an arm64 builder and try to start pushing PRs for it.

powderluv:
Nightly builds pass with this change too https://github.com/llvm/torch-mlir/actions/runs/5072926336/jobs/9111281918

PR #2159, opened 2023-05-24
Add ARM64 release builds

powderluv:
Creates a build_linux_arm64 job that builds the release on an arm64 self-hosted runner.  Drop Python 3.10 support
Pass  TM_TORCH_VERSION to choose the Stable PyTorch version (since arm64 doesn't have nightly builds)

### DIFF
--- a/.github/workflows/buildRelease.yml
+++ b/.github/workflows/buildRelease.yml
@@ -12,12 +12,12 @@ on:
 
 jobs:
   build_linux:
-    name: Manylinux Build
+    name: Manylinux x86_64 Build
     runs-on: a100
     strategy:
       matrix:
         package: [ torch-mlir, torch-mlir-core ]
-        py_version: [ cp38-cp38, cp310-cp310, cp311-cp311 ]
+        py_version: [ cp38-cp38, cp311-cp311 ]
         exclude:
           - package: torch-mlir-core
             py_version: cp38-cp38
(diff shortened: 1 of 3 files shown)


powderluv:
This depends on https://github.com/llvm/torch-mlir/pull/2038 for the PyTorch Stable builds

.github/workflows/buildRelease.yml
```diff
@@ -86,6 +86,76 @@ jobs:
...
+
+    - name: Get torch-mlir
+      uses: actions/checkout@v3
+      with:
+        submodules: 'true'
+        fetch-depth: 0
+
+    - uses: ./.github/actions/setup-build
+      with:
+        cache-suffix: 'release'
```
ashay, line 113: We should use a different suffix so that the key for cache artifacts does not collide between x64 and arm64 builds.  Tangentially related, do we want to disable caching for release builds?

.github/workflows/buildRelease.yml
```diff
@@ -86,6 +86,76 @@ jobs:
         name: wheels
         path: dist
 
+  build_linux_arm64:
```
ashay, line 89: In a separate PR, is it worthwhile to put the parts that are common across all hosts into a separate, parameterized action?  It is getting harder to know what changed between the build_linux and build_linux_arm64 jobs.

ashay (approved):
LGTM, but we should update the cache suffix between the x64 and arm64 jobs.

.github/workflows/buildRelease.yml
powderluv: Yeah this may be a nice cleanup

.github/workflows/buildRelease.yml
powderluv: Good catch. We should just disable for release builds

ashay: Addressed in https://github.com/llvm/torch-mlir/pull/2168, although it looks like this PR might need to be rebased.

powderluv:
@maxbartel I can't seem to add you as a reviewer but adding @mgehre-amd

This patch pulls in a part of your changes in https://github.com/llvm/torch-mlir/pull/2038 to unblock the ARM64 builder so we can build nightly builds with stable / nightly not just CI.  It also removes the need to rename all the requirements.txt files in the original PR. If land this it should drastically simplify https://github.com/llvm/torch-mlir/pull/2038 to add the testing based on if it is stable / nightly.

.github/workflows/buildRelease.yml
```diff
@@ -44,11 +44,80 @@ jobs:
...
+
+    - name: Get torch-mlir
+      uses: actions/checkout@v3
+      with:
+        submodules: 'true'
+        fetch-depth: 0
+
+    - uses: ./.github/actions/setup-build
+      with:
+        cache-suffix: 'release'
```
ashay, line 112: This can now be cache-enabled: 'false'.

build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -295,14 +298,33 @@ function test_in_tree() {
 
 function setup_venv() {
   local python_version="$1"
-  echo ":::: Setting up VENV with Python: $python_version"
+  local torch_version="$2"
+
+  echo ":::: Setting up VENV with Python: $python_version PyTorch $2"
```
ashay, line 303: Nit: I presume you meant to use $torch_version instead of $2.

ashay (approved):
LGTM, save for some small changes.

.github/workflows/buildRelease.yml
powderluv: ah nice. will update

build_tools/python_deploy/build_linux_packages.sh
powderluv: ah true. both worked. will update.

PR #2160, opened 2023-05-24
Add alias analysis for cast-like ops to maximize-value-semantics

ramiro050:
When `use_tracing=True` is used to import a model into Torch-MLIR, several casts get inserted in the IR to bridge the untyped inputs and outputs with the typed body of the computation. These casts create extra aliases of tensors that cause the current analysis in `maximize-value-semantics` to fail.

In particular, the `maximize-value-semantics` analysis assumes that the only valid alias right after an overwrite is the overwritten alias. So, if there is a use of a casted version of the overwritten alias after the overwrite, the analysis fails.

This commit improves the analysis by identifying all cast-like aliases of the overwritten alias and allowing such aliases to be used after an overwrite.

Because this issue only arises when using tracing, it cannot be currently tested e2e, so only lit test is added.

### DIFF
--- a/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
+++ b/lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
@@ -28,6 +28,28 @@ static Value assertNonValueTensor(Value tensor) {
   return tensor;
 }
 
+// A cast-like op is an op that does not modify the contents, shape, and dtype
+// of the input tensor. In other words, it is an op that only serves to encode
+// compile time information, but at runtime the op behaves like a no-op.
+static bool isCastLikeOp(Operation *op) {
+  return isa<TensorStaticInfoCastOp>(op);
+}
+
+// Given a `value`, this function goes up the use-def chain and finds the
+// largest sequence of consecutive cast-like ops. The returned set contains all
+// the aliases that are identical to `value`, and have only been transformed by
+// cast-like ops.
+static DenseSet<Value> getCastLikeAliasesOf(Value value) {
+  Operation *currentOp = value.getDefiningOp();
+  DenseSet<Value> result;
+  while (isCastLikeOp(currentOp)) {
+    Value operand = assertNonValueTensor(currentOp->getOperand(0));
+    result.insert(operand);
+    currentOp = operand.getDefiningOp();
+  }
+  return result;
+}
+
 namespace {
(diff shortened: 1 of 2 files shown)


lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
```diff
@@ -28,6 +28,28 @@ static Value assertNonValueTensor(Value tensor) {
...
+// compile time information, but at runtime the op behaves like a no-op.
+static bool isCastLikeOp(Operation *op) {
+  return isa<TensorStaticInfoCastOp>(op);
+}
+
+// Given a `value`, this function goes up the use-def chain and finds the
+// largest sequence of consecutive cast-like ops. The returned set contains all
+// the aliases that are identical to `value`, and have only been transformed by
+// cast-like ops.
+static DenseSet<Value> getCastLikeAlisesOf(Value value) {
```
vivekkhandelwal1, line 42: Typo: `Alises` -> `Aliases`

lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
```diff
@@ -88,9 +110,13 @@ class AbstractlyInterpretCopyToNonValueTensorOpUsersWithinABlock
...
         // To simplify the analysis, we only support the case where the
         // only aliases used after an overwrite are the aliases generated
-        // after plus the alias being overwritten.
+        // after plus the alias being overwritten and any aliases that are
+        // simply a cast of the overwritten alias.
         availableAliases.clear();
-        availableAliases.insert(assertNonValueTensor(overwrite.getOverwritten()));
+        Value overwritten = overwrite.getOverwritten();
+        availableAliases.insert(assertNonValueTensor(overwritten));
+        DenseSet<Value> castLikeAliases = getCastLikeAlisesOf(overwritten);
```
vivekkhandelwal1, line 118: Similarly here

vivekkhandelwal1 (changes requested):
Rest LGTM!

mgehre-amd (approved):
LGTM, thanks!

lib/Dialect/Torch/Transforms/MaximizeValueSemantics.cpp
ramiro050: Thanks! :)

vivekkhandelwal1 (approved):

PR #2161, opened 2023-05-25
[TCP] Support for tcp.cast op

zezhang:
As titled.

Testing:

```
./build/bin/llvm-lit externals/llvm-external-projects/torch-mlir-dialects/test/Conversion/TcpToLinalg/ -v
./build/bin/llvm-lit externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/ -v
./build/bin/llvm-lit test/Conversion/TorchToTcp/ -v
python -m e2e_testing.main --config=tcp -v -s
```


externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -514,4 +514,28 @@ def Tcp_Atan2Op : Tcp_BinaryElementwiseOp<"atan2", [SameOperandsAndResultElement
...
+
+  let summary = "TCP Cast operation";
+
+  let description = [{
+    Performs elementwise type conversion.
+  }];
+
+  let arguments = (ins
+    Tcp_Tensor:$in,
+    // Align with enum SignednessSemantics
```
navahgar, line 527: Is there a better way to handle attributes that are enum?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -514,4 +514,28 @@ def Tcp_Atan2Op : Tcp_BinaryElementwiseOp<"atan2", [SameOperandsAndResultElement
...
+    OptionalAttr<UI32Attr>:$in_dtype,
+    OptionalAttr<UI32Attr>:$out_dtype
```
navahgar, line 529: Lets not call these `dtype`. I believe MLIR uses `elementType` instead.

Also, we should probably encode in the name that this is only meant for integer types.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
```diff
@@ -174,24 +174,108 @@ createLinalgPayloadForElementwiseOp(Operation *op,
...
+  if (auto castOp = dyn_cast<CastOp>(op)) {
+    auto inputType =
+        castOp.getIn().getType().dyn_cast<RankedTensorType>().getElementType();
+    auto outputType = resultTensorType.getElementType();
+    // check for same type
+    if (inputType.getIntOrFloatBitWidth() ==
+            outputType.getIntOrFloatBitWidth() &&
+        ((!castOp.getInDtype() && !castOp.getOutDtype()) ||
+         (castOp.getInDtype() && castOp.getOutDtype() &&
+          castOp.getInDtype().value() == castOp.getOutDtype().value())))
```
navahgar, line 206: nit: since you are adding `{ }` for the else cases below, it is better to add that here as well, to keep it uniform and more readable.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
```diff
@@ -174,24 +174,108 @@ createLinalgPayloadForElementwiseOp(Operation *op,
...
+      return payloadArgs[0];
+    // To I1 (Bool)
```
navahgar, line 208: nit: To make it more readable, consider formatting this like:

```
   } else if (outputType.isInteger(1)) { // to I1 (Bool)
```
OR
```
   } else if (outputType.isInteger(1)) {
      // to I1 (Bool)
```

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
```diff
@@ -174,24 +174,108 @@ createLinalgPayloadForElementwiseOp(Operation *op,
...
+      if (inputType.isa<mlir::FloatType>()) {
+        return {b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UNE,
+                                        payloadArgs[0], cstZero)};
+      } else if (inputType.isa<mlir::IntegerType>()) {
+        return {b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::ne,
+                                        payloadArgs[0], cstZero)};
+      }
+    }
+    // TO FP
+    else if (!castOp.getOutDtype()) {
```
navahgar, line 221: It is better to use the condition `outputType.isa<mlir::FloatType>()` here instead.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
```diff
@@ -174,24 +174,108 @@ createLinalgPayloadForElementwiseOp(Operation *op,
...
+                IntegerType::SignednessSemantics::Unsigned)
+          return {b.create<arith::UIToFPOp>(loc, outputType, payloadArgs[0])};
+        // Signed INT to FP
+        else if (castOp.getInDtype().value() ==
+                 IntegerType::SignednessSemantics::Signed)
+          return {b.create<arith::SIToFPOp>(loc, outputType, payloadArgs[0])};
+      }
+    }
+    // TO INT
+    else if (castOp.getOutDtype()) {
```
navahgar, line 245: same as above

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -417,13 +417,207 @@ class ConvertAtenAtan2Op : public OpConversionPattern<AtenAtan2Op> {
...
+
+    // The non_blocking arg should be a constant `False`.
+    bool nonBlocking;
+    if (!matchPattern(op.getNonBlocking(), m_TorchConstantBool(&nonBlocking))) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: non_blocking arg must be a constant");
+    } else if (nonBlocking) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: non_blocking arg is expected to be false");
+    }
```
navahgar, line 452: nit: To simplify the code, we can probably combine the two checks into one.

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -417,13 +417,207 @@ class ConvertAtenAtan2Op : public OpConversionPattern<AtenAtan2Op> {
...
+
+    // The copy arg should be a constant `False`.
+    bool copy;
+    if (!matchPattern(op.getCopy(), m_TorchConstantBool(&copy))) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: copy arg must be a constant");
+    } else if (copy) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: copy arg is expected to be false");
+    }
```
navahgar, line 462: same as above

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -417,13 +417,207 @@ class ConvertAtenAtan2Op : public OpConversionPattern<AtenAtan2Op> {
...
+      if (!matchPattern(op.getMemoryFormat(),
+                        m_TorchConstantInt(&memoryFormat)))
+        return rewriter.notifyMatchFailure(
+            op, "unimplemented: the memory format should be specified in "
+                "an integer constant");
+      if (memoryFormat != torch_upstream::MemoryFormat::Contiguous &&
+          memoryFormat != torch_upstream::MemoryFormat::Preserve)
+        return rewriter.notifyMatchFailure(
+            op, "unimplemented: only none, contiguous and preserve "
+                "memory_format is supported");
```
navahgar, line 476: same as above

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -417,13 +417,207 @@ class ConvertAtenAtan2Op : public OpConversionPattern<AtenAtan2Op> {
...
+    }
+
+    // FP -> FP
+    if (inputType.getDtype().isa<mlir::FloatType>() &&
+        outputType.getDtype().isa<mlir::FloatType>())
+      rewriter.replaceOpWithNewOp<tcp::CastOp>(
+          op, resultType, adaptor.getSelf(), IntegerAttr{}, IntegerAttr{});
+    // FP -> INT
+    else if (inputType.getDtype().isa<mlir::FloatType>()) {
+      if (outputType.getDtype().isSignlessInteger())
```
navahgar, line 486: Instead of checking for these conditions, you can directly use the method `getSignedness()` to get the signedness of the input and output types. That would simplify this and the code below a lot.

navahgar (commented):
Took one pass over it.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: @burmako What is the recommended approach for defining attributes that represent enums like this?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -514,4 +514,28 @@ def Tcp_Atan2Op : Tcp_BinaryElementwiseOp<"atan2", [SameOperandsAndResultElement
...
+  let summary = "TCP Cast operation";
+
+  let description = [{
+    Performs elementwise type conversion.
+  }];
+
+  let arguments = (ins
+    Tcp_Tensor:$in,
+    // TODO: Better way to align with enum SignednessSemantics
+    OptionalAttr<UI32Attr>:$in_int_element_type,
```
navahgar, line 528: If this is only meant to represent the `SignednessSemantics`, we should probably name it accordingly, say something like `in_int_signedness`. 

My earlier comment was just that `dtype` is not a term used in core MLIR.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -514,4 +514,28 @@ def Tcp_Atan2Op : Tcp_BinaryElementwiseOp<"atan2", [SameOperandsAndResultElement
   let assemblyFormat = "$in1 `,` $in2 attr-dict `:` type($in1) `,` type($in2) `->` type($out)";
 }
 
+def Tcp_CastOp:  Tcp_Op<"cast", [Pure, Elementwise, SameOperandsAndResultShape]> {
+
+  let summary = "TCP Cast operation";
+
+  let description = [{
+    Performs elementwise type conversion.
```
navahgar, line 522: Can you add more explanation as to the type of casts we do here?

This is probably a good place to document all casting cases that are covered.

navahgar (commented):
Mostly LGTM. Just a couple of minor comments to address.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
burmako: Here's how we're doing this in StableHLO:
  * Definition of an enum attribute: https://github.com/openxla/stablehlo/blob/d67da5d85045e03893684c35ae5e730695289ae2/stablehlo/dialect/StablehloEnums.td#L98-L122.
  * Use of this attribute: https://github.com/openxla/stablehlo/blob/d67da5d85045e03893684c35ae5e730695289ae2/stablehlo/dialect/StablehloOps.td#L1570.

navahgar: Thanks a lot for those pointers @burmako. Thats very helpful.

@zezhang Can we follow a similar approach to represent signedness in `tcp.cast`?

navahgar:
@burmako Can you please review this PR?

navahgar (commented):
LGTM. Thanks for taking this up. I don't have any other comments. 

I will let @burmako approve the PR.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -514,4 +515,41 @@ def Tcp_Atan2Op : Tcp_BinaryElementwiseOp<"atan2", [SameOperandsAndResultElement
...
+           ui32, si32, i32,
+           ui64, si64, i64, 
+           i1 (Boolean)
+      FP: f16, bf16, f32, f64
+
+    Supported Conversion Mode:
+      FP -> FP
+      INT -> FP
+      FP -> INT
+      INT -> INT
```
burmako, line 537: There are quite a few corner cases here, e.g. what happens if the conversion overflows the destination type or what happens if a negative number is converted to an unsigned destination type. Consider explicitly speccing them at some point.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -514,4 +515,41 @@ def Tcp_Atan2Op : Tcp_BinaryElementwiseOp<"atan2", [SameOperandsAndResultElement
...
+  }];
+
+  let arguments = (ins
+    Tcp_Tensor:$in,
+    OptionalAttr<Tcp_SignednessAttr>:$in_int_signedness,
+    OptionalAttr<Tcp_SignednessAttr>:$out_int_signedness
```
burmako, line 543: Have you considered an alternative design with multiple ops like in the Arith dialect? I.e. `sitofp`, `uitofp`, `fptosi`, `fptoui`, `extsi`, `extui`, `extf`, `trunci`, `truncf`?

externals/llvm-external-projects/torch-mlir-dialects/lib/Dialect/Tcp/IR/TcpOps.cpp
```diff
@@ -155,4 +156,43 @@ LogicalResult ConcatOp::verify() {
...
+    if (!getOutIntSignedness())
+      return emitOpError(
+          "out_int_signedness attr must be set when output is INT");
+    if (outputType.getElementType().isInteger(1) &&
+        getOutIntSignedness().value() != Signedness::Signless)
+      return emitOpError("out_int_signedness attr must be set to "
+                         "Signedness::Signless when output is i1");
+  }
+
+  return success();
```
burmako, line 195: What if inputType and/or outputType aren't integer or floating-point tensors? As it's currently written, the verifier will succeed but that's probably undesirable?

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
```diff
@@ -174,24 +174,97 @@ createLinalgPayloadForElementwiseOp(Operation *op,
...
+        if (inputType.getIntOrFloatBitWidth() >
+            outputType.getIntOrFloatBitWidth())
+          return {b.create<arith::TruncFOp>(loc, outputType, payloadArgs[0])};
+        return {b.create<arith::ExtFOp>(loc, outputType, payloadArgs[0])};
+      }
+      // INT -> FP
+      else if (inputType.dyn_cast<mlir::IntegerType>()) {
+        // Signless or Unsigned INT to FP
+        if (castOp.getInIntSignedness().value() == Signedness::Signless ||
+            castOp.getInIntSignedness().value() == Signedness::Unsigned)
```
burmako, line 234: Why does signless mean unsigned? This was a bit unexpected.

burmako (commented):
Reviewed the enum attributes part + left a few design questions. Please feel free to ignore those - they are mostly for my curiousity.

zezhang:
Thanks @burmako for reviewing this. I have addressed your comments. Could you take another look and approve it if everything looks good for now. Please let me know if you have any other questions.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
```diff
@@ -514,4 +515,41 @@ def Tcp_Atan2Op : Tcp_BinaryElementwiseOp<"atan2", [SameOperandsAndResultElement
...
 }
 
+def Tcp_CastOp:  Tcp_Op<"cast", [Pure, Elementwise, SameOperandsAndResultShape]> {
+
+  let summary = "TCP Cast operation";
+
+  let description = [{
+    Performs elementwise type conversion.
+
+    Supported torch/mlir element type:
```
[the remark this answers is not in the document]
navahgar: TCP should adhere to the core types in MLIR. The comment is just misleading here. 

@zezhang Can you fix the comment to talk only about MLIR core types?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: We debated about this. The summary was that we felt that is probably too low level of an abstraction for TCP. And it is better to represent "casting" as a single op so that transformations are simpler.

Let us know your thoughts on this. Are we missing anything obvious here?

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Thats a good point. @zezhang Can you add a TODO here to handle other corner cases?

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
navahgar: IIRC, this bit is meant for `i1` (`bool`) case only. Do you see any issues with that?

@zezhang Can you document this in the op? Also, it would be good to add a check in the verifier to ensure that it is `signless` only when the integer width is 1.

externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/unary.mlir
```diff
@@ -183,3 +183,63 @@ func.func @test_atan_f32(%arg0 : tensor<?x?xf32>) -> tensor<?x?xf32> {
...
+  return %0 : tensor<?x?xf32>
+}
+
+// -----
+
+func.func @test_cast_i32_f32(%arg0 : tensor<?x?xf32>) -> tensor<?x?xi16> {
+  // expected-error@+1{{tcp.cast' op out_int_signedness attr must be set when output is INT}}
+  %0 = tcp.cast %arg0 : tensor<?x?xf32> -> tensor<?x?xi16>
+  return %0 : tensor<?x?xi16>
+}
```
burmako, line 201: Consider adding more negative tests to cover all the ways how verification can fail. Apologies - I didn't notice this in the previous review.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
burmako: I can't say I have hard data on what approach should be preferred in this case. The optional signedness attributes felt a bit awkward, but having a bunch of ops instead of just one also has downsides. I think that the current design looks good for getting started and getting more experience with it.

externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp
burmako: "IIRC, this bit is meant for i1 (bool) case only. Do you see any issues with that?". Ah, this was non-obvious :) Additional documentation would help with this.

burmako (approved):
Overall, the PR looks good, and I left just one additional comment.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td
navahgar: Sounds good. Lets keep this as one of the design points that needs to be revisited.

PR #2162, opened 2023-05-25
setup-python only on x64 machines

powderluv:
there are no arm64 Python binaries. Should also work on x86//64 macOS runners.

### DIFF
--- a/.github/actions/setup-build/action.yml
+++ b/.github/actions/setup-build/action.yml
@@ -15,6 +15,7 @@ runs:
 
   steps:
   - name: Set up Python
+    if: ${{ runner.arch == 'X64' }}
     uses: actions/setup-python@v4
     with:
       python-version: '3.11'


powderluv:
CI run is here: https://github.com/llvm/torch-mlir/actions/runs/5075572008

powderluv:
This should allow arm64 builds to proceed (https://github.com/llvm/torch-mlir/actions/runs/5073693909/jobs/9113023363)

.github/actions/setup-build/action.yml
```diff
@@ -15,6 +15,7 @@ runs:
 
   steps:
   - name: Set up Python
+    if: ${{ runner.arch == 'X64' }}
```
ashay, line 18: Should we add a similar guard condition around the `pip install` lines below as well?

Also, let's add a comment to explain why this line is necessary/sufficient.

ashay: Oh, I see.  Instead of setting up Python explicitly, are we using the Python version installed on the system?

powderluv: Yes. Because GitHub doesn't provide arm64 python

.github/actions/setup-build/action.yml
```diff
@@ -15,16 +15,19 @@ runs:
...
     uses: actions/setup-python@v4
     with:
       python-version: '3.11'
 
   - name: Install MLIR Python depends
+    if: ${{ runner.os != 'Linux' }}
```
ashay, line 24: I'm not sure how this works.  Don't we need to install the Python dependencies to be able to build MLIR's Python bindings?

powderluv: so on macOS we need the system install since we are not dockerized. On Linux both for CI and Release we are dockerized so all the deps are in the docker image. This is a no-op on Linux but it fails to install on arm64 since github doesn't have a Python release for arm64 yet.

ashay: Ah, I see.  Thanks!

ashay (approved):

powderluv:
ah i merged this into the other PR instead of main.  Going to land on main now

Issue #2135, opened 2023-05-19
Static shapes not working with TMTensor::Attention


gpetters94:
I'm running into an issue with static shapes with the new `tm_tensor.attention` op.

This MLIR compiles to tm_tensor without issue:
```
#loc = loc(unknown)
module attributes {torch.debug_module_name = "ScaledDotProductAttentionDifferentModule"} {
  func.func @forward(%arg0: !torch.vtensor<[?,?,?,?],f32> loc(unknown), %arg1: !torch.vtensor<[?,?,?,?],f32> loc(unknown), %arg2: !torch.vtensor<[?,?,?,?],f32> loc(unknown)) -> !torch.vtensor<[?,?,?,?],f32> {
    %none = torch.constant.none loc(#loc)
    %float0.000000e00 = torch.constant.float 0.000000e+00 loc(#loc)
    %false = torch.constant.bool false loc(#loc)
    %0 = torch.aten.scaled_dot_product_attention %arg0, %arg1, %arg2, %none, %float0.000000e00, %false, %none : !torch.vtensor<[?,?,?,?],f32>, !torch.vtensor<[?,?,?,?],f32>, !torch.vtensor<[?,?,?,?],f32>, !torch.none, !torch.float, !torch.bool, !torch.none -> !torch.vtensor<[?,?,?,?],f32> loc(#loc1)
    return %0 : !torch.vtensor<[?,?,?,?],f32> loc(#loc)
  } loc(#loc)
} loc(#loc)
#loc1 = loc("/home/gap/code/mlir-npcomp/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/basic.py":3735:15)
```

But this fails:
```
#loc = loc(unknown)
module attributes {torch.debug_module_name = "ScaledDotProductAttentionDifferentModule"} {
  func.func @forward(%arg0: !torch.vtensor<[3,2,8,4],f32> loc(unknown), %arg1: !torch.vtensor<[3,2,16,4],f32> loc(unknown), %arg2: !torch.vtensor<[3,2,16,4],f32> loc(unknown)) -> !torch.vtensor<[3,2,8,4],f32> {
    %none = torch.constant.none loc(#loc)
    %float0.000000e00 = torch.constant.float 0.000000e+00 loc(#loc)
    %false = torch.constant.bool false loc(#loc)
[...]

PR #2166, opened 2023-05-25
Remove incorrect type checking for Torch::ScaledDotProductAttention

gpetters94:
This might need better input size verification, but this compiles cross attention cases properly, fixing #2135.


pashu123 (approved):
LGTM!

PR #2170, opened 2023-05-25
[Torch Dialect] add support for AtenIsnanOp

JianzheXiao:
as title: this PR adds support and decomposition for aten.isnanr.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -351,6 +351,24 @@ class DecomposeAtenZeroOp
...
 } // namespace
 
+namespace {
+class DecomposeAtenIsnanOp : public OpRewritePattern<AtenIsnanOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenIsnanOp op,
+                                PatternRewriter &rewriter) const override {
+    Value input = op.getSelf();
+    if (!input.getType().isa<ValueTensorType>()) {
```
qingyunqu, line 361: I think it's no need to check this?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -215,6 +215,12 @@ def aten〇lift_fresh_copy〡shape(self: List[int]) -> List[int]:
 def aten〇_log_softmax_backward_data〡shape(grad_output: List[int], output: List[int], dim: int, input_dtype: int) -> List[int]:
     return upstream_shape_functions.unary(grad_output)
 
+def aten〇isnan〡shape(self: List[int]) -> List[int]:
+    return upstream_shape_functions.unary(self)
+
+def aten〇ne〇Tensor〡shape(self: List[int], other: List[int]) -> List[int]:
```
qingyunqu, line 221: I think it's better to add an e2e test for AtenNeTensorOp.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
JianzheXiao: Yes fixed

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -215,6 +215,12 @@ def aten〇lift_fresh_copy〡shape(self: List[int]) -> List[int]:
...
+    return upstream_shape_functions.unary(self)
```
zhekunz2, line 222: `AtenNeTensor` is a broadcast-able op. Shape function should look like:
```suggestion
    return upstream_shape_functions.broadcast(self, other)
```

python/torch_mlir_e2e_test/test_suite/elementwise_comparison.py
```diff
@@ -578,6 +578,48 @@ def ElementwiseNeIntScalarModule_basic(module, tu: TestUtils):
...
 # ==============================================================================
 
+class ElementwiseNeFloatTensorModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([2, 3], torch.float32, True),
```
zhekunz2, line 588: Can you also add a test case with dynamic shape?

JianzheXiao: Added, but because it will be failed with dynamic shape I will not put it in the PASS_SET right now

qingyunqu (approved):
LGTM

ramiro050 (approved):

PR #2176, opened 2023-05-26
Bump LLVM

makslevental:
Tosa added an `acc_type` to [`AvgPool2d`](https://mlir.llvm.org/docs/Dialects/TOSA/#tosaavg_pool2d-mlirtosaavgpool2dop); this implementation pulls straight from the verifier [for their op](https://github.com/llvm/llvm-project/blob/6c9c69bd761653372999c09c7862290f8793068a/mlir/lib/Dialect/Tosa/IR/TosaOps.cpp#L144).

cc @powderluv


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3971,10 +3971,25 @@ class ConvertAtenPoolingBaseOp : public OpConversionPattern<AtenOpT> {
...
           op, "Failed to process inputs for pooling");
 
-    auto pooledOutput =
-        rewriter
-            .create<TosaOpT>(op->getLoc(), outputTy, input, kernel, stride, pad)
-            .getResult();
+    Value pooledOutput;
+    static_assert(std::is_same<TosaOpT, tosa::MaxPool2dOp>::value ||
+                      std::is_same<TosaOpT, tosa::AvgPool2dOp>::value,
+                  "Expected either tosa::MaxPool2dOp or tosa::AvgPool2dOp");
```
makslevental, line 3977: There's probably some slicker way to "spell" this so feel free to drop a hint (or link) if your template-fu is better than mine.

powderluv:
@ashay since we need a green commit sync. @makslevental is trying to prefetch the next bump since there are changes upstream he requires.

ashay:
Hey, just saw this.  From last night's run, our most recent green commit for LLVM is [10518fea](https://github.com/llvm/llvm-project/commit/10518feaf33760e77be46af62c41a1caf4ef1da2).

makslevental:
@ashay last night's run of what? That commit is from April 4th while mlir-hlo has https://github.com/llvm/llvm-project/commit/daeee56798c5 as of yesterday...

ashay:
We run nightly builds and tests of LLVM to check if they work for Linux and Windows with Release + Debug configurations.  The last night's run picked the LLVM commit with hash 10518fea, which landed 13 hours ago.

makslevental:
> We run nightly builds and tests of LLVM to check if they work for Linux and Windows with Release + Debug configurations. The last night's run picked the LLVM commit with hash 10518fea, which landed 13 hours ago.

Ah sorry you're right - I clicked the diffusion link and didn't notice the timestamp from github. Okay cool I'll kick these tests again tomorrow (or Monday...).

lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
```diff
@@ -341,5 +341,34 @@ template std::optional<Value> getConstTensor<int64_t>(PatternRewriter &,
                                                       Operation *,
                                                       ArrayRef<int64_t> vec,
                                                       ArrayRef<int64_t> shape);
+
+LogicalResult getAvgPool2dAccType(PatternRewriter &rewriter, Value input,
+                                  TypeAttr &accType) {
+  auto inputTy = llvm::dyn_cast<ShapedType>(input.getType());
+  if (!inputTy)
+    return failure();
+  auto inputETy = inputTy.getElementType();
```
eric-k256, line 350: (Background)
The goal with the change we made to TOSA was to allow either fp32 or fp16 accumulation for average pool to trade off speed/range of intermediate value. The accumulator size isn't visible outside the op, so it needs this new attribute. At first we're leaving accumulator as always fp32 while finding the right time/method to choose fp16.

For legalizing from torch, there is an easier way than what the validator does. You can reduce the type choice to something like:
```
  // Tosa supports FP16 and FP32 accumulator type for FP16 input. When the time
  // FP16 is supported, the accumulator type can be selected based on trade-off
  // between performance and accuracy. Set to FP32 by default.
 accType = inputETy.isa<FloatType>()
                          ? mlir::TypeAttr::get(rewriter.getF32Type())
                          : mlir::TypeAttr::get(rewriter.getIntegerType(32));
```

lib/Dialect/TorchConversion/Transforms/BackendTypeConversion.cpp
```diff
@@ -53,12 +53,12 @@ setupValueTensorToBuiltinTensorConversion(ConversionTarget &target,
 static void setupTorchBoolToI1Conversion(ConversionTarget &target,
                                          TypeConverter &typeConverter) {
   target.addLegalOp<TorchConversion::ToI1Op, TorchConversion::FromI1Op>();
-  typeConverter.addConversion([](Torch::BoolType type) -> Optional<Type> {
+  typeConverter.addConversion([](Torch::BoolType type) -> std::optional<Type> {
```
makslevental, line 56: LLVM finally removed `Optional` as an alias for `std::optional`.

makslevental:
@ashay do you know what's going on here? Each time I've run these tests for the last couple of days they all go for the full 6 hours and then timeout.

ashay:
Hey @makslevental, just to be sure, running `python -m e2e_testing.main --config=stablehlo --verbose --sequential` locally on your machine works (as in, doesn't stall forever), doesn't it?

From the logs, it looks some of the StableHLO tests are failing, and I wonder if they are waiting for some user input, causing CI to stall until it finally times out.  The reason only the ubuntu in-tree builds fail is because only those builds run the e2e tests.

Perhaps you already saw this, but here's an excerpt from the CI logs (notice the timestamps):

```
2023-06-03T01:11:35.1513153Z Job is about to start running on the runner: torch-mlir-2 (repository)
...
2023-06-03T01:30:38.6616536Z :::: Run StableHLO e2e integration tests
...
2023-06-03T01:30:48.0309566Z ValueError: invalid itemsize in generic type tuple
2023-06-03T01:30:49.2288101Z LLVM ERROR: Failed to infer result type(s).
2023-06-03T07:11:50.4682407Z ##[error]The operation was canceled.
```

If you're able to reproduce the error locally, we might learn whether the Python process hangs when it sees a failure, causing CI to ultimately timeout.

mgehre-amd:
When I run this locally, I see that `python -m e2e_testing.main --config=stablehlo -v -s -f ArgmaxModule_basic` crashes with `LLVM ERROR: Failed to infer result type(s).`.

vivekkhandelwal1:
Hi @makslevental, when can we get this patch merged? It's my turn this week for the llvm-bump. And it's waiting for this patch to be merged.

makslevental:
@powderluv @ashay @vivekkhandelwal1 @ramiro050 @mgehre-amd 

I can't make it work and unfortunately I am out of time. I added `STABLEHLO_CRASHING_SET` and with this list I am able to pass (i.e., no segfaults) sequentially but then it continues to hang when multiprocessing. Something must be not threadsafe 🤷 somewhere and corrupting something; when running under multiprocessing I see a `LLVM ERROR: Failed to infer result type(s)` whereas when running sequentially I do not. In addition there are a slew of new [stablehlo fails](https://gist.github.com/makslevental/87dd7e80a0b84377f7f17df43df25aab#file-gistfile1-txt-L3717).

I have incorporated @eric-k256's comments so the rest of the PR is good to go. Good luck.

mgehre-amd:
I will give it quick try.

mgehre-amd:
Adding @sjarus because I think Eric is out this week

mgehre-amd:
All builds passed, waiting for reviewer approval.

ashay (approved):
LGTM!

Issue #1135, opened 2022-08-02
Experiment: coordinate LLVM commits between torch-mlir and onnx-mlir


sstamenova:
Following up on a couple of conversations we've had about coordinating LLVM commits across projects (in [onnx-mlir](https://github.com/onnx/onnx-mlir/issues/1558) and in [LLVM](https://discourse.llvm.org/t/coordinate-llvm-commits-for-different-project/63990/64)), we had a follow up discussion in the Torch MLIR community meeting yesterday (see [this comment](https://discourse.llvm.org/t/coordinate-llvm-commits-for-different-project/63990/64) for more details).

The short-term solution that we decided to run as an experiment is to use Torch MLIR as the "source of truth" for what commit to use for other projects with the caveat that when the commit of LLVM is chosen for Torch MLIR, it would have gone through some testing that would pretty much guarantee that the commit would satisfy any requirements that onnx-mlir has as well. Based on the discussion, it should be fairly trivial to find a version of MHLO that uses the same LLVM commit as well, though MHLO might need some fixes for onnx-mlir (to make the shared library build work, for example).

We are going to pick the commit based on some internal testing that we do at Microsoft when we pick the LLVM commit that we merge with. This includes several platforms (Ubuntu, Windows, CentOS) and build tools (clang, VS, gcc) and both static and shared library builds. We'll communicate the commit here and we'll use this commit to update Torch MLIR as we are able (though ideally after we communicate the commit someone else will do the update some of the time). Then when it is time for onnx-mlir to update to a newer LLVM commit (which could also be done by us, though ideally not always), that project (as well as any others that want to stay in sync) can choose any of the recent LLVM commits on this list and we'll have at least one point a month where Torch MLIR and onnx-mlir are synced to the same LLVM commit.

The goal is to improve the process for everyone for the short-term while also working on a longer-term solution in LLVM (for example, pre-commit checks, etc.) and this is a relatively cheap first step towards all of these projects being in sync with the same LLVM dependencies.

A second step would be to start tagging the "green" commits automatically (see https://discourse.llvm.org/t/coordinate-llvm-commits-for-different-project/63990/65) instead of "announcing" them manually, but that will require some additional work we need to evaluate.

@AlexandreEichenberger @sjarus @stellaraccident @silvasean @joker-eph @ashay 

So to recap:
1. Once a week, we will identify a "green" LLVM commit and post it here
2. Once a week, someone (sometimes us, though hopefully not always) will update Torch MLIR to this commit 
[...]

PR #2177, opened 2023-05-26
Actually migrate to StableHLO

makslevental:
Not sure what the story is but after https://github.com/llvm/torch-mlir/pull/1840 I believe we should be pointing to https://github.com/openxla/stablehlo/ instead of https://github.com/tensorflow/mlir-hlo, which is no longer being updated.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "externals/mlir-hlo"]
 	path = externals/mlir-hlo
 	url = https://github.com/tensorflow/mlir-hlo.git
+[submodule "externals/stablehlo"]
+	path = externals/stablehlo
+	url = https://github.com/openxla/stablehlo.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,13 +119,13 @@ endif()
 
 if (TORCH_MLIR_ENABLE_STABLEHLO)
   set(STABLEHLO_BUILD_EMBEDDED ON)
-  add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/externals/mlir-hlo
-    ${CMAKE_CURRENT_BINARY_DIR}/mlir-hlo
+  add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo
+    ${CMAKE_CURRENT_BINARY_DIR}/stablehlo
     EXCLUDE_FROM_ALL)
-  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/mlir-hlo/include)
-  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/mlir-hlo)
-  include_directories(${CMAKE_CURRENT_BINARY_DIR}/mlir-hlo/include)
-  include_directories(${CMAKE_CURRENT_BINARY_DIR}/mlir-hlo)
+  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo/include)
+  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo)
+  include_directories(${CMAKE_CURRENT_BINARY_DIR}/stablehlo/include)
+  include_directories(${CMAKE_CURRENT_BINARY_DIR}/stablehlo)
 endif()
 
 set(TORCH_MLIR_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}")


makslevental:
So I ran this locally with `TORCH_MLIR_ENABLE_STABLEHLO=OFF` (whoops, duh) and thought it would be a simple thing. But it's not - the blocker seems to be [MLIRBufferTransforms](https://github.com/tensorflow/mlir-hlo/blob/0cea8b109116cac6de387e715ab2fd6a75bc1d68/transforms/CMakeLists.txt) which probably can't/won't/shouldn't go to stablehlo. So now I'm just wondering what the plan is since [mlir-hlo](https://discourse.llvm.org/t/sunsetting-the-mlir-hlo-repository/70536) is being sunsetted. [I.e., we're effectively stuck on an LLVM commit](https://github.com/llvm/torch-mlir/pull/2176).

powderluv:
I think there was some other reason to use the old repo (which everytime we bump topples our git with index.lock issues).  @ramiro050 do you know why we need the tf/mhlo repo ?

ashay:
Huh.. I was unaware that MHLO was being sunset.  Thanks for the link!

The reason we were pointing to the MHLO repo was because we used the MHLO passes to lower from MHLO to Linalg, after converting from StableHlo to MHLO (see pass pipeline [here](https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir_e2e_test/stablehlo_backends/linalg_on_tensors.py#L43)).  Now that MHLO is going away, does StableHlo include the lowering-to-linalg pass?

ashay:
Oh, I see.  It got moved to XLA: https://github.com/openxla/xla/blob/main/xla/mlir_hlo/mhlo/transforms/mhlo_passes.td#L107.

burmako:
Hi folks! We are still updating the MLIR-HLO repository, but indeed we are planning to sunset it in the near future. We have not yet discussed the exact date, but will share both the exact details and the proposed date [on Discourse](https://discourse.llvm.org/t/sunsetting-the-mlir-hlo-repository/70536).

"The reason we were pointing to the MHLO repo was because we used the MHLO passes to lower from MHLO to Linalg". Meanwhile, I think you may be interested in the work that IREE folks are doing to switch from the MLIR-HLO repository to the StableHLO repository (cc @kuhar who's driving this work). Here's the tracking issue: https://github.com/openxla/iree/issues/12678, and here's the OpenXLA RFC about this which goes into a bit more detail: https://groups.google.com/a/openxla.org/g/openxla-discuss/c/EWuUbyL5n3c.

powderluv:
Wait is https://github.com/openxla/xla/blob/main/xla/mlir_hlo/mhlo/transforms/mhlo_passes.td#L107 the canonical spot for all mhlo->linalg lowerings ? I am not sure how viable it would be to add lowerings to XLA for Pytorch->StableHLO->LinAlg .

@burmako where do we add any new stablehlo->linalg lowerings ? Is it in the XLA repo ?


@ashay @qingyunqu @ZihengJiang @Vremold @tanyokwok should we fully switch to stablhlo and remove mhlo as a submodule ? (the mhlo repo still topples over git with the `index.lock` issue every time it is updated)

ashay:
Oh hey Eugene, thanks for the links to the IREE issue!  Minor question, what's the reason for IREE making a copy of the MHLO source files instead of, say, a submodule dependency?  Would it be a bad idea for Torch-MLIR to add a submodule dependency on XLA?

More importantly, I noticed that XLA integrates StableHlo in what looks like a [patch applied to the last update](https://github.com/openxla/xla/commit/0738064c9ae083befd531b868bf1fa749255f317), but this presents two challenges.  First, to build Torch-MLIR, we'd need the StableHLO commit hash that XLA uses, and while it can be parsed out of workspace.bzl, I wonder if you'd be open to saving it in a plaintext file, similar to how [StableHLO points to the LLVM commit](https://github.com/openxla/stablehlo/blob/main/build_tools/llvm_version.txt).

The other challenge would be in sharing [green commits](https://github.com/llvm/torch-mlir/issues/1135) between Torch-MLIR and ONNX-MLIR (and potentially, Triton?) so that all involved projects can use a single LLVM commit.  While we currently create [green commit branches in the MHLO repo](https://github.com/tensorflow/mlir-hlo/branches/all), we'd probably need to setup a similar system for StableHLO and/or XLA, depending on whether we link XLA and StableHLO using commit hashes versus copying the StableHLO source into XLA.  Let me know your thoughts on either approach (or other approaches).

Anush, yeah, it makes sense to drop the MHLO submodule, but I wonder if we can just depend on XLA, which would point to the appropriate commit hash of StableHLO instead of adding submodule dependencies on both.

powderluv:
I think an implicit goal of StableHLO was to not depend on XLA (?) and be standalone hence my earlier question on where do we put these passes.  I would rather copy / fork if we have to than take an XLA dep because a few passes live there. 

Also @stellaraccident fyi

kuhar:
> I think an implicit goal of StableHLO was to not depend on XLA (?) and be standalone hence my earlier question on where do we put these passes. I would rather copy / fork if we have to than take an XLA dep because a few passes live there.

In IREE we ported the code to understand what it would take to lower directly to linalg  + other dialects. This allowed us to iterate on this quickly and after a little over a month of work we are almost done.

Along the way we learned that it takes more effort than we anticipated: the C++ api is close but not exactly the same as MHLO, StableHLO is doesn't provide folders and canonicalzers, there's some utility code that needed to be ported, the printed form is often different which requires modifications to tests, CHLO needs to be lowered to MHLO, etc.

IMO it would make sense to upstream the work IREE's done to StableHLO, provided that this aligns with the direction of the StableHLO project itself. It would be a shame to duplicate porting this effort across other projects.

burmako:
"where do we add any new stablehlo->linalg lowerings ? Is it in the XLA repo ?". I think that these lowerings could live in the StableHLO repository, and I've just wrote this up as [an OpenXLA RFC](https://groups.google.com/a/openxla.org/g/openxla-discuss/c/KsRp9euuuB0). Thank you, @kuhar, for making this RFC possible through your excellent work! 

"I think an implicit goal of StableHLO was to not depend on XLA (?) and be standalone hence my earlier question on where do we put these passes. I would rather copy / fork if we have to than take an XLA dep because a few passes live there". The RFC goes into details about goals of StableHLO, but in a nutshell: yes, the goal of StableHLO is to be framework- and compiler-agnostic. This is why I think it can be a great fit for this purpose - it's relatively lightweight to depend on, supports both CMake and Bazel builds and doesn't involve heavyweight CI.

"The other challenge would be in sharing green commits between Torch-MLIR and ONNX-MLIR (and potentially, Triton?) so that all involved projects can use a single LLVM commit". Happy to discuss the logistics of this! Although, based on some recent discussions, I think that Stella may already have some opinion on how this could be done, so let's ask her first: @stellaraccident.

powderluv:
Folks - should we now switch over to stableHLO proper and drop mhlo ?

stellaraccident:
> Folks - should we now switch over to stableHLO proper and drop mhlo ?

Those passes are effectively a testing-only dep, right? (You can convert to stablehlo without them but verifying on the ref backend needs them)

If there isn't consensus from the stablehlo side, I would just fork whatever is necessary into the torch-mlir repo and drop the dep on mhlo. Ideally, they can just go into stablehlo, even as an optional contrib kind of thing.

In the Turbine work, I've already had to fork around and not use stablehlo at all because there isn't a reasonable path to use that part of the codebase without unusable deps.

stellaraccident:
I'll also say that I haven't experienced much trouble with sharing commits if we leave projects like torch-mlir and stablehlo pretty lightweight (just dialects and high level conversion passes)... The stuff doesn't drift that much at that level. I regularly use both projects at other commits (but only by excluding the testing infra). It may be worthwhile to further isolate testing and lowering in the projects so that things stay pretty portable across versions.

burmako:
@qingyunqu To the best of my knowledge, the StableHLO-to-Linalg work has been completed, and I was planning to migrate it from IREE into the StableHLO repository later this month. If this is time-sensitive on your side, I can hurry up on my side as well.

@stellaraccident "In the Turbine work, I've already had to fork around and not use stablehlo at all because there isn't a reasonable path to use that part of the codebase without unusable deps". Can you tell more about this? The StableHLO repository was designed to specifically avoid this problem of the MLIR-HLO repository, so if you can elaborate on what dependencies you find unusable, we may be able to address that.

Issue #2178, opened 2023-05-26
`add_` e2e test crashes with torchdynamo backend


ramiro050:
Test:

```python
class Add_Module(torch.nn.Module):

    def __init__(self):
        super().__init__()
        self.tensor = torch.ones(2, 3)

    @export
    @annotate_args([
        None,
        ([-1, -1], torch.float32, True),
    ])
    def forward(self, x):
        return torch.ops.aten.add_(x, self.tensor)


@register_test_case(module_factory=lambda: Add_Module())
def Add_Module_basic(module, tu: TestUtils):
    module.forward(tu.rand(2, 3))
```

Running:

```
python -m e2e_testing.main -v -c torchdynamo -f "Add_"
```

results in

```
Thread 1 "python" received signal SIGSEGV, Segmentation fault.
0x00007ffff7aea331 in _mlir__mlir_ciface_Add_Module ()
(gdb) bt
#0  0x00007ffff7aea331 in _mlir__mlir_ciface_Add_Module ()
#1  0x00007ffff7fb6f7a in ?? () from /lib/x86_64-linux-gnu/libffi.so.8
#2  0x00007ffff7fb640e in ?? () from /lib/x86_64-linux-gnu/libffi.so.8
#3  0x00007ffff7fb6b0d in ffi_call ()
   from /lib/x86_64-linux-gnu/libffi.so.8
#4  0x00007ffff7b24d4d in ?? ()
   from /usr/lib/python3.10/lib-dynload/_ctypes.cpython-310-x86_64-linux-gnu.so
#5  0x00007ffff7b24125 in ?? ()
   from /usr/lib/python3.10/lib-dynload/_ctypes.cpython-310-x86_64-linux-gnu.so
#6  0x00005555556a531b in _PyObject_MakeTpCall ()
#7  0x000055555569e42a in _PyEval_EvalFrameDefault ()
#8  0x00005555556bb27c in ?? ()
#9  0x000055555569ad4d in _PyEval_EvalFrameDefault ()
#10 0x00005555556ae655 in _PyFunction_Vectorcall ()
#11 0x000055555569ad4d in _PyEval_EvalFrameDefault ()
#12 0x00005555556ae655 in _PyFunction_Vectorcall ()
#13 0x0000555555698dc4 in _PyEval_EvalFrameDefault ()
#14 0x00005555556ae655 in _PyFunction_Vectorcall ()
#15 0x0000555555698bda in _PyEval_EvalFrameDefault ()
[...]

PR #2179, opened 2023-05-26
Add `ReadOnly` trait to `copy.to_vtensor`

ramiro050:
Before inlining a global slot, the users of the global slot are checked to see if they are `ReadOnly` or `MemoryEffectFree` to make sure that the global slot is not being mutated. Because the op `copy.to_vtensor` currently does not have the `ReadOnly` trait, if a global slot is passed to `copy.to_vtensor`, the pass `InlineGlobalSlots` will fail.

The op `copy.to_vtensor` is `ReadOnly`, since it does not modify the contents of the input tensor; it simply makes a new copy. This commit adds the trait as well as an e2e test that generates the case of a global slot being passed to a `copy.to_vtensor`.

### DIFF
--- a/e2e_testing/main.py
+++ b/e2e_testing/main.py
@@ -32,6 +32,7 @@ from .xfail_sets import (
     STABLEHLO_PASS_SET,
     TOSA_PASS_SET,
     LTC_XFAIL_SET,
+    LTC_CRASHING_SET,
     TORCHDYNAMO_XFAIL_SET,
     TORCHDYNAMO_CRASHING_SET
 )
@@ -104,7 +105,7 @@ def main():
     elif args.config == "lazy_tensor_core":
         config = LazyTensorCoreTestConfig()
         xfail_set = LTC_XFAIL_SET
-        crashing_set = set()
+        crashing_set = LTC_CRASHING_SET
     elif args.config == "torchdynamo":
         config = TorchDynamoTestConfig(RefBackendLinalgOnTensorsBackend())
         xfail_set = TORCHDYNAMO_XFAIL_SET
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -298,6 +298,9 @@ TORCHDYNAMO_CRASHING_SET = {
     "ToCopyModule_basic",
     "TransposeIntModule_basic",
     "TransposeIntNegDimsModule_basic",
+
+    # See https://github.com/llvm/torch-mlir/issues/2178
+    "Add_Module_basic"
 }
 
 STABLEHLO_PASS_SET = {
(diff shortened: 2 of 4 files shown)


pashu123 (approved):
LGTM!

mgehre-amd:
The PR itself looks good, but I'm still wondering why this failed in the first places.
It seems to be related to the fact that the `add_` function modifies the first argument in-place, but `self.tensor` in `return torch.ops.aten.add_(x, self.tensor)` is the second argument, which is not modified in-place.
Is this because the whole operation is regarded as "non-value semantics" instead of just its first operand?

ramiro050:
> Seems to be crashing in LTC tests:
> 
> ```
> Running Add_Module_basic...
> [F tensor.cpp:361] Check failed: lazy_tensor Input tensor is not a lazy tensor: CPUFloatType
> ```

Yeah, I've [created an issue](https://github.com/llvm/torch-mlir/issues/2178) for it and added it to a `crashing_set` for LTC. Unclear to me why it is crashing.

ramiro050:
> The PR itself looks good, but I'm still wondering why this failed in the first places. It seems to be related to the fact that the `add_` function modifies the first argument in-place, but `self.tensor` in `return torch.ops.aten.add_(x, self.tensor)` is the second argument, which is not modified in-place. Is this because the whole operation is regarded as "non-value semantics" instead of just its first operand?

The entire computation going into torch-mlir always begins with non-value semantics. Then slowly through a few different passes, an attempt is made to convert the graph to value semantics. What happened in this case is the following.

First, the `inline-global-slots` pass tried to inline the global slot, but because the slot is being used by an in-place op `add_`, the pass did not inline anything. The current implementation for `inline-global-slots` is not smart enough to know that `add_` only modifies the first argument, so it conservatively fails. Then, the pass `reduce-op-variants` converts the `add_` to `add`, inserting the appropriate overwrites and copies to `vtensor`. From this change we end up with something like

```mlir
   torch.global_slot "private" @_param_constant82 : !torch.tensor
  func.func @forward(%arg0: !torch.vtensor<[1,384,12],f32>) -> !torch.vtensor {
    %int1 = torch.constant.int 1
    %0 = torch.global_slot.get @_param_constant82 : !torch.tensor
    %1 = torch.copy.to_vtensor %0 : !torch.vtensor
    %2 = torch.aten.add.Tensor %arg0, %1, %int1 : !torch.vtensor<[1,384,12],f32>, !torch.vtensor, !torch.int -> !torch.vtensor
    return %2 : !torch.vtensor
  }
```

Now we have a global slot being passed to the `copy.to_vtensor` op. The second time the `inline-global-slots` pass takes place, it fails again because `copy.to_tensor` does not have a `ReadOnly` trait, so the pass does not know if the op will modify the tensor or not. This failure continues to happen until the `lower-to-backend-contract` pass reaches the iteration limit, resulting in the error you were seeing.

PR #2180, opened 2023-05-26
[Tcp] Enable Tcp by default in mlir-tcp branch

navahgar:
This PR enables Tcp by default in the cmake as well as bazel builds in the `mlir-tcp` branch.


sjain-stanford (approved):
Thanks for fixing this.

PR #2190, opened 2023-06-02
CI: Spot fixes related to nightly and stable PyTorch builds

ashay:
This PR contains two patches:

 - CI: Skip (redundant) libtorch build when using stable PyTorch version

> When we use PyTorch stable builds, there is no need to build libtorch
> from source, making the stable-pytorch-with-torch-binary-OFF
> configuration redundant with stable-pytorch-with-torch-binary-ON.  This
> patch drops the redundant configuration from CI.

 - CI: Simplify guard conditions for creating and using libtorch cache

> Whether libtorch is enabled or not is predicated on a host of conditions
> such as the platform, in-tree versus out-of-tree build, and stable
> versus nightly PyTorch builds.  Instead of repeating these conditions to
> guard whether to create or use the libtorch cache artifacts (and getting
> them almost incorrect), this patch predicates the relevant pipeline
> steps to whether libtorch is enabled, thus making the conditions far
> simpler.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -92,7 +92,7 @@ jobs:
         arch: x64
 
     - name: Try to Restore PyTorch Build Cache
-      if: matrix.os-arch != 'windows-x86_64'
+      if: ${{ matrix.torch-binary == 'OFF' }}
       id: cache-pytorch
       uses: actions/cache/restore@v3
       with:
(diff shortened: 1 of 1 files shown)


ashay:
@maxbartel Somehow I am not able to add you as a reviewer.  Could you still take a look at these changes please?

powderluv (approved):

powderluv:
landing to catch the next builds

maxbartel:
LGTM. Sorry I wasn't fast enough because of time zone differences

PR #2194, opened 2023-06-03
[Torch Dialect] Add canonicalize pattern for aten.is_floating_point

qingyunqu:


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1860,6 +1860,26 @@ void Aten__Getitem__TOp::getCanonicalizationPatterns(
   });
 }
 
+//===----------------------------------------------------------------------===//
+// AtenIsFloatingPointOp
+//===----------------------------------------------------------------------===//
+
```
mgehre-amd, line 1866: Curious, would this also work as a ::fold by returning BoolAttr?
This would have the advantage that one doesn't need to explicitly call the canonicalize pass.

qingyunqu: Yes, maybe implement this as folder is more suitable. I will write it as folder.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1860,6 +1860,25 @@ void Aten__Getitem__TOp::getCanonicalizationPatterns(
...
+OpFoldResult AtenIsFloatingPointOp::fold(FoldAdaptor adaptor) {
+  auto operandType = getSelf().getType().dyn_cast<BaseTensorType>();
+  if (!operandType)
+    return nullptr;
+  if (operandType.hasDtype()) {
+    if (operandType.getDtype().isa<mlir::FloatType>()) {
+      return IntegerAttr::get(IntegerType::get(getContext(), 1), true);
+    } else {
+      return IntegerAttr::get(IntegerType::get(getContext(), 1), false);
+    }
```
ramiro050, line 1876: nit: less repetitive to write

```c++
bool isFloat = operandType.getDtype().isa<mlir::FloatType>();
return IntegerAttr::get(IntegerType::get(getContext(), 1), isFloat);

ramiro050 (approved):
can you add a lit test here: https://github.com/llvm/torch-mlir/blob/main/test/Dialect/Torch/canonicalize.mlir

Other than the lit test and the small change request below, LGTM

PR #2198, opened 2023-06-05
Torch: Fold RuntimeAssertOp when condition is true

mgehre-amd:
The runtime.assert op gets introduced by some decompositions.
In the cases we saw, the condition got simplified to `true`, so we can remove the assert again.

This makes two tests pass with TOSA.


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -502,6 +502,27 @@ void PrimIfOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+  patterns.add(+[](RuntimeAssertOp op, PatternRewriter &rewriter) {
+    bool value;
+    if (!matchPattern(op.getCondition(), m_TorchConstantBool(&value)))
+      return failure();
+
+    if (value) {
+        rewriter.eraseOp(op);
+        return success();
+    }
+    // Even if the condition is statically false, the assert might never be
```
qingyunqu, line 520: the comment is just a little confusing. I think "If the condition is statically false, left it as unfolded." is better.

ramiro050 (approved):
Can you also add a lit test in: https://github.com/llvm/torch-mlir/blob/main/test/Dialect/Torch/canonicalize.mlir

Other than that, LGTM

PR #2200, opened 2023-06-05
[MLIR][TOSA] Fix aten.div conversion for integer inputs

cmcgirr-amd:
When converting the following `torch.aten.div` operation:
```
 %int128 = torch.constant.int 128
 %0 = torch.aten.div.Scalar %arg0, %int128 : !torch.vtensor<[?,?],si64>, !torch.int -> !torch.vtensor<[?,?],f32>
```
The following TOSA IR would be generated:
```
  %0 = "torch_c.to_builtin_tensor"(%arg0) : (!torch.vtensor<[?,?],si64>) -> tensor<?x?xi64>
  %1 = "torch.constant.int"() {value = 128 : i64} : () -> !torch.int
  %2 = "tosa.const"() <{value = dense<128> : tensor<i64>}> : () -> tensor<i64>
  %3 = "tosa.cast"(%0) : (tensor<?x?xi64>) -> tensor<?x?xf32>
  %4 = "tosa.cast"(%2) : (tensor<i64>) -> tensor<f32>
  %5 = "tosa.div"(%3, %4) : (tensor<?x?xf32>, tensor<f32>) -> tensor<?x?xf32>
```
Causing an error in the TOSA operation verifier:
```
'tosa.div' op operand #0 must be tensor of 32-bit signless integer values, but got 'tensor<?x?xf32>'
        return torch.ops.aten.div(x, 128)
```
The issue is the wrong TOSA operation was selected for the conversion. As a ReciprocalOp and MulOp are required to implement the floating point division rather than the TOSA integer division.

The fix is to check the output type of the division operation and select the appropriate TOSA conversion based on the output type. As in this case the two input integers are casted immediately to floating point. 

The same issue can be found in the TorchToLinalg pass as well. A simple up-cast on the self operand fixes the issue ensuring both inputs are floating point in this particular case.

### DIFF
--- a/e2e_testing/xfail_sets.py
+++ b/e2e_testing/xfail_sets.py
@@ -230,6 +230,9 @@ TORCHDYNAMO_XFAIL_SET = {
     # ERROR: 'torch.aten.div.Tensor' op operand #1 must be Any Torch tensor type, but got '!torch.float'
     "ElementwiseDivScalarModule_basic",
 
+    # ERROR: 'torch.aten.div.Tensor' op operand #1 must be Any Torch tensor type, but got '!torch.int'
+    "ElementwiseAtenDivIntScalarModule_basic",
+
     # ERROR: 'torch.aten.mul.Tensor' op operand #1 must be Any Torch tensor type, but got '!torch.int'
     "ElementwiseMulScalarModule_int",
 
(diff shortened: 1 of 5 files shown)


sjarus (approved):
That looks fine. Do you need an optimized form that tests the denominator for a power of 2 const and performs a right shift instead ?

mgehre-amd:
> That looks fine. Do you need an optimized form that tests the denominator for a power of 2 const and performs a right shift instead ?

Sounds interesting, but not something we need to right now. And I think we can still do it on TOSA level (e.g. as a canonicalization).

PR #2202, opened 2023-06-05
[MLIR][TORCH] Add E2E support for `aten.tril` op

vivekkhandelwal1:


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1056,30 +1081,19 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
   }
 
   if (auto triu = dyn_cast<AtenTriuOp>(op)) {
-    // Check if the rank of the input tensor is valid.
-    AtenTriuOp::Adaptor adaptor(operands);
-    auto inputType = adaptor.getSelf().getType().cast<RankedTensorType>();
-    uint64_t inputRank = inputType.getRank();
-    if (inputRank < 2) {
-      triu.emitError("too few dimensions to compute triangular part of matrix");
```
vivekkhandelwal1, line 1064: This explicit condition is not needed because this op by definition fails on an input tensor with rank < 2.

ashay (approved):
LGTM!

PR #2204, opened 2023-06-05
[MLIR][TOSA] Support unchanged dimensions in torch.aten.broadcast_to operation.

ttjost:
In `torch.aten.broadcast_to`, passing `-1` as the size for a dimension means not changing the size of that dimension.
This is currently not support by TOSA lowering, so this PR fixes such cases.


sjarus (approved):
Thanks for this contribution! This seems to make sense where the input dim is 1, another known dimsize or a ? . 

The main question is how to validate the shape alignment during a dim broadcasting binary or ternary op, but that test would need to be implemented around such an elementwise op; this change does not appear to impact the handling of that (other than to ensure rank broadcasting - including of dynamic dims - beforehand), so LGTM.

PR #2207, opened 2023-06-05
[Torch Dialect] Support recompose aten.einsum

JianzheXiao:


lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -476,6 +476,7 @@ static void markDecomposedOpsAsIllegal(MLIRContext *context,
   target.addIllegalOp<PrimsSqueezeOp>();
   target.addIllegalOp<AtenMovedimIntOp>();
   target.addIllegalOp<AtenOneHotOp>();
+  target.addIllegalOp<AtenEinsumOp>();
```
mgehre-amd, line 479: In DecomposeOps, we use `addPatternIfTargetOpIsIllegal`. I.e. the op will not be decomposed when the target declares it as legal. Wouldn't this conflict with declaring it illegal here?

ramiro050: We set the ops back to legal if they are in the `backendLegalOpsSet`:

https://github.com/llvm/torch-mlir/blob/ae7f5775162bed34293392a11e366a97540cd65a/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L485-L488

There's probably a clearer way of doing this.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2963,6 +2963,252 @@ class DecomposeAtenLinearOp : public OpRewritePattern<AtenLinearOp> {
...
+          torchTensors[0] = rewriter.create<AtenTransposeIntOp>(loc, op.getType(),torchTensors[0], zero, one);
+          size_t temp = lhsShape[lhsBatchingDims[0]];
+          lhsShape[lhsBatchingDims[0]] = lhsShape[0];
+          lhsShape[0] = temp;
+        }
+      } else {
+        lhsShape.insert(lhsShape.begin(), 1);
+      }
+
+      if (!rhsContractingDims.empty()) {
```
ramiro050, line 3102: Will take a closer look tomorrow, but this pattern has multiple cases, which should be accompanied by lit tests making sure each case works as expected: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
JianzheXiao: I accidentally closed this PR, I will reopen one and add more test cases to make sure every case is tested

PR #2208, opened 2023-06-05
[Torch Dialect] Add torch.arange scalar support

zhekunz2:
Problem:
```
def forward(self, x):
        end = torch.ops.prim.NumToTensor(x.shape[1])
        return torch.arange(0, end, 1)
```
throws error: 
```
'torch.derefine' op operand type '!torch.number' and result type '!torch.union<float, int>' are cast incompatible
```

Solution:
Fix `isValidSubType` for `!torch.number` subtype.
`isValidSubType(!torch.number, !torch.union<float, int> = true`


lib/Dialect/Torch/IR/TorchTypes.cpp
```diff
@@ -36,6 +36,18 @@ bool Torch::isValidSubtype(Type subtype, Type type) {
...
 
+  if (auto number = subtype.dyn_cast<NumberType>()) {
+    if (auto unionType = type.dyn_cast<UnionType>()) {
+      bool hasFloatType = false;
+      bool hasIntType = false;
+      for (auto containedType : unionType.getContainedTypes()) {
+        hasFloatType |= containedType.isa<FloatType>();
+        hasIntType |= containedType.isa<IntType>();
+      }
+      return hasIntType && hasFloatType;
```
qingyunqu, line 47: I think this is not a correct condition of `isValidSubType`. More precisely, I think `torch.derefine` op's input type should not be a subtype of result type.
In my understanding, if input type may become one of result union types, the `torch.derefine` op is `cast compatible`. @ramiro050 @silvasean How about your idea?

qingyunqu (changes requested):

lib/Dialect/Torch/IR/TorchTypes.cpp
qingyunqu: So I think the correct solution is that changing `DerefineOp::areCastCompatible` function to something like:
```
bool DerefineOp::areCastCompatible(mlir::TypeRange inputs,
                                   mlir::TypeRange outputs) {
  return isValidSubtype(inputs[0], outputs[0]) || mayBecomeSameType(inputs[0], outputs[0]);
}
```

PR #2210, opened 2023-06-06
Print name of the backend when tests fail to help debugging issues in CI

mgehre-amd:
When an end2end test fails or unexpectedly passes in CI, it sometimes takes me a bunch of scrolling to understand in which e2e config the test failed.
With this change, we show the name of the end2end config at the end of output.

### DIFF
--- a/python/torch_mlir_e2e_test/reporting.py
+++ b/python/torch_mlir_e2e_test/reporting.py
@@ -263,7 +263,8 @@ class SingleTestReport:
 
 def report_results(results: List[TestResult],
                    expected_failures: Set[str],
-                   verbose: bool = False):
+                   verbose: bool = False,
+                   config: str = ""):
     """Print a basic error report summarizing various TestResult's.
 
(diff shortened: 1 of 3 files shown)


python/torch_mlir_e2e_test/reporting.py
```diff
@@ -310,7 +311,7 @@ def report_results(results: List[TestResult],
         results_by_outcome['XPASS']) != 0
 
     if had_unexpected_results:
-        print('\nUnexpected outcome summary:')
+        print(f'\nUnexpected outcome summary: ({config})')
```
silvasean, line 314: Please add/update test somewhere in here: https://github.com/llvm/torch-mlir/tree/main/python/test/torchscript_e2e_test

silvasean (approved):

python/torch_mlir_e2e_test/reporting.py
silvasean: LGTM! Thanks

silvasean (approved):

PR #2205, opened 2023-06-05
Decompose aten.sign and support resulting AtenGeScalarOp in TOSA

mgehre-amd:
This decompose aten.sign into comparisons,
and adds support for lowering aten.ge.scalar into TOSA.

PR #2214, opened 2023-06-07
[Stablehlo] Add torch.aten.sign op to stablehlo

AmosLewis:
Find this issue in Deberta model at https://github.com/nod-ai/SHARK/issues/1442


mgehre-amd:
I think alternatively, you could declare aten.sign as allowed by stable hlo in the `BACKEND_LEGAL_OPS` dict, which should disable my decomposition.

AmosLewis:
> Hi @AmosLewis, there's an overlapping PR for this op, here: #2205. It's decomposing this op into a set of another op, so you might need to add the lowering for this op. You can cherry-pick and try that patch locally.

This patch worked for my case.

PR #2224, opened 2023-06-10
[TCP] Enable type promotion and scalar support for elementwise ops

zezhang:
- Add scalar support for existing binary elementwise ops
- Enable type promotion for existing binary elementwise ops (Directly cast input type to output type if necessary)
- Drop alpha=1 constraint in add/sub ops

Testing:

```
./build/bin/llvm-lit externals/llvm-external-projects/torch-mlir-dialects/test/Conversion/TcpToLinalg/ -v
./build/bin/llvm-lit externals/llvm-external-projects/torch-mlir-dialects/test/Dialect/Tcp/ -v
./build/bin/llvm-lit test/Conversion/TorchToTcp/ -v
python -m e2e_testing.main --config=tcp -v -s
```


lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -60,54 +60,144 @@ class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
...
+                          .template dyn_cast<torch::Torch::ValueTensorType>()
+                          .getDtype();
+    auto outputType = op.getType()
+                          .template dyn_cast<torch::Torch::ValueTensorType>()
+                          .getDtype();
+
+    if (isa<AtenAddScalarOp>(op) || isa<AtenSubScalarOp>(op)) {
+      RankedTensorType tensorResultType =
+          RankedTensorType::get({}, adaptor.getOther().getType());
+      rhs = torch_to_tcp::scalarToTcpTensor(rewriter, op, tensorResultType,
```
asaadaldien, line 84: We should move `scalarToTcpTensor` to be its own `OpConversionPattern` that converts 
`torch.const.int -> tcp.const` and just use `op.getOther`

zezhang: Currently, `tcp.const` only outputs `RankedTensorType`. For ops that still requires the scalar input, always converting `torch.const.int -> tcp.const` will cause problems in the default `backendTypeConversion` pass due to the type mismatch. For elementwise ops with the scalar input,  we use `getConstTensor` utils which creates a `tcp::ConstOp` to get the const tensor first, then cast and broadcast its type and shape. For other places which still requires the `torch.const` input, it will be automatically converted to `arith.const` op. Please let me know if this makes any sense or I misunderstand something.

asaadaldien: Just to make sure I understand correctly: which cases we don't want to lower `torch.const.int -> tcp.const` ?

zezhang: I guess any op that still requires scalar (torch.const) as its input will have this problem. I tried to implement the `torch.const-> tcp.const` conversion pattern for both int and fp, but fails bunch of e2e tests. For example, if we convert `torch.const.int -> tcp.const`, `aten.add.scalar` (which requires a scalar as its second input) will produce following error: BackendTypeConversion.cpp:118: auto setupTorchFloatToF64Conversion(mlir::ConversionTarget &, mlir::TypeConverter &)::(anonymous class)::operator()(mlir::OpBuilder &, mlir::Float64Type, mlir::ValueRange, mlir::Location) const: Assertion `inputs[0].getType().isa<Torch::FloatType>()' failed.

asaadaldien: Aren't we converting `aten.add.scalar -> tcp.add` and broadcasting the scalar to same shape as the lhs ?

navahgar: I agree with @zezhang's point that converting all `torch.const-> tcp.const` would not work because `tcp.const` always results in a tensor. 

This would be a problem for any op that takes in a scalar. For example, `tcp.broadcast` op takes in broadcast sizes as a Variadic list of scalars (https://github.com/llvm/torch-mlir/blob/mlir-tcp/externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpOps.td#L202). If they are constants, we can't be using `tcp.const` to return `index` types (at least with the current specification of `tcp.const`).

Are you suggesting we change `tcp.const` to allow returning scalars as well?

asaadaldien: > Are you suggesting we change tcp.const to allow returning scalars as well?

Yes, I can see use writing less conversion pattern code if we allow `tcp.const` to return scalar or tensor types e.g
```mlir
%0 = tcp.constant 2 : f32
%1 = tensor.from_ements %0 : tensor<f32>
```
will add folding pattern to do:
```mlir
%0 = tcp.constant 2 : tensor<f32>
```
and so on, otherwise every op we will need to have handle scalar operand case.

navahgar (commented):
Made one pass over it. PTAL.

lib/Conversion/TorchToTcp/Elementwise.cpp
navahgar: Agree that it would help reduce the conversion pattern code. But can we base design decisions on that factor?

So far, all TCP ops return only Tensors, none of them produce scalars. My worry is that it might be a bigger change than just updating `tcp.const`, which might have other design implications. We should probably carefully consider this and do it in a separate PR, if there is an agreement. Wdyt?

asaadaldien: > But can we base design decisions on that factor?

My concern is current design converting  `TorchOp -> TCPOp` requires  special handling of `TCPOp` based on if operands are scalar or tensors seems like a gap we have in the `TCP` dialect and how it models constants.

>We should probably carefully consider this and do it in a separate PR, if there is an agreement. Wdyt?

Yes, my comment here is more around understanding current Torch -> TCP approach so I can provide useful feedback on this PR its nonblocking indeed we shouldn't address design changes here!

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -66,6 +68,7 @@ Value torch_to_tcp::broadcastShapeInLeadingDims(
 }
 
 // The parameters input and target are expected to be of RankedTensorType.
+// TODO: didn't cover tailing dims (e.g. [5, 3, 2] * [1, 2])
```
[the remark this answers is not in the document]
zezhang: Delete the comment here since we have added the new utils.

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -83,6 +86,52 @@ Value torch_to_tcp::broadcastInLeadingDimsToMatchShape(
   return result;
 }
 
+// The parameters input are expected to be of RankedTensorType.
+Value torch_to_tcp::broadcastToMatchShapeAndType(
```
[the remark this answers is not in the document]
zezhang: I agree with you. `broadcastInLeadingDimsToMatchShape` currently is not being used anymore. I forget to remove it.

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -83,6 +86,52 @@ Value torch_to_tcp::broadcastInLeadingDimsToMatchShape(
...
+  while (numAxes-- > 0) {
+    if (inputShape[inputAxes] == 1 && targetShape[targetAxes] != 1) {
+      axes.push_back(targetAxes);
+      dimSizes.push_back(rewriter.createOrFold<tensor::DimOp>(
+          op->getLoc(), target, targetAxes));
+      inputShape[inputAxes] = targetShape[targetAxes];
+    }
+    inputAxes--;
+    targetAxes--;
+  }
```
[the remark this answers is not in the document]
zezhang: Please refer to the above comment to see if it makes sense.

lib/Conversion/TorchToTcp/Utils.cpp
```diff
@@ -83,6 +86,52 @@ Value torch_to_tcp::broadcastInLeadingDimsToMatchShape(
...
+  SmallVector<int64_t> inputShape(inputType.getShape().begin(),
+                                  inputType.getShape().end());
+  SmallVector<int64_t> targetShape(targetType.getShape().begin(),
+                                   targetType.getShape().end());
+  Operation *op = input.getDefiningOp();
+  SmallVector<int64_t> axes;
+  SmallVector<Value> dimSizes;
+  int64_t inputAxes = inputShape.size() - 1;
+  int64_t targetAxes = targetShape.size() - 1;
+  int64_t numAxes = std::min(inputType.getRank(), targetType.getRank());
```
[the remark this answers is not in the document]
zezhang: Consider the following case:
lhs = [5,4,3,3,1] --> rank = 5
rhs = [4,3,1,2] --> rank = 4

Currently, to math both operands `lhs` and `rhs`, we do` broadcastToMatchShapeAndType(lhs, rhs)` and then `broadcastToMatchShapeAndType(rhs, lhs)`

When we firstly try to match `lhs` with `rhs`, even if `lhs` has higher rank, we still need to broadcast its last dim to get a [5,4,3,3,2] tensor.
Then, we match `rhs` with `lhs`, rhs firstly get extended to [1,4,3,1,2], then brodcasted to [5,4,3,3,2].

As we can see, broadcast can be needed even if `lhs` and `rhs` have different ranks (especially when `rank(lhs) > rank(rhs)`). Therefore, we should always start with the last dim in both operands to solve the rank mismatch issue.

Please let me know if this makes any sense.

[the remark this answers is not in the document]
navahgar: Aah. I see your point. But that just seems to be an artifact of the way these utilities are setup, and not a fundamental issue.

How about we do this? Given two values, `x`, and `y`, we can make this utility update both the values such that they have a matching shape.

So, the signature of this utility will be:
```
std::pair<Value, Value> broadcastToMatchShape(ConversionPatternRewriter &rewriter, Value val1, Value val2) {
   ...
   auto rankVal1 = ...
   auto rankVal2 = ...
   if (rankVal1 < rankVal2)
     val1 = broadcastRankInLeadingDIms(...)
   else if (rankVal2 < rankVal1)
     val2 = broadcastRankInLeadingDims(...)

  // Now both val1 and val2 have the same rank.
  // Broadcast the shape on both `val1` and `val2` to ensure that they have the same shape.
  ...
  return std::make_pair(val1, val2);
}
```

IMO that is going to lead to simpler code for shape broadcasting as well and will be easier to understand. Wdyt?

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -60,54 +60,144 @@ class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
...
+    auto inputAType = op.getSelf()
+                          .getType()
+                          .template dyn_cast<torch::Torch::ValueTensorType>()
+                          .getDtype();
+    auto outputType = op.getType()
+                          .template dyn_cast<torch::Torch::ValueTensorType>()
+                          .getDtype();
+
+    if (isa<AtenAddScalarOp>(op) || isa<AtenSubScalarOp>(op)) {
+      RankedTensorType tensorResultType =
```
navahgar, line 82: nit: The variable `tensorResultType` sounds like it represents the result of this entire op. Can you rename it to something thats more representative, say, `rank0TensorType`?

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -60,54 +60,144 @@ class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
...
+          RankedTensorType::get({}, adaptor.getOther().getType());
+      rhs = torch_to_tcp::scalarToTcpTensor(rewriter, op, tensorResultType,
+                                            op.getOther());
+      if (adaptor.getOther().getType().template isa<mlir::FloatType>())
+        // FP rhs is treated as fp64
```
navahgar, line 87: Why is this the case? What about f32 additions, for example?

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -60,54 +60,144 @@ class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
...
+        rhs = torch_to_tcp::castTensorToDtype(rewriter, rewriter.getF64Type(),
+                                              outputType, rhs,
+                                              resultType.getElementType());
+      else if (adaptor.getOther().getType().template isa<mlir::IntegerType>())
+        // INT rhs is treated as si64
```
navahgar, line 92: Why is this getting casted to `si64`? There are no signed integers in TCP, right?

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -60,54 +60,144 @@ class ConvertAtenAddSubOp : public OpConversionPattern<AtenOpT> {
...
+        alpha = torch_to_tcp::castTensorToDtype(rewriter, rewriter.getF64Type(),
+                                                outputType, alpha,
+                                                resultType.getElementType());
+      else if (adaptor.getAlpha().getType().template isa<mlir::IntegerType>())
+        // INT alpha is treated as si64
+        alpha = torch_to_tcp::castTensorToDtype(
+            rewriter, rewriter.getIntegerType(64, true), outputType, alpha,
+            resultType.getElementType());
+      else
+        return rewriter.notifyMatchFailure(op, "Unsupported alpha data type");
```
navahgar, line 129: This pattern keeps getting repeated several times. Can we make a utility out of it?

lib/Conversion/TorchToTcp/Elementwise.cpp
zezhang: Renamed to `scalarToTensorType`

lib/Conversion/TorchToTcp/Elementwise.cpp
zezhang: Looks like torch.constant.int will be treated as [64-bit signed value](https://github.com/llvm/torch-mlir/blob/f6a6cfea4e4f71867127504c0315082e1eb8e354/include/torch-mlir/Dialect/Torch/IR/TorchOps.td#L719). So I just use si64 as its type to perform the cast if necessary.

lib/Conversion/TorchToTcp/Elementwise.cpp
zezhang: Looks like torch.constant.float will be treated as [64 bit floating point value](https://github.com/llvm/torch-mlir/blob/f6a6cfea4e4f71867127504c0315082e1eb8e354/include/torch-mlir/Dialect/Torch/IR/TorchOps.td#L741). So I just use fp64 as its type to perform the cast if necessary.

lib/Conversion/TorchToTcp/Elementwise.cpp
```diff
@@ -47,6 +47,29 @@ getTcpSignednessAttr(MLIRContext *context,
...
+    resultValue = torch_to_tcp::castTensorToDtype(
+        rewriter, rewriter.getF64Type(), outputType, resultValue,
+        convertedOutputType);
+  else if (convertedscalarValue.getType().template isa<mlir::IntegerType>())
+    // INT scalarValue is treated as si64
+    resultValue = torch_to_tcp::castTensorToDtype(
+        rewriter, rewriter.getIntegerType(64, true), outputType, resultValue,
+        convertedOutputType);
+  else
+    resultValue = nullptr;
```
navahgar, line 69: Why is this needed? What case would require setting the result to {}?

navahgar (commented):
Mostly LGTM. Just one final comment.

navahgar (approved):
LGTM. Thanks for addressing all the comments.

Please take a look at the build failures.

PR #2227, opened 2023-06-12
[Torch Dialect] fix torch.uint8's dtype infer

qingyunqu:
This is first PR for https://github.com/llvm/torch-mlir/pull/2184.
This PR fix `dtype inference` on torch.uint8.


ramiro050 (approved):
Thanks!

PR #2230, opened 2023-06-12
[Torch Dialect] Support Einsum Op

JianzheXiao:
As title, support torch.aten.einsum op

Right now only support Static Shape, because of the known issue, the fixed solution is here: https://github.com/llvm/torch-mlir/pull/2154

Co-authored-by: Jiawei Wu [wujiawei.aml@bytedance.com](mailto:wujiawei.aml@bytedance.com)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,263 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+      }
+
+      lhsFinalShape.push_back(lhsMiddle);
+      lhsFinalShape.push_back(lhsShape[lhsShape.size() - 1]);
+      
+      finalShape = dotResultShape;
+      // for (auto &it: dotResultShape) {
+      //   finalShape.push_back(rewriter.create<Torch::ConstantIntOp>(
+      //         loc, rewriter.getI64IntegerAttr(it)));
+      // }
```
qingyunqu, line 584: Maybe this comment should be removed.

qingyunqu (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+
+    SmallVector<SmallVector<char>> inputTokens;
+    SmallVector<char> resultTokens;
+    SmallVector<char> inputToken;
+    SmallVector<SmallVector<Value>> inputShapes;
+
+    size_t index = 0;
+    enum EquationVariable { kIsInput, kIsResult };
+    EquationVariable currentVariable = kIsInput;
+    while (index < equation.size()) {
```
ramiro050, line 408: I would move `inputToken` inside the `while` loop since it is only used in here. I don't know what the final state of `inputToken` is after calling `std::move` on it, so having it initialized with every `while` iteration seems safer to me.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+      } else if ((index < (equation.size() - 1)) &&
+                 (equation.substr(index, 2).find("->") != std::string::npos)) {
+        inputTokens.emplace_back(std::move(inputToken));
+        currentVariable = kIsResult;
+        index++;
+      } else {
+        return rewriter.notifyMatchFailure(op, "unexpected character encountered");
+      }
+      index++;
+    }
```
ramiro050, line 426: Can you turn this into a helper function? How we actually parse the equation is pretty independent from how we decompose `Einsum`, so it makes sense to have the logic separate. Also, can you add a string above the helper function explaining the algorithm?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+
+    SmallVector<Value> torchTensors;
+    if (!getListConstructElements(op.getTensors(), torchTensors)) {
+      return rewriter.notifyMatchFailure(
+          op, "input should comes from a PrimListConstructOp");
+    }
+
+    int64_t totalRank = torchTensors.size();
```
ramiro050, line 434: nit: I would call this something like `numOfTensors`, since it is not really a rank

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+    for (int i = 0; i < totalRank; i++) {
+      BaseTensorType tensorType = torchTensors[i].getType().cast<BaseTensorType>();
+      ArrayRef<int64_t> inputShape = tensorType.getSizes();
```
ramiro050, line 437: There is no guarantee that tensors will have sizes at this point, so you need to check `tensorType.hasSizes()` first

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+      int64_t inputRank = inputShape.size();
+      SmallVector<Value> inputValueShape;
+      for (int j = 0; j < inputRank; j++) {
+        Value SizeInt = rewriter.create<AtenSizeIntOp>(
```
ramiro050, line 441: nit: lowercase first letter: `sizeInt`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+          rhsContractingDims, rhsBatchingDims;
+      SmallVector<char> dotResultTokens;
+      SmallVector<Value> dotResultShape;
+      SmallVector<Value> lhsShape = inputShapes[0];
+      SmallVector<Value> rhsShape = inputShapes[1];
+      SmallVector<char> lhsTokens, rhsTokens;
+      lhsTokens = inputTokens[0];
+      rhsTokens = inputTokens[1];
+      collectOperandDims(lhsShape, lhsTokens, rhsTokens, lhsContractingDims,
+                        lhsBatchingDims, dotResultTokens, dotResultShape);
```
ramiro050, line 491: nit: formatting seems off. Make sure you run `git clang-format` to format the code

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+                              : rewriter.create<Torch::ConstantIntOp>(
+                                    loc, rewriter.getI64IntegerAttr(
+                                            inputShape[j])));
+      }
+
+      inputShapes.push_back(inputValueShape);
+    }
+
+
+    auto collectOperandDims =
```
ramiro050, line 456: Can you add a small comment explaining what this does?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4636,6 +4889,7 @@ class DecomposeComplexOpsPass
     addPatternIfTargetOpIsIllegal<DecomposeAtenRandLikeOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenHardsigmoidOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenRelu6Op>(patterns);
+    addPatternIfTargetOpIsIllegal<DecomposeAtenEinsumOp>(patterns);
```
ramiro050, line 4892: Needs to also be added here: https://github.com/llvm/torch-mlir/blob/0caaf8d32abc73e6dee6ff3b39573cc7fdc2929f/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L364

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+            } else if (isOtherToken) {
+              batchingDims.push_back(en.index());
+            } else {
+              dotResultTokens.push_back(en.value());
+              dotResultShape.push_back(operandShape[en.index()]);
+            }
+          }
+        };
+
+    for (int i = 0; i < totalRank - 1; i++) {
```
ramiro050, line 480: Can you add a comment explaining why only loop until `i < totalRank - 1` (I would've expected to loop until `i < totalRank`)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+                              batchingShapeDim);
+      }
+      // Lowering to dot_general does not support a mismatch between the number
+      // of result dims and the number of non-contracting dims.
+      Value constZero =
+          rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(0));
+      Value constOne =
+          rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
+      Value constTwo =
+          rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(2));
```
ramiro050, line 510: nit: I would extract these out of the loop to both make the loop body smaller and avoid making unnecessary IR

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+      SmallVector<Value> lhsFinalShape, rhsFinalShape, finalShape;
+      if (!lhsContractingDims.empty()) {
```
ramiro050, line 512: All these `if-else`s seem to be a little similar. They all do some creation of ints, a transpose, and a swap of values. Is there a common logic that can be extracted to avoid repetition?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -384,6 +384,259 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
 };
 } // namespace
 
+namespace {
+// support static shape only
```
ramiro050, line 388: This decomposition is quite complex. Can you add a string explaining how this decomposition works? Namely, how different cases are identified and what they are decomposed into

ramiro050 (changes requested):
This is a really complex decomposition. Good job getting it working! Initial set of comments

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
JianzheXiao: Removed

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
JianzheXiao: > I would move `inputToken` inside the `while` loop since it is only used in here. I don't know what the final state of `inputToken` is after calling `std::move` on it, so having it initialized with every `while` iteration seems safer to me.

Here we are trying to split input token and result token, the state will not gonna change once it become kIsResult, so it couldn't be put in the while loop, otherwise the state will be initialize every time

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
JianzheXiao: Sure add this part to 

> Can you turn this into a helper function? How we actually parse the equation is pretty independent from how we decompose `Einsum`, so it makes sense to have the logic separate. Also, can you add a string above the helper function explaining the algorithm?

Sure now change this part to helper func

JianzheXiao:
@ramiro050  I just refactor the code, could you please review it again?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -187,6 +187,117 @@ static SmallVector<int64_t> computeDimsOrderForMoveDim(int64_t srcDimInt,
...
+    } else if (equation.substr(index, 1).find(",") != std::string::npos) {
+      inputTokens.push_back(inputToken);
+      inputToken.clear();
+    } else if ((index < (equation.size() - 1)) &&
+               (equation.substr(index, 2).find("->") != std::string::npos)) {
+      inputTokens.push_back(inputToken);
+      inputToken.clear();
+      currentVariable = kIsResult;
+      index++;
+    }
```
ramiro050, line 213: I don't think the equations are guaranteed to be parsable. There should be an `else` branch that returns an error.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -187,6 +187,117 @@ static SmallVector<int64_t> computeDimsOrderForMoveDim(int64_t srcDimInt,
...
+    index++;
+  }
+  return;
+}
+
+// Helper function to process tensor shape.
+static Value processShape(
+    PatternRewriter &rewriter, Location loc, Operation *op, Value inputTensor,
```
ramiro050, line 221: nit: no need to pass `Location loc` if you're passing `Operation *op`. You can extract location from `op`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -187,6 +187,117 @@ static SmallVector<int64_t> computeDimsOrderForMoveDim(int64_t srcDimInt,
...
+            loc, op->getResultTypes(), inputTensor, contractingDim, indexDim);
+        usedOtherDim += 1;
+      }
+    }
+    finalShape.push_back(contractingDimProduct);
+  }
+  return inputTensor;
+}
+
+static Value createReshapedTensor(PatternRewriter &rewriter, Location loc,
```
ramiro050, line 292: nit: I don't think this is really needed, since it does not save that many lines of code

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -389,6 +500,182 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+
+    for (size_t i = 0; i < inputTensors.size(); i++) {
+      BaseTensorType tensorType =
+          inputTensors[i].getType().cast<BaseTensorType>();
+      if (!tensorType.hasSizes()) {
+        return rewriter.notifyMatchFailure(
+            op, "unimplemented: input tensor must have known sizes");
+      }
+      ArrayRef<int64_t> inputShape = tensorType.getSizes();
+      int64_t inputShapeSize = inputShape.size();
```
ramiro050, line 543: nit: unnecessary variable. I would just do `inputShape.size()` in the `for` loop

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -187,6 +187,117 @@ static SmallVector<int64_t> computeDimsOrderForMoveDim(int64_t srcDimInt,
...
+      inputToken.clear();
+      currentVariable = kIsResult;
+      index++;
+    }
+    index++;
+  }
+  return;
+}
+
+// Helper function to process tensor shape.
```
ramiro050, line 219: This is a complicated function. Can you add more detail about what this functions does? What are contracting dims, batching dims, final shape, etc

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -187,6 +187,117 @@ static SmallVector<int64_t> computeDimsOrderForMoveDim(int64_t srcDimInt,
...
+static Value processShape(
+    PatternRewriter &rewriter, Location loc, Operation *op, Value inputTensor,
+    SmallVector<Value> &shape, SmallVector<int64_t> &contractingDims,
+    SmallVector<int64_t> &batchingDims, SmallVector<Value> &finalShape,
+    SmallVector<char> &tokens) {
```
ramiro050, line 224: nit: pass the vectors that are constant by `const` reference to make it clear what is mutated by this function

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -389,6 +500,182 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+    }
+
+    auto collectOperandDims = [resultTokens](
+                                  SmallVector<Value> operandShape,
+                                  SmallVector<char> operandTokens,
+                                  SmallVector<char> others,
+                                  SmallVectorImpl<int64_t> &contractingDims,
+                                  SmallVectorImpl<int64_t> &batchingDims,
+                                  SmallVector<char> &dotResultTokens,
+                                  SmallVector<Value> &dotResultShape) {
```
ramiro050, line 564: nit: pass non-mutated vectors by `const` reference

python/torch_mlir_e2e_test/test_suite/reshape_like.py
```diff
@@ -710,4 +710,62 @@ def forward(self, a):
 
 @register_test_case(module_factory=lambda: ReshapeAliasCollapseModule())
 def ReshapeAliasCollapseModule_basic(module, tu: TestUtils):
-    module.forward(tu.rand(2, 4))
\ No newline at end of file
+    module.forward(tu.rand(2, 4))
+
+# ==============================================================================
+
+class EinsumStaticModule(torch.nn.Module):
```
ramiro050, line 717: Can you also add an e2e test for the single tensor reduction and transposition? (something like `iijk->kj`)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -389,6 +500,182 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+      SmallVector<Value> rhsShape = inputShapes[1];
+      SmallVector<char> lhsTokens = inputTokens[0];
+      SmallVector<char> rhsTokens = inputTokens[1];
+      Value lhsTensor = inputTensors[0];
+      Value rhsTensor = inputTensors[1];
+      // Step 1: split input equation to input/result tokens
+      collectOperandDims(lhsShape, lhsTokens, rhsTokens, lhsContractingDims,
+                         lhsBatchingDims, dotResultTokens, dotResultShape);
+      collectOperandDims(rhsShape, rhsTokens, lhsTokens, rhsContractingDims,
+                         rhsBatchingDims, dotResultTokens, dotResultShape);
```
ramiro050, line 604: I'm a bit confused by what is happening here. Can you explain this step to me? In particular, why are we collecting operands twice with `lhs` and `rhs` changing positions?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -187,6 +187,117 @@ static SmallVector<int64_t> computeDimsOrderForMoveDim(int64_t srcDimInt,
...
+      currentVariable = kIsResult;
+      index++;
+    }
+    index++;
+  }
+  return;
+}
+
+// Helper function to process tensor shape.
+static Value processShape(
```
ramiro050, line 220: is there a more descriptive name we can give to this function? It's not clear to me what it does based on the name alone

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -389,6 +500,182 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+
+      // Step 3: reshape the input tensors, the final shape should
+      // be[batchingDims, otherDims, contractingDims]
+      Value lhs = createReshapedTensor(rewriter, loc, op, op.getType(),
+                                       lhsTensor, lhsFinalShape);
+      Value rhs = createReshapedTensor(rewriter, loc, op, op.getType(),
+                                       rhsTensor, rhsFinalShape);
+      Value result;
+
+      // Step 4: use AtenMmOp/AtenBmmOp to get the result, loop util we```cpp
```
ramiro050, line 636: nit: remove ```cpp

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
JianzheXiao: > I'm a bit confused by what is happening here. Can you explain this step to me? In particular, why are we collecting operands twice with `lhs` and `rhs` changing positions?

In this function we are trying to find which dim should be contracted and which dim should be keep,  for example if we have equation "bcwd,bcdh->bcwh",  we could find b and c should be batched and d should be contracted

JianzheXiao:
@ramiro050 Could you please review this pr again? Thank you!

JianzheXiao:
@ramiro050  Could you please review this pr again? Thank you!

Vremold (commented):
A few comments. Can we make sure that all einsum ops like matmul or Hadamard product pass this decomposition pattern, e.g., an einsum op with equation "abc,abc->"?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -410,6 +528,199 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+        if (dotResultTokens.size() != resultTokens.size()) {
+          return rewriter.notifyMatchFailure(op,
+                                            "rank reducing einsum not supported");
+        }
+        int64_t resultSize = 0;
+        for (char resultToken : resultTokens) {
+          auto *foundIt = std::find(dotResultTokens.begin(), dotResultTokens.end(),
+                                    resultToken);
+          if (foundIt == dotResultTokens.end()) {
+            return rewriter.notifyMatchFailure(
```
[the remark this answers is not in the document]
JianzheXiao: Here we are trying to compare whether the result we calculated are equal with the resultTokens, I think there is no easy way to make this validation at the beginning

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -410,6 +528,199 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+      Value rhs = rewriter.create<AtenReshapeOp>(loc, op.getType(), rhsTensor, rhsReshapedDims);
+      Value result;
+
+      // Step 4: use AtenMmOp/AtenBmmOp to get the result, loop util we
+      // get the final result
+      if (!lhsBatchingDims.empty()) {
+        rhs = rewriter.create<AtenTransposeIntOp>(loc, op.getType(), rhs, constOne, constTwo);
+        result = rewriter.create<AtenBmmOp>(loc, op.getType(), lhs, rhs);
+      } else {
+        rhs = rewriter.create<AtenTransposeIntOp>(loc, op.getType(), rhs, constZero, constOne);
```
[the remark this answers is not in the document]
JianzheXiao: Fixed and add an e2e test to verify

JianzheXiao:
> A few comments plus a question. Can we make sure that all einsum ops like matmul or Hadamard product pass this decomposition pattern, e.g., an einsum op with equation "abc,abc->"?

matmul should be okay but right now the algorithm are not able to deal with the case like abc, abc-> abc

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -410,6 +528,199 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+// Decompose AtenEinsumOp to AtenMmOp or AtenBmmOp
+// Step 1: split input equation to input/result tokens and find batchingDims and
+// contractingDims for future use
+// Step 2: transpose the input tensors to [batchingDims[0,1,2],
+// otherDims[0,1,2], contractingDims[0,1,2]]
+// Step 3: reshape the input tensors, the final shape should
+// be[batchingDims, otherDims, contractingDims]
+// Step 4: use AtenMmOp/AtenBmmOp to get the result, loop util we get the final
+// result
+// notice: support static shape only
```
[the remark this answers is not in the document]
JianzheXiao: When I used dynamic shape, the FinalizingBufferize will always failed to legalize operation 'mhlo.dynamic_reshape', maybe we need to find the reason

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -187,6 +187,387 @@ static SmallVector<int64_t> computeDimsOrderForMoveDim(int64_t srcDimInt,
...
+                                   Value rhs, SmallVector<char> &rhsTokens,
+                                   Value &result,
+                                   SmallVector<char> &resultTokens,
+                                   SmallVector<char> &finalResultTokens) {
+  auto lhsType = lhs.getType().cast<BaseTensorType>();
+  auto rhsType = rhs.getType().cast<BaseTensorType>();
+  Type promotedDType;
+
+  // promote dtype
+  if (lhsType.hasDtype() && rhsType.hasDtype()) {
```
ramiro050, line 388: The type promotion should be handled by the dtype functions. What's the reason for not having shape and dtype functions?

ramiro050 (changes requested):
Looks mostly good to me, just one question

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
JianzheXiao: the problem here is that we have some cases that einsum operands can have differing dtypes, it becomes an issue when we attempt to decompose it into mm or bmm operations, dtype mismatches would result in errors. Also, shape inference for the einsum operation, which necessitates the analysis of the equation string that specifies the subscripts for summation, but right now it's not supported by abstract_interp_lib_gen

JianzheXiao: > The type promotion should be handled by the dtype functions. What's the reason for not having shape and dtype functions?

the problem here is that we have some cases that einsum operands can have differing dtypes, it becomes an issue when we attempt to decompose it into mm or bmm operations, dtype mismatches would result in errors. Also, shape inference for the einsum operation, which necessitates the analysis of the equation string that specifies the subscripts for summation, but right now it's not supported by abstract_interp_lib_gen

ramiro050: I would've imagined that the result type is simply the promotion of all the dtypes in the input tensors list. An example of how to handle a list of tensors is

https://github.com/llvm/torch-mlir/blob/53fc995639e26f40465516dded26548928bd2d1a/projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py#L3651

Also, let's add an e2e test that has inputs of different dtypes to test this behavior.

For the shape information, I think if we just leave the result types of the different `mm` and `bmm` ops as having no shape information, the shape propagation passes should take care of filling those in.

JianzheXiao: > I would've imagined that the result type is simply the promotion of all the dtypes in the input tensors list. An example of how to handle a list of tensors is
> 
> https://github.com/llvm/torch-mlir/blob/53fc995639e26f40465516dded26548928bd2d1a/projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py#L3651
> 
> Also, let's add an e2e test that has inputs of different dtypes to test this behavior.
> 
> For the shape information, I think if we just leave the result types of the different `mm` and `bmm` ops as having no shape information, the shape propagation passes should take care of filling those in.

I have removed the original type promotion part and added the corresponding type function. Additionally, I found that the Torch frontend does not allow for different input types in einsum, so it might be an issue arising during our model conversion process.

ramiro050 (approved):
Thanks for the hard work on this op 🙂

qingyunqu (approved):

PR #2233, opened 2023-06-14
[MLIR][TORCH] Add support for negative index values for index.Tensor op

vivekkhandelwal1:
This commit adds the support for index.Tensor op when the index values are negative. This commit wraps around the index values by checking their values at run time.

Fixes https://github.com/openxla/iree/issues/13951.


ramiro050 (approved):
Thanks!

PR #2234, opened 2023-06-14
TorchToTosa: Legalization for torch.aten.sqrt

ttjost:
It supports both integer and floating point types.
For integer, the output type will automatically be converted to floating point, therefore, there is an extra cast needed for those cases.


sjarus:
Is this quite what you want to do ? The [TOSA spec](https://www.mlplatform.org/tosa/tosa_spec.html) doesn't define pow, reciprocal or rqsrt for integers. [Casting](https://www.mlplatform.org/tosa/tosa_spec.html#_cast) to fp could work, but there will be an impact on numerical stability due to rounding behavior. 

TFLite to TOSA has an implementation of quantized integer rqsrt that use numerically precise tosa.table() based implementation which may be better suited here: https://github.com/tensorflow/tensorflow/blob/master/tensorflow/compiler/mlir/tosa/transforms/legalize_tfl.cc#L2518 

@eric-k256

sjarus (commented):
Left comment.

eric-k256 (commented):
+1 on what Suraj said. I would expect that if we are being called to do a sqrt on an integer doing it as a table with integer output is likely closer to what the original network was looking for.

ttjost:
Thank you very much for your comments.
The problem is that `torch` expects the output element type to be floating point. So no matter which method we use, the output type is a FP tensor.
I would expect that using `TableOp` would give less numerical stability than casting it to a FP type and perform `pow(x, 0.5)` legalization, specially since `ElementwiseSqrtIntModule_basic` test uses `torch.int` as input element type, which is a supported data type for the `TableOp` operation.
Am I missing something?

mgehre-amd:
Ping.
I think the important aspect here is that torch.sqrt(integer value) behaves like (torch.sqrt(cast-to-float(integer)), i.e. we are expected to do a floating point sqrt here.

eric-k256:
> Ping. I think the important aspect here is that torch.sqrt(integer value) behaves like (torch.sqrt(cast-to-float(integer)), i.e. we are expected to do a floating point sqrt here.

I agree that this is the key point. Assuming the behavior described of torch.sqrt on integers, then this legalization is fine.

If we added support for quantized int sqrt handling, that would get us to a point where we had to make a decision. If we were dealing with quantized integer sqrt on a system that did not support floating-point ops, then tosa.table would be the right legalization. We would have to choose the different paths based on the output type of the operation (or profile based, or through some external mechanism)

Given that it would take time to get PyTorch changed, it seems reasonable to allow the proposed legalization in, and postpone dealing with quantized sqrt until it exists.

eric-k256 (approved):

PR #2240, opened 2023-06-16
Add make_fx_tosa variant to end2end tests

mgehre-amd:
This adds
- `use_make_fx` flag to torch_mlir.compile
- A `make_fx_tosa` configuration to e2e tests
- `MAKE_FX_TOSA_PASS_SET ` to xfail sets, which is defined on top of the TOSA_PASS_SET


silvasean (approved):

mgehre-amd:
This seems to hit a bug in upstream tosa::AddOp::fold. I'm checking whether this is already resolved upstream.

mgehre-amd:
Fix for the crash is here: https://reviews.llvm.org/D154615

mgehre-amd:
While we wait for the next LLVM bump, #2289 turns out to also avoid hitting the bug.

PR #2246, opened 2023-06-20
Add support for aten.tile operator(not e2e support)

georgeuser:


include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
```diff
@@ -12089,3 +12089,27 @@ def Torch_QuantizedLinearOp : Torch_Op<"quantized.linear", [
   }];
 }
 
+def Torch_AtenTileOp : Torch_Op<"aten.tile", [
```
ramiro050, line 12092: Do you just need the ODS? If so, no e2e test is needed.

georgeuser: yes , I've just add the ODS here.

ramiro050: Then there's no need to add all the other changes, just the changes in `GeneratedTorchOps.td`

georgeuser: Please see if this commit is Ok.

ramiro050 (approved):

ramiro050:
Make sure you're generating the ODS using `build_tools/update_torch_ods.sh`. The CI is failing because the current ODS does not match the output of `build_tools/update_torch_ods.sh`

PR #2251, opened 2023-06-20
Fix GHA workflows on `mlir-tcp`

sjain-stanford:
As titled. Should fix issues like [this](https://github.com/llvm/torch-mlir/actions/runs/5314715542/jobs/9647290496?pr=2224).


zezhang (approved):
Thanks for looking into this!

PR #2255, opened 2023-06-21
[MLIR][TORCH] Add E2E support for aten.embedding_dense_backward op

vivekkhandelwal1:
This commit adds the support for aten.embedding_dense_backward op by decomposing it into a set of other torch ops.
This commit also removes this op from the list of decompositions used by the dynamo compilation. Earlier, it was added to the list because there was no lowering present for this op in Torch-MLIR.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4533,6 +4533,127 @@ class DecomposeAtenSignOp : public OpRewritePattern<AtenSignOp> {
 };
 } // namespace
 
+// The `embedding_dense_backward` op is decomposed as follows:
```
mgehre-amd, line 4536: I wonder if managing the decomposition here instead of relying on the upstream decomposition in dynamo is a scalable approach in the long term.

I was actually expecting us to do the opposite and move our decompositions into pytorch over time.
(Also considering that the python code for the decomposition is shorter and more readable than the corresponding MLIR C++ code).

vivekkhandelwal1: Hi @mgehre-amd, this comment (https://github.com/nod-ai/SHARK/pull/1567#issuecomment-1601062045) might provide more context. Please let me know if you still have any doubts or thoughts.

vivekkhandelwal1: > I wonder if managing the decomposition here instead of relying on the upstream decomposition in dynamo is a scalable approach in the long term.

Sorry, I missed that your query is specific to the dynamo. Yeah, we can use the upstream PyTorch decomposition it's just a choice. I can revert the dynamo changes to use the upstream decomposition instead of the Torch-MLIR one. @ramiro050 Should I do that?

vivekkhandelwal1: It seems the Long term roadmap (https://github.com/llvm/torch-mlir/blob/main/docs/images/roadmap_frontend.png) also mentions that we should reuse the PyTorch infra. I think for a temporary solution to unblock us this patch can be merged(as there is no harm, and there exists a lot of such instances already) and then we can take a look at fixing the issue in upstream PyTorch.

ramiro050: I think fixing things upstream would require very simple changes (changing `_unsafe_index` to `index_impl` or adding type annotations). If the decomposition is broken for us, it is likely broken for other PyTorch users, so I would prefer we fix it upstream than have a workaround here.

vivekkhandelwal1:
Closing this PR. Since it's not required anymore. https://github.com/pytorch/pytorch/pull/104127 fixed this.

PR #2256, opened 2023-06-21
[Torch Dialect] avoid assertion failure when PrimNumToTensorScalarOp'…

qingyunqu:
…s input is torch.number

### DIFF
--- a/lib/Dialect/Torch/Transforms/SimplifyDtypeCalculations.cpp
+++ b/lib/Dialect/Torch/Transforms/SimplifyDtypeCalculations.cpp
@@ -171,6 +171,11 @@ public:
       return rewriter.notifyMatchFailure(
           op, "`PrimNumToTensorScalarOp` already has a dtype");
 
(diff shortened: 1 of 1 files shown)


lib/Dialect/Torch/Transforms/SimplifyDtypeCalculations.cpp
```diff
@@ -171,6 +171,11 @@ class RefineNumToTensorScalarOpType
       return rewriter.notifyMatchFailure(
           op, "`PrimNumToTensorScalarOp` already has a dtype");
 
+    if (op.getA().getType().isa<Torch::NumberType>()) {
+      return rewriter.notifyMatchFailure(
+          op,
+          "`PrimNumToTensorScalarOp`'s input should has concrete Scalar Type.");
```
ramiro050, line 177: nit: typo "has" -> "have"

ramiro050 (approved):
Good catch!

PR #2258, opened 2023-06-22
Fix single-element tuple construction in abstract interp library

ramiro050:
Single element tuples in Python need a comma after the element. However, the `registry.py` file, which generates the expected abstract interpretation function signatures, was not inserting the comma. This commit changes the expected signature generator to add a comma after the last element in any non-empty default tuple argument.


qingyunqu (approved):
LGTM

pashu123 (approved):

PR #2259, opened 2023-06-22
[Torch Dialect] Support aten.native_dropout

qingyunqu:


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4654,6 +4706,7 @@ class DecomposeComplexOpsPass
     addPatternIfTargetOpIsIllegal<DecomposeAten_ToCopyOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenCopyOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenDropoutOp>(patterns);
+    addPatternIfTargetOpIsIllegal<DeomposeAtenNativeDropoutOp>(patterns);
```
ramiro050, line 4709: Need to also add here: https://github.com/llvm/torch-mlir/blob/38fb99df658faf46e1266a684b0900d183167371/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L364

ramiro050 (approved):
Thanks! Just one small change request

PR #2262, opened 2023-06-25
[Stablehlo] fix promoteType() when input doesn't have DefiningOp

qingyunqu:


Vremold:
Thanks for the correction.

PR #2264, opened 2023-06-25
[Torch Dialect] add folder for aten.add

qingyunqu:


qingyunqu:
Also add an e2e test which doesn't pass previously because lacking folder for `aten.add`.

ramiro050 (approved):

PR #2268, opened 2023-06-26
[Stablehlo]Add support for AvgPool1dOp

JianzheXiao:
Add support for AvgPool1dOp in Stablehlo
Add support for AvgPool1dOp in Linalg As well


lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -31,7 +31,52 @@ using namespace mlir::torch::Torch;
 // vector.
 template <typename OpTy>
 static LogicalResult
-checkAndGetPoolingParameters(OpTy op, ConversionPatternRewriter &rewriter,
+checkAndGetPooling1dParameters(OpTy op, ConversionPatternRewriter &rewriter,
```
ramiro050, line 34: This function is almost identical to the 2d case. I think it should work for both 1d and 2d if you change

```
SmallVector<Value, 1> kernelSizeTorchInt;
```

to

```
SmallVector<Value> kernelSizeTorchInt;
```

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -72,10 +117,72 @@ checkAndGetPoolingParameters(OpTy op, ConversionPatternRewriter &rewriter,
   return success();
 }
 
+
+// Creates a pooling operation based on the type specified by `OpTy` and
+// arguments passed.
+template <typename OpTy>
+static LogicalResult createPooling1dOp(
```
ramiro050, line 124: This lowering is very similar to the 2d case. The common logic should be identified and refactored to avoid repetition.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -366,6 +473,86 @@ class ConvertAtenMaxPool2dWithIndicesOp
 };
 } // namespace
 
+
+namespace {
+class ConvertAtenAvgPool1dOp : public OpConversionPattern<AtenAvgPool1dOp> {
```
ramiro050, line 478: This pattern shares a lot of code with `ConvertAtenAvgPool2dOp`. Common logic should be identified and refactored to avoid repetition.

lib/Conversion/TorchToStablehlo/Pooling.cpp
```diff
@@ -373,6 +373,164 @@ LogicalResult ConvertAtenOp<AtenMaxPool2dWithIndicesOp>::matchAndRewrite(
   return success();
 }
 
+
+// AtenAvgPool1dOp
+template <>
+LogicalResult ConvertAtenOp<AtenAvgPool1dOp>::matchAndRewrite(
```
ramiro050, line 379: Same comment as above. Common logic should be identified and refactored to avoid repetition.

ramiro050 (changes requested):
Thanks for working on this! There is a bit of code repetition in this PR that needs to be refactored. Avoiding repetition makes things easier to maintain, since you only need to fix bugs in a single place rather than in multiple places.

lib/Conversion/TorchToLinalg/Pooling.cpp
JianzheXiao: Thank you! Fixed

lib/Conversion/TorchToLinalg/Pooling.cpp
JianzheXiao: Fixed,  have already merge these two functions.

lib/Conversion/TorchToStablehlo/Pooling.cpp
JianzheXiao: > Same comment as above. Common logic should be identified and refactored to avoid repetition.

Thank you! Have already merged this two Conversion by create a new template

lib/Conversion/TorchToLinalg/Pooling.cpp
JianzheXiao: > This lowering is very similar to the 2d case. The common logic should be identified and refactored to avoid repetition.

Same, Fixed by Merge them together

qingyunqu:
@ramiro050 cloud you please review this PR again? Thanks!

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -72,12 +72,13 @@ checkAndGetPoolingParameters(OpTy op, ConversionPatternRewriter &rewriter,
...
 }
 
+
 // Creates a pooling operation based on the type specified by `OpTy` and
 // arguments passed.
 template <typename OpTy>
 static LogicalResult createPoolingOp(
     Operation *op, ConversionPatternRewriter &rewriter, Value self,
-    bool supportNonFPInput, bool ceilMode,
+    bool supportNonFPInput, bool ceilMode, int64_t dimension,
```
ramiro050, line 81: nit: I would call this `dimensionality` to avoid confusion with the use of `dimension` to denote a specific axis of a tensor

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -177,8 +182,9 @@ class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
     Value maxPool2d, paddedInput;
     if (failed(createPoolingOp<linalg::PoolingNchwMaxOp>(
             op, rewriter, self, /*supportNonFPInput=*/false, ceilMode,
-            kernelSizeIntValues, strideInts, paddingInts, dilationInts,
-            smallestFPValueAttr, outTensorShape, paddedInput, maxPool2d)))
+            /*dimension*/2, kernelSizeIntValues, strideInts, paddingInts,
```
ramiro050, line 185: nit: comment is missing an `=`: `/*dimension=*/`

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -253,8 +259,9 @@ class ConvertAtenMaxPool2dWithIndicesOp
     SmallVector<Value, 4> outTensorShape;
     if (failed(createPoolingOp<linalg::PoolingNchwMaxOp>(
             op, rewriter, self, /*supportNonFPInput=*/false, ceilMode,
-            kernelSizeIntValues, strideInts, paddingInts, dilationInts,
-            smallestFPValueAttr, outTensorShape, paddedInput, maxPool2d)))
+            /*dimension*/2, kernelSizeIntValues, strideInts, paddingInts,
```
ramiro050, line 262: Same comment: missing `=`

ramiro050 (approved):
Thanks for this! Other than a few nits, LGTM

lib/Conversion/TorchToLinalg/Pooling.cpp
JianzheXiao: Fixed

lib/Conversion/TorchToLinalg/Pooling.cpp
JianzheXiao: > 

Fixed

PR #2286, opened 2023-07-06
[Torch Dialect] fix DerefineOp's type check

qingyunqu:
More general fix for previous PR: https://github.com/llvm/torch-mlir/pull/2208


ramiro050:
The issue was not with the `derefine` op, but rather with the way the abstract interpretation library was handling the `!torch.number` type. I have made a fix here: https://github.com/llvm/torch-mlir/pull/2309

PR #2289, opened 2023-07-06
DecomposeComplexOps: Use static shape if available

mgehre-amd:
It seems generally better to preserve the information we have. This makes the live easier for TOSA, which doesn't handle dynamic shapes very well.


eric-k256 (approved):
Looks good to me.

PR #2291, opened 2023-07-08
Add support for custom ops and quantized matmul

gpetters94:
This is consolidation of work by @jinchen62 and @makslevental, who can explain it better than I can, but it (or something equivalent) is necessary for int8 and int4 model support downstream.


include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -829,6 +829,7 @@ def Torch_DerefineOp : Torch_Op<"derefine", [
 }
 
 def Torch_OperatorOp : Torch_Op<"operator", [
+    ReadOnly,
```
mgehre-amd, line 832: How do we know that all currently unsupported ops are ReadOnly?

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1606,6 +1606,97 @@ OpFoldResult ValueTensorLiteralOp::fold(FoldAdaptor adaptor) {
...
+
+    OpOperand *use = constOp.getResult().use_begin().getOperand();
+    auto op = dyn_cast<OperatorOp>(use->getOwner());
+    if (!op)
+      return failure();
+
+    if (use->getOperandNumber() != 1)
+      return failure();
+
+    if (op.getName().str() != "brevitas.matmul_rhs_group_quant") {
```
mgehre-amd, line 1623: Instead of using torch.operator here, can we add `brevitas.matmul_rhs_group_quant` to TorchOps.td?

powderluv:
The TorchExtension piece should land separately. I don't think we should merge these two into this PR (other than for testing end to end)

makslevental:
> The TorchExtension piece should land separately. I don't think we should merge these two into this PR (other than for testing end to end)

wholly agree

PR #2295, opened 2023-07-10
[Stablehlo] add converter for aten.scatter.src op

Vremold:
Since for now `hlo-legalize-to-linalg` pass doesn't support `stablehlo::Scatter` op, this patch won't pass relevant E2E tests. Instead, we add a MLIR unittest case.


qingyunqu (approved):
LGTM

PR #2297, opened 2023-07-11
[Torch-Dialect] emit aten.narrow.Tensor op and decompose it to aten.narrow op

Vremold:
This patch emits `aten.narrow.Tensor` op and decomposes it into `aten.narrow` op. 
To make relevant e2e tests pass, this patch also strength the folder for `aten.ScalarImplicit` and `aten.int.tensor` op.
And this patch also adds a converter to stablehlo for `aten.tensor.int` op.


Vremold:
Could you please review the PR? @ramiro050

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -900,6 +900,11 @@ def aten〇sort〡dtype(self_rank_dtype: Tuple[int, int], dim: int = -1, descend
 def aten〇narrow〡shape(self: List[int], dim: int, start: int, length: int) -> List[int]:
     return upstream_shape_functions.slice(self, dim, start, start + length, 1)
 
+# This shape function is a little hacky, because we don't know the start index which is determined by a tensor param.
+# Let's assume it to be 0.
+def aten〇narrow〇Tensor〡shape(self: List[int], dim: int, start: List[int], length: int) -> List[int]:
+    return upstream_shape_functions.slice(self, dim, 0, length, 1)
```
ramiro050, line 906: I think we can do

```python
self[dim] = length
return self
```

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
Vremold: Done.

ramiro050: I don't see the changes, but other than this, LGTM

ramiro050 (approved):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
Vremold: > I don't see the changes, but other than this, LGTM

Sorry for that. I mistakenly modify the shape inference function for `aten.narrow` rather than `aten.narrow.Tensor.` I will fix it. Thanks for your approval.

PR #2301, opened 2023-07-12
[Stablehlo]Support None index Tensor in AtenIndexTensor

JianzheXiao:
As title, support none index tensors at the beginning or end of the indices in AtenIndexTensor


Vremold (commented):
I think a better way to handle NoneType index might be writing a 'converter' (like a canonicalizer for `aten::intex.Tensor`) in Torch-Dialect to eliminate none value in `indices` param. What's your opinion? @ramiro050

zhekunz2:
> I think a better way to handle NoneType index might be writing a 'converter' (like a canonicalizer for `aten::intex.Tensor`) in Torch-Dialect to eliminate none value in `indices` param. What's your opinion? @ramiro050

I don't think we should eliminate none values. `None` in `indices` has semantic meaning that select all elements from that dimension.

ramiro050:
> I think a better way to handle NoneType index might be writing a 'converter' (like a canonicalizer for `aten::intex.Tensor`) in Torch-Dialect to eliminate none value in `indices` param. What's your opinion? @ramiro050

I think it's necessary to know which indices are `None` when lowering to the backend. Depending on where the `None` indices are in the list (all together or interspersed with index tensors) changes the indexing. See: https://numpy.org/doc/stable/user/basics.indexing.html#combining-advanced-and-basic-indexing

mgehre-amd:
We have a PR for that here https://github.com/Xilinx/torch-mlir/pull/98/files#diff-552495083150eaaece8db6c1824216ddd5780a3a0bb46499a92799e1e66d5a02 but I didn't have time for upstreaming yet.

Vremold:
> I think a better way to handle NoneType index might be writing a 'converter' (like a canonicalizer for `aten::intex.Tensor`) in Torch-Dialect to eliminate none value in `indices` param. What's your opinion? @ramiro050

What I mean here is that we can replace the none value with manually generated index (probably a iota op with specific shape), since we've already know what the behaviour is.

ramiro050:
> What I mean here is that we can replace the none value with manually generated index (probably an iota op with specific shape), since we've already know what the behaviour is.

I think I understand now. Yeah, I think this would be nice to have! It would simplify things for the backends quite a bit

PR #2302, opened 2023-07-12
Filter out empty strings while generating function signature

nithinsubbiah:


makslevental (approved):
wow how annoying! good job tracking it down

PR #2306, opened 2023-07-14
[Torch-Dialect] add canonicalizer for prim::ListConstruct op

Vremold:


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2085,6 +2085,36 @@ void PrimTupleUnpackOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+void PrimListConstructOp::getCanonicalizationPatterns(
+    RewritePatternSet &patterns, MLIRContext *context) {
+  patterns.add(+[](PrimListConstructOp op, PatternRewriter &rewriter) {
+    if (isListPotentiallyMutated(op.getResult())) 
+      return failure();
+    SmallVector<Value> elements = llvm::to_vector<4>(op.getElements());
+    if (elements.size() == 0)
+      return failure();
+
+    auto listUnpackOp = elements[0].getDefiningOp<PrimListUnpackOp>();
```
mgehre-amd, line 2101: Why do we only need to check the first element?

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2085,6 +2085,36 @@ void PrimTupleUnpackOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+    if (!listUnpackOp)
+      return failure();
+    SmallVector<Value, 4> listUnpackResults;
+    for (auto v : listUnpackOp.getResults()) {
+      listUnpackResults.emplace_back(v);
```
mgehre-amd, line 2106: Instead of adding those into `listUnpackResults`, can we directly compare against `elements`?

Vremold: Good idea! The new patch fixes this.

test/Dialect/Torch/canonicalize.mlir
```diff
@@ -965,6 +965,15 @@ func.func @torch.prim.If$fold_same_result$subset_of_results(%arg0: !torch.bool,
...
 }
 
+// CHECK-LABEL: func.func @prim.ListConstruct$fold_list(
+// CHECK-SAME:      %[[ARG0:.*]]: !torch.list<tensor>) -> !torch.list<tensor> {
+// CHECK:         return %[[ARG0]] : !torch.list<tensor>
+func.func @prim.ListConstruct$fold_list(%arg0: !torch.list<tensor>) -> !torch.list<tensor> {
+  %0:2 = torch.prim.ListUnpack %arg0 : !torch.list<tensor> -> !torch.tensor, !torch.tensor
+  %1 = torch.prim.ListConstruct %0#0, %0#1 : (!torch.tensor, !torch.tensor) -> !torch.list<tensor>
+  return %1 : !torch.list<tensor>
+}
```
qingyunqu, line 975: I think it's necessary to check that `%arg0`'s length is same as `NumResults` of `prim.Listunpack`.

Vremold: We only replace the later ListConstruct op with the former ListUnpack's operand. ListUnpack op will not be eliminated.

Vremold:
Could you review the PR for me? @ramiro050

mgehre-amd (approved):
LGTM

PR #2307, opened 2023-07-14
Don't crash when the input to aten.copy is unranked

mgehre-amd:
This can happen when the input comes from an unsupported operator.
`convertTensorToDtype` below relies on that information.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -3162,6 +3162,11 @@ public:
       return rewriter.notifyMatchFailure(
           op, "expected result type to have a dtype");
     }
+    auto srcTy = op.getSrc().getType().cast<BaseTensorType>();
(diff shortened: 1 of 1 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3162,6 +3162,11 @@ class DecomposeAtenCopyOp : public OpRewritePattern<AtenCopyOp> {
       return rewriter.notifyMatchFailure(
           op, "expected result type to have a dtype");
     }
+    auto srcTy = op.getSrc().getType().cast<BaseTensorType>();
+    if (!srcTy.hasSizes() || !srcTy.hasDtype()) {
+      return rewriter.notifyMatchFailure(
+          op, "expected src type to have a known rank");
```
ramiro050, line 3168: nit: "... known rank **and dtype**"

ramiro050 (approved):
Thanks!

PR #2309, opened 2023-07-14
Fix handling of `!torch.number` in abstract interpretation library

ramiro050:
In PyTorch, the `NumberType` is equal to `Union[int, float, complex]`. However, the abstract interpretation library was treating the `NumberType` as `Union[int, float]`, resulting in type mismatches when reifying certain dtype functions. This commit fixes the type inconsistency by having the abstract interpretation functions take as an input a `Union[int, float, complex]` for the ops that take `!torch.number` inputs.


qingyunqu (approved):
Great! Thanks!
LGTM

PR #2313, opened 2023-07-16
[Torch Dialect] emit aten.nonzero op, recompose aten.index.Tensor with aten.nonzero_numpy to aten.masked_select

Vremold:
In detail, this PR includes:
1. emit `aten.nonzero` and `aten.nonzero_numpy` op;
2. add missing shape infer function for `aten.masked_select` op
3. add recomposition rule for `aten.index.Tensor` op. When it's indices come from `aten.nonzero_numpy` op and the shapes of associated tensor are exactly same, it's equivalent to a `aten.masked_select` op. But for now, no backend supports `aten.masked_select` or `aten.nonzero_numpy` op, this PR only adds a MLIR test case.


lib/Dialect/Torch/Transforms/RecomposeComplexOps.cpp
```diff
@@ -418,6 +418,116 @@ class RecomposeChunkListUnpack : public OpRewritePattern<PrimListUnpackOp> {
...
+    Value mask = nonzeroNumpyOp.getSelf();
+    auto inputType = input.getType().cast<BaseTensorType>();
+    auto maskType = mask.getType().cast<BaseTensorType>();
+
+    // Because this pass is run before reduce-op-variants and
+    // maximize-value-semantics pass, operation operands might already be
+    // converted through static_info_cast or copy.to_tensor op. We need to
+    // restore the shape information to judge whether we can convert the op.
+    auto originalMaskInfo = getOriginalValue(mask);
+    auto originalInputInfo = getOriginalValue(input);
```
ramiro050, line 472: I'm a bit confused by the need for `getOriginalValue`. Can you show me the IR you're seeing that causes issues? I would expect that returning `failure()` when the tensors don't have shape information would work, since the pipeline is applied multiple times, so at some point the shape information would propagate all the way.

lib/Dialect/Torch/Transforms/RecomposeComplexOps.cpp
```diff
@@ -418,6 +418,116 @@ class RecomposeChunkListUnpack : public OpRewritePattern<PrimListUnpackOp> {
...
+              "have same-ranked shape");
+    }
+
+    for (int64_t i = 0; i < originalInputSizes.size(); ++i) {
+      if (originalInputSizes[i] != originalMaskSizes[i] ||
+          originalInputSizes[i] == ShapedType::kDynamic ||
+          originalMaskSizes[i] == ShapedType::kDynamic) {
+        return rewriter.notifyMatchFailure(
+            op, "inputs of aten::index.tensor and aten::nonzero_numpy should "
+                "have same static shape");
```
ramiro050, line 497: Why doesn't this work for dynamic sizes?

lib/Dialect/Torch/Transforms/RecomposeComplexOps.cpp
```diff
@@ -418,6 +418,116 @@ class RecomposeChunkListUnpack : public OpRewritePattern<PrimListUnpackOp> {
     return success();
   }
 };
+
+// recompose `aten::nonzero_numpy` + `prim::ListUnpack` + `prim::ListConstruct` + `aten::index.Tensor` with
```
ramiro050, line 422: nit: there is no `ListConstruct` being handled in this pattern

lib/Dialect/Torch/Transforms/RecomposeComplexOps.cpp
```diff
@@ -418,6 +418,116 @@ class RecomposeChunkListUnpack : public OpRewritePattern<PrimListUnpackOp> {
...
+    Value boolMask =
+        rewriter.create<AtenToDtypeOp>(loc, boolMaskType, mask, i1DtypeValue,
+                                       falseValue, falseValue, noneValue);
+
+    rewriter.replaceOpWithNewOp<AtenMaskedSelectOp>(op, op.getType(), input,
+                                                    boolMask);
+
+    // Separately handle possible prim::ListUnpack op, if it's a user of the
+    // aten::nonzero_numpy op.
+    if (nonzeroNumpyOp.getResult().hasOneUse()) {
```
ramiro050, line 516: A couple of things on this:

1. Can you add a lit test for this case?
2. This should be a separate pattern, since it is fully independent of the `AtenIndexTensorOp`

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/torch_ods_gen.py
```diff
@@ -446,6 +446,8 @@ def emit_with_mutating_variants(key, **kwargs):
     emit("aten::mse_loss_backward : (Tensor, Tensor, Tensor, int) -> (Tensor)")
     emit("aten::upsample_nearest2d_backward : (Tensor, int[], int[], float?, float?) -> (Tensor)")
     emit("aten::cross_entropy_loss : (Tensor, Tensor, Tensor?, int, int, float) -> (Tensor)")
+    emit("aten::nonzero : (Tensor) -> (Tensor)")
```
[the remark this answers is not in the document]
Vremold: Okay

test/Dialect/Torch/recompose-complex-ops.mlir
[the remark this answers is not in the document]
Vremold: It‘s okay. In my side, it's hard for stablehlo to represent the dynamic semantics of aten.masked_select op. If any folks support it's lowering to linalg, tosa or other backend, we'll be happy to see that.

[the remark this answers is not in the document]
ramiro050: Yeah, I'm not really sure how we would represent such an op in `linalg`. I also don't know of anyone currently working on adding support for those ops.

PR #2321, opened 2023-07-18
Fixes Windows DLL crash

AyaanShah2204:
[LLVM bump](https://github.com/llvm/torch-mlir/commit/ebda6111000710470bad61ac9c467b57b2744d1a) brought in this [commit](https://reviews.llvm.org/D120368) that required inline extension for the func dialect to be declared explicitly.

This was causing the Windows dll to crash in SHARK in all snapshots after https://github.com/llvm/torch-mlir/releases/tag/snapshot-20230620.875

### DIFF
--- a/lib/InitAll.cpp
+++ b/lib/InitAll.cpp
@@ -9,6 +9,7 @@
 
 #include "torch-mlir/InitAll.h"
 
+#include <mlir/Dialect/Func/Extensions/InlinerExtension.h>
 #include "mlir/Dialect/Func/IR/FuncOps.h"
 #include "mlir/IR/Dialect.h"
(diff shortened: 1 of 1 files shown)


lib/InitAll.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "torch-mlir/InitAll.h"
 
+#include <mlir/Dialect/Func/Extensions/InlinerExtension.h>
```
stellaraccident, line 12: Please match the quote format of the other includes.

stellaraccident (approved):
Small formatting nut. Otherwise, lgtm.

lib/InitAll.cpp
AyaanShah2204: pushed a commit for the same

PR #2327, opened 2023-07-20
TorchToTosa: aten.embedding: Allow indices with any rank

mgehre-amd:
It's actually fine to not check the rank of the indices, because the conversion anyways flattens the index tensor to be (1, numElements) before applying tosa::gather, and then anyways reshapes the output tensor to the output shape of the aten.embedding.


eric-k256 (approved):
Looks good, and one more test on the passing list. Thanks.

PR #2328, opened 2023-07-21
Ignore constants in the legality error

mgehre-amd:
This improves the error message when lowering to tosa fails by pointing at the actual op that is missing support.
The constants are never a problem because
- they are pure, so they will fold away if unused
- they don't return tensor types, so they cannot be used directly by the ReturnOp
Thus there is always some other op using them, and that op is the interesting one.


eric-k256 (approved):
Looks good

PR #2336, opened 2023-07-24
fix cmake torch-mlir-capi linking and bazel build

qingyunqu:
fix `torch-mlir-capi-torch-test` and `bazel` after merge.


ashay (approved):
LGTM, thanks for fixing this!

PR #2342, opened 2023-07-25
[Torch Dialect] add support for adaptive_avgpool_1d

JianzheXiao:
as tittle , add support to decompose from adaptive_avgpool_1d to avgpool_1d


JianzheXiao:
@ramiro050  Could you please help me review this PR? Thank you so much!

lib/Conversion/TorchToStablehlo/Pooling.cpp
```diff
@@ -35,7 +35,7 @@ static Value createInitialValueForAtenPoolingOp(Operation *op, Type elementTy,
                                                 PatternRewriter &rewriter) {
   auto constType = RankedTensorType::get({}, elementTy);
   // Avg pooling
-  if (isa<AtenAvgPool1dOp, AtenAdaptiveAvgPool2dOp, AtenAvgPool2dOp, AtenCumsumOp>(op)) {
```
ramiro050, line 38: Is there a reason for adding it here? Given the decomposition, support for `AtenAvgPool1dOp` should be all that is needed

ramiro050 (approved):
Just a small comment, but other than that LGTM

lib/Conversion/TorchToStablehlo/Pooling.cpp
JianzheXiao: You are right, I just removed it from the file

qingyunqu:
@ashay Cloud you please help fix the ci failure (git checkout error)?

ashay:
I think @powderluv might have to reset the runners to work around the error.

PR #2344, opened 2023-07-26
[Torch Dialect] replace none-index in aten.Index.Tensor's  param by manually generating it

Vremold:
This patch is an extension for PR #2301    . Besides, this patch also supports when non-none indices are not adjacent in position. 
By the way, this patch also fixes a little bug in the shape infer function for aten.Index.Tensor.

Co-authored-by: Jiawei Wu <wujiawei.aml@bytedance.com>
Co-authored-by: Jianzhe Xiao <jianzhe.xiao@bytedance.com>


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4634,6 +4634,203 @@ class DecomposeAtenTypeAsOp : public OpRewritePattern<AtenTypeAsOp> {
...
+    auto inputSizes = inputType.getSizes();
+    int64_t inputRank = inputSizes.size();
+    auto outputType = op.getType().cast<BaseTensorType>();
+    if (!outputType.hasSizes()) {
+      rewriter.notifyMatchFailure(
+          op, "only output with shape information is supported");
+    }
+    auto outputRank = outputType.getSizes().size();
+
+    bool needCaninialize = false;
```
mgehre-amd, line 4698:
```suggestion
    bool needCanonicalization = false;
```

mgehre-amd:
Do you think that the code might become easier when the pattern just replaces one `None` at a time (and then the pattern gets applied multiple times)?

Vremold:
> Do you think that the code might become easier when the pattern just replaces one `None` at a time (and then the pattern gets applied multiple times)?

I don't think so. One reason is that code at the loop might be simplified, but the overall code structure has not changed much. Many additional checks will also be introduced due to the multiple pass running. Another is that this pattern also need to do something on the non-none index.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
Vremold: Done as you wish.

Vremold:
Could you review this PR ? @ramiro050

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4634,6 +4634,203 @@ class DecomposeAtenTypeAsOp : public OpRewritePattern<AtenTypeAsOp> {
...
+      for (int i = 0; i < inputRank; i++) {
+        dims.emplace_back(i);
+      }
+    }
+
+    // manually generate new indices.
+    SmallVector<Value> listElements(inputRank);
+    Value none = rewriter.create<Torch::ConstantNoneOp>(loc);
+    auto int64Dtype = getDtypeIntValueForType(
+        rewriter, loc, rewriter.getIntegerType(64, true));
```
ramiro050, line 4765: nit: add `/*var_name=*/` for `true`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4634,6 +4634,203 @@ class DecomposeAtenTypeAsOp : public OpRewritePattern<AtenTypeAsOp> {
...
+    for (i = inputRank - 1; i >= 0; --i) {
+      int64_t oldI = dims[i];
+      if (!indexUsed[oldI]) {
+        auto resultType = ValueTensorType::get(
+            context, {inputSizes[oldI]}, rewriter.getIntegerType(64, true));
+        auto dim = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(i));
+        auto end = rewriter.create<Torch::AtenSizeIntOp>(loc, newInput, dim);
+        auto v = rewriter.create<Torch::AtenArangeOp>(
+            loc, resultType, end, int64Dtype, none, none, none);
```
ramiro050, line 4779: nit: `/*var_name=*/` for `none`s

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4634,6 +4634,203 @@ class DecomposeAtenTypeAsOp : public OpRewritePattern<AtenTypeAsOp> {
...
+    Value none = rewriter.create<Torch::ConstantNoneOp>(loc);
+    auto int64Dtype = getDtypeIntValueForType(
+        rewriter, loc, rewriter.getIntegerType(64, true));
+
+    int64_t trailingDimCnt = 0;
+    int64_t i;
+    // handle trailing none index.
+    for (i = inputRank - 1; i >= 0; --i) {
+      int64_t oldI = dims[i];
+      if (!indexUsed[oldI]) {
```
ramiro050, line 4772: nit: break early to avoid indentation

```cpp
if (indexUsed[oldI])
  break;

// rest of code

File: lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp [hunk 5]
```diff
@@ -4634,6 +4634,203 @@ class DecomposeAtenTypeAsOp : public OpRewritePattern<AtenTypeAsOp> {
...
+        }
+        listElements[i] = *vInfo;
+        trailingDimCnt++;
+      } else
+        break;
+    }
+    // handle non-none index in between.
+    for (; i >= 0; --i) {
+      int64_t oldI = dims[i];
+      if (indexUsed[oldI]) {
```
ramiro050, line 4793: nit: break early

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4634,6 +4634,203 @@ class DecomposeAtenTypeAsOp : public OpRewritePattern<AtenTypeAsOp> {
...
+            loc, rewriter.getI64IntegerAttr(i));
+        auto end = rewriter.create<Torch::AtenSizeIntOp>(loc, newInput, dim);
+        auto v = rewriter.create<Torch::AtenArangeOp>(
+            loc, resultType, end, int64Dtype, none, none, none);
+        auto vInfo =
+            unsqueezeTensorAtTrailingDim(op, rewriter, v, outputRank - 1 - i);
+        if (failed(vInfo)) {
+          return rewriter.notifyMatchFailure(op, "failed to unsqueeze tensor");
+        }
+        listElements[i] = *vInfo;
```
ramiro050, line 4820: this code is very similar to the handling of trailing nones above. It should be refactored

ramiro050 (changes requested):
This is great! A few initial comments

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -2263,6 +2263,54 @@ def IndexTensorModule3dInput_basic(module, tu: TestUtils):
...
+
+class IndexTensorStaticNonContiguousWithNoneModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([2, 3, 4, 5, 32], torch.float32, True),
```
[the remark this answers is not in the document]
Vremold: Yes. Dynamic shape of input tensor The decomposition supports dynamic shape.

Vremold:
The requested changes are done.

Vremold:
Could you review the PR for me? @ramiro050

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4634,6 +4634,211 @@ class DecomposeAtenTypeAsOp : public OpRewritePattern<AtenTypeAsOp> {
...
+// Strictly speaking, this pattern is not a decomposition. The goal of this
+// pattern is to eliminate none index in aten.Index.Tensor's `indices` param for
+// the ease of various backend. The detailed steps are:
+//    1. reorder input tensor so that the non-none index appears at adjacent
+//    positions.
+//    2. manually generate index tensor with some ops like iota, to replace the
+//    none index in `indices`
+//    3. replace the old aten.Index.Tensor with a new
+//    aten.Index.Tensor_hacked_twin.
+class DecomposeAtenIndexTensorOp : public OpRewritePattern<AtenIndexTensorOp> {
```
ramiro050, line 4648: We should remove the decomposition for `HackedTwin` -> `aten.index.Tensor`, to avoid a cycle of decompositions. Also, `AtenIndexTensorOp` should be added to

https://github.com/llvm/torch-mlir/blob/574450687d8361bfa055b9cc09e0bef91541d4c3/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L364

to make it clear to backends that they only need to worry about the hacked twin variant

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
Vremold: Removing decomposition for `aten.Index.Tensor_hacked_twin` op is reasonable. Since the pass only works when the indices of `aten.Index.Tensor` contains none value. If we mark `aten.Index.Tensor` illegal explicitly, we will fail to lower some `aten.Index.Tensor` ops.

Vremold: A possible solution may be that we fill the `indices` argument using the methods in this PR until it has the same length as the input_tensor's rank. But it will cause many e2e tests fail, because all backends only handle aten.Index.Tensor for now.

ramiro050: > But if we mark aten.Index.Tensor illegal explicitly, we will fail to lower some aten.Index.Tensor ops, since the pass only works when the indices argument contains none value.

We should have the pattern also decompose into the hacked twin when there are no `None`s in the original `aten.Index.Tensor`. It would just be a renaming. We want to have at most one variant around at the end to make it easy for backends to know what needs handling.

> But it will cause many e2e tests fail, because all backends only handle aten.Index.Tensor for now.

Yeah, we need to update the backends to lower the hacked variant instead of `aten.Index.Tensor`. I think this would be straight forward. We just need to remove the logic that handles the `None` cases and replace every `AtenIndexTensorOp` with the hacked twin op

Vremold:
All done.

ramiro050 (approved):
This is awesome! Thanks 🙂

PR #2355, opened 2023-07-29
[Torch Dialect] emit aten.tile op and decompose it into aten.repeat

Vremold:
This PR is an extension of PR #2246


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -442,6 +442,16 @@ def aten〇repeat〡shape(self: List[int], repeats: List[int]) -> List[int]:
...
 
+def aten〇repeat〡shape(self: List[int], repeats: List[int]) -> List[int]:
+    return _repeat_like_shape_helper(self, repeats)
+
+def aten〇tile〡shape(self: List[int], dims: List[int]) -> List[int]:
+    dims_length = len(dims)
+    self_length = len(self)
+    if dims_length < self_length:
+        dims = [1] * (self_length - dims_length) + dims
+    return _repeat_like_shape_helper(self, dims)
```
ramiro050, line 453: nit: you can call shape functions inside another shape function:

```python
return aten〇repeat〡shape(self, dims)
```

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -442,6 +442,16 @@ def aten〇repeat〡shape(self: List[int], repeats: List[int]) -> List[int]:
         out.append(self[i] * repeats[i + leading_rank])
     return out
 
+def aten〇repeat〡shape(self: List[int], repeats: List[int]) -> List[int]:
+    return _repeat_like_shape_helper(self, repeats)
+
+def aten〇tile〡shape(self: List[int], dims: List[int]) -> List[int]:
```
ramiro050, line 448: nit: can you add a shape function test for this to make sure the edge case `dims_length < self` works correctly? Here's an example of a test: https://github.com/llvm/torch-mlir/blob/32b1c3dac0dc17ce0b5571c2811c827adf07879e/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py#L345

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -1462,6 +1462,27 @@ def RepeatModule_basic(module, tu: TestUtils):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([3, 1, 2], torch.float32, True),
+    ])
+    def forward(self, x):
+        return x.tile([3, 4])
```
ramiro050, line 1476: Can you also add a test for the `dims_size > input_rank`?

ramiro050 (changes requested):

Vremold:
The comments are addressed in the latest commit.

ramiro050 (approved):
Thanks!

PR #2358, opened 2023-07-31
[Torch Dialect] emit aten.masked_scatter and aten.masked_scatter_ op

Vremold:
In fact, I have a decomposition rule developed locally for aten.masked_scatter op. However it relies on the backend support for aten.masked_select op, i.e., it will pass no e2e tests. Will it be accepted if I propose it as a PR?


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -939,6 +939,12 @@ def aten〇narrow〇Tensor〡shape(self: List[int], dim: int, start: List[int],
 def aten〇slice_scatter〡shape(self: List[int], src: List[int], dim: int = 0, start: Optional[int] = None, end: Optional[int] = None, step: int = 1) -> List[int]:
     return self
 
+def aten〇masked_scatter〡shape(self: List[int], mask: List[int], source: List[int]) -> List[int]:
+    return self
+
+def aten〇masked_scatter_〡shape(self: List[int], mask: List[int], source: List[int]) -> List[int]:
```
ramiro050, line 945: no need to add underscore variant, since torch-mlir will replace it with the regular one

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1825,6 +1831,14 @@ def aten〇scatter〇value〡dtype(self_rank_dtype: Tuple[int, int], dim: int, i
     self_rank, self_dtype = self_rank_dtype
     return self_dtype
 
+def aten〇masked_scatter_〡dtype(self_rank_dtype: Tuple[int, int], mask_rank_dtype: Tuple[int, int], source_rank_dtype: Tuple[int, int]) -> int:
```
ramiro050, line 1834: no need to add underscore variant, since torch-mlir will replace it with the regular one

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1825,6 +1831,14 @@ def aten〇scatter〇value〡dtype(self_rank_dtype: Tuple[int, int], dim: int, i
...
+    self_rank, self_dtype = self_rank_dtype
+    return self_dtype
+
+def aten〇masked_scatter〡dtype(self_rank_dtype: Tuple[int, int], mask_rank_dtype: Tuple[int, int], source_rank_dtype: Tuple[int, int]) -> int:
```
ramiro050, line 1838: can you add a test for this similar to the other dtype functions?

ramiro050 (changes requested):
This code here (the ODS + shape/dtype functions) is fine, but the decomposition would require testing. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

ramiro050 (approved):

PR #2361, opened 2023-07-31
Update Torch ODS list with new ops

GlebKazantaev:
### Description
* Updated torch ODS gen list with new ops & regenerated GeneratedTorchOps.td
* Add some shape inference functions required by LTC
* Updated update_torch_ods.sh to use env PYTHONPATH


ramiro050:
The emit changes and the build changes look good to me. Is there someone familiar with the LTC changes that can review those?

GlebKazantaev:
@antoniojkim could you please take a look at the ltc part?

mabrowning (approved):
LTC Shape inference functions look good, although we do seem to have 2 different styles. I wonder if we could automate the `meta` device dispatch?

antoniojkim (approved):

antoniojkim:
> LTC Shape inference functions look good, although we do seem to have 2 different styles. I wonder if we could automate the `meta` device dispatch?

The ultimate goal is to be able to use structured kernels (i.e. basically the same thing as the meta device dispatch) for everything. But something that seems to have fallen under the radar as LTC is no longer under active development. I say its fine for now. If there is a bit need, we can do it ourselves

GlebKazantaev:
@ramiro050 it seems like "ubuntu-x86_64, in-tree, ON, stable" version doesn't raise any errors in case something failed. For example:
```
****** Unexpectedly Passed tests - 6 tests
    XPASS - "ElementwiseAtenFloorDivideBroadcastModule_basic"
    XPASS - "ElementwiseAtenFloorDivideModule_basic"
    XPASS - "HardtanhBackward_basic"
    XPASS - "RandLikeDtypeModule_basic"
    XPASS - "RandLikeModule_basic"
    XPASS - "UniformNoCorrelationModule_basic"
``` 
https://github.com/llvm/torch-mlir/actions/runs/5743311773/job/15567309988?pr=2361
but `Build and Test / build-test (ubuntu-x86_64, in-tree, ON, stable) (pull_request)` marked as green.

powderluv:
This seems to be breaking the CI. Lets revert and re-land.

PR #2363, opened 2023-08-01
Add support for aten.tensor op

vivekkhandelwal1:


vivekkhandelwal1:
Hi @ramiro050, the torch-mlir doesn't have support for the aten.tensor op. The issue with adding the shape and dtype info for the op is that it expects the list to be of any type (`t` in torch-mlir), and a shape/dtype func can't be declared with that type. So, I thought reducing this op into the corresponding variant is better. Please let me know if this approach is fine or if we can do something else to add support for this.

ramiro050:
My main concern with this approach is that it does not scale nicely to handle nested lists (you need a new op for every level of nesting for every dtype). I gotta think a bit about how we can improve the shape/dtype library to handle this case, but in the meantime you can just add dtype handling for `AtenTensorOp` similar to the handling of `NumToTensorScalarOp`: 

https://github.com/llvm/torch-mlir/blob/c0d8248ec7fc5d7791e5b7339aaacb1cde4b89d1/lib/Dialect/Torch/Transforms/SimplifyDtypeCalculations.cpp#L163

vivekkhandelwal1:
> My main concern with this approach is that it does not scale nicely to handle nested lists (you need a new op for every level of nesting for every dtype). I gotta think a bit about how we can improve the shape/dtype library to handle this case, but in the meantime you can just add dtype handling for `AtenTensorOp` similar to the handling of `NumToTensorScalarOp`:
> 
> https://github.com/llvm/torch-mlir/blob/c0d8248ec7fc5d7791e5b7339aaacb1cde4b89d1/lib/Dialect/Torch/Transforms/SimplifyDtypeCalculations.cpp#L163

But the shape info still can't be added, right?

ramiro050:
> But the shape info still can't be added, right?

You can also add shape handling in the shape function simplification pass:

https://github.com/llvm/torch-mlir/blob/c0d8248ec7fc5d7791e5b7339aaacb1cde4b89d1/lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp

vivekkhandelwal1:
Closing this for now, since it's not needed anymore.

PR #2370, opened 2023-08-02
Change Python version from 3.10 to 3.11 in installation instructions

ramiro050:


ashay (approved):
LGTM!

Related, we should also update the Python version in our Docker images (we currently use 3.10.4).  For instance, here's a line from the out-of-tree build:

```
Defaulting to /usr/bin/python3 (expected for CI builds)
:::: Python version Python 3.10.4
```

ramiro050:
Good catch! I'll update all the other 3.10 mentions in a follow up patch

PR #2373, opened 2023-08-03
Prevent failed stable CI job from cancelling nightly jobs

ramiro050:
The CI jobs that use stable PyTorch are currently not required to pass in order for a patch to get merged in `main`. This commit makes sure that if a CI job for stable PyTorch fails, it does not cancel the other required jobs.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -22,6 +22,7 @@ concurrency:
 # macos  - arm64  - llvm in-tree     - pytorch binary - build only    # cross compile, can't test arm64
 jobs:
   build-test:
+    continue-on-error: ${{ matrix.torch-version == 'stable' }}
     strategy:
       fail-fast: true
       matrix:


ashay:
I think you might need to put single quotes (not double quotes, apparently that's not allowed by GitHub's parser) around `stable`.  [Here](https://github.com/llvm/torch-mlir/actions/runs/5754437236/workflow) is GitHub's rudimentary validation of the script.

Related, should we mark the stable PyTorch builds as optional in the repo admin settings?  I imagine that intention of this PR is to allow merging even if the stable PyTorch build failed, but let me know if that's not the case.

ramiro050:
> Related, should we mark the stable PyTorch builds as optional in the repo admin settings? I imagine that intention of this PR is to allow merging even if the stable PyTorch build failed, but let me know if that's not the case.

I believe they are already optional. In the list of CI jobs, the only "Required" ones are nightly for MacOS and Ubuntu (in-tree, out-of-tree).

ashay:
> I believe they are already optional. In the list of CI jobs, the only "Required" ones are nightly for MacOS and Ubuntu (in-tree, out-of-tree).

I see.  Perfect, thanks for confirming!

ashay (approved):

glebk-cerebras (approved):

PR #2384, opened 2023-08-07
[Torch Dialect] Fix ui8 lower to uitofp

JianzheXiao:
The second pr for #2184
As MentionedThe change will lower aten.to.dtype(ui8) -> f32 to arith.uitofp, instead of arith.sitofp.


ramiro050 (approved):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -965,10 +965,14 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
     Value input = payloadArgs[0];
+    Type srcOriginalDtype =
+        converter->convertType(atenToDtype.getSelf().getType())
+            .cast<RankedTensorType>()
+            .getElementType();
     Type dtype = converter->convertType(atenToDtype.getType())
                      .cast<RankedTensorType>()
                      .getElementType();
-    Value result = convertScalarToDtype(b, loc, input, dtype);
+    Value result = convertScalarToDtype(b, loc, input, dtype, srcOriginalDtype);
```
qingyunqu, line 975: I think we can add an e2e test which has `unsigned` dtype.

PR #2388, opened 2023-08-09
[Torch Dialect] add folder for aten.any.bool

JianzheXiao:
Aviod error occured in runtime.assert in stablehlo


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1437,6 +1437,23 @@ OpFoldResult AtenBoolIntOp::fold(FoldAdaptor adaptor) {
...
 }
 
+//===----------------------------------------------------------------------===//
+// AtenAnyBoolOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenAnyBoolOp::fold(FoldAdaptor adaptor) {
+  SmallVector<bool> inputList;
+  if (!matchPattern(getSelf(), m_TorchListOfConstantBools(inputList)))
+    return nullptr;
```
qingyunqu, line 1447: I think we can implement a more aggressive folder in future. We can match a `torch.List.Construct` and if we find that any operand of `torch.List.Construct` is `torch.constant.bool true`, just fold it to `true`.

JianzheXiao: Fixed

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1437,6 +1437,22 @@ OpFoldResult AtenBoolIntOp::fold(FoldAdaptor adaptor) {
   return nullptr;
 }
 
+//===----------------------------------------------------------------------===//
+// AtenAnyBoolOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenAnyBoolOp::fold(FoldAdaptor adaptor) {
+  auto inputConsturct = getSelf().getDefiningOp<Torch::PrimListConstructOp>();
```
ramiro050, line 1445: Drive by comment: need to check that the defining op is indeed a `PrimListConstructOp` and that the list is not mutated. Similar to this: 

https://github.com/llvm/torch-mlir/blob/139e820e15fc441fe90f595ee927035304b5b606/lib/Dialect/Torch/IR/TorchOps.cpp#L1931-L1933

ramiro050, line 1445: nit: inputConst**r**u~r~ct

JianzheXiao:
@powderluv Hi! Could you please help fix the CI failure (git checkout error)?

PR #2391, opened 2023-08-10
Add handling of namespaces to library generator

ramiro050:
When using custom ops, sometimes PyTorch will insert namespaces to the abstract interpretation function name in the format: `__torch__.{namespace_1}.{namespace_2}...{op_name}`.  The extra namespaces are not part of the abstract interpretation function name, so it needs to be removed before generating the library of MLIR snippets of abstract interpretation functions. This commit adds support for removing the namespace information.


ramiro050:
@nithinsubbiah, this should fix your issue.

vivekkhandelwal1 (approved):

PR #2398, opened 2023-08-15
[TOSA] Fix conversion for depthwise convolutions

simon-camp:


test/Conversion/TorchToTosa/linear.mlir
```diff
@@ -0,0 +1,74 @@
+// RUN: torch-mlir-opt <%s -convert-torch-to-tosa -split-input-file -verify-diagnostics | FileCheck %s
```
eric-k256, line 1: Rather than adding lit tests, preference in the torch-mlir project is to use e2e testing. 
The file e2e_testing/xfail_sets.py has the TOSA_PASS_SET defined in it, which is the working e2e tests for the TOSA backend. 

Looking at the test suite, (python/torch_mlir_e2e_test/test_suite), the ConvolutionModule2DGroups_basic should test the new functionality you've added. You should only need to enable this test, and then can verify that it passes. Presumably it fails without your change.

Once the e2e test is working, this file is no longer needed.

eric-k256 (changes requested):
The code itself looks good. The only change needed is the e2e testing change.

test/Conversion/TorchToTosa/linear.mlir
simon-camp: I've added tests for depthwise and grouped convolutions with statically shaped weights and updated `xfail_sets.py` accordingly.

eric-k256 (approved):
Thanks for the test update, looks good to me.
The build failure appears unrelated to this change, I can see other builds hitting the same failure.

marbre:
@eric-k256 @silvasean Does one of you squash an merge or should I take over?

eric-k256:
Either way can work, but to save time, I squashed and merged.

PR #2402, opened 2023-08-17
Prototype passes for lowering quantized group matmul

jinchen62:


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -427,6 +427,166 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
...
 } // namespace
 
+namespace {
+class ConvertCustomQuantizedMatmulOp : public OpConversionPattern<OperatorOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(OperatorOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (op.getName().str() != "quant.matmul_rhs_group_quant") {
```
ramiro050, line 437: I don't think torch-mlir is the right place for patterns like this one (matching on `OperatorOp`). A few of the reasons being:

1. it cannot be e2e tested
2. It reserves a name for a custom op preventing other users of torch-mlir from using their own custom version of `quant.matmul_rhs_group_quant`
3. It would open the gates for people to add their own custom op patterns, increasing the complexity of the codebase in unwanted ways

stellaraccident: I think this is exactly what torch-mlir is for: The PyTorch ecosystem has a large variety of custom ops and trying to stick to some "pure" subset is not pragmatic or aligned with how PyTorch is actually used.

With that said, I think that this patch needs a bit of work:

1. The right way to test these kinds of conversions is with lit, not e2e tests. There needs to be lit tests to verify.
2. I think that we need either: a) opt-in conversions for these kinds of patterns that preserves optionality or b) not sitting on generic names.

This kind of codebase works best (and is the most aligned with PyTorch) if new things emerge at the edges of the codebase and migrate "core-ward" as things go. For this, I would prefer that if this op is showing up in some PyTorch downstream (I think this is coming from brevitas), then it would be better if it was named that as that wouldn't stop other people from doing anything else. Then I would have a separate CPP file with these patterns like `BrevitasCustomConversions.cpp` to keep it all isolated.

So three action items from me:

1. Lit tests please for these custom conversions.
2. Namespace things for the projects that are driving it so that others can have similar things without stomping on common names prematurely.
3. Segregate custom op lowerings into their own CPP file (i.e. `BrevitasCustomConversions.cpp`) to keep things tidy as we see where things go.

My approach to flood-gates is to see if we actually have a problem vs blocking progress. In this case, segregating into its own CPP file isolates things, and if 3 months from now, we find we have 10 of these things, then we can organize it a bit better/add opt-in/etc.

ramiro050: > I think this is exactly what torch-mlir is for: The PyTorch ecosystem has a large variety of custom ops and trying to stick to some "pure" subset is not pragmatic or aligned with how PyTorch is actually used.

My concern/opposition was regarding patterns on `OperatorOp`. I've done a lot of work on developing the custom op story for torch-mlir, so I am in full agreement with you that having support for custom ops is of high importance. Currently users can create custom ops and have them reach the backend contract. How the custom ops get from the backend contract to a backend is still to be designed, and is the kind of feature that will have big impact on users of torch-mlir, so I think we need to come up with a good design.

> The right way to test these kinds of conversions is with lit, not e2e tests. There needs to be lit tests to verify.

This is a big change from the way things have operated in torch-mlir (see our [testing guidelines](https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing)). I think it is important to set contribution expectations clearly in order to be fair to all contributors. I'm not opposed to having guidelines evolve to adapt to the new user requirements, but this is the type of change that we should get the community involved in.

> This kind of codebase works best (and is the most aligned with PyTorch) if new things emerge at the edges of the codebase and migrate "core-ward" as things go. 

While I agree with the sentiment, I think we need to set expectations clearly with community involvement. My understanding of the plan outlined here is that we will have essentially a Brevitas dialect in torch-mlir. There have been folks working on another custom dialect in torch-mlir (TCP), and attempts to merge into main have been blocked because a full e2e flow supporting a model like Resnet is still missing. While I understand that the dialects, TCP and Brevitas, are slightly different, they are still a form of custom dialect that is being added to torch-mlir, and I think they should be held to similar standards.

> My approach to flood-gates is to see if we actually have a problem vs blocking progress. In this case, segregating into its own CPP file isolates things, and if 3 months from now, we find we have 10 of these things, then we can organize it a bit better/add opt-in/etc.

My approach, at least in torch-mlir, is to try to not block progress, but at the same time think of long term implications. All of this to mean, I think this is the type of addition that would benefit from an RFC :) 

Lastly, I do want to emphasize that I care about the custom op functionality in torch-mlir. While my involvement in torch-mlir is a bit limited at the moment, I am happy to help as much as possible to develop a good solution for users.

silvasean: I think if we are merging this into main it would make sense to add this to the torch dialect just as a regular op, rather than using OperatorOp. This would also include migrating the shape and dtype functions for this op into main along with this conversion pattern, otherwise it leaves lots of details in unspecified downstream flows, which fragments the community and prevents users from remixing each other's contributions.

The point of OperatorOp is for ops that are not recognized by torch-mlir. It's not about choosing a subset, but more about using the right tool (OperatorOp vs recognized op) depending on the level of "core-ness". The namespacing at the Torch operator registry level should be enough to disambiguate such cases and avoid conflicts.



I am also wary of not e2e testing this -- how difficult would it be to somehow enable that? What's the specific technical blocker? My experience is that lit testing for this kind of change basically becomes change-detector tests that people blindly update, which moves meaningful test failures into downstream whole-hog integration tests, which is contrary to the principle of catching failures as early as possible.

Having "dangling" e2e-untested functionality in our main branch is hard to back out from. I think if you want to steer the project more in that direction @stellaraccident that's okay, but we should acknowledge it is a shift in project strategy and maybe modify some documentation accordingly (e.g. here: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#testing). Up until now I have been trying really hard to avoid "dangling" functionality, but I can see value in a strategy that tries to manage it -- could have velocity advantages, among other things.

lib/Dialect/TorchConversion/Transforms/ConvertCustomQuantOp.cpp
```diff
@@ -0,0 +1,225 @@
...
+    Type newResultType = getTypeConverter()->convertType(op.getType(0));
+    auto resultType = newResultType.cast<RankedTensorType>();
+    if (!resultType) {
+      return failure();
+    }
+    auto resultShape = resultType.getShape();
+    Type elementType = resultType.getElementType();
+
+    // expand lhs
+    std::vector<int64_t> lhs_expandedShape = {lhsShape[0], lhsShape[1],
```
stellaraccident, line 103: Nit: `lhsExpandedShape`

lib/Dialect/TorchConversion/Transforms/ConvertCustomQuantOp.cpp
```diff
@@ -0,0 +1,225 @@
...
+    }
+    auto lhsShape = lhsType.getShape();
+    int lhs_reduct_dim_size = lhsShape.back();
+
+    Value q_rhs = adaptor.getOperands()[1];
+    auto rhsType = q_rhs.getType().cast<RankedTensorType>();
+    if (!rhsType) {
+      return failure();
+    }
+    auto rhsShape = rhsType.getShape();
```
stellaraccident, line 57: Nit: These should all be `camelCase` per LLVM style.

stellaraccident: (here and throughout, please)

lib/Dialect/TorchConversion/Transforms/UnpackQuantTensor.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+    OpOperand *use = constOp.getResult().use_begin().getOperand();
+    auto op = dyn_cast<OperatorOp>(use->getOwner());
+    if (!op)
+      return failure();
+
+    if (use->getOperandNumber() != 1)
+      return failure();
+
+    if (op.getName().str() != "quant.matmul_rhs_group_quant") {
```
stellaraccident, line 41: We should really invert this pattern: It should be anchoring on the `quant.matmul_rhs_group_quant` and then getting the producer for the weights and applying the unpacking.

lib/Dialect/TorchConversion/Transforms/UnpackQuantTensor.cpp
```diff
@@ -0,0 +1,141 @@
...
+    ValueTensorType newRhsType = ValueTensorType::get(
+        rewriter.getContext(), tensorShape, unpackedElementType);
+
+    auto elements = constOp.getValueAttr().dyn_cast<DenseIntElementsAttr>();
+    if (!elements)
+      return failure();
+
+    auto attrType = RankedTensorType::get(tensorShape, unpackedElementType);
+
+    // This is terrible but idk what else to do.
```
stellaraccident, line 96: This is going to be quite bad for performance and memory usage. I think that ultimately, we don't want this pass doing this kind of constant twiddling at all, and we want to rely on constant evaluation in backends that support it.

Ideally, what we should be doing is materializing IR that does the conversion from a quantized type to a pure integer type. That will look substantially different from what you have here.

test/Dialect/TorchConversion/unpack-quant-tensor.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: torch-mlir-opt %s -torch-unpack-quant-tensor | FileCheck %s
+
+// CHECK-LABEL: func @forward
```
stellaraccident, line 3: I do need to ask you to actually verify the converted constants, given that the main thing the pass is doing is bit twiddling.

stellaraccident: Since you also changed the type legality code to include i4, I would like to see at least a variant of that.

test/Dialect/TorchConversion/convert-custom-quant-op.mlir
```diff
@@ -0,0 +1,12 @@
+// RUN: torch-mlir-opt %s -torch-convert-custom-quant-op -verify-diagnostics | FileCheck %s
+
+// CHECK-LABEL: func @forward
```
stellaraccident, line 3: This needs to check the conversion result, not just that it converts.

include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
```diff
@@ -57,6 +57,9 @@ std::unique_ptr<OperationPass<ModuleOp>> createFuncBackendTypeConversionPass();
 std::unique_ptr<OperationPass<func::FuncOp>>
 createFinalizingBackendTypeConversionPass();
 
+std::unique_ptr<OperationPass<func::FuncOp>> createUnpackQuantTensorPass();
```
stellaraccident, line 60: Can we please add a comment:

```
// These passes do a one-off conversion of a specific kind of quantized group matmul as a prototype.
// Generalized quantized operation handling will likely obviate them but that are being carried for now in order
// to unblock progress on full integrations. See <<insert new issue here>> for the plan to support a more
// generalized lowering for these graphs.
```

stellaraccident (changes requested):
Hi - there are a number of issues I've noted with this patch as written that we need to address.

In addition, it is pretty clear to me that this is a stopgap on the way to more robust handling of these kinds of quantized graphs. I've suggested verbiage around the passes to indicate as such, and we should have an issue tracking it.

In addition, I would recommend renaming the patch to "Prototype passes for lowering quantized group matmuls."

include/torch-mlir/Dialect/TorchConversion/Transforms/Passes.h
stellaraccident: Can you also include verbiage like this in the Passes.td file for the passes and add that they should not be included in default lowering flows until further along.

lib/Conversion/TorchToLinalg/Linear.cpp
stellaraccident: > Having "dangling" e2e-untested functionality in our main branch is hard to back out from. I think if you want to steer the project more in that direction @stellaraccident that's okay, but we should acknowledge it is a shift in project strategy and maybe modify some documentation accordingly (e.g. here: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#testing). Up until now I have been trying really hard to avoid "dangling" functionality, but I can see value in a strategy that tries to manage it -- could have velocity advantages, among other things.

Thanks, Sean -- I do appreciate the uphill battle of keeping things tidy. Looking at the entire design of these quantization schemes, I think there is a lot of work to do here to make a robust pipeline for lowering these *kinds* of things, and I've asked that https://github.com/llvm/torch-mlir/issues/2417 be opened to elaborate and track this. It is actually quite helpful for me (and I assume others) to see the, let's call it "direct", way to cut this into the system in order to get some early momentum.

I think that there is a fairly big difference between having dangling features enabled in the default flows and having some passes available that let us paper over a fairly wide area to get started by a subset of the community who opts in. The former needs to be coherent and rationalized while the latter really is "it does what the lit test says". It might just be me, but when I see something like this that is breaking ground on an entire area of work, I like to see it isolated vs baked in. For that, having some "on the side" passes/patterns for getting started, as long as properly called out as such helps unlock the full pipeline. As an example, having had the programs that this work enables available has enabled multiple months of work across three different companies to get a backend in shape to be able to even do this, and it has caused the refinement of a lot of thought about *how to do it*.

> My approach, at least in torch-mlir, is to try to not block progress, but at the same time think of long term implications. All of this to mean, I think this is the type of addition that would benefit from an RFC :)
> Lastly, I do want to emphasize that I care about the custom op functionality in torch-mlir. While my involvement in torch-mlir is a bit limited at the moment, I am happy to help as much as possible to develop a good solution for users.

Thanks Ramiro - we're all trying to do the right thing here. For me, I've watched this slice of work trying to get a foothold for several months, and it is a real chicken and egg situation: we don't know exactly what we need until we get some backend capabilities straightened out (when this started, sub-byte types didn't work *at all* properly in *any* backend and it tooks months to straighten that). Then we come back and straighten more of the frontend out. We've got to be able to support that cycle pragmatically and without disrupting the default flows. That is just LLVM dev policy kind of stuff. It isn't clear to me yet that we will end up with a "Brevitas" dialect or rather this is just the starting point. I can almost see a way to generalize all of this, but not there yet. Let's get a couple more laps on the feature and see where it goes.

stellaraccident (approved):
This looks reasonable to me at this point for what it is, but let's let the open comment thread get responses before landing.

lib/Conversion/TorchToLinalg/Linear.cpp
powderluv: As discussed in the community meeting we will land this as a stepping stone towards better support for these kind of quantized matmuls etc. Opened https://github.com/llvm/torch-mlir/issues/2417 for tracking the overall effort.

PR #2408, opened 2023-08-21
[Torch Dialect] revert canonicalizer for PrimListConstructOp

Vremold:
This a reverter for PR #2306 . PrimListConstructOp has so many complex usage scenarios. Canonializing it might not be a good idea. 
Here is an example. Suppose we have a IR like:
```mlir
%499 = torch.aten.chunk %498, %int4, %int0 : !torch.tensor, !torch.int, !torch.int -> !torch.list<tensor>
%500:4 = torch.prim.ListUnpack %499 : !torch.list<tensor> -> !torch.tensor, !torch.tensor, !torch.tensor, !torch.tensor
%501 = torch.prim.ListConstruct %500#0, %500#1, %500#2, %500#3 : (!torch.tensor, !torch.tensor, !torch.tensor, !torch.tensor) -> !torch.list<tensor>
%502 = torch.aten.cat %501, %int-1 : !torch.list<tensor>, !torch.int -> !torch.tensor
```
If we canonicalize the ListConstructOp, the `aten.cat` op will directly use the result from `aten.chunk`. It might fail some passes, since we often assume `aten.cat`'s operand comes from a ListConstructOp.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
@@ -413,7 +413,6 @@ def Torch_PrimListConstructOp: Torch_Op<"prim.ListConstruct", [
   );
 
   let hasVerifier = 1;
-  let hasCanonicalizer = 1;
 
   let assemblyFormat = [{
     $elements attr-dict `:` functional-type(operands, results)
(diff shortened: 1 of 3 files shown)


ramiro050 (approved):
Good point. Thanks for reverting!

qingyunqu (approved):

PR #2411, opened 2023-08-21
Add missing early return.

stellaraccident:
Found with more strict build flags.


ramiro050 (approved):

ramiro050:
Just realized that this was addressed in a PR made this morning: https://github.com/llvm/torch-mlir/pull/2409

PR #2413, opened 2023-08-22
[Torch Dialect][stablehlo] emit aten.rand op and add converter to stablehlo

Vremold:


ramiro050 (approved):
Thanks! I think you need to add the new test to the `linalg` xfail set

PR #2416, opened 2023-08-23
Add E2E support for aten.elu

123epsilon:
Implements a decomposition for `aten.elu` and passes e2e tests for all backends. 

Newbie to torch-mlir so I would love some feedback on this PR, in particular for the dtype propagation function. When implementing this function the dtype propagation stops in its tracks if I allow the `dtypes` array to have more than two elements - i.e. `dtypes = [self_dtype, alpha_dtype, scale_dtype, input_scale_dtype]` fails but `dtypes = [self_dtype, input_scale_dtype]` succeeds - from what I can tell the dtype propagation fails even at the earliest passes in the former case despite the proper dtype function being successfully inlined in the IR, I'm not sure why. I'm not sure if this is by design, a bug, or (more likely) something that I'm doing wrong. Intuitively I think the dtype function for this op should take into account all parameter dtypes when doing the final type promotion so I would appreciate it if anyone has any feedback on the additions to `abstract_interp_lib_gen.py` below.

cc @stellaraccident @vivekkhandelwal1 @ramiro050


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -2710,6 +2713,23 @@ def aten〇leaky_relu〡dtype(self_rank_dtype: Tuple[int, int], negative_slope:
...
+    alpha_dtype = get_dtype_of_scalar(alpha)
+    scale_dtype = get_dtype_of_scalar(scale)
+    input_scale_dtype = get_dtype_of_scalar(input_scale)
+
+    if any([is_float_dtype(d) for d in [alpha_dtype, scale_dtype, input_scale_dtype]]):
+        assert not is_integer_dtype(self_dtype)
+
+    dtypes = [self_dtype, input_scale_dtype]
+    return promote_dtypes(ranks, dtypes)
+
```
vivekkhandelwal1, line 2732: Hi @123epsilon, I think the dtype computation for this op does not involve the scalar types. I tried different variants for this op, and it turned out that:
1.) The input tensor always has to have floating-point values(f16, f32, f64).
2.) Irrespective of the scalar types, the result type is the same as the input dtype.

123epsilon: Ok, I've amended the dtype function - in the dtype tests however we still require checking whether the tensor is an int tensor in the case that the params are floating point - otherwise we do not match the behavior from [here](https://github.com/pytorch/pytorch/blob/256fed02e930210dbcd7e5e23fcf142362098c2a/torch/_refs/nn/functional/__init__.py#L231C34-L231C34) and the dtype tests fail, similarly with asserting that the tensor is not of bool type.

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -2710,6 +2713,17 @@ def aten〇leaky_relu〡dtype(self_rank_dtype: Tuple[int, int], negative_slope:
...
 
+@check_dtype_function(
+    _check_tensors_with_the_same_dtype(num_of_tensors=1, error_types={torch.bool}, alpha=1, scale=1, input_scale=2) +
+    _check_tensors_with_the_same_dtype(num_of_tensors=1, error_types={torch.bool, torch.int8, torch.uint8, torch.int16, torch.int32, torch.int64}, alpha=1.0, scale=1.0, input_scale=2.0))
+def aten〇elu〡dtype(self_rank_dtype: Tuple[int, int], alpha: Union[int, float, complex] = 1, scale: Union[int, float, complex] = 1, input_scale: Union[int, float, complex] = 1) -> int:
+    self_rank, self_dtype = self_rank_dtype
+    assert self_dtype != torch.bool
+    param_dtypes = [get_dtype_of_scalar(p) for p in [alpha, scale, input_scale]]
+    if any([is_float_dtype(d) for d in param_dtypes]):
+        assert not is_integer_dtype(self_dtype)
```
vivekkhandelwal1, line 2724: I think this is not required.

python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -473,6 +473,26 @@ def forward(self, x):
 def ElementwiseLeakyReluStaticModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(4, 5, 6, low=-1))
 
+# ==============================================================================
+
+
+class ElementwiseEluModule(torch.nn.Module):
```
vivekkhandelwal1, line 479: Also, add a test with non-default values of alpha, scale, and input_scale.

vivekkhandelwal1 (changes requested):
@123epsilon, you also have to add this op here: https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L490

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
123epsilon: Without this logic we don't pass the `check_dtype` tests due to the fact that the op fails in certain cases where dtypes [can not be properly promoted](https://github.com/pytorch/pytorch/blob/256fed02e930210dbcd7e5e23fcf142362098c2a/torch/_refs/nn/functional/__init__.py#L231C34-L231C34) from a weaker type to a stronger type. Do you still want me to remove this logic? I also notice that similar logic is present in the `leaky_relu` dtype function, also mirroring the corresponding pytorch logic [here](https://github.com/pytorch/pytorch/blob/256fed02e930210dbcd7e5e23fcf142362098c2a/torch/_refs/nn/functional/__init__.py#L335).

python/torch_mlir_e2e_test/test_suite/elementwise.py
123epsilon: Done

123epsilon:
> @123epsilon, you also have to add this op here: https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L490

Just pushed this change

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
vivekkhandelwal1: Thanks for the context!

vivekkhandelwal1 (approved):
LGTM

vivekkhandelwal1:
I'm extremely sorry for not doing a squash and merge.

PR #2418, opened 2023-08-25

vivekkhandelwal1:
Set PyTorch and TorchVision version to nightly release 2023-08-24.


powderluv:
this may need another update

vivekkhandelwal1:
> this may need another update

The CI is failing for the stable torch build.

CC: @ashay @powderluv

ashay:
Is it right that the two tests were passing in nightly builds until recently and that they always passed in stable builds?  If so, I wonder if we can raise an issue with upstream PyTorch to address the problem in nightly builds (and hold off on updating PyTorch in the meantime)?

If we need the PyTorch update soon, we could go back to [Ramiro's change](https://github.com/llvm/torch-mlir/pull/2373/files) or even disable TorchDynamo tests in the stable builds, but I worry that we'll have broken stable builds.

powderluv:
@maxbartel / @mgehre-amd how do you guys want to resove this ? We don't want to hold up nightly rolls if stable breaks.

vivekkhandelwal1:
> Is it right that the two tests were passing in nightly builds until recently and that they always passed in stable builds? If so, I wonder if we can raise an issue with upstream PyTorch to address the problem in nightly builds (and hold off on updating PyTorch in the meantime)?
> 
> If we need the PyTorch update soon, we could go back to [Ramiro's change](https://github.com/llvm/torch-mlir/pull/2373/files) or even disable TorchDynamo tests in the stable builds, but I worry that we'll have broken stable builds.

Earlier, those two tests were passing for the nightly build about a week ago. After which they started failing due to upstream PyTorch changes. Right now, if I remove them from the xfail set then those two tests would fail for nightly while pass for the stable build.

maxbartel:
I guess such cases need to be special handled like https://github.com/llvm/torch-mlir/blob/5282324c68aceff2d186fb16043b2ec76beb5c9d/python/torch_mlir_e2e_test/test_suite/__init__.py#L19-L24. Those are the same tests, but for the linalg lowering test.

mgehre-amd:
What do you think about removing `torchdynamo` tests from the stable pipeline? dynamo still changes a lot in pytorch, and it's not our focus right now in the way we use stable pytorch.

powderluv:
Folks - if we need to enable / disable tests (or other fixes) on Stable builds please do jump in and help. We have added Stable build testing and want to support it but that is not automatic and requires on-going maintenance to not hold up the main branch Pytorch rolls.

mgehre-amd:
> Folks - if we need to enable / disable tests (or other fixes) on Stable builds please do jump in and help. We have added Stable build testing and want to support it but that is not automatic and requires on-going maintenance to not hold up the main branch Pytorch rolls.

I opened https://github.com/llvm/torch-mlir/pull/2426

Abhishek-Varma (approved):

PR #2419, opened 2023-08-25
[MLIR][TORCH] Add E2E support for view_as_real op

brucekimrokcmu:
Implemented torch.view_as_real op and passes e2e tests for linalg, torchdynamo backends. (xfailed for tosa and stablehlo)
This PR revised how complex dtype is supported, and thus affect `simplify-dtype-calculations.mlir` test in https://github.com/llvm/torch-mlir/pull/1673. Essentially, `unk` should be replaced with `complex<f32>` in the test.

Since PyTorch doesn't seem to support complex32, the dtype is not considered. [[here](https://pytorch.org/docs/stable/tensor_attributes.html)].

cc @stellaraccident @vivekkhandelwal1  @ramiro050


lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1225,7 +1227,8 @@ class ConvertAtenContiguousOp : public OpConversionPattern<AtenContiguousOp> {
       return failure();
 
     Type resultType = getTypeConverter()->convertType(op.getType());
-    rewriter.replaceOpWithNewOp<tensor::CastOp>(op, resultType, adaptor.getSelf());
+    rewriter.replaceOpWithNewOp<tensor::CastOp>(op, resultType,
+                                                adaptor.getSelf());
```
qedawkins, line 1231: Looks like there are a few of these formatting changes. Did you add them yourself or did the linter do this? Should be fine, just wondering where they are coming from.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1427,6 +1430,94 @@ class ConvertAtenViewAsComplexOp
...
+        typeConverter->convertType(op.getType()).cast<RankedTensorType>();
+
+    RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+    auto inputElementType = getElementTypeOrSelf(input.getType());
+    Type elementType;
+    if (inputElementType.isa<ComplexType>()) {
+      elementType = resultType.getElementType();
+    } else {
+      return op.emitError("only ComplexType is allowed as input type");
+    }
```
qedawkins, line 1460: nit: just return if the input is non-complex and get the result element type afterwards
```suggestion
    if (!inputElementType.isa<ComplexType>()) {
      return op.emitError("only ComplexType is allowed as input type");
    }
    Type elementType = resultType.getElementType();
```

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1427,6 +1430,94 @@ class ConvertAtenViewAsComplexOp
...
+    const TypeConverter *typeConverter = getTypeConverter();
+    MLIRContext *context = rewriter.getContext();
+
+    auto input = adaptor.getSelf();
+
+    RankedTensorType resultType =
+        typeConverter->convertType(op.getType()).cast<RankedTensorType>();
+
+    RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+    auto inputElementType = getElementTypeOrSelf(input.getType());
```
qedawkins, line 1454: nit: Since you are casting to RankedTensorType you can just get the element type directly.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1427,6 +1430,94 @@ class ConvertAtenViewAsComplexOp
...
+    for (unsigned i = 0; i < resultType.getRank(); i++) {
+      outputExpr.push_back(getAffineDimExpr(i, context));
+    }
+    AffineMap outputMap =
+        AffineMap::get(resultType.getRank(), 0, outputExpr, op->getContext());
+
+    SmallVector<AffineExpr> inputExpr;
+    for (unsigned i = 0; i < resultType.getRank() - 1; i++) {
+      inputExpr.push_back(getAffineDimExpr(i, context));
+    }
```
qedawkins, line 1479: You can probably reuse the expressions set up in for the output rather than generating these affine expressions twice. Or setup the input map first and then append the final dim to the same list for the output.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1427,6 +1430,94 @@ class ConvertAtenViewAsComplexOp
...
+    Type elementType;
+    if (inputElementType.isa<ComplexType>()) {
+      elementType = resultType.getElementType();
+    } else {
+      return op.emitError("only ComplexType is allowed as input type");
+    }
+
+    SmallVector<OpFoldResult> resultShape =
+        tensor::getMixedSizes(rewriter, loc, input);
+    resultShape.push_back(
```
qedawkins, line 1464: nit: I would add a brief comment here to explain why we're appending a dim of size `2` here. Helps to understand the semantics of the op when just viewing the code.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1427,6 +1430,94 @@ class ConvertAtenViewAsComplexOp
...
+    auto realVar =
+        rewriter
+            .create<linalg::GenericOp>(
+                loc, outTensor.getType(), input, outTensor, indexingMaps,
+                iteratorTypes,
+                [&](OpBuilder &b, Location loc, ValueRange args) {
+                  SmallVector<Value> indices;
+                  for (int i = 0; i < inputType.getRank(); i++) {
+                    indices.push_back(b.create<linalg::IndexOp>(loc, i));
+                  }
```
qedawkins, line 1498: Indices does not appear to be used anywhere, only the last index.

python/torch_mlir_e2e_test/test_suite/reshape_like.py
```diff
@@ -8,6 +8,7 @@
 from torch_mlir_e2e_test.registry import register_test_case
 from torch_mlir_e2e_test.annotations import annotate_args, export
 
+
```
qedawkins, line 11: There are a lot of whitespace changes in here. Wondering the same thing as the earlier formatting comment.

qedawkins (changes requested):
Looks mostly good to me, just some nits and a question about all of the unrelated formatting changes.

python/torch_mlir_e2e_test/test_suite/reshape_like.py
brucekimrokcmu: I probably manually put test here initially and moved to basic.py

qedawkins (approved):
Overall this looks good to me, thanks for fixing the handling of complex types here! I have not worked on this project in a while so it might be worth getting a quick scan from someone who has.

python/torch_mlir/_dynamo_fx_importer.py
```diff
@@ -146,7 +146,11 @@ def _convert_dtype_to_mlir_type(dtype: torch.dtype) -> str:
         return "!torch.qint8"
     if dtype == torch.quint8:
         return "!torch.quint8"
+    # if dtype == torch.complex32:
+    #     return "complex<f32>"
```
vivekkhandelwal1, line 150: Please remove this

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -924,6 +924,26 @@ def aten〇view_as_complex〡dtype(self_rank_dtype: Tuple[int, int]) -> int:
...
+    return out
+
+def aten〇view_as_real〡dtype(self_rank_dtype: Tuple[int, int]) -> int:
+    self_rank, self_dtype = self_rank_dtype
+    if self_dtype == torch.complex64:
+        return torch.float
+    elif self_dtype == torch.complex128:
+        return torch.double
+    else:
+        assert False, "Unsupported dtype"
```
vivekkhandelwal1, line 942: Don't we need to add complex32 as well?

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -924,6 +924,26 @@ def aten〇view_as_complex〡dtype(self_rank_dtype: Tuple[int, int]) -> int:
     else:
         assert False, "Unsupported dtype"
 
+def aten〇view_as_real〡shape(self: List[int]) -> List[int]:
+    out: List[int] = []
+    n = len(self)
+    for i in range(n):
+        out.append(self[i])
+    out.append(2)
```
vivekkhandelwal1, line 932: Instead, do 
```
return self + [2]
```

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -924,6 +924,26 @@ def aten〇view_as_complex〡dtype(self_rank_dtype: Tuple[int, int]) -> int:
...
+    if self_dtype == torch.complex64:
+        return torch.float
+    elif self_dtype == torch.complex128:
+        return torch.double
+    else:
+        assert False, "Unsupported dtype"
+
+def aten〇view_as_real〡has_value_semantics() -> None:
+    return
+
```
vivekkhandelwal1, line 946: Why do we need this?

python/torch_mlir_e2e_test/framework.py
```diff
@@ -181,6 +181,7 @@ def __init__(self):
         torch.manual_seed(0)
 
     # TODO: Add zeros/ones/etc. as convenient.
+
```
vivekkhandelwal1, line 184: delete this

vivekkhandelwal1 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
qedawkins: Pytorch does not appear to support complex32 as noted here: https://pytorch.org/docs/stable/tensor_attributes.html

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
brucekimrokcmu: deleted it. Works without it

vivekkhandelwal1 (approved):
LGTM

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -924,6 +924,18 @@ def aten〇view_as_complex〡dtype(self_rank_dtype: Tuple[int, int]) -> int:
     else:
         assert False, "Unsupported dtype"
 
+def aten〇view_as_real〡shape(self: List[int]) -> List[int]:
+    return self + [2]
+def aten〇view_as_real〡dtype(self_rank_dtype: Tuple[int, int]) -> int:
```
ramiro050, line 929: This needs testing similar to other dtype functions:
https://github.com/llvm/torch-mlir/blob/4636abbad5cb9fee653c6768a9298f3e0b57638e/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py#L1343

lib/Dialect/Torch/IR/TorchTypes.cpp
```diff
@@ -404,20 +404,8 @@ static Type convertDtypeToBuiltinElementType(MLIRContext *context, Type dtype) {
...
-    // imag value is of float type then the resulting complex value is of
-    // complex<float> type.
-    auto floatType = complexType.getElementType().dyn_cast<mlir::FloatType>();
-    if (floatType.getWidth() == 32)
-      return ComplexType::get(mlir::FloatType::getF16(context));
-    else if (floatType.getWidth() == 64)
-      return ComplexType::get(mlir::FloatType::getF32(context));
-    else if (floatType.getWidth() == 128)
-      return ComplexType::get(mlir::FloatType::getF64(context));
+  } else if (dtype.isa<mlir::ComplexType>()){
```
ramiro050, line 407: nit: formatting is off. Make sure to run `git clang-format` on your changes

python/torch_mlir/_dynamo_fx_importer.py
```diff
@@ -146,7 +146,11 @@ def _convert_dtype_to_mlir_type(dtype: torch.dtype) -> str:
         return "!torch.qint8"
     if dtype == torch.quint8:
         return "!torch.quint8"
+    # if dtype == torch.complex32:
```
ramiro050, line 149: nit: remove commented out code

ramiro050 (changes requested):
Thanks! Can you also update

https://github.com/llvm/torch-mlir/blob/34a0897e1b61a19c48282192f2437dd24507660e/include/torch-mlir/Dialect/Torch/IR/TorchTypes.td#L131

We can do something like `| torch.complex{width}| complex<{width/2}>|`

PR #2422, opened 2023-08-28
[MLIR][TORCH] add E2E support for aten.min op

123epsilon:
Adds a definition for `aten.min` as a reduction operation, passes all backends except for `tosa`.


python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -572,6 +572,77 @@ def ReduceAmaxKeepDim_basic(module, tu: TestUtils):
...
+    @annotate_args([
+        None,
+        ([-1, -1, -1], torch.float32, True),
+    ])
+    def forward(self, a):
+        return torch.ops.aten.min(a)
+@register_test_case(module_factory=lambda: ReduceMinFloatModule())
+def ReduceMinFloatModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5))
+
```
vivekkhandelwal1, line 608: I don't see any difference in both these tests, except the former one has limits for the random values. I think we need only one of these.

vivekkhandelwal1 (changes requested):
Rest LGTM, but I would like someone from stablehlo side to review this as well. Probably @Vremold @tanyokwok

python/torch_mlir_e2e_test/test_suite/reduction.py
123epsilon: You're correct, I copied this set of tests from the corresponding tests for `AtenMaxOp`, but I'll remove one.

vivekkhandelwal1 (approved):
LGTM

PR #2424, opened 2023-08-28
[MLIR][TORCH] add E2E support for aten.rand

123epsilon:
Implements a decomposition for aten.rand that fills an empty tensor with uniformly generated values between 0 and 1. This passes tests in all backends except for `tosa`, there is an existing implementation (merged yesterday) that is specific to `stablehlo` from @Vremold. Please let me know if the implementation is correct :)


vivekkhandelwal1 (changes requested):
@123epsilon, you need to add the e2e tests for this op to ensure the correctness of the lowering.

123epsilon:
@vivekkhandelwal1 there is actually already an e2e test for this op, `RandModule_basic` in `rng.py` in the test suite - previously it only worked for stablehlo but this decomposition extends it to other backends. Please see #2413

Vremold (approved):
LGTM. Please delete the converter from torch dialect to stablehlo dialect registered at [`lib/Conversion/TorchToStablehlo/basic.cpp`](https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchToStablehlo/Basic.cpp#L1556-L1584)

PR #2425, opened 2023-08-29
[MLIR][TORCH] add E2E support for aten.new_full

123epsilon:
Adds a decomposition for `aten.new_full`, passes all backends with the exception that `Int2D_basic` and `Int3D_basic` do not pass in the `tosa` backend.


python/torch_mlir_e2e_test/test_suite/constant_alloc.py
```diff
@@ -1093,6 +1093,164 @@ def forward(self, a):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, a):
+        return torch.ops.aten.new_full(a, (3,4), 10)
```
vivekkhandelwal1, line 1186: I think these 2 tests are not required.

vivekkhandelwal1 (changes requested):

python/torch_mlir_e2e_test/test_suite/constant_alloc.py
123epsilon: Just removed these 👍

PR #2426, opened 2023-08-31
build_tools/python_deploy/build_linux_packages.sh: Disable dynamo testing for stable pytorch

mgehre-amd:


powderluv (approved):

powderluv:
Thank you

vivekkhandelwal1:
The issue is not only limited to the torchdynamo tests. It happens for the make_fx_tosa tests as well. I think, for now, special handling of the failing tests is fine. But if we want to disable the torchdynamo tests for stable build then that's also fine.

mgehre-amd:
> The issue is not only limited to the torchdynamo tests. It happens for the make_fx_tosa tests as well. I think, for now, special handling of the failing tests is fine. But if we want to disable the torchdynamo tests for stable build then that's also fine.

The difference for us it that we use `make_fx` + `tosa` as a stable way to get into `tosa`. Once `dynamo` has matured enough to be adopted, it will become important again to test it with the then-stable version of pytorch.

vivekkhandelwal1 (approved):
Hi @mgehre-amd, can we merge this PR, since the other PRs are failing CI for stable torch dynamo tests?

PR #2436, opened 2023-09-04
[Torch Dialect] emit aten.max.other op

Vremold:
This patch also adds a canonicalizer for `aten.max.other` op (`aten.max.other` -> `aten.maximum`).


qingyunqu (approved):
LGTM

PR #2442, opened 2023-09-06
Fall cleaning: Re-organize sources so that each target type is contai…

stellaraccident:
…ned and optional.

While working on https://github.com/openxla/iree/pull/14917, I noticed that it is somewhat hard to take a dependency on torch-mlir such that one only builds deps for the target(s) of interest (in this case Linalg). I noticed that some ifdef'ey optionality was added for stablehlo, but this was not mirrored for the others. Further, it does the switching very deep in the dependency graph vs having top-level directories and defines gating entire features. In addition, I noticed that a lot of things in the Linalg path were broken down to a fine level of detail but were not actually shared/shareable outside of that target. I opted to clump these together into TorchToLinalg. It is easy enough to "promote" them to common with this new organization if the need arises.

General approach:

* Isolate each conversion target in one of TorchToLinalg, TorchToStablehlo, TorchToTosa.
* Gate each by top-level CMake flags and defines.
* Common conversions go in a Common/ directory (currently Arith and SCF).
* Pull target specific conversions out of TorchConversion/Transforms and put in their top-level directory.
* General maintenance on the build graph and registration stuff that had bitrotted and was blocking progress.

The main functional change for people taking a source dep is that `#include "torch-mlir/Conversion/Passes.h"` no longer is a one stop shop: For optional conversions, you have to include the dedicated `Passes.h` of each and take a library dep. See `InitAll.cpp` which does it right (and *is* a one stop shop still).


maxbartel:
@stellaraccident Is there a chance that this lands? It nicely decouples the "representation" of Torch in MLIR and the various backends.

stellaraccident:
I ran into some CI problems and then meant to check with folks if this was desirable -- and then never did.

I can finish this off if folks think it is a good idea.

maxbartel:
I guess this is more a nice to have then a necessity... I think this would be a good idea, but it also seems like there is some work left to do.

PR #2444, opened 2023-09-07
implemented complex tensor aten mul

brucekimrokcmu:
Supports torch complex dtype tensor mul - `torch.mul`

Note that current torch-mlir does not support multiplication between two different dtype/precision.
(This will be addressed in the future)

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -567,6 +567,8 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     Value rhs = convertScalarToDtype(b, loc, payloadArgs[1], dtype);
     if (dtype.isa<mlir::FloatType>()) {
       return b.create<arith::MulFOp>(loc, lhs, rhs);
+    } else if(dtype.isa<mlir::ComplexType>()) {
+      return b.create<complex::MulOp>(loc, lhs, rhs);
     } else {
       return b.create<arith::MulIOp>(loc, lhs, rhs);
     }
--- a/python/torch_mlir_e2e_test/test_suite/elementwise.py
+++ b/python/torch_mlir_e2e_test/test_suite/elementwise.py
@@ -1042,6 +1042,28 @@ def ElementwiseMulTensorIntModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
+class ElementwiseMulTensorComplexModule(torch.nn.Module):
+
+    def __init__(self):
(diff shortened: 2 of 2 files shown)


stellaraccident (approved):
Thanks!

ramiro050:
> Note that current torch-mlir does not support multiplication between two different dtype/precision.

What error are you seeing on this? I would expect this to work, since the arguments to `AtenMulOp` get casted to the result dtype before multiplied.

brucekimrokcmu:
> > Note that current torch-mlir does not support multiplication between two different dtype/precision.
> 
> What error are you seeing on this? I would expect this to work, since the arguments to `AtenMulOp` get casted to the result dtype before multiplied.

Right, but the result dtype seems wrong. 
```
error: "aten::mul"("/home/brucekimrok/CLionProjects/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/elementwise.py":1057:15): 'complex.mul' op requires the same type for all operands and results
note: "aten::mul"("/home/brucekimrok/CLionProjects/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/elementwise.py":1057:15): see current operation: %12 = "complex.mul"(%arg2, %arg3) : (complex<f32>, complex<f64>) -> complex<f32>

```

ramiro050:
Ah, I see. Yeah, there is no handling for casting complex values to a common type. This needs to be added here: 

https://github.com/llvm/torch-mlir/blob/a8fd275a006b4c3284991bf7ac3608d4f8ecf8cf/lib/Conversion/Utils/Utils.cpp#L251

Not sure how you got all the way down to generating `"complex.mul"(%arg2, %arg3) : (complex<f32>, complex<f64>)`, since `convertScalarToDtype` should crash here

https://github.com/llvm/torch-mlir/blob/a8fd275a006b4c3284991bf7ac3608d4f8ecf8cf/lib/Conversion/Utils/Utils.cpp#L324

But yeah, adding support there should fix it.

ramiro050:
Regarding the result type, when I test things locally, I do get the correct type promoted result type:

```mlir
torch.aten.mul.Tensor %arg0, %arg1 : !torch.vtensor<[?,?],complex<f32>>, !torch.vtensor<[?,?],complex<f64>> -> !torch.vtensor<[?,?],complex<f64>>
```

brucekimrokcmu:
@ramiro050  Thanks a lot! I will make a separate PR to address this since I'm in the middle of a test that requires this to land first.

PR #2457, opened 2023-09-12
[MLIR][TORCH] Add E2E support for aten.empty_strided decomposition op 

brucekimrokcmu:
Add decompose empty_strided op. 
Referring to https://github.com/llvm/torch-mlir/pull/1776, this decomposition op only supports default stride values, because accessing the tensor or indexing over that, the indices are determined by the strides. 
In MLIR, this is not implicitly supported but assumes that the strides are default while iterating over the tensor.

p.s. commented out `check_dtype_function` in `aten.linear` in `abstract_interp_lib_gen.py` because it throws value error. @vivekkhandelwal1


ramiro050:
@brucekimrokcmu, please don't merge changes without review. I've reverted your commit.

brucekimrokcmu:
My bad. I thought it was enabled to be merged because the review was already made.

PR #2459, opened 2023-09-12
[MLIR][TORCH] Add E2E support for aten.empty_strided decomposition op (redo PR)

brucekimrokcmu:
_Making the same PR with  #2457, as I accidentally thought the review was already made and merged it (reverted)._

Add decompose empty_strided op.
Referring to https://github.com/llvm/torch-mlir/pull/1776, this decomposition op only supports default stride values, because accessing the tensor or indexing over that, the indices are determined by the strides.
In MLIR, this is not implicitly supported but assumes that the strides are default while iterating over the tensor.


python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -3424,7 +3432,7 @@ def aten〇atan〡dtype(self_rank_dtype: Tuple[int, int]) -> int:
         return torch.float32
     return self_dtype
 
-@check_dtype_function(_check_two_tensor_op())
+# @check_dtype_function(_check_two_tensor_op())
```
ramiro050, line 3435: What error are you getting? In general, we should not remove tests when they fail. (something to check is if you locally have the same PyTorch version that Torch-MLIR expects: https://github.com/llvm/torch-mlir/commit/c1f379f3bf9185621ff99d8223bed62e914a5140)

ramiro050 (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
brucekimrokcmu: Locally, these are the PyTorch version.
```
8ef4b6af4eb6df3b8377109404fef649b0543245
torch==2.2.0.dev20230911
torchvision==0.17.0.dev20230911
```
And when I `update_abstract_interp_lib.sh`, I get the below error. 
I think it's because `torch.matmul` between different complex dtype is not supported. 
I can exclude such such test case if that's reasonable.
```
ValueError: For dtype function 'aten〇linear〡dtype' with invocation Invocation(TensorOfShape(1, dtype=torch.complex64, device=meta), TensorOfShape(1, dtype=torch.complex128, device=meta)): Expected result dtype torch.comple4, got torch.complex128
```

ramiro050: Can you check the output of `pip list | grep torch`?

brucekimrokcmu: Ah you are right. I'm switching to `venv` from `conda` to avoid this mistake of version mismatch. I will test again and let you know.

stellaraccident:
Note that the Windows CI issue was fixed by https://github.com/llvm/torch-mlir/commit/107ed0dec9f96fc5615c99af01ca25ad68641a29 last night. If this PR is rebased, it will pass. I looked at the logs and am ok landing in the current state (it will pass at head).

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4416,6 +4416,53 @@ class DecomposeAtenNewEmptyStridedOp
 };
 } // namespace
 
+namespace {
+class DecomposeAtenEmptyStridedOp
```
ramiro050, line 4420: This pattern is almost identical to the pattern for `DecomposeAtenNewEmptyStridedOp`. We should avoid copying and pasting code when possible to make maintenance easier. Here, we can avoid repetition by having `AtenNewEmptyStridedOp` decompose into `atenEmptyStridedOp`.

stellaraccident: @brucekimrokcmu can you send an NFC (non-functional change) PR when you have time to address Ramiro's comments on code organization?

brucekimrokcmu: @ramiro050 Hi Ramiro, 
I was initially thinking if I could simply decompose `AtenEmptyStridedOp` into `AtenNewEmptyStridedOp` to avoid any repetition. Could you look at my approach below and give me feed back how it can be improved?

Here's how `new_empty_strided` is decomposed:
`AtenNewEmptyStridedOp` -> `AtenNewEmptyOp` -> `AtenEmptyMemoryFormatOp`
Here, `self` tensor is passed, and is also required by `AtenNewEmptyOp`, according table gen.
 
On the other hand `AtenEmptyStridedOp` does not pass a tensor, and `AtenEmptyMemoryFormatOp` requires no `self` in table gen. So it cannot be decomposed into `AtenNewEmptyOp`, but would have to be decomposed into `AtenEmptyMemoryFormatOp` directly. 

So here, it is decomposed as the following: 
`AtenEmptyStridedOp` -> `AtenEmptyMemoryFormatOp` 

Please let me know your thoughts and I will make NFC PR accordingly.

ramiro050: Yeah, all paths lead to `AtenEmptyMemoryFormatOp`. The `AtenNew...` ops have [value semantics in the table gen generated](https://github.com/llvm/torch-mlir/blob/40913a36c297239523863867f2e9b8639176993f/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td#L8227), which means that the `self` argument is not modified, so we can drop it (it is only used to [set `dtype` and `device` args](https://pytorch.org/docs/stable/generated/torch.Tensor.new_empty.html) if they're set to `None` in `AtenNew...`. `device` can be ignored, since there is no concept of device in torch-mlir. `dtype` can be handled [like this](https://github.com/brucekimrokcmu/torch-mlir/blob/dcd56980c4f4cff2d68dad231f83b29ba4b3e8c8/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L3288-L3297)).

The flow to avoid too much code repetition should be:
- `AtenEmptyStridedOp` -> `AtenEmptyMemoryFormatOp`
- `AtenNewEmptyStridedOp` (dropping `self` and properly setting `dtype` arg) -> `AtenEmptyStridedOp`

brucekimrokcmu: That sounds good! I can work on those after I finish the priorities I have :) Thanks for the feedback @ramiro050

ramiro050: Sure!

PR #2463, opened 2023-09-13
[Bazel] Replace mlir-hlo with stablehlo

sjain-stanford:
Aligns with https://github.com/llvm/torch-mlir/pull/2460 and fixes bazel build.

GHA workflow: https://github.com/sjain-stanford/torch-mlir/actions/runs/6178894329


navahgar (approved):
LGTM

PR #2465, opened 2023-09-14
Merge `main` into `mlir-tcp`

sjain-stanford:
Merges https://github.com/llvm/torch-mlir/commit/3d974ed9883eac4c3651ac7799a49da5ad9c597b into `mlir-tcp`, resolves conflicts and fixes remaining `mlir-hlo` references in `StablehloToTcp`. 

The corresponding hashes are:

```
stablehlo: https://github.com/openxla/stablehlo/tree/77a59815a82b34f7b08ed2d42a711d9920682d0e
llvm: https://github.com/llvm/llvm-project/tree/4acc3ffbb0af5631bc7916aeff3570f448899647
```

Note: 

CI was stuck indefinitely on TCP e2e integration testing step, in particular on `ElementwiseMulTensorComplexModule_basic...` due to `torch.complex64` type which we don't support yet (but also didn't catch in `Tcp` dialect's verifier). This test is not in `TCP_PASS_SET` so it is expected to fail. But in this case it is not failing, just stuck. It hit [this](https://github.com/llvm/torch-mlir/blob/275697de26aa7e9e7455c2a494e0d6fd3a934a9f/externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp#L177) `llvm_unreachable` but didn't fail, just stuck indefinitely and then does timeout at 6 hours.
```
Compiling ElementwiseMulTensorComplexModule_basic...
unsupported element type in createLinalgPayloadForElementwiseOp for tcp.mul
UNREACHABLE executed at /main_checkout/torch-mlir/externals/llvm-external-projects/torch-mlir-dialects/lib/Conversion/TcpToLinalg/Elementwise.cpp:178!
```

So for now I changed the type constraints for `Tcp_MulOp` in `TcpOps.td` from `Tcp_Tensor` to `Tcp_FloatOrIntTensor`, which causes this test to fail (expectedly).


zezhang (approved):
Thanks for working on this!

sjain-stanford:
Note to myself: 
> Once CI is clean, AVOID `Squash and Merge` on the GH UI, instead do a `git merge` locally and push to upstream `mlir-tcp`. This is needed since this repo doesn’t allow regular merge and we’d like to preserve the upstream commit history.

PR #2466, opened 2023-09-15
Bump llvm-project to f66cd9e9556a53142a26a5c21a72e21f1579217c.

stellaraccident:
Picks up DenseResourceElementsAttr python support and fixes minf/maxf C++ rename.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,4 +3,4 @@
 	url = https://github.com/llvm/llvm-project.git
 [submodule "externals/stablehlo"]
 	path = externals/stablehlo
-	url = https://github.com/openxla/stablehlo.git
+	url = https://github.com/shark-infra/stablehlo.git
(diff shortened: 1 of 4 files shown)


vivekkhandelwal1 (approved):

.gitmodules
```diff
@@ -3,4 +3,4 @@
 	url = https://github.com/llvm/llvm-project.git
 [submodule "externals/stablehlo"]
 	path = externals/stablehlo
-	url = https://github.com/openxla/stablehlo.git
+	url = https://github.com/shark-infra/stablehlo.git
```
ramiro050, line 6: What's the reasoning for this?

qingyunqu: I have the same confusion about this.

stellaraccident: Wait, I'm confused and did something wrong. Let me revert this.

Reconstructing what I think was going on here: my local stablehlo git repo was reporting a commit not found and that was because my repository had a remote of a throwaway fork of stablehlo that I had used locally when doing the switch from mhlo. I saw that it was the old fork and updated, thinking that I had accidentally committed the use of the throwaway fork. I switched it to the fork we use for patching on another project, thinking I needed to land some patches on the branch.

In any case, unintentional sequence of events matching work I was doing on a peer project. I apologize -- the change was not intentional (and that repo it is pointing to is a pure mirror of stablehlo).

stellaraccident: Resetting it here: https://github.com/llvm/torch-mlir/pull/2477/files

PR #2467, opened 2023-09-15
[Tcp] Handle int inputs in sqrt

navahgar:
This PR adds support for integer inputs in `tcp.sqrt`.

### DIFF
--- a/externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
+++ b/externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
@@ -88,8 +88,7 @@ class Tcp_UnaryElementwiseOp<string mnemonic, list<Trait> traits = []> :
     Tcp_Op<mnemonic, !listconcat(traits, [
         Pure,
         Elementwise,
-        SameOperandsAndResultShape,
-        SameOperandsAndResultElementType])> {
+        SameOperandsAndResultShape])> {
 }
 
 class Tcp_BinaryElementwiseOp<string mnemonic, list<Trait> traits = []> :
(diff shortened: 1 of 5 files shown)


externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
```diff
@@ -88,8 +88,7 @@ class Tcp_UnaryElementwiseOp<string mnemonic, list<Trait> traits = []> :
     Tcp_Op<mnemonic, !listconcat(traits, [
         Pure,
         Elementwise,
-        SameOperandsAndResultShape,
-        SameOperandsAndResultElementType])> {
+        SameOperandsAndResultShape])> {
```
sjain-stanford, line 91: Since all elementwise ops except `tcp.sqrt` adhere to the `SameOperandsAndResultElementType` trait, is it better to leave this as-is, and let `Tcp_SqrtOp` just be a `Tcp_Op` that uses a relaxed set of traits?

sjain-stanford (approved):
LGTM. You could also drop the complex types from `TcpBase.td` and `TcpOps.td`.

externals/llvm-external-projects/torch-mlir-dialects/include/torch-mlir-dialects/Dialect/Tcp/IR/TcpBase.td
navahgar: It is still useful to be able to identify all the unary elementwise ops using the common base class type. 

If there are more such ops in future, we could add another base type that can be common to all of them.

navahgar:
> You could also drop the complex types from `TcpBase.td` and `TcpOps.td`.

I wanted to do that as a separate PR. Will do as a follow up.

PR #2470, opened 2023-09-18
[NFC] Clean-up `ConvertAtenViewOp` in linalg backend

ramiro050:
While trying to fix a bug in the `ConvertAtenViewOp` pattern in the linalg backend, I realized that the pattern had become quite complex and had accumulated some dead code, making it hard to reason about.

This commit simplifies the pattern quite a bit. The main changes are:
1. All the static helper functions in the `ConvertAtenViewOp` class have been simplified, both in their signature and their body. Each one now performs simple calculations on arrays, and take the least number of arguments necessary.
2. The body of [the `while` loop](https://github.com/ramiro050/torch-mlir/blob/9fce566b0cb64ff2b198693d1f6ee9580b8fa01f/lib/Conversion/TorchToLinalg/DataMovement.cpp#L407) inside the main pattern has been changed to work on `MutableArrayRef` slices, to avoid having to keep track of `start` and `end` indices for the input and output shape arrays.
3. All the heuristics used to determine the mapping between the input and output dimensions are now in [this relatively short `if-else` section](https://github.com/ramiro050/torch-mlir/blob/9fce566b0cb64ff2b198693d1f6ee9580b8fa01f/lib/Conversion/TorchToLinalg/DataMovement.cpp#L428-L460), making it easy to see what is going on.
4. Dead code was eliminated + updates to some of the documentation comments

This commit does not add any new functionality to the `ConvertAtenViewOp` pattern.

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -182,14 +182,14 @@ public:
   // computes the size of a collapsed dynamic dim if necessary.
   // TODO(ramiro050): returning product here feels awkward, but I
   // can't think of better way
-  static FailureOr<int64_t>
+  static LogicalResult
   collapseToSingleDimHelper2(ArrayRef<int64_t> xs, ArrayRef<int64_t> ys,
                              SmallVector<int64_t> &xIndices,
                              SmallVector<int64_t> &yIndices) {
     // TODO(ramiro050): name variable
     auto blah = [](int64_t expectedReductionProduct,
                    ArrayRef<int64_t> arrayToReduce,
-                   SmallVector<int64_t> &reductionIndices) -> FailureOr<int64_t> {
+                   SmallVector<int64_t> &reductionIndices) -> LogicalResult {
       int64_t reductionProduct = 1;
       for (auto [index, elem] : llvm::enumerate(arrayToReduce)) {
         reductionIndices.push_back(index);
@@ -203,7 +203,7 @@ public:
           reductionProduct != expectedReductionProduct) {
         return failure();
       }
-      return reductionProduct;
+      return success();
     };
 
     if (xs.size() == 1) {
@@ -273,8 +273,8 @@ public:
     return success(xTotalSize == yTotalSize);
   }
 
-  static void solveDynamicSize(SmallVector<int64_t> &inputShape,
-                               SmallVector<int64_t> &outputShape) {
+  static void solveSingleDynamicSize(MutableArrayRef<int64_t> inputShape,
+                                     MutableArrayRef<int64_t> outputShape) {
     int64_t inputDynamicDimCount = llvm::count(inputShape, kUnknownSize);
     int64_t outputDynamicDimCount = llvm::count(outputShape, kUnknownSize);
     if (inputDynamicDimCount + outputDynamicDimCount != 1)
@@ -311,7 +311,7 @@ public:
   // TODO: think of a way better way to at least detect when this assumption
   // is violated for the cases of dynamic dimensions.
   static FailureOr<SmallVector<int64_t>>
-  getStaticInformation(SmallVector<int64_t> &inputShape,
+  getStaticInformation(MutableArrayRef<int64_t> inputShape,
                        SmallVector<Value> outputSizeTorchInt, AtenViewOp op,
                        PatternRewriter &rewriter) {
     SmallVector<int64_t> outputShape(outputSizeTorchInt.size(), kUnknownSize);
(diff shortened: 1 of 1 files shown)


ramiro050:
@vivekkhandelwal1 @qedawkins @Abhishek-Varma , can one of you PTAL?

qedawkins:
> @vivekkhandelwal1 @qedawkins @Abhishek-Varma , can one of you PTAL?

Sorry for taking so long here. I was in the middle of reviewing just as you sent this.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -349,105 +337,37 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
-    // 2. Mark dims in unchangedDims for size list items where the output dim
-    // size comes from a `torch.aten.size.int(inputTensor, inputDim)`. We
-    // naively assume this means the corresponding dimension is not expanded or
-    // collapsed. Note this may technically not always be true.
-    // TODO: think of a way better way to at least detect when this assumption
-    // is violated for the cases of dynamic dimensions.
-    SmallVector<int64_t> outputShape(resultRank, kUnknownSize);
-    SmallVector<ReassociationIndices> unchangedDims;
-    std::optional<int64_t> inferredDimension;
-    for (auto en : llvm::enumerate(outputSizeTorchInt)) {
```
qedawkins, line 380: I think the comment that used to be here was still helpful. Correct me if I misread, but it seems like 2) is still true.
```
    // 2. Mark dims in unchangedDims for size list items where the output dim
    // size comes from a `torch.aten.size.int(inputTensor, inputDim)`. We
    // naively assume this means the corresponding dimension is not expanded or
    // collapsed. Note this may technically not always be true.
```

qedawkins (approved):
Thanks for the cleanup! Looks good to me with a few comments. As an additional aside, it might also be helpful to add a few `LLVM_DEBUG` statements throughout the lowering (e.g. `unchangedDims` and prints on how it decides to reassociate) if it's easy to add them for the next time someone needs to work on this lowering.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -177,144 +181,129 @@ namespace {
...
+        return failure();
+      xIndices.assign({0});
+      yIndices.assign(llvm::to_vector(llvm::seq<int64_t>(0, yDims.size())));
+    } else if (yDims.size() == 1) {
+      if (!isValidReduction(yDims[0], xDims))
+        return failure();
+      yIndices.assign({0});
+      xIndices.assign(llvm::to_vector(llvm::seq<int64_t>(0, xDims.size())));
+    } else {
+      return failure();
```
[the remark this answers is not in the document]
ramiro050: Done!

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -177,144 +181,129 @@ namespace {
...
+  // with the indices of the dims arrays that correspond to the subsets found.
+  //
+  // An error is returned if two subsets of dims with total number of elements
+  // equal to each other is not found.
+  static LogicalResult mapStaticallyKnownDims(ArrayRef<int64_t> xDims,
+                                              ArrayRef<int64_t> yDims,
+                                              SmallVector<int64_t> &xIndices,
+                                              SmallVector<int64_t> &yIndices) {
+    int64_t xTotalSize = xDims[0];
+    int64_t yTotalSize = yDims[0];
```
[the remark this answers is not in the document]
ramiro050: Done!

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: Indeed, I got too carried away removing lines haha. I've added the comment back

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -482,119 +414,75 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+                     llvm::count(inputShapeSlice, kUnknownSize) > 0) {
+            outputShapeSlice[0] = kUnknownSize;
           }
-          hasDynamic = false;
-          outputDim = outputAssociations.back().back() + 1;
-          inputDim = inputAssociations.back().back() + 1;
-          continue;
+        } else if (succeeded(mapStaticallyKnownDims(
+                       inputShapeSlice, outputShapeSlice, inputSliceIndices,
+                       outputSliceIndices))) {
```
qedawkins, line 461: There is a little bit of action at a distance here combined with the fact that this is any empty block. Might be worth adding a comment like
```
/// Maps the smallest number of input and output dimensions in the slice
/// statically known to have the same number of elements.
```

qedawkins (approved):
LGTM! Nice red commit :)

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: Yeah, not a fan of the empty blocks here. The comment does help

PR #2475, opened 2023-09-19
[Torch Dialect] Add Support for aten.unflatten.int

JianzheXiao:
As title, Add support for aten.unflatten.int, support dim to be negative and one of the sizes' elements to be -1


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1452,6 +1452,90 @@ class DecomposeAtenFlattenUsingIntsOp
...
+          int64_t sizeInt;
+          if (!matchPattern(sizesTorchInt[j], m_TorchConstantInt(&sizeInt))) {
+            return rewriter.notifyMatchFailure(
+                op, "Size is expected to be a constant");
+          }
+          if (sizeInt == -1) {
+            if (inferred) {
+              return rewriter.notifyMatchFailure(
+                  op, "Only One of Sizes' elements can be -1");
+            };
```
ramiro050, line 1509: nit: it is better to do the error checking at the beginning to avoid creating IR when the pattern is not matched.

You should be able to do something like:

```c++
SmallVector<int64_t> sizesInts;
if (!matchPattern(op.getSizes(), m_TorchListOfConstantInts(sizesInts)))

if (llvm::count(sizesInts, -1) > 1)
```

lib/Dialect/Torch/Utils/Utils.cpp
```diff
@@ -199,7 +199,7 @@ bool Torch::isViewLikeOp(Operation *op) {
   // that it does not return a view and treat those as having value
   // semantics.
   return isa<AtenBroadcastToOp, AtenContiguousOp, AtenDetachOp, AtenExpandAsOp,
-             AtenExpandOp, AtenFlattenUsingIntsOp, AtenPermuteOp, AtenReshapeOp,
+             AtenExpandOp, AtenFlattenUsingIntsOp, AtenUnflattenIntOp, AtenPermuteOp, AtenReshapeOp,
```
ramiro050, line 202: nit: make sure to run `git clang-format` so that this list is nicely formatted

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -605,6 +605,27 @@ def adaptive_avg_pool1d(self: List[int], out: List[int]):
 
     return shape
 
+
+# TODO: This should be upstreamed.
+# See https://github.com/pytorch/pytorch/pull/76889 for an example.
```
ramiro050, line 610: no need for this comment

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -605,6 +605,27 @@ def adaptive_avg_pool1d(self: List[int], out: List[int]):
...
+def unflatten(self: List[int], dim: int, sizes: List[int]):
```
ramiro050, line 611: This function has non-trivial logic. Can you add tests to it? See for an example: 

https://github.com/llvm/torch-mlir/blob/a520d39f84f159838cc2f4f88a08c9cd611635fd/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py#L469-L472

Also, no need to refactor into a helper function, since the shape function `aten〇unflatten〇int〡shape`mis the only one using this.

ramiro050 (changes requested):
Thanks for this! A few initial comments

ramiro050 (approved):
Thanks!

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -620,6 +620,30 @@ def aten〇adaptive_avg_pool2d〡shape(self: List[int], output_size: List[int])
 def aten〇flatten〇using_ints〡shape(self: List[int], start_dim: int = 0, end_dim: int = -1) -> List[int]:
     return upstream_shape_functions.flatten(self, start_dim, end_dim)
 
+@check_shape_function([
+    Invocation(TensorOfShape(3, 6, 8), 1, [3, 2]),
+    Invocation(TensorOfShape(3, 6, 8), 1, [3, -1]),  # contain one -1 in sizes
+    Invocation(TensorOfShape(3, 6, 8), -1, [2, -1, 2]),  # dim = -1
+    ErrorInvocation(TensorOfShape(3, 6, 8), 1, [-1, -1, 2]), # contain two -1 in sizes
+])
+def aten〇unflatten〇int〡shape(self: List[int], dim: int, sizes: List[int]) -> List[int]:
```
qingyunqu, line 629: I think we can reuse `aten.view`'s shape function:
```
def aten〇unflatten〇int〡shape(self: List[int], dim: int, sizes: List[int]) -> List[int]:
    if dim < 0:
        dim = dim + len(self)
    unflatten_shape: List[int] = [self[dim]]
    unflatten_shape_output = upstream_shape_functions.view(unflatten_shape, sizes)
    shape: List[int] = []
    for i in range(len(self)):
        if i != dim:
            shape.append(self[i])
        else:
            shape = shape + unflatten_shape_output
    return shape
```

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1452,6 +1452,94 @@ class DecomposeAtenFlattenUsingIntsOp
...
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value self = op.getSelf();
+    MLIRContext *context = op.getContext();
+    std::optional<unsigned> maybeRank = getTensorRank(self);
+    if (!maybeRank)
+      return rewriter.notifyMatchFailure(op, "unimplemented: unranked tensor");
+    unsigned inputRank = *maybeRank;
+    BaseTensorType inputTensorType = self.getType().cast<BaseTensorType>();
+    if (!inputTensorType.hasSizes())
```
qingyunqu, line 1471: Check `outputTensorType` is more restrict than `inputTensorType`. Also, we should check inputTensor should have `ValueSemantics`.

qingyunqu (changes requested):

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
ramiro050: Good catch!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
qingyunqu: => 
`auto inputTensorType = self.getType().cast<Torch::ValueTensorType>();`
`if (!inputTensorType || !inputTensorType.hasSizes()) {`
`  return rewriter.notifyMatchFailure(op,`
`                                         "Expected input type having sizes");`
`}`

python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -623,8 +623,23 @@ def aten〇adaptive_avg_pool2d〡shape(self: List[int], output_size: List[int])
...
+    Invocation(TensorOfShape(3, 6, 8), 1, [3, -1]),  # contain one -1 in sizes
+    Invocation(TensorOfShape(3, 6, 8), -1, [2, -1, 2]),  # dim = -1
+])
 def aten〇unflatten〇int〡shape(self: List[int], dim: int, sizes: List[int]) -> List[int]:
-    return self[:dim] + sizes + self[dim + 1:]
+    if dim < 0:
+        dim += len(self)
+    unflatten_shape: List[int] = [self[dim]]
+    unflatten_shape_output = upstream_shape_functions.view(unflatten_shape, sizes)
+    shape: List[int] = []
```
qingyunqu, line 636: `return self[:dim] + unflatten_shape_output + self[dim+1:]` maybe more simple

qingyunqu (approved):
LGTM

PR #2476, opened 2023-09-20
Fixing implicit double to float casts.

benvanik:
MSVC (and other compilers with implicit narrowing warnings) don't like this type mismatch.


ramiro050 (approved):
Thanks!

PR #2478, opened 2023-09-21
add e2e support for torch.eye operations (aten.eye, aten.eye.m)

saienduri:


python/torch_mlir_e2e_test/test_suite/constant_alloc.py
```diff
@@ -215,6 +215,186 @@ def OnesModuleCPUDevice_basic(module, tu: TestUtils):
...
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+    ])
+    def forward(self):
+        return torch.ones(3, dtype=torch.float32, pin_memory=False)
```
ramiro050, line 281: there's a couple of tests that return `torch.ones` rather than `torch.eye`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -377,6 +377,117 @@ class DecomposeAtenZeroOp
...
+      return rewriter.notifyMatchFailure(
+          op, "Only tensor types input are currently supported");
+    if (!outType.hasDtype()) {
+      return rewriter.notifyMatchFailure(op, "result should have dtype");
+    }
+    if (n < 0) {
+      return op.emitError("n must be greater or equal to 0");
+    }
+    if (m < 0) {
+      return op.emitError("m must be greater or equal to 0");
```
ramiro050, line 423: nit: for errors in patterns, the convention here is to use `rewriter.notifyMatchFailure`. Errors should be emitted by whole passes rather than single patterns.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -377,6 +377,117 @@ class DecomposeAtenZeroOp
...
+    auto arangeType =
+        ValueTensorType::get(context, llvm::ArrayRef(n), si64Type);
+    Value range_n = rewriter.create<AtenArangeOp>(
+        loc, arangeType, op.getN(), /*dtype=*/int64Dtype, /*layout=*/none,
+        /*device=*/op.getDevice(), /*pin_memory=*/none);
+    
+
+    auto arangeType1 =
+        ValueTensorType::get(context, llvm::ArrayRef(m), si64Type);
+    Value range_m = rewriter.create<AtenArangeOp>(
```
ramiro050, line 440: nit: variables should be camelcase

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -377,6 +377,117 @@ class DecomposeAtenZeroOp
...
+        loc, arangeType1, op.getM(), /*dtype=*/int64Dtype, /*layout=*/none,
+        /*device=*/none, /*pin_memory=*/none);
+    auto range_mType = range_m.getType().cast<BaseTensorType>();
+    if (!range_mType.hasSizes()) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: input must have known sizes");
+    }
```
ramiro050, line 447: There shouldn't be a need to check this. The type assigned as the return type is the first argument into `rewriter.create<AtenArangeOp>`; namely, `arangeType1`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -377,6 +377,117 @@ class DecomposeAtenZeroOp
...
+    if (failed(unsqzTensorInfo)) {
+      return rewriter.notifyMatchFailure(op,
+                                         "cannot generate unsqueeze tensor");
+    }
+    Value unsqzRangeN = *unsqzTensorInfo;
+    auto range_nType = unsqzRangeN.getType().cast<BaseTensorType>();
+    if (!range_nType.hasSizes()) {
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: input must have known sizes");
+    }
```
ramiro050, line 464: Same comment as above, this check should not be needed

ramiro050 (changes requested):
Thanks for this! A few comments

python/torch_mlir_e2e_test/test_suite/constant_alloc.py
saienduri: fixed

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
saienduri: fixed

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
saienduri: fixed

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
saienduri: fixed

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
saienduri: fixed

saienduri:
Hey @ramiro050, thanks for the feedback! I made the changes, and I think it is good to go. Please let me know if anything else is needed.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -377,6 +377,107 @@ class DecomposeAtenZeroOp
...
+namespace {
+class DecomposeAtenEyeOp
+    : public OpRewritePattern<AtenEyeOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenEyeOp op,
+                                PatternRewriter &rewriter) const override {
+    Value n = op.getN();
+    Value m = op.getN();
+    rewriter.replaceOpWithNewOp<AtenEyeMOp>(op, op.getType(), n, m, op.getDtype(), op.getLayout(), op.getDevice(), op.getPinMemory());
```
ramiro050, line 389: nit: the formatting is off. Make sure you use `git clang-format` to format your changes

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -377,6 +377,107 @@ class DecomposeAtenZeroOp
...
+    
+
+    auto arangeType1 =
+        ValueTensorType::get(context, llvm::ArrayRef(m), si64Type);
+    Value rangeM = rewriter.create<AtenArangeOp>(
+        loc, arangeType1, op.getM(), /*dtype=*/int64Dtype, /*layout=*/none,
+        /*device=*/none, /*pin_memory=*/none);
+
+      
+
```
ramiro050, line 445: nit: unnecessary empty lines

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -377,6 +377,107 @@ class DecomposeAtenZeroOp
...
+    auto si64Type = IntegerType::get(context, 64, IntegerType::Signed);
+    auto arangeType =
+        ValueTensorType::get(context, llvm::ArrayRef(n), si64Type);
+    Value rangeN = rewriter.create<AtenArangeOp>(
+        loc, arangeType, op.getN(), /*dtype=*/int64Dtype, /*layout=*/none,
+        /*device=*/op.getDevice(), /*pin_memory=*/none);
+    
+
+    auto arangeType1 =
+        ValueTensorType::get(context, llvm::ArrayRef(m), si64Type);
```
ramiro050, line 439: There is no guarantee at this point in the lowering that things have value semantics. It is better to create new types using

```
arangeType1 = outType.getWithSizesAndDtype(llvm::ArrayRef(m), si64Type);
```

python/torch_mlir_e2e_test/test_suite/constant_alloc.py
ramiro050, line 1: You need to add your tests to this list to make the CI pass (tosa tests are `xfail`ed by default, so we need to manually set them to `xpass`)

https://github.com/llvm/torch-mlir/blob/6d74e8cccd581ef75cbe9de6f76db2672cffb103/e2e_testing/xfail_sets.py#L1280

ramiro050 (approved):
Thanks! Just a few small changes, but other than that it LGTM

python/torch_mlir_e2e_test/test_suite/constant_alloc.py
saienduri: @ramiro050, so I've added the tests to the pass set. But, there is still one failing so I put it in the xfail set with a description on the error as you can see in the xfail_sets.py file. Not sure why it fails, as it only fails with tosa backend. Does this work? (now 8 tests pass, and 2 xfail)

ramiro050: Yeah, this should be fine. The make_fx path is probably adding extra ops to the graph that are causing things to fail. Given that all your tests pass on the linalg backend, there's good reason to believe your decomposition is good.

saienduri: Ok cool, sounds good. Thanks for the feedback for this PR!

saienduri:
Hey @ramiro050, can you merge this whenever you get a chance (I don't have write access). Thanks!

PR #2482, opened 2023-09-23
CI: reconcile differences between RollPyTorch and pre-merge checks

ashay:
The pre-merge checks that build PyTorch from source use an out-of-tree
build whereas, before this patch, the RollPyTorch workflow used an
in-tree build.  This caused inconsistencies in the build output, so this
patch reconciles the differences.

### DIFF
--- a/.github/workflows/RollPyTorch.yml
+++ b/.github/workflows/RollPyTorch.yml
@@ -24,8 +24,20 @@ jobs:
     - name: Get torch-mlir
       uses: actions/checkout@v3
       with:
-        submodules: 'true'
+        submodules: 'false'
         token: ${{ secrets.WORKFLOW_INVOCATION_TOKEN }}
+
+    - name: Get LLVM and StableHlo submodules
+      run: |
+        set -eo pipefail
+        cd ${GITHUB_WORKSPACE}
+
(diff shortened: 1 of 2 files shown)


ashay:
[Here](https://github.com/llvm/torch-mlir/actions/runs/6282879507/job/17062786337) is the most recent run.

.github/workflows/RollPyTorch.yml
```diff
@@ -24,9 +24,21 @@ jobs:
...
+        set -eo pipefail
+        cd ${GITHUB_WORKSPACE}
+
+        # Fetching the submodules concurrently may cause problems, so we fetch
+        # them one after another.
+        rm -f .git/modules/externals/llvm-project/index.lock
+        rm -f .git/modules/externals/stablehlo/index.lock
+        git submodule update --init --recursive externals/llvm-project
+        git submodule update --init --recursive externals/stablehlo
+
```
ashay, line 41: This change is unrelated to this PR, but it helps work around the index.lock issue arising from concurrent submodule fetches.

vivekkhandelwal1 (approved):
Thanks @ashay for the fix.

powderluv (approved):

powderluv:
Thank you

PR #2485, opened 2023-09-26
[RFC, WIP] Collapse leading unit dims on input shapes

dan-garvey:
additionally uses derived output shape as a hint in get input and output shapes.

Covers some cases like
before:
```1,-1
-1,-1```
after:
```1,-1
1,-1```

I left some notes in the code for now so anyone reviewing can see the thought process

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -308,15 +308,25 @@ public:
   // `aten.view`.
   static std::pair<SmallVector<int64_t>, SmallVector<int64_t>>
   getInputAndOutputShape(Value inputTorchTensor,
-                         SmallVector<Value> outputSizeTorchInt) {
+                         SmallVector<Value> outputSizeTorchInt, ArrayRef<int64_t> resultTypeShape) {
     SmallVector<int64_t> inputShape(
         inputTorchTensor.getType().cast<BaseTensorType>().getSizes());
+    //SmallVector<int64_t> outputShape(resultTypeShape.begin(), resultTypeShape.end());
     SmallVector<int64_t> outputShape(outputSizeTorchInt.size(), kUnknownSize);
-    for (auto [outputDim, outputDimSize] :
-         llvm::enumerate(outputSizeTorchInt)) {
+    for (const auto &it :
+         llvm::enumerate(llvm::zip_equal(outputSizeTorchInt, resultTypeShape))) {
+      int64_t outputDim = it.index();
+      Value outputDimSize = std::get<0>(it.value());
+      int64_t resultShapeSize = std::get<1>(it.value());
+    // for (auto [outputDim, outputDimSize] :
+    //      llvm::enumerate(outputSizeTorchInt)) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lib/Conversion/TorchToLinalg/DataMovement.cpp [hunk 1]
```diff
@@ -284,15 +284,25 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
   // `aten.view`.
   static std::pair<SmallVector<int64_t>, SmallVector<int64_t>>
   getInputAndOutputShape(Value inputTorchTensor,
```
ramiro050, line 286: Are the changes here necessary? (from doing some experimenting locally, the handling of leading dims seems to work without this)

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -472,7 +491,27 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+	  // outputShape = [2,3]
+	  // inReassociation = [{0},...]
+	  // outReassociation = [{0},...]
+	  // inputShape =  [1 + 1, 1]
+	  // outputShape = [1,     1 + 1]
+	  // inputShape = [1, 1, 2, ?, 2]
+	  // outputShape = [2, ?, 2]
+        }
+	
+	if (inputSliceIndices.empty() || outputSliceIndices.empty()) {
```
ramiro050, line 514: We should keep the original design of having the `else` branch result in an error. It makes it very clear that we are trying a bunch of things in the `if-else` block, and if none of them work, we return with an error.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -472,7 +491,27 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
           inputSliceIndices.push_back(0);
           outputSliceIndices.push_back(0);
           assumedDynamicDimNotSplit = true;
-        } else {
+        } else if (inputShapeSlice[0] == 1 && outputShapeSlice[0] == kUnknownSize) {
+	  int64_t idx = 0;
+          while (idx < inputShapeSlice.size() && inputShapeSlice[idx] == 1) {
+            inputSliceIndices.push_back(idx++);
+	  }
+	  if (idx < inputShapeSlice.size() && inputShapeSlice[idx] == kUnknownSize) {
```
ramiro050, line 499: We should also have a `checkDimEqualHelper` call in here, similar to the `else if (inputShapeSlice[0] == kUnknownSize)` case above

ramiro050 (changes requested):
Looking good. Missing e2e test

lib/Conversion/TorchToLinalg/DataMovement.cpp
dan-garvey: This was actually the easier of the two problems we had to solve, and I just forgot to mention we fixed it, but the idea here is that if we have information about the resultType already, we should use that to calculate outputDims. I'll get an example to show

PR #2486, opened 2023-09-27
add e2e support for aten.xlogy.Tensor

saienduri:


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3139,6 +3139,47 @@ class DecomposeAtenMishOp : public OpRewritePattern<AtenMishOp> {
 };
 } // namespace
 
+namespace {
+// Decompose `aten.mish` op into `aten.tanh` and `aten.softplus` ops.
+// Mish(x) = x * Tanh(Softplus(x))
```
ramiro050, line 3144: nit: comment is for the wrong op

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3139,6 +3139,47 @@ class DecomposeAtenMishOp : public OpRewritePattern<AtenMishOp> {
...
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    auto context = op.getContext();
+    auto resultType = op.getType().cast<BaseTensorType>();
+    Value constZero = rewriter.create<Torch::ConstantIntOp>(
+        loc, rewriter.getI64IntegerAttr(0));
+
+    auto eqType = ValueTensorType::get(
+        context, op.getType().cast<BaseTensorType>().getSizes(),
+        IntegerType::get(context, 1));
```
ramiro050, line 3158: There is no guarantee that the types will have value semantics at this stage. It is better to create the type from the existing types:

```c++
auto resultType = op.getType().cast<BaseTensorType>();
auto eqType = resultType.getWithSizesAndDtype(
    resultType.getOptionalSizes(), IntegerType::get(context, 1));
```

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3139,6 +3139,47 @@ class DecomposeAtenMishOp : public OpRewritePattern<AtenMishOp> {
...
+          loc, eqType, op.getSelf(), constZero);
+
+    Value logOther = rewriter.create<AtenLogOp>(loc, resultType, op.getOther());
+    Value selfMulLogOther = rewriter.create<AtenMulTensorOp>(loc, resultType, op.getSelf(), logOther);
+    Value rhs = rewriter.create<AtenWhereScalarSelfOp>(
+        loc, resultType, eqTensor, constZero, selfMulLogOther);
+
+    auto isnanType = ValueTensorType::get(
+        context, op.getType().cast<BaseTensorType>().getSizes(),
+        IntegerType::get(context, 1));
```
ramiro050, line 3169: Same comment as above about creating new types

python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -1130,6 +1130,51 @@ def ElementwiseMishModule_basic(module, tu: TestUtils):
...
+        ([-1, -1], torch.float32, True),
+        ([-1, -1], torch.float32, True),
+    ])
+    def forward(self, x, y):
+        return torch.ops.aten.xlogy(x, y)
+
+
+@register_test_case(module_factory=lambda: ElementwiseXlogyTensorModule())
+def ElementwiseXlogyTensorModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 5), tu.rand(3, 5))
```
ramiro050, line 1152: Can you also add tests that explicitly test the zero and NAN case? (i.e. doing `module.forward(torch.zeros(3, 5), ...)` and `module.forward(... , torch.tensor(float('nan')))`

ramiro050 (changes requested):
Thanks!

saienduri:
NaN case needs some work (not sure if NaN constant is being created properly). will reopen PR when resolved

josephbak:
I came across this while looking into `aten.xlogy.Tensor` — it still has no lowering path on current `main` (built from source and confirmed it passes through decompose + convert-to-linalg untouched). Looks like the work here was nearly done and just stalled on the NaN-constant handling.

I'd like to try picking it up. I want to dig into that NaN-constant issue first, since it seems to be the crux; if I can sort it out, I'll open a fresh PR with the review feedback here applied (the `getWithSizesAndDtype` change and the zero/NaN test cases).

Is a decomposition for `xlogy` still wanted, or has interest moved on since this closed?

Thanks!

PR #2496, opened 2023-09-29
Elide dynamic broadcast checks when in strict symbolic shapes mode.

stellaraccident:
When importing dynamic shaped programs from Dynamo, via torch.compile or torch.export, we can assume that strict symbolic shape checks have been done prior to generating torch IR. Among other shape checking, this eliminates the case where an unknown dimension can be dynamically '1' in a way that signals a broadcast.

Adds a `isAssumingStrictSymbolicShapes` utility which consults a `torch.assume_strict_symbolic_shapes` attribute on an enclosing scope and returns true if present.

In the linalg pipeline, many runtime checks are elided when this returns true.


stellaraccident:
(still working on some unit test updates)

ramiro050 (approved):
Nice!

PR #2474, opened 2023-09-19
[linalg] Fix handling of trailing size-1 dimensions in aten.view

ramiro050:
This commit adds to the lowering of `aten.view` handling for the
following cases:

- `(..., a.size(i))` -> `(..., a.size(i), 1, ..., 1)`
- `(..., a.size(i), 1, ..., 1)` -> `(..., a.size(i))`

Fixes: https://github.com/llvm/torch-mlir/issues/2448

PR #2501, opened 2023-10-03
[linalg] Add handling for leadin and trailing size-1 dims in ViewOp

ramiro050:
After trying to generalize https://github.com/llvm/torch-mlir/pull/2474 to handle leading size-1 dims, it became clear that the approach used there was not the right one. Therefore, this PR:

- Reverts https://github.com/llvm/torch-mlir/pull/2474
- Adds to the lowering of `aten.view` handling for the following cases:
  - `(..., a.size(i))` -> `(..., a.size(i), 1, ..., 1)`
  - `(..., a.size(i), 1, ..., 1)` -> `(..., a.size(i))`
  - `(a.size(i), ...)` -> `(1, ..., 1, a.size(i), ...)`
  - `(1, ..., 1, a.size(i), ...)` -> `(a.size(i), ...)`

I recommend reviewing commits separately.

### DIFF
--- a/python/torch_mlir_e2e_test/test_suite/reshape_like.py
+++ b/python/torch_mlir_e2e_test/test_suite/reshape_like.py
@@ -672,6 +672,108 @@ class ViewNegativeStaticModule(torch.nn.Module):
 def ViewNegativeStaticModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(1, 128))
 
+class ViewSizeDimFollowedByExpandedOnesModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1], torch.float32, True),
+    ])
+
+    def forward(self, a):
+        return a.view(a.size(0), 1, 1, 1)
+
+@register_test_case(module_factory=lambda: ViewSizeDimFollowedByExpandedOnesModule())
+def ViewSizeDimFollowedByExpandedOnesModule_basic(module, tu: TestUtils):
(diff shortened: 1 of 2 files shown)


python/torch_mlir_e2e_test/test_suite/reshape_like.py
```diff
@@ -672,6 +672,74 @@ def forward(self, a):
...
+        ([1, 1, 1, -1], torch.float32, True),
+    ])
+
+    def forward(self, a):
+        return a.view(a.size(3))
+
+@register_test_case(module_factory=lambda: ViewSizeDimLedByCollapsedOnesModule())
+def ViewSizeDimLedByCollapsedOnesModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(1, 1, 1, 128))
+
```
qedawkins, line 742: It might be worth adding a test for both at the same time, e.g. `?` -> `1, 1, ?, 1, 1`

qedawkins (approved):
Thanks for the fix. Agree with the revert and the new fix looks good to me.

python/torch_mlir_e2e_test/test_suite/reshape_like.py
ramiro050: Good idea. Added!

brucekimrokcmu (approved):
LGTM

qedawkins (approved):

dan-garvey (approved):
Thanks, I was dragging my feet a little as I had some other kinks to work out but this looks like it will work

PR #2505, opened 2023-10-05
[TorchToLinalg] Improve broadcast lowerings in strict symbolic modes

qedawkins:
With strict symbolic shapes, we can assume numpy-style dynamic broadcasts never occur. This improves the lowering in the presence of this assumption.


Abhishek-Varma (approved):
Thanks for this. Looks great!

PR #2509, opened 2023-10-10
Add aten.unflatten.int support and its torch-to-tosa lowering

zezhang:
Add aten.unflatten.int op
Add its torch-to-tosa lowering
Update the TorchToTosa/basic.mlir tests

To test e2e tosa lowering:

`python -m e2e_testing.main -v -c=tosa`


navahgar (approved):
LGTM

I only see a unit test for TorchToTosa. What about other tests? Any e2e tests expected to pass? Can you confirm that we are missing anything?

zezhang:
> Can you also add an e2e test similar to the flatten ones:
> 
> https://github.com/llvm/torch-mlir/blob/9b5a4afadd5df5b7f770e2524f9214741707f546/python/torch_mlir_e2e_test/test_suite/basic.py#L307

Just added an e2e test. Note that the aten.unflatten.int op only has TOSA lowering for now, and the e2e test will fail on the Linalg-on-Tensors backend. @ramiro050 Please let me know if you think this is ok.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -304,6 +304,29 @@ def AddmmModule_differentRankBroadcastable(module, tu: TestUtils):
...
+        super().__init__()
+        self.unflat = torch.nn.Unflatten(1, (2, 3))
+
+    @export
+    @annotate_args([
+        None,
+        ([1, 6, 4], torch.float32, True),
+    ])
+    def forward(self, x):
+        return self.unflat(x)
```
ramiro050, line 319: nit: can you use `torch.ops.aten.unflatten(x, 1, (2, 3))` to make it clear that `aten.unflatten` is being tested? See: https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#forward-method

ramiro050 (approved):
Thanks! Other than a small change request, LGTM

PR #2510, opened 2023-10-11
Add compilation support for dynamic shapes

arnab-oss:
Add compilation support for dynamic shapes

### DIFF
--- a/python/torch_mlir/__init__.py
+++ b/python/torch_mlir/__init__.py
@@ -116,7 +116,7 @@ class TensorPlaceholder:
             dynamic_axes = []
         shape = []
         for i, dim in enumerate(tensor.shape):
-            if i in dynamic_axes:
+            if i in dynamic_axes or isinstance(dim, torch.SymInt):
                 shape.append(-1)
             else:
                 shape.append(dim)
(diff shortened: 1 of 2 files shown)


test/python/importer/jit_ir/node_import/dynamic_add.py
ramiro050, line 1: This test seems to pass without the changes in this PR

python/torch_mlir/__init__.py
```diff
@@ -177,10 +177,10 @@ def _canonicalize_args(example_args: _example_args_for_one_method):
         if not isinstance(example_args, Sequence):
             example_args = [example_args]
         for arg in example_args:
-            if not isinstance(arg, (TensorPlaceholder, torch.Tensor)):
-                raise Exception(f"Only Tensor's, TensorPlaceholder's, or sequences of "
-                                f"Tensor's and TensorPlaceholder's are supported as "
-                                f"example args for method inputs. "
+            if not isinstance(arg, (TensorPlaceholder, torch.Tensor, torch.SymInt)):
```
ramiro050, line 180: Do you have an example of the use-case you're trying to handle with these changes?

ramiro050 (changes requested):

PR #2512, opened 2023-10-12
Add aten.isclose support and its torch-to-tosa lowering

zezhang:
Add aten.isclose op
Add its torch-to-tosa lowering
Update the TorchToTosa/basic.mlir tests


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3866,6 +3866,63 @@ LogicalResult ConvertAtenOp<AtenLeTensorOp>::matchAndRewrite(
...
+  auto rhsAbsOp =
+      rewriter.create<tosa::AbsOp>(op->getLoc(), selfType, rhsSubOp);
+
+  auto lhsAbsOp =
+      rewriter.create<tosa::AbsOp>(op->getLoc(), otherType, adaptor.getOther());
+  auto constType =
+      RankedTensorType::get(otherType.getShape(), rewriter.getF32Type());
+  auto rtolAttr = DenseElementsAttr::get(constType, static_cast<float>(rtol));
+  auto rtolConstOp =
+      rewriter.create<tosa::ConstOp>(op->getLoc(), constType, rtolAttr);
```
navahgar, line 3910: Are there any existing libraries we can use to create `tosa::ConstOp`?

navahgar (changes requested):
One comment inlined below.

Can you also add e2e test for this?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
zezhang: Added the e2d test.

**tosa::getTosaConstTensorSingleF32** doesn't create the the proper shape, and **tosa::getConstTensor** requires us to provide a const vector with the same size of the tensor. I think it's just easier to create a const op in place. Let me what you think. I can use the **tosa::getConstTensor** util if you think it's better.

navahgar: Why do we need to create a const of the shape corresponding to the other tensor. We should be able to create a rank-0 tensor with that const (using `getTosaConstTensorSingleF32`) and use it any binary op, since Tosa ops should support implicit broadcasting. Am I missing something here?

zezhang: Ah.. forget TOSA can do implicit broadcast. Just updated.

navahgar (approved):
LGTM

navahgar:
Requested a review from @ramiro050, in case he has any other comments.

python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -4558,3 +4558,26 @@ def forward(self, x):
...
+        ([5, 5], torch.float32, True),
+        ([5, 5], torch.float32, True),
+    ])
+    def forward(self, x, y):
+        return torch.isclose(x, y)
+
+
+@register_test_case(module_factory=lambda: IscloseStaticModule())
+def IscloseStaticModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(5, 5), tu.rand(5, 5))
```
ramiro050, line 4583: This will likely always result in the `torch.isclose == false`. Might be worth adding a test with the same tensor as input (for example `torch.ones` for `x` and `y`) to make sure the equal case works as expected too

ramiro050 (approved):
Small change, but other than that LGTM

PR #2514, opened 2023-10-13
add fallback for unimplemented cases in aten view

saienduri:


qedawkins (changes requested):
I took a look at the semantics of tensor.reshape and it looks like it takes a 1D tensor to indicate the shape: https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorreshape-tensorreshapeop

This supports unranked tensors, but we shouldn't really need that here. You can just construct the reshape target shape with `tensor.from_elements` based on the shape list values (requiring a static list length; we require ranked tensors in torch-mlir at the moment anyway so this is reasonable).

Also I'm now remembering why we didn't do this before with tensor.collapse_shape -> tensor.expand_shape. Currently tensor.expand_shape doesn't take either a destination or a list of shape values and thus is incapable of representing reshapes like `?x?x?` -> `?x?`.

saienduri:
Already implemented with collapse, expand

PR #2515, opened 2023-10-16
Adding reshape fall back case for unimplemented cases in aten view for time being

saienduri:


lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -330,19 +330,27 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+    Value sizeTensor = rewriter.create<tensor::FromElementsOp>(loc, RankedTensorType::get({}, sizeType), op.getSize())
+              .getResult();
 
     // TODO: add support for case inputRank 0 expanded to size 1
     if (inputRank == 0)
-      return rewriter.notifyMatchFailure(
-          op, "unimplemented: input rank 0 is not supported");
+      // return rewriter.notifyMatchFailure(
+      //     op, "unimplemented: input rank 0 is not supported");
+      rewriter.replaceOpWithNewOp<tensor::ReshapeOp>(op, resultType, input, sizeTensor);
```
ramiro050, line 342: `tensor::ReshapeOp` is not a good op to lower to. See discussion here: https://github.com/llvm/torch-mlir/pull/1168

Do you have examples of the cases you need to handle that currently fail? I can help brainstorm approaches

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -330,19 +330,27 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+      return success();
```
ramiro050, line 343: The tests are failing because this return is not part of the `if` above (`if` needs braces)

lib/Conversion/TorchToLinalg/DataMovement.cpp
saienduri: Others (Gaurav, Mahesh) are working on another approach with collapsing and expanding. This is just a short term fix until we get that patch landed. Specifically, we are trying to lower llama and currently blocked. I'll add the specific issue to this thread once I get it.

lib/Conversion/TorchToLinalg/DataMovement.cpp
saienduri: Thank you! (didn't catch that silly mistake)

lib/Conversion/TorchToLinalg/DataMovement.cpp
dan-garvey:
```
Llama_2_7b_chat_hf_torch.mlir:958:10: error: failed to legalize operation 'torch.aten.view' that was explicitly marked illegal
    %5 = torch.aten.view %3, %4 : !torch.vtensor<[1,?],si64>, !torch.list<int> -> !torch.vtensor<[1,?],si64>
         ^
Llama_2_7b_chat_hf_torch.mlir:958:10: note: see current operation: %55 = "torch.aten.view"(%53, %54) : (!torch.vtensor<[1,?],si64>, !torch.list<int>) -> !torch.vtensor<[1,?],si64>
Llama_2_7b_chat_hf_torch.mlir:12546:10: error: failed to legalize operation 'torch.aten.view' that was explicitly marked illegal
    %6 = torch.aten.view %4, %5 : !torch.vtensor<[1,?],si64>, !torch.list<int> -> !torch.vtensor<[1,?],si64>
         ^
Llama_2_7b_chat_hf_torch.mlir:12546:10: note: see current operation: %69 = "torch.aten.view"(%67, %68) : (!torch.vtensor<[1,?],si64>, !torch.list<int>) -> !torch.vtensor<[1,?],si64>
```

I think this is the same issue that I solved locally by adding that logic that you were asking about in #2485

dan-garvey: its still weird to me that this fails. Let me double check I'm on main too

dan-garvey: yeah does still happen ToM,

https://storage.googleapis.com/shark_tank/dan/repro.mlir
https://storage.googleapis.com/shark_tank/dan/repro_readable.mlir

iree-compile --iree-input-type=torch repro.mlir

ramiro050: Do you have a small reproducer of the error?

ramiro050: > This is just a short term fix until we get that patch landed.

`tensor.reshape` is not a good op to lower to because it just pushes further down the stack the actual analysis of figuring out what is expanding and what is contracting. AFAICT, `tensor.reshape` is not supported by IREE (I'm assuming you're trying to run llama using IREE).

ramiro050: > yeah does still happen ToM,
> 
> https://storage.googleapis.com/shark_tank/dan/repro.mlir https://storage.googleapis.com/shark_tank/dan/repro_readable.mlir
> 
> iree-compile --iree-input-type=torch repro.mlir

Do you have a small reproducer? (a single MLIR function with a single call to the `torch.view` causing problems)

MaheshRavishankar: @ramiro050 these are all good points. I agree that `tensor.reshape` is just bumping the problem down the stack, but this is blocking things downstream. So here is my proposal
1) We use the lowering to `tensor.reshape` for the time being.
2) To make this work in IREE, we need a lowering from `tensor.reshape` to `flow.tensor.reshape`. The semantics of the latter is pretty similar to the former, so that should be straigh-forward.
3) We then try to land https://github.com/llvm/llvm-project/pull/69267 (I am reviewing it right now, sorry it fell through the cracks a few times, but I am looking at it right now). With that we can come and fix up the lowering here.  We might actually not need to do that either. Downstream from here (like in IREE), we could convert `tensor.reshape` to `tensor.collapse_shape` + `tensor.expand_shape` to represent the `tensor.reshape`.... 

Mainly since (3) is going to take some time, (1) and (2) will unblock us for now.

ramiro050: It would be good to know what is blocking things downstream. @dan-garvey mentioned above that the issue might be fixed by making a small change to how the output shape is calculated. If so, this would be preferred to adding a new op that downstream projects have to handle. A small reproducer of what needs solving would be very useful here.

If the issue turns out to be very hard to handle given the current setup, then the plan you've outlined makes sense.

dan-garvey: I'll aim to have that up for re"view" today

ramiro050: Perfect! Thanks @dan-garvey

PR #2520, opened 2023-10-18
Add flags to CMake build that allow for StableHLO to be provided externally

sogartar:


stellaraccident (commented):
I think I would have done this a bit differently: just have a cache variable for EXPLICIT_STABLEHLO_SOURCE_DIR. If that is defined and not empty, use it. Otherwise, user a default in tree path.

sogartar:
This is available in this PR, but there is also the option to let the user entirely specify all required targets. I want this since there is a use case where multiple projects what to shared the same StableHLO in one build.

PR #2524, opened 2023-10-20
add uncovered cases in view lowering

dan-garvey:
removes unecessary checks from empty strided


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4434,13 +4436,6 @@ class DecomposeAtenEmptyStridedOp
   LogicalResult matchAndRewrite(AtenEmptyStridedOp op,
                                 PatternRewriter &rewriter) const override {
     SmallVector<int64_t> sizeListInts, strideListInts;
-    if (!matchPattern(op.getSize(), m_TorchListOfConstantInts(sizeListInts)))
```
ramiro050, line 4437: These are necessary. The checks below expect `strideListInts` and `sizeListInts` to be populated

ramiro050 (changes requested):

ramiro050 (changes requested):
I made the comment last week but forgot to submit the review lol

Also, do you have an example of the case being handled that was not working before? (if it's not possible to get a PyTorch test because it only fails when tracing, an MLIR example is okay)

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -467,16 +478,34 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+            outputSliceIndices.push_back(idx++);
+	  }
+	  if (idx < outputShapeSlice.size() && outputShapeSlice[idx] == kUnknownSize) {
+	    checkDimEqualHelper(rewriter, loc, inputSize[inputDim], 
+			    outputSizeInt[idx]);
+            inputShape[inputDim] = outputShape[idx];
+            outputSliceIndices.push_back(idx);
+            inputSliceIndices.push_back(0);
+            assumedDynamicDimNotSplit = true;
+	  }
```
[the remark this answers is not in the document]
dan-garvey: Yeah its weird. lines 489 and 503 were particularly key in making it work. The output shape was getting mangled by mapStaticallyKnownDims iirc. I was also unable to create a pytorch test that could fail it. But the tracing fails. I'll add a lit test.

test/Conversion/TorchToLinalg/view.mlir
```diff
@@ -73,4 +73,21 @@ func.func @torch.aten.view$expandInferredDim(%arg0: !torch.vtensor<[2,6],f32>) -
...
+// CHECK: %[[COLLAPSE_CAST:.*]] = tensor.cast %[[COLLAPSED]] : tensor<?x4096xf32> to tensor<?x4096xf32>
+// CHECK: %[[BUILTIN_TENSOR_CAST:.*]] = torch_c.from_builtin_tensor %cast_3 : tensor<?x4096xf32> -> !torch.vtensor<[?,4096],f32>
+// CHECK: return %[[BUILTIN_TENSOR_CAST]] : !torch.vtensor<[?,4096],f32>
+
+func.func @torch.aten.view$collapseUnitDim(%arg0: !torch.vtensor<[1,?],si64>, %arg1: !torch.vtensor<[1,?,4096],f32>) -> !torch.vtensor<[?,4096],f32> {
+  %int1 = torch.constant.int 1
+  %0 = torch.aten.size.int %arg0, %int1 : !torch.vtensor<[1,?],si64>, !torch.int -> !torch.int
+  %int4096 = torch.constant.int 4096
+  %1 = torch.prim.ListConstruct %0, %int4096 : (!torch.int, !torch.int) -> !torch.list<int>
+  %2 = torch.aten.view %arg1, %1 : !torch.vtensor<[1,?,4096],f32>, !torch.list<int> -> !torch.vtensor<[?,4096],f32>
```
ramiro050, line 91: The issue here is not that we are not properly handling trailing and leading size-1 dimensions, but that we are not properly detecting the unchanged dynamic dimension. I think it would make more sense to do something like:

```diff
@@ -377,6 +377,11 @@ public:
     // collapsed. Note this may technically not always be true.
     // TODO: think of a way better way to at least detect when this assumption
     // is violated for the cases of dynamic dimensions.
+    bool inputHasOneDynDim = llvm::count(inputShape, kUnknownSize) == 1;
+    bool outputHasOneDynDim = llvm::count(outputShape, kUnknownSize) == 1;
+    bool singleDynDimsAreEqual =
+        inputHasOneDynDim && outputHasOneDynDim &&
+        productReduce(inputShape) == productReduce(outputShape);
     SmallVector<std::pair<int64_t, int64_t>> unchangedDims;
     for (auto [outputDim, outputDimSize] :
          llvm::enumerate(outputSizeTorchInt)) {
@@ -385,6 +390,14 @@ public:
       if (matchPattern(outputDimSize,
                        m_TorchTensorSizeInt(op.getSelf(), &inputDim))) {
         unchangedDims.push_back(std::make_pair(inputDim, outputDim));
+      } else if (singleDynDimsAreEqual &&
+                 outputShape[outputDim] == kUnknownSize) {
+        // If the input and output have a single dynamic dimension and the
+        // product of the other dimensions is the same, then we know that the
+        // dynamic dimension is unchanged.
+        inputDim = std::distance(inputShape.begin(),
+                                 llvm::find(inputShape, kUnknownSize));
+        unchangedDims.push_back(std::make_pair(inputDim, outputDim));
       }
     }
     // Mark the end of the input/output shapes
```

ramiro050 (changes requested):

test/Conversion/TorchToLinalg/view.mlir
```diff
@@ -73,4 +73,20 @@ func.func @torch.aten.view$expandInferredDim(%arg0: !torch.vtensor<[2,6],f32>) -
...
+  }
+
+// CHECK-LABEL: func.func @torch.aten.view$collapseUnitDim(
+// CHECK: %[[CASTED:.*]] = tensor.cast %1 : tensor<1x?x4096xf32> to tensor<1x?x4096xf32>
+// CHECK: %[[COLLAPSED:.*]] = tensor.collapse_shape %[[CASTED]] {{\[\[}}0, 1], [2]] : tensor<1x?x4096xf32> into tensor<?x4096xf32>
+// CHECK: %[[COLLAPSE_CAST:.*]] = tensor.cast %[[COLLAPSED]] : tensor<?x4096xf32> to tensor<?x4096xf32>
+// CHECK: %[[BUILTIN_TENSOR_CAST:.*]] = torch_c.from_builtin_tensor %cast_3 : tensor<?x4096xf32> -> !torch.vtensor<[?,4096],f32>
+// CHECK: return %[[BUILTIN_TENSOR_CAST]] : !torch.vtensor<[?,4096],f32>
+
+func.func @torch.aten.view$collapseUnitDim(%arg0: !torch.vtensor<[1,?],si64>, %arg1: !torch.vtensor<[1,?,4096],f32>) -> !torch.vtensor<[?,4096],f32> {
```
ramiro050, line 85: This case can be expressed as an e2e test. The test below fails without these changes and passes with them:

```python
class ViewSizeFromOtherTensor(torch.nn.Module):
    def __init__(self):
        super().__init__()

    @export
    @annotate_args([
        None,
        ([1, -1], torch.float32, True),
        ([1, -1, 10], torch.float32, True),
    ])

    def forward(self, x, y):
        return torch.ops.aten.view(y, (torch.ops.aten.size(x, 1), 10))

@register_test_case(module_factory=lambda: ViewSizeFromOtherTensor())
def ViewSizeFromOtherTensor_basic(module, tu: TestUtils):
    module.forward(tu.rand(1, 7), tu.rand(1, 7, 10))
```

Let's use the e2e one and remove this one.

ramiro050 (approved):
Other than the test changes, this LGTM

PR #2537, opened 2023-10-28
Rework AtenEmptyStridedOp checks

dan-garvey:
Now using Value instead of Ints. Trades compile failure for a runtime assert


lib/Dialect/Torch/Utils/Utils.cpp
```diff
@@ -333,3 +333,25 @@ bool Torch::isAssumingStrictSymbolicShapes(Block *block) {
   }
   return false;
 }
+
+void Torch::checkDefaultStrideHelper(PatternRewriter &rewriter,
+                                     SmallVector<Value> sizeListValues,
+                                     SmallVector<Value> strideListValues,
```
qedawkins, line 339: You can use `SmallVector<OpFoldResult>` instead of values, and then use `rewriter.createOrFold` calls to create the ops down below and evaluate the assert at compile time when everything is static.

qedawkins (changes requested):
A large portion of the changes here look like something is wonky with the linter and it makes it a harder to review. Any idea what's going on here?

lib/Dialect/Torch/Utils/Utils.cpp
dan-garvey: Oh cool. I'll do that.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -404,16 +405,15 @@ class DecomposeAtenGluOp : public OpRewritePattern<AtenGluOp> {
...
 public:
   using OpRewritePattern::OpRewritePattern;
   LogicalResult matchAndRewrite(AtenZeroOp op,
                                 PatternRewriter &rewriter) const override {
     Value zero = rewriter.create<ConstantIntOp>(op.getLoc(),
                                                 rewriter.getI64IntegerAttr(0));
-    rewriter.replaceOpWithNewOp<AtenFillScalarOp>(op, op.getType(), op.getSelf(),
-                                                  zero);
+    rewriter.replaceOpWithNewOp<AtenFillScalarOp>(op, op.getType(),
+                                                  op.getSelf(), zero);
```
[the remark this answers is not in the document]
dan-garvey: I think it was just not applied to this file previously or something

lib/Dialect/Torch/Utils/Utils.cpp
dan-garvey: @qedawkins so I tried changing those SmallVectors but there's undefined conversions from Value to OpFoldResult that make that and several other things in this function not work. I made as many things as I could as "createOrFold", but I'm not clear on how that gets handled exactly. I suppose one thing I could do is forcibly fail a test, get the ir, correct it, and then see if it folds away.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4433,37 +4451,51 @@ class DecomposeAtenNewEmptyStridedOp
...
-    for (unsigned i = 0; i < strideListInts.size(); i++) {
-      int64_t defaultStride = 1;
-      for (unsigned j = i + 1; j < sizeListInts.size(); j++)
-        defaultStride *= sizeListInts[j];
-      if (defaultStride != strideListInts[i]) {
-        isDefaultStride = false;
-        break;
+    if (matchPattern(op.getSize(), m_TorchListOfConstantInts(sizeListInts)) &&
+        matchPattern(op.getStride(),
+                     m_TorchListOfConstantInts(strideListInts))) {
```
qedawkins, line 4456: You could probably just take `op.getSize()` and `op.getStride()` directly in the helper and do the matching there. That way you don't have to duplicate this code in multiple places. You'll need to return `LogicalResult` from the helper if you do that though because it would then be possible for the helper to fail.

qedawkins (approved):

PR #2540, opened 2023-11-01
[Torch] Fix mixP case for non value semantic ops

zhekunz2:
NonValueSemantic Ops like Add_, div_, etc. expect result DType to be the same as the first input. However, current implementation would result in wrong result type for case like:

```python
a = torch.randn(3, 3).half() # float16
b = torch.randn(3, 3) # float32
a += b # i.e. torch.ops.aten.add_(a, b)
```
torch expects `a` to be float16, but dtype refinement would infer float32 type, since it's replaced by `aten.add`.

### DIFF
--- a/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
+++ b/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
@@ -243,6 +243,13 @@ public:
            "Torch JIT operators shouldn't have regions or successors");
 
     Operation *newOp = rewriter.create(state);
+    // Note: need to convert result to first input's dtype because mix precision
+    // compute would result in different behaviors.
+    // For example: 
+    // a = torch.randn(3, 3).half() # float16 
+    // b = torch.randn(3, 3) # float32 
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -243,8 +243,13 @@ class ReduceTrailingUnderscoreInplaceVariant : public RewritePattern {
            "Torch JIT operators shouldn't have regions or successors");
 
     Operation *newOp = rewriter.create(state);
-    auto tensor =
-        rewriter.create<CopyToValueTensorOp>(op->getLoc(), newOp->getResult(0));
+    Value none = rewriter.create<ConstantNoneOp>(op->getLoc());
```
ramiro050, line 246: can you add a small comment here explaining the reasoning

ramiro050 (approved):
Nice catch! Kinda crazy that it took this long for this issue to surface.

qingyunqu (approved):

PR #2541, opened 2023-11-01
Add torch vision to requirements.txt

newling:
Before this PR: following the instructions in torch-mlir/docs/development.md and running the example in the section "Testing MLIR output in various dialects" hits an error about missing package.


powderluv:
btw we removed this to avoid packaging it into the nightly build

newling:
> btw we removed this to avoid packaging it into the nightly build

Ok, I'll abandon this and maybe just add a comment in development.md

PR #2543, opened 2023-11-02
[Torch] Add fold rule for AtenMaskedFillTensorOp to AtenMaskedFillScalarOp

zhekunz2:


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -162,6 +162,41 @@ static Value getScalarIntValue(Value input, Location loc,
...
+    return nullptr;
+
+  std::optional<unsigned> inputRank = getTensorRank(input);
+  if (!inputRank || *inputRank != 0)
+    return nullptr;
+
+  if (auto valueTensorLiteralOp = input.getDefiningOp<ValueTensorLiteralOp>()) {
+    auto val = valueTensorLiteralOp.getValue()
+                   .cast<DenseElementsAttr>()
+                   .getSplatValue<float>();
```
ramiro050, line 188: I think this should be `double`, since the input type can be an `f64`

qingyunqu: I think `getSplatValue<APFloat>().getValueAsDouble()` is more general for different float type.

qingyunqu (approved):
LGTM

qingyunqu:
@ramiro050 Could you please review this PR again?

ramiro050 (approved):
Thanks!

PR #2544, opened 2023-11-02
[Torch Dialect] support aten.isinf

qingyunqu:
Also fix linalg lowering from `UEQ` to `OEQ`.  
I will check other comparison's lowering later.


ramiro050 (approved):
Thanks!

PR #2550, opened 2023-11-03
Decomposition of aten.pixel_shuffle with static input shape

newling:
For static tests (that is when the shape is know) for example:

 ```
 @annotate_args([None, ([3, 18, 2, 2], torch.float32, True)])
 ```
 
 The e2e passes. But only if the replacement op's return type is set as undefined (optional shape and type must be explicitly made unset), otherwise there's a error about the function return type. 
 
 For dynamic cases, for example if the above is replaced with 
 
  ```
 @annotate_args([None, ([-1, -1, -1, -1], torch.float32, True)])
 ```

There is a failure to lower to linalg from torch ("view op explicitly labelled as illegal"). This seems to be because the support for lowering from torch to linalg with dynamic shapes is limited.


newling:
@vivekkhandelwal1 I have an e2e test case working, but only for static shapes (see my comment above). I am not sure what to do about the dynamic shape case, is my thought that torch->linalg for aten.view is limited, and that's why I'm having problems? 

More generally, I'd like to understand dynamic shapes end-to-end in torch-mlir, any pointers?

vivekkhandelwal1:
> @vivekkhandelwal1 I have an e2e test case working, but only for static shapes (see my comment above). I am not sure what to do about the dynamic shape case, is my thought that torch->linalg for aten.view is limited, and that's why I'm having problems?
> 
> More generally, I'd like to understand dynamic shapes end-to-end in torch-mlir, any pointers?

Hi @newling, the support for dynamic shapes in Torch-MLIR is limited, especially for view-like ops. I think, for now, you can follow either of the below two paths:
1.) Add a condition in your decomp restricting the support for only static shapes (Easy and fast path, but limited support).
2.) Directly add a `Torch -> Linalg` lowering for this op, if it seems feasible (Hard and slow path, but you will get complete support for the op).

For the `Torch -> Linalg` support you can refer to this: https://github.com/pytorch/pytorch/blob/fa9045a8725214c05ae4dcec5a855820b861155e/aten/src/ATen/native/cpu/PixelShuffleKernel.cpp#L16-L53

Also, unfortunately, there is no documentation for the dynamic shapes in Torch-MLIR. It's just that ops in Torch-MLIR broadly fall into one or a combination of the below four categories:
1.) Full-Dynamic Support
2.) Partial-Dynamic Support
3.) Full-Static Support
4.) Partial-Static Support.

You can go through some of the op lowerings for view-like ops to understand the functioning or dynamic-shape support in Torch-MLIR.

newling:
Thanks for these information gems @vivekkhandelwal1 

I am tempted to go with option 1 (just support static) for now, and leave comments in the code about what the issues currently are with dynamic shapes, and what possible remedies there are (like go directly to linalg, or maybe extend lowering of reshape). 

Hopefully the use cases that @qedawkins  has in mind are static.

vivekkhandelwal1:
> Thanks for these information gems @vivekkhandelwal1
> 
> I am tempted to go with option 1 (just support static) for now, and leave comments in the code about what the issues currently are with dynamic shapes, and what possible remedies there are (like go directly to linalg, or maybe extend lowering of reshape).
> 
> Hopefully the use cases that @qedawkins has in mind are static.

If it's required for Llama-2 then the use case would probably be dynamic. But @qedawkins can comment more on this,

qedawkins:
> > Thanks for these information gems @vivekkhandelwal1
> > I am tempted to go with option 1 (just support static) for now, and leave comments in the code about what the issues currently are with dynamic shapes, and what possible remedies there are (like go directly to linalg, or maybe extend lowering of reshape).
> > Hopefully the use cases that @qedawkins has in mind are static.
> 
> If it's required for Llama-2 then the use case would probably be dynamic. But @qedawkins can comment more on this,

Ah no, this mostly shows up in vision models, and I would guess most of the time we would see static dims for the ones that are actually shuffled. There might be some dynamic outer/batch like dims though.

newling:
Ok, so I've got it supporting just static shapes. I've left a comment in the code about why, and possible next steps. 

I don't seem to be able to add reviewers, and I also need approval to start the workflows. How I get these permissions?

qedawkins:
> Ok, so I've got it supporting just static shapes. I've left a comment in the code about why, and possible next steps.
> 
> I don't seem to be able to add reviewers, and I also need approval to start the workflows. How I get these permissions?

Good question. IIRC there is a process to be added to the repo after you have landed a commit, but I don't know if the process has changed.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1092,9 +1096,223 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+    // the semantics of pixel_shuffle, by the time the lowering sees the view
+    // ops, this special context is lost.
+    //
+    // There are a few possible options to fixing the current approach, as I see
+    // it:
+    //
+    // 1) Lower pixel_shuffle to linalg directly. This would mean a bit of
+    // duplication of the logic for lowering from aten.permute to linalg, and
+    // would not be the most 'gradual' lowering approach. But perhaps the most
+    // straightforward and understable.
```
qedawkins, line 1156: There are likely other benefits of doing this. In particular, the kind of reshape done by this operation is relatively structured. This makes the analysis needed on the reshape ops to recover the optimal expand/collapse_shape ops much easier (especially in dynamic cases).

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1092,9 +1096,223 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+    //
+    // 2) Rather than use aten.reshape ops in the decomposition below, insert
+    // tensor.expand_shape and tensor.collapse_shape ops directly. This would
+    // require inserting unrealized_conversion_cast ops to convert the types of
+    // the operands from !torch.tensor to to the MLIR built-in tensor type,
+    // which isn't ideal (unrealized_conversion_cast should only be used in
+    // conversion passes AFAIK (?)).
```
qedawkins, line 1163: Torch-mlir cannot decompose to a tensor operation. One important part of torch-mlir is the ability to lower to different target IRs (or even mix-and-match), thus all torch decompositions must be strictly in terms of torch ops.

In regards to `builtin.unrealized_conversion_cast` though, there is a `torchc.to_builtin` and `torchc.from_builtin` (might have gotten the names wrong) that allow conversion between the builtin tensor type and the torch tensor type.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1092,9 +1096,223 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+    //
+    // 3) Create 2 new ops in the torch dialect, with the same semantics as
+    // tensor.expand_shape and tensor.collapse_shape, but with !torch.tensor
+    // operands and results. Then use these ops in the decomposition below,
+    // instead of aten.reshape ops.
```
qedawkins, line 1168: It would definitely interesting to have something along the lines of a `TorchExt` extension of the torch dialect (or just a set of add on operations) that torch operators can decompose to. This could help unify lowerings for particularly complicated ops by first decomposing them a bit if there is no good choice already in ATen.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1092,9 +1096,223 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+    }
+    auto inShape = maybeSizes.value();
+    auto inRank = inShape.size();
+
+    // TODO: Handle dynamic shapes.
+    //
+    // Currently the decomposition of inputs with dynamic shapes results in a
+    // failure when lowering to linalg. Specifically, the error:
+    //
+    // "Unhandled case of expand/collapse. "
```
qedawkins, line 1136: There is a WIP patch to help with these unhandled dynamic cases (requires some upstream changes to tensor.expand_shape), but it will still generate somewhat non-optimal IR by essentially doing a full collapse -> full expand.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1092,9 +1096,223 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+         permutation);
+
+     Value B = rewriter.create<AtenPermuteOp>(
+         loc, getTypeFromShape(postPermuteShape), A, permuteDimsOrder);
+
+     Value outShapeList =
+         rewriter.create<PrimListConstructOp>(loc, listType, outShape);
+
+     // TODO(jn) figure out why the type of the returned value must be made
+     // undefined.
```
qedawkins, line 1295: Can you elaborate a bit on what was going wrong here?

qedawkins (commented):
Mostly looks good, thanks for the patch! It has been a while since I've worked in depth with these decompositions and there might be some helpers already in place to help simplify code in a few places (looking for another reviewer who might know).

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -227,8 +228,7 @@ class DecomposeAtenAmaxOp : public OpRewritePattern<AtenAmaxOp> {
     }
     // For every dimension included in `dim` of the op, iterated over in
     // reverse order, we create a call to aten.max.dim.
-    std::sort(dims.begin(), dims.end());
-    std::reverse(dims.begin(), dims.end());
+    std::sort(dims.rbegin(), dims.rend());
```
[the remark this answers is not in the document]
qedawkins: (nothing actionable here, just a note)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
qedawkins: Ah, I read the earlier part of the comment more closely. Yes, +1 to this approach.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
newling: Makes sense, so this isn't an option because we it'd break the lowering to tosa, for example. Good to know that about the to_builtin and from_builtin.  Out of curiosity, do you know why torch-mlir decided against using the builtin tensor type to represent a value semantics tensor?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
newling: I'd be happy to explore this as a follow-up task if other reviewers don't see any problem with this approach. One concern I have is that introducing a new "concept" to torch-mlir like this might make the lowering paths less "obvious".

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
newling: Yeah. In a way it is good that the full collapse path isn't in torch-mlir yet -- if it was this PR would have just worked with dynamic shapes and we would have missed the possible better loop nesting.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1072,9 +1075,171 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+    auto maybeSizes = inType.getOptionalSizes();
+    if (!maybeSizes) {
+      return rewriter.notifyMatchFailure(
+          op, "Expected input tensor to have known rank.");
+    }
+    auto inShape = maybeSizes.value();
+    auto inRank = inShape.size();
+
+    // At least 3 dimensions are needed
+    // (case when leading_dims is empty).
```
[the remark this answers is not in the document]
newling: I'll update the comment in the code as:

The input tensor must have at least 3 dimensions: (1) the channel dimension which gets smaller by 'factor*factor', (2) the H channel which gets larger by 'factor' and (3) the W channel which get larger by 'factor'. The total number of dimensions is 3 + N, where N is the number of leading dimensions (N >= 0).

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
newling: Below, "finalShape" is  `!torch.vtensor<[3,4,6],si64>`.

But I think "op" has type `!torch.vtensor` (?), and so if I replace 

` rewriter.replaceOpWithNewOp<AtenReshapeOp>(op, finalType, B, outShapeList);`

in the code with 

`rewriter.replaceOpWithNewOp<AtenReshapeOp>(op, finalShape, B, outShapeList);`

I get the error:

```
        error: unknown: type of return operand 0 ('!torch.vtensor<[3,4,6],si64>') doesn't match function result type ('!torch.vtensor') in function @forward
        note: unknown: see current operation: "func.return"(%11) : (!torch.vtensor<[3,4,6],si64>) -> ()
```

Sorry, my variable names "finalShape" and "finalType" are bad.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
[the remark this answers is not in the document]
newling: That makes sense, noted for next time 👍

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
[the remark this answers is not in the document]
newling: (The leading dims can be empty)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
qedawkins: Ah, so the reason for this is that the rewrite that you do on this operation must replace all original uses of `aten.pixel_shuffle` with values of the same type. Like you said, the op has type `!torch.vtensor`, so we see something like this
```
%0 = torch.aten.pixel_shuffle ... -> !torch.vtensor
%1 = "some_user" %0 : !torch.vtensor -> !torch.int
```
If you don't use the same type, the resulting IR will look like this
```
%0 = torch.aten.pixel_shuffle ... -> !torch.vtensor<[2, 3, 6]xf32>
%1 = "some_user" %0 : !torch.vtensor -> !torch.int
```
and thus there is a type mismatch between %0 and it's user, %1. This is because all operations carry the types of their operands, and you can't change the type of an op operand on the fly.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1092,9 +1096,223 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+    // the operands from !torch.tensor to to the MLIR built-in tensor type,
+    // which isn't ideal (unrealized_conversion_cast should only be used in
+    // conversion passes AFAIK (?)). UPDATE: not a viable option as we cannot
+    // make assumptions about the dialects that torch is lowered to.
+    //
+    // 3) Create 2 new ops in the torch dialect, with the same semantics as
+    // tensor.expand_shape and tensor.collapse_shape, but with !torch.tensor
+    // operands and results. Then use these ops in the decomposition below,
+    // instead of aten.reshape ops.
+    //
```
qedawkins, line 1170: Also, typically long TODO comments like this are best for github issues rather than inline in the code. Can you keep a short-form of `1)` as a TODO comment and move the rest to an issue? For example
```
// TODO: Lower pixel_shuffle to linalg directly. Pixel shuffle does a static reshape that is hard to recover through pure torch (view) ops, especially in dynamic cases.
```

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "PassDetail.h"
 
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
```
qedawkins, line 12: nit: What do you need to use from Tensor?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1092,9 +1096,223 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+     Value B = rewriter.create<AtenPermuteOp>(
+         loc, getTypeFromShape(postPermuteShape), A, permuteDimsOrder);
+
+     Value outShapeList =
+         rewriter.create<PrimListConstructOp>(loc, listType, outShape);
+
+     // TODO(jn) figure out why the deduced return type (like
+     // !torch.vtensor<[3,4,6],si64>) cannot be used for the type of the
+     // replacement of the pixel_shuffle op's output. The pattern seems to
+     // require the generic !torch.vtensor type.
```
qedawkins, line 1300: I think the reason you need to cast back to an unranked tensor is because you are missing a shape conversion function here: https://github.com/llvm/torch-mlir/blob/main/projects/pt1/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py

There should be an entry for `aten.pixel_shuffle` in `include/torch-mlir/Dialect/Torch/IR/JITOperatorRegistryDump.txt` that has shape/dtype/(decomposition) signature functions that you can add. Here is an example for aten._fw_primal
```
  shape_function_signature = def aten〇_fw_primal〡shape(self: List[int], level: int) -> List[int]:
  decomposition_function_signature = def aten〇_fw_primal〡decomposition(self: Tensor, level: int) -> Tensor:
  dtype_function_signature = def aten〇_fw_primal〡dtype(self_rank_dtype: Tuple[int, int], level: int) -> int:

```

qedawkins (changes requested):
I left the comments that still need addressing as unresolved. Mainly some comment cleanup and missing shape/dtype refinement functions (sorry for the segmented review).

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
qedawkins: I think it was some attempt to reflect the more general programming model that pytorch offers, and since a non-value semantics tensor was going to be needed anyway, I'm guessing the idea was that swapping between builtin types and a torch type would make things more difficult. I wasn't around for those design decisions though :)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
newling: Nice, I was wondering where to implement the type inference. That fixes the problem.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
newling: Removed (was experimenting with unrealized_conversion to use expand_shape and collapse_shape)

qedawkins (approved):
LGTM! Thanks for pushing this through

qedawkins:
I think some linting check wants you to run
```
./build_tools/update_abstract_interp_lib.sh
```

newling:
Thanks for the help @qedawkins! 

> I think some linting check wants you to run
> 
> ```
> ./build_tools/update_abstract_interp_lib.sh
> ```

Yeah, that was a very useful diagnostic message from CI 😍

newling:
@qedawkins would you mind kickstarting CI? It was failing with an "unexpected pass" failure, so I added the 2 tests to projects/pt1/e2e_testing/xfail_sets.py (a bit of a shot in the dark) 

General question: Is there a simple way to run all the tests locally, with a single command?

qedawkins:
There are some notes on testing here: https://github.com/llvm/torch-mlir/blob/main/docs/development.md#testing
(I don't remember off the top of my head what the command you're looking for is)

newling:
There isn't much material about updating xfail_sets.py... but I'm optimistic I've got it right now (added to LTC_XFAIL_SET). Can you please restart the workflow @qedawkins  ?

qedawkins:
> There isn't much material about updating xfail_sets.py... but I'm optimistic I've got it right now (added to LTC_XFAIL_SET). Can you please restart the workflow @qedawkins ?

Sorry about needing approval each time to run the CI :(
Let's get this in and get you direct access

projects/pt1/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -412,6 +412,23 @@ def aten〇sum〇dim_IntList〡shape(self: List[int], dim: Optional[List[int]],
...
+    upscale_factor_squared = upscale_factor * upscale_factor
+    assert self[-3] % (upscale_factor_squared) == 0, "number of input channels  must be divisible by upscale_factor^2 in pixel_shuffle"
+    
+    out = self[0:-3]
+    out.append(self[-3] // upscale_factor_squared)
+    out.append(self[-2] * upscale_factor)
+    out.append(self[-1] * upscale_factor)
+    return out
+
+def aten〇pixel_shuffle〡dtype(self_rank_dtype: Tuple[int, int], upscale_factor: int) -> int:
```
ramiro050, line 427: nit: all dtype functions here are tested e2e. Can you add a test to this one? It should be a matter of adding something like `@check_dtype_function(_check_tensors_with_the_same_dtype(num_of_tensors=1, upscale_factor=1))`

projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -655,6 +656,39 @@ def TransposeIntNegDimsModule_basic(module, tu: TestUtils):
...
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([None, ([3, 18, 2, 2], torch.float32, True)])
+    def forward(self, x):
+        return torch.ops.aten.pixel_shuffle(x, 3)
+
+@register_test_case(module_factory=lambda: PixelShuffleModuleStatic_3_18_2_2())
+def PixelShuffleModuleStatic_3_18_2_2_basic(module, tu: TestUtils):
```
ramiro050, line 669: nit: no need to write the shape in the test name. I think the important bit here is the input dtype, so I would call the tests something like `PixelShuffleModuleStaticInt64` and `PixelShuffleModuleStaticFloat32`

ramiro050 (commented):
Some small requests

projects/pt1/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
ramiro050: You can run the tests by generating the shape and dtype library: `./build_tools/update_abstract_interp_lib.sh`

projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
newling: Initially my motivation for the second test was for the rank-3 case (no leading dimensions). Making it a different dtype was just an extra, to get coverage in a different direction.  I've gone with names `Rank4Float32` and `Rank3Int64`

projects/pt1/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
newling: Done!

qedawkins:
(Merging because it looks like Ramiro's concerns were addressed and there is no PR blocker in place)

projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
ramiro050: Makes sense, thanks!

PR #2551, opened 2023-11-04
Add verification for torch permute op

newling:
- adds support for an optional verifier to the generated torch op tablegen (GeneratedTorchOps.td)
- uses the above to add a verifier for the torch permute op. 

Motivation: I hit an unclear error from linalg while developing a decomposition pass for pixel_shuffle. The error would have been clearer if the problem had been detected in the invalid  aten.permute op. 

Testing: new tests added. To run added tests, from the base directory run

```
 ./build/bin/llvm-lit  test/Dialect/Torch/invalid.mlir
 ```

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -2859,6 +2859,94 @@ LogicalResult ShapeCalculateYieldShapesOp::verify() {
   return success();
 }
 
+LogicalResult AtenPermuteOp::verify() {
+
+  // Verification of the permute op for input & output dimensions with
+  // statically known sizes.
+
+  SmallVector<Value> permutation;
+  auto permutationObtained = getListConstructElements(getDims(), permutation);
+  if (!permutationObtained) {
+    return success();
+  }
+
+  auto outType = getResult().getType().cast<BaseTensorType>();
+  auto inType = getSelf().getType().cast<BaseTensorType>();
+
+  if (!outType.hasSizes() || !inType.hasSizes()) {
+    return success();
+  }
+
+  auto outShape = outType.getOptionalSizes().value();
+  auto inShape = inType.getOptionalSizes().value();
+
+  auto outRank = outShape.size();
+
+  if (outRank != inShape.size()) {
+    return emitOpError(
+               "expected input and output tensors to have same rank, but ")
+           << inShape.size() << " != " << outRank << '.';
+  }
+
+  if (outRank != permutation.size()) {
+    return emitOpError() << "expected permutation to have size equal result "
+                            "tensor rank. The permutation has "
+                         << permutation.size()
+                         << " elements, the output has rank " << outRank << '.';
+  }
+
+  auto inBounds = [outRank](auto index) {
+    return index >= 0 && static_cast<uint64_t>(index) < outRank;
+  };
+
+  // Initialization of the reverse permutation. -1 denotes an unknown
+  // permutation index.
+  SmallVector<int64_t> reversePermutation(outRank, -1);
+
+  // In this loop:
+  //  (1) check that the permutation indices are in bounds, and not duplicated.
+  //  (2) populate reversePermutation (to check for duplicates).
+  //  (3) check that the input and output shapes agree with the permutation. For
+  //  example, if the permutation is (1,2,0) and the input shape is (2,3,5),
+  //  then the output shape must be (3,5,2).
+
+  for (uint64_t to = 0; to < outRank; ++to) {
+    int64_t from;
+    auto fromIsSet = matchPattern(permutation[to], m_TorchConstantInt(&from));
+
+    // if 'from' is the unkwown index, continue.
+    if (from == -1) {
+      continue;
+    }
+
+    if (fromIsSet) {
+      if (!inBounds(from)) {
+        return emitError("observed invalid index in permutation (")
+               << from << ") for input tensor of rank " << outRank << '.';
+      }
+      if (reversePermutation[from] != -1) {
+        return emitOpError("has a duplicate dimension (")
+               << from << ") in its permutation " << getDims() << '.';
+      }
+      reversePermutation[from] = to;
+
+      auto dimSizesDefined = inShape[from] != -1 && outShape[to] != -1;
+      auto dimSizesDifferent = inShape[from] != outShape[to];
+
+      if (dimSizesDefined && dimSizesDifferent) {
+        return emitOpError("has a permutation which is not compatible with the "
+                           "input and output shapes. ")
+               << "The input shape in dimension " << from << " is "
+               << inShape[from] << ", and the output shape in dimension " << to
+               << " is " << outShape[to]
+               << " : they should be the same with this permutation. ";
+      }
+    }
+  }
+
+  return success();
+}
+
 //===----------------------------------------------------------------------===//
 // DtypeCalculateYieldDtypesOp
 //===----------------------------------------------------------------------===//
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -2877,8 +2877,8 @@ LogicalResult AtenPermuteOp::verify() {
     return success();
   }
 
-  auto outShape = outType.getOptionalSizes().value();
-  auto inShape = inType.getOptionalSizes().value();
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: include/torch-mlir/Dialect/Torch/IR/TorchOps.td [hunk 1]
```diff
@@ -1326,4 +1326,30 @@ def Torch_DtypeCalculateYieldDtypesOp : Torch_Op<"dtype.calculate.yield.dtypes",
   let hasVerifier = 1;
 }
 
+def Torch_AtenPermuteOp : Torch_Op<"aten.permute", [
+    AllowsTypeRefinement,
+    ReadOnly
+  ]> {
+  let summary = "Generated op for `aten::permute : (Tensor, int[]) -> (Tensor)`";
```
newling, line 1333: I just c&p'd this from the generated version. Would need to adjust summary of op.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2838,6 +2838,88 @@ LogicalResult ShapeCalculateYieldShapesOp::verify() {
...
+  //  (2) populate reversePermutation (to check for duplicates).
+  //  (3) check that the input and output shapes agree with the permutation. For
+  //  example, if the permutation is (1,2,0) and the input shape is (2,3,5),
+  //  then the output shape must be (3,5,2).
+
+  for (uint64_t to = 0; to < outRank; ++to) {
+    int64_t from;
+    auto fromIsSet = matchPattern(permutation[to], m_TorchConstantInt(&from));
+    if (fromIsSet) {
+      if (!inBounds(from)) {
```
ramiro050, line 2896: We need to account for the fact that `-1` is a valid value in the permutation list. The case failing in the CI is

```python
def forward(self, x):
    return torch.ops.aten.permute(x, (0, -1, 1))
```

which is valid in PyTorch.

ramiro050 (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
ramiro050: Input: `x = torch.rand(3, 4, 2)`

newling:
Thanks for reviewing @ramiro050 , I'll fix this soon. 

How can I reproduce this locally? 

image

i.e. it seems to pass for me locally. 

More generally, how can I run as many of the CI tests as possible locally before creating a PR. Is this the max set: 


```
cmake --build build --target check-torch-mlir
cmake --build build --target check-torch-mlir-python
./projects/pt1/tools/e2e_test.sh
```

?

newling:
@ramiro050 can you please start CI?

ramiro050:
> How can I reproduce this locally?

Ah, should've started with that. You need to specify the right config:

```python
python -m e2e_testing.main -v -f "PermuteNegativeIndexModule_basic" -c make_fx_tosa
```

I'm not sure if we have a script that runs all the e2e tests. However, you can see all the available configs by doing `python -m e2e_testing.main -h` and calling the script without any filter `-f` (this runs all the tests for a specific config).

lib/Dialect/Torch/IR/TorchOps.cpp
newling: Fixed, and added a test for this case in test/Dialect/Torch/ops.mlir

newling:
@ramiro050 would you mind taking another look at this sometime this week?

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2838,6 +2838,94 @@ LogicalResult ShapeCalculateYieldShapesOp::verify() {
...
+  if (outRank != permutation.size()) {
+    return emitOpError() << "expected permutation to have size equal result "
+                            "tensor rank. The permutation has "
+                         << permutation.size()
+                         << " elements, the output has rank " << outRank << '.';
+  }
+
+  auto inBounds = [outRank](auto index) {
+    return index >= 0 && static_cast<uint64_t>(index) < outRank;
+  };
```
ramiro050, line 2879: nit: use `isValidDim` to be consistent with codebase

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2838,6 +2838,94 @@ LogicalResult ShapeCalculateYieldShapesOp::verify() {
...
+  for (uint64_t to = 0; to < outRank; ++to) {
+    int64_t from;
+    auto fromIsSet = matchPattern(permutation[to], m_TorchConstantInt(&from));
+
+    // if 'from' is the unkwown index, continue.
+    if (from == -1) {
+      continue;
+    }
+
+    if (fromIsSet) {
```
ramiro050, line 2901: nit: return early to avoid indentation

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2838,6 +2838,94 @@ LogicalResult ShapeCalculateYieldShapesOp::verify() {
...
+  //  then the output shape must be (3,5,2).
+
+  for (uint64_t to = 0; to < outRank; ++to) {
+    int64_t from;
+    auto fromIsSet = matchPattern(permutation[to], m_TorchConstantInt(&from));
+
+    // if 'from' is the unkwown index, continue.
+    if (from == -1) {
+      continue;
+    }
```
ramiro050, line 2899: You should check that the match succeeded before doing anything else (Without looking into the implementation of `matchPattern`, we don't know what happens to the `from` value in the case of failure. Moreover, leaving it uninitialized is also a bit dangerous)

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2838,6 +2838,94 @@ LogicalResult ShapeCalculateYieldShapesOp::verify() {
...
+
+  auto outType = getResult().getType().cast<BaseTensorType>();
+  auto inType = getSelf().getType().cast<BaseTensorType>();
+
+  if (!outType.hasSizes() || !inType.hasSizes()) {
+    return success();
+  }
+
+  auto outShape = outType.getOptionalSizes().value();
+  auto inShape = inType.getOptionalSizes().value();
```
ramiro050, line 2860: You can just do `getSizes()`

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2838,6 +2838,94 @@ LogicalResult ShapeCalculateYieldShapesOp::verify() {
...
+        return emitError("observed invalid index in permutation (")
+               << from << ") for input tensor of rank " << outRank << '.';
+      }
+      if (reversePermutation[from] != -1) {
+        return emitOpError("has a duplicate dimension (")
+               << from << ") in its permutation " << getDims() << '.';
+      }
+      reversePermutation[from] = to;
+
+      auto dimSizesDefined = inShape[from] != -1 && outShape[to] != -1;
```
ramiro050, line 2912: For unknown dimensions in tensors, we have `kUnknownSize`. In PyTorch, `-1` is a special value, but in Torch-MLIR-specific types, -1 being used for dynamic dimensions is an implementation detail that should not be relied on.

ramiro050 (approved):
I just have some small requests, but other than that LGTM. I appreciate the thorough testing 🙂

lib/Dialect/Torch/IR/TorchOps.cpp
newling: Good spot, I've fixed this now.

lib/Dialect/Torch/IR/TorchOps.cpp
newling: Ok. Just to check I understand, is `kUnknown` to `TorchTensorType` as `kDynamic` is to `ShapedTensor` ? 

Using dynamic shapes in MLIR is all new to me.

newling:
Thank you for the useful comments @ramiro050 😃

lib/Dialect/Torch/IR/TorchOps.cpp
ramiro050: That sounds right. In torch-mlir, anytime you deal with `BaseTensorType`s (or its two child classes `ValueTensorType`, `NonValueTensorType`), the dynamic dimensions will have value `kUnknownSize`.

ramiro050 (approved):

PR #2557, opened 2023-11-06
Specify path of e2e_test.sh after directory change

newling:
Is there a way to disable some of CI for docs-only PR's?


ramiro050 (approved):
At the moment our CI runs for every change, unfortunately. However, this LGTM

PR #2560, opened 2023-11-07
Fix for unused variable failure when trying to bump torch-mlir in IREE

saienduri:
Due to blob being an unused variable, we are not able to bump torch-mlir in iree. With this PR, we remove this unused variable.


lib/Conversion/TorchToArith/TorchToArith.cpp
```diff
@@ -192,6 +192,7 @@ class ConvertTorchTensorLiteralOp
           auto shapedType =
               RankedTensorType::get(type.getShape(), builtinTensorElemTy);
           AsmResourceBlob *blob = elements.getRawHandle().getBlob();
+          (void)blob;
```
ramiro050, line 195: Let's remove `blob` instead since it is no longer used

ramiro050 (changes requested):

lib/Conversion/TorchToArith/TorchToArith.cpp
saienduri: ok sure

ramiro050 (approved):

saienduri:
Hey @ramiro050, I don't have write access, so can you merge this in whenever you get a chance? Thanks

stellaraccident (approved):

PR #2563, opened 2023-11-09
[Torch Dialect] support AtenArangeStartOutOp in ReduceOpVariants like…

qingyunqu:
… AtenBernoulli_FloatOp

It fixing case like: `%2110 = torch.aten.arange.start_out %int1, %int1517, %int1, %2109 : !torch.int, !torch.int, !torch.int, !torch.tensor -> !torch.tensor`.
`aten.arange.start_out` doesn't have value semantics also, means`%2110` is an alias for %2109.
So I decompose it to `aten.arange.start` + `torch.contents.overwrite`.  
The complex decomposition logic is target to handle cases like view and dtype cast which I add in e2e tests.

### DIFF
--- a/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
+++ b/lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
@@ -191,6 +191,16 @@ private:
 // Reduce Ops without value semantics but the corresponding without trailing
 // underscore variant doesn't exist.
 namespace {
+
+// int(ceil((end - start) / step))
+Value calculateArangeResultNumElements(PatternRewriter &rewriter, Location loc,
+                                       Value start, Value end, Value step) {
+  Value sub = rewriter.create<AtenSubOp>(
+      loc, Torch::NumberType::get(rewriter.getContext()), end, start);
+  Value div = rewriter.create<AtenDivOp>(loc, sub, step);
+  return rewriter.create<AtenCeilFloatOp>(loc, div);
+}
+
 class ReduceNonValueSemanticOps : public RewritePattern {
 public:
   ReduceNonValueSemanticOps(MLIRContext *context)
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -191,26 +191,76 @@ class ConvertHasValueSemanticsOpsToValueTensors : public RewritePattern {
...
     if (isa<AtenBernoulli_FloatOp>(op)) {
-      newOp = rewriter.create<ValsemVariantAtenBernoulliFloatOp>(
+      Operation *newOp = rewriter.create<ValsemVariantAtenBernoulliFloatOp>(
           loc, op->getResultTypes(), op->getOperands());
+      auto tensor =
+          rewriter.create<CopyToValueTensorOp>(loc, newOp->getResult(0));
+      createOverwriteTensorContents(rewriter, loc, tensor, op->getOperand(0));
+      rewriter.replaceOp(op, op->getOperand(0));
+      return success();
+    } else if (isa<AtenArangeStartOutOp>(op)) {
```
ramiro050, line 220: What case is this fixing? (do you have MLIR showing the before and after)? The `AtenBernoulli_FloatOp` has special handling here because there is no value semantics variant of that op (we replace it with a Torch-MLIR version of it `ValsemVariantAtenBernoulliFloatOp`). The handling for `AtenArangeStartOutOp` looks more like a decomposition.

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
qingyunqu: It fixing case like: `%2110 = torch.aten.arange.start_out %int1, %int1517, %int1, %2109 : !torch.int, !torch.int, !torch.int, !torch.tensor -> !torch.tensor`.
`aten.arange.start_out` doesn't have value semantics, means`%2110` is an alias for %2109.
So I decompose it to `aten.arange.start` + `torch.contents.overwrite`.  
The complex decomposition logic is target to handle cases like view and dtype cast which I add in e2e tests.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -191,26 +191,76 @@ class ConvertHasValueSemanticsOpsToValueTensors : public RewritePattern {
...
+
+      auto dtype = rewriter.create<PrimDtypeOp>(loc, out);
+      auto device = rewriter.create<PrimDeviceOp>(loc, out);
+      auto shape = rewriter.create<AtenSizeOp>(
+          loc, Torch::ListType::get(Torch::IntType::get(ctx)), out);
+      auto none = rewriter.create<ConstantNoneOp>(loc);
+      auto cstFalse = rewriter.create<ConstantBoolOp>(loc, false);
+      Value newArange = rewriter.create<AtenArangeStartStepOp>(
+          loc, op->getResult(0).getType(), start, end, step, dtype,
+          /*dtype=*/none, device, /*pin_memory=*/none);
```
ramiro050, line 249: nit: this `/*dtype` should be `/*layout`

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -191,26 +191,76 @@ class ConvertHasValueSemanticsOpsToValueTensors : public RewritePattern {
...
+      Value reshape = rewriter.create<AtenReshapeOp>(
+          loc, op->getResult(0).getType(), newArange, shape);
+      auto toDtype = rewriter.create<AtenToDtypeOp>(
+          loc, op->getResult(0).getType(), reshape, dtype,
+          /*non_blocking=*/cstFalse,
+          /*copy=*/cstFalse, /*memory_format=*/none);
```
ramiro050, line 255: I would have expected specifying `dtype` in `AtenArangeStartStepOp` to take care of having the right output type. What error where you seeing without this? (it could be that `AtenArangeStartStepOp` is not handling the `dtype` argument properly, and if so, we should fix it there).

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -191,26 +191,76 @@ class ConvertHasValueSemanticsOpsToValueTensors : public RewritePattern {
...
+      auto tensor =
+          rewriter.create<CopyToValueTensorOp>(loc, newOp->getResult(0));
+      createOverwriteTensorContents(rewriter, loc, tensor, op->getOperand(0));
+      rewriter.replaceOp(op, op->getOperand(0));
+      return success();
+    } else if (isa<AtenArangeStartOutOp>(op)) {
+      Value start = op->getOperand(0);
+      Value end = op->getOperand(1);
+      Value step = op->getOperand(2);
+      Value out = op->getOperand(3);
```
ramiro050, line 224: nit: use `if (auto arangeOutOp = dyn_cast<AtenArangeStartOutOp>(op))` to avoid using `getOperand` and used the named getters instead (`arangeOutOp.getStart()`, ...)

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
ramiro050: I see, thanks

ramiro050 (changes requested):
A few comments

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
qingyunqu: Oh, you are right. I have removed the postpositional `AtenToDtypeOp`.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
qingyunqu: Done.

ramiro050 (approved):
Thanks!

Issue #2559, opened 2023-11-07
Support dynamic shapes in pixel_shuffle


newling:
This is a follow-up for PR #2550 

                                                                                                                                                                                                                                                                             
  Currently the decomposition of inputs with dynamic shapes results in a
  failure when lowering to linalg. Specifically, the error:

  "Unhandled case of expand/collapse. "

  is hit in the lowering of aten.view to the linalg dialect (see
  ConvertAtenViewOp in TorchToLinalg/DataMovement.cpp where this error is
  thrown).

  The issue is that there is no way for the lowering from torch to linalg
  to know that the view ops (reshape ops) created in this decomposition can
  be mapped directly to tensor.expand_shape and tensor_collapse_shape. We
  DO know that this is the case: the first reshape is an expand_shape
  and the second one is a collapse_shape. But we only know this because of
  the semantics of pixel_shuffle, by the time the lowering sees the view
  ops, this special context is lost.

  There are a few possible options which we've considered
  
  1) Lower pixel_shuffle to linalg directly. This would mean a bit of
  duplication of the logic for lowering from aten.permute to linalg, and
  would not be the most 'gradual' lowering approach. But perhaps the most
  straightforward and understanble.

  2) [BAD] Rather than use aten.reshape ops in the decomposition below, insert
  tensor.expand_shape and tensor.collapse_shape ops directly. This would
  require inserting unrealized_conversion_cast ops to convert the types of
  the operands from !torch.tensor to to the MLIR built-in tensor type,
  which isn't ideal (unrealized_conversion_cast should only be used in
  conversion passes AFAIK (?)). This is not a viable option as we cannot
  make assumptions about the dialects that torch is lowered to.

  3) Create 2 new ops in the torch dialect, with the same semantics as
  tensor.expand_shape and tensor.collapse_shape, but with !torch.tensor
  operands and results. Then use these ops in the decomposition below,
  instead of aten.reshape ops.

ramiro050:
For some context, do you have the calls to `aten.view` that are causing problem? In other words, the MLIR at the `output_type=torch` level of the `aten.view` op with the definition of its arguments.

newling:
Some more info on the dynamic cases which fail (some cases actually work)

Consider what happens when SHAPE below takes different values:
```
class PixelShuffleModuleDynamic_0(torch.nn.Module):
    def __init__(self):
        super().__init__()

    @export
    @annotate_args([None, (SHAPE, torch.float32, True)])
    
[...]

PR #2572, opened 2023-11-13
Support for prims collapse op (lowering to linalg)

newling:
Steps taken:
1) add generator code to torch_ods_gen.py, run update_torch_ods.sh
2) add (custom) shape and type inference generator code to abstract_interp_lib_gen.py, run update_abstract_interp_lib.sh
3) Implement lowering to tensor.collapse_dims. Requires the `start` and `end` values to be constant, else lowering fails
4) Update xfail_sets.py (append to LTC_XFAIL_SET) after running /tools/e2e_test.sh --filter Collapse --verbose -c XX for all support backends (XX). 

Motivation: 
- Supporting the collapse operation will be useful for lowering of pixel_shuffle (see Issue #2559)


projects/pt1/python/torch_mlir_e2e_test/test_suite/reshape_like.py
```diff
@@ -122,6 +122,48 @@ def forward(self, a):
...
+
+
+@register_test_case(
+    module_factory=lambda: CollapseStaticModule())
+def CollapseStaticModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(2,3,4))
+
+# ==============================================================================
+#
+class CollapseDynamicModule(torch.nn.Module):
```
ramiro050, line 147: I imagine the lowering works for fully dynamic inputs too, right? If so, we should have a test case that is fully dynamic (this one collapses static with dynamic). Also, can we test the case where the collapse range encompasses more than two elements?

ramiro050 (changes requested):
Thanks!

projects/pt1/python/torch_mlir_e2e_test/test_suite/reshape_like.py
newling: Added these 2 tests. I also added a test where the input is rank-1, which revealed a bug (basically upstream MLIR doesn't allow collapse_shape if nothing collapses) Always a good idea asking for extra tests :-D

ramiro050 (approved):

projects/pt1/python/torch_mlir_e2e_test/test_suite/reshape_like.py
ramiro050: Nice!

Issue #1876, opened 2023-02-13
Run PyTorch end-to-end op correctness tests


silvasean:
Chillee pointed me at this code for how to run the PyTorch official op-level correctness tests.

https://github.com/pytorch/pytorch/blob/701412a4ec2602b5cb3a2fabbb74e43e51defcde/test/test_proxy_tensor.py#L1468

We should do something similar for Torch-MLIR. We should probably go through the dynamo integration only to allow us to easily tests both static and dynamic shapes easily, and to be forward-looking.

Additionally, looking forward a bit more, we can retire most of our e2e tests. Some will remain, such as scalar/list ops and resnet/etc.

Assigning to Ramiro to scope and prioritize.

PR #2573, opened 2023-11-13
(WIP) Refactor lowering of aten.view op to the tensor dialect

newling:
WIP -- will be split into 2 parts. 

There are no significant functional change to the lowering in this PR. There are some extremely minor improvements in the generated tensor dialect IR (better inference of 1's in shape, fewer casts) but the main contribution is a slightly more compact (-90 lines of non-comment code) which (in my opinion) is a bit simpler. 

Also added multiple unit tests for the lowering. These new tests all pass (with very minor changes in IR) with the pre-PR algorithm.

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -28,6 +28,7 @@
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 
+#include <algorithm>
 #include <numeric>
 
 using namespace mlir;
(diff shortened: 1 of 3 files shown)


test/Conversion/TorchToLinalg/view.mlir
```diff
@@ -73,4 +151,137 @@ func.func @torch.aten.view$expandInferredDim(%arg0: !torch.vtensor<[2,6],f32>) -
...
+// CHECK:  %[[CAST:.*]] = tensor.cast %[[EXPAND]] : tensor<1x1x8x1x1xf32> to tensor<?x?x8x1x?xf32>
+// CHECK: %[[BUILTIN_TENSOR_CAST:.*]] = torch_c.from_builtin_tensor %[[CAST]] : tensor<?x?x8x1x?xf32> -> !torch.vtensor<[?,?,8,1,?],f32>
+// CHECK: return %[[BUILTIN_TENSOR_CAST]] : !torch.vtensor<[?,?,8,1,?],f32>
+
+func.func @torch.aten.view$expandWithOnes(%arg0: !torch.vtensor<[8],f32>) -> !torch.vtensor<[?,?,8,1,?],f32> {
+    %int8 = torch.constant.int 8
+    %int1 = torch.constant.int 1
+    %int-1 = torch.constant.int -1
+    %0 = torch.prim.ListConstruct %int-1, %int-1, %int8, %int1, %int-1 : (!torch.int, !torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
+    %1 = torch.aten.view %arg0, %0 : !torch.vtensor<[8],f32>, !torch.list<int> -> !torch.vtensor<[?,?,8,1,?],f32>
```
qedawkins, line 171: Is this a legal view op? My understanding was that you can only have one inferred (`-1`) dimension size.
(IIRC reshape lowers to view: https://pytorch.org/docs/stable/generated/torch.reshape.html#torch.reshape)

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -171,439 +172,348 @@ class ConvertAtenFlattenUsingIntsOp
...
-      // Match torch.aten.size.int(inputTensor, inputDim) with constant inputDim
-      if (matchPattern(outputDimSize,
-                       m_TorchTensorSizeInt(inputTorchTensor, &inputDim))) {
-        outputShape[outputDim] = inputShape[inputDim];
-      } else if (matchPattern(outputDimSize,
-                              m_TorchConstantInt(&outputDimSizeInt))) {
-        if (outputDimSizeInt != -1) {
-          outputShape[outputDim] = outputDimSizeInt;
-        }
+  getInputAndOutputShape(Value inTorchTensor,
```
qedawkins, line 244: It's a little hard to follow, but is this primarily just changing variable names and adding whitespace? I think the names before were reasonable, and keeping `inputDim` and `outputDimSizeInt` separate, while technically not necessary, helps readability due to the different semantic meanings in each case.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -171,439 +172,348 @@ class ConvertAtenFlattenUsingIntsOp
...
-         llvm::enumerate(outputSizeTorchInt)) {
-      int64_t inputDim;
-      int64_t outputDimSizeInt;
-      // Match torch.aten.size.int(inputTensor, inputDim) with constant inputDim
-      if (matchPattern(outputDimSize,
-                       m_TorchTensorSizeInt(inputTorchTensor, &inputDim))) {
-        outputShape[outputDim] = inputShape[inputDim];
-      } else if (matchPattern(outputDimSize,
-                              m_TorchConstantInt(&outputDimSizeInt))) {
-        if (outputDimSizeInt != -1) {
```
qedawkins, line 306: Why are we able to drop this check? If a dimension size is `-1` then it needs to be inferred from the rest of the shape of the view, meaning it either can be statically resolved as an earlier step or we need to keep it dynamic.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -171,439 +172,348 @@ class ConvertAtenFlattenUsingIntsOp
...
-        xIndicesResult.push_back(nextXIndex++);
-      } else {
-        if (nextYIndex == yDims.size() || yDims[nextYIndex] == kUnknownSize)
-          return failure();
-        yTotalSize *= yDims[nextYIndex];
-        yIndicesResult.push_back(nextYIndex++);
+    auto nUnknowns = unknowns.size();
+
+    auto toProduct =
+        productReduce(to) / std::pow<int64_t>(kUnknownSize, nUnknowns);
```
qedawkins, line 204: I'm not sure about this change. What if torch-mlir decides to change `kUnknownSize` to, say, INT64_MIN? This is what happened in upstream llvm and this could be a tricky bug to catch.

qedawkins (commented):
A few initial comments. Will let @ramiro050 take a look too as he wrote the most recent refactor of this lowering.

test/Conversion/TorchToLinalg/view.mlir
newling: I'm not sure. I don't have a good understanding of how the classic pytorch '-1' and dynamic shape '?', as created by dynamo, are related.  But this IR does look a bit confused..

lib/Conversion/TorchToLinalg/DataMovement.cpp
qedawkins: +1 to that then. My understanding of those kinds of sentinel constants is that you shouldn't assume they are a specific value, just that they are static, meaning arithmetic with them is probably a bad idea. If this was present in the old code then removing the division would definitely be nice cleanup!

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: I feel the same way about `tryFillOneDirection` and `tryFillBothDirections`. What is the reason for not keeping `calculateSingleDynamicSize` the way it is?

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -171,439 +172,348 @@ class ConvertAtenFlattenUsingIntsOp
...
+          viewOp, "size list length mismatches with the result type rank");
     }
-    SmallVector<Value> outputSizeInt = getTypeConvertedValues(
-        rewriter, loc, typeConverter, outputSizeTorchInt);
-    if (resultRank != (int64_t)outputSizeInt.size()) {
+
+    if (inRank == 0 || outRank == 0) {
       return rewriter.notifyMatchFailure(
-          op, "desired size list length mismatches with the result type rank");
+          viewOp, "It is impossible to expand from or collapse to a scalar");
```
ramiro050, line 306: This is not true. It is valid to have a rank-zero inputs and outputs (they are just not supported in this pattern yet):

```python
>>> import torch
>>> torch.ops.aten.view(torch.tensor(1), [1,1,1])
tensor([[[1]]])
>>> torch.ops.aten.view(torch.tensor(1), [])
tensor(1)
```

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -171,439 +172,348 @@ class ConvertAtenFlattenUsingIntsOp
...
+    SmallVector<Value> outSizeTorchInt;
+    if (!getListConstructElements(viewOp.getSize(), outSizeTorchInt)) {
+      return rewriter.notifyMatchFailure(viewOp,
                                          "unimplemented: the target size is "
                                          "not constructed from ListConstruct");
     }
-    if (llvm::count_if(outputSizeTorchInt, [](Value size) -> bool {
-          int64_t sizeInt;
-          if (matchPattern(size, m_TorchConstantInt(&sizeInt)))
-            return sizeInt == -1;
```
ramiro050, line 350: Why is this check gone?

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -171,439 +172,348 @@ class ConvertAtenFlattenUsingIntsOp
...
-                       m_TorchTensorSizeInt(op.getSelf(), &inputDim))) {
-        unchangedDims.push_back(std::make_pair(inputDim, outputDim));
-      } else if (singleDynDimsAreEqual &&
-                 outputShape[outputDim] == kUnknownSize) {
-        // If the input and output have a single dynamic dimension and the
-        // product of the other dimensions is the same, then we know that the
-        // dynamic dimension is unchanged.
-        inputDim = std::distance(inputShape.begin(),
-                                 llvm::find(inputShape, kUnknownSize));
-        unchangedDims.push_back(std::make_pair(inputDim, outputDim));
```
ramiro050, line 400: is this case gone?

ramiro050 (changes requested):
I'll take a closer look tomorrow, in the meantime a few comments

newling (commented):
@ramiro050 Thank you for your comments. 

I would like to factorize this PR into separate parts. The first part will be improving test coverage of the existing code. Does that sound reasonable? I don't want to over test and make future changes painful, but there seem to be a few important paths through the code which aren't currently tested (at least with the lit tests I've seen). 

I'm relatively new to working on github -- I wonder, do you know if there are any options with github to show test coverage in PRs? In my previous job we added code coverage to our project (in phabricator), and it made it really nice as a developer and reviewer to get an idea of which lines of code were being tested.

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: For these cases, the error should be prefixed with `unimplemented: ...`. This means, the behavior is valid in PyTorch, but in torch-mlir we don't yet know how to handle it

ramiro050:
> I would like to factorize this PR into separate parts.

Sounds good. I'm all for small PRs 🙂 

> I'm relatively new to working on github -- I wonder, do you know if there are any options with github to show test coverage in PRs? In my previous job we added code coverage to our project (in phabricator), and it made it really nice as a developer and reviewer to get an idea of which lines of code were being tested.

We don't have any such functionality, unfortunately. We also don't have good ways of testing error cases. The approach we've taken in the project is to test every success path through the pattern. After that, any additional bug fix should include a test for the bug that was found. The approach is far from perfect, but it has worked reasonably well.

Our goal is to eventually just plug into the PyTorch testing and trust that it is thorough enough: https://github.com/llvm/torch-mlir/issues/1876 (assigned to me, but I have not had time to look into it).

newling:
Additional tests to include... 
https://github.com/llvm/torch-mlir/issues/2854
https://github.com/llvm/torch-mlir/issues/2877

PR #2575, opened 2023-11-14
[E2E] add nan case in elementwise comparison e2e tests

qingyunqu:


projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -1241,7 +1241,7 @@
     "HardswishRandomModule_basic",
     "BucketizeTensorStaticFloatModule_basic",
     "BucketizeTensorStaticModule_basic",
-    "ElementwiseLeFloatTensorModule_basic",
+    # "ElementwiseLeFloatTensorModule_basic",
```
qingyunqu, line 1244: @AmosLewis Do you know how to fix this case in torch-to-tosa conversion?

AmosLewis: If you want to add Nan test Case, you should add a new "ElementwiseLeFloatTensorNanModule_basic" test instead of rewrite the existing one.

qingyunqu: ok, done.

AmosLewis:
LGTM

PR #2576, opened 2023-11-14
Lowering to linalg of prims split_dim op 

newling:
Adds support for lowering to prims split_op. 

Very similar design to collapse op lowering in https://github.com/llvm/torch-mlir/pull/2572, with 
some small differences,  because the split_dim op (in pytorch) is view-changing whereas the collapse 
is not. The difference means that 

1) it must be registered in the function Torch::isViewLikeOp(Operation *op) 
2) it must be be added to the "expected fail" set for the torch dynamo backend.


newling:
Hi @ramiro050 -- following up from yesterday's ops.prims.collapse https://github.com/llvm/torch-mlir/pull/2572 here is the split_dim lowering to linalg. It is basically the same as the collapse lowering, but it doesn't work (described in summary in my first comment). 

If you have any ideas as to why it doesn't work, I would really appreciate any feedback... I've been starting at it for too long now, and thought it might be worth getting an expert to take look.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1757,6 +1759,49 @@ class ConvertAtenDetachOp : public OpConversionPattern<AtenDetachOp> {
...
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    auto aRankedTensorType = adaptor.getA().getType().cast<RankedTensorType>();
+    const TypeConverter *typeConverter = getTypeConverter();
+
+    auto resultRankedTensorType =
+        typeConverter->convertType(op.getType()).cast<RankedTensorType>();
+
+    // The dimension being split ust be statically known.
```
ramiro050, line 1779: nit: `ust` -> `must`

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1805,6 +1850,10 @@ void mlir::torch::torch_to_linalg::populateUncategorizedPatternsAndLegality(
   patterns.add<ConvertAtenNllLossForwardOp>(typeConverter, context);
   target.addIllegalOp<AtenBatchNormOp>();
   patterns.add<ConvertAtenBatchNormOp>(typeConverter, context);
+
+  target.addIllegalOp<PrimsSplitDimOp>();
+  patterns.add<ConvertPrimsSplitDimOp>(typeConverter, context);
```
ramiro050, line 1855: nit: remove empty lines to make it consistent

projects/pt1/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -3694,6 +3720,8 @@ def prims〇squeeze〡dtype(a_rank_dtype: Tuple[int, int], dimensions: List[int]
 
 
 
+
+
```
ramiro050, line 3724: nit: remove extra whitespace

projects/pt1/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -905,6 +930,7 @@ def aten〇squeeze〇dim〡shape(self: List[int], dim: int) -> List[int]:
 def prims〇squeeze〡shape(a: List[int], dimensions: List[int]) -> List[int]:
     return upstream_shape_functions.squeeze_dims(a, dimensions)
 
+
```
ramiro050, line 933: nit: remove extra whitespace

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -341,6 +341,7 @@ def ElementwiseUnsqueezeBroadcastModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(4), tu.rand())
 
 
+
```
ramiro050, line 344: nit: changes to this file not needed

projects/pt1/python/torch_mlir/dialects/torch/importer/jit_ir/build_tools/abstract_interp_lib_gen.py
```diff
@@ -177,6 +177,8 @@ def aten〇glu〡shape(self: List[int], dim: int = -1) -> List[int]:
     assert self[dim] % 2 == 0, "glu's dim size must be multiply of 2"
     return self[:dim] + [self[dim] // 2] + self[dim+1:]
 
+
+
```
ramiro050, line 181: nit: remove whitespace

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1298,6 +1299,7 @@ class ConvertElementwiseOp : public ConversionPattern {
 //     nll_loss_forward[i] = -(input[i][indi]);
 // TODO: `weight`operand is still to be taken care of.
 namespace {
+
```
ramiro050, line 1302: nit: remove whitespace

ramiro050:
> I see that one generated class has the ValueSemantic trait the other. Is that relevant?

Yeah, this is likely the source of the error. The op does not have value semantics, but it is just a view-like op (it does not modify the contents of the tensor), so you just need to add it here:

https://github.com/llvm/torch-mlir/blob/e81282ae8f23a34b5a7dbef94bea3dfdf945ebcd/lib/Dialect/Torch/Utils/Utils.cpp#L196

ramiro050 (changes requested):
A few initial comments

newling:
> > I see that one generated class has the ValueSemantic trait the other. Is that relevant?
> 
> Yeah, this is likely the source of the error. The op does not have value semantics, but it is just a view-like op (it does not modify the contents of the tensor), so you just need to add it here:
> 
> https://github.com/llvm/torch-mlir/blob/e81282ae8f23a34b5a7dbef94bea3dfdf945ebcd/lib/Dialect/Torch/Utils/Utils.cpp#L196

That fixed it. Thank you! Will tidy up and address remaining comments soon.

ramiro050 (approved):
Thanks!

PR #2584, opened 2023-11-20
Additional tests for view lowering

newling:
The logic for lowering the aten view op to linalg is fairly complex, and IMO a bit undertested. In this PR I have tried to follow all non-failing paths through the lowering and add unit tests. 

There is 1 logical change to the lowering: redundant tensor.cast ops (same source and destination type) are folded.


ramiro050 (approved):
Thanks, this is great!

PR #2585, opened 2023-11-21
Initial TorchOnnxToTorch conversion pipeline.

stellaraccident:
Adds a pipeline to convert custom ops and metadata represented as `torch.operator` custom ops to corresponding `torch` ops where possible.

This is part of a multi-part approach for building ONNX import in as a regular feature of torch-mlir. It is focused on the conversions vs the infra. We will end up maintaining a [pure-python importer](https://github.com/nod-ai/SHARK-Turbine/blob/main/python/shark_turbine/importers/onnx_importer.py) to go with this in torch-mlir, and we will also maintain test case generation utilities derived from it.

I have left substantial documentation in the README of the conversion directory, including the recommended approach that we will take to keep building this out.

(note that this organizes the code to coincide with the refactoring in #2442 versus the current flat arrangement)

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -14,17 +14,15 @@ set(LinkedLibs
   MLIRTosaDialect
   MLIRSupport
 
-  # Dialect passes.
-  TorchMLIRTorchPasses
+  # Dialects.
+  TorchMLIRTMTensorDialect
+  TorchMLIRTorchDialect
   TorchMLIRTorchConversionDialect
 
-  # Dialects.
-  TorchMLIRTorchDialect
-  TorchMLIRTorchConversionPasses
-
-  # TMTensor.
+  # Dialect passes.
   TorchMLIRTMTensorPasses
-  TorchMLIRTMTensorDialect
+  TorchMLIRTorchConversionPasses
+  TorchMLIRTorchPasses
 
   # Conversion passes.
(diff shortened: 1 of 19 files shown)


lib/CMakeLists.txt
```diff
@@ -14,16 +14,21 @@ set(LinkedLibs
   MLIRTosaDialect
   MLIRSupport
 
+  # Dialect passes.
   TorchMLIRTorchPasses
   TorchMLIRTorchConversionDialect
 
+  # Dialects.
```
vivekkhandelwal1, line 21: Both these should be dialects and the other one should be under `Dialect Passes.`

vivekkhandelwal1 (approved):
LGTM

silvasean:
Drive-by nit: double-check that `torch.onnx` won't collide with any actual PyTorch op. We may need `torch._tm_onnx` or some such.

stellaraccident:
> Drive-by nit: double-check that `torch.onnx` won't collide with any actual PyTorch op (not sure who own that namespace now, or how it is used -- hopefully it is isomorphic to this proposal). We may need `torch._tm_onnx` or some such though.

Thanks. I did double check. I think this is good enough for now and once we get enough support to be real, we can assess/ask/etc. if we have to change it, we are at the level of a fancy sed command in the future, so I think it is fine to defer for now in favor of just moving forward.

PR #2589, opened 2023-11-21
[MLIR][TORCH] Add support for unsigned integer types

vivekkhandelwal1:
Refer: https://github.com/pytorch/pytorch/issues/58734


stellaraccident (approved):
Thanks. We can probably also add the fp8 types as they say on another issue that those are coming too. Can be another patch.

PR #2590, opened 2023-11-21
Improve decomposition of pixel_shuffle (support dynamic shapes)

newling:
Here the aten.reshape ops are replaced with prims.collapse and prims.split_dim ops, which means that the cases where the lowering of reshape from torch to linalg which are not supported, are avoided. 

Essentially, by using the collapse and split_dim ops instead of the reshape op, we are not "losing" the information that the reshapes do not arbitrarily mix dimensions. 

3 additional tests added: 
- fully dynamic, 
- dynamic only the spatial dimensions, 
- dynamic only in the non-spatial dimensions.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -1095,16 +1095,22 @@ public:
 };
 } // namespace
 
-// Decompose aten.pixel_shuffle into: aten.permute and aten.reshape operations.
+// Decompose aten.pixel_shuffle into: prims.split_dim, aten.permute, and
+// prims.collapse operations.
 //
-// If input is a tensor of shape (*leading_dims, C*r*r, H, W), where
-// leading_dims is of size N, then
+// If input is a tensor of shape
+//     (*leading_dims, C*r*r, H, W),
+//
+// where leading_dims is of size N, then
 //    X = pixel_shuffle(input, upscale_factor)
 //
 // gets replaced with
-//    A = input.reshape(*leading_dims, C, r, r, H, W)
-//    B = A.permute(0, ..., N, N+3, N+1, N+4, N+2)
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1169,98 +1181,87 @@ class DecomposeAtenPixelShuffleOp
...
         loc, Torch::ListType::get(Torch::IntType::get(op->getContext())),
         permutation);
 
-    Value B = rewriter.create<AtenPermuteOp>(
-        loc, getTypeFromShape(postPermuteShape), A, permuteDimsOrder);
-
-    Value outShapeList =
-        rewriter.create<PrimListConstructOp>(loc, listType, outShape);
+    // Split input channel inC -> (inC, factorSquared)
+    auto X = rewriter
```
qedawkins, line 1239: nit: Use a more descriptive name than `X`. (Maybe `shuffledValue`?)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1169,98 +1181,87 @@ class DecomposeAtenPixelShuffleOp
...
-      const auto intShape = getIntShape(vals);
-      return ValueTensorType::get(vals[0].getContext(),
-                                  llvm::ArrayRef(intShape), inOptionalDType);
-    };
+    SmallVector<Value> partiallyExpandedShape = leadingDims;
+    partiallyExpandedShape.insert(partiallyExpandedShape.end(),
+                                  {outC, factorSquared, inH, inW});
 
+    SmallVector<Value> prePermuteShape = leadingDims;
     prePermuteShape.insert(prePermuteShape.end(),
```
qedawkins, line 1213: nit: use `SmallVector::append(<initializer list>)`, same for the below.

qedawkins (approved):
LGTM

PR #2591, opened 2023-11-21
[Torch Dialect]Add Support for AtenGroupNormOp and AtenNativeGroupNormOp

JianzheXiao:


projects/pt1/python/torch_mlir_e2e_test/test_suite/norm_like.py
```diff
@@ -243,6 +243,48 @@ def NativeBatchNormNoneWeightModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
+class GroupNormModule(torch.nn.Module):
```
ramiro050, line 246: No need to have this test. The goal of these e2e tests is to make sure each individual op works as intended. This is achieved by calling the op through `torch.ops.aten...`. When you use `torch.nn....`, PyTorch will often generate extra logic (code for checking inputs, dispatching, etc) that can lead to failures in torch-mlir unrelated to the op being added (I believe this is what is happening here). Using the `nn` module version also makes it harder from looking at the test to know what Torch-MLIR op it corresponds to.

More info on this here: https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#forward-method

ramiro050 (changes requested):
Initial comment. Will take a closer look in a bit

projects/pt1/python/torch_mlir_e2e_test/test_suite/__init__.py
```diff
@@ -10,7 +10,8 @@
 from torch_mlir._version import torch_version_for_comparison, version
 
 COMMON_TORCH_MLIR_LOWERING_XFAILS = {
-    "NativeGroupNormModule_basic",
+    # "GroupNormModule_basic" failed because some list operator are not canonicalized.
+    "GroupNormModule_basic",
```
ramiro050, line 14: I thought this was failing because of the use of an `torch.nn` module for the op. What error are you getting?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3327,6 +3327,165 @@ class DecomposeConstantTensorAllocLikeOp : public OpRewritePattern<OpTy> {
...
+      groupNormOutput = rewriter.create<AtenMulTensorOp>(
+          loc, inputType, groupNormOutput, weightReshaped);
+    }
+    if (!bias.getType().isa<Torch::NoneType>()) {
+      auto biasReshaped = rewriter.create<AtenViewOp>(
+          loc, baseType, bias, /*shape=*/viewShapeSizeList);
+      groupNormOutput = rewriter.create<AtenAddTensorOp>(
+          loc, inputType, groupNormOutput, biasReshaped,
+          /*alpha=*/cstOne);
+    }
```
ramiro050, line 3474: can we have e2e tests for the case where `weight` and `bias` are `None` to check that this works as expected?

ramiro050 (changes requested):
Looks mostly good to me. Just two comments

projects/pt1/python/torch_mlir_e2e_test/test_suite/__init__.py
qingyunqu: Yes, it will generate some code like:
```
%0 = torch.vtensor.literal(dense<0.000000e+00> : tensor<4xf32>) : !torch.vtensor<[4],f32>
    %1 = torch.vtensor.literal(dense<1.000000e+00> : tensor<4xf32>) : !torch.vtensor<[4],f32>
    %float1.000000e-05 = torch.constant.float 1.000000e-05
    %str = torch.constant.str "Expected more than 1 value per channel when training, got input size {}"
    %str_0 = torch.constant.str "builtins.ValueError"
    %2 = torch.prim.ListConstruct %int4, %int2 : (!torch.int, !torch.int) -> !torch.list<int>
    %3 = torch.prim.ListConstruct %int6, %int7 : (!torch.int, !torch.int) -> !torch.list<int>
    %4 = torch.aten.list.t %3 : !torch.list<int> -> !torch.list<int>
    %5 = torch.aten.add.t %2, %4 : !torch.list<int>, !torch.list<int> -> !torch.list<int>
    %6 = torch.aten.__getitem__.t %5, %int0 : !torch.list<int>, !torch.int -> !torch.int
    %7 = torch.aten.len.t %5 : !torch.list<int> -> !torch.int
    %8 = torch.aten.sub.int %7, %int2 : !torch.int, !torch.int -> !torch.int
    %9 = torch.prim.Loop %8, %true, init(%6) {
    ^bb0(%arg1: !torch.int, %arg2: !torch.int):
      %37 = torch.aten.add.int %arg1, %int2 : !torch.int, !torch.int -> !torch.int
      %38 = torch.aten.__getitem__.t %5, %37 : !torch.list<int>, !torch.int -> !torch.int
      %39 = torch.aten.mul.int %arg2, %38 : !torch.int, !torch.int -> !torch.int
      torch.prim.Loop.condition %true, iter(%39 : !torch.int)
    } : (!torch.int, !torch.bool, !torch.int) -> !torch.int
    %10 = torch.aten.eq.int %9, %int1 : !torch.int, !torch.int -> !torch.bool
    torch.prim.If %10 -> () {
      %37 = torch.aten.format(%str, %5) : !torch.str, !torch.list<int> -> !torch.str
      torch.prim.RaiseException %37, %str_0 : !torch.str, !torch.str
      torch.prim.If.yield
    } else {
      torch.prim.If.yield
    }
```

If we want to resolve it, maybe we should add some canonicalize functions for `ListType`(but I think it's a little hard to resolve reference problem).  
So I agree to your idea. It's **no need** to add this testcase. It's a problem which would not appear in torchscript or which should be resolved by dynamo/fx.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
qingyunqu: done

ramiro050 (approved):
LGTM!

PR #2595, opened 2023-11-28
[MLIR][TORCH] Fix TypeConverter automatically convert the ui8 to i8

AmosLewis:
When converting vtenor to buildin tensor, the converter wrongly converts the ui8 to i8.
  This will lead to other typecast issues in later usage of i8. Here is an example:
  https://github.com/nod-ai/SHARK-Turbine/issues/110


AmosLewis:
It could be fixed by https://github.com/llvm/torch-mlir/pull/2597

PR #2601, opened 2023-11-30
[MLIR][ONNX] Add OnnxToTorch support for q-z ops (specific ops in description)

saienduri:
This commit adds the OnnxToTorch support for Reciprocal, Round, ScatterElements, Sigmoid, Sin, Tanh, Sqrt, Sub, Sum, Where, Xor, Squeeze, Unsqueeze ops.
For reviewers, the ops that weren't trivial and probably require extra review are Sum, Squeeze, and Unsqueeze.


include/torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h
```diff
@@ -54,6 +54,21 @@ struct OpBinder {
     return success();
   }
 
+  ParseResult tensorOperands(SmallVector<Value> &dimList) {
+    int64_t numOperands = op->getNumOperands();
+    if (numOperands < 3)
```
stellaraccident, line 59: Why this failure? Seems like they're should not be a limit here

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -26,4 +28,392 @@ using namespace mlir::torch::onnx_c;
...
+                  return success();
+                });
+  patterns.onOp(
+      "ScatterElements", 18,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        SmallVector<Value> valList;
+        int64_t axis;
+        std::string reduction;
+        if (binder.tensorOperands(valList) ||
```
stellaraccident, line 72: Recommend passing the number of expected to the helper and validating that there (vs hard coded 3).

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -26,4 +28,392 @@ using namespace mlir::torch::onnx_c;
...
+        if (binder.op->getNumOperands() == 1) {
+          Torch::ValueTensorType resultType;
+          Value x;
+          if (binder.tensorOperand(x) || binder.tensorResultType(resultType))
+            return failure();
+          rewriter.replaceOp(binder.op, x);
+          return success();
+        }
+        if (binder.op->getNumOperands() == 2) {
+          Torch::ValueTensorType resultType;
```
stellaraccident, line 182: Add comment: short circuit to binary add.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -26,4 +28,392 @@ using namespace mlir::torch::onnx_c;
...
+              binder.tensorResultType(resultType))
+            return failure();
+          Value const1 = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getType<Torch::IntType>(),
+              rewriter.getIntegerAttr(rewriter.getIntegerType(64), 1));
+          rewriter.replaceOpWithNewOp<Torch::AtenAddTensorOp>(
+              binder.op, resultType, x, y, const1);
+          return success();
+        }
+        // When binder.op->getNumOperands() > 2
```
stellaraccident, line 195: I'm not sure this choice of how to translate is the best idea: you are introducing a potentially very expensive data movement step to compute the sun, which is not implied by the original op def.

I'd recommend just producing n-1 add ops. Do your setup for the first two but don't return. Then process 2..n but adding to the last result and return the last result at the end. One exit, no data movement.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -26,4 +28,392 @@ using namespace mlir::torch::onnx_c;
...
+        Value dimValueList = rewriter.create<Torch::PrimListConstructOp>(
+            binder.getLoc(),
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+            dimList);
+        rewriter.replaceOpWithNewOp<Torch::PrimsSqueezeOp>(
+            binder.op, resultType, data, dimValueList);
+        return success();
+      });
+  patterns.onOp(
+      "Unsqueeze", 13,
```
stellaraccident, line 332: I don't think we want to be punting this to view. Let's just unroll it to multiple unsqueeze. That will preserve the canonical single dim case and be easier on the tools downstream.

If too hard to do in one step, remove from this patch and do in a follow-up.

stellaraccident (changes requested):
Thanks - several comments to address. If some are too disruptive to do in one patch, feel free to drop the op fun this patch and do in a dedicated follow-up.

include/torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h
saienduri: I was thinking if numOperands < 3, the other tensorOperand methods should be used. But, will change so this method can be used regardless of numOperannds.

stellaraccident: The weird coupling isn't worth it and will cause hard to diagnose errors. Leave it as a caller choice too do one vs the other.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
saienduri: I thought this was a neat way to lower, but this makes total sense. I will definitely consider data movement costs going forward. Made necessary changes

stellaraccident (approved):
Thanks!

There are a couple of commented out regions that it would be good to remove before landing.

PR #2603, opened 2023-12-01
[TorchToLinalg] NFC: Move Utils.h to an externally accessible location

qedawkins:


qedawkins:
> LGTM. Mind if I ask what the motivation is?

Downstream IREE composes its own lowering pipeline of the Torch dialect so that we can lower directly to certain IREE ops. Essentially these utils are just useful for doing some of those kinds of lowerings. This isn't really meant to make these utils part of a stable API if that's a concern though.

PR #2611, opened 2023-12-05
[MLIR][ONNX] Add OnnxToTorch support for bitwise and math ops

vivekkhandelwal1:
This commit adds the OnnxToTorch support for BitwiseXor, BitwiseOr, Div, Equal, Cast, Ceil, Floor, Cos, and Clip op.
This commit also adds the TorchToLinalg support for aten.clamp.Tensor and aten.clamp_min.Tensor op.

Signed-Off By: vivekkhandelwal1424@gmail.com


include/torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h
```diff
@@ -54,6 +54,15 @@ struct OpBinder {
     return success();
   }
 
+  ParseResult getIdxTensorOperand(Value &valueIdx, int64_t idx) {
```
stellaraccident, line 57: Try to match style of others and name in a way that is self documenting. Recommend `tensorOperandAtIndex`

include/torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h
```diff
@@ -54,6 +54,15 @@ struct OpBinder {
...
+    if (op->getNumOperands() <= idx)
```
stellaraccident, line 58: Recommend inverting for better readabilitu: idx >= op->getNumOperands()

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -195,4 +207,126 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          Value operand;
+          if (binder.tensorOperand(operand) ||
+              binder.tensorResultType(resultType))
+            return failure();
+          Value cstNone =
+              rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+          rewriter.replaceOpWithNewOp<Torch::AtenClampOp>(
+              binder.op, resultType, operand, /*min=*/cstNone, /*max=*/cstNone);
+        } else if (binder.op->getNumOperands() == 2) {
+          Value operand0, operand1;
```
stellaraccident, line 267: Recommend naming these as used: source, max or equivalent

(And for the three arg case below)

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -195,4 +207,126 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+            binder.op, resultType, lhs, rhs);
+        return success();
+      });
+  patterns.onOp(
+      "Cast", 19, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value operand;
+        int64_t dtypeInt;
+        if (binder.tensorOperand(operand) ||
+            binder.s64IntegerAttr(dtypeInt, "to") ||
```
stellaraccident, line 228: Can we verify that these dtype codes line up between on x PyTorch? (I'm not sure I see what is making sure they do)

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -195,4 +207,126 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          if (binder.getIdxTensorOperand(operand0, 0) ||
+              binder.getIdxTensorOperand(operand1, 1) ||
+              binder.getIdxTensorOperand(operand2, 2) ||
+              binder.tensorResultType(resultType))
+            return failure();
+          rewriter.replaceOpWithNewOp<Torch::AtenClampTensorOp>(
+              binder.op, resultType, operand0, /*min=*/operand1,
+              /*max=*/operand2);
+        }
+        return success();
```
stellaraccident, line 284: Missing else branch for wrong number is operands. Recommend moving the success/failure return into each branch: this will cause a compiler earning if unintended fall through.

stellaraccident (changes requested):
Mostly naming suggestions but one functional error due to a missing else branch so marking request changes.

stellaraccident (approved):
I'm fine post submit reviewing if you just use an inline switch statement. Would like another look prior to landing if you use a map.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1015,6 +1015,56 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+      auto minPromoted = convertScalarToDtype(b, loc, payloadArgs[1], dtype);
+      Value pred;
+      if (dtype.isa<mlir::FloatType>()) {
+        pred = b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::ULT, result,
+                                       minPromoted);
+      } else if (dtype.isa<mlir::IntegerType>()) {
+        pred = b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::slt, result,
+                                       minPromoted);
+      } else {
+        clampTensor.emitError(
```
stellaraccident, line 1043: Missing return?

I think this function needs to be refactored. It is very important that:

A. Any failure returns happen before any ops are created. Pattern match predictes should be at the top.
B. The pattern be made resilient to multiple application and partial application.

We really shouldn't be emitting errors during pattern application like this due to the above. We can notifyMatchResult or mutate to at an attribute indicating what is wrong (with a pass at the end to validate and report errors.

Since this kind of thing is already being done here, I'm not going to request that you did all error emission in this patch, but could you please raise an issue to refactor out all error reporting from these patterns?

You do need to move all checks that will return failure to the top before creating any ops though. I expect that if you have a unit test that tests this failure case, it will crash as written.

stellaraccident (approved):

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1702,13 +1708,25 @@ def aten〇clamp_min〡dtype(self_rank_dtype: Tuple[int, int], min: Union[int, f
         return torch.int64
     return self_dtype
 
+def aten〇clamp_min〇Tensor〡dtype(self_rank_dtype: Tuple[int, int], min_rank_dtype: Tuple[int, int]) -> int:
```
ramiro050, line 1711: drive by comment: all dtype functions should be tested. Should be a matter of adding something like

```python
@check_dtype_function(_check_tensors_with_the_same_dtype(num_of_tensors=2))
```

Same comment for dtype function below.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
vivekkhandelwal1: Opened an issue here:https://github.com/llvm/torch-mlir/issues/2627

PR #2613, opened 2023-12-05
Implement torch.aten.argmin

frafranz:
I have lowered the torch.aten.argmin operator to linalg via decomposition into torch.aten.min.dim.


stellaraccident:
@qedawkins does this need to be an optional decomposition? There are many ways to implement these and we need to ensure that the frontend doesn't prematurely lower them universally.

qedawkins:
> @qedawkins does this need to be an optional decomposition? There are many ways to implement these and we need to ensure that the frontend doesn't prematurely lower them universally.

In traditional Aten fashion it's really hard to tell the difference between these two ops. At least wrt to the lowering to Linalg, as far as I can tell `AtenMaxDimOp` and `AtenMinDimOp` is exactly what I would expect from an arg___ anyway. StableHLO appears to have a lowering for `AtenArgmaxOp`, but I don't know if that's just an artifact from before the argmax lowering was added.

So this is probably fine globally.

stellaraccident:
> > @qedawkins does this need to be an optional decomposition? There are many ways to implement these and we need to ensure that the frontend doesn't prematurely lower them universally.
> 
> In traditional Aten fashion it's really hard to tell the difference between these two ops. At least wrt to the lowering to Linalg, as far as I can tell `AtenMaxDimOp` and `AtenMinDimOp` is exactly what I would expect from an arg___ anyway. StableHLO appears to have a lowering for `AtenArgmaxOp`, but I don't know if that's just an artifact from before the argmax lowering was added.
> 
> So this is probably fine globally.

Great, thanks. That's what I was looking to verify.

frafranz:
> > > @qedawkins does this need to be an optional decomposition? There are many ways to implement these and we need to ensure that the frontend doesn't prematurely lower them universally.
> > 
> > 
> > In traditional Aten fashion it's really hard to tell the difference between these two ops. At least wrt to the lowering to Linalg, as far as I can tell `AtenMaxDimOp` and `AtenMinDimOp` is exactly what I would expect from an arg___ anyway. StableHLO appears to have a lowering for `AtenArgmaxOp`, but I don't know if that's just an artifact from before the argmax lowering was added.
> > So this is probably fine globally.
> 
> Great, thanks. That's what I was looking to verify.

Just to join in the discussion: What exactly do you mean by optional decomposition? Optional in the sense that the op is not decomposed if argmin is not made illegal in the lower to backend contract?

stellaraccident:
> > > > @qedawkins does this need to be an optional decomposition? There are many ways to implement these and we need to ensure that the frontend doesn't prematurely lower them universally.
> > > 
> > > 
> > > In traditional Aten fashion it's really hard to tell the difference between these two ops. At least wrt to the lowering to Linalg, as far as I can tell `AtenMaxDimOp` and `AtenMinDimOp` is exactly what I would expect from an arg___ anyway. StableHLO appears to have a lowering for `AtenArgmaxOp`, but I don't know if that's just an artifact from before the argmax lowering was added.
> > > So this is probably fine globally.
> > 
> > 
> > Great, thanks. That's what I was looking to verify.
> 
> Just to join in the discussion: What exactly do you mean by optional decomposition? Optional in the sense that the op is not decomposed if argmin is not made illegal in the lower to backend contract?

Yes, that can be the mechanism. I was mainly guarding to ensure that we are not prematurely lowering at the ingress point vs further down. When done too high up, lowerings like that need to be truly universal and not lose information.

frafranz:
> > > > > @qedawkins does this need to be an optional decomposition? There are many ways to implement these and we need to ensure that the frontend doesn't prematurely lower them universally.
> > > > 
> > > > 
> > > > In traditional Aten fashion it's really hard to tell the difference between these two ops. At least wrt to the lowering to Linalg, as far as I can tell `AtenMaxDimOp` and `AtenMinDimOp` is exactly what I would expect from an arg___ anyway. StableHLO appears to have a lowering for `AtenArgmaxOp`, but I don't know if that's just an artifact from before the argmax lowering was added.
> > > > So this is probably fine globally.
> > > 
> > > 
> > > Great, thanks. That's what I was looking to verify.
> > 
> > 
> > Just to join in the discussion: What exactly do you mean by optional decomposition? Optional in the sense that the op is not decomposed if argmin is not made illegal in the lower to backend contract?
> 
> Yes, that can be the mechanism. I was mainly guarding to ensure that we are not prematurely lowering at the ingress point vs further down. When done too high up, lowerings like that need to be truly universal and not lose information.

Ah thanks, that’s good to know!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -885,13 +886,26 @@ class DecomposeAtenArgMaxOp : public OpRewritePattern<AtenArgmaxOp> {
       input = rewriter.create<AtenFlattenUsingIntsOp>(loc, flattenType, input,
                                                       dim, end);
     }
-    Value maxResult =
+
+    static_assert(std::is_same<OpTy, AtenArgmaxOp>() ||
+                  std::is_same<OpTy, AtenArgminOp>());
+    Value resultArg;
+    if constexpr (std::is_same<OpTy, AtenArgmaxOp>()) {
```
qedawkins, line 893: I think a more typical way to do this is to template the conversion target as well, that way you don't need this constexpr. For example, see the implementation of the pooling lowerings: https://github.com/llvm/torch-mlir/blob/6244f301fb063be15dc5df27f77567fb8852e31f/lib/Conversion/TorchToLinalg/Pooling.cpp#L377

projects/pt1/python/torch_mlir_e2e_test/test_suite/argmin.py
```diff
@@ -0,0 +1,72 @@
+#  Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
```
qedawkins, line 1: You can move all of these tests to `reduction.py`. [torch.max](https://pytorch.org/docs/stable/generated/torch.max.html) actually returns the index as well (I assume the same is true for min) so testing of related ops is hidden in the reduction test suite.

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -445,6 +445,9 @@ def aten〇std〇correction〡shape(self: List[int], dim: Optional[List[int]] =
 def aten〇argmax〡shape(self: List[int], dim: Optional[int] = None, keepdim: bool = False) -> List[int]:
     return upstream_shape_functions.argmax(self, dim, keepdim)
 
+def aten〇argmin〡shape(self: List[int], dim: Optional[int] = None, keepdim: bool = False) -> List[int]:
+    return upstream_shape_functions.argmax(self, dim, keepdim)
```
qedawkins, line 449: Is there an `argmin` shape function? I assume it would do the same thing but would be a little less strange to look at :)

projects/pt1/python/torch_mlir_e2e_test/test_suite/argmin.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+@register_test_case(module_factory=lambda: ArgminModule())
+def ArgminModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4))
+
+@register_test_case(module_factory=lambda: ArgminModule())
+def ArgminModule_multiple_mins(module, tu: TestUtils):
+    """To cover the special case that the minimal value occurs more than once.
+    The pytorch convention is here to consider the first occurence as the argmax.
```
qedawkins, line 34: nit: argmax -> argmin

qedawkins: Also for testing this, I would use integers instead. It can double as a test for integer types as well.

qedawkins (changes requested):
Looks good, a few comments/nits.

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
frederik-h: There is none. see [here](https://github.com/pytorch/pytorch/blob/fcf6a76108be8e7b6db528b631a7b8ebdc7470ac/torch/jit/_shape_functions.py#L1068). I had the same issue with the aten.min.dim operator recently. I decided that it is not worth the effort to add it to pytorch, but perhaps I was wrong. @frafranz  Perhaps add a comment here to explain that using the argmax shape function for argmin is intentional?

frafranz: That is a good idea, I will add that comment!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
frafranz: Thanks, I already had the feeling there was a more elegant solution for this. Will change this

projects/pt1/python/torch_mlir_e2e_test/test_suite/argmin.py
frafranz: Will do that, then I should move the argmax tests as well which are so far in a separate file.

projects/pt1/python/torch_mlir_e2e_test/test_suite/argmin.py
frafranz: Thanks a lot for catching this, will change to integers.

frafranz: And will add an analogous test for argmax

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
qedawkins: Comment is fine, definitely don't go to the effort of adding to pytorch. I just figured it was worth asking, thanks for the attention to detail!

qedawkins (approved):
LGTM! Thanks for the patch :)

I can merge it for you once the checks pass.

PR #2615, opened 2023-12-06
[TorchToLinalg] Fix integer type handling for aten.mm

qedawkins:
Despite aten.mm requiring the input and output types match, we still opt to maintain signedness semantics in case later passes try to do any sort of integer type narrowing.


vivekkhandelwal1 (approved):
LGTM

PR #2617, opened 2023-12-07
[TorchToLinalg] Drop constexpr from ifs in argmin/max.dim

qedawkins:
MSVC-19 does not support constexprs of lambda captured constexpr values like this: https://godbolt.org/z/ej65rMzdr
Instead, this just drops the constexpr from the if statements.

See the discussion in https://discord.com/channels/689900678990135345/1062405112292712499/1182338050664185999


Groverkss (approved):
LGTM

PR #2619, opened 2023-12-08
Advance llvm-project and stablehlo.

stellaraccident:
llvm-project: bbd2b08b95fe76bea138c1b03c1cd42ed3ee04df
stablehlo: ab709fe48de88c67717abfbd7ef17425eb95ddaf

These commits were chosen in order to account for an MLIR API break from https://github.com/llvm/llvm-project/commit/3dbac2c007c114a720300d2a4d79abe9ca1351e7 which required a patch to stablehlo. We integrate a bit beyond that commit to deal with some revert/reapply cycles in the intervening range which were discovered in another downstream.

Further, it requires adaptation to the stablehlo API breaks introduced from https://github.com/openxla/stablehlo/pull/1872 which are along for the ride.

Since some stablehlo builders were changed to directly take int64_t array refs, also traced that up some call stacks to eliminate some signed/unsigned mismatches that result.

Also adds a few TOSA tests to the passing set that seem to work now.

### DIFF
--- a/.github/workflows/buildAndTest.yml
+++ b/.github/workflows/buildAndTest.yml
@@ -17,7 +17,7 @@ concurrency:
   # (postsubmit). The workflow name is prepended to avoid conflicts between
   # different workflows.
   group: ${{ github.workflow }}-${{ github.event.number || github.sha }}
-  cancel-in-progress: true
+  #cancel-in-progress: true
 
 
 # Provisioned Jobs:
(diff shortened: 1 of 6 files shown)


qingyunqu (approved):
LGTM

eric-k256:
I don't see anything wrong in what you've done. I built this PR locally, and can see the 5 tests unexpectedly pass. Switching to your commit adding them to the pass list and everything runs with no failures.

 I retriggered the github checks to make sure it wasn't a weird glitch specific to that build.

stellaraccident:
> I don't see anything wrong in what you've done. I built this PR locally, and can see the 5 tests unexpectedly pass. Switching to your commit adding them to the pass list and everything runs with no failures.
> 
> I retriggered the github checks to make sure it wasn't a weird glitch specific to that build.

I think this is a difference between nightly PT and stable. Trying something else.

Issue #2622, opened 2023-12-08
Some linalg tests failing on Conv2d grouped/dilation for PyTorch nightly after LLVM bump


stellaraccident:
In the LLVM bump (https://github.com/llvm/torch-mlir/pull/2619), we started seeing these three tests failing for linalg and dynamo:

```
        "Conv2dWithPaddingDilationStrideStaticModule_grouped",
        "Conv2dWithPaddingDilationStrideStaticModule_grouped_multiplier",
        "ConvolutionModule2DGroups_basic",
```

I added them to the LINALG_XFAIL_SET and TORCHDYNAMO_XFAIL_SET for PyTorch nightly (denoted by version >= 2.2.0.dev20231204). These work on stable.

These are failing with numerical issues, and I do not immediately understand how bumping the LLVM version in isolation would result in *just* the PyTorch nightly failing and not the stable. Possibly something subtly different that is sending it down a bad path. Should be triaged and fixed.

ubfx:
Might this have to do with this PR: https://github.com/llvm/llvm-project/pull/73855 ?

I will try and submit a patch for the grouped conv2d lowering in torch-mlir.

PR #2623, opened 2023-12-08
Grouped conv2d: Use MLIR Op which matches memory layout of weight dimensions

ubfx:
The linalg Op `linalg.conv_2d_ngchw_fgchw` had a bug where

1. Weights were accessed as G,F,C,H,W instead of as F,G,C,H,W
2. Output was accessed as N,F,G,H,W instead of as N,G,F,H,W

Now this has been fixed in https://github.com/llvm/llvm-project/pull/73855 which broke the torch-mlir lowering to that Op. 

This patch switches lowering in torch-mlir to the newly introduced `linalg.conv_2d_ngchw_gfchw` op which accesses weights in an order that is compatible with PyTorch's memory layout.

Fix https://github.com/llvm/torch-mlir/issues/2622

### DIFF
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -848,15 +848,14 @@ public:
                                                       indices);
       };
 
-      // expand F,C,H,W -> F/G,G,C,H,W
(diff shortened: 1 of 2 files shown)


vivekkhandelwal1 (approved):
LGTM

Issue #1471, opened 2022-10-10
Support returning single element tuple 


ZihengJiang:
For example:
```
class Mod(torch.nn.Module):

    def forward(self, a):
       return (torch.tanh(a), )
```

When such module get compiled with torch-mlir, it will raise such error:

```
torch_mlir.compiler_utils.TorchMlirCompilerError: Lowering TorchScript IR -> Torch Backend IR failed with the followingdiagnostics:
error: failed to legalize unresolved materialization from '!torch.tuple<tensor>' to '!torch.tensor' that remained live after conversion
note: see current operation: %3 = "builtin.unrealized_conversion_cast"(%2) : (!torch.tuple<tensor>) -> !torch.tensor
note: see existing live user here: "func.return"(%3) : (!torch.tensor) -> ()
```

ramiro050:
Hi @ZihengJiang, in torch-mlir there is no support for returning a tuple with a single element in it because at the backend contract level, the IR produced would be impossible to differentiate from the IR produced by a function that returns a single tensor.

For example, if you have a function

```python
def some_func(...):
    return (a, b, c)
```

this becomes the following in MLIR:

```mlir
func @some_func(...) -> (!torch.vtensor, !torch.vtensor, !torch.vtensor) { ... }
```

Therefore, one would expect a function

```python
def some_func(...):
    return (a,)
```

to become:

```mlir
func @some_func(...) -> (!torch.vtensor) { ... }
```

However, this is the same MLIR that we would expect from a function that returns a single tensor. Because the signature is ambiguous, we can only support one of the cases, and torch-mlir chose to support returning a single tensor rather than a tuple with a single element in it.

ZihengJiang:
Hi @ramiro050 , thanks for the clarification! In this case, can we convert the single element tuple to a single tensor automatically instead of raising the error?

silvasean:
> Hi @ramiro050 , thanks for the clarification! In this case, can we convert the single element tuple to a single tensor automatically instead of raising the error?

We cannot just automatically do this. It would have to be done by the user before their code reaches Torch-MLIR. We have some code to do this on FX graphs that we have been copying from project to project: https://github.com/iree-org/iree-torch/blob/94911c8fc068d135de2a810699d399e10b1cc43f/torchdynamo_poc/utils.py#L55

ZihengJiang:
um, ok. In my case, I only have some torch scripted models so modifying them might be hard. Thank you!

[...]

PR #2628, opened 2023-12-11
Fix return tuple with one element

qingyunqu:


ramiro050:
See: https://github.com/llvm/torch-mlir/issues/1471#issuecomment-1273913856

PR #2630, opened 2023-12-12
Sort the TOSA passing test list

### LINKED ISSUE #2620: Convert TOSA passing test set to xfail set
The TOSA e2e test set has this comment:

```
# Write the TOSA set as a "passing" set as it is very early in development
# and very few tests work yet.
```

And then it lists a few hundred passing tests in a big unsorted list. Recommend converting this to an xfail set or (at a minimum) sorting the list.

eric-k256:
For easier tracking of issues, sort the TOSA passing list. It is still significantly smaller then the XFAIL list would be.

Resolves #2620, at least until the xfail list gets smaller than the passing list.


stellaraccident (approved):
Thanks

PR #2632, opened 2023-12-12
lower torch.diagonal op

frafranz:
The new code compiles, but all of my tests fail with the following message:

```
<unknown>:0: error: unsupported by backend contract: tensor with unknown rank
<unknown>:0: note: see current operation: %2 = "torch.tensor_static_info_cast"(%arg0) : (!torch.vtensor<[?,?],f32>) -> !torch.vtensor
<unknown>:0: note: this is likely due to a missing transfer function in abstract_interp_lib_gen.py
// -----// IR Dump After LowerToBackendContract Failed (torch-lower-to-backend-contract) //----- //
module attributes {torch.debug_module_name = "DiagonalModule"} {
  func.func @forward(%arg0: !torch.vtensor<[?,?],f32>) -> !torch.vtensor {
    %int0 = torch.constant.int 0
    %int1 = torch.constant.int 1
    %0 = torch.tensor_static_info_cast %arg0 : !torch.vtensor<[?,?],f32> to !torch.vtensor
    %1 = torch.copy.to_tensor %0 : !torch.tensor
    %2 = torch.aten.diagonal %1, %int0, %int0, %int1 : !torch.tensor, !torch.int, !torch.int, !torch.int -> !torch.tensor
    %3 = torch.copy.to_vtensor %2 : !torch.vtensor
    return %3 : !torch.vtensor
  }
}
```


vivekkhandelwal1:
Hi @frafranz, you need to add this op similar to: https://github.com/llvm/torch-mlir/commit/5e9582b055ff3fda1529f0a376fca75b72da8ac5

Specifically, add your op to this list: https://github.com/llvm/torch-mlir/blob/5e9582b055ff3fda1529f0a376fca75b72da8ac5/lib/Dialect/Torch/Utils/Utils.cpp#L200, and then re-build and try to run the test.

frafranz:
> Hi @frafranz, you need to add this op similar to: [5e9582b](https://github.com/llvm/torch-mlir/commit/5e9582b055ff3fda1529f0a376fca75b72da8ac5)
> 
> Specifically, add your op to this list:
> 
> https://github.com/llvm/torch-mlir/blob/5e9582b055ff3fda1529f0a376fca75b72da8ac5/lib/Dialect/Torch/Utils/Utils.cpp#L200
> 
> , and then re-build and try to run the test.

Ah I missed that, thanks @vivekkhandelwal1 ! Will give it a try

frafranz:
> > Hi @frafranz, you need to add this op similar to: [5e9582b](https://github.com/llvm/torch-mlir/commit/5e9582b055ff3fda1529f0a376fca75b72da8ac5)
> > Specifically, add your op to this list:
> > https://github.com/llvm/torch-mlir/blob/5e9582b055ff3fda1529f0a376fca75b72da8ac5/lib/Dialect/Torch/Utils/Utils.cpp#L200
> > 
> > , and then re-build and try to run the test.
> 
> Ah I missed that, thanks @vivekkhandelwal1 ! Will give it a try

This solved indeed the error. Now however my tests run infinitely. Looking at the mlir dumps, I noticed that this is still in the torch.nn dialect, which I did not see previously in debugging (and gives errors when I try to call mlir-opt on this dump)

```
func.func private @__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule.forward(%arg0: !torch.nn.Module<"__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule"> loc(unknown), %arg1: !torch.tensor {torch.type_bound = !torch.vtensor<[?,?],f32>} loc(unknown)) -> !torch.tensor {
    %int1 = torch.constant.int 1 loc(#loc)
    %int0 = torch.constant.int 0 loc(#loc)
    %1 = torch.aten.diagonal %arg1, %int0, %int0, %int1 : !torch.tensor, !torch.int, !torch.int, !torch.int -> !torch.tensor loc(#loc2)
    return %1 : !torch.tensor loc(#loc)
  } loc(#loc)
  torch.class_type @__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule {
    torch.attr private "training" : !torch.bool loc(#loc)
    torch.attr private "_is_full_backward_hook" : !torch.optional<bool> loc(#loc)
    torch.method "forward", @__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule.forward loc(#loc)
  } loc(#loc)
  %true = torch.constant.bool true loc(#loc)
  %none = torch.constant.none loc(#loc)
  %0 = torch.nn_module {
    torch.slot "training", %true : !torch.bool loc(#loc)
    torch.slot "_is_full_backward_hook", %none : !torch.none loc(#loc)
  } : !torch.nn.Module<"__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule"> loc(#loc)
} loc(#loc)
#loc1 = loc("/proj/rdi/staff/frhaniel/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/diagonal.py":24:15)
#loc2 = loc("aten::diagonal"(#loc1))
```

vivekkhandelwal1:
> > > Hi @frafranz, you need to add this op similar to: [5e9582b](https://github.com/llvm/torch-mlir/commit/5e9582b055ff3fda1529f0a376fca75b72da8ac5)
> > > Specifically, add your op to this list:
> > > https://github.com/llvm/torch-mlir/blob/5e9582b055ff3fda1529f0a376fca75b72da8ac5/lib/Dialect/Torch/Utils/Utils.cpp#L200
> > > 
> > > , and then re-build and try to run the test.
> > 
> > Ah I missed that, thanks @vivekkhandelwal1 ! Will give it a try
> 
> This solved indeed the error. Now however my tests run infinitely. Looking at the mlir dumps, I noticed that this is still in the torch.nn dialect, which I did not see previously in debugging (and gives errors when I try to call mlir-opt on this dump)
> 
> ```
> func.func private @__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule.forward(%arg0: !torch.nn.Module<"__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule"> loc(unknown), %arg1: !torch.tensor {torch.type_bound = !torch.vtensor<[?,?],f32>} loc(unknown)) -> !torch.tensor {
>     %int1 = torch.constant.int 1 loc(#loc)
>     %int0 = torch.constant.int 0 loc(#loc)
>     %1 = torch.aten.diagonal %arg1, %int0, %int0, %int1 : !torch.tensor, !torch.int, !torch.int, !torch.int -> !torch.tensor loc(#loc2)
>     return %1 : !torch.tensor loc(#loc)
>   } loc(#loc)
>   torch.class_type @__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule {
>     torch.attr private "training" : !torch.bool loc(#loc)
>     torch.attr private "_is_full_backward_hook" : !torch.optional<bool> loc(#loc)
>     torch.method "forward", @__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule.forward loc(#loc)
>   } loc(#loc)
>   %true = torch.constant.bool true loc(#loc)
>   %none = torch.constant.none loc(#loc)
>   %0 = torch.nn_module {
>     torch.slot "training", %true : !torch.bool loc(#loc)
>     torch.slot "_is_full_backward_hook", %none : !torch.none loc(#loc)
>   } : !torch.nn.Module<"__torch__.torch_mlir_e2e_test.test_suite.diagonal.DiagonalModule"> loc(#loc)
> } loc(#loc)
> #loc1 = loc("/proj/rdi/staff/frhaniel/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/diagonal.py":24:15)
> #loc2 = loc("aten::diagonal"(#loc1))
> ```

There might be an issue with the lowering you have added. @AmosLewis, can you please help here in debugging?

AmosLewis:
It looks like it's not the test run infinitely. You have got the mlir dump from the e2e tests. Which means your tests fail and test run has stop not run infinitely. And based on the mlir file you got, you can notice that there is torch.tensor , which means it is a torchscript mlir. If it has torch.vtensor, it is a torch backend mlir. So the lowering stop at torchscript. So the bug is when you lowering torchscript to torch backend/torch dialect. To figure out using which command to debug, you can look at the end of you e2e tests [lower torch.diagonal op (WIP) · llvm/torch-mlir@1b603a6 (github.com)](https://github.com/llvm/torch-mlir/actions/runs/7179569116/job/19550014088?pr=2632).  If you scroll all the way down, you will find 4 e2e test you add failed. And it mentioned how to debug it : torch-mlir-opt -pass-pipeline='builtin.module(torchscript-module-to-torch-backend-pipeline{backend-legal-ops=aten.flatten.using_ints extra-library=})' /tmp/DiagonalWithOffsetModule.mlir I will add -debug and maybe the -mlir-print-ir-after-all -mlir-disable-threading most of time. They will print where is the failure is notified. It should in the cpp code you added.

stellaraccident:
Is this patch still making progress? (It is in draft state so I have been ignoring)

frafranz:
It is, I was busy with other topics in between. Will get back to it after the holidays!

frafranz:
With the recent commit, the lowering passes all previous e2e test cases and the two new ones I added.

qedawkins:
> With the recent commit, the lowering passes all previous e2e test cases and the two that I added.
> 
> Note that I used a `tensor::ExtractOp` inside a `linalg::GenericOp`, while my initial idea was to define suitable affine maps so that no additional extraction is needed within the generic but only a `linalg::YieldOp`. Taking a 2D matrix as an example, I had in mind something like `affine_map<(d0, d1) -> (d0, d0)>` for the input and `affine_map<(d0, d1) -> (d0)>` for the output, so that `output[d0] = input[d0, d0]`. But during my experiments I got the impression that affine maps cannot be used for cases like this one. So now, again for the 2D case, I use a default one-dimensional affine map for the output, do not pass an input to the generic and perform the extraction as `output[d0] = input[d0, d0]`.
> 
> I don't see any issues concerning performance with the solution using extract, since the number of extracted values coincides with the dimensionality of the output. Please let me know if you think otherwise @stellaraccident @qedawkins @ramiro050

Affine maps can be used like that, but I would not be surprised of later parts of the stack (especially with linalg) get tripped up by it. `tensor.extract` makes sense to me for this op, and typically it's easier to see whether there is a better approach after collecting some data (i.e. land this and run some models with it).

As a rule of thumb, having something that runs at all is infinitely faster than having nothing, so IMO it's fine for first iterations of things to not hit all the happiest paths, unless we're doing something obviously bad or regressing performance somehow.

frafranz:
I have just added one more e2e test, to verify that also negative parameters `dim1` and `dim2` are parsed correctly.

frafranz:
> > With the recent commit, the lowering passes all previous e2e test cases and the two that I added.
> > Note that I used a `tensor::ExtractOp` inside a `linalg::GenericOp`, while my initial idea was to define suitable affine maps so that no additional extraction is needed within the generic but only a `linalg::YieldOp`. Taking a 2D matrix as an example, I had in mind something like `affine_map<(d0, d1) -> (d0, d0)>` for the input and `affine_map<(d0, d1) -> (d0)>` for the output, so that `output[d0] = input[d0, d0]`. But during my experiments I got the impression that affine maps cannot be used for cases like this one. So now, again for the 2D case, I use a default one-dimensional affine map for the output, do not pass an input to the generic and perform the extraction as `output[d0] = input[d0, d0]`.
> > I don't see any issues concerning performance with the solution using extract, since the number of extracted values coincides with the dimensionality of the output. Please let me know if you think otherwise @stellaraccident @qedawkins @ramiro050
> 
> Affine maps can be used like that, but I would not be surprised of later parts of the stack (especially with linalg) get tripped up by it. `tensor.extract` makes sense to me for this op, and typically it's easier to see whether there is a better approach after collecting some data (i.e. land this and run some models with it).
> 
> As a rule of thumb, having something that runs at all is infinitely faster than having nothing, so IMO it's fine for first iterations of things to not hit all the happiest paths, unless we're doing something obviously bad or regressing performance somehow.

Yes, that makes sense. Thanks for sharing your view!

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1834,6 +1834,147 @@ class ConvertAtenViewAsRealOp : public OpConversionPattern<AtenViewAsRealOp> {
...
+    for (auto i = 0; i < inputRank; i++)
+      if (i==dim1)
+        // TODO: limits will change here, just from 0 to diagSize, pass this somehow?
+        inExprs.push_back(diagStart1+getAffineDimExpr(dim1, rewriter.getContext())); // used operator overload instead of add constant function for affine dim expr
+      else if (i==dim2)
+        // TODO: limits will change here, just from 0 to diagSize, pass this somehow?
+        inExprs.push_back(diagStart2+getAffineDimExpr(dim1, rewriter.getContext())); // used operator overload
+      else
+        inExprs.push_back(getAffineDimExpr(i, rewriter.getContext())); // iterate over unchanged dims
+
```
frederik-h: I think you could just use `AffineMap::getMultiDimIdentityMap(inputRank, context);` to get the inExprs map and then `auto d = inExprs.getResult(i);` `inExprs.replace(d , diagStart1 + d, inputRank, 0);` and the same for diagStart 2. Perhaps less efficient (which should not matter at all here), but easier to read.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1834,6 +1834,147 @@ class ConvertAtenViewAsRealOp : public OpConversionPattern<AtenViewAsRealOp> {
...
+          rewriter.create<arith::MinSIOp>(loc, rewriter.create<arith::SubIOp>(
+              loc,
+              dim2Size, 
+              rewriter.create<arith::ConstantOp>(loc, rewriter.getI64IntegerAttr(offset))
+          ), 
+          dim1Size), 
+          rewriter.create<arith::ConstantOp>(loc, rewriter.getI64IntegerAttr(0))
+      );
+    
+    // TODO: below version also works, need to cast diagSize to Value below
```
frederik-h: Remove TODO? Version below seems good to me.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1834,6 +1834,147 @@ class ConvertAtenViewAsRealOp : public OpConversionPattern<AtenViewAsRealOp> {
...
+      return rewriter.notifyMatchFailure(op, "dim1 out of range");
+    dim2 = toPositiveDim(dim2, inputRank);
+    if (!isValidDim(dim2, inputRank))
+      return rewriter.notifyMatchFailure(op, "dim2 out of range");
+    if (dim1 == dim2) 
+      return rewriter.notifyMatchFailure(op, "diagonal dimensions cannot be identical");
+
+    bool offsetNegative = offset < 0;
+    Value dim1Size, dim2Size;
+    dim1Size = getDimOp(rewriter, loc, adaptor.getSelf(), dim1);
```
frederik-h: There is a  `getTensorSizes` function to get all sizes at once. `inputSizes[dim]` is  a bit more readable than using getDimOp in a complex expression as you do below.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1834,6 +1834,147 @@ class ConvertAtenViewAsRealOp : public OpConversionPattern<AtenViewAsRealOp> {
...
+      if (i==dim1)
+        // TODO: limits will change here, just from 0 to diagSize, pass this somehow?
+        inExprs.push_back(diagStart1+getAffineDimExpr(dim1, rewriter.getContext())); // used operator overload instead of add constant function for affine dim expr
+      else if (i==dim2)
+        // TODO: limits will change here, just from 0 to diagSize, pass this somehow?
+        inExprs.push_back(diagStart2+getAffineDimExpr(dim1, rewriter.getContext())); // used operator overload
+      else
+        inExprs.push_back(getAffineDimExpr(i, rewriter.getContext())); // iterate over unchanged dims
+    
+    // for diagonal result, one of the dims is deleted and one is moved to the end
```
frederik-h: Use upper case first letter and period at the end as in your other comments above?

frederik-h (commented):
Looks good to me, but I'll request other reviews for approval.

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -59,6 +59,36 @@ def aten〇triu〡shape(self: List[int], diagonal: int = 0) -> List[int]:
...
+    ErrorInvocation(TensorOfShape(2, 3, 4), dim1=1, dim2=1), # `dim1` and `dim2` equal.
+    ErrorInvocation(TensorOfShape(2, 3, 4), dim1=3, dim2=1), # `dim1` out of bounds.
+])
+def aten〇diagonal〡shape(self: List[int], offset: int = 0, dim1: int = 0, dim2: int = 1) -> List[int]:
+    assert len(self) >= 2, "input must have at least two dimensions"
+    dim1 = upstream_shape_functions.maybe_wrap_dim(dim1, len(self))
+    dim2 = upstream_shape_functions.maybe_wrap_dim(dim2, len(self))
+    assert dim1 != dim2, "diagonal dimensions cannot be identical"
+    
+    diagonal: List[int] = []
```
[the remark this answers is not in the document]
frafranz: That's true, I will consider it when I work on other changes.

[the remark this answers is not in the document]
frafranz: Yes, I can change that.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1834,6 +1834,147 @@ class ConvertAtenViewAsRealOp : public OpConversionPattern<AtenViewAsRealOp> {
...
+        loc, getAsOpFoldResult(outputDims), elementType);
+    
+    SmallVector<AffineExpr> inExprs;
+    SmallVector<AffineExpr> diagExprs;
+    // Input indexing (i1, ... , in, j1, j2) -> (i1, ... , in, j1+off1, j1+off2) (also ranges have to change with this, is this done automatically?)
+    // Output indexing (i1, ... , in, j1, j2) -> (i1, ... , in, j1), i.e. output indexing is just reduction of last dim
+    for (auto i = 0; i < inputRank; i++)
+      if (i==dim1)
+        // TODO: limits will change here, just from 0 to diagSize, pass this somehow?
+        inExprs.push_back(diagStart1+getAffineDimExpr(dim1, rewriter.getContext())); // used operator overload instead of add constant function for affine dim expr
```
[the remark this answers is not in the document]
frafranz: This refers to an outdated version of my code, the passage looks different after the last commits (and there shouldn't be TODOs left anymore)

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1834,6 +1834,147 @@ class ConvertAtenViewAsRealOp : public OpConversionPattern<AtenViewAsRealOp> {
...
+    Value outMatrix = rewriter.create<tensor::EmptyOp>(
+        loc, getAsOpFoldResult(outputDims), elementType);
+    
+    SmallVector<AffineExpr> inExprs;
+    SmallVector<AffineExpr> diagExprs;
+    // Input indexing (i1, ... , in, j1, j2) -> (i1, ... , in, j1+off1, j1+off2) (also ranges have to change with this, is this done automatically?)
+    // Output indexing (i1, ... , in, j1, j2) -> (i1, ... , in, j1), i.e. output indexing is just reduction of last dim
+    for (auto i = 0; i < inputRank; i++)
+      if (i==dim1)
+        // TODO: limits will change here, just from 0 to diagSize, pass this somehow?
```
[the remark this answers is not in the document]
frafranz: This as well refers to an older commit of DataMovement.cpp, the loops look different now (the loop over the diagonal has size diagSize)

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1834,6 +1834,147 @@ class ConvertAtenViewAsRealOp : public OpConversionPattern<AtenViewAsRealOp> {
...
+    outputDims.push_back(diagSize); // TODO: use this or line below depending on which implementation we use above
+    // outputDims.push_back(rewriter.create<arith::ConstantOp>(loc, rewriter.getI64IntegerAttr(diagSize)));
+    
+    Value outMatrix = rewriter.create<tensor::EmptyOp>(
+        loc, getAsOpFoldResult(outputDims), elementType);
+    
+    SmallVector<AffineExpr> inExprs;
+    SmallVector<AffineExpr> diagExprs;
+    // Input indexing (i1, ... , in, j1, j2) -> (i1, ... , in, j1+off1, j1+off2) (also ranges have to change with this, is this done automatically?)
+    // Output indexing (i1, ... , in, j1, j2) -> (i1, ... , in, j1), i.e. output indexing is just reduction of last dim
```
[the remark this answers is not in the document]
frafranz: You're right, this sounds clearer. After the recent commits, this is solved differently.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1834,6 +1834,147 @@ class ConvertAtenViewAsRealOp : public OpConversionPattern<AtenViewAsRealOp> {
...
+        outputDims.push_back(getDimOp(rewriter, loc, adaptor.getSelf(), i));
+    outputDims.push_back(diagSize); // TODO: use this or line below depending on which implementation we use above
+    // outputDims.push_back(rewriter.create<arith::ConstantOp>(loc, rewriter.getI64IntegerAttr(diagSize)));
+    
+    Value outMatrix = rewriter.create<tensor::EmptyOp>(
+        loc, getAsOpFoldResult(outputDims), elementType);
+    
+    SmallVector<AffineExpr> inExprs;
+    SmallVector<AffineExpr> diagExprs;
+    // Input indexing (i1, ... , in, j1, j2) -> (i1, ... , in, j1+off1, j1+off2) (also ranges have to change with this, is this done automatically?)
```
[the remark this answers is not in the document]
frafranz: This comment is not part of the newer commits (and the indexing is done differently now)

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1834,6 +1834,147 @@ class ConvertAtenViewAsRealOp : public OpConversionPattern<AtenViewAsRealOp> {
...
+    // Output indexing (i1, ... , in, j1, j2) -> (i1, ... , in, j1), i.e. output indexing is just reduction of last dim
+    for (auto i = 0; i < inputRank; i++)
+      if (i==dim1)
+        // TODO: limits will change here, just from 0 to diagSize, pass this somehow?
+        inExprs.push_back(diagStart1+getAffineDimExpr(dim1, rewriter.getContext())); // used operator overload instead of add constant function for affine dim expr
+      else if (i==dim2)
+        // TODO: limits will change here, just from 0 to diagSize, pass this somehow?
```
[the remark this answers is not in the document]
frafranz: This is solved differently in the newer commits.

lib/Conversion/TorchToLinalg/DataMovement.cpp
frafranz: This would have been indeed more readable. In the newer commits I used a MultiDimIdentityMap

lib/Conversion/TorchToLinalg/DataMovement.cpp
frafranz: In the newer commits, I have deleted the below alternative and just worked with the version, in which computations / case distinctions are made with the rewriter.

lib/Conversion/TorchToLinalg/DataMovement.cpp
frafranz: Thanks, that is a good remark. The comment does not occur anymore in the recent commits

lib/Conversion/TorchToLinalg/DataMovement.cpp
frafranz: Thanks for the hint, I could change this together with other modifications.

frafranz:
Just ran CI again after rebase on the latest main, from my side this PR is ready to be merged but let me know if you think otherwise! @ramiro050 @qedawkins

ramiro050 (approved):
LGTM, thanks!

PR #2614, opened 2023-12-06
Implement e2e support for aten.acos op

frederik-h:
See also issue #2604.

The implementation requires acos support in the MLIR Math dialect. I have created the PR llvm/llvm-project#74584 to implement this. Marking this PR as a draf until the LLVM core PR gets merged.

PR #2633, opened 2023-12-12
Revert "Implement e2e support for aten.acos op"

vivekkhandelwal1:
Reverts llvm/torch-mlir#2614


frederik-h:
Thanks for the help! But it seems that the commits are not responsible for the failure. The CI run for the PR was also successful before I merged it and I was just able to build a fresh clone of the repository locally, including those commits.

vivekkhandelwal1:
I don't think we need to revert this PR. The CI failure is due to some machine related issues.

PR #2634, opened 2023-12-12
[onnx] Lowerings from `onnx.selu`

rsuderman:
Lowerings for `selu` lowerings for ONNX to the corresponding torch implementations.
Torch has no `selu` implementation and instead uses a generalize `elu` with the input scale
set to `1.0`.

### DIFF
--- a/include/torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h
+++ b/include/torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h
@@ -113,6 +113,25 @@ struct OpBinder {
     return failure();
   }
 
+  ParseResult f32FloatAttr(float &value, StringRef nameSuffix,
+                           float defaultValue = 0.0f) {
+    SmallString<64> name("torch.onnx.");
+    name.append(nameSuffix);
+    auto attr = op->getAttr(name);
+    if (!attr) {
+      value = defaultValue;
+      return success();
+    }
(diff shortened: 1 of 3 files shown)


rsuderman:
> Hi Rob, onnx.relu lowering has already been addressed here: #2601. We are using this issue to monitor which ops have already been taken/implemented: [nod-ai/SHARK-Turbine#215](https://github.com/nod-ai/SHARK-Turbine/issues/215).

Sounds good, I can remove. We should probably avoid checking off implementations before landing a PR and mark as in progress with a issue / username. It makes it a bit tricky when working on grouped operations.

include/torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h
```diff
@@ -113,6 +113,25 @@ struct OpBinder {
     return failure();
   }
 
+  ParseResult f32FloatAttr(float &value, StringRef nameSuffix,
+                           float defaultValue = 0.0f) {
```
wu-s-john, line 117: Nice, I needed something like this for another task!

wu-s-john (approved):

vivekkhandelwal1 (approved):
LGTM

include/torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h
rsuderman: Awesome! I'll land now so you can pull it in.

PR #2635, opened 2023-12-12
Implement lowering of aten.cosh op.

godot73:


test/Conversion/TorchToLinalg/elementwise.mlir
ramiro050, line 1: No need to have this test. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

ramiro050 (approved):
Looks good, just one small change

ramiro050:
The CI is failing because `AbstractInterpLib.cpp` does not match the autogenerated one. Run `./build_tools/update_abstract_interp_lib.sh` to regenerate the file.

godot73:
Another CI test was failing around the GeneratedTorchOps.td. Resolved it by editing projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/torch_ods_gen.py instead of the .td file.

PR #2637, opened 2023-12-13
[onnx] Add torch-mlir-import-onnx tool.

stellaraccident:
Simple Python console script to import an ONNX protobuf to the torch dialect for additional processing.

For installed wheels, this can be used with something like:

```
torch-mlir-import-onnx test/python/onnx_importer/LeakyReLU.onnx
```

Or from a dev setup:

```
python -m torch_mlir.tools.import_onnx ...
```


test/python/onnx_importer/import_onnx_tool.runlit
```diff
@@ -0,0 +1,3 @@
+# RUN: %PYTHON -m torch_mlir.tools.import_onnx %S/LeakyReLU.onnx | FileCheck %s
```
vivekkhandelwal1, line 1: I think there's some typo here with `%S/%s`. Rest looks good.

vivekkhandelwal1 (approved):

test/python/onnx_importer/import_onnx_tool.runlit
stellaraccident: Believe it or not, no this is right :/

Capital 'S' refers to the directory that contains the file (lower case 's').

I have to look it up every time: https://llvm.org/docs/CommandGuide/lit.html#substitutions

vivekkhandelwal1: Thanks for the info. I think I need to learn these macros.

PR #2641, opened 2023-12-14
[onnx] Lowerings from `onnx.transpose`

rsuderman:
Lowerings for `transpose` from ONNX to `aten`. Implementation depends on making multiple `aten.transpose`
operations swapping pairs of dimensions. As `onnx.transpose` can swap around any dimensions it may
require constructing multiple `aten.transpose`.


dan-garvey (approved):
if you can fix the merge conflict and ping me I'll reapprove. 

I like the use of swap here = )

stellaraccident (approved):

rsuderman:
I should be good to land without a re-approve. Usually merges do not invalidate.

PR #2642, opened 2023-12-14
[onnx] Lowerings from `onnx.tan`

rsuderman:
Started work on the `tan` lowerings for ONNX to Torch. Uses `sin` and `cos` to represent a `tan`.


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -26,4 +26,20 @@ using namespace mlir::torch::onnx_c;
...
+      "Tan", 7, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value operand;
+        if (binder.tensorOperand(operand) ||
+            binder.tensorResultType(resultType))
+          return failure();
+        auto loc = binder.op->getLoc();
+        Value sin = rewriter.create<Torch::AtenSinOp>(loc, resultType, operand);
+        Value cos = rewriter.create<Torch::AtenCosOp>(loc, resultType, operand);
+        rewriter.replaceOpWithNewOp<Torch::AtenDivTensorOp>(
```
vivekkhandelwal1, line 41: You can directly create `AtenTanOp`.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
rsuderman: `AtenTanOp` had not been added and I was not certain whether it was an `aten` operation in `torch`. Found a reference and added `AtenTanOp` with the corresponding lowering to `math.tan`

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1311,7 +1315,7 @@ class ConvertElementwiseOp : public ConversionPattern {
   LogicalResult
   matchAndRewrite(Operation *op, ArrayRef<Value> operands,
                   ConversionPatternRewriter &rewriter) const override {
-    if (!isa<AtenTanhOp, AtenReluOp, AtenPreluOp, AtenGeluOp,
+    if (!isa<AtenTanOp, AtenTanhOp, AtenReluOp, AtenPreluOp, AtenGeluOp,
```
ramiro050, line 1318: You also need to add the op to

https://github.com/llvm/torch-mlir/blob/6fbd8ea0c8a97e57d03cd3598cef5941d14f93c4/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L1966

ramiro050 (changes requested):
Changes that add e2e support for new torch ops need dtype/shape functions and e2e tests. Similar to: https://github.com/llvm/torch-mlir/pull/2635

More info in the docs:
- https://github.com/llvm/torch-mlir/blob/main/docs/adding_abstract_interpretation_functions.md
- https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md

lib/Conversion/TorchToLinalg/Uncategorized.cpp
rsuderman: I believe I have the remaining boilerplate based on the sample PR.

ramiro050 (approved):
Thanks!

PR #2643, opened 2023-12-14
[onnx] Add support for `onnx.sinh`

rsuderman:
Adds a lowering from `onnx.sinh` to `aten.sinh`. This includes adding the `aten.sinh` operator.


vivekkhandelwal1:
I think it's better to add the missing Torch->Linalg support as well for this op. Otherwise, we will have an issue in the downstream pipeline. Also, I think this op is missing from the mlir. You need to add it there as well, like: https://github.com/llvm/llvm-project/pull/74584

rsuderman:
> I think it's better to add the missing Torch->Linalg support as well for this op. Otherwise, we will have an issue in the downstream pipeline. Also, I think this op is missing from the mlir. You need to add it there as well, like: [llvm/llvm-project#74584](https://github.com/llvm/llvm-project/pull/74584)

Review is sent to upstreamed llvm [pr](https://github.com/llvm/llvm-project/pull/75517). Either I can shelve this into it lands and integrates or we can land partial support. Either works for me.

stellaraccident (approved):

PR #2644, opened 2023-12-14
Update PYTHONPATH in development.md

yinrun:
Modify PYTHONPATH to new related directory in docs.

### DIFF
--- a/build_tools/write_env_file.sh
+++ b/build_tools/write_env_file.sh
@@ -13,7 +13,7 @@ portable_realpath() {
 
 td="$(portable_realpath "$(dirname "$0")"/..)"
 build_dir="$(portable_realpath "${TORCH_MLIR_BUILD_DIR:-$td/build}")"
-python_packages_dir="$build_dir/python_packages"
+python_packages_dir="$build_dir/tools/torch-mlir/python_packages"
 
 write_env_file() {
   echo "Updating $build_dir/.env file"
--- a/docs/development.md
+++ b/docs/development.md
@@ -115,7 +115,7 @@ export PYTHONPATH=`pwd`/build/tools/torch-mlir/python_packages/torch_mlir:`pwd`/
 ### Windows PowerShell
 
 ```shell
-$env:PYTHONPATH = "$PWD/build/python_packages/torch_mlir;$PWD/projects/pt1/examples"
+$env:PYTHONPATH = "$PWD/build/tools/torch-mlir/python_packages/torch_mlir;$PWD/projects/pt1/examples"
 ```
 
 ## Testing MLIR output in various dialects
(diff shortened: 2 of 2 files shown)


stellaraccident (commented):
This was unintended fallout to require this change. I'd like to track through and figure out why it is different. Wdyt? Fix it to how it was or just accept that it is different now and land this doc update?

yinrun:
> This was unintended fallout to require this change. I'd like to track through and figure out why it is different. Wdyt? Fix it to how it was or just accept that it is different now and land this doc update?

I am not sure what's you CMAKE_CURRENT_BINARY_DIR returns in the main branch, in my environment, it's related to 
`PWD`/torch-mlir/build/tools/torch-mlir. Can you help to check?

dan-garvey:
So I also checked this and the path updated to is correct for me. It seems like the last person to update it did it with an out of tree llvm build. I'm not sure if that would change it. 

Regardless, I don't think it needs to be fixed to how it was unless its breaking anything.

zjgarvey:
In case this is the preferred fix, here are other places to correct:

https://github.com/llvm/torch-mlir/blob/main/docs/development.md?plain=1#L118
https://github.com/llvm/torch-mlir/blob/main/docs/development.md?plain=1#L129 (with an additional typo to be corrected)
https://github.com/llvm/torch-mlir/blob/main/build_tools/write_env_file.sh#L16

srcarroll:
I made the most recent change to the documentation.  Sorry if it's wrong or I'm misunderstanding the follow up discussions here, but I'm a bit confused. The top level cmake sets `TORCH_MLIR_PYTHON_PACKAGES_DIR` to `${CMAKE_CURRENT_BINARY_DIR}/python_packages` by default unless specified by externally ( https://github.com/llvm/torch-mlir/blob/main/CMakeLists.txt#L170). Since this is in the top level cmake then the current binary dir is `build`.  Verified that it is set to `build/python_packages` on latest main.

Both the `torch_mlir` and `torch_mlir_e2e` python modules use this variable for their build locations (https://github.com/llvm/torch-mlir/blob/main/python/CMakeLists.txt#L96 and https://github.com/llvm/torch-mlir/blob/main/projects/pt1/python/torch_mlir_e2e_test/CMakeLists.txt#L11).  Again, for me on latest main, these are put in `build/python_packages/torch_mlir` as expected.  I have no `build/tools/torch_mlir`.

The changes I made work for me (aside from the typo @zjgarvey pointed out) and are consistent with the cmake structure mentioned above, including `e2e_test.sh` so I don't know what I'm missing.  I wonder if anyone here is setting `TORCH_MLIR_PYTHON_PACKAGES_DIR` manually (I am not)?  I am using an out of tree llvm (which is updated to what torch-mlir expects), but can't imagine how that affects this. 

Moreover, since this variable can be set differently by downstream users, it might be a good idea for the documentation to reflect that and if the path is going to be hardcoded relative to a build dir in `write_env_file.sh`, then it should be default. Otherwise, it should be simple enough to make it more general.

dan-garvey (approved):
seems fine to merge. If someone changes the layout in CMAKE at some point we can always change it then. May as well be correct in the docs for now

stellaraccident:
I'm fine with fixing the docs to what actually works. I also think there are things we can improve about how this is done, but that is orthogonal.

PR #2650, opened 2023-12-15
[TorchToLinalg] Lower aten.cat to tensor.concat

qedawkins:
This replaces the lowering of aten.cat with tensor.concat, allowing more efficient handling of concatenations in downstream flows. The refbackend populates concat decomposition patterns that can be used to recover the previous lowering.

Depends on #2654


dan-garvey (approved):
@qedawkins  do we need to add this to any of the default pass pipelines inside of torch-mlir? If not please merge.

qedawkins:
> @qedawkins do we need to add this to any of the default pass pipelines inside of torch-mlir? If not please merge.

Shouldn't need any other torch-mlir changes. If other downstream users were expecting the existing decomposition, the pass added to the refbackend should reproduce that exact lowering and is implemented upstream.

PR #2659, opened 2023-12-16
[aten] Make `torch.aten.matmul` to `linalg` work for non-broadcasting case

rsuderman:
Broadcasting for `torch.aten.matmul` is optional so a MxN with NxK
matmul should be legalized to a `linalg.matmul`.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -191,8 +191,10 @@ public:
     Value lhs = adaptor.getSelf();
     Value rhs = adaptor.getOther();
 
-    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter))) {
+      llvm::errs() << "NOT VALID\n";
       return failure();
+    }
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -191,8 +191,10 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
     Value lhs = adaptor.getSelf();
     Value rhs = adaptor.getOther();
 
-    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter))) {
+      llvm::errs() << "NOT VALID\n";
       return failure();
```
ramiro050, line 196: is the `llvm::errs()` here a remnant from debugging? If not, I think we should use `notifyMatchFailure` to be consistent with other patterns.

ramiro050 (approved):
Other than the one question below, LGTM

lib/Conversion/TorchToLinalg/Linear.cpp
rsuderman: Woops, slipped through. Deleted

PR #2661, opened 2023-12-16
[RFC] general support for Adaptive Pooling Ops

zjgarvey:
Adaptive pooling ops can only be decomposed into their non-adaptive counterparts in trivial cases. 

For example, the current decomposition for AtenAdaptiveAvgPool1dOp in DecomposeComplexOps.cpp supports outSize = inSize (i.e., do literally nothing), and outSize = 1 (i.e., do a batched average). 

The reason adaptive pooling ops are difficult to lower to linalg is that they are not constantly strided. They are computed by taking an input tensor of shape (N, C, Hin), and an output size Hout, and computing the output tensor at position (n,c, h) in the following way:

1. compute st(h) = (h*Hin)//Hout
2. compute en(h) = 1 + ((h+1)*Hin -1)//Hout
3. apply a computation (max or avg) to the slice:  INPUT[n, c, st(h):en(h)]

The provided sample implementation (for ConvertAtenAdaptiveAvgPool1dOp) uses tensor.extract to access the input tensor inside the payload of a linalg generic op. This is likely an unattractive use of linalg generic ops, which is why I am asking for some more targeted feedback on the validity of this approach before attempting to support the many other adaptive pooling ops. 

Specifically:

- Is the performance of this implementation bad enough to warrant targeting different dialects entirely? e.g. TMtensor/linalg ext/ etc.
- If the provided implementation is of acceptable performance to the community, then is it permissable to remove the Adaptive pooling decompositions from DecomposeComplexOps.cpp? Based on the current structure of the -torch-decompose-complex-ops pass, it does not seem possible to only decompose the adaptive ops in special cases (it seems to get stuck in an infinite loop on a match failure). I would be happy to instead incorporate the case logic into the conversion directly, and remove the decompositions once they are rendered completely obsolete. 

As long as this approach is acceptable, I can clean up the implementation with some helper functions, and quickly add support for each of the remaining Adaptive pooling ops.


qedawkins:
Thanks for the PR! Some comments before taking a deeper look at the code.
> Adaptive pooling ops can only be decomposed into their non-adaptive counterparts in trivial cases.
> 
> For example, the current decomposition for AtenAdaptiveAvgPool1dOp in DecomposeComplexOps.cpp supports outSize = inSize (i.e., do literally nothing), and outSize = 1 (i.e., do a batched average).
> 
> The reason adaptive pooling ops are difficult to lower to linalg is that they are not constantly strided. They are computed by taking an input tensor of shape (N, C, Hin), and an output size Hout, and computing the output tensor at position (n,c, h) in the following way:
> 
>     1. compute st(h) = (h*Hin)//Hout
> 
>     2. compute en(h) = 1 + ((h+1)*Hin -1)//Hout
> 
>     3. apply a computation (max or avg) to the slice:  INPUT[n, c, st(h):en(h)]
> 
> 
> The provided sample implementation (for ConvertAtenAdaptiveAvgPool1dOp) uses tensor.extract to access the input tensor inside the payload of a linalg generic op. This is likely an unattractive use of linalg generic ops, which is why I am asking for some more targeted feedback on the validity of this approach before attempting to support the many other adaptive pooling ops.

I wouldn't quite say `tensor.extract` is unattractive, but it certainly isn't a happy path. Just that when you need it, you need it. Also without working out all the details myself, the other way to do something like this is with some indexing map tricks (e.g. `affine_map<(d0, d1, d2) -> (d0, d1 + d2)>` and then masking off the out of range values, kind of like you're doing now, except without tensor.extract).

> 
> Specifically:
> 
>     * Is the performance of this implementation bad enough to warrant targeting different dialects entirely? e.g. TMtensor/linalg ext/ etc.

I think this comes down to seeing workloads. Having a default/fallback lowering anyway is always helpful + reusable.

> 
>     * If the provided implementation is of acceptable performance to the community, then is it permissable to remove the Adaptive pooling decompositions from DecomposeComplexOps.cpp? Based on the current structure of the -torch-decompose-complex-ops pass, it does not seem possible to only decompose the adaptive ops in special cases (it seems to get stuck in an infinite loop on a match failure). I would be happy to instead incorporate the case logic into the conversion directly, and remove the decompositions once they are rendered completely obsolete.
> 

I haven't followed all of the work surrounding decompositions, but per my understanding, it should be possible to make certain decompositions opt-in, and I'd expect that's what applies here. StableHLO and TOSA might still be relying on the decomposition in which case we can't remove it just yet (haven't checked, just guessing).

> 
> As long as this approach is acceptable, I can clean up the implementation with some helper functions, and quickly add support for each of the remaining Adaptive pooling ops.

High level the lowering looks reasonable and I don't have an immediate alternative suggestion (again probably need to take a closer look). Plus supporting the op, even if not at the speed of light, is worth having. So in short, I don't have an immediate answer, but normally these things bounce up and down the stack a few times before we land on new ops and/or start disabling other functional paths.

zjgarvey:
Thanks for the response, @qedawkins.

> I wouldn't quite say tensor.extract is unattractive, but it certainly isn't a happy path. Just that when you need it, you need it. Also >without working out all the details myself, the other way to do something like this is with some indexing map tricks (e.g. >affine_map<(d0, d1, d2) -> (d0, d1 + d2)> and then masking off the out of range values, kind of like you're doing now, except >without tensor.extract).

That's a good point, I'll look into the affine map idea, at least in the static-dim case. If it is possible to do at all, then the strides for the "minimum_start_index" and "maximum_end_index" would still be directly computed from the input and output sizes. If either of those are dynamic dims, I'm not sure if a dynamic stride value is allowed in a linalg generic's affine maps. 

> I haven't followed all of the work surrounding decompositions, but per my understanding, it should be possible to make certain > decompositions opt-in, and I'd expect that's what applies here. StableHLO and TOSA might still be relying on the decomposition > in which case we can't remove it just yet (haven't checked, just guessing).

I think in the meantime, I'll leave the existing decompositions (for adaptive avg 1 and 2 d), but include all of the adaptive pool lowerings here even if they don't get used right away. It will at least be a pure benefit for those ops, like adaptive max, which have no support whatsoever yet.

If you end up taking a closer look and see anything that could be improved, I would greatly appreciate the advice.

dan-garvey:
@zjgarvey you'll have to check the builder error, there are several test suites you will not have tested locally. You'll either need to add them as xfails or fix as appropriate

lib/Conversion/TorchToLinalg/PopulatePatterns.h
```diff
@@ -45,6 +45,9 @@ void populateLinearPatternsAndLegality(TypeConverter &typeConverter,
 void populatePoolingPatternsAndLegality(TypeConverter &typeConverter,
                                         RewritePatternSet &patterns,
                                         ConversionTarget &target);
+void populateAdaptivePoolingPatternsAndLegality(TypeConverter &typeConverter,
+                                        RewritePatternSet &patterns,
```
ramiro050, line 49: nit: indentation is off. Make sure to format all your changes with `git clang-format`

ramiro050:
> I haven't followed all of the work surrounding decompositions, but per my understanding, it should be possible to make certain decompositions opt-in, and I'd expect that's what applies here. StableHLO and TOSA might still be relying on the decomposition in which case we can't remove it just yet (haven't checked, just guessing).

Correct. You can add the op to the list below to prevent it from decomposing

https://github.com/llvm/torch-mlir/blob/be3e74b647375d1c72f075bebf77dc31e74477a2/projects/pt1/python/torch_mlir/__init__.py#L243-L253

**Note** this is only for `torch_mlir.compile` (which is used by our e2e tests). There is a TODO there about moving this list deeper into torch-mlir so that it is independent of the Python API. On the command line, you can specify the op as

https://github.com/llvm/torch-mlir/blob/be3e74b647375d1c72f075bebf77dc31e74477a2/projects/pt1/python/torch_mlir/__init__.py#L453

And from C++ like this

https://github.com/llvm/torch-mlir/blob/be3e74b647375d1c72f075bebf77dc31e74477a2/lib/Dialect/Torch/Transforms/Passes.cpp#L133

projects/pt1/python/torch_mlir_e2e_test/test_suite/pooling.py
```diff
@@ -776,6 +864,25 @@ def AvgPool1dStaticModule_basic(module, tu: TestUtils):
 
 # ==============================================================================
 
+class AdaptiveAvgPool1dGeneralDynamic(torch.nn.Module):
+    
+    def __init__(self):
+        super().__init__()
+        self.aap1d = torch.nn.AdaptiveAvgPool1d(7)
```
ramiro050, line 871: nit: can you use named arguments to make easier to tell what the 7 means? Same comment for the other tests

lib/Conversion/TorchToLinalg/AdaptivePooling.cpp
```diff
@@ -0,0 +1,237 @@
...
+  LogicalResult
+  matchAndRewrite(AtenAdaptiveAvgPool1dOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+
+    Location loc = op->getLoc();
+    const TypeConverter *typeconverter = getTypeConverter();
+
+    // input tensor and output shape
+    Value input = adaptor.getSelf();
+    Value outputShape = adaptor.getOutputSize();
```
ramiro050, line 69: In MLIR conversions, the `adaptor` gives access to the operands in the dialect _after_ conversion (if it's possible to convert the operand). For example, `op.getSelf()` returns a `!torch.vtensor<...>`, but `adaptor.getSelf()` returns a `tensor<...>`.

This means that for helper functions that work in the torch dialect world, any arguments used should be obtained from `op` instead of `adaptor`. (In this case, `getListConstructElements` works at the torch dialect level, so you should pass `op.getOutputSize()`).

Note: things work here because there is no conversion for `!torch.list`s, so they stay in the torch dialect world, but it is still a good practice to use `op` for the input dialect and `adaptor` for the target dialect.

lib/Conversion/TorchToLinalg/AdaptivePooling.cpp
```diff
@@ -0,0 +1,237 @@
...
+    RankedTensorType OutputType =
+        typeconverter->convertType(op.getResult().getType())
+            .cast<RankedTensorType>();
+
+    // get rank of input (same as rank of output)
+    int64_t rank = input.getType().cast<RankedTensorType>().getRank();
+    // input operand should be NCH (i.e. rank 3)
+    if (rank != 3) {
+      return rewriter.notifyMatchFailure(op, "only supports input type NCH");
+    }
```
ramiro050, line 87: When possible, error checking should happen _before_ any IR is created (functions that take a `rewriter` as an argument, like `getDimOp`, generate IR). Otherwise we risk leaving the graph in an invalid state after a match failure. This check should be one of the very first things this patterns does.

lib/Conversion/TorchToLinalg/AdaptivePooling.cpp
```diff
@@ -0,0 +1,237 @@
...
+      Kiter = rewriter.create<tensor::EmptyOp>(
+          loc, RankedTensorType::get({ShapedType::kDynamic}, dummyType), Kmax);
+    } else {
+      int64_t HinInt = InputType.getShape()[rank - 1];
+      int64_t HoutInt = OutputType.getShape()[rank - 1];
+      int64_t correction = ((HinInt - 1) % HoutInt == 0) ? 0 : 1;
+      int64_t Kmax = 1 + (HinInt - 1) / HoutInt + correction;
+      Kiter = rewriter.create<tensor::EmptyOp>(
+          loc, RankedTensorType::get({Kmax}, dummyType), ValueRange{});
+    }
```
ramiro050, line 112: No need to have special handling for static shapes. If the sizes are static, a backend compiler should be able to easily optimize away the ops generated in the dynamic case.

lib/Conversion/TorchToLinalg/AdaptivePooling.cpp
```diff
@@ -0,0 +1,237 @@
...
+      dynOutputSizes.push_back(HoutIndex);
+      kwTensor = rewriter.create<tensor::EmptyOp>(
+          loc, RankedTensorType::get({ShapedType::kDynamic}, elementType),
+          HoutIndex);
+    } else {
+      kwTensor = rewriter.create<tensor::EmptyOp>(
+          loc,
+          RankedTensorType::get({OutputType.getShape()[rank - 1]}, elementType),
+          ValueRange{});
+    }
```
ramiro050, line 141: You can use the helpers in

https://github.com/llvm/torch-mlir/blob/be3e74b647375d1c72f075bebf77dc31e74477a2/lib/Conversion/Utils/Utils.cpp#L124-L131

to create tensors without having to worry about dynamic/static cases.

projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -20,7 +20,11 @@
     # 'linalg.depthwise_conv_2d_nchw_chw' op inferred input/output operand #1 has shape's dimension #0 to be 4, but found 8
     "Conv2dWithPaddingDilationStrideStaticModule_depthwise_multiplier",
     "IscloseStaticModule_basic",
-    "IscloseStaticModuleTrue_basic"
+    "IscloseStaticModuleTrue_basic",
+    "AdaptiveMaxPool2dNonUnitOutputSizeStaticModule_basic",
+    "AdaptiveMaxPool2dNonUnitOutputSizeDynamicModule_basic",
+    "AdaptiveMaxPool2dUnitOutputSizeStaticModule_basic",
+    "AdaptiveMaxPool2dUnitOutputSizeDynamicModule_basic"
```
ramiro050, line 27: Is the plan for this patch to also add 2d support? If not, it makes more sense to me to wait until 2d support is added to add these tests. It makes it easier when reviewing patches to see what cases are being tested.

ramiro050 (changes requested):
Thanks for this! I haven't thought too much about the algorithm yet. But I do have some initial comments regarding style/best practices.

lib/Conversion/TorchToLinalg/AdaptivePooling.cpp
zjgarvey: Yeah, that's a good point. I was playing with this to see what the adaptor would give, and forgot to change it back.

lib/Conversion/TorchToLinalg/AdaptivePooling.cpp
zjgarvey: Ah! This is very helpful, thanks.

lib/Conversion/TorchToLinalg/AdaptivePooling.cpp
zjgarvey: I am now using createInitTensor, and the OpFoldResult builder for EmptyOp in a few places now. Very helpful, thank you.

projects/pt1/e2e_testing/xfail_sets.py
zjgarvey: Perhaps I should make this a draft pr? I did actually intend to implement all the adaptive ops right away, as long as people didn't think the general idea was too gross. 

If I can make some helper functions to do a lot of the boilerplate stuff in this conversion, it would be very easy to write conversions for all of the other adaptive ops. 

What do you think would be better? Just getting this pull request through and then making a new pr for the other adaptive ops, or put all of them in one pr?

projects/pt1/python/torch_mlir_e2e_test/test_suite/pooling.py
zjgarvey: Let me know if my correction addresses this properly for the cited example. If so, I'll add the same kind of change to other adaptive pooling tests.

zjgarvey:
Ah, one minute. The last commit broke some tests and I didn't realize it since I was building just torch-mlir-opt.

zjgarvey:
> Ah, one minute. The last commit broke some tests and I didn't realize it since I was building just torch-mlir-opt.

The tests should pass now: I needed to update the linalg generic's output shape to accommodate the earlier simplifications.

projects/pt1/e2e_testing/xfail_sets.py
dan-garvey: In general smaller patches are better. But as ramiro suggested, I'd keep the tests with the commit that adds the code it tests.

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/AdaptivePooling.cpp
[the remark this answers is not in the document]
zjgarvey: Okay, I can do that.

zjgarvey:
Thanks again, @ramiro050 , for the helpful comments. 

I believe I addressed each of your comments in the last few commits. Although I had to do a bit of work to make the conversion compatible with the decomposition:

1. Since the decomposition pass actually had a RuntimeAssertOp instead of a match failure, I decided to change some of the functionality there in order to allow it to pass to my conversion. 
2. I decided to add a case to the decomposition where the input size and output size are static and the output size evenly divides the input size (in which case the strides are static). 
3. I opted to remove the case where the input size and output size are possibly dynamic, but equal. This is truly an uninteresting case (imo: anyone trying to use an adaptive pooling op as an identity operator deserves the minor drop in performance). Unless this breaks something elsewhere, removing that trivial case makes the rest of the case logic for the decomposition much simpler.
4. I moved some match failures to occur as early as possible
5. I used clang-format on the DecomposeComplexOps.cpp file.

I removed the AdaptiveMaxPool2d tests as requested, but added a few tests for AdaptiveAvgPool1d to account for the modifications made to the decomposition.

zjgarvey:
```
Unexpected outcome summary: (make_fx_tosa)

****** Unexpectedly Passed tests - 1 tests
    XPASS - "AdaptiveAvgPool1dStaticEvenMultiple_basic"

Summary:
    Passed: 351
    Expectedly Failed: 742
    Unexpectedly Passed: 1
Error: Process completed with exit code 1.
```
Is the solution to this issue to simply add "AdaptiveAvgPool1dStaticEvenMultiple_basic" to [xfail_sets.py#L1323](https://github.com/llvm/torch-mlir/blob/main/projects/pt1/e2e_testing/xfail_sets.py#L1323):

```python
MAKE_FX_TOSA_PASS_SET = (TOSA_PASS_SET | {
### Tests additionally passing in make_fx_tosa
    "AdaptiveAvgPool1dNonUnitOutputSizeStaticModule_basic",
    "AdaptiveAvgPool1dUnitOutputSizeStaticModule_basic",
    "NativeGroupNormBackwardModule_basic",
    "SliceWholeTensorModule_basic",
    "TensorFloatModule_basic",
    "TensorIntModule_basic",
}) - 
```

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -459,8 +462,188 @@ class ConvertAtenAvgPoolOp : public OpConversionPattern<OpTy> {
...
+    RankedTensorType inputType = input.getType().cast<RankedTensorType>();
+    RankedTensorType outputType =
+        typeConverter->convertType(op.getResult().getType())
+            .cast<RankedTensorType>();
+
+    // get elementType of input tensor
+    Type elementType = inputType.getElementType();
+
+    // make an iteration space of size kMax = 1 + ceildiv (hIn - 1) , hOut
+    Type dummyType = rewriter.getI1Type();
```
ramiro050, line 530: nit: we should use a more descriptive name, such as `boolType`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4575,19 +4584,34 @@ class DecomposeAtenAdaptiveAvgPool1dOp
...
     unsigned rank = *maybeRank;
-    Value sizeDim = rewriter.create<Torch::ConstantIntOp>(
-        loc, rewriter.getI64IntegerAttr(rank - 1));
-    Value inputSize = rewriter.create<AtenSizeIntOp>(loc, input, sizeDim);
 
     Value outputShape = op.getOutputSize();
     SmallVector<Value> outputShapeSizesTorchInt;
     getListConstructElements(outputShape, outputShapeSizesTorchInt);
     Value outputSize = outputShapeSizesTorchInt[0];
+    int64_t outputSizeInt;
```
ramiro050, line 4598: I'm a bit confused by this. Why do we need to change `DecomposeCOmplexOps`? Your changes to the [backend legal ops set](https://github.com/llvm/torch-mlir/pull/2661/files#diff-ed49f70c396f47c0d651d573b089e78db6177242f2d9fd799cefee6212fc6036) should prevent this pattern from running when using `linalg` as the backend.

ramiro050 (changes requested):

ramiro050:
> Is the solution to this issue to simply add "AdaptiveAvgPool1dStaticEvenMultiple_basic" to [xfail_sets.py#L1323](https://github.com/llvm/torch-mlir/blob/main/projects/pt1/e2e_testing/xfail_sets.py#L1323):

Yep!

> I used clang-format on the DecomposeComplexOps.cpp file.


Make sure you use `git clang-format {YOUR_RANGE_OF_COMMITS}`. This ensures that the formatting only happens to your code changes, avoiding changing the rest of the file.

> Since the decomposition pass actually had a RuntimeAssertOp instead of a match failure, I decided to change some of the functionality there in order to allow it to pass to my conversion.
> I decided to add a case to the decomposition where the input size and output size are static and the output size evenly divides the input size (in which case the strides are static).
> I opted to remove the case where the input size and output size are possibly dynamic, but equal. This is truly an uninteresting case (imo: anyone trying to use an adaptive pooling op as an identity operator deserves the minor drop in performance). Unless this breaks something elsewhere, removing that trivial case makes the rest of the case logic for the decomposition much simpler.

This sounds to me like a step backwards (going from runtime checks to compile-time checks that depend on statically shaped inputs), which we should normally try to avoid. This patch should not really have to change anything in `DecomposeComplexOps.cpp`. We just need to remove prevent the `linalg` pipeline from using the pattern, which should happen by using the `backend_legal_ops` set.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
zjgarvey: I thought that while running linalg as the backend for the e2e tests, it still tries to apply the -torch-decompose-complex-ops pass. If it runs into the runtime assert op, it causes the test to get stuck in a loop and it doesn't pass the IR to my -torch-to-linalg conversion. Trying to revert the changes to just the DecomposeComplexOps.cpp file and running the e2e tests, the terminal just spits out:

```
unimplemented: only support cases where input and output size are equal for non-unit output size
```

and then sits forever. 

If I use a match failure instead, then it just keeps the unconverted IR, which is legal, and gets passed to my conversion and the tests all pass.

zjgarvey:
I'll try to summarize what exactly I wanted out of this (and what the PR currently does):

0. The torch-to-linalg conversion provided currently covers all cases.
1. The decomposition pass in DecomposeComplexOps.cpp should work when possible, and should pass to my conversion when not. I am doing this because the named linalg op is likely more performant than my conversion. If I am not mistaken, that is the result of adding the op to the [backend legal ops set](https://github.com/llvm/torch-mlir/pull/2661/files#diff-ed49f70c396f47c0d651d573b089e78db6177242f2d9fd799cefee6212fc6036).
2. The original decomposition only supports two edge cases, so I added support for the important cases where output_size evenly divides input_size.
3. My understanding is that the runtime assert op, in the original decomposition, is used to check that the possibly dynamic input size is equal to the static output size (in which case the decomposition will be valid, but reduces to an identity op). My conversion covers the case where the runtime assert fails, and this is the reason I removed the runtime assert and returned match failures instead.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
zjgarvey: Ah!

The reason for my confusion is that I spelled the op wrong in the backend legal ops set, so it basically wasn't doing anything.

Let me make some changes.

zjgarvey:
I think I've cleared up a misconception about the backend legal ops set vs. the [target.addIllegalOp](https://github.com/llvm/torch-mlir/pull/2661/files#diff-1276e4f8fd0cb432c26bd260869a85ef26061e347780ce6ded6ce39be9c8a139) in the LowerToBackendContract.cpp file (which I had commented out). From what I understand, there are two workable routes:

1. Uncomment the target.addIllegalOp<Adap...>, add aten.adaptive_avg_pool1d to backend legal ops set, and revert the decomposition to the original. This means the decomposition will not be used for the linalg backend.
2. Keep the target.addIllegalOp<Adap...> commented (or delete it), keep the decomposition changes (modulo formatting issue). This will actually use the decomposition, which may be desirable since the decomposition is likely faster in the cases where it applies.

Which one is preferable?

zjgarvey:
For the sake of progress, I opted to revert the changes to DecomposeComplexOps.cpp. If anyone thinks the alternative approach is desirable for performance purposes, I would be happy to put up a separate PR.

ramiro050 (approved):
Thanks for the hard work on this! LGTM

ramiro050:
> Keep the target.addIllegalOp<Adap...> commented (or delete it), keep the decomposition changes (modulo formatting issue). This will actually use the decomposition, which may be desirable since the decomposition is likely faster in the cases where it applies.

We currently don't really have a nice way of doing this (and in practice, it has not really come up. That is, we have not had any requests for having decompositions some of the time and direct lowerings the rest of the time). If we were to add this functionality, we should modify the Python API to allow users to specify which ops they want to _try_ decompositions on before resorting to direct lowerings. And we also need to figure out how to mark things legal/illegal properly.

Until necessity for such a feature comes up, using `backend_legal_ops` should suffice.

ramiro050:
I can merge this for you if you're happy with it.

zjgarvey:
> I can merge this for you if you're happy with it.

Thanks, I would appreciate that! And thanks again for all of the careful comments and advice.

PR #2664, opened 2023-12-16
[Doc] Fix typo + python_package build location change

jiangqucheng:
Based on the instructions above in `Development.md`, the build result of Python packages "torch_mlir", and "torch_mlir_e2e_test" didn't show up in `pwd/build/python_packages/torch_mlir`, but in another location that I added. 
I found this path has been changed quite often in recent commits. To keep safe, both paths are kept. I have tested, and adding both locations at the same time won't hurt. 
Also, change typo "tpython_packages", and add hyphens.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -42,7 +42,7 @@ cmake -GNinja -Bbuild \
   -DLLVM_TARGETS_TO_BUILD=host \
   externals/llvm-project/llvm
 ```
-The following additional quality of life flags can be used to reduce build time:
+The following additional quality-of-life flags can be used to reduce build time:
 * Enabling clang on Linux
 ```shell
   -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -109,13 +109,13 @@ cmake --build build
 ### Linux and macOS
 
 ```shell
-export PYTHONPATH=`pwd`/build/python_packages/torch_mlir:`pwd`/projects/pt1/examples
+export PYTHONPATH=`pwd`/build/python_packages/torch_mlir:`pwd`/build/tools/torch-mlir/python_packages/torch_mlir:`pwd`/projects/pt1/examples
```
wu-s-john, line 112: No wonder I had weird issues with running integration tests 😂

wu-s-john (approved):
Thanks for catching this. I was about to make the same PR as well.

dan-garvey:
I think we should probably fix the underlying problem in cmake that is causing the divergence of paths for different users. Can you two confirm the build command you use so we can triage? From what I understand, its an issue with out of tree llvm builds?

PR #2669, opened 2023-12-17
Allow printing all IR in `torch_mlir.compile`

rikhuijzer:
This PR adds the `enable_ir_printing` option to `torch_mlir.compile`, which can be used to print the IR for all intermediate passes.

When running the added test file via:
```shell
$ python test/python/compile.py 2> tiny.stderr
```
the file `tiny.stderr` is about 700 KB.


dan-garvey:
Thanks for contributing this!

PR #2676, opened 2023-12-19
Fix typo in DecomposeBernoulli() match failure messages.

godot73:


godot73:
These seem to be typos to me. Before copying the message, creating a PR to check.

ramiro050 (approved):
Good catch!

PR #2678, opened 2023-12-20
[onnx] `onnx.qlinear` decomposition to `aten` operations

rsuderman:
Explicitly decomposes the `qlinear` operation. It is possible to
use the existing `aten` operation for the scalar case however no
per-channel variant is available.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -7,7 +7,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/IR/ImplicitLocOpBuilder.h"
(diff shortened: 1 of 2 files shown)


stellaraccident (changes requested):
Let's discuss a bit more. While valid, I'd like to understand the lowering flow from here to make sure that we aren't reducing granularity at the frontend in a way that will make a pattern matching issue later on.

(If just trying to unblock some progress and need this while getting the rest worked out, feel free to proceed and note accordingly that a proper translation is pending a full plan for handling the quant schemes)

vivekkhandelwal1 (commented):
Hi @rsuderman, I think in this lowering you have made an assumption that the zero point and scale would be a tensor. What about the case when they could be scalars?

rsuderman:
> Hi @rsuderman, I think in this lowering you have made an assumption that the zero point and scale would be a tensor. What about the case when they could be scalars?

This is actually out of date so I am closing. They are required to be tensors AFAICT (either rank-0 or rank-1)

Issue #2646, opened 2023-12-14
Implement lowering of torch.aten.exponential to linalg.


godot73:
torch.aten.exponential is in the [op-to-implement](https://github.com/nod-ai/SHARK-Turbine/issues/119#issue-1971327299.) list.

godot73:
High-level direction:

When x is sampled from uniform(0, 1), f(x) := - ln(1-x)/lambda produces exponentially distributed random variables. Thus, we can lower torch.aten.exponential(x, lambda) by decomposing it into:

%one_minus_x = sub(%one - x)
%log_one_minus_x = log(%one_minus_x)
%neg_log_one_minus_x = neg(%log_one_minus_x)
%result = sub(%neg_log_one_minus_x, %lambda)

PR #2680, opened 2023-12-20
Implement lowering of torch.aten.exponential

godot73:
https://github.com/llvm/torch-mlir/issues/2646

Decompose aten.exponential() into: -exp(1-x)/lambda


vivekkhandelwal1 (approved):
LGTM

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3591,6 +3591,51 @@ class DecomposeAtenBernoulliTensorOp
...
+    Type resultType = op.getType();
+
+    // Create a uniform random op with low and high set to 0.0 and 1.0,
+    // respectively.
+    Value none = rewriter.create<ConstantNoneOp>(loc);
+    Value zero =
+        rewriter.create<ConstantFloatOp>(loc, rewriter.getF64FloatAttr(0.0));
+    Value one =
+        rewriter.create<ConstantFloatOp>(loc, rewriter.getF64FloatAttr(1.0));
+    Value emptyTensor = rewriter.create<AtenFullLikeOp>(
```
godot73, line 3619: Question about AtenFullLikeOp.

In other creation cases of AtenUniformOp, they first create AtenFullLikeOp or AtenEmptyMemoryFormatOp, and pass dtype, layout, device, etc from the op being replaced. On the other hand, here we are just passing none for all of them.

Is there any concern about this?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3591,6 +3591,51 @@ class DecomposeAtenBernoulliTensorOp
 };
 } // namespace
 
+namespace {
+// Decompose exponential() to do inverse transform sampling.
+// - https://en.wikipedia.org/wiki/Inverse_transform_sampling
+// With the exponential distribution, F(x) = 1 - exp(-lambda * x). Thus,
+// exponential() = - ln(1 - uniform(0, 1)) / lambda.
+class DecomposeAtenExponentialOp : public OpRewritePattern<AtenExponentialOp> {
```
stellaraccident, line 3599: While mathematically valid, I don't think we want to be taking such a lowering path for such an op. Specifically, creating a temporary tensor and populating it with a random distribution to represent this calculation is surprising.

I think that instead of trying to decompose, the target of this op should be an elementwise generic that bottoms out on the `math.exp` op (https://github.com/llvm/llvm-project/blob/main/mlir/include/mlir/Dialect/Math/IR/MathOps.td#L444). Such builtins are implemented on target devices with an algorithm of choice vs having that pre-determined to a (most probably not efficient) form at the frontend.

stellaraccident (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
godot73: Not sure I understand your intent. For a clarified discussion, let me first put together my understanding and my question below.

[1]
Many of deterministic calculus ops (like aten.exp(), aten.sin(), aten.tanh(), etc) are elementwise lowered from [Uncategorized.cpp](https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L1330) to the corresponding math dialect ops. 

Unlike those deterministic ops, aten.exponential() is a random number generator with the [exponential probability distribution](https://en.wikipedia.org/wiki/Exponential_distribution), which cannot be directly implemented through the math.exp op. A closer example would be [**AtenRandnGeneratorOp**](https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L5409), which is decomposed from the same DecomposeComplexOps.cpp file. Due to their stochastic behavior that produces varying output values, it seems to make sense to decompose them using [AtenUniformOp](https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L5442).

[2]
However, I still couldn't figure out what to do with `size`, `dtype`, etc. While aten::randn* ops have [placeholders for `size`, `dtype`, `layout`, ...](https://github.com/llvm/torch-mlir/blob/main/projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/torch_ods_gen.py#L385) but aten::exponential doesn't seem to have them, according to //include/torch-mlir/Dialect/Torch/IR/JITOperatorRegistryDump.txt.

JitOperator 'aten::exponential : (Tensor, float, Generator?) -> (Tensor)':
  MLIR op name = torch.aten.exponential
  MLIR cpp class name = AtenExponentialOp
  namespace = aten
  unqualified_name = exponential
  ...

Thus, my PR is passing none to all these optional operands. Please correct me if I'm mistaken about your suggested approach.

stellaraccident: Ah, I had completely misunderstood this op while taking a pass through. What you say makes sense, and sorry for the pushback. In general, I'm on the lookout for premature lowering and didn't look close enough.

This makes sense to me.

stellaraccident (approved):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
stellaraccident: Seems ok to me.

PR #2681, opened 2023-12-21
[fx] Upstream the turbine FxImporter to torch-mlir.

stellaraccident:
Changes made during upstreaming:

* Removed comments attributing some copied code back to torch-mlir (since it is now repatriated).
* Re-organized imports.
* Inlined RefMapping/RefTracker and TypeSubclassMap from an external utility module.
* Added FxImporter class comments.
* Updated stack trace extraction to be fail safe.
* Added an entry-point for `import_frozen_exported_program` which uses the shiny new upstream `torch.export.export()` API (versus the lower-level/older API that Turbine is presently using). This necessitated a small FX rewrite to line external state management up with current conventions.
* Adapted one of Turbine's importer tests to go with this initial submission. Turbine unfortunately has a lot of more-integration-ey tests, and I would like to extract those as more of unit tests of the importer features and upstream them that way vs trying to copy directly. For now, one overall test with the initial submission gets us moving.

I acknowledge that there are some code quality things that could be improved in this submission: this was authored over the course of many months (and often via some trial and error). I would like to keep it relatively converged with the downstream for the next few steps while getting the test suite upstreamed. And then it will be easier to take a hygienic pass through the code.

Including co-authors for contributors in the git log of the original repository.

### DIFF
--- a/test/python/fx_importer/basic_test.py
+++ b/test/python/fx_importer/basic_test.py
@@ -1,0 +1,80 @@
+# Copyright 2023 Advanced Micro Devices, Inc
+#
+# Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+# See https://llvm.org/LICENSE.txt for license information.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+# Also available under a BSD-style license. See LICENSE.
+
+# RUN: %PYTHON %s | FileCheck %s
+
+from typing import Optional
+
+import torch
+import torch.export
+import torch.nn as nn
+
+from torch_mlir.extras.fx_importer import FxImporter
+from torch_mlir import ir
+from torch_mlir.dialects import torch as torch_d
+
+
+def export_and_import(
+    f,
+    *args,
+    fx_importer: Optional[FxImporter] = None,
+    constraints: Optional[torch.export.Constraint] = None,
+    **kwargs,
+):
+    context = ir.Context()
+    torch_d.register_dialect(context)
+
+    if fx_importer is None:
+        fx_importer = FxImporter(context=context)
+    prog = torch.export.export(f, args, kwargs, constraints=constraints)
+    fx_importer.import_frozen_exported_program(prog)
+    return fx_importer.module_op
+
+
+def run(f):
+    print(f"{f.__name__}")
+    print("-" * len(f.__name__))
+    f()
+    print()
+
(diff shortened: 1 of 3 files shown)


rsuderman (approved):

test/python/fx_importer/basic_test.py
```diff
@@ -0,0 +1,80 @@
...
+    fx_importer: Optional[FxImporter] = None,
+    constraints: Optional[torch.export.Constraint] = None,
+    **kwargs,
+):
+    context = ir.Context()
+    torch_d.register_dialect(context)
+
+    if fx_importer is None:
+        fx_importer = FxImporter(context=context)
+    prog = torch.export.export(f, args, kwargs, constraints=constraints)
```
sjain-stanford, line 33: Nice, I was wondering if we'd have to take a `shark_turbine.aot` dependency (i.e. `export` or `CompiledModule` API) to get the dynamo exported program, but looks like not :)

Just for my understanding, could you clarify/document what's different between the native `torch.export.export` in comparison to Turbine's AOT exporter, in terms of the constraints/knobs they expose (e.g. [dynamic dim constraints](https://pytorch.org/docs/stable/export.html#torch.export.dynamic_dim)). In other words, do you anticipate relying more directly on the `torch.export` path going forward for all of torch-mlir's use-cases?

stellaraccident: The turbine side adds a lot for assembling multi graph programs and having options to exploit iree features. But it does all of this via hooks into the fx_importer when needed.

Turbine uses a predecessor API to the now public torch `ExportedProgram` and will eventually upgrade to the same thing as here under the hood. The main difference is in how tensors from "outside" are handled. The predecessor API just let them flow through and the importer has to do a bit of work to resolve them. The`ExportedProgram` API does the extra accounting at the FX level. It also looks like it has the beginnings of new features like call graphs. Mainly a v0/V1 difference if I read it properly. Figured I'd just start upstream on the V1 thing.

stellaraccident:
> Thanks for getting this upstreamed so quickly! Excited to try this out.
> 
> [This is not blocking, please feel free to address in a follow-on] It would be nice to have docs similar to [onnx_importer.md](https://github.com/llvm/torch-mlir/blob/main/docs/importers/onnx_importer.md) for the uninitiated (like me).

Yes there is more we can do there. Followons.

PR #2682, opened 2023-12-21
[onnx] Lower onnx.HardSigmoid to torch

wu-s-john:
The expression for HardSigmoid in Onnx (https://onnx.ai/onnx/operators/onnx__HardSigmoid.html): max(0, min(1, alpha * x + beta))

is inherently different from HardSigmoid in Torch (https://pytorch.org/docs/stable/generated/torch.nn.Hardsigmoid.html) which is: if x < -3 -> 0
elif x > 3 -> 1
else x/6 + 1/2

That being said, it was just better to compute out the entire expression when translating the Onnx expression to Torch mlir, which is done in this PR. Some of the logic is shared from the files in `DecomposeComplexOps`. Therefore, refactored some shared logic between `DecomposeComplexOps` and `DefaultDomainGToP` and put it in a `Utils` file.


test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
```diff
@@ -198,4 +198,70 @@ func.func @test_less_or_equal(%arg0: !torch.vtensor<[3,4,5],f32>, %arg1: !torch.
...
+    // CHECK: %[[MIN_EXPRESSION:.*]] = torch.aten.minimum %[[ONE_TENSOR:.*]], %[[ALPHA_MULTI_X_PLUS_BETA:.*]] : !torch.vtensor<[],f32>, !torch.vtensor<[3,4,5],f32> -> !torch.vtensor<[3,4,5],f32>
+    // CHECK: %[[INT_0:.*]] = torch.constant.int 0
+    // CHECK: %[[TENSOR_DIMENSION_LIST_FOR_ZERO:.*]] = torch.prim.ListConstruct  : () -> !torch.list<int>
+    // CHECK: %[[NONE_FOR_ZERO:.*]] = torch.constant.none
+    // CHECK: %[[INT_TYPE_FOR_TENSOR_ZERO:.*]] = torch.constant.int 6
+    // CHECK: %[[ZERO_TENSOR:.*]] = torch.aten.full %[[TENSOR_DIMENSION_LIST_FOR_ZERO:.*]], %[[INT_0:.*]], %[[INT_TYPE_FOR_TENSOR_ZERO:.*]], %[[NONE_FOR_ZERO:.*]], %none_0, %none_0 : !torch.list<int>, !torch.int, !torch.int, !torch.none, !torch.none, !torch.none -> !torch.vtensor<[],f32>
+    // CHECK: torch.aten.maximum %[[ZERO_TENSOR:.*]], %[[MIN_EXPRESSION:.*]] : !torch.vtensor<[],f32>, !torch.vtensor<[3,4,5],f32> -> !torch.vtensor<[3,4,5],f32>
+  %0 = torch.operator "onnx.HardSigmoid"(%arg0) : (!torch.vtensor<[3,4,5],f32>) -> !torch.vtensor<[3,4,5],f32>
+  return %0 : !torch.vtensor<[3,4,5],f32>
 }
```
stellaraccident, line 267: Mind adding a terminating CR.

lib/Dialect/Torch/Utils/Utils.cpp
```diff
@@ -471,3 +471,32 @@ LogicalResult Torch::checkDefaultStrideHelper(Operation *op,
...
+  assert(inputType.hasDtype() && "input must have dtype");
+  SmallVector<int64_t> sizes;
+  BaseTensorType rank0TensorTy =
+      inputType.getWithSizesAndDtype(ArrayRef(sizes), inputType.getDtype())
+          .cast<BaseTensorType>();
+  Value dimList = rewriter.create<PrimListConstructOp>(
+      loc, Torch::ListType::get(Torch::IntType::get(inputType.getContext())),
+      ValueRange{});
+  return createInitTensor(rewriter, loc, rank0TensorTy, scalar, dimList);
+}
```
stellaraccident, line 502: Mind adding a terminating CR.

stellaraccident (approved):
Thanks. We may want to add something later that makes this optional: some folks may want to match the original and not deal with the decomposition, but until we get there, I'd rather be right than wrong.

stellaraccident (approved):

PR #2683, opened 2023-12-21
Mentioned helpful tooling to convert Onnx models to Torch MLIR

wu-s-john:
- Going through the `#torch-mlir` channel on the `llvm` discord, I realize that there are some useful commands that would be extremely helpful in creating Onnx lowers to Torch MLIR. Seems a lot of people are contributing to this. So, I thought it would be good to add this information to the docs.

These tools helped streamlined the development of this PR: https://github.com/llvm/torch-mlir/pull/2682


stellaraccident (approved):
Thank you for the doc update! Would you mind reflowing and breaking at 80 columns before we land?

stellaraccident (approved):

PR #2690, opened 2023-12-21
[onnx] Fix importer variable names to make `mlir` legal

rsuderman:
Some names for `onnx` identifiers are not legal in `mlir-ir`. Sanitize
so that the generated `ir` is legal.

### DIFF
--- a/python/torch_mlir/extras/onnx_importer.py
+++ b/python/torch_mlir/extras/onnx_importer.py
@@ -470,8 +470,6 @@ class ContextCache:
             name = "_" + name  
         return re.sub("[:/]", "_", name)  
 
-
-
     def tensor_proto_to_attr(self, tp: onnx.TensorProto) -> Attribute:
         tensor_type = self.tensor_proto_to_builtin_type(tp)
         if tp.HasField("raw_data"):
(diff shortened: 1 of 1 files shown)


python/torch_mlir/extras/onnx_importer.py
```diff
@@ -464,13 +464,23 @@ def type_proto_to_type(self, tp: onnx.TypeProto) -> IrType:
             # See TypeProto: sequence_type, map_type, optional_type, sparse_tensor_type.
             raise OnnxImportError(f"Unsupported ONNX TypeProto: {tp}")
 
+    def _sanitize_name(self, name):
+        try:
+            name = f"__{int(name)}__"
```
stellaraccident, line 469: I generally avoid exceptions in the critical/common path if feasible. How about:

```
if not name.isidentifier():
  name = "_" + name
return re.sub("[:/]", "_")
```

stellaraccident (approved):

python/torch_mlir/extras/onnx_importer.py
rsuderman: Done

PR #2693, opened 2023-12-22
Implement aten.reflection_pad2d lowering to linalg

frederik-h:


stellaraccident:
Checking whether this PR is ready for review or still being iterated (please use the "Ready for review" button if so).

frederik-h:
> Checking whether this PR is ready for review or still being iterated (please use the "Ready for review" button if so).

I would probably unify a few things (e.g. the handling of the "horizontal" and "vertical" dimensions?) and I am on vacation until next week and hence I would not respond to change requests. I think it is better to wait with the review.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -244,6 +244,283 @@ class ConvertAtenReflectionPad1dOp
...
+    auto createSubs = [&](std::initializer_list<Value> values) {
+      assert(values.size() >= 2);
+      return std::accumulate(values.begin() + 1, values.end(), data(values)[0],
+                             createSub);
+    };
+
+    // Enums for specifying the coordinates of a tile.  An "h" prefix
+    // is used to stand for "horizontal" and "v" for "vertical"
+    // throughout.
+    enum PadHLoc { LEFT = 0, RIGHT = 1, HCENTER = 2 };
```
frederik-h, line 327: There are many places in the implementation where I am doing the same thing for the "vertical" and "horizontal" dimension. I first thought that it might improve readability to keep this distinction. I have now changed my mind and I believe that using two dimensional arrays to unify the code for the vertical and horizontal dimensions would simplify things. For instance, assignments like `extractHOffset[LEFT] = one;` and `extractVOffset[LEFT] = one;` below would be replaced by `extractOffset[dim][FIRST] = ONE` inside a loop that iterates the dimensions. But I have not made the change yet. Perhaps it is no necessary now? It would make sense to do this for the implementation of reflection_pad3d. I think the 2d and 1d operators could then be decomposed into it.

frederik-h:
> > Checking whether this PR is ready for review or still being iterated (please use the "Ready for review" button if so).
> 
> I would probably unify a few things (e.g. the handling of the "horizontal" and "vertical" dimensions?) and I am on vacation until next week and hence I would not respond to change requests. I think it is better to wait with the review.

@stellaraccident I have decided not to implement the changes that I mentioned unless a reviewer thinks they are necessary because I think the implementation should be generalized to handle reflection_pad3d later and I assume that the 1d and 2d versions can be decomposed into it then. I have however fixed the lazy tensor core issue. That is, the PR is ready for review.

lib/Conversion/TorchToLinalg/DataMovement.cpp
qedawkins: I don't know the exact semantics of `reflection_pad3d` (and would have to see the decomposition you have in mind) but I would be surprised if a decomposition into pad3d would produce cleaner code than a direct lowering of 1d and 2d (fewer reshapes and such).

WRT the suggested changes here, I think it's fine to leave it as is and make the changes when adding support for 3d (if you're the one planning to implement support for 3d).

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -244,6 +244,283 @@ class ConvertAtenReflectionPad1dOp
...
+    //     x_1,1 x_1,2 ... x_1,m
+    //     x_2,1 x_2,2 ... x_2,m
+    //              .
+    //              .
+    //              .
+    //     x_n,1 x_n,2 ... x_n,m
+    //
+    // The padding tile consists of the columns 2, ..., m + 1
+    // of the input in reverse order. The first columns
+    // gets skipped because this this is the column trough
```
qedawkins, line 419: typo nit: columns -> column and trough -> through

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1831,6 +1851,20 @@ def aten〇reflection_pad1d〡dtype(self_rank_dtype: Tuple[int, int], padding: L
...
+                       ErrorInvocation(TensorOfShape(2, 3, 4), padding=[]),
+                       ErrorInvocation(TensorOfShape(2, 3, 4), padding=[2]),
+                       ErrorInvocation(TensorOfShape(2, 3, 4), padding=[2,1]),
+                       ErrorInvocation(TensorOfShape(2, 3, 4), padding=[3,2,1]),
+                       Invocation(TensorOfShape(5, 5, 4), padding=[1,2,3,4]),
+                       ErrorInvocation(TensorOfShape(2, 3, 4), padding=[5,4,3,2,1])])
+def aten〇reflection_pad2d〡dtype(self_rank_dtype: Tuple[int, int], padding: List[int]) -> int:
+    self_rank, self_dtype = self_rank_dtype
+
+    assert len(padding) == 4, 'padding size expected to be 4'
```
qedawkins, line 1864: Is this the right place for this assert? This seems closer to an op verification or a check during the lowering; the dtype should be independent of this.

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1286,6 +1286,26 @@ def aten〇reflection_pad1d〡shape(self: List[int], padding: List[int]) -> List
...
+def aten〇reflection_pad2d〡shape(self: List[int], padding: List[int]) -> List[int]:
+    assert len(self) >= 2
+    vdim = self[-2]
+    hdim = self[-1]
+    padding_left = padding[0]
+    padding_right = padding[1]
+    padding_top = padding[2]
+    padding_bottom = padding[3]
+    assert padding_left < hdim and padding_right < hdim
+    assert padding_top < vdim  and padding_bottom < vdim
```
qedawkins, line 1305: Looking at this shape function, this seems like the right place for the `assert(len(padding) == 4)` check.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -244,6 +244,283 @@ class ConvertAtenReflectionPad1dOp
...
+      Value tile = rewriter.create<tensor::ExtractSliceOp>(
+          loc, input, extractOffsets, extractShape, allOneStrides);
+
+      // Reverse the tile along the horizontal, vertical, or both
+      // dimensions
+      auto inputMap = AffineMap::getMultiDimIdentityMap(numDims, context);
+      if (shouldHReflect(horizontalPos))
+        inputMap =
+            reflectDim(inputMap, numDims, hDim, getHPadArgument(horizontalPos));
+      if (shouldVReflect(verticalPos))
```
qedawkins, line 490: nit: Since this spans 2 lines I would add braces around the body of the `if`.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -244,6 +244,283 @@ class ConvertAtenReflectionPad1dOp
...
+  LogicalResult
+  matchAndRewrite(AtenReflectionPad2dOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    SmallVector<int64_t> padInts;
+    if (!matchPattern(op.getPadding(), m_TorchListOfConstantInts(padInts)))
+      return rewriter.notifyMatchFailure(
+          op, "only support constant int pad ranges");
```
qedawkins, line 300: The shape function already checks for this, but it might also be worth importing these asserts here as well
```
    assert padding_left < hdim and padding_right < hdim
    assert padding_top < vdim  and padding_bottom < vdim
```
given that we know the padding values statically and it could help provide better error reporting than a crash later on in the shape function.

qedawkins (changes requested):
Looks mostly good to me, just have some nits and requests for error reporting.

lib/Conversion/TorchToLinalg/DataMovement.cpp
frederik-h: Ok, I'll put it a bit below where the necessary definitions are available.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -244,6 +244,283 @@ class ConvertAtenReflectionPad1dOp
...
+// center left [[2,1],[5,4],[8,7]]
+// center: copy of the input tensor
+// center right: [[2,1],[5,4],[8,7]]
+// bottom left: [[5,4],[2,1]]
+// center bottom: [[2,3,2]]
+// center right: [[2,1]]
+//
+// The lowering uses a tensor.extract_slice operation to create each tile,
+// a linalg.generic for the reflection, and a tensor.insert_slice to
+// insert the tile in the resulting tensor.
```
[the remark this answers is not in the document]
frederik-h: Ok, thanks for the pointer. I have noticed the operation but it did not seem fitting to me. Looking at it now, I think it could be possible to use it, but I am not really sure about the details. Perhaps this should be reconsidered in the future.

frederik-h:
> Looks mostly good to me, just have some nits and requests for error reporting.

Thanks for the review!

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -244,6 +244,294 @@ class ConvertAtenReflectionPad1dOp
...
+      tile = rewriter
+                 .create<linalg::GenericOp>(
+                     loc, llvm::cast<RankedTensorType>(tile.getType()), tile,
+                     tile, ArrayRef({inputMap, idMap}), iteratorTypes,
+                     [](OpBuilder &b, Location nestedLoc, ValueRange args) {
+                       b.create<linalg::YieldOp>(nestedLoc, args[0]);
+                     })
+                 .getResult(0);
+
+      // Insert the tile in the resultTensor
```
qedawkins, line 514: mini-nit: Add period to the end of sentence.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -244,6 +244,294 @@ class ConvertAtenReflectionPad1dOp
...
+
+      SmallVector<Value> extractOffsets(numDims, zero);
+      extractOffsets[hDim] = extractHOffset[horizontalPos];
+      extractOffsets[vDim] = extractVOffset[verticalPos];
+
+      Value tile = rewriter.create<tensor::ExtractSliceOp>(
+          loc, input, extractOffsets, extractShape, allOneStrides);
+
+      // Reverse the tile along the horizontal, vertical, or both
+      // dimensions
```
qedawkins, line 494: mini-nit: Add period to the end of sentence.

qedawkins (approved):
LGTM, thanks!

lib/Conversion/TorchToLinalg/DataMovement.cpp
frederik-h: Not sure if it is worth to re-run the CI for this though ;-). Anyway, I fixed it.

qedawkins: Oh right, my bad. I forgot Torch-MLIR CI takes forever. Thanks for fixing it anyway :D

frederik-h:
> I can merge for you if you need, unless you want to add a PR description first.

I just did it myself. It was the first thing that I did in the morning and I did not notice that squashing is not the default. Now I have produced a mess :-(. Sorry to everyone!  I think there is no reasonable way to fix this, i.e. reverting and adding one commit would not be cleaner than just keeping the commits.

stellaraccident:
No worries - it happens. It is off hours and the contribution patterns are such that I decided to just fix this. I squashed the head commits and force pushed. Seemed better than letting it linger and messing people up in the future.

stellaraccident:
I've gone ahead and disabled rebase merging. We can selectively enable if needed in the future.

frederik-h:
> No worries - it happens. It is off hours and the contribution patterns are such that I decided to just fix this. I squashed the head commits and force pushed. Seemed better than letting it linger and messing people up in the future.

Makes sense. Thanks so much for the help!

PR #2695, opened 2023-12-22
[MLIR][ONNX] Add OnnxToTorch support for Maxpool Op

AmosLewis:
https://github.com/nod-ai/SHARK-Turbine/issues/244

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -8,25 +8,12 @@
 //===----------------------------------------------------------------------===//
 
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
+#include "torch-mlir/Conversion/TorchOnnxToTorch/Utils.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 
 using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::onnx_c;
-
-static Value createConstantIntList(OpBinder binder,
-                                   ConversionPatternRewriter &rewriter,
-                                   SmallVector<int64_t> cstInput) {
-  SmallVector<Value> cstValue;
-  for (int64_t i : cstInput) {
-    cstValue.push_back(rewriter.create<Torch::ConstantIntOp>(
-        binder.getLoc(), rewriter.getI64IntegerAttr(i)));
-  }
-  return rewriter.create<Torch::PrimListConstructOp>(
-      binder.getLoc(),
-      Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
-      cstValue);
-}
 
 // Simple rewrites for the default domain.
 // See: https://onnx.ai/onnx/operators/
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -14,6 +14,20 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::onnx_c;
 
+static Value createConstantIntList(OpBinder binder,
+                                   ConversionPatternRewriter &rewriter,
+                                   SmallVector<int64_t> cstInput) {
+  SmallVector<Value> cstValue;
+  for (int64_t i : cstInput) {
+    cstValue.push_back(rewriter.create<Torch::ConstantIntOp>(
+        binder.getLoc(), rewriter.getI64IntegerAttr(i)));
+  }
+  return rewriter.create<Torch::PrimListConstructOp>(
+      binder.getLoc(),
+      Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+      cstValue);
+}
+
 // Simple rewrites for the default domain.
 // See: https://onnx.ai/onnx/operators/
 // For operators that are effectively version invariant, we register with
@@ -148,6 +162,86 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
 		    binder.op, resultType, lhs, rhs);
 		  return success();
 		});
+  patterns.onOp(
+      "MaxPool", 12, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        std::string autoPad;
+        if (binder.customOpNameStringAttr(autoPad, "auto_pad", "NOTSET"))
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "auto_pad bind failure");
+        if (autoPad != "NOTSET") {
+          // TODO: Add support for `auto_pad` != "NOTSET"
+          return rewriter.notifyMatchFailure(
+              binder.op, "unsupported conversion: auto_pad != NOTSET");
+        }
+
+        Torch::ValueTensorType resultType;
+        Value operand;
+        bool ceilMode;
+        int64_t storageOrder;
+        // TODO: Add support for indices output and storage_order
+        if (binder.tensorOperand(operand) ||
+            binder.s64BoolAttr(ceilMode, "ceil_mode", false) ||
+            binder.s64IntegerAttr(storageOrder, "storage_order", 0) ||
+            binder.tensorResultType(resultType))
+          return rewriter.notifyMatchFailure(
+              binder.op,
+              "operand/ceil_mode/storage_order/resultType bind failure");
+        if (storageOrder != 0)
+          return rewriter.notifyMatchFailure(
+              binder.op, "storage_order setting is not supported.");
+        // Determine the rank of input tensor.
+        std::optional<unsigned> maybeRank = Torch::getTensorRank(operand);
+        if (!maybeRank)
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "Unimplemented: unranked tensor");
+        unsigned rank = *maybeRank;
+
+        SmallVector<int64_t> kernel, padding, strides, dilations;
+        if (binder.s64IntegerArrayAttr(kernel, "kernel_shape", {}))
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "kernel_shape bind failure");
+        if (kernel.size() != rank - 2)
+          return rewriter.notifyMatchFailure(
+              binder.op, "kernel list size does not match the number of axes");
+        if (binder.s64IntegerArrayAttr(padding, "pads", {0}))
+          return rewriter.notifyMatchFailure(binder.op, "pads bind failure");
+        if (padding.size() != 1 && padding.size() != rank - 2)
+          return rewriter.notifyMatchFailure(
+              binder.op, "padding list size does not match the number of axes");
+        if (binder.s64IntegerArrayAttr(strides, "strides", {1}))
+          return rewriter.notifyMatchFailure(binder.op, "strides bind failure");
+        if (strides.size() != 1 && strides.size() != rank - 2)
+          return rewriter.notifyMatchFailure(
+              binder.op, "strides list size does not match the number of axes");
+        if (binder.s64IntegerArrayAttr(dilations, "dilations", {}))
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "dilations bind failure");
+
+        Value kernelSizeList = createConstantIntList(binder, rewriter, kernel);
+        Value paddingList = createConstantIntList(binder, rewriter, padding);
+        Value stridesList = createConstantIntList(binder, rewriter, strides);
+        Value dilationsList =
+            createConstantIntList(binder, rewriter, dilations);
+        Value cstCeilMode =
+            rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), ceilMode);
+
+        if (rank == 3)
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "Unimplemented: AtenMaxPool1dOp");
+        if (rank == 4) {
+          rewriter.replaceOpWithNewOp<Torch::AtenMaxPool2dOp>(
+              binder.op, resultType, operand, kernelSizeList, stridesList,
+              paddingList, dilationsList, cstCeilMode);
+          return success();
+        }
+        if (rank == 5) {
+          rewriter.replaceOpWithNewOp<Torch::AtenMaxPool3dOp>(
+              binder.op, resultType, operand, kernelSizeList, stridesList,
+              paddingList, dilationsList, cstCeilMode);
+          return success();
+        }
+        return failure();
+      });
   patterns.onOp("Greater", 16,
                 [](OpBinder binder, ConversionPatternRewriter &rewriter) {
                   Torch::ValueTensorType resultType;
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -155,11 +155,9 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         if (binder.customOpNameStringAttr(autoPad, "auto_pad", "NOTSET"))
           return rewriter.notifyMatchFailure(binder.op,
                                              "auto_pad bind failure");
-        if (autoPad != "NOTSET") {
-          // TODO: Add support for `auto_pad` != "NOTSET"
+        if (autoPad != "NOTSET")
           return rewriter.notifyMatchFailure(
               binder.op, "unsupported conversion: auto_pad != NOTSET");
-        }
 
         Torch::ValueTensorType resultType;
         Value operand;
@@ -227,7 +225,7 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
               paddingList, dilationsList, cstCeilMode);
           return success();
         }
-        return failure();
+        return rewriter.notifyMatchFailure(binder.op, "No rank is matched.");
       });
   patterns.onOp("Greater", 16,
                 [](OpBinder binder, ConversionPatternRewriter &rewriter) {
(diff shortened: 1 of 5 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -99,6 +99,113 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        }
+        if (binder.s64IntegerArrayAttr(dilations, "dilations", {})) {
+          return failure();
+        }
+
+        Torch::ValueTensorType resultType;
+        Value operand;
+        bool ceilMode;
+        int64_t storageOrder;
+        // TODO: Add support for indices output and storage_order
```
stellaraccident, line 121: I would have expected that we would have failed the match on an unsupported storage order and/or indices output as the op is most likely invalid. Is this not so?

stellaraccident (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -99,6 +99,113 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        SmallVector<int64_t> dilations;
+        if (binder.customOpNameStringAttr(autoPad, "auto_pad", "NOTSET"))
+          return failure();
+        if (autoPad != "NOTSET") {
+          // TODO: Add support for `auto_pad` != "NOTSET"
+          return rewriter.notifyMatchFailure(
+              binder.op, "unsupported conversion: auto_pad != NOTSET");
+        }
+        if (binder.s64IntegerArrayAttr(dilations, "dilations", {})) {
+          return failure();
```
rsuderman, line 114: Try to return a `notifyMatchFailure`. Makes future users have an easier time tracking down errors.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -99,6 +99,113 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        }
+        if (strides.size() != 1 && strides.size() != rank - 2) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "strides list size does not match the number of axes");
+        }
+
+        SmallVector<Value> cstKernel, cstPadding, cstStrides, cstDilations;
+        for (int64_t i : kernel) {
+          cstKernel.push_back(rewriter.create<Torch::ConstantIntOp>(
```
rsuderman, line 159: This loop is replicated a bunch. Could be worth making a static helper, especially considering we turn each of these values into a list later anyway.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -99,6 +99,113 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+            cstDilations);
+        Value cstCeilMode =
+            rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), ceilMode);
+
+        if (rank == 3) {
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "Unimplemented: AtenMaxPool1dOp");
+        } else if (rank == 4) {
+          rewriter.replaceOpWithNewOp<Torch::AtenMaxPool2dOp>(
```
rsuderman, line 197: No point to make these `else if` statements as each returns.

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
AmosLewis: I didn't add the support for this too because from the test files generated, there is no test for this 2. And for resnet support for now, it probably not necessary. And I haven't figured out a way to setup a storageorder in torch_mlir tensor.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
AmosLewis: why? there are differences between 1d,2d,3d maxpool, there are not the same maxpool op

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
rsuderman: I think you need to still need to report a match failure if the value for `storage_order` is not the expected value. I think the check below validates that we get the value but does not alter behaviour depending on its value.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -148,6 +162,90 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        if (binder.s64IntegerArrayAttr(padding, "pads", {0})) {
+          return rewriter.notifyMatchFailure(binder.op, "pads bind failure");
+        }
+        if (padding.size() != 1 && padding.size() != rank - 2) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "padding list size does not match the number of axes");
+        }
+        if (binder.s64IntegerArrayAttr(strides, "strides", {1})) {
+          return rewriter.notifyMatchFailure(binder.op, "strides bind failure");
+          ;
```
rsuderman, line 214: Extra newline.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -148,6 +162,90 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+              "operand/ceil_mode/storage_order/resultType bind failure");
+        // Determine the rank of input tensor.
+        std::optional<unsigned> maybeRank = Torch::getTensorRank(operand);
+        if (!maybeRank)
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "Unimplemented: unranked tensor");
+        unsigned rank = *maybeRank;
+
+        SmallVector<int64_t> kernel, padding, strides, dilations;
+        if (binder.s64IntegerArrayAttr(kernel, "kernel_shape", {})) {
```
rsuderman, line 197: For single line if statements we should try to minimize lines: e.g.

```
if (<condition>)
```

This was we avoid all the extra brackets.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
rsuderman: Yes this is the case, its more about reability. See the following case:

```
if (<condition1>) {
  do stuff.
  return
} else if (<condition2>) {
  do stuff
  return;
}
```

is the same as 

```
if (<condition1>) {
  do stuff.
  return
} 

if (<condition2>) {
  do stuff
  return;
}
```

It seems silly but it tends to make code more readable for a control flow standpoint.

rsuderman: Alternatively you case do a switch / case statement

```
switch(rank) {
  case 3:
    <make 2d pool>
    return success();
  case 4:
    <make 3d pool>
    return success();
  case 5:
    <make 3d pool>
    return success();
  default:
    <print errror>
    return failure();
}
```

test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
```diff
@@ -160,6 +162,71 @@ func.func @test_matmul_4d(%arg0: !torch.vtensor<[1,2,3,4],f32>, %arg1: !torch.vt
 
 // -----
 
+// CHECK-LABEL:   func.func @test_maxpool_2d_default(
+// CHECK-SAME:                                       %[[VAL_0:.*]]: !torch.vtensor<[1,3,32,32],f32>) -> !torch.vtensor<[1,3,31,31],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 12 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
```
rsuderman, line 166: For `lit` tests we want to remove extraneous details wherever possible, ie. this check should be relatively minimal:
```
// CHECK-SAME: %[[VAL_0:.*]]: !torch.vtensor<[1,3,32,32],f32>
```

As the rest remains unchanged.

test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
```diff
@@ -160,6 +162,71 @@ func.func @test_matmul_4d(%arg0: !torch.vtensor<[1,2,3,4],f32>, %arg1: !torch.vt
...
+// CHECK:           %[[VAL_4:.*]] = torch.constant.int 0
+// CHECK:           %[[VAL_5:.*]] = torch.prim.ListConstruct %[[VAL_4]] : (!torch.int) -> !torch.list<int>
+// CHECK:           %[[VAL_6:.*]] = torch.constant.int 1
+// CHECK:           %[[VAL_7:.*]] = torch.prim.ListConstruct %[[VAL_6]] : (!torch.int) -> !torch.list<int>
+// CHECK:           %[[VAL_8:.*]] = torch.prim.ListConstruct  : () -> !torch.list<int>
+// CHECK:           %[[VAL_9:.*]] = torch.constant.bool false
+// CHECK:           %[[VAL_10:.*]] = torch.aten.max_pool2d %[[VAL_0]], %[[VAL_3]], %[[VAL_7]], %[[VAL_5]], %[[VAL_8]], %[[VAL_9]] : !torch.vtensor<[1,3,32,32],f32>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool -> !torch.vtensor<[1,3,31,31],f32>
+// CHECK:           return %[[VAL_10]] : !torch.vtensor<[1,3,31,31],f32>
+// CHECK:         }
+func.func @test_maxpool_2d_default(%arg0: !torch.vtensor<[1,3,32,32],f32>) -> !torch.vtensor<[1,3,31,31],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 12 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
```
rsuderman, line 179: Not all of the attributes are required for the tests. You should be able to strip down to the following:

```
func.func @test_maxpool_2d_default(%arg0: !torch.vtensor<[1,3,32,32],f32>) -> !torch.vtensor<[1,3,31,31],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 12 : si64} {
```

test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
```diff
@@ -160,6 +162,71 @@ func.func @test_matmul_4d(%arg0: !torch.vtensor<[1,2,3,4],f32>, %arg1: !torch.vt
 
 // -----
 
+// CHECK-LABEL:   func.func @test_maxpool_2d_default(
+// CHECK-SAME:                                       %[[VAL_0:.*]]: !torch.vtensor<[1,3,32,32],f32>) -> !torch.vtensor<[1,3,31,31],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 12 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
+// CHECK:           %[[VAL_1:.*]] = torch.constant.int 2
```
rsuderman, line 167: Try to replace the variable names with relavant names. Leaving as `VAL_N` makes it difficult for people to review the test if it fails as what binds where can be confusing.

test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
```diff
@@ -160,6 +162,71 @@ func.func @test_matmul_4d(%arg0: !torch.vtensor<[1,2,3,4],f32>, %arg1: !torch.vt
...
+// CHECK:           %[[VAL_2:.*]] = torch.constant.int 2
+// CHECK:           %[[VAL_3:.*]] = torch.prim.ListConstruct %[[VAL_1]], %[[VAL_2]] : (!torch.int, !torch.int) -> !torch.list<int>
+// CHECK:           %[[VAL_4:.*]] = torch.constant.int 0
+// CHECK:           %[[VAL_5:.*]] = torch.prim.ListConstruct %[[VAL_4]] : (!torch.int) -> !torch.list<int>
+// CHECK:           %[[VAL_6:.*]] = torch.constant.int 1
+// CHECK:           %[[VAL_7:.*]] = torch.prim.ListConstruct %[[VAL_6]] : (!torch.int) -> !torch.list<int>
+// CHECK:           %[[VAL_8:.*]] = torch.prim.ListConstruct  : () -> !torch.list<int>
+// CHECK:           %[[VAL_9:.*]] = torch.constant.bool false
+// CHECK:           %[[VAL_10:.*]] = torch.aten.max_pool2d %[[VAL_0]], %[[VAL_3]], %[[VAL_7]], %[[VAL_5]], %[[VAL_8]], %[[VAL_9]] : !torch.vtensor<[1,3,32,32],f32>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool -> !torch.vtensor<[1,3,31,31],f32>
```
rsuderman, line 176: Strip operation call details when not import. I.e. this line should be simplified to:
```
// CHECK:           %[[VAL_10:.*]] = torch.aten.max_pool2d %[[VAL_0]], %[[VAL_3]], %[[VAL_7]], %[[VAL_5]], %[[VAL_8]], %[[VAL_9]]
```

test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
```diff
@@ -160,6 +162,71 @@ func.func @test_matmul_4d(%arg0: !torch.vtensor<[1,2,3,4],f32>, %arg1: !torch.vt
 
 // -----
```
rsuderman, line 164: Overall try to make the style of the lit tests match the previous tests above. We try to make these easy to read as failures become easier to understand.

test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
```diff
@@ -160,6 +162,71 @@ func.func @test_matmul_4d(%arg0: !torch.vtensor<[1,2,3,4],f32>, %arg1: !torch.vt
...
+// CHECK:           %[[VAL_10:.*]] = torch.aten.max_pool2d %[[VAL_0]], %[[VAL_3]], %[[VAL_7]], %[[VAL_5]], %[[VAL_8]], %[[VAL_9]] : !torch.vtensor<[1,3,32,32],f32>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool -> !torch.vtensor<[1,3,31,31],f32>
+// CHECK:           return %[[VAL_10]] : !torch.vtensor<[1,3,31,31],f32>
+// CHECK:         }
+func.func @test_maxpool_2d_default(%arg0: !torch.vtensor<[1,3,32,32],f32>) -> !torch.vtensor<[1,3,31,31],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 12 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
+  %0 = torch.operator "onnx.MaxPool"(%arg0) {torch.onnx.kernel_shape = [2 : si64, 2 : si64]} : (!torch.vtensor<[1,3,32,32],f32>) -> !torch.vtensor<[1,3,31,31],f32>
+  return %0 : !torch.vtensor<[1,3,31,31],f32>
+}
+
+// -----
+
```
rsuderman, line 185: Repeat the comments above for the other tests below

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
AmosLewis: I have created a static helper. But it will used across files. So I think it worth creating a Utils.h/cpp instead a static function for single file

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -148,6 +149,86 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+              paddingList, dilationsList, cstCeilMode);
+          return success();
+        }
+        if (rank == 5) {
+          rewriter.replaceOpWithNewOp<Torch::AtenMaxPool3dOp>(
+              binder.op, resultType, operand, kernelSizeList, stridesList,
+              paddingList, dilationsList, cstCeilMode);
+          return success();
+        }
+        return failure();
```
rsuderman, line 230: Include a `rewriter.notifyMatchFailure` call that no rank is matched.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -148,6 +149,86 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
 		  return success();
 		});
+  patterns.onOp(
+      "MaxPool", 12, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        std::string autoPad;
+        if (binder.customOpNameStringAttr(autoPad, "auto_pad", "NOTSET"))
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "auto_pad bind failure");
+        if (autoPad != "NOTSET") {
+          // TODO: Add support for `auto_pad` != "NOTSET"
```
rsuderman, line 159: Comment is unnecessary as match failure message is descriptive enough.

rsuderman (approved):

rsuderman:
LGTM with some small nits

PR #2696, opened 2023-12-23
[MLIR][ONNX] Add OnnxToTorch support for Slice Op

wu-s-john:


wu-s-john:
Implement https://github.com/nod-ai/SHARK-Turbine/issues/259

@stellaraccident @vivekkhandelwal1 @rsuderman please review

wu-s-john:
I don't know why there are testing errors detected tbh.

I looked at failures more carefully and it seems that the `test_transpose_all_permutations_4` produced the wrong code that the test was expecting. However, it actually generated the right outputs. Not sure what's going on. Possibly, a data race is occurring?

stellaraccident:
This looks in reasonable shape sans test failures and conflicts. Assigning to either Rob or Vivek to triage / get it landed. (a lot of people are on holiday -- apologies for delays)

rsuderman:
I'll pull this down and take a look.

rsuderman:
@wu-s-john Managed to fix through the errors in the follow branch:

https://github.com/rsuderman/torch-mlir/tree/slice

Looks like you were throwing an `i64` as the element type for a torch tensor when they enforce signedness. It also appears that `torch-mlir` still uses `-1` for its dynamic shape signifier which triggered the other assertion failure.

As a useful trick, LLVM includes assertions that catch these errors. So `cmake -DLLVM_ENABLE_ASSERTIONS=ON .` often helps track down these failures.

I'll add some extra changes that I would recommend on top of my diff.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -870,6 +870,156 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        if (!(endsTy.getRank() == 1 && startsTy.getRank() == 1 &&
+              startsTy.getDimSize(0) == endsTy.getDimSize(0))) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected the rank of starts and ends tensors to be 1 "
+                         "and their dimensions to match");
+        }
+
+        // Infer Axes
+        int64_t numAxes;
+        if (binder.getNumOperands() >= 4) {
```
rsuderman, line 914: Rather than an if-else, set the value to the default then only change if the operand exists.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -870,6 +870,156 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          if (operandTy.getRank() != endsTy.getDimSize(0))
+            return rewriter.notifyMatchFailure(
+                binder.op,
+                "When axes argument are not applied, expected the tensor "
+                "operand rank to match the sizes of starts and ends");
+          numAxes = operandTy.getRank();
+        }
+
+        // Infer Steps
+        if (binder.getNumOperands() >= 5) {
```
rsuderman, line 936: Same thing here

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -870,6 +870,156 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+              startsTy.getDimSize(0) == endsTy.getDimSize(0))) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected the rank of starts and ends tensors to be 1 "
+                         "and their dimensions to match");
+        }
+
+        // Infer Axes
+        int64_t numAxes;
+        if (binder.getNumOperands() >= 4) {
+          if (binder.tensorOperandAtIndex(axes, 3)) {
```
rsuderman, line 915: Bind / get the axes outside of this if and in a check if there are 3 or more operands. It makes it clear that we always get this for different number of operands.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -870,6 +870,156 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          auto stepsTy =
+              stepsTorchTy.toBuiltinTensor().dyn_cast<RankedTensorType>();
+
+          if (!(stepsTy && stepsTy.getDimSize(0) == endsTy.getDimSize(0)))
+            return rewriter.notifyMatchFailure(
+                binder.op, "When steps is provided, steps size should match "
+                           "the size of starts and ends");
+        }
+
+        auto resultTy =
```
rsuderman, line 950: Move this to the top with the rest of the type validation.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -870,6 +870,156 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            resultTorchType.toBuiltinTensor().dyn_cast<RankedTensorType>();
+        if (!resultTy)
+          return failure();
+
+        Location loc = binder.getLoc();
```
rsuderman, line 955: Move this to the top with the operand declarations.

wu-s-john:
> Looks like you were throwing an i64 as the element type for a torch tensor when they enforce signedness

Ooo thanks for catching this. It seemed like other places in the code were using i64 and it seemed natural to use this. Thanks for the tip for using `cmake -DLLVM_ENABLE_ASSERTIONS=ON`. I will make sure to not have these errors next time.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -870,6 +870,156 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          Value item = rewriter.create<Torch::AtenItemOp>(
+              loc, rewriter.getType<Torch::IntType>(), sel);
+          return item;
+        };
+
+        Value defaultStepSize;
+        if (!steps) {
+          defaultStepSize = rewriter.create<Torch::ConstantIntOp>(
+              loc, rewriter.getType<Torch::IntType>(),
+              rewriter.getIntegerAttr(rewriter.getIntegerType(64), 1));
```
rsuderman, line 976: Change to 
```
rewriter.getIntegerAttr(rewriter.getIntegerType(64), /*isSigned=*/true));
```

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -870,6 +870,156 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        auto startsTorchTy = starts.getType().cast<Torch::ValueTensorType>();
+        auto startsTy =
+            startsTorchTy.toBuiltinTensor().dyn_cast<RankedTensorType>();
+
+        auto endsTorchTy = ends.getType().cast<Torch::ValueTensorType>();
+        auto endsTy =
+            endsTorchTy.toBuiltinTensor().dyn_cast<RankedTensorType>();
+
+        if (!operandTy)
+          return failure();
```
rsuderman, line 903: You should include an error message

rsuderman (changes requested):

wu-s-john:
@rsuderman 

Made corrections to the PR. Addressed most of your comments. Overall, your concern was to simplify the logic for nonbinded arguments and supply them with default ones early one. Implemented these and the code looks a lot cleaner now! I think you would like it!

PR #2697, opened 2023-12-26
Implement lowering of torch.aten.logit

ikalinic:
Closes nod-ai/SHARK-Turbine#290


stellaraccident (approved):
Thanks for the contribution. If the CI passes, lgtm and we can land.

ikalinic:
@stellaraccident thank you for the review! It seems that my e2e test has an issue, I'm passing a Tensor instead of a float to the value of eps. While developing, I wasn't able to run all e2e test with the appropreate python script, it kept getting stuck. Is there a way to run a single e2e test?

stellaraccident:
> @stellaraccident thank you for the review! It seems that my e2e test has an issue, I'm passing a Tensor instead of a float for the value of eps. While developing, I wasn't able to run all e2e test with `e2e_test.sh`, it kept getting stuck. Is there a way to run a single e2e test?
> 
> As for the test, do I need to specify `[0]` as the size for eps arg annotation? Also, I'm not sure how to get a float and not a tensor with tu.rand

There are a lot of people out on vacation right now, so we're having a bit of trouble keeping up with reviews and tech support. Let me check in tomorrow and see if we have someone online who can help.

ikalinic:
Hi @stellaraccident, happy New Year! Is there any update on the help with this PR?

Groverkss:
> Hi @stellaraccident, happy New Year! Is there any update on the help with this PR?

Hi! Stella is on vacation, but I think the folks who work on torch-mlir are back from vacation. @vivekkhandelwal1 Could you help with this or assign this to someone to help out?

vivekkhandelwal1:
Hi @ikalinic, you need to add `ElementwiseLogitModule_basic ` to https://github.com/llvm/torch-mlir/blob/main/projects/pt1/e2e_testing/xfail_sets.py#L1366

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -7,6 +7,9 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/IR/BuiltinAttributeInterfaces.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/Support/LogicalResult.h"
```
vivekkhandelwal1, line 12: Please remove these includes, they are not required.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -25,6 +28,7 @@
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/ADT/APSInt.h"
+#include "llvm/Support/raw_ostream.h"
```
vivekkhandelwal1, line 31: Please remove this include, they are not required.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1971,6 +1974,81 @@ class ConvertTensorStaticInfoCastOp
...
+                    Value truncEps =
+                        b.create<arith::TruncFOp>(loc, inputElementType, eps);
+                    Value oneMinusEps =
+                        b.create<arith::SubFOp>(loc, oneValue, truncEps);
+
+                    Value min =
+                        b.create<arith::MinimumFOp>(loc, input, oneMinusEps);
+                    Value clampedInput =
+                        b.create<arith::MaximumFOp>(loc, min, truncEps);
+
```
vivekkhandelwal1, line 2034: Are you sure that these values will always be Floating-point values? Otherwise ops like `...FOp` won't work.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1971,6 +1974,81 @@ class ConvertTensorStaticInfoCastOp
...
+                    zI = clampedInput;
+                  }
+
+                  Value probability =
+                      b.create<arith::SubFOp>(loc, oneValue, zI);
+                  Value odds = b.create<arith::DivFOp>(loc, zI, probability);
+                  Value result = b.create<math::LogOp>(loc, odds);
```
vivekkhandelwal1, line 2041: Same as above

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -1480,6 +1480,29 @@ def ElementwiseLog1pModule_basic(module, tu: TestUtils):
...
+        ([], torch.float32, True),
+    ])
+    def forward(self, a, b):
+        return torch.logit(a, eps=b)
+
+
+@register_test_case(module_factory=lambda: ElementwiseLogitModule())
+def ElementwiseLogitModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4), tu.rand(low=1e-6, high=1e-3).item())
+
```
vivekkhandelwal1, line 1501: If integer inputs are allowed then add a test with `integer dtype`.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
ikalinic: The logit (https://pytorch.org/docs/stable/special.html#torch.special.logit) function is defined as:
```math
\begin{align}
    y_i &= \ln({z_i \over 1 - z_i}) \\
    z_i &= \begin{cases}
        x_i & \text{if eps is None} \\
        eps & if x_i < eps \\
        x_i & if eps \leq x_i \leq 1 - eps \\
        1 - eps & if x_i > 1 - eps
    \end{cases}
\end{align}
```
So this means that the input should be a floating point value in the range of `[0, 1)` or clamped to `[eps, 1-eps]` if `eps` is defined.

vivekkhandelwal1: I tried the PyTorch op with integer inputs, and it didn't throw a runtime error. You should add a check at the top restricting the inputs to be of Floating-point type.

ikalinic: Got it, I thought the type checks were implicit, thanks!

Also, why must the `ElementwiseLogitModule_basic` be put in xfail_sets?

vivekkhandelwal1: There are several configs for which the e2e tests are run. This test is failing for the LTC backend and for that reason you need to add it to the `LTC_XFAIL_SET`.

projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -532,6 +532,7 @@
     "ElementwiseLtDiffWidthScalarModule_basic",
     "ElementwiseLtFloatScalarModule_basic",
     "ElementwiseLtIntScalarModule_basic",
+    "ElementwiseLogitModule_basic",
```
ikalinic, line 535: I've added `ElementwiseLogitModule_basic` to the wrong set, sorry about that

vivekkhandelwal1 (approved):
LGTM

PR #2698, opened 2023-12-26
[MLIR][ONNX] Add OnnxToTorch support for Reshape Op

saienduri:
This commit adds the OnnxToTorch support for Reshape op.


stellaraccident (approved):
Thank you!

Issue #2689, opened 2023-12-21
Support lowering `fb-opt` from `onnx` to `torch` dialect


rsuderman:
The following operations fail to lower with examples. These operations are supposed to be finished and will need to be correct.

- `onnx.Cast`
```
%208 = torch.operator "onnx.Cast"(%207) {torch.onnx.to = 1 : si64} : (!torch.vtensor<[?,?,?,?],si64>) -> !torch.vtensor<[?,?,?,?],f32>
```
- `onnx.Div`
```
    %227 = torch.operator "onnx.Div"(%222, %226) : (!torch.vtensor<[1,64,768],f32>, !torch.vtensor<[1,64,1],f32>) -> !torch.vtensor<[1,64,768],f32>
```
- `onnx.Equal`
```
%205 = torch.operator "onnx.Equal"(%204, %175) : (!torch.vtensor<[4],si64>, !torch.vtensor<[4],si64>) -> !torch.vtensor<[4],i1>
```
- `onnx.reduceMean`
```
%221 = torch.operator "onnx.ReduceMean"(%220) {torch.onnx.axes = [-1 : si64]} : (!torch.vtensor<[1,64,768],f32>) -> !torch.vtensor<[1,64,1],f32>
```

The following operations have no implementations:

- onnx.Concat
- onnx.CumSum
- onnx.Expand
- onnx.Gather
- onnx.Max
- onnx.Mul
- onnx.Pow
- onnx.Reshape
- onnx.Slice
- onnx.Split

renxida:
Made a [gist](https://gist.github.com/renxida/8a9f23c20c473a23ca36712e777652b7)with some test cases for the ones with examples. Looks like Div works now. I'll be tackling onnx.Equal.

renxida:
Equal from [this commit](https://github.com/llvm/torch-mlir/commit/0b4422a2532a1ed673de33c17c86f038e9354969) has an unecessarily high opset version (19) when i'm pretty sure we can support all the way down to 1

stellaraccident:
I noticed a couple of others like that too. Would be good to do a scrub: https://github.com/llvm/torch-mlir/issues/2700

PR #2701, opened 2023-12-27
Fix since-opset too high

### LINKED ISSUE #2700: ONNX Gemm and LeakyRelu have wrong from version
Both of these can be `1`, not the latest version. Figuring out the version to set the pattern to requires some judgment. In this case, I consulted these docs:

* https://onnx.ai/onnx/operators/text_diff_Gemm_1_13.html : Shows that there is nothing that changed in the op that the pattern can't handle as-is.
* https://onnx.ai/onnx/operators/text_diff_LeakyRelu_1_16.html : Just some commenting and an added dtype

(found when running some stuff in the test suite)

renxida:
Addresses two of the ops from https://github.com/llvm/torch-mlir/issues/2689

@vivekkhandelwal1 let me know if the since values for Cast and Equal are set to 19 for compatibility reasons and if I should write separate lowerings for the older versions - I merely reduced the since-version to 1 in your code because they looked backward compatible to me.

Fixes #2700

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -377,7 +377,7 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
         return success();
       });
   patterns.onOp(
-      "Cast", 19, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+      "Cast", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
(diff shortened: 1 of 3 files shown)


renxida:
Stella found some more in #2700 

Both of these can be 1, not the latest

> https://onnx.ai/onnx/operators/text_diff_Gemm_1_13.html : Shows that there is nothing that changed in the op that the pattern can't handle as-is.
> https://onnx.ai/onnx/operators/text_diff_LeakyRelu_1_16.html : Just some commenting and an added dtype

stellaraccident (approved):

PR #2704, opened 2023-12-29
Layer_Normalization

afalkenberg1:
Checking in 3 files for Layer_Normalization
There are limits to the implementation still:
onnx.Layer_Normalization produces 3 results while the 2nd and the 3rd is optional.
For this initial build we are ignoring those results since they are not supported as part of the torch layer_norm. Also added a function to check for the multiple result types in case we would use them later.


afalkenberg1:
unfortunately this does not merge correctly

afalkenberg1:
Identified the problem so I am going to close this one for now.

PR #2710, opened 2023-12-29
OnnxToTorch support for onnx.InstanceNormalization op

aldesilv:
https://github.com/nod-ai/SHARK-Turbine/issues/327


vivekkhandelwal1:
Hi @aldesilv, I did take a look at your code and I think there's an issue with the dtype check for the instance norm op. The PyTorch doesn't support the device cpu/meta (both of which I tried) for the tensors to this op. I think for now you can just add a comment saying the device is not supported hence unable to check the dtype function and remove the check.

@ramiro050, would you like to add something to this?

vivekkhandelwal1 (changes requested):
Hi @aldesilv, can you please add some e2e tests for quantized::instance_norm op so that your TorchToLinalg lowering can be verified?

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/torch_ods_gen.py
```diff
@@ -424,6 +424,12 @@ def emit_with_mutating_variants(key, **kwargs):
     emit(
         "aten::batch_norm : (Tensor, Tensor?, Tensor?, Tensor?, Tensor?, bool, float, float, bool) -> (Tensor)"
     )
+    emit(
+        "aten::instance_norm : (Tensor, Tensor?, Tensor?, Tensor?, Tensor?, bool, float, float, bool) -> (Tensor)"
```
ramiro050, line 428: I don't see any patterns for this op. Do we need it in this PR?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -251,6 +251,39 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+		   name.append("epsilon");
+
+		   auto attr = binder.op->getAttr(name);
+		   float eps;
+		   if (attr) {
+		     auto epsAttr = dyn_cast<FloatAttr>(attr);
+                     eps = epsAttr.getValue().convertToFloat();
+		   } else {
+                     eps = 1e-05f;
+		   }
```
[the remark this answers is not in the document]
ramiro050: +1 to adding e2e test

aldesilv:
> Hi @aldesilv, can you please add some e2e tests for quantized::instance_norm op so that your TorchToLinalg lowering can be verified?

when using quantized instance norm (torch.ao.nn.quantized.InstanceNorm2d) in an e2e test, 
NotImplementedError: Could not run 
'quantized::instance_norm' with arguments from the 'CPU' backend. This could be because the operator doesn't exist for this backend, or was omitted during the selective/custom build process (if using custom build).

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/torch_ods_gen.py
aldesilv: no

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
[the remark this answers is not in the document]
aldesilv: will add a test in torch-mlir/test/Conversion/TorchToLinalg

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -335,38 +335,66 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+                   }
+
+		   auto epsValue = rewriter.create<Torch::ConstantFloatOp>(binder.getLoc(),
+                                     rewriter.getF64FloatAttr(eps));
+                   
+                   auto outputScale = rewriter.create<Torch::ConstantFloatOp>(binder.getLoc(),
+                                        rewriter.getF64FloatAttr(1.0f));
+                   auto outputZeroPoint = rewriter.create<Torch::ConstantIntOp>(
+                                            binder.getLoc(), rewriter.getI64IntegerAttr(0));
+	           rewriter.replaceOpWithNewOp<Torch::QuantizedInstanceNormOp>(
```
ramiro050, line 358: Is this the right op to decompose into? From looking at the [ONNX docs for `InstanceNormalization`](https://onnx.ai/onnx/operators/onnx__InstanceNormalization.html), it seems that it performs the same computation as the `torch.nn.functional.batch_norm` op, and we already have support for `batch_norm` in torch-mlir

ramiro050 (changes requested):
Appologies for the late review. I have one question

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
aldesilv: [the paper ](https://arxiv.org/pdf/1701.02096.pdf) says that batch normalization "applies the normalization to a whole batch of images instead of single ones."

ramiro050: I'm not familiar with the paper. From the ONNX docs, `InstanceNormalization` takes inputs of shape `(NxCx...)`, where `N` is the batch dimension, so `torch.nn.functional.batch_norm` should be a good match.

aldesilv: from what I understand, instance norm calculates the mean and variance per image, and then normalizes with that instead of where batch normalization does by batches.  The onnx op carries instance normalization based on that paper, but they are different ops.

ramiro050: Ah, I see what you mean. You are right, they do different things. PyTorch does have an instance version as well: `'aten::instance_norm`. We should be using this one rather than the quantized one. This op should be testable e2e.

aldesilv: aten::instance_norm op uses running mean, running var, and momentum unlike the onnx.InstanceNormalization. quantized::instance_norm has zero point and output_scale, but the rest of the args are the same as the onnx instance normalization.

ramiro050: > aten::instance_norm op uses running mean, running var, and momentum

That's fine, the running statistics are optional. Your lowering can just match when they are `None`. The `momentum` can be ignored by being set to 0, I believe.

My opposition to using the quantized one is that `onnx.InstanceNormalization` is not a quantized operation. From the docs, the ONNX op takes regular float tensors and produces regular float tensors. On the other hand, the PyTorch version (basing from [the quantized::batch_norm docs](https://pytorch.org/docs/stable/generated/torch.quantized_batch_norm.html), since there are no docs for quantized::instance_norm) takes quantized tensors as inputs (tensors with dtype `torch.q...`). So, it is not the right operation to lower `onnx.InstanceNormalization` to.

aldesilv: Ok. the e2e test fails because it says mulf expects a floating point like operand but gets a torch.none, but why is the instance norm e2e test failing?

ramiro050: The e2e instance norm test is failing because the conversion of `instance_norm` to linalg is generating invalid IR. In particular, it is generating a `mulf` op with invalid inputs.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1826,6 +1826,199 @@ class ConvertAtenBatchNormOp : public OpConversionPattern<AtenBatchNormOp> {
 };
 } // namespace
 
+namespace {
+class ConvertAtenInstanceNormOp
```
ramiro050, line 1830: nit: I think this pattern can be written as a decomposition (the linalg.generics seem to recreate the behavior of torch ops we already support). Using a decomposition would simplify the pattern and have the added benefit that the other 2 backends in torch-mlir get the lowering for free.

projects/pt1/python/torch_mlir_e2e_test/test_suite/norm_like.py
```diff
@@ -489,3 +489,19 @@ def forward(self, x):
...
+        super().__init__()
+        self.inorm = torch.nn.InstanceNorm2d(100)
+
+    @export
+    @annotate_args([
+        None,
+        ([20, 100, 35, 45], torch.float32, True),
+    ])
+    def forward(self, x):
+        return self.inorm(x)
```
ramiro050, line 503: nit: use `torch.ops.aten.instance_norm`. See: https://github.com/llvm/torch-mlir/blob/main/docs/adding_an_e2e_test.md#forward-method

test/Conversion/TorchToLinalg/instancenorm.mlir
ramiro050, line 1: This test is not needed. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

ramiro050 (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
aldesilv: how do I write a decomposition?

ramiro050: Here is an example that does a norm operation: https://github.com/llvm/torch-mlir/pull/1332

projects/pt1/python/torch_mlir_e2e_test/test_suite/norm_like.py
```diff
@@ -489,3 +489,40 @@ def forward(self, x):
 def LayerNormNormalizeOverAllDimsModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(2, 2, 3))
 
+class InstanceNormModule(torch.nn.Module):
```
ramiro050, line 492: Is this testing a case not tested by the other test below? If not, then we should remove it

lib/Conversion/TorchToLinalg/Uncategorized.cpp
ramiro050, line 1: This should no longer be needed now that we have the decomposition.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -6656,6 +6801,7 @@ class DecomposeComplexOpsPass
         DecomposeAtenAddCLikeOp<AtenAddcmulOp, AtenMulTensorOp>>(patterns);
     addPatternIfTargetOpIsIllegal<
         DecomposeAtenAddCLikeOp<AtenAddcdivOp, AtenDivTensorOp>>(patterns);
+    addPatternIfTargetOpIsIllegal<DecomposeAtenInstanceNormOp>(patterns);
```
ramiro050, line 6804: Can you also add the op here: 

https://github.com/llvm/torch-mlir/blob/d6e1d836ca8d49da25ad5e2f10d6816bfbb6ba2f/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L367

ramiro050 (changes requested):
Looks good! Just a couple small changes

ramiro050 (approved):
Thanks!

vivekkhandelwal1 (approved):

PR #2712, opened 2023-12-30
[Stablehlo] add refbackend

qingyunqu:


stellaraccident (commented):
I'm fine with giving this a try, but the stablehlo ref backend significantly increases the surface area of the stablehlo dep, and keeping it up to date has already been a challenge.

I think this pushes us to a policy change: if when integrating a new version of llvm, if a corresponding stablehlo version doesn't work, we will land with stablehlo disabled and fix it in post commit. If this becomes a pattern, we can factor the CI a bit differently to test these other configurations in a non gating way.

I think that removing the requirement that every commit is available in lockstep will give us all what we want without needing to engage in integrate heroics.

qingyunqu:
> I'm fine with giving this a try, but the stablehlo ref backend significantly increases the surface area of the stablehlo dep, and keeping it up to date has already been a challenge.
> 
> I think this pushes us to a policy change: if when integrating a new version of llvm, if a corresponding stablehlo version doesn't work, we will land with stablehlo disabled and fix it in post commit. If this becomes a pattern, we can factor the CI a bit differently to test these other configurations in a non gating way.
> 
> I think that removing the requirement that every commit is available in lockstep will give us all what we want without needing to engage in integrate heroics.

I agree with you. I think it's ok to disable stablehlo temporarily.
BTW, this PR is ready. Could you please review it?

lib/CMakeLists.txt
```diff
@@ -29,6 +29,9 @@ set(LinkedLibs
   # Conversion passes.
   TorchMLIRConversionPasses
   TorchMLIRTorchOnnxToTorch
+
+  StablehloPasses
+  StablehloLinalgTransforms
```
stellaraccident, line 34: Can you move these to a `list(APPEND` following that is gated on `TORCH_MLIR_ENABLE_STABLEHLO`?

lib/InitAll.cpp
```diff
@@ -29,6 +29,9 @@
 #include "torch-mlir/Dialect/TorchConversion/Transforms/Passes.h"
 #include "torch-mlir/RefBackend/Passes.h"
 
+#include "stablehlo/conversions/linalg/transforms/Passes.h"
```
stellaraccident, line 32: Surround in `#ifdef TORCH_MLIR_ENABLE_STABLEHLO`

lib/InitAll.cpp
```diff
@@ -52,6 +55,9 @@ void mlir::torch::registerAllPasses() {
   mlir::torch::onnx_c::registerTorchOnnxToTorchPasses();
   mlir::torch::TMTensor::registerPasses();
 
+  mlir::stablehlo::registerChloLegalizeToStablehloPass();
```
stellaraccident, line 58: And here (Surround in `#ifdef TORCH_MLIR_ENABLE_STABLEHLO`)

stellaraccident (approved):
Thanks! Before landing, would you mind conditionalizing like I mention in the comments?

qingyunqu:
> Thanks! Before landing, would you mind conditionalizing like I mention in the comments?

Thanks, I forgot these config.

projects/pt1/python/torch_mlir_e2e_test/stablehlo_backends/linalg_on_tensors.py
```diff
@@ -0,0 +1,57 @@
...
+# Linalg-on-Tensors backend contract accepted by RefBackend.
+STABLEHLO_TO_LINALG_FUNC_PIPELINE = ",".join([
+    "func.func(chlo-legalize-to-stablehlo)",
+    "canonicalize",
+    "stablehlo-legalize-to-linalg"
+])
+
+
+class LinalgOnTensorsStablehloBackend(StablehloBackend):
+    """Main entry-point for the linalg-on-tensors based TOSA backend.
```
ScottTodd, line 28: Why is this file mentioning TOSA?

qingyunqu: It's a typo, I will fix it.

lib/CMakeLists.txt
```diff
@@ -31,6 +31,10 @@ set(LinkedLibs
   TorchMLIRTorchOnnxToTorch
 )
 
+if(TORCH_MLIR_ENABLE_STABLEHLO)
+list(APPEND LinkedLibs StablehloPasses StablehloLinalgTransforms)
+endif()
```
ScottTodd, line 36: I'm having trouble compiling some part of this after syncing past this commit (f3e8199). My build is fine before this commit (8e2e5ee) and after this commit with `-DTORCH_MLIR_ENABLE_STABLEHLO=OFF`.

Error logs: https://gist.github.com/ScottTodd/332e4e76b744d0bfdd75c68cc8191970#file-torch_mlir_build-txt-L1784
System: Windows 10, MSVC (Visual Studio Build Tools 2022 Preview - amd64)

Do StableHLO and torch-mlir have Windows CI builds? I don't see issues in IREE's downstream Windows CI or locally, so either that is behind this version or something else is different about how the projects are configured.

qingyunqu: It seems a different behavior between different cxx compiler. If you do not need stablehlo backend, I think you could set `TORCH_MLIR_ENABLE_STABLEHLO=OFF` when building.

qingyunqu: Torch-MLIR doesn't have Windows CI builds now. I don't have an idea to fix this compiling error.

qingyunqu: It seems the same error as https://github.com/openxla/stablehlo/issues/1772

PR #2715, opened 2024-01-02
Replication pad2d

lisaliu1:


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -137,17 +140,17 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
-		  Torch::ValueTensorType resultType;
+                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+                  Torch::ValueTensorType resultType;
                   Value lhs, rhs;
                   if (binder.tensorOperands(lhs, rhs) ||
                       binder.tensorResultType(resultType)) {
                     return failure();
                   }
                   rewriter.replaceOpWithNewOp<Torch::AtenMulTensorOp>(
-		    binder.op, resultType, lhs, rhs);
```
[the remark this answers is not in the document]
lisaliu1: thanks. The format changes are the results of clang-format, which is recommended in the torch-mlir build instructions. Next time, I'll just try clang-format on my own changes.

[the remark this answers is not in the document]
qedawkins: Yeah I've seen this a few times, but seems like something got messed up with torch-mlir's clang format somewhere along the way and we need a reset. I don't have a great idea of how to avoid this so I guess it's fine for now, but this really does mess up the blame log.

qedawkins (changes requested):
I'm noticing a lot of formatting changes here that make it harder to review like @wu-s-john noted, some of which are probably not formatting certain comments the way we want. Reading the code, this looks like it is the case, but are there any changes to `lib/Conversion/TorchToLinalg/DataMovement.cpp` or `lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp` that aren't NFC/formatting changes? If so, it seems like something might be wrong with torch-mlir's clang format and it would be good to solve in a separate issue/PR where we walk through all of the changes introduced by clang-format and make sure they are what we want. Can also preserve the blame log that way.

Some git-fu that might help (back up your branch before doing this so as to not lose work).
```
# Get the most recent commit that changed this file that isn't from this PR
git log -p lib/Conversion/TorchToLinalg/DataMovement.cpp

# (I think you want 6660a26594dc82cd3dd6fc33c9269ff09ecd263a, but the above will make it clear)
# Check out the commit for just the file you want to revert changes to
git checkout <commit hash> lib/Conversion/TorchToLinalg/DataMovement.cpp
# Commit the changes
git commit -m "revert data movement changes"
```
From https://stackoverflow.com/questions/2733873/how-can-i-revert-a-single-file-to-a-previous-version

Alternatively you could squash all of your commits down to one, revert it locally, and then add back the relevent parts one by one (`git add -p`).

projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -565,6 +565,105 @@ def __init__(self):
...
+        return torch.ops.aten.replication_pad2d(x, (1, 2, 0, 4))
+
+
+@register_test_case(module_factory=lambda: ReplicationPad2dModule_3())
+def ReplicationPad2dModule_top0(module, tu: TestUtils):
+    module.forward(tu.rand(1, 1, 3, 3, low=-1))
+
+# ==============================================================================
+
+class ReplicationPad2dModule_4(torch.nn.Module):
```
qedawkins, line 648: nit: I would give more descriptive names to the modules themselves as well (i.e., attach `bottom0` to the module name in addition to the test function. Makes it easier to see what the test is trying to do.

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1271,6 +1271,15 @@ def pad_shape_fn(input: List[int], pad: List[int]):
 def aten〇constant_pad_nd〡shape(self: List[int], pad: List[int], value: float = 0) -> List[int]:
     return pad_shape_fn(self, pad)
 
+def aten〇replication_pad2d〡shape(self: List[int], padding: List[int]) -> List[int]:
+    assert len(self) >= 2
+    return pad_shape_fn(self, padding)
+
+def aten〇replication_pad2d〡dtype(self_rank_dtype: Tuple[int, int], padding: List[int]) -> int:
+    self_rank, self_dtype = self_rank_dtype
+    assert len(padding) == 4, 'padding size expected to be 4'
```
qedawkins, line 1280: I would move this assert to the shape function, given that the dtype does not depend on the padding shape at all.

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
```diff
@@ -79,6 +81,227 @@ class ConvertAtenConstantPadNdOp
...
+          highPadding[2] = padInts[3];
+          vRightSlice = torch_to_linalg::getPaddedTensor(op, rewriter, vRightSlice, lowPadding, highPadding, bottomRightValue);
+        }
+        for (auto i=0; i<padInts[1]; ++i) {
+          tensorsRight.push_back(vRightSlice);
+        }
+        Value rightPadTile = rewriter.create<tensor::ConcatOp>(loc, 3, tensorsRight);
+        tensorsRes.push_back(rightPadTile);
+      }     
+      Value resTensor = rewriter.create<tensor::ConcatOp>(loc, 3, tensorsRes);
```
qedawkins, line 297: I left a comment on the PR for this commit, but unfortunately rebase/merge mix ups lost the comment: https://github.com/llvm/torch-mlir/commit/0860c41ee2a0bdec41f544f19eba170cf646c3ce

But the `tensor.pad` op was designed to be able to support this kind of dynamic padding: https://mlir.llvm.org/docs/Dialects/TensorOps/#tensorpad-tensorpadop
I'll say the same thing here as I did there, this is fine for now because I'm not sure how well the rest of the stack supports such pad ops, but it's likely we'll want to rework this lowering at some later point to use pad directly instead of this sequence of extracts/concats. It includes a body to allow dynamically determining the padding value to use based on the current indices.

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
```diff
@@ -79,6 +81,227 @@ class ConvertAtenConstantPadNdOp
...
+      auto createAdds = [&](std::initializer_list<Value> values) {
+        assert(values.size() >= 2);
+        return std::accumulate(values.begin() + 1, values.end(), data(values)[0],
+                             createAdd);
+      };
+      auto createSub = [&](Value x, Value y) {
+        return rewriter.create<arith::SubIOp>(loc, x, y);
+      };
+
+      //extract left and right pad tiles
```
qedawkins, line 188: nit:
```suggestion
      // Extract left and right pad tiles.
```

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
```diff
@@ -79,6 +81,227 @@ class ConvertAtenConstantPadNdOp
...
+          if ((i == BOTTOM) && (vSize > 0))
+            hasBottomPadding = true;
+          if ((j == LEFT) && (hSize > 0))
+            hasLeftPadding = true;
+          if ((j == RIGHT) && (hSize > 0))
+            hasRightPadding = true;
+        }
+      }
+
+      //some generic helper functions
```
qedawkins, line 174: nit: Something like
```suggestion
      // Some generic helper functions to aid in constructing basic arithmetic.
```

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
```diff
@@ -79,6 +81,227 @@ class ConvertAtenConstantPadNdOp
...
+      int64_t hDim = numDims - 1;
+      int64_t vDim = numDims - 2;
+      Value hDimSize = inputShape[hDim];
+      Value vDimSize = inputShape[vDim];
+
+      enum tileHLoc { LEFT = 0, HCENTER = 1, RIGHT = 2 };
+      enum tileVLoc { TOP = 0, VCENTER = 2, BOTTOM = 1, };
+      // vTile denotes the vertical size of the tile
+      // hTile denotes the horizontal size of the tile
+      // the padding results are composed of following tiles
```
qedawkins, line 129: nit: 
```suggestion
      // The padding results are composed of following tiles:
```

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
```diff
@@ -79,6 +81,227 @@ class ConvertAtenConstantPadNdOp
 };
 } // namespace
 
+namespace {
+
+  // Lower aten.replication_pad2d operator into a sequence of
+  // tensor.extract_slice and tensor_insert_slice operations.
```
qedawkins, line 87: It looks like you're lowering into `aten.cat` ops rather than `insert_slice`. Does this comment need updating?

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
```diff
@@ -79,6 +81,227 @@ class ConvertAtenConstantPadNdOp
...
+          //pad vCenterRightSlice on the top
+          SmallVector<int64_t> lowPadding(4, 0);
+          SmallVector<int64_t> highPadding(4, 0);
+          lowPadding[2] = padInts[2];
+          vRightSlice = torch_to_linalg::getPaddedTensor(op, rewriter, vRightSlice, lowPadding, highPadding, topRightValue);
+        }
+        if (hasBottomPadding) {
+          Value bottomRightValue = rewriter.create<tensor::ExtractOp> (loc, input, ValueRange{zero, zero, vDimSizeMinusOne, hDimSizeMinusOne});
+
+          //pad vCenterRightSlice or vRightTopPaddedSlice at the bottom
```
qedawkins, line 285: nit:
```suggestion
          // Pad vCenterRightSlice or vRightTopPaddedSlice at the bottom.
```

qedawkins (commented):
A few other comments specific to the pattern.

projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
lisaliu1: Thanks, corresponding changes have been made per your suggestions.

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
lisaliu1: Thanks, suggested changes have been made.

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
lisaliu1: Thanks, suggested changes have been made.

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
lisaliu1: Thanks, suggested changes have been made.

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
lisaliu1: Thanks, suggested changes have been made.

lib/Conversion/TorchToLinalg/TensorConstructors.cpp
lisaliu1: Thanks, suggested changes have been made.

qedawkins (approved):
Thanks for dropping the formatting changes, LGTM!

PR #2717, opened 2024-01-02
add basic cumsum. this doesn't support the exclusive and reverse attrs

### LINKED ISSUE #2711: onnx.CumSum
working on this

renxida:
fixes #2711


vivekkhandelwal1:
@renxida, generally, we don't merge PRs until the issues are resolved unless it's very urgent and critical and I don't see any urgentness of merging this PR. Ideally, it would be best if you had first addressed the comments, asked for a re-review and then you would have merged it. Also, from the next time please merge the PR only when you get at least one approval from the reviewer.

Also, as I have pointed out this lowering won't work downstream, please see my previous comment.

PR #2597, opened 2023-11-29
Move handling of integer signedness to the backend conversions

ramiro050:
The function `getTypeForScalarType` currently takes an argument to specify the signedness of integer types. This is leakage of backend specific requirements into the torch dialect world. Because `getTypeForScalarType` is a utility function for the torch dialect, it should only produce types that match the sign conventions used by PyTorch (regular integers are signed and unsigned integers are unsigned).

This commit removes the signedness argument from
`getTypeForScalarType`, and moves the backend specific handling of integer types to the backend code.

PR #2718, opened 2024-01-03
[linalg] Added `aten.clamp` support with integers to `torch-to-linalg`

rsuderman:
The lowering for `aten.clamp` did not support integer types. Added
support for integer types including a signed integer test.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -1013,55 +1013,42 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     Type dtype = converter->convertType(clamp.getType())
                      .cast<RankedTensorType>()
                      .getElementType();
-    if (dtype.isa<mlir::FloatType>()) {
-      auto result = payloadArgs[0];
-      if (!min.getType().isa<Torch::NoneType>()) {
-        auto minPromoted = convertScalarToDtype(b, loc, min, dtype);
-        auto pred = b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::ULT,
-                                            result, minPromoted);
-        result = b.create<arith::SelectOp>(loc, pred, minPromoted, result);
-      }
-      if (!max.getType().isa<Torch::NoneType>()) {
-        auto maxPromoted = convertScalarToDtype(b, loc, max, dtype);
-        auto pred = b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGT,
-                                            result, maxPromoted);
-        result = b.create<arith::SelectOp>(loc, pred, maxPromoted, result);
-      }
-      return result;
+    if (!dtype.isa<mlir::FloatType, mlir::IntegerType>()) {
+      clamp.emitError("unimplement type for clamp");
+      return nullptr;
     }
 
-    if (auto intTy = dtype.dyn_cast<mlir::IntegerType>()) {
-      auto result = payloadArgs[0];
-
-      if (!min.getType().isa<Torch::NoneType>()) {
-        auto minPromoted =
-            convertScalarToDtype(b, loc, min, dtype,
-                                 /*srcOriginalDtype=*/std::nullopt,
-                                 /*dstOriginalDtype=*/b.getI64Type());
-        auto pred = b.create<arith::CmpIOp>(loc,
-                                            intTy.isUnsigned()
-                                                ? arith::CmpIPredicate::ult
-                                                : arith::CmpIPredicate::slt,
-                                            result, minPromoted);
-        result = b.create<arith::SelectOp>(loc, pred, minPromoted, result);
-      }
-      if (!max.getType().isa<Torch::NoneType>()) {
-        auto maxPromoted =
-            convertScalarToDtype(b, loc, max, dtype,
-                                 /*srcOriginalDtype=*/std::nullopt,
-                                 /*dstOriginalDtype=*/b.getI64Type());
-        auto pred = b.create<arith::CmpIOp>(loc,
-                                            intTy.isUnsigned()
-                                                ? arith::CmpIPredicate::ugt
-                                                : arith::CmpIPredicate::sgt,
-                                            result, maxPromoted);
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1016,20 +1009,59 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    }
+
+    if (auto intTy = dtype.dyn_cast<mlir::IntegerType>()) {
+      auto result = payloadArgs[0];
+
+      if (!min.getType().isa<Torch::NoneType>()) {
+        auto minPromoted =
+            convertScalarToDtype(b, loc, min, dtype,
+                                 /*srcOriginalDtype=*/std::nullopt,
+                                 /*dstOriginalDtype=*/b.getI64Type());
```
ramiro050, line 1040: These parameters are used to handle signedness properly when truncating/extending, so we should be passing the `clamp.getType().cast<BaseTensorType>().getDtype()`

test/Conversion/TorchToLinalg/elementwise.mlir
```diff
@@ -102,3 +102,20 @@ func.func @elementwise_sinh(%arg0: !torch.vtensor<[3],f32>) -> !torch.vtensor<[3
...
+
+// CHECK-LABEL:    func.func @elementwise_clamp
+// CHECK:            linalg.generic
+// CHECK:            arith.trunci
+// CHECK:            arith.cmpi slt
+// CHECK:            arith.select
+// CHECK:            arith.trunci
+// CHECK:            arith.cmpi sgt
+// CHECK:            arith.select
+func.func @elementwise_clamp(%arg0: !torch.vtensor<[1,3,8,8],si8>) -> !torch.vtensor<[1,3,8,8],si8> {
```
ramiro050, line 116: For these type of changes, we use e2e tests rather than lit tests. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

Tests should be added here: https://github.com/llvm/torch-mlir/blob/1778314620b796de7a7aba61f00396cecbd29a0b/projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py#L860

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1016,20 +1009,59 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+        auto pred = b.create<arith::CmpIOp>(loc,
+                                            intTy.isUnsigned()
```
ramiro050, line 1042: In torch-mlir, this should always be signless. See: https://github.com/llvm/torch-mlir/pull/2597. I believe what you want to check is the signedness of `clamp.getType().cast<BaseTensorType>().getDtype()`.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1016,20 +1009,59 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
-                                          result, minPromoted);
-      result = b.create<arith::SelectOp>(loc, pred, minPromoted, result);
-    }
-    if (!max.getType().isa<Torch::NoneType>()) {
-      auto maxPromoted = convertScalarToDtype(b, loc, max, dtype);
-      auto pred = b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGT,
-                                          result, maxPromoted);
-      result = b.create<arith::SelectOp>(loc, pred, maxPromoted, result);
+
+    Type dtype = converter->convertType(clamp.getType())
```
ramiro050, line 1013: The logic here is a bit repetitive. I think we can do something like

```c++
    if (!dtype.isa<mlir::FloatType, mlir::IntegerType>())
      return clamp.emitError(...);

    Type dstOriginalDtype = clamp.getType().cast<BaseTensorType>().getDtype();
    auto clamp = [&](Value input, Value clamp, bool isMax) -> Value {
        auto clamp =
          convertScalarToDtype(b, loc, clamp, dtype,
                               /*srcOriginalDtype=*/std::nullopt,
                               /*dstOriginalDtype=*/dstOriginalDtype);
        Value cmp = isMax
          ? arith::CmpFPredicate::UGT : arith::CmpFPredicate::ULT;
        Value pred;
        if (dtype.isa<mlir::FloatType>()) {
          pred = b.create<arith::CmpFOp>(loc, cmp, input, clamp);
        } else if (dtype.isa<mlir::IntegerType>()) {
          cmp = convertToIntegerPredicate(cmp, dstOriginalDtype.isUnsigned());
          pred = b.create<arith::CmpIOp>(loc, cmp, input, clamp);
        }
        return b.create<arith::SelectOp>(loc, pred, clamp, input);
    };

    auto result = payloadArgs[0];
    if (!min.getType().isa<Torch::NoneType>())
      result = clamp(result, min, /*isMax=*/false);
    if (!max.getType().isa<Torch::NoneType>())
      result = clamp(result, max, /*isMax=*/true);
    return result;
```

ramiro050 (changes requested):
Thanks!

lib/Conversion/TorchToLinalg/Uncategorized.cpp
rsuderman: I updated the check using the dtype. I have a separate PR in progress that should include a signedness helper.

test/Conversion/TorchToLinalg/elementwise.mlir
rsuderman: Updated and removed the old tests.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
rsuderman: SG

lib/Conversion/TorchToLinalg/Uncategorized.cpp
rsuderman: Taken and modified due to `convertToIntegerPredicate` not being available outside of `ArithOps.cpp`

ramiro050 (approved):
Thanks!

PR #2721, opened 2024-01-03
Do not try to legalize transposed convolution

srinathava:
Currently transposed convolution is not handled correctly by `TorchToTosa`. This PR allows transposed convolutions to pass through the conversion so that they can be handled by other conversion passes later in a pipeline.

An example input which produces a compilation error is:

```
func.func @forward(%input: !torch.vtensor<[1,64,1,100],f32>) -> !torch.vtensor<[1,64,2,200],f32> {
  %int1 = torch.constant.int 1
  %int2 = torch.constant.int 2
  %weight = torch.vtensor.literal(dense<0.0> : tensor<64x64x3x3xf32>) : !torch.vtensor<[64,64,3,3],f32>
  %bias = torch.vtensor.literal(dense<0.0> : tensor<64xf32>) : !torch.vtensor<[64],f32>
  %stride = torch.prim.ListConstruct %int2, %int2 : (!torch.int, !torch.int) -> !torch.list<int>
  %int1x1 = torch.prim.ListConstruct %int1, %int1 : (!torch.int, !torch.int) -> !torch.list<int>
  %output = torch.aten.convolution %input, %weight, %bias, %stride, %int1x1, %int1x1, %true, %int1x1, %int1 : !torch.vtensor<[1,64,1,100],f32>, !torch.vtensor<[64,64,3,3],f32>, !torch.vtensor<[64],f32>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool, !torch.list<int>, !torch.int -> !torch.vtensor<[1,64,2,200],f32>
  return %output : !torch.vtensor<[1,64,2,200],f32>
}
```

This MLIR produces an error about a cast operation with a size mismatch when passed through `torch-to-tosa`:

```
 error: 'tensor.cast' op operand type 'tensor<1x64x1x50xf32>' and result type 'tensor<1x64x2x200xf32>' are cast incompatible
```


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -5208,6 +5207,15 @@ class ConvertTorchToTosa : public ConvertTorchToTosaBase<ConvertTorchToTosa> {
     INSERT_ATENOP_PATTERN(AtenIscloseOp);
 #undef INSERT_ATENOP_PATTERN
 
+    auto isTransposedConvOp = [&](AtenConvolutionOp op) {
+      bool transposed = false;
+      if (matchPattern(op.getTransposed(), m_TorchConstantBool(&transposed)))
+        return transposed;
+      return false;
+    };
+    target.addDynamicallyLegalOp<AtenConvolutionOp>(isTransposedConvOp);
```
ramiro050, line 5216: This doesn't feel right to me. The `Conversion` pattern `TorchToTosa` is expected to output a graph that does not contain any torch dialect ops

https://github.com/llvm/torch-mlir/blob/bafe8ca55757a1ea70dffb39823748c3e96e30dc/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L5000

We should fix the conversion pattern for transposed convolution to return a `notifyMatchFailure` when the inputs are for a convolution that is currently unsupported.

If support for the convolution in your example MLIR is needed, then we should add support for it rather than letting things propagate down.

ramiro050 (changes requested):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
srinathava: Hi Mario, back to the drawing board for me :) I'll convert this into a `notifyMatchFailure` for now.

sjarus: Sorry I'm late to see this coming back from vacation. I agree with @ramiro050 , let's coordinate the fix here internally @srinathava - I'll help you.

srinathava: Hi Ramiro, can you take another look with the new changes?

ramiro050 (approved):
LGTM!

PR #2722, opened 2024-01-03
[torch-mlir] update e2e test class documentation

aartbik:
The doc seems copy-and-paste from the linalg-on-tensors class


ramiro050 (approved):
Thanks!

jpienaar (commented):
Thanks!

PR #2725, opened 2024-01-04
[MLIR] [ONNX] lowering for onnx tile op and sign op

PhaneeshB:


PhaneeshB:
cc: @vivekkhandelwal1 for review

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1343,4 +1343,32 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                                                           data, dimValueList);
         return success();
       });
+
+  patterns.onOp("Tile", 13,
```
vivekkhandelwal1, line 1347: You can add this for version 6. Both the versions don't have any change so it's better to support the older one.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1343,4 +1343,32 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                });
+
+  patterns.onOp(
+      "Sign", 13, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+      Torch::ValueTensorType resultType;
+      Value operand;
+      if (binder.tensorOperand(operand) || binder.tensorResultType(resultType))
+          return failure();
+
+      rewriter.replaceOpWithNewOp<Torch::AtenSignOp>(binder.op, resultType, operand);
```
vivekkhandelwal1, line 1371: Have you used clang-format?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1343,4 +1343,32 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                  binder.getLoc(),
+                  Torch::ListType::get(rewriter.getType<Torch::IntType>()), repeatDims);
+
+                  rewriter.replaceOpWithNewOp<Torch::AtenTileOp>(
+                      binder.op, resultType, operand, repeatDimsToList.getResult(0));
+                  return success();
+                });
+
+  patterns.onOp(
+      "Sign", 13, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
```
vivekkhandelwal1, line 1365: Similarly, you can support it for version number 9.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1343,4 +1343,32 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+                  Torch::ValueTensorType resultType;
+                  Value operand;
+                  Value repeatDims;
+                  if (binder.tensorOperands(operand, repeatDims) ||
+                      binder.tensorResultType(resultType))
+                    return failure();
+                  Torch::PrimTolistOp repeatDimsToList = rewriter.create<Torch::PrimTolistOp>(
+                  binder.getLoc(),
+                  Torch::ListType::get(rewriter.getType<Torch::IntType>()), repeatDims);
```
Shukla-Gaurav, line 1357: I think you have to extract value from repeatDims tensor and then create a list out of it. 
you can use `torch.index_select` to extract values.

PhaneeshB: Thanks for pointing this out, I had used torch.aten.item to extract individual elements for the 1-D tensor, will change it to index_select

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
PhaneeshB: Oh, I had the impression that it would be part of the CI.
used now, thanks for pointing this out 🤝

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
PhaneeshB: It seems there should be a better way to do this 
The earlier implementation had `PrimTolistOp` (which I think comes from [Torch.Tensor.toList](https://pytorch.org/docs/stable/generated/torch.Tensor.tolist.html#torch.Tensor.tolist) )
There is also `Aten_TensorToListOp` which would be the exact fit here but I think it is not yet lowered to linalg.

I could possibly take this up next as there is a recurring need to convert `1-D tensors` to `List(int)` for onnx to torch lowering

vivekkhandelwal1 (approved):
LGTM

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: I think we can assign `Aten_TensorToListOp` to any folk from the turbine camp, would be a good learning exercise for them. @kumardeepakamd, do you want to assign this to someone?

rsuderman (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
PhaneeshB: Marking this resolved will update with `Aten_TensorToListOp` in a future PR as needed.

PhaneeshB:
@vivekkhandelwal1 @Shukla-Gaurav @AmosLewis @rsuderman 
would be great help if any of you can help me merge this
Thanks in Advance :)

PR #2731, opened 2024-01-05
[torch][quant] Support quantize and dequantize for torch

rsuderman:
Handle both `torch.dequantize` and `torch.quantize_per_tensor` including
the op based quantization parameter tracking. This includes adding
`qint32` to torch types as it was missing during the initial type
inclusion.

For testing we only have `torch.int8` and `torch.float` types on
function boundaries as the `qint8` types require passing the scale
and zero point quantization information which is not supported yet.


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1316,6 +1316,86 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
   }
 
+  if (isa<AtenDequantizeTensorOp, AtenDequantizeSelfOp>(op)) {
+    auto value = payloadArgs[0];
+    auto qtensor = op->getOperand(0);
+    auto qtensorTy = qtensor.getType().cast<ValueTensorType>().getDtype();
+    auto makeQTensor =
+        qtensor.getDefiningOp<Aten_MakePerTensorQuantizedTensorOp>();
+    if (!makeQTensor) {
+      qtensor.dump();
```
qedawkins, line 1326: Leftover debug print?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1316,6 +1316,86 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+    } else {
+      value = b.create<arith::SIToFPOp>(loc, payloadArgs[1].getType(), value);
+    }
+
+    Value zp = makeQTensor.getZeroPoint();
+    zp = converter->materializeTargetConversion(
+        b, loc, converter->convertType(zp.getType()),
+        makeQTensor.getZeroPoint());
+    zp = b.create<arith::SIToFPOp>(loc, value.getType(), zp);
+    value = b.create<arith::SubFOp>(loc, value, zp);
```
qedawkins, line 1342: If we're converting both the value and the zero point to floating point, could we instead do the subtraction on integers and then do the conversion? Maybe signed/unsigned extending to the bitwidth of the target floating point type first. Just a suggestion, and not sure that it really matters.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1316,6 +1316,86 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
+
+    Value scale = makeQTensor.getScale();
+    scale = converter->materializeTargetConversion(
+        b, loc, converter->convertType(scale.getType()),
+        makeQTensor.getScale());
+    scale = b.create<arith::TruncFOp>(loc, value.getType(), scale);
```
qedawkins, line 1348: Is the scale expected to always be a larger bit width than the target of `dequantize_per_tensor`? Or could we have, say, `f32` and `f32`? You can use `convertScalarToDtype(b, loc, scale, value.getType())` to handle other cases if so.

lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -559,3 +559,20 @@ FailureOr<Type> torch_to_linalg::getBackendTypeForScalarType(
...
+    return false;
+  if (isa<QInt8Type>(type))
+    return false;
+  if (isa<QUInt8Type>(type))
+    return true;
+  if (isa<QInt32Type>(type))
+    return false;
+  if (auto intTy = dyn_cast<IntegerType>(type))
+    return intTy.isUnsigned();
+  assert(false);
```
qedawkins, line 576: nit: use llvm_unreachable

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -4136,6 +4136,51 @@ def ElementwiseBitwiseAndScalarInt8Module_basic(module, tu: TestUtils):
...
+    def forward(self, x):
+        qx = torch._make_per_tensor_quantized_tensor(x, 0.1, 8)
+        qx = torch.dequantize(qx)
+        return qx
+
+@register_test_case(module_factory=lambda: ElementwiseDequantizePerTensorModule())
+def ElementwiseDequantizePerTensorModule_basic(module, tu: TestUtils):
+    module.forward(tu.randint(3, 4, low=-128, high=127).to(torch.int8))
+
+# ==============================================================================
```
qedawkins, line 4182: Out of curiosity, what is `aten.dequantize_self_per_tensor`? Is it worth trying to test here?

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -4136,6 +4136,51 @@ def ElementwiseBitwiseAndScalarInt8Module_basic(module, tu: TestUtils):
...
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float, True),
+    ])
+    def forward(self, x):
+        scale = 0.04
+        zp = -110
+        dtype = torch.qint8
+        q = torch.quantize_per_tensor(x, scale, zp, dtype)
+        return q.int_repr()
```
qedawkins, line 4154: IIUC this `int_repr` is load bearing to the test because we don't support returning `qint*` types; maybe worth adding a note? It's clearer from the lowering for dequantization.

qedawkins (approved):
LGTM, just a few nits.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
rsuderman: Removed.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1316,6 +1316,86 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
 
+  if (isa<AtenDequantizeTensorOp, AtenDequantizeSelfOp>(op)) {
+    auto value = payloadArgs[0];
+    auto qtensor = op->getOperand(0);
+    auto qtensorTy = qtensor.getType().cast<ValueTensorType>().getDtype();
+    auto makeQTensor =
+        qtensor.getDefiningOp<Aten_MakePerTensorQuantizedTensorOp>();
+    if (!makeQTensor) {
+      qtensor.dump();
+      op->emitError("dequantizing tensor of unknown scale / zeorpoint");
```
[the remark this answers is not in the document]
rsuderman: Fixed. Changed to `unimplemented` though the error is more likely due to incorrect munging of side channeled information.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
rsuderman: Performed in fixed point math. It should not technically matter except in extreme cases of `int32` dequantization, but this is good enough.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
rsuderman: This is more of a quirk of `torch` number values. They are always assumed to be `64-bit`, so we should pretty much always expect them to be truncated.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
qedawkins: Sounds good to me

lib/Conversion/TorchToLinalg/Utils.cpp
rsuderman: Added.

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
rsuderman: Added a comment - and yes, its because IREE does not yet support adding the sidechannel information required for `quint*` types.

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
rsuderman: Its the same as `aten.quantize_per_tensor`. The only difference is whether its the lib call or member call. AFAICT we cannot actually generate the none-self version.

PR #2733, opened 2024-01-08
Fixing implicit double->float truncation warnings.

benvanik:
Floating-point literals should use the correct type specifier.


qedawkins (approved):

benvanik:
(mind merging? I'm not part of the cool kids club)

qedawkins:
I can't believe the CI finished that fast. 2-3 hours is standard...

PR #2734, opened 2024-01-09
torch-to-tosa lowering support for AtenLinalgVectorNormOp

zezhang:
This PR add torch-to-tosa lowering support for AtenLinalgVectorNormOp

e2e test:
python -m e2e_testing.main --config=tosa

LIT tests:


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -681,15 +681,63 @@ class ConvertAtenReductionOp : public OpConversionPattern<AtenOpT> {
...
+          OpConversionPattern<AtenOpT>::getTypeConverter()->convertType(
+              self.getType()),
+          self);
+      auto powVal = rewriter.create<tosa::PowOp>(
+          op.getLoc(),
+          OpConversionPattern<AtenOpT>::getTypeConverter()->convertType(
+              self.getType()),
+          absVal, ordValue);
+      self = powVal;
+    }
```
navahgar, line 719: Why is this in `ConvertAtenReductionOp` and not in say `ConvertAtenMultipleDimsReductionOp` (which is what we use to instantiate the convertor for this op) or a new one for this op, say `ConvertAtenLinalgVectorNormOp`?

navahgar: Since you are special casing for `AtenLinalgVectorNormOp` anyways here, it seems having a separate convertor for this op would be better. 

@ramiro050 Do you have any recommendations for this?

ramiro050: I think we just need to define a `conversionFuncT` for the full norm reduction, and pass that as the template parameter rather than `mlir::tosa::convertReduceSumOp`. It seems to me that that was the original intention with this template pattern.

zezhang: Replace the specialization with `mlir::tosa::convertLinalgVectorNormOp` function

navahgar: @zezhang I think what @ramiro050 was referring to was to define a new reduction function for norm, something like `convertReduceNormOp` that will replace `convertReduceSumOp` used here: https://sourcegraph.com/github.com/llvm/torch-mlir/-/blob/lib/Conversion/TorchToTosa/TorchToTosa.cpp?L5111

zezhang: Yes, that's what I did. `mlir::tosa::convertLinalgVectorNormOp` is the new `conversionFuncT` added in  `TosaLegalizeCommon.cpp` so that the original `ConvertAtenReductionOp` template remain unchanged.

navahgar: Oh yeah. I misread your changes earlier. Thanks for clarifying. That matches my understanding as well.

ramiro050 (approved):
LGTM!

PR #2735, opened 2024-01-09
Added aten.pool_max3d support to torch-to-linalg

wu-s-john:
Added verification logic to the abstract_interpreter_lib_gen.py

Also made some unit tests

Initially, I thought we can use `linalg::pooling_ndhwc_max` to help implement this problem. However, on a 5-dimensional matrix it does the pooling on dimensions (2, 3, 4) which is not what we want. We want pooling on dimensions (3, 4, 5).

To achieve this, we would need to lower our code using the `linalg` dialect.

Turns out the pooling code in `linalg` looks like this.

```
func @max_pooling_ncdhw(%I: memref<?x?x?x?x?xf32>, %K: memref<3xindex>, %O: memref<?x?x?x?x?xf32>,
                        %strides: memref<3xindex>, %dilations: memref<3xindex>) {
    %N = memref.dim %I, %c0 : memref<?x?x?x?x?xf32>
    %C = memref.dim %I, %c1 : memref<?x?x?x?x?xf32>
    %D = memref.dim %I, 2 : memref<?x?x?x?x?xf32>
    %H = memref.dim %I, 3 : memref<?x?x?x?x?xf32>
    %W = memref.dim %I, 4 : memref<?x?x?x?x?xf32>

    %kernel_d = memref.load %K[%c0] : memref<3xindex>
    %kernel_h = memref.load %K[%c1] : memref<3xindex>
    %kernel_w = memref.load %K[2] : memref<3xindex>
    %stride_d = memref.load %strides[%c0] : memref<3xindex>
    %stride_h = memref.load %strides[%c1] : memref<3xindex>
    %stride_w = memref.load %strides[2] : memref<3xindex>
    %dilation_d = memref.load %dilations[%c0] : memref<3xindex>
    %dilation_h = memref.load %dilations[%c1] : memref<3xindex>
    %dilation_w = memref.load %dilations[2] : memref<3xindex>

    linalg.generic {
        indexing_maps = [
            affine_map<(n, c, d, h, w, kd, kh, kw) -> (n, c, d * %stride_d + kd * %dilation_d, h * %stride_h + kh * %dilation_h, w * %stride_w + kw * %dilation_w)>,  // Map for input tensor
            affine_map<(n, c, d, h, w, kd, kh, kw) -> (kd, kh, kw)>,                                              // Map for kernel tensor
            affine_map<(n, c, d, h, w, kd, kh, kw) -> (n, c, d, h, w)>                                            // Map for output tensor
        ],
        iterator_types = ["parallel", "parallel", "parallel", "parallel", "parallel", "reduction", "reduction", "reduction"],
        doc = "3D Max Pooling NCDHW with Strides, Dilations, and Kernel Size"
    } ins(%I, %K : memref<?x?x?x?x?xf32>, memref<3xindex>) outs(%O : memref<?x?x?x?x?xf32>) {
        ^bb0(%input_elem: f32, %kernel_elem: index, %output_elem: f32):
            %max_val = arith.maxf %input_elem, %output_elem : f32
            linalg.yield %max_val : f32
    }
    return
}

```

This was implemented based on it's source code with the adjustments mentioned above:
https://github.com/llvm/llvm-project/blob/4ca1b5e094280ef1af40412e3cfcb62dc3cf15bc/mlir/include/mlir/Dialect/Linalg/IR/LinalgNamedStructuredOps.yaml#L5647

Issues related to this can be found here

https://github.com/nod-ai/SHARK-Turbine/issues/324


wu-s-john:
@rsuderman @vivekkhandelwal1 @stellaraccident please review when you have time.

I maybe be missing things for this implementation so please let me know if I am missing anything. Integration tests may fail as a result.

vivekkhandelwal1:
Hi @wu-s-john, the tests in pooling.mlir are failing. Probably your changes might have resulted in some changes in the existing lowering or the newly added tests would not be matching the checks, please take a look at that.

wu-s-john:
@vivekkhandelwal1 and @rsuderman . 

It seems that the tests are passing. Please review!

test/Conversion/TorchToLinalg/pooling.mlir
ramiro050, line 1: lit tests are not needed for these type of PRs. See: https://github.com/llvm/torch-mlir/blob/main/docs/architecture.md#dos-and-donts-for-unit-vs-end-to-end-testing

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -617,9 +617,82 @@ def aten〇_unsafe_view〡shape(self: List[int], size: List[int]) -> List[int]:
 def aten〇resize_〡shape(self: List[int], size: List[int], memory_format: Optional[int] = None) -> List[int]:
     return size
 
+def _max_pool3d(
```
ramiro050, line 620: This function has some complex logic. Can you add tests for it? See example:

https://github.com/llvm/torch-mlir/blob/9e2a2851898fbd83df28674e2561d52910f3e83e/projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py#L454

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -137,57 +162,159 @@ static LogicalResult createPoolingOp(
...
 
-
 namespace {
-class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
+template <typename OpTy, int Dim>
+class ConvertAtenMaxPoolOp : public OpConversionPattern<OpTy> {
+  using OpConversionPattern<OpTy>::OpConversionPattern;
+
+private:
+  LogicalResult writePoolingMax3D(AtenMaxPool3dOp &op,
```
ramiro050, line 174: nit: I would call this `createPoolingMax3D` to match the `createPoolingOp` name

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -137,57 +162,159 @@ static LogicalResult createPoolingOp(
...
+                                  typename OpTy::Adaptor &adaptor,
+                                  ConversionPatternRewriter &rewriter,
+                                  TypedAttr &smallestFPValueAttr, Value &self,
```
ramiro050, line 177: nit: `smallestFPValueAttr` does not need to be a parameter. This function only applies to `AtenMaxPool3dOp`, so the smallest FP value should always be the same. I would just generate it inside this helper.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -137,57 +162,159 @@ static LogicalResult createPoolingOp(
...
+                                  Type &elementType,
```
ramiro050, line 178: Same comment for `elementType`. You can get this from `op`, which is already a parameter

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -137,57 +162,159 @@ static LogicalResult createPoolingOp(
...
-
 namespace {
-class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
+template <typename OpTy, int Dim>
+class ConvertAtenMaxPoolOp : public OpConversionPattern<OpTy> {
+  using OpConversionPattern<OpTy>::OpConversionPattern;
+
+private:
+  LogicalResult writePoolingMax3D(AtenMaxPool3dOp &op,
+                                  typename OpTy::Adaptor &adaptor,
```
ramiro050, line 175: nit: I think `adaptor` is unsused

lib/Conversion/TorchToLinalg/Pooling.cpp
ramiro050, line 177: nit: you can get `self` from `op`

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -137,57 +162,159 @@ static LogicalResult createPoolingOp(
...
                              outTensorInitialized, stridesAttr, dilationAttr)
                .getResult(0);
-
   return success();
 }
 
-
 namespace {
-class ConvertAtenMaxPool2dOp : public OpConversionPattern<AtenMaxPool2dOp> {
+template <typename OpTy, int Dim>
```
ramiro050, line 169: I don't think we need the `Dim` parameter. You can use `if (std::is_same<OpTy, AtenMaxPool2dOp>()) ...`. This also makes it more explicit in the code how each op is being handled.

ramiro050 (changes requested):
Looks mostly good. A few initial comments

test/Conversion/TorchToLinalg/pooling.mlir
wu-s-john: Wait, what do u mean by lit tests? I feel that the unit test I constructed is useful as the lowering from torch to linalg for maxpool3d as it was not a trivial transformation.

lib/Conversion/TorchToLinalg/Pooling.cpp
wu-s-john: Good idea! Thanks!

lib/Conversion/TorchToLinalg/Pooling.cpp
wu-s-john: `self` and `elementType` are actually derived from adaptor. So, I will keep that.

lib/Conversion/TorchToLinalg/Pooling.cpp
wu-s-john: Good idea! Fixed!

wu-s-john:
@ramiro050 Addressed all comments! Please check again!

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -617,9 +617,120 @@ def aten〇_unsafe_view〡shape(self: List[int], size: List[int]) -> List[int]:
...
+
+    if len(input) == 4:
+        return [nInputPlane, outputDepth, outputHeight, outputWidth]
+    else:
+        return [nbatch, nInputPlane, outputDepth, outputHeight, outputWidth]
+    
 def aten〇max_pool2d〡shape(self: List[int], kernel_size: List[int], stride: List[int] = (), padding: List[int] = (0, 0,), dilation: List[int] = (1, 1,), ceil_mode: bool = False) -> List[int]:
     return upstream_shape_functions.max_pool2d(self, kernel_size, stride, padding, dilation, ceil_mode)
 
+@check_shape_function([
```
wu-s-john, line 724: @renxida Since you were implementing `conv_tbc` as an operator for torch-mlir, I would recommend creating unit tests for it's shape function. 

Coming from a math and ML background, this is my first time seeing `conv_tbc`. It would be good to have this unit tests to see when `conv_tbc` will be invalid or valid. 

I feel that I exercised the validity of the shape pretty comprehensively for `max_pool3d`.

lib/Conversion/TorchToLinalg/Pooling.cpp
ramiro050: Ah, missed that. Makes sense to keep it.

test/Conversion/TorchToLinalg/pooling.mlir
ramiro050: Usually the lit tests (CHECKS in `*.mlir` files) for conversion patterns in torch-mlir turn into just an MLIR version of the C++ code. Every `rewriter.create` turns into a `CHECK`. So, we have not found these types of tests that useful for catching errors in the lowerings. The e2e tests are the ones that catch such bugs because we are able to compare directly to PyTorch.

However, looking at your lit test again, I do think it is useful to see the affine maps created, so it's fine with me if we leave it.

ramiro050 (approved):
LGTM! Any idea why the CI is failing?

wu-s-john:
> LGTM! Any idea why the CI is failing?

Checking!

rsuderman:
> > LGTM! Any idea why the CI is failing?
> 
> Checking!

The CI was having some issues. I think Stella managed to fix it up now. See now passing presubmits.

PR #2738, opened 2024-01-09
[ONNX][MLIR] Add support for pad op in the onnx pipeline

Shukla-Gaurav:
This commit adds mapping from `onnx.pad` op to `torch.pad` op. Currently it does not support `axes` parameter of `onnx.pad` op.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -708,6 +708,100 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
                       binder.op, resultType, operand, constAlpha);
                   return success();
                 });
+  patterns.onOp(
+      "Pad", 19, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value data, pads, constantValue, axes;
+        std::string mode;
+
+        // TODO: The `axes` parameter is not supported yet.
+        if (!binder.tensorOperandAtIndex(axes, 3)) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "The axes parameter is not supported yet");
+        }
+        if (binder.tensorOperandAtIndex(data, 0) ||
+            binder.tensorOperandAtIndex(pads, 1) ||
+            binder.tensorOperandAtIndex(constantValue, 2) ||
+            binder.tensorResultType(resultType) ||
+            binder.customOpNameStringAttr(mode, "mode", "constant"))
+          return failure();
+        Location loc = binder.getLoc();
+
+        // Get pads shape and rank. The pads tensor is expected to be 1-D
+        // tensor.
+        auto padsTensorType = pads.getType().cast<Torch::ValueTensorType>();
+        if (!padsTensorType || !padsTensorType.hasSizes()) {
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "Expect non empty pad tensor");
+        }
+        ArrayRef<int64_t> padsShape = padsTensorType.getSizes();
+        int64_t padsRank = padsShape.size();
+        if (padsRank != 1) {
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "Expect 1-D pad tensor");
+        }
+
+        // Extract all the values of 1-D pad tensor and create a list of all
+        // these values as torch.pad op expects pad list.
+        int64_t padsSize = padsShape[0];
+        Value constZero = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(0));
+        SmallVector<Value> padsTensorValue;
+        SmallVector<int64_t> emptyShape;
+        Type padsElemType =
+            Torch::ValueTensorType::get(padsTensorType.getContext(), emptyShape,
+                                        padsTensorType.getOptionalDtype());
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -469,6 +469,87 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        Value data, pads, constantValue;
+        std::string mode;
+        if (binder.tensorOperandAtIndex(data, 0) ||
+            binder.tensorOperandAtIndex(pads, 1) ||
+            binder.tensorOperandAtIndex(constantValue, 2) ||
+            binder.tensorResultType(resultType) ||
+            binder.customOpNameStringAttr(mode, "mode", "constant"))
+          return failure();
+        Location loc = binder.getLoc();
+
```
vivekkhandelwal1, line 484: Please add a check for the `axes` to exist or not and if it exists then add a failure message.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -469,6 +469,87 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          Value index =
+              Torch::createRank0Tensor(rewriter, loc, padsTensorType, pos);
+          padsTensorValue.emplace_back(
+              rewriter.create<Torch::AtenIndexSelectOp>(loc, padsElemType, pads,
+                                                        constZero, index));
+        }
+        // Rearrange pad tensor to satisfy torch.pad op semantics.
+        SmallVector<Value> padsRearrange;
+        for (unsigned i = 0; i < padsSize / 2; i++) {
+          padsRearrange.emplace_back(padsTensorValue[padsSize / 2 - 1 - i]);
```
vivekkhandelwal1, line 517: better to use paranthesis here

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -469,6 +469,87 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+                    padsRearrange)
+                .getResult(0);
+        Value modeVal = rewriter.create<Torch::ConstantStrOp>(
+            loc, rewriter.getStringAttr(mode));
+
+        auto constValueType =
+            constantValue.getType().cast<Torch::ValueTensorType>();
+        if (!constValueType) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expect constant value to be a tensor");
```
vivekkhandelwal1, line 535: I think this error message is not correct because the constant value would always be a tensor whenever present. You should instead say that it should be non-none.

vivekkhandelwal1 (commented):
For future purpose, it would be better if you can add a comment at the beginning explaining your lowering, since it's not a direct op conversion.

Shukla-Gaurav:
@vivekkhandelwal1 thanks for the review, I have added comments at different places about the algorithm. Can you please review again?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -548,6 +548,103 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        // these values as torch.pad op expects pad list.
+        int64_t padsSize = padsShape[0];
+        Value constZero = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(0));
+        SmallVector<Value> padsTensorValue;
+        SmallVector<int64_t> emptyShape;
+        Type padsElemType =
+            Torch::ValueTensorType::get(padsTensorType.getContext(), emptyShape,
+                                        padsTensorType.getOptionalDtype());
+        for (unsigned i = 0; i < padsSize; ++i) {
```
rsuderman, line 594: Generally `llvm` tends to use `uint32_t` instead of `unsigned` when doing unsigned types (or `uint64_t`). This is for bit depth clarity.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -548,6 +548,103 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          Value pos = rewriter.create<Torch::ConstantIntOp>(
+              loc, rewriter.getI64IntegerAttr(i));
+          Value index =
+              Torch::createRank0Tensor(rewriter, loc, padsTensorType, pos);
```
rsuderman, line 598: Considering the number of invocations, we likely want to not use this helper. This is what results in the excessive number of `ConstantNoneOps` and `ConstructListOp`. If we directly invoke we can reuse the same values between them, making the test more succinct.

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
Shukla-Gaurav: Thanks for pointing out this :) 
There is another better op `torch.select.int` to handle this which does not even need the index to be a 0-d tensor.

rsuderman (approved):
Awesome, that cleaned things up a ton!

PR #2739, opened 2024-01-09
Handle torch.none type in tosa.clamp op 

zezhang:
This PR updates the torch-to-tosa conversion with following changes:

- Support torch.none as min/max input argument for tosa.clamp op
- Support negative value as start index for tosa.slice op
- Add tosa.logical_or lowering support

e2e test:
python -m e2e_testing.main --config=tosa

LIT tests:


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3986,34 +3985,70 @@ LogicalResult ConvertAtenOp<AtenClampOp>::matchAndRewrite(
...
+      max_int = rewriter.getI64IntegerAttr(std::numeric_limits<int64_t>::max());
+      max_fp = rewriter.getF32FloatAttr(std::numeric_limits<float>::max());
+    } else {
+      if (!matchPattern(op.getMax(), m_TorchConstantFloat(&fp_max)))
+        return rewriter.notifyMatchFailure(
+            op,
+            "unimplemented: value `fp_max` should be a torch constant float");
+      max_int = rewriter.getI64IntegerAttr(static_cast<int64_t>(fp_max));
+      max_fp = rewriter.getF32FloatAttr(static_cast<float>(fp_max));
+    }
```
navahgar, line 4032: We are repeating this code a couple of times. Can we restructure this better to avoid code duplication?

navahgar (commented):
Mostly LGTM. One comment regarding refactoring below.

navahgar:
Can you also update the PR title to be more clear? You can even make the title just about handling none in clamp, since that is the biggest change. Then, you can add the other two in the PR description.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
zezhang: Just updated the format. PTAL.

navahgar (commented):
LGTM. 

@ramiro050, @sjarus - in case you have any other comments here.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -8,6 +8,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "torch-mlir/Conversion/TorchToTosa/TorchToTosa.h"
+#include "mlir/Support/LogicalResult.h"
```
ramiro050, line 11: nit: move lower down with other `mlir` imports (they are also alphabetized within each group)

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3337,8 +3338,7 @@ LogicalResult ConvertAtenOp<AtenSliceTensorOp>::matchAndRewrite(
     return rewriter.notifyMatchFailure(op, "start must be a Scalar constant");
 
   if (start < 0)
-    return rewriter.notifyMatchFailure(op, "Currently unsupported: start < 0");
-
+    start = toPositiveDim(start, selfType.getShape()[dim]);
```
ramiro050, line 3341: nit: contrary to what the name would suggest, `toPositiveDim` is not guaranteed to output a positive dim. You have to verify this:

https://github.com/llvm/torch-mlir/blob/aee1fca2517b8bff3b18e3b01beaafe8d57a7dd8/lib/Conversion/TorchToStablehlo/Reduction.cpp#L238-L241

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3986,34 +3986,72 @@ LogicalResult ConvertAtenOp<AtenClampOp>::matchAndRewrite(
...
-  if (op.getMin().getType().isa<Torch::FloatType>()) {
-    double fp_min, fp_max;
-    if (!matchPattern(op.getMin(), m_TorchConstantFloat(&fp_min)))
-      return rewriter.notifyMatchFailure(
-          op, "unimplemented: value `fp_min` should be a torch constant float");
 
-    if (!matchPattern(op.getMax(), m_TorchConstantFloat(&fp_max)))
-      return rewriter.notifyMatchFailure(
-          op, "unimplemented: value `fp_max` should be a torch constant float");
+  auto getValAttr = [&](Value operand, bool isFloat, IntegerAttr &intAttr,
```
ramiro050, line 3990: There's a lot of repetitive control flow here. I think we can avoid a lot of it by doing

```c++
  IntegerAttr min_int = rewriter.getI64IntegerAttr(std::numeric_limits<int64_t>::min());
  IntegerAttr max_int = rewriter.getI64IntegerAttr(std::numeric_limits<int64_t>::max());
  FloatAttr min_fp = rewriter.getF32FloatAttr(std::numeric_limits<float>::lowest());
  FloatAttr max_fp = rewriter.getF32FloatAttr(std::numeric_limits<float>::max());

  auto getValAttr = [&](Value operand, IntegerAttr &intAttr,
                        FloatAttr &fpAttr) -> LogicalResult {
    double valFloat;
    int64_t valInt;
    if (matchPattern(operand, m_TorchConstantFloat(&valFloat))) {
      intAttr = rewriter.getI64IntegerAttr(static_cast<int64_t>(valFloat));
      fpAttr = rewriter.getF32FloatAttr(static_cast<float>(valFloat));
    } else if (matchPattern(operand, m_TorchConstantInt(&valInt))) {
      intAttr = rewriter.getI64IntegerAttr(valInt);
      fpAttr = rewriter.getF32FloatAttr(static_cast<float>(valInt));
    } else {
      return failure();
    }
    return success();
  };

  LogicalResult minAttrResult = getValAttr(op.getMin(), min_int, min_fp);
  LogicalResult maxAttrResult = getValAttr(op.getMax(), max_int, max_fp);
  if (failed(minAttrResult) && failed(maxAttrResult)) {
        op, "either `min` or `max` should be a torch constant");
  }
  if (failed(minAttrResult) && checkNotNone(rewriter, op, op.getMin())) {
    return rewriter.notifyMatchFailure(op, "min attr should be a torch constant");
  }
  if (failed(maxAttrResult) && checkNotNone(rewriter, op, op.getMax())) {
    return rewriter.notifyMatchFailure(op, "max attr should be a torch constant");
  }
```

ramiro050 (changes requested):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
zezhang: This looks a lot cleaner! Thanks for the suggestion!

ramiro050 (approved):
Thanks!

PR #2747, opened 2024-01-12
[ONNX][MLIR] support constantOfShape op

PhaneeshB:


lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1472,4 +1472,99 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        auto resultDType = resultType.getDtype();
+        if (!attr) {
+          if (!resultDType.isa<FloatType>()) {
+            return rewriter.notifyMatchFailure(
+                binder.op,
+                "`value` attr and result type (FloatType) mismatch.");
+          }
+          value_const = rewriter.create<Torch::ConstantFloatOp>(
+              binder.getLoc(), rewriter.getF64FloatAttr(0.0));
+        } else if (attr.isa<mlir::DenseElementsAttr>()) {
```
rsuderman, line 1524: We want to avoid all of these `if {} else if {} else if {} else{}` chains. Often we can know how the changes sequence and avoiding all the nesting / overlappiing.

For instance, your `if (!attr) { ...}` case can be seen as setting the default, then each other case (e.g. `if(attr.isa<DenseElementsAttr>()) {...}` can handle the individual cases. You can also do an upfront `if(!isa<Type1, Type2, Type3>(attr)( {return failure();}` to upfront validate the supported types.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1472,4 +1472,99 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          } else if (resultDType.isa<mlir::IntegerType>()) {
+            if (!isa<IntegerType>(denseAttrEleType)) {
+              return rewriter.notifyMatchFailure(
+                  binder.op,
+                  "`value` attr and result type (IntegerType) mismatch.");
+            }
+            int64_t intValue = denseAttr.getSplatValue<IntegerAttr>().getSInt();
+            value_const = rewriter.create<Torch::ConstantIntOp>(
+                binder.getLoc(), rewriter.getI64IntegerAttr(intValue));
+          } else {
```
rsuderman, line 1547: If you use the approach above you can handle the `value_const = ...` cases, then just check that `value_const` was set to something without using an `else`

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1472,4 +1472,99 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          }
+        } else if (attr.isa<mlir::FloatAttr>()) {
+          float value = attr.dyn_cast<FloatAttr>().getValue().convertToFloat();
+          value_const = rewriter.create<Torch::ConstantFloatOp>(
+              binder.getLoc(), rewriter.getF64FloatAttr(value));
+        } else if (attr.isa<mlir::IntegerAttr>()) {
+          int64_t value = attr.dyn_cast<IntegerAttr>().getSInt();
+          value_const = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(value));
+        } else {
```
rsuderman, line 1560: Just another observation. By doing the following up front
```
if (!isa<DenseElementsAttr, FloatAttr, IntegerAttr>) {
              binder.op,
              "`value` attr not float / int / tensor. not supported yet.");
}
```

We can validate the inputs / support upfront, clearly indicating what is supported.

rsuderman (changes requested):

PhaneeshB:
Hi @rsuderman  @vivekkhandelwal1 
this PR is ready for review please take a look when y'all can, it's needed for the initial 3 models going through ONNX
Thanks

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
PhaneeshB: I agree those if else chains were ugly, thanks for this review, pushed to try and get rid of them.
Most of the nesting was to ensure that any mismatch in resultDType and value attr dtype shold fail right away but I realise I might have been overly cautious there. Proceeding with an assumption for now and will fix if that break later on. 
Kept only `DenseElementsAttr ` as acceptable type since ONNX specifies value attr to be a TENSOR type.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
PhaneeshB: Got rid of the else part and decoupled element dtype discovery from constant op creation step.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
PhaneeshB: Updated 🙏

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1472,4 +1472,106 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          attr = DenseElementsAttr::get(
+              resultType.toBuiltinTensor().clone(resultDType),
+              rewriter.getFloatAttr(resultDType, 0.0));
+        }
+        if (!isa<DenseElementsAttr>(attr)) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "`value` attr needs to be a tensor.");
+        }
+
+        auto denseAttr = attr.dyn_cast_or_null<DenseElementsAttr>();
```
rsuderman, line 1537: Nit: you might as well use `attr.cast<DenseElementsAttr>()` as you already validated the type above.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1472,4 +1472,106 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        if (denseAttrEleType.isa<FloatType>()) {
+          fvp.dType = FillValueParam::FLOAT;
+          fvp.floatVal =
+              denseAttr.getSplatValue<FloatAttr>().getValue().convertToFloat();
+        }
+
+        // Create the constant Op for value
+        switch (fvp.dType) {
+        case FillValueParam::INT:
+          value_const = rewriter.create<Torch::ConstantIntOp>(
```
rsuderman, line 1560: Why not move these instantiations into the `if` statements above? We could forgo the entire `switch` and `struct` types you defined.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1472,4 +1472,106 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+              binder.getLoc(), rewriter.getI64IntegerAttr(fvp.intVal));
+          break;
+
+        case FillValueParam::FLOAT:
+          value_const = rewriter.create<Torch::ConstantFloatOp>(
```
rsuderman, line 1565: Ditto

rsuderman (changes requested):
Just some minor changes and it should be good.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
PhaneeshB: Done

rsuderman (approved):

Issue #2164, opened 2023-05-25
Getting illegal expression error in when using torch.arange


Abhishek-TyRnT:
Tried to use torch.arange operator, but it was giving illegal expression error.

Following is the code to reproduce the error
```
from torch.nn import Module
import torch
import torch_mlir

class ArangeDtypeIntModule(torch.nn.Module):
    def __init__(self):
        super().__init__()

    def forward(self):
        return torch.arange(0.2, 5.0, dtype=torch.int64)


model = ArangeDtypeIntModule()
example_args = []
ans = model()
print(ans)
x = torch_mlir.compile(model, example_args,torch_mlir.OutputType.TOSA)

with open("output.mlir", 'w') as f:
    f.write(str(x))
```

Here is the error IR Dump

```
models/issue.py:10:15: error: failed to legalize operation 'torch.aten.arange.start_step' that was explicitly marked illegal
        return torch.arange(0.2, 5.0, dtype=torch.int64)
              ^
models/issue.py:10:15: note: see current operation: %5 = "torch.aten.arange.start_step"(%3, %4, %0, %2, %1, %1, %1) : (!torch.float, !torch.float, !torch.int, !torch.int, !torch.none, !torch.none, !torch.none) -> !torch.vtensor<[5],si64>
// -----// IR Dump After ConvertTorchToTosa Failed (convert-torch-to-tosa) //----- //
func.func @forward() -> !torch.vtensor<[5],si64> {
  %int1 = torch.constant.int 1
  %none = torch.constant.none
  %int4 = torch.constant.int 4
  %float2.000000e-01 = torch.constant.float 2.000000e-01
  %float5.000000e00 = torch.constant.float 5.000000e+00
  %0 = torch.aten.arange.start_step %float2.000000e-01, %float5.000000e00, %int1, %int4, %none, %none, %none : !torch.float, !torch.float, !torch.int, !torch.int, !torch.none, !torch.none, !torch.none -> !torch.vtensor<[5],si64>
  return %0 : !torch.vtensor<[5],si64>
}
```

When I looked at ConvertTorchToTosa Pass. I noticed that arangeOp was not added
https://github.com/llvm/torch-mlir/blob/f0b7b63be06fc0ed5bdf43e02967cd47a53a24fc/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L4541

Is this because it was not implemented or something else is the issue?

ramiro050:
[...]

PR #2749, opened 2024-01-13
Added support for torch arange float module

Abhishek-TyRnT:
Added Support for float dtype in in torch.arange in TOSA Dialect

This resolves the following issue :- 
https://github.com/llvm/torch-mlir/issues/2164

The following test cases are passing after this change

1. ArangeDtypeIntModule_basic
2. ArangeFloatModule_basic
3. ArangeNegativeStartFloatModule_basic
4. ArangeStartFloatModule_basic
5. ArangeStartNegativeStepFloatModule_basic
6. ArangeStartOutDtypeModule_basic
7. ArangeStartStepFloatModule_basic


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -26,6 +26,7 @@
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionDialect.h"
 #include "torch-mlir/Dialect/TorchConversion/Transforms/BackendTypeConversion.h"
+#include <iostream>
```
dan-garvey, line 29: is this required?

Abhishek-TyRnT: That was added for debugging. It was accidently left there. I will get rid of it

eric-k256 (approved):
Looks good, thanks!

Abhishek-TyRnT:
CI doesn't seem to pick up jobs... Is there any reason?

newling (commented):
This isn't going to work for large 64-bit integers, because casting to double will be lossy (mantissa of double is 52 bits, so 2^53 + 1 can't be represented as a double).

Abhishek-TyRnT:
You are right. I will try to find a way to fix this.

Abhishek-TyRnT:
I guess there will be no loss if we use long double instead of double. What do you think?

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -4067,28 +4067,60 @@ LogicalResult ConvertAtenOp<AtenArangeStartStepOp>::matchAndRewrite(
...
-  SmallVector<int64_t> values(resultShape, start);
-  for (unsigned i = 1; i < resultShape; i++)
-    values[i] += i * step;
-  Value result =
-      tosa::getConstTensor<int64_t>(rewriter, op, values, resultShape).value();
+  int64_t resultShape = ceil((end - start) / step);
+  Value result;
+  if (is_all_inp_int)
+  {
+    SmallVector<int64_t> values(resultShape, start);
```
newling, line 4109: @Abhishek-TyRnT  

For the all integer case, you can calculate the result shape for the all-int case here, using only the integer values (start_int, step_int, end_int) which should be available. 

 I think that's better than using long double, which is not guaranteed to be strictly larger than double by C++ (AFAIK).

Abhishek-TyRnT: I made the change please check.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -4067,28 +4067,62 @@ LogicalResult ConvertAtenOp<AtenArangeStartStepOp>::matchAndRewrite(
...
-  SmallVector<int64_t> values(resultShape, start);
-  for (unsigned i = 1; i < resultShape; i++)
-    values[i] += i * step;
-  Value result =
-      tosa::getConstTensor<int64_t>(rewriter, op, values, resultShape).value();
+  
+  Value result;
+  if (is_all_inp_int)
+  {
+    int64_t resultShape = ceil((float)(end_int - start_int) / (float)(step_int));
```
newling, line 4109: There is some casting precision loss here still. I think the resultShape can be determined without casting. 

I think this variable should be called resultSize, not resultShape. 


You could do away with it altogether, just have 

SmallVector<int64_t> values;
for (int64_t i = start_int; i < end_int; i+= step_int){
  values.push_back(i);
}

newling (commented):
Looks fine, just the final comment for the int case.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
Abhishek-TyRnT: Done

newling (commented):
The logic looks good to me. I think the code can be polished a little bit though :-)

Abhishek-TyRnT:
Hi @newling. The int64 test suddenly started failing cause mlir couldn't typecast float32 or float64 to int64. So I manually typecasted them Torch-to-Tosa pass.

Abhishek-TyRnT:
The CI failure doesn't seem it to be a issue with TorchToTosa but something in stablehlo unexpectedly passing.

newling:
Ok. You might need to update the xfail set? https://github.com/llvm/torch-mlir/pull/2927
Have you rebased onto main?

Abhishek-TyRnT:
I just checked, that tested has been added there. It should hopefully pass now

Abhishek-TyRnT:
Now that it has passed. Can you please merge it. @newling

newling:
Hi @Abhishek-TyRnT I've made a PR against this which should fix the link checks. It also makes some stylistic improvements (IMO) which you can consider pulling in if you like: https://github.com/Abhishek-TyRnT/torch-mlir/pull/1

PR #2751, opened 2024-01-13
[onnx] Support lowering quantize linear to `torch`

rsuderman:
We can map the per_tensor case to the `torch.aten.quantize_per_linear`
operation. In this case we extract the `scale` and `zeropoint` values
and directly invoke the quantization, then return the integer
representation value.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -44,7 +44,9 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
 
                   auto scaleTy = scale.getType().dyn_cast<Torch::ValueTensorType>();
                   if (!scaleTy || !scaleTy.hasSizes()) return rewriter.notifyMatchFailure(binder.op, "requires known rank");
-                  if (!resultType.hasDtype()) return rewriter.notifyMatchFailure(binder.op, "requires known resulty dtype");
+                  if (!resultType.hasDtype())
+                    return rewriter.notifyMatchFailure(
+                        binder.op, "requires known result dtype");
 
                   if (scaleTy.getSizes().size() == 0) {
                     Type qTy = resultType.getDtype();
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -8,6 +8,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -29,6 +30,54 @@ using namespace mlir::torch::onnx_c;
...
+                      binder.tensorResultType(resultType))
+                    return failure();
+
+                  Value operand = operands[0];
+                  Value scale = operands[1];
+                  Value zeropoint = operands[2];
+
+                  auto scaleTy = scale.getType().dyn_cast<Torch::ValueTensorType>();
+                  if (!scaleTy || !scaleTy.hasSizes()) return rewriter.notifyMatchFailure(binder.op, "requires known rank");
+                  if (!resultType.hasDtype()) return rewriter.notifyMatchFailure(binder.op, "requires known resulty dtype");
```
dan-garvey, line 47: typo

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -29,6 +30,54 @@ using namespace mlir::torch::onnx_c;
...
+
+                    if (qTy.isUnsignedInteger(8)) {
+                      qTy = rewriter.getType<Torch::QUInt8Type>();
+                    } else if (qTy.isSignedInteger(8)) {
+                      qTy = rewriter.getType<Torch::QInt8Type>();
+                    } else if (qTy.isSignedInteger(32)) {
+                      qTy = rewriter.getType<Torch::QInt32Type>();
+                    } else {
+                      return rewriter.notifyMatchFailure(binder.op, "unsupported result dtype");
+                    }
```
dan-garvey, line 60: are these the types we most commonly see or is there some other limitation?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
dan-garvey: I think = P

dan-garvey (approved):
ping me for another +1 after the nit

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
rsuderman: There are the only supported quantized types. While there are 4-bit versions, they only occur as pure integer types which lack the sidechannel scale / zero-point information.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
rsuderman: Fixed.

PR #2752, opened 2024-01-13
Onnx range

kumardeepakamd:
Implemented ONNX.Range. The spec says the data type for start, limit, delta are 0-D can be double, float, int16, int32, int64, All int types mapped to !torch.int and all float types mapped to !torch.float

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -33,9 +33,6 @@ using namespace mlir::torch::onnx_c;
 namespace {
 template <typename T>
 Value getItemOp(OpBinder binder, ConversionPatternRewriter &rewriter,
-                Value &ofItem);
-template <typename T>
-Value getItemOp(OpBinder binder, ConversionPatternRewriter &rewriter,
                 Value &ofItem) {
   return rewriter.create<Torch::AtenItemOp>(binder.getLoc(),
                                             rewriter.getType<T>(), ofItem);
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1336,4 +1351,50 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            binder.tensorOperandAtIndex(delta, 2) ||
+            binder.tensorResultType(resultType))
+          return failure();
+
+        // Convert a 0-dimensional/Scalar Tensor ([]) to Scalar Torch Numeric
+        // Value torch.tensor(1.1) equivalent in ONNX to 1.1 as an example
+        // type of start, limit, delta can be one of: double, float, int16,
+        // int32, int64 Assuming start, limit and delta to be same type (could
+        // they be different?)
+        Torch::BaseTensorType startTesorType =
```
vivekkhandelwal1, line 1373: Typo here: `startTesorType` -> `startTensorType`

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -27,6 +27,21 @@ using namespace mlir::torch::onnx_c;
...
+//  Templatized function to get an item op of a type
+namespace {
+template <typename T>
+Value getItemOp(OpBinder binder, ConversionPatternRewriter &rewriter,
+                Value &ofItem);
+template <typename T>
+Value getItemOp(OpBinder binder, ConversionPatternRewriter &rewriter,
+                Value &ofItem) {
+  return rewriter.create<Torch::AtenItemOp>(binder.getLoc(),
+                                            rewriter.getType<T>(), ofItem);
```
vivekkhandelwal1, line 41: I think there's no need to declare this utility here since you're also defining it here.

vivekkhandelwal1 (approved):
Except minor comments, everything else looks good.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
kumardeepakamd: Good eyes. :-) Fixed.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
kumardeepakamd: That is right, was thinking of having a common place for other CPP files to share initially. Declaration removed. I see that createTorchTransposeOp and onnxDtypeIntToTorchDtypeInt are static fucntions in /torch-mlir/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp but they could be useful for other Domain*cpp files too. Maybe we should have DefaultDomainCommon.h and .cpp and put common utility functions there?

vivekkhandelwal1: Yeah, that's a good idea. We can do that in a follow-up PR.

vivekkhandelwal1 (approved):

PR #2758, opened 2024-01-15
Add support for `si8` and `ui8` to onnx importer

rsuderman:
si8 and ui8 types are relatively common to encounte in ONNX model
imports of quantized models. Added these datatypes to our `onnx`
importer.


dan-garvey (approved):

dan-garvey:
Do you see any value in adding some kind of test that these types get properly converted to DenseElementAttrs?

rsuderman:
> Do you see any value in adding some kind of test that these types get properly converted to DenseElementAttrs?

It would be a good idea to, I mainly encountered this via a full end-to-end model so I am not aware of any similar existing unittests.

stellaraccident (approved):

PR #2760, opened 2024-01-15
[onnx] Added flatten

daveliddell:
[https://github.com/nod-ai/SHARK-Turbine/issues/328](url)


lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1311,6 +1311,83 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        auto baseType = Torch::ValueTensorType::getWithLeastStaticInformation(
+            binder.op->getContext());
+        if (axis >= rank) {
+          Value rankConst = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(rank));
+          collapsedRight = rewriter.create<Torch::AtenUnsqueezeOp>(
+              binder.getLoc(), baseType, operand, rankConst);
+        }
+
+        // Otherwise, collapse the right range into a single dimension:
```
rsuderman, line 1357: Move this inside of the else state so that the else statement looks like 
```
} else {
```

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1311,6 +1311,83 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        }
+
+        // If the left range is empty, add a dim of size 1 to the
+        // left side of the shape:
+        // result = torch.unsqueeze(cr, 0)
+        Value result;
+        Value zero = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getI64IntegerAttr(0));
+        if (axis <= 0)
+          result = rewriter.create<Torch::AtenUnsqueezeOp>(
```
rsuderman, line 1375: You might as well make this 
```
if (axis <= 0) {
	rewriter.replaceOpWithNewOp<Torch::AtenUnsqueezeOp>(
        	binder.op, resultType, collapsedRight, zero);
        return success();
}
```

Then you can just follow up with the other case.

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
daveliddell: I thought making the right and left symmetric would be easier to understand, as they're basically the same thing, but I can change it to this way, too.  If I do that, I probably should replace the rewriter.create with rewriter.replaceOpWithNewOp there, too, shouldn't I?

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
daveliddell: Gotcha.  Not very familiar with the conventions yet

daveliddell:
Thanks for your review, @rsuderman!  I've made the changes you've requested.

test/Conversion/TorchOnnxToTorch/simple_ops_a_to_f.mlir
```diff
@@ -1020,3 +1020,116 @@ func.func @ints_constant() -> !torch.vtensor<[2], si64> attributes {torch.onnx_m
   return %0 : !torch.vtensor<[2],si64>
 }
 
+// CHECK-LABEL: @test_flatten_4d_axis_2
+func.func @test_flatten_4d_axis_2(%arg0: !torch.vtensor<[2,3,4,5],f32>) -> !torch.vtensor<[6,20],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 13 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
+  // CHECK: %[[RIGHT_START:.*]] = torch.constant.int 2
```
newling, line 1025: A small nit. These tests will fail if the IR gets printed as 

torch.constant.int 3
torch.constant.int 2

instead of 

torch.constant.int 2
torch.constant.int 3

This could in theory happen in something in MLIR upstream changes or someone changes the onnx->torch-mlir code you've implemented here for whatever reason. 

If it doesn't matter in what order 2 lines appear, you can use CHECK-DAG. So  

// CHECK-DAG: torch.constant.int 3
// CHECK-DAG: torch.constant.int 2

will match both of the above.

newling (commented):
Looks great to me.

test/Conversion/TorchOnnxToTorch/simple_ops_a_to_f.mlir
daveliddell: @newling Thanks for the review, James!  Yes, I'm aware of the potential for reordering, as I've encountered it in the past. I agree that if the -convert-torch-onnx-to-torch pass gets more complicated and isn't just the rewrites here, then the ordering could be changed.  However, all the other torch-onnx tests are relying on order (that is, I don't see any CHECK-DAGs anywhere), so the first person to try such a thing will have fun cleaning up all the tests!  :-D  I take the reliance on order preservation to mean that we intend not to add fancy order-changing optimizations to this pass.

rsuderman: Better to improve practice now and fix up other tests later, so I would change to `CHECK-DAG`. MLIR does not guarantee to maintain operation order between passes so independent of our own changes it is possible that the operation ordering could change.

rsuderman (approved):

test/Conversion/TorchOnnxToTorch/simple_ops_a_to_f.mlir
daveliddell: @newling @rsuderman  Now using CHECK-DAG for all lines that could potentially appear in any order: not just the constants but the first (RHS) collapse intermediate results as well.  Lit tests (still) pass.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
daveliddell: Made this change as suggested, plus also changed the final create to replaceOpWithNewOp.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
daveliddell: I have made this change

AmosLewis:
The flatten op need to be relook.
```
DarkNet53.default.onnx.torch.mlir:238:12: error: failed to legalize operation 'torch.operator' that was explicitly marked illegal
    %234 = torch.operator "onnx.Flatten"(%233) {torch.onnx.axis = 1 : si64} : (!torch.vtensor<[1,1024,1,1],f32>) -> !torch.vtensor<[1,1024],f32> 
           ^
DarkNet53.default.onnx.torch.mlir:238:12: note: see current operation: %1229 = "torch.operator"(%1228) <{name = "onnx.Flatten"}> {torch.onnx.axis = 1 : si64} : (!torch.vtensor<[1,1024,1,1],f32>) -> !torch.vtensor<[1,1024],f32>
```

PR #2761, opened 2024-01-16
[onnx] `onnx.MatMulInteger` lowering to `torch.mm` and `quint*` types

rsuderman:
Torch does not have an equivalent matmul operation for integers. Instead
it sidechannels the information via its quantized types. For this
lowering we setup these sidechannels then invoke `torch.mm`.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -136,6 +136,79 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
                       binder.op, resultType, lhs, rhs);
                   return success();
                 });
+  patterns.onOp(
+      "MatMulInteger", 10,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        llvm::SmallVector<Value> operands;
+        if (binder.tensorOperands(operands, 4) ||
+            binder.tensorResultType(resultType))
+          return failure();
+
+        Value lhs = operands[0];
+        Value rhs = operands[1];
+        Value lhsZp = operands[2];
+        Value rhsZp = operands[3];
+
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -137,6 +137,79 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        Torch::ValueTensorType resultType;
+        llvm::SmallVector<Value> operands;
+        if (binder.tensorOperands(operands, 4) ||
+            binder.tensorResultType(resultType))
+          return failure();
+
+        Value lhs = operands[0];
+        Value rhs = operands[1];
+        Value lhsZp = operands[2];
+        Value rhsZp = operands[3];
```
vivekkhandelwal1, line 152: The `a_zero_point` and `b_zero_point` are optional args, they might not be present always.

vivekkhandelwal1: And, indexing this would result in a crash.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
rsuderman: Changed to conditionally pull the values (see above). Do we know if the optional representation is via a `None` value or absect operands? I had run on the assumption that `onnx` used none types for missing parameters.

vivekkhandelwal1: No, the way onnx operators are imported the optional value is either present or absent.

renxida (approved):
LGTM. Learning tons about how to write code in style 😎

vivekkhandelwal1 (approved):

PR #2769, opened 2024-01-18
[torch] `torch.dequantize` for per channel tensors to` linalg`

rsuderman:
Support a lowering for dequantization for per channel tensors from
`torch` dialect to a linalg decomposition. Tested via a numerical
`torch` test.


renxida (approved):
"ElementwiseDequantizePerChannelModule_basic",
needs to be inserted at line 1392 of projects/pt1/e2e_testing/xfail_sets.py
to make CI pass

otherwise LGTM!

PR #2773, opened 2024-01-19
Implement lowering of torch.aten.lerp.Scalar

ikalinic:
Closes nod-ai/SHARK-Turbine#356


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -986,6 +986,24 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     auto weightedDelta = b.create<arith::MulFOp>(loc, delta, weight);
     return b.create<arith::AddFOp>(loc, start, weightedDelta);
   }
+  if (auto lerpScalar = dyn_cast<AtenLerpScalarOp>(op)) {
```
ramiro050, line 989: This can be done as a decomposition. This way all three torch-mlir backend (tosa, stablehlo, linalg) benefit from this addition.

ramiro050 (changes requested):

ikalinic:
Hi @ramiro050, thanks for the review! I've decomposed the AtenLerpScalar op inside DecomposeComplexOps.cpp, so the changes are ready for a review.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1895,6 +1896,37 @@ class DecomposeAtenLeakyReluBackwardOp
...
+    if (!resType.hasDtype()) {
+      return rewriter.notifyMatchFailure(op, "result should have dtype");
+    }
+    if (!op.getType()
+             .cast<ValueTensorType>()
+             .getDtype()
+             .isa<mlir::FloatType>()) {
+      op.emitError("unimplemented: non-floating point dtype");
+      return failure();
+    }
```
vivekkhandelwal1, line 1916: You don't need to constraint this to the floating point.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1895,6 +1896,37 @@ class DecomposeAtenLeakyReluBackwardOp
...
+    auto dtype = resType.getDtype();
+    auto start = convertScalarToDtype(rewriter, loc, op.getSelf(), dtype);
+    auto end = convertScalarToDtype(rewriter, loc, op.getEnd(), dtype);
+    auto weight = convertScalarToDtype(rewriter, loc, op.getWeight(), dtype);
```
vivekkhandelwal1, line 1920: This is not required, also `Self` and `End` are tensors, not scalars.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1895,6 +1896,37 @@ class DecomposeAtenLeakyReluBackwardOp
...
+    auto delta = rewriter.create<AtenSubFloatOp>(loc, end, start);
+    auto weightedDelta = rewriter.create<AtenMulFloatOp>(loc, delta, weight);
+    auto lerp = rewriter.create<AtenAddFloatIntOp>(loc, start, weightedDelta);
```
vivekkhandelwal1, line 1923: Just use ops: `AtenSubTensorOp`, `AtenMulScalarOp`, `AtenAddTensorOp`.

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -545,6 +545,30 @@ def forward(self, x):
 def ElementwiseLeakyReluStaticModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(4, 5, 6, low=-1))
 
+
+# ==============================================================================
+
+
+class ElementwiseLerpScalarFloatModule(torch.nn.Module):
```
vivekkhandelwal1, line 552: Please add a test case for integer type as well.

vivekkhandelwal1 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ikalinic: @vivekkhandelwal1 Should I refactor this to be a check if complex type is result type or should I remove this check completely?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1895,6 +1896,37 @@ class DecomposeAtenLeakyReluBackwardOp
...
 } // namespace
 
+namespace {
+class DecomposeAtenLerpScalarOp : public OpRewritePattern<AtenLerpScalarOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenLerpScalarOp op,
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    auto resType = op.getType().cast<ValueTensorType>();
```
ramiro050, line 1906: nit: cast to `BaseTensorType`. Types at this stage are not guaranteed to be `ValueTensorType`s.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ramiro050: I don't see any constraints on the result type in the [documentation for this op](https://pytorch.org/docs/stable/generated/torch.lerp.html), so no need to have this check.

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -12,6 +12,7 @@
 #include "mlir/IR/BuiltinDialect.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
+#include "torch-mlir/Conversion/Utils/Utils.h"
```
ramiro050, line 15: nit: I don't think this is being used?

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -545,6 +545,48 @@ def forward(self, x):
...
+class ElementwiseLerpScalarIntModule(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.float32, True),
+        ([-1, -1], torch.float32, True),
```
ramiro050, line 561: To test `int` types, these should be `torch.int64`s or `torch.int32`s

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -545,6 +545,48 @@ def forward(self, x):
...
+    ])
+    def forward(self, a, b):
+        return torch.ops.aten.lerp(a, b, weight=2)
+
+@register_test_case(module_factory=lambda: ElementwiseLerpScalarIntModule())
+def ElementwiseLerpScalarIntModule_basic(module, tu: TestUtils):
+    module.forward(tu.rand(5,3), tu.rand(5,3))
```
ramiro050, line 568: you can generate random int tensors with `tu.randint(5, 3, low=1, high=10)`

ramiro050 (changes requested):

ikalinic:
Hi @ramiro050, thanks again for the review! As for the `ElementwiseLerpScalarIntModule` test which should test `int` types, the only input that is supported as int is the weight parameter (which is a scalar), so I am testing the weight as an int in this test case. By passing Tensors with any int dtype, all of the configurations fail with the following message:
```
module.forward(tu.randint(5, 3, low=0, high=10, dtype=torch.int32), tu.randint(5, 3, low=0, high=10, dtype=torch.int32))
          File "/home/ikalinic/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 261, in __call__
            output = self.__wrapped__(*args, **kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "/home/ikalinic/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/elementwise.py", line 564, in forward
            return torch.ops.aten.lerp(a, b, weight=2)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "/home/ikalinic/anaconda3/envs/torch-mlir/lib/python3.11/site-packages/torch/_ops.py", line 825, in __call__
            return self_._op(*args, **(kwargs or {}))
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        RuntimeError: "lerp_kernel_scalar" not implemented for 'Int'
```
Should I still pass int tensors with `tu.randint` and add all the test case to all the xfail sets? Or does this happen because I need to handle int Tensors (cast them to float dtype of same size) inside my actual lowering logic?
Thanks

ramiro050:
> the only input that is supported as int is the weight parameter

Whoops, my bad. I did not notice that the scalar type was the one being tested. The tests are fine as is, then.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ikalinic: This isn't being used by my patch, but I think that clang-format moved this (the build fails without this import)

ramiro050 (approved):
LGTM!

PR #2777, opened 2024-01-20
[NFC] Cleanup of `torch-to-linalg` `pooling` code

rsuderman:
The pooling implementation template behavior can encounter template
specialization failures on some platforms due to nested templates.
Removed the template nesting to support a wider range of compilers.


stellaraccident (approved):

rsuderman:
Someone else landed a fix

PR #2778, opened 2024-01-20
g++ build fix

newling:
Introduced in 704cfdaf0893fa4078ab744a915a56367282dcaa of @wu-s-john 

g++ compiler error: 

Pooling.cpp:177:13: error: explicit specialization in non-namespace scope ‘class 

Design looks good, g++ is just freaking out for no good reason. Un-nesting the template classes fixes the error. 

We don't have g++ CI. This hopefully happens infrequently enough that we can just fix manually. My service to those folks who really like building with g++... :)


stellaraccident:
Thanks!

stellaraccident (approved):

PR #2779, opened 2024-01-20
AtenAdaptiveMaxPool2d Conversion to Linalg

zjgarvey:
The logic here is very similar to the conversion for AdaptiveAvgPool1d #2661 with a few modifications:

1. buffVal = -inf instead of 0
2. the main linalg generic op accumulates a max, instead of a sum, to the first output tensor
3. avg pooling requires dividing the sum pool by the kernel width, which we stored as an auxilliary tensor (kSizeTensor). Here, the auxiliary tensor will be recording the indices. Strangely enough, the only signature available for this function is to return indices, and it appears that they must be computed whether the user desires them or not. See [pytorch/torch/nn/functional.py](https://github.com/pytorch/pytorch/blob/main/torch/nn/functional.py#L1174).

Before writing other adaptive pooling conversions, the logic of this decomposition should be rolled into a helper function that will work for both max and avg pooling ops. Even the auxiliary tensor should likely be automated. This code was written in a slightly more tedious way than strictly necessary (often using loops to fill SmallVectors up to rank-2, which is only two in this case), in order to more easily facilitate the transition to a helper function.


include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
```diff
@@ -6743,6 +6743,31 @@ def Torch_Aten_AdaptiveAvgPool3dBackwardOp : Torch_Op<"aten._adaptive_avg_pool3d
   }];
 }
 
+def Torch_AtenAdaptiveMaxPool2dOp : Torch_Op<"aten.adaptive_max_pool2d", [
```
newling, line 6746: Too bad we don't have the flexibility with the automatic generation to make the adaptive pooling ops 'inherit' from the same class.

newling (commented):
Why not implement the helper functions first, and then add this op? 

Just a general software concern... there's a good chance this refactoring won't be done, and then someone will modify the average pool op to make it more efficient etc. without seeing the note here, resulting in divergence.

include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
zjgarvey: Yeah, I can do that if it makes more sense. Someone wanted this specific op lowered in [SHARK-Turbine#116](https://github.com/nod-ai/SHARK-Turbine/issues/116). 

I'm planning to refactor in the next patch I put up for this, but figured I would do this as a test run for the max ops and get support for this op ASAP while working on the rest.

zjgarvey:
> Why not implement the helper functions first, and then add this op?
> 
> Just a general software concern... there's a good chance this refactoring won't be done, and then someone will modify the average pool op to make it more efficient etc. without seeing the note here, resulting in divergence.

Yeah, I can do that if it makes more sense. Someone wanted this specific op lowered in https://github.com/nod-ai/SHARK-Turbine/issues/116.

I'm planning to refactor in the next patch I put up for this, but figured I would do this as a test run for the max ops and get support for this op ASAP while working on the rest.

newling:
> > Why not implement the helper functions first, and then add this op?
> > Just a general software concern... there's a good chance this refactoring won't be done, and then someone will modify the average pool op to make it more efficient etc. without seeing the note here, resulting in divergence.
> 
> Yeah, I can do that if it makes more sense. Someone wanted this specific op lowered in [nod-ai/SHARK-Turbine#116](https://github.com/nod-ai/SHARK-Turbine/issues/116).
> 
> I'm planning to refactor in the next patch I put up for this, but figured I would do this as a test run for the max ops and get support for this op ASAP while working on the rest.

Makes sense, thanks for the context.

newling (approved):
I'm ok with this landing, as there is a plan to follow it up with the factorization with the similar op(s). 

@zjgarvey let me know if you'd like me squash and merge this.

zjgarvey:
> I'm ok with this landing, as there is a plan to follow it up with the factorization with the similar op(s).
> 
> @zjgarvey let me know if you'd like me squash and merge this.

Yeah, please do. I've already almost finished the refactor + support for the remaining ops. I'll post that PR as soon as it's done.

Issue #2764, opened 2024-01-16
Importing ONNX model for facebook/opt-125M into torch-mlir fails


kumardeepakamd:
Issue seen: Error: onnx_importer.py", line 350, in _get_immediate_tensor

Steps to reproduce the issue:
a)Save following code as model.py:

```
from transformers import OPTModel, AutoTokenizer
import torch

class optModel(torch.nn.Module):
    def __init__(self):
        super().__init__()
        self.model = OPTModel.from_pretrained(
            "facebook/opt-125M",
            num_labels=2,
            output_attentions=False,
            output_hidden_states=False,
            torchscript=True,
        )
        self.model.eval()

    def forward(self, tokens):
        return self.model.forward(tokens)[0]

tokenizer = AutoTokenizer.from_pretrained("facebook/opt-125M")
test_input = torch.tensor([tokenizer.encode("The Manhattan bridge")])

model = optModel()
onnx_program = torch.onnx.export(model, test_input, "opt-125M.onnx")
```

b) run: 'python ./model.py' to get opt-125M.onnx 9'   (This may take a minute)
c) run: 'python -m torch_mlir.tools.import_onnx opt-125M.onnx -o opt-125M.torch-onnx.mlir

You will see an error:
  _torch_pytree._register_pytree_node(
Traceback (most recent call last):
  File "/proj/gdba/kumar/anaconda3/envs/vaiml/lib/python3.10/site-packages/torch_mlir/extras/onnx_importer.py", line 350, in _get_immediate_tensor
    initializer = self._gi.initializer_map[name]
KeyError: '/decoder/Constant_3_output_0'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/proj/gdba/kumar/anaconda3/envs/vaiml/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/proj/gdba/kumar/anaconda3/envs/vaiml/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/proj/gdba/kumar/anaconda3/envs/vaiml/lib/python3.10/site-packages/torch_mlir/tools/import_onnx/__main__.py", line 77, in <module>
    _cli_main()
  File "/proj/gdba/kumar/anaconda3/envs/vaiml/lib/python3.10/site-packages/torch_mlir/tools/import_onnx/__main__.py", line 73, in _cli_main
[...]

PR #2780, opened 2024-01-21
Fix onnx importer to treat Constant values as static

daveliddell:
Fixes  https://github.com/llvm/torch-mlir/issues/2764

In the case of OPT, there are ConstantOfShape ops whose input shape is not static (that is, an initializer), but rather comes from a Constant op.  The importer can't handle such non-static input shapes.

The fix here is to create initializers for a subset of Constant ops (ones with "value" attributes), so that their outputs can be used statically.  Additionally, there was no case for creating a splat of int64, so I added that as well.


python/torch_mlir/extras/onnx_importer.py
```diff
@@ -276,10 +276,13 @@ def import_node(self, node: onnx.NodeProto):
         with InsertionPoint(self._b), Location.name(node.name):
             op_type = node.op_type
             # Handle special op types that materialize to non-op IR constructs.
+            # Handlers return True if the op was handled, else this function
+            # should process it as a general node.
             special_key = f"_handle_node_{op_type}"
             if hasattr(self, special_key):
-                getattr(self, special_key)(node)
-                return
+                wasHandled = getattr(self, special_key)(node)
```
stellaraccident, line 283: Nit: Use snake_case

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -605,9 +633,10 @@ def tensor_proto_to_attr(self, tp: onnx.TensorProto) -> Attribute:
 }
 
 
-def _get_attr(node: onnx.NodeProto, attr_name: str) -> onnx.AttributeProto:
+def _get_attr(node: onnx.NodeProto, attr_name: str, isRequired: bool = True) -> onnx.AttributeProto:
     for attr in node.attribute:
         if attr.name == attr_name:
             return attr
-    else:
+    if isRequired:
```
stellaraccident, line 640: Nit: snake_case

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -605,9 +633,10 @@ def tensor_proto_to_attr(self, tp: onnx.TensorProto) -> Attribute:
...
         raise OnnxImportError(f"Required attribute {attr_name} not found in {node}")
+    return None
```
stellaraccident, line 642: Nit: newline at end

stellaraccident (approved):
A few small nits then good to go. Thanks!

stellaraccident:
Interesting that this kind of thing doesn't seem to be in the onnx test suite... (would have caused an xfail to start passing)

python/torch_mlir/extras/onnx_importer.py
daveliddell: D'oh!  Why isn't my editor doing my thinking for me?!  Fixed

python/torch_mlir/extras/onnx_importer.py
daveliddell: Fixed.  I thought we didn't like snakes here

PR #2781, opened 2024-01-21
implement support for lowering exp from onnx to torch

kritgpt:
implements nod-ai/SHARK-Turbine#312.


AmosLewis:
https://github.com/llvm/torch-mlir/pull/2792

PR #2782, opened 2024-01-22
[ONNX][MLIR] Fix padding size constraint for onnx.maxpool op

Shukla-Gaurav:


newling (commented):
Would it make sense to add a test for this here, in the torch-mlir repo?

Shukla-Gaurav:
> Would it make sense to add a test for this here, in the torch-mlir repo?

Sure, let me add a test here. Thanks!

newling (approved):
Thanks @Shukla-Gaurav!

PR #2789, opened 2024-01-23
[MLIR][TORCH] Support for `onnx.LayerNormalization`

vivekkhandelwal1:

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -653,12 +653,9 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
       "LayerNormalization", 17,
       [](OpBinder binder, ConversionPatternRewriter &rewriter) {
         Torch::ValueTensorType yType, meanType, invStdDevType;
-        Value x;
-        Value scale;
-        Value b;
-        int64_t axis;
+        Value x, scale, b;
+        int64_t axis, stashType;
         float epsilon;
-        int64_t stashType;
         if (binder.tensorOperandAtIndex(x, 0) ||
             binder.tensorOperandAtIndex(scale, 1) ||
             binder.tensorOperandAtIndex(b, 2) ||
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -649,49 +649,75 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         }
         return failure();
       });
-  patterns.onOp("LayerNormalization", 17,
-                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-                  Torch::ValueTensorType Y_type;
-                  Torch::ValueTensorType Mean_type;
-                  Torch::ValueTensorType InvStdDev_type;
-                  Value X;
-                  Value Scale;
-                  Value B;
-                  int64_t axis;
-		  float epsilon;
-                  int64_t stash_type;
-                  if (binder.tensorOperandAtIndex(X, 0) ||
-                      binder.tensorOperandAtIndex(Scale, 1) ||
-                      binder.tensorOperandAtIndex(B, 2) ||
-                      binder.tensorResultTypeAtIndex(Y_type, 0) || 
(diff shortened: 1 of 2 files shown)


afalkenberg1:
Hello yes that was an open issue since the mean and stddev are optional outputs ; 
nevertheless where do you get the information on how to derive the missing types ? 

auto reducedType = X_type.getWithSizesAndDtype(
              reducedShape, X_type.getOptionalDtype());

afalkenberg1 (commented):
Ok looks good ; there was another called AtenLayerNorm only which I considered before. It might be a better match if only one output is generated. Nevertheless it also has a cuda flag which needs to be supported. So overall we might be better of the native version in both cases.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -649,49 +649,73 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
-                    Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
-                    normalized);
-                  rewriter.replaceOpWithNewOp<Torch::AtenNativeLayerNormOp>(
-                      binder.op, Y_type, Mean_type, InvStdDev_type, X, normalized_shape, Scale, B, constEpsilon);
-                  return success();
-              });
+  patterns.onOp(
+      "LayerNormalization", 17,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType Y_type;
```
rsuderman, line 655: You can declare all of these on the same line:
```
Torch::ValueTensorType yType, meanType, invStdDevType;
```

Also, llvm style  uses camelCase instead of under_scores. So use the renamed vars above.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -649,49 +649,73 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        Torch::ValueTensorType Mean_type;
+        Torch::ValueTensorType InvStdDev_type;
+        Value X;
```
rsuderman, line 658: Ditto.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -649,49 +649,73 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            reducedShape[i] = X_shape[i];
+          auto reducedType = X_type.getWithSizesAndDtype(
+              reducedShape, X_type.getOptionalDtype());
+          Value y = rewriter.create<Torch::AtenNativeLayerNormOp>(
+              binder.getLoc(), Y_type, /*meanType=*/reducedType,
+              /*invStdDevType=*/reducedType, X, normalized_shape, Scale, B,
+              constEpsilon).getResult0();
+          rewriter.replaceOp(binder.op, y);
+          return success();
+        } else if (numResults == 3) {
```
rsuderman, line 707: The `if` condition above contains a `return` so we do not need to use the `else`

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
afalkenberg1: Yes I wondered about that, there might be violations of that in lots of other places as well.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -649,49 +649,75 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
-                    normalized);
-                  rewriter.replaceOpWithNewOp<Torch::AtenNativeLayerNormOp>(
-                      binder.op, Y_type, Mean_type, InvStdDev_type, X, normalized_shape, Scale, B, constEpsilon);
-                  return success();
-              });
+  patterns.onOp(
+      "LayerNormalization", 17,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType yType, meanType, invStdDevType;
+        Value x;
```
rsuderman, line 656: Same thing as the `Torch::ValueTensorType`, you can declare the `Value` vars all on the same row.

rsuderman (approved):
Just one last nit and the nit can be landed.

PR #2790, opened 2024-01-23
Implement trace

frafranz:
The lowering decomposes AtenTraceOp into an AtenDiagonalOp followed by AtenSumOp.

The progress is tracked in https://github.com/nod-ai/SHARK-Turbine/issues/333.


projects/pt1/python/torch_mlir_e2e_test/test_suite/reduction.py
```diff
@@ -1230,3 +1230,56 @@ def forward(self, input, target):
...
+
+    @export
+    @annotate_args([
+        None,
+        ([-1, -1], torch.int64, True),
+    ])
+    def forward(self, a):
+        return torch.ops.aten.trace(a)
+
+@register_test_case(module_factory=lambda: TraceIntModule())
```
newling, line 1274: I see your logic accounts for rank-2 non-square tensors. Perhaps there should test for this case.

newling (commented):
Looks good!

projects/pt1/python/torch_mlir_e2e_test/test_suite/reduction.py
frafranz: Hi James, thanks for your review! Exactly, the logic accounts for rank-2 tensors. A test case for non-square and float-valued tensors is already contained, I could add one for int but do not think the handling of non-square tensors will depend on the dtype.

newling: So sorry, I missed the float tests above! My bad.

newling (approved):

projects/pt1/python/torch_mlir_e2e_test/test_suite/reduction.py
frafranz: No worries! :)

frafranz:
Perhaps @qedawkins @vivekkhandelwal1 could have another look whether you agree with this decomposition of trace into diagonal and sum? You two have already provided useful advice in my previous PRs, thanks again for that!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1057,6 +1058,58 @@ class DecomposeAtenEinsumOp : public OpRewritePattern<AtenEinsumOp> {
...
+    Location loc = op.getLoc();
+    Value self = op.getSelf();
+    std::optional<unsigned> inRank = getTensorRank(self);
+    if (!inRank) {
+      return rewriter.notifyMatchFailure(
+          op, "Expected input tensor to have known rank.");
+    }
+    if (inRank != 2)
+      return rewriter.notifyMatchFailure(
+          op, "Expected input tensor to have rank 2.");
```
vivekkhandelwal1, line 1083: If you want then you can merge these two conditions.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1057,6 +1058,58 @@ class DecomposeAtenEinsumOp : public OpRewritePattern<AtenEinsumOp> {
...
+    Type elementType = inputType.getOptionalDtype();
+    Type diagonalType = inputType.getWithSizesAndDtype(
+        llvm::ArrayRef(diagonalShape), elementType);
+
+    Value diagonal = rewriter.create<AtenDiagonalOp>(
+        loc, diagonalType, /*input=*/self, /*offset=*/zero, /*dim1=*/zero,
+        /*dim2=*/one);
+    Value dtype = getDtypeIntValueForType(rewriter, loc, elementType);
+    Value sum = rewriter.create<AtenSumOp>(loc, outputType, /*self=*/diagonal,
+                                           /*dtype=*/dtype);
```
vivekkhandelwal1, line 1106: You can pass the `dtype` as None since the `AtenDiagonalOp` doesn't make any changes to the input dtype.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1057,6 +1058,58 @@ class DecomposeAtenEinsumOp : public OpRewritePattern<AtenEinsumOp> {
...
+    rewriter.replaceOpWithNewOp<tensor::CastOp>(op, output.getType(), sum);
```
vivekkhandelwal1, line 1107: This is wrong, tensor::cast works on `RankedTensorType` not `ValueTensorType`. Also, you don't need do any explicit casting here. Just do:
`rewriter.replaceOp(op, sum);`

vivekkhandelwal1 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
frafranz: Indeed, I now resorted to only checking the second one which is stricter than the first.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
frafranz: You're right! I simplified this in my new commit.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
frafranz: Thanks for making me aware of this! This simplified the decomposition even more.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "PassDetail.h"
 
+#include "mlir/Dialect/Tensor/IR/Tensor.h"
```
vivekkhandelwal1, line 12: Please remove this, it's not required.

vivekkhandelwal1 (approved):
Please address one small comment, rest LGTM.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
frafranz: Done, good catch!

frafranz:
> Please address one small comment, rest LGTM.

Hi @vivekkhandelwal1 thanks for your review, I also addressed this comment now and rebased. From my side we're ready to land! (someone else would have to merge it for me)

PR #2793, opened 2024-01-24
Describe how to get --debug and --debug-only flags in dev notes 

newling:
Change should be visible : https://github.com/newling/torch-mlir/blob/docs_update/docs/development.md


vivekkhandelwal1 (approved):
LGTM

Issue #2765, opened 2024-01-16
Importing llama2 ONNX model into torch-mlir generates invalid MLIR


kumardeepakamd:
Issue: func.func @<>() in imported torch onnx MLIR

Have ability to get llama2 model. I do following, but you can follow your own way that works.
pip install huggingface-cli huggingface-hub
Run huggingface-cli login and enter your HF token
Login to a machine with NVIDIA GPU and latest drivers.
#Set Hugging Face Home, an example below, change path to your own
export HF_HOME=<your path with enough disk space>/HF_HOME
Save following file as model.py
```
from transformers import LlamaForCausalLM, LlamaTokenizer
import torch

class optModel(torch.nn.Module):

    def __init__(self):
        super().__init__()
        self.model = LlamaForCausalLM.from_pretrained(
            "meta-llama/Llama-2-7b-hf")
        self.model.eval()

    def forward(self, tokens):
        attention_mask = torch.ones(tokens.shape, dtype=torch.long)
        return self.model.forward(input_ids=tokens, attention_mask=attention_mask)

tokenizer = LlamaTokenizer.from_pretrained("meta-llama/Llama-2-7b-hf")
test_input = tokenizer.encode("The Manhattan bridge", return_tensors="pt")

model = optModel()
onnx_program = torch.onnx.export(model, test_input, "llama-7b.onnx")
```
The last line export ONNX mode, in case you use a different script, stick in last line to your setup.

run: 'python ./model.py'
run: 'python -m torch_mlir.tools.import_onnx llama-7b.onnx -o llama-7b.torch-onnx.mlir'
Examine (head llama-7b.torch-onnx.mlir) you will see that mlir has func.func @<>() and body of the function is missing.

daveliddell:
Hi @vivekkhandelwal1 , I have a fix for this, working on getting into a PR now

kumardeepakamd:
@vivekkhandelwal1 assigning it back to you for you to fix the crash in torch MLIR, use Dave's PR of course to be able to get to crash stage.

vivekkhandelwal1:
> @vivekkhandelwal1 assigning it back to you for you to fix the crash in torch MLIR, use Dave's PR of course to be able to get to crash stage.

Hi @kumardeepakamd, this issue is fixed by: https://github.com/llvm/torch-mlir/pull/2803.

vivekkhandelwal1:
Closing this issue since the PR is merged.

daveliddell:
[...]

PR #2794, opened 2024-01-24
Make the onnx importer more robust for internal/external and large models

daveliddell:
Fix for https://github.com/llvm/torch-mlir/issues/2765

The onnx docs say that you can't do shape inference using the in-memory API for models > 2 GB.  This fix replaces that API with the file-based API.  Since the new API generates an intermediate file, also added a --keep switch to keep that file, which I delete by default.


daveliddell:
Also tested on my two-node test case.  It works on small models, too!  :-D

python/torch_mlir/tools/import_onnx/__main__.py
```diff
@@ -48,9 +48,15 @@ def main(args):
...
 
 
-def load_onnx_model(file_path: Path) -> onnx.ModelProto:
-    raw_model = onnx.load(file_path)
-    inferred_model = onnx.shape_inference.infer_shapes(raw_model)
+def load_onnx_model(file_path: Path, keep: bool) -> onnx.ModelProto:
+    # Do shape inference via files instead of in memory in order to handle
+    # models > 2 GB. See https://github.com/onnx/onnx/blob/main/docs/PythonAPIOverview.md#shape-inference-a-large-onnx-model-2gb
+    # for details about this technique.
+    inferred_path = file_path.with_stem(file_path.stem + '-inferred-shape')
```
stellaraccident, line 55: I think this is likely to interplay badly with automation and setups where the source directory is not read-write. We ultimately likely need to be a bit more switchy and have a flag for `--with-external-data` (as described here: https://github.com/onnx/onnx/blob/main/docs/PythonAPIOverview.md#loading-an-onnx-model-with-external-data)

Thinking of how to not boil the ocean. How about something like this:

* Use a `with tempfile.TemporaryDirectory(dir=flags.temp_dir, delete=not flags.keep_temps) as td:`
* `temp_path = Path(td.name, "inferred.onnx")`
* `onnx.shape_inference.infer_shapes_path(file_path, temp_path)`
* `inferred_model = onnx.load(temp_path, load_external_data=False)`
* `if flags.external_data: load_external_data_for_model(inferred_model, td.name)`

Or something like that. It can be made a lot better if anyone cares, but I'm not thrilled to do that without a reason. I think this will at least avoid some of the copies.

I hate proto based languages. Such a poor design decision.

stellaraccident (changes requested):

python/torch_mlir/tools/import_onnx/__main__.py
daveliddell: Oh, good idea!  I thought about this problem briefly but promptly forgot to deal with it.  On a previous product, we had the same issue with a temp index file.  We tried /tmp, user's home dir, and cwd, but had unhappy customers coming to us each time.  I think we finally gave them a temp_dir flag and the the wailing quieted down to a low grumbling.  :-D  I'll make these changes tomorrow when I'm feeling better.  Thanks for the feedback!

daveliddell: Wait, I think I'm still unclear on what we're doing with external data and whether this is related to shape inference or just something else we should be supporting (for large models).  This statement in particular would seem to imply that we don't have a choice with external data location when doing shape inference:

> Current shape_inference supports models with external data, but for those models larger than 2GB, please use the model path for onnx.shape_inference.infer_shapes_path and the external data needs to be under the same directory.

stellaraccident: We don't really support it yet (I think) but I was trying to leave the door opened since we'll need to get it right soon.

There are a few modes for transferring back and forth and we'll probably need to elaborate them.

stellaraccident: It is implicated here because if you save the graph to a directory different from the data, you need to use the APIs explicitly to tell it where the original external data was.

daveliddell: @stellaraccident New code with external data fails in MLIR verify.  Considering that the original code works, this is a regression.  :-(  Debugging it

daveliddell: Ok, I think I'm ready for you again.  After tweaking the export size, now everything is working, small models and llama.  Added unit tests to test the command line on two small models in both internal and external data scenarios.

Unfortunately, I had to sacrifice running the checker on large models, as I couldn't find a way of pointing the file-based checker to the external data that llama does seem to generate.  In-memory checker fails in the same way as in-memory shape inference due to the 2 GB limit.

We could do away with the temp directory for the shape-inferred model and just put the inferred model wherever the user requests it.  If that location defaults to the same dir as the original model, then I would be able to run the checker.  Seems a bit beyond what's needed for now, but your call.

test/python/onnx_importer/command_line_test.py
```diff
@@ -0,0 +1,145 @@
...
+        model_data_dir.mkdir(exist_ok=True)
+        convert_model_to_external_data(
+            onnx_model, all_tensors_to_one_file=True,
+            location=data_dir_name + "/data.bin",
+            size_threshold=48,
+            convert_attribute=True)
+        onnx.save(onnx_model, model_file)
+        temp_dir = run_path / "temp"
+        temp_dir.mkdir(exist_ok=True)
+        p = subprocess.run([
```
stellaraccident, line 125: Pro tip: For these cases, I often don't use a literal sub-process but just call into the main entrypoint from importing the model, giving it explicit command line arguments.

(can ignore / I might fix in a followup)

stellaraccident (approved):
Thanks for sweating through this. A couple of comments that are optional.

test/python/onnx_importer/command_line_test.py
stellaraccident: Little more clarity since you asked offline:

```
from torch_mlir.tools.import_onnx import __main__

...


args = __main__.parse_arguments([1, 2, 3])
__main__.main(args)
```

Something like that. Often if I'm doing it, I'll have an entry point in the module just for testing instead of needing to poke in in two steps like that.

stellaraccident (approved):

daveliddell:
@stellaraccident You're too fast for me!  :-D  Since the ideal features of tempfile.TemporaryDirectory weren't available in 3.11, I was thinking of dropping in tempfile.mkdtemp as a good substitute for hard-coded temp dir name with equivalent behavior.  Anyway, good to get this in quickly.  Thanks for the magic formula for calling main.  Works great and way faster, too!

stellaraccident:
Year, that's the way. But we can follow-up with it. What you have works and we can improve it as we go. Thanks for the work!

PR #2799, opened 2024-01-25
[torch-mlir][torch] add encoding field to torch type

aartbik:
This adds an encoding field to the torch type, using the interfaces for printing, parsing, and verification. Note that although this change prepares adding sparsity to the torch type (as illustrated by the round trip and invalid tests), nothing in this change depends on the actual contents of the encoding field!


stellaraccident:
Not to reopen old wounds, but I'm pretty sure we got it wrong upstream by adding a generic encoding attribute. For a very purpose built type like this, could we just call it "sparsity" and constrain the attribute type accordingly?

test/Dialect/Torch/sparse.mlir
ramiro050, line 1: nit: I think it makes sense to make the checks with the other tensor type checks here

https://github.com/llvm/torch-mlir/blob/66447b70aeb3dbb26e31437798ae3d71fb56da7f/test/Dialect/Torch/ops.mlir#L16-L29

aartbik:
> For a very purpose built type like this, could we just call it "sparsity" and constrain the attribute type accordingly?

I actually thought the encoding + interface (rather than a specific encoding with concrete implementation) was something everyone loved and wanted  ;-)

But I am totally open to renaming this to sparse_encoding or sparsity or something like that and restrict it to those concrete attributes only if you are more comfortable with that. What I love it that it is very light-weight, i.e. one optional attribute with very minor machinery to make it work (and I hope you love that part at least too ;-)

test/Dialect/Torch/invalid_sparse.mlir
ramiro050, line 1: Similar comment as above, this can be placed in https://github.com/llvm/torch-mlir/blob/main/test/Dialect/Torch/invalid.mlir

ramiro050 (approved):
Some minor comments. Other than addressing Stella's comment, the changes here LGTM

stellaraccident:
> > For a very purpose built type like this, could we just call it "sparsity" and constrain the attribute type accordingly?
> 
> I actually thought the encoding + interface (rather than a specific encoding with concrete implementation) was something everyone loved and wanted ;-)
> 
> But I am totally open to renaming this to sparse_encoding or sparsity or something like that and restrict it to those concrete attributes only if you are more comfortable with that. What I love about the direction is that it is very light-weight, i.e. one optional attribute with very minor machinery to make it work (and I hope you love that part at least too ;-)

Yeah, everyone loves it so much that they think it is theirs... And then who decides? :)

Torch is domain specific and doesn't any away from naming things for what they are vs the universe of possibilities that they may be some day.

Let's call it "sparse_encoding", strongly type the attribute and then represent it in asm with a keyword like "sparse #CSR...".

Still light, but named for purpose and removes a lot of type checking etc.

test/Dialect/Torch/sparse.mlir
aartbik: Perfect. I was not sure if you would prefer a new file for this, but what you suggest is very clean and makes perfect sense. Thanks!

test/Dialect/Torch/invalid_sparse.mlir
aartbik: SGTM

aartbik:
> Yeah, everyone loves it so much that they think it is theirs... And then who decides? :)

Yes, you are absolutely right on that (I was always hoping for an AttributeSet or something like that as way out, but that never materialized.

> Torch is domain specific and doesn't shy away from naming things for what they are vs the universe of possibilities that they may be some day.

That is a very good philosophy. Okay, sparse_encoding coming up!

aartbik:
All addressed, PTAL

ramiro050 (approved):
Thanks!

rdadolf:
Late to the discussion, but I just wanted to say +1 for the patch.

It looks [*very* familiar](https://github.com/rdadolf/torch-mlir/commit/b5975b75bea186e36ce052713ad718f76f6449b3) :smile:. Except you did it better.

I'm also glad we're moving to a more sparsity-specific name. The encoding was overloaded with unrelated things that I'm glad we won't need to overlap with.

PR #2801, opened 2024-01-25
[MLIR][ONNX] Add OnnxToTorch support for g-z ops

saienduri:
This commit adds the OnnxToTorch support for Mean, IsInf, IsNaN, and PRelu ops.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -850,4 +850,109 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
             binder.op, resultType, tensor, /*memory_format=*/noneVal);
         return success();
       });
+  patterns.onOp(
+      "Mean", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Value noneVal = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+        if (binder.op->getNumOperands() == 1) {
+          Torch::ValueTensorType resultType;
+          Value x;
+          if (binder.tensorOperand(x) || binder.tensorResultType(resultType))
+            return failure();
+          rewriter.replaceOpWithNewOp<Torch::AtenMeanOp>(binder.op, resultType,
+                                                         x, /*dtype=*/noneVal);
+          return success();
+        }
+        Torch::ValueTensorType resultType;
+        SmallVector<Value> valList;
+        int64_t numOperands = binder.op->getNumOperands();
+        if (binder.tensorOperands(valList, numOperands) ||
+            binder.tensorResultType(resultType))
+          return failure();
+        Value const1 = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(64), 1));
+        // Short circuit to binary add
+        if (numOperands == 2) {
+          Value sum = rewriter.create<Torch::AtenAddTensorOp>(
+              binder.getLoc(), resultType, valList[0], valList[1], const1);
+
+          rewriter.replaceOpWithNewOp<Torch::AtenMeanOp>(
+              binder.op, resultType, sum, /*dtype=*/noneVal);
+          return success();
+        }
+        // When binder.op->getNumOperands() > 2
+        auto baseType = Torch::ValueTensorType::getWithLeastStaticInformation(
+            binder.op->getContext());
+        Value curr = rewriter.create<Torch::AtenAddTensorOp>(
+            binder.getLoc(), resultType, valList[0], valList[1], const1);
+        for (int i = 2; i < numOperands; i++) {
+          if (i == numOperands - 1) {
+            curr = rewriter.create<Torch::AtenAddTensorOp>(
+                binder.getLoc(), resultType, curr, valList[i], const1);
+          } else {
+            curr = rewriter.create<Torch::AtenAddTensorOp>(
+                binder.getLoc(), baseType, curr, valList[i], const1);
+          }
+        }
+        rewriter.replaceOpWithNewOp<Torch::AtenMeanOp>(binder.op, resultType,
+                                                       curr, /*dtype=*/noneVal);
+        return success();
+      });
+  patterns.onOp(
+      "IsInf", 10, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value tensor;
+        int64_t neg;
+        int64_t pos;
+        if (binder.tensorOperand(tensor) ||
+            binder.s64IntegerAttr(neg, "detect_negative", 1) ||
+            binder.s64IntegerAttr(pos, "detect_positive", 1) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+        if (neg == 0) {
+          // replace all negative infs with 0
+          tensor = rewriter.create<Torch::AtenReluOp>(
+              binder.getLoc(),
+              dyn_cast<Torch::ValueTensorType>(tensor.getType()), tensor);
+        }
+        if (pos == 0) {
+          // first use neg op to flip positive inf to negative inf. Then relu to
+          // replace all positive infs with 0.
+          Value flip = rewriter.create<Torch::AtenNegOp>(
+              binder.getLoc(),
+              dyn_cast<Torch::ValueTensorType>(tensor.getType()), tensor);
+          tensor = rewriter.create<Torch::AtenReluOp>(
+              binder.getLoc(), dyn_cast<Torch::ValueTensorType>(flip.getType()),
+              flip);
+        }
+        rewriter.replaceOpWithNewOp<Torch::AtenIsinfOp>(binder.op, resultType,
+                                                        tensor);
+        return success();
+      });
+  patterns.onOp("IsNaN", 9,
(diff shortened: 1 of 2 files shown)


saienduri:
Actually noticed a change has to be made for Mean, will convert back to PR when ready.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -850,4 +850,110 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        if (numOperands == 2) {
+          Value sum = rewriter.create<Torch::AtenAddTensorOp>(
+              binder.getLoc(), resultType, valList[0], valList[1], const1);
+
+          rewriter.replaceOpWithNewOp<Torch::AtenDivScalarOp>(
+              binder.op, resultType, sum, numOperandsConstant);
+          return success();
+        }
+        // When binder.op->getNumOperands() > 2
+        auto baseType = Torch::ValueTensorType::getWithLeastStaticInformation(
```
vivekkhandelwal1, line 885: What's the difference between `baseType` and `resultType`?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -850,4 +850,110 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        if (binder.tensorOperands(valList, numOperands) ||
+            binder.tensorResultType(resultType))
+          return failure();
+        Value const1 = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(64), 1));
+        // Short circuit to binary add
+        if (numOperands == 2) {
+          Value sum = rewriter.create<Torch::AtenAddTensorOp>(
+              binder.getLoc(), resultType, valList[0], valList[1], const1);
```
vivekkhandelwal1, line 878: You can keep this out of this if condition and as a result this value can be used in the later part of code in place of `curr`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -850,4 +850,110 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        rewriter.replaceOpWithNewOp<Torch::AtenDivScalarOp>(
+            binder.op, resultType, curr, numOperandsConstant);
+        return success();
+      });
+  patterns.onOp(
+      "IsInf", 10, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value tensor;
+        int64_t neg;
+        int64_t pos;
```
vivekkhandelwal1, line 907: Combine them to a single line.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -850,4 +850,110 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            binder.s64IntegerAttr(neg, "detect_negative", 1) ||
+            binder.s64IntegerAttr(pos, "detect_positive", 1) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+        if (neg == 0) {
+          // replace all negative infs with 0
+          tensor = rewriter.create<Torch::AtenReluOp>(
+              binder.getLoc(),
+              dyn_cast<Torch::ValueTensorType>(tensor.getType()), tensor);
```
vivekkhandelwal1, line 918: Can't you directly use `resultType` here?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -850,4 +850,110 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        }
+        if (pos == 0) {
+          // first use neg op to flip positive inf to negative inf. Then relu to
+          // replace all positive infs with 0.
+          Value flip = rewriter.create<Torch::AtenNegOp>(
+              binder.getLoc(),
+              dyn_cast<Torch::ValueTensorType>(tensor.getType()), tensor);
```
vivekkhandelwal1, line 925: same here

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -850,4 +850,110 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+                  }
+                  rewriter.replaceOpWithNewOp<Torch::AtenIsnanOp>(
+                      binder.op, resultType, tensor);
+                  return success();
+                });
+  patterns.onOp("PRelu", 1,
+                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+                  Torch::ValueTensorType resultType;
+                  Value tensor;
+                  Value slope;
```
vivekkhandelwal1, line 950: Declare them in the same line

test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
```diff
@@ -572,3 +572,57 @@ func.func @test_not_2d(%arg0: !torch.vtensor<[3,4],i1>) -> !torch.vtensor<[3,4],
...
   }
+
+// CHECK-LABEL: func.func @test_mean_one_input
+  func.func @test_mean_one_input(%arg0: !torch.vtensor<[3],f32>) -> !torch.vtensor<[3],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 13 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
+    %0 = torch.operator "onnx.Mean"(%arg0) : (!torch.vtensor<[3],f32>) -> !torch.vtensor<[3],f32>
+    return %0 : !torch.vtensor<[3],f32>
+  }
+
+// CHECK-LABEL: func.func @test_mean_two_inputs
+  func.func @test_mean_two_inputs(%arg0: !torch.vtensor<[3],f32>, %arg1: !torch.vtensor<[3],f32>) -> !torch.vtensor<[3],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 13 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
```
vivekkhandelwal1, line 583: Please add a test for the case with more than two inputs.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
saienduri: baseType allows us to perform add operations without specifying shape information. This can help us in this situation because we will need to complete a series of add operations. In between, there might be broadcasts which can lead to resultType's shape not being the one needed at some of the add operations.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
saienduri: Here we use resultType, but later in the code we use baseType so need both currently.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
saienduri: resultType would be i1 in this case, but I want to stick with f32 until we actually call the aten isInf method, so nothing is lost.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
vivekkhandelwal1: Hi @saienduri, I don't see any difference between 'Value sum' at line no. 947, and the `Value curr` at line no. 957. That's why I said that we can keep only only the first one by taking it out of the if condition.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -920,4 +920,110 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        Torch::ValueTensorType resultType;
+        SmallVector<Value> valList;
+        int64_t numOperands = binder.op->getNumOperands();
+        Value numOperandsConstant = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(64), numOperands));
+        if (binder.tensorOperands(valList, numOperands) ||
+            binder.tensorResultType(resultType))
+          return failure();
+        Value const1 = rewriter.create<Torch::ConstantIntOp>(
```
vivekkhandelwal1, line 942: Rename to `constOne`.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
saienduri: ah makes sense, done

vivekkhandelwal1 (approved):

PR #2802, opened 2024-01-25
[FxImporter] make FxImporter to fit python<=3.9

qingyunqu:
As that torch with py3.9 is also used widely.

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -5,10 +5,16 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 # Also available under a BSD-style license. See LICENSE.
 
+from sys import version_info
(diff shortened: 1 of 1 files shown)


python/torch_mlir/extras/fx_importer.py
```diff
@@ -5,10 +5,16 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 # Also available under a BSD-style license. See LICENSE.
 
+from sys import version_info
```
stellaraccident, line 8: Instead of doing a complicated version check, wdyt about:

```
try:
  from types import None type
except ImportError:
  NoneType = type(None)
```

stellaraccident: With a comment that less than 3.10 doesn't have it

stellaraccident (approved):

PR #2804, opened 2024-01-25
[Torch Dialect] add fold pattern for aten.clone

qingyunqu:


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1654,6 +1654,18 @@ void AtenMaskedFillTensorOp::getCanonicalizationPatterns(
...
 }
 
+//===----------------------------------------------------------------------===//
+// AtenCloneOp
+//===----------------------------------------------------------------------===//
+
+void AtenCloneOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                              MLIRContext *context) {
+  patterns.add(+[](AtenCloneOp op, PatternRewriter &rewriter) {
+    rewriter.replaceOp(op, op.getSelf());
```
ramiro050, line 1664: I think we need to make sure the type of `self` is a `torch.vtensor`. If this transformation is performed before all tensors have value semantics, it would not be correct.

ramiro050 (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
qingyunqu: ok, done.

newling (commented):
Generally if no new operations are created, it is implemented as a fold method not a canonicalize method. 

See for example AtenRoundOp (hasFolder, fold).

qingyunqu:
> Generally if no new operations are created, it is implemented as a fold method not a canonicalize method.
> 
> See for example AtenRoundOp (hasFolder, fold).

ok, done.

ramiro050 (approved):
LGTM!

PR #2808, opened 2024-01-25
Converts all Adaptive Pooling Ops to Linalg

zjgarvey:
The previous conversions for AtenAdaptiveAvgPool1dOp and AtenAdaptiveMaxPool2dOp are refactored into a general templated conversion that works for all of the AtenAdaptive...PoolNdOp's. 

New support is added for the following ops:

1. AtenAdaptiveMaxPool1d
2. AtenAdaptiveMaxPool3d
3. AtenAdaptiveAvgPool3d

Support is also provided for passing inputs without batch dimensions. For example, applying adaptive_avg_pool2d to an input tensor of rank 3. 

After [pytorch #118162](https://github.com/pytorch/pytorch/pull/118162) gets down to torch-mlir, I'll add a test for AdaptiveMaxPool1d with return_indices (which will pass with that upstream fix).


zjgarvey:
The AdaptiveAvgPool2d tests are failing on dynamo with the expected error message:
```
Exception: Unsupported: missing default value for argument 0 in schema for aten.div.Tensor_mode
```
So I know to add those to the appropriate place in the xfail sets.

However, the dynamo failures for AdaptiveAvgPool3d are more cryptic and say:
```
error: unknown: failed to legalize operation 'torch.constant.int'
```
I'll do some digging to figure out what is going on there. I would like to assume it is secretly encountering the same issue as the other tests, but I'd like to figure out why it isn't giving the same error message.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -626,230 +626,68 @@ This is problematic for linalg ops for a few reasons:
...
-Current TODO's:
-    1. gather most of the boilerplate out of this op and make it into an
-adaptive pooling helper function.
-    2. figure out what to do with the conflicting decompositions in
-DecomposeComplexOps.cpp
-    3. Implement more efficient passes for when the kernel-size, input spatial
-dims, and output spatial dims are constant.
 */
 
+// TODO: infer template Dim from OpTy
```
newling, line 631: This would be good, see 

```
 template <typename T> struct DimensionTraits;
 ```
 
 in this file. Motivation: compile time

File: lib/Conversion/TorchToLinalg/Pooling.cpp [hunk 2]
```diff
@@ -626,230 +626,68 @@ This is problematic for linalg ops for a few reasons:
...
-    if (rank != 4) {
-      return rewriter.notifyMatchFailure(op, "only supports input type NCHW");
+    int64_t rank = inputType.getRank();
+    // get number of non-spatial dims
+    int64_t nonSpatial;
+    if (rank == Dim + 2 || rank == Dim + 1) {
+      nonSpatial = rank - Dim;
+    } else {
+      return rewriter.notifyMatchFailure(
+          op, "only supports inputs with one or two non-spatial dims");
```
newling, line 668: Is this a constraint imposed by torch, or a temporary constraint? 

Simpler:

int64_t nonSpatial = rank - Dim;
if (nonSpatial != 1 && nonSpatial != 2) { ... }

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -861,67 +699,82 @@ class ConvertAtenAdaptiveMaxPool2dOp
...
+      auto smallestFPValueAttr = rewriter.getFloatAttr(
+          elementType,
+          APFloat::getInf(
+              elementType.cast<mlir::FloatType>().getFloatSemantics(),
+              /*Negative=*/true));
+      buffVal = rewriter.create<arith::ConstantOp>(loc, elementType,
+                                                   smallestFPValueAttr);
+      auxTensor = rewriter.create<tensor::EmptyOp>(
+          loc, getAsOpFoldResult(outputSizes), auxTensorElementType);
+    } else {
```
newling, line 729: I would have

else if (isAvgPool) { ... }

else {
  unreachable error of some sort 
}

newling: Likewise for the other 2 places where there is a branching between the reduction types. 

Did you consider using old-fashioned polymorphism with virtual functions instead of if-else approach? It might be neater  (This isn't a review request, I'm just curious!).

newling (changes requested):
Looks like a nice improvement! Just a few minor suggestions.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -626,230 +626,68 @@ This is problematic for linalg ops for a few reasons:
...
-    const TypeConverter *typeConverter = getTypeConverter();
+    // check if we are doing a max pool or avg pool.
+    constexpr bool isMaxPool =
+        (llvm::is_one_of<OpTy, AtenAdaptiveMaxPool1dOp, AtenAdaptiveMaxPool2dOp,
+                         AtenAdaptiveMaxPool3dOp>::value);
+    static_assert(
+        isMaxPool ||
+            llvm::is_one_of<OpTy, AtenAdaptiveAvgPool1dOp,
+                            AtenAdaptiveAvgPool2dOp, AtenAdaptiveAvgPool3dOp,
+                            Aten_AdaptiveAvgPool3dOp>::value,
```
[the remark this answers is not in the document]
zjgarvey: > AdaptiveAvgPool3dOp appears twice here

In fact, Aten_AdaptiveAvgPool3dOp and AtenAdaptiveAvgPool3dOp are different. For some reason, dynamo was tracing the former rather than the latter, which was the reason for the cryptic torchdynamo failure mentioned in my previous comment. Frankly, it might be a good idea to add support for the other underscored ops just in case, i.e., Dim = 1 and 2.

> Rather do
> 
> constexpr bool isAvgPool = (llvm:: ... 
> static_assert(isMaxPool || isAvgPool, ... );

Yeah, that looks neater.

lib/Conversion/TorchToLinalg/Pooling.cpp
zjgarvey: Ah that's a good point. This is already part of the shape information for these ops (from torch), so perhaps it should be worded differently or removed. I imagine if you tried to put something of rank 2 into an adaptive pool 3d op, it would cause errors in python before getting to this conversion. I'll test that when I get back to my PC, but it's likely that this check should just be removed, or at least replaced with a check for negative nonSpatial value, since that's the only thing that would cause genuine problems with this conversion even if the shape information is loosened later on.

lib/Conversion/TorchToLinalg/Pooling.cpp
zjgarvey: > I would have
> 
> else if (isAvgPool) { ... }
> 
> else { unreachable error of some sort }

Is there a benefit to this change beyond readability? I can't imagine that the compiler would get past the static_assert at the beginning.

> Did you consider using old-fashioned polymorphism with virtual functions instead of if-else approach? It might be neater (This isn't a review request, I'm just curious!).

Hm, I hadn't considered that. Originally, I was offloading pieces to helper functions (similar to createPoolingOp in this file), but realized that I either had to duplicate a bunch of boilerplate or pass a lot of variables to each function, so I opted to just write a few if statements instead.  

Although, I'm thinking now that it might be worth the bit of redundancy to do something more modular, since it may be inconvenient to edit this pass in the future if one of these pytorch functions gets modified. For example, if adaptive max pooling with and without indices are separated, then someone would likely need to read through the whole pass to provide a fix for that.

I don't think it would take too long to convert the if statement cases into polymorphic helper functions, so I'm happy to try out this approach if you want.

lib/Conversion/TorchToLinalg/Pooling.cpp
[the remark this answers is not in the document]
newling: > In fact, Aten_AdaptiveAvgPool3dOp and AtenAdaptiveAvgPool3dOp are different
Ouch, that's sure to confuse future me. I'll try to remember that.

lib/Conversion/TorchToLinalg/Pooling.cpp
newling: I guess there's not much harm is checking things multiple times

newling (approved):
This looks good to me. Thanks for making the suggested changes (the template to determine dim and isMax is very nice). @zjgarvey feel free to get the opinion of a more experienced torch-mlir-er if you like.

zjgarvey:
The helper functions require quite a few arguments, but overall I think this polymorphism approach is probably better. Let me know what you think, @newling.

newling:
> The helper functions require quite a few arguments, but overall I think this polymorphism approach is probably better. Let me know what you think, @newling.

Nice! I think it makes the main function easier to follow. I have made a PR https://github.com/zjgarvey/torch-mlir/pull/1 against this which may/mayn't be a better design.

zjgarvey:
> Nice! I think it makes the main function easier to follow. I have made a PR [zjgarvey#1](https://github.com/zjgarvey/torch-mlir/pull/1) against this which may/mayn't be a better design.

Yeah, I can see the advantage there especially if I can move all of the helper functions to the external structure and not need to define derived classes for the conversion. I'll check this out, thanks for the prototyping @newling .

zjgarvey:
I'm going to make a few more changes right now to move the rest of the methods to the external helper.

zjgarvey:
@rsuderman I just updated this PR so that it should be fine to merge without problems. 

Would you mind taking a look?

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -794,65 +812,123 @@ class ConvertAtenAdaptiveAvgPool1dOp
...
     // get rank of input (same as rank of output)
-    int64_t rank =
-        adaptor.getSelf().getType().cast<RankedTensorType>().getRank();
-    // input operand should be NCHW (i.e. rank 4)
-    if (rank != 4) {
-      return rewriter.notifyMatchFailure(op, "only supports input type NCHW");
+    const int64_t rank = inputType.getRank();
+    // get number of non-spatial dims
+    const int64_t nonSpatial = rank - Dim;
+    if (nonSpatial < 0) {
```
renxida, line 903: noice. 

early returns > else

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -629,159 +629,177 @@ This is problematic for linalg ops for a few reasons:
...
-DecomposeComplexOps.cpp
-    3. Implement more efficient passes for when the kernel-size, input spatial
-dims, and output spatial dims are constant.
 */
 
 namespace {
-class ConvertAtenAdaptiveAvgPool1dOp
-    : public OpConversionPattern<AtenAdaptiveAvgPool1dOp> {
+
+class AdaptivePoolingHelper {
```
renxida, line 636: "Helper" usually refers to helper functions. I think a better name for this would be something that makes it clear that this eventually templates / inherits down to something that provides an OpConversionPattern.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -629,159 +629,177 @@ This is problematic for linalg ops for a few reasons:
...
-    2. figure out what to do with the conflicting decompositions in
-DecomposeComplexOps.cpp
-    3. Implement more efficient passes for when the kernel-size, input spatial
-dims, and output spatial dims are constant.
 */
 
 namespace {
-class ConvertAtenAdaptiveAvgPool1dOp
-    : public OpConversionPattern<AtenAdaptiveAvgPool1dOp> {
+
```
renxida, line 635: My understanding of the design is that we have AdaptivePoolingHelper, which is inherited by an AdaptiveMaxPoolingHelper and an AdaptiveAvgPoolingHelper, each of which gets templated to the 1-3d case.

The code was a little hard to read for me and I wonder if there's an easier way to do this.

renxida: I don't know enough to comment on the tradeoff between:
1. code duplication
2. somebody reading this code having to jump around due to inheritance and templating
but generally in this codebase i see a lot of places where duplication is preferred over inheritance / templating for cases like 1d-3d. E.g. the conv1d-3d ops.

renxida (commented):
I'm by no means familiar with this codebase and most of my comments are "i wonder" type comments. I would so love it if @zjgarvey could explain the thought processes behind some of the decisions. Maybe we should @ somebody more experienced to comment on it afterwards.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -629,159 +629,177 @@ This is problematic for linalg ops for a few reasons:
...
+public:
+  // Constructor for AdaptiveMaxPoolingHelper. Just forwards all arguments
+  // (except the OpConversionPattern) to the base class constructor.
+  template <typename... Args>
+  AdaptiveMaxPoolingHelper(const OpConversionPattern<OpTy> &ocp, Args &&...args)
+      : AdaptivePoolingHelper(std::forward<Args>(args)...),
+        opConversionPattern(ocp) {}
+
+  LogicalResult auxTensorSetup(OpTy op, const SmallVector<Value> &outputSizes,
```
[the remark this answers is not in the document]
zjgarvey: Hi, the auxiliary tensor is a tensor constructed in the first linalg generic op that is not the pooling operation. Both adaptive avg and adaptive max need some auxiliary tensor. I'm doing avg pooling first with a sum pool so as not to repeatedly compute and divide by the kernel volume on each iteration of the loop; therefore, I use the aux tensor keep track of what the kernel volume is for the sum pooling operation (so that it can later tensor divide sumpool/kernelVol). For max pooling, the function unfortunately always needs to return indices, so the indices (i.e., the integer position in the flattened input at which the local max occurs) need to be recorded in an aux tensor.

I offloaded this logic to a helper function to facilitate switching between those two uses (or possibly not doing anything if no aux tensor is required for a future op).

projects/pt1/python/torch_mlir_e2e_test/test_suite/pooling.py
```diff
@@ -1164,3 +1353,106 @@ def forward(self,x):
...
+        ([1, 512, 10, 16, 17], torch.float32, True)
+    ])
+    def forward(self,x):
+        return self.amp3d(x)
+
+@register_test_case(
+    module_factory=lambda: AdaptiveMaxPool3dStaticWithIndices())
+def AdaptiveMaxPool3dStaticWithIndices_basic(
+        module, tu: TestUtils):
+    module.forward(tu.rand(1, 512, 10, 16, 17))
```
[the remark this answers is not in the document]
zjgarvey: Yeah, I tried to add enough tests to cover most of the non-trivial cases, but it would certainly be better to randomize some of the input/output dims. 

I'm not sure if that is what you meant by a numeric test beyond these e2e tests, which seem quite numeric to me since they compare the python output with the compiled-from-mlir output.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -935,64 +1010,70 @@ class ConvertAtenAdaptiveMaxPool2dOp
...
-                                                    inputSpatialSizes[1]);
-          Value indexOut1 =
-              b.create<arith::AddIOp>(loc, indexOut0, inputElementIndices[3]);
-          Value indexOut1Int = castIndexToInt64(b, loc, indexOut1);
-          Value indexOut2 =
-              b.create<arith::SelectOp>(loc, cond1, indexOut1Int, maxIndex);
-          Value out2 = b.create<arith::SelectOp>(loc, cond1, inElt, res);
-          b.create<linalg::YieldOp>(loc, ValueRange({out2, indexOut2}));
+          Value out2, auxOut;
+          // customize for max vs. avg:
```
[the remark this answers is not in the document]
zjgarvey: It customizes two parts of the linalg generic op payload: the pooling op itself and also the aux tensor computation. 

I think I am going to add a comment at the beginning to try and give a summary of what each one is supposed to do.

lib/Conversion/TorchToLinalg/Pooling.cpp
zjgarvey: I think it is just used to store some helper functions and help reduce the number of args passed between helper functions by storing them in the class. 

The additional AdaptivePoolingOpsTraits classes are used to get the dim from the optype and alias the helper class name to something generic so it can be referenced in the conversion pattern.

lib/Conversion/TorchToLinalg/Pooling.cpp
zjgarvey: > I don't know enough to comment on the tradeoff between:
> 
>     1. code duplication
> 
>     2. somebody reading this code having to jump around due to inheritance and templating
> 
> 
> but generally in this codebase i see a lot of places where duplication is preferred over inheritance / templating for cases like 1d-3d. E.g. the conv1d-3d ops.

That is a good question. Do you think simply adding some more/better commentary in the code would make this more tractable? 

> My understanding of the design is that we have AdaptivePoolingHelper, which is inherited by an AdaptiveMaxPoolingHelper and an AdaptiveAvgPoolingHelper, each of which gets templated to the 1-3d case.
> 
> The code was a little hard to read for me and I wonder if there's an easier way to do this.

Yeah, I found out I left in something which probably makes it a bit more complicated for no reason, so let me make a change and add some better comments. 

Before this, I tried a few different implementations, and felt like this was probably a smidge less readable than some other options, but much easier to adapt in the future if need be.

projects/pt1/python/torch_mlir_e2e_test/test_suite/pooling.py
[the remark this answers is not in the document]
renxida: oh cool! i didn't realize that the e2e tests did that. ty!

newling (approved):
Apologies for the delay in reviewing. This looks good to me!

zjgarvey:
@renxida do you think the comments I added in the last commit are sufficient to clear up the readability concerns you had earlier?

renxida (approved):

PR #2811, opened 2024-01-26
[torch] Support `torch.convolution` quantized lowering to `linalg`

rsuderman:
Linalg has quantized specific operations. We can lower to these
operations when there is a known zeropoint and scale operations. This
allows the `convolution` to occur with lower bitwidth's, improving the
overall performance.


renxida:
There's a test that has been running for 1hr40 already i wonder what's going on there.

I had a long-running test that didn't stop after 6 hrs a few days ago. but the issue ended up fixing itself, after the following changes were made, among others:
- me fixing a duplicate pattern add
- me fixing a LLVM submodule version issue
- anush restarting CI.

renxida (approved):
LGTM

PR #2812, opened 2024-01-26
[RFC] Clang format refresh

qedawkins:
After noticing a number of commits with unrelated formatting changes,
I think something was changed with clang-format at one point and we're
seeing a number of unrelated changes. Doing a refresh can help avoid
this.

The changes made here came from
```
find lib -iname *.h -o -iname *.cpp  | xargs clang-format -i --style=llvm
find include -iname *.h -o -iname *.cpp  | xargs clang-format -i --style=llvm
find projects -iname *.h -o -iname *.cpp  | xargs clang-format -i --style=llvm
```

Looking for comments on the right way to go about this, given sweeping changes
like this can interrupt ongoing work. Also some of the formatting changes to comments
needs some closer attention to make sure they aren't mangled.


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -39,8 +40,9 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
             binder.f32FloatAttr(beta, "beta", 0.5f) ||
             binder.tensorResultType(resultType))
           return failure();
-        
-        // HardSigmoid computes the following expression: max(0, min(1, alpha * x + beta))
+
+        // HardSigmoid computes the following expression: max(0, min(1, alpha *
+        // x + beta))
```
ramiro050, line 45: nit: move equation to next line

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -110,22 +110,29 @@ LogicalResult prepareArgumentsForSlicingOp(OpTy op, OpAdaptor adaptor,
...
+// padding_right = 1 tensor([[[3., 2., 1., 0., 1., 2., 3., 2.],
 //          [7., 6., 5., 4., 5., 6., 7., 6.]]])
-// Checks: 1) Each of padding_left and padding_right must be non-negative less than size of last dimension 
-// Implementation: a) Construct a result tensor of shape of input tensor except for the last dimension.
-//                    The last dimension of the result tensor should be last dimension of input tensor +
-//                    left padding size + right padding size. INitialize result tensor to all zeros
-//                 b) Setup affine map to take slice from input tensor of size left padding starting from
-//                    second column onwards as first column is reflection boundary
+// Checks: 1) Each of padding_left and padding_right must be non-negative less
+// than size of last dimension Implementation: a) Construct a result tensor of
```
ramiro050, line 117: insert newline before `a)`

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -158,91 +165,109 @@ class ConvertAtenReflectionPad1dOp
...
+        getConstant(rewriter, loc, padInts[PAD_LEFT], indexType);
+    tileWidth[PAD_RIGHT] =
+        getConstant(rewriter, loc, padInts[PAD_RIGHT], indexType);
     tileWidth[PAD_CENTER] = lastDimSize;
 
     extractOffset[PAD_LEFT] = one;
-    // for (1,2,4) input, padding (3,1) lastDimSize=4, 4 - 1 - 1 = 2   [3,5, 6,7], so start offset to 6, which is right
-    // lasDimSize - (tileWidth[PAD_RIGHT] + one)
-    extractOffset[PAD_RIGHT] = createISub(lastDimSize, createIAdd(tileWidth[PAD_RIGHT], one)); 
+    // for (1,2,4) input, padding (3,1) lastDimSize=4, 4 - 1 - 1 = 2   [3,5,
```
ramiro050, line 193: insert newline before `[3,5,`

projects/ltc/csrc/base_lazy_backend/mlir_lowering_context.cpp
ramiro050, line 1: The lazy tensor code seems to follow a different format from the rest of torch-mlir (for example, snake_case variable names, instead of camel case). And I remember the placement of the pointer `*` character being discussed in a previous PR. @antoniojkim, @ke1337, is clang-format fine with you?

ramiro050 (approved):
Thanks for this! We should be having the CI check that changes are properly formatted before submitting. I'm not sure why we never added that. I'm happy to look into adding it to the CI

ramiro050:
Can you also make sure that running `git clang-format main` on your branch does not add any new changes?

qedawkins:
> Thanks for this! We should be having the CI check that changes are properly formatted before submitting. I'm not sure why we never added that. I'm happy to look into adding it to the CI

For some reason I had assumed CI was checking this, but I guess not. No wonder this got so bad. If you have cycles to set that up that would be really helpful!

.git-blame-ignore-revs
```diff
@@ -0,0 +1,11 @@
...
+# Such commits do not meaningfully contribute to git history, and include
+# large-scale mechanical changes like code formatting style changes.
+#
+# To set this file as the default ignore file for 'git blame', run:
+# ```shell
+#   git config blame.ignoreRevsFile .git-blame-ignore-revs
+# ```
+
+# Refresh clang-format
+2844576efc9b81f1900f567732f9e67526f1d3f6
```
qedawkins, line 11: I'll have to send this as a follow up so that I can get the commit hash right after squashing (easier to keep the comment updates and the mechanical stuff separate).

ramiro050:
> For some reason I had assumed CI was checking this, but I guess not. No wonder this got so bad. If you have cycles to set that up that would be really helpful!

Turned out to be simple: https://github.com/llvm/torch-mlir/pull/2816

projects/ltc/csrc/base_lazy_backend/mlir_lowering_context.cpp
qedawkins: @antoniojkim @ke1337 I'm going to land this today to avoid letting the problem get worse. We can address other formatting concerns in a follow up.

PR #2815, opened 2024-01-26
Fix illegal use of TypeRange

rsuderman:
TypeRange is an ArrayRef<Type> and therefore cannot be safely instantiated from a list initializer.


renxida (approved):
LGTM

PR #2817, opened 2024-01-27
[torch-mlir] provide FX traced graph importer for sparse tensors

aartbik:
Note that we are waiting for actual FX traced graph support for sparse tensors. For details see

https://github.com/pytorch/pytorch/issues/117188

Until then, however, we provide this clever importer that builds the FX traced graph for for the dense case and then puts a sparse annotation back on the parameters.

With import test.

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -207,10 +207,35 @@ SYMBOLIC_OP_TO_TORCH_OP = {
 }
 
 
-"""Check whether an object in our graph is symbolic"""
+def sparsity_encoding(shape: torch.Size, sparse_layout : torch.layout) -> str:
+   """Returns sparse tensor encoding for the given sparse layout as string.
+
+   Note that we currently rely on string parsing to construct the MLIR tensor
+   type (both for dense as well as sparse). In the long run this should be
+   replaced by proper IR building, obviously.
+
+   Also note this method currently just supports 2-dim sparse formats. This should
+   be generalized to the torch.sparse encodings for prefix dense batch dimensions
+   and suffix dense subtensor dimensions. Since MLIR supports a superset of what
+   is currently implememented in torch.sparse, this should not a be problem.
+   """
+
+   if len(shape) != 2:
+       raise RuntimeError(f"Unsupported sparse rank {len(shape)}")
+
+   if sparse_layout is torch.sparse_coo:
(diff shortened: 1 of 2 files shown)


python/torch_mlir/extras/fx_importer.py
```diff
@@ -207,10 +207,35 @@
 }
 
 
-"""Check whether an object in our graph is symbolic"""
+def sparsity_encoding(shape: torch.Size, sparse_layout : torch.layout) -> str:
+   """Returns sparse tensor encoding for the given sparse layout as string.
+
+   Note that we currently rely on string parsing to construct the MLIR tensor
+   type (both for dense as well as sparse). In the long run this should be
+   replaced by proper IR building, obviously.
```
stellaraccident, line 215: We can use upstream apis but we resort to string pasting for torch APIs because torch-mlir doesn't expose sufficient python APIs for all of this -- and doesn't have a super good way to export them to downstreams even if it did.

stellaraccident (commented):
I wasn't quite ready to accept contributions to this part yet. Let me synchronize the downstream fork tomorrow and make sure we don't get into a merge conflict.

python/torch_mlir/extras/fx_importer.py
aartbik: Ah, I will remove the comment in that case (given that it is uncertain we will change that soon)

aartbik:
> I wasn't quite ready to accept contributions to this part yet. Let me synchronize the downstream fork tomorrow and make sure we don't get into a merge conflict.

I suspect conflicts will be very light, since I am only touching a few tensor type building parts. But let me know when you are ready.

stellaraccident (commented):
Should be fine but I'm just trying to juggle less. I'm juggling a downstream pr and almost have the spice flowing to not need the fork. Give me a few hours

aartbik:
> Should be fine but I'm just trying to juggle less. I'm juggling a downstream pr and almost have the spice flowing to not need the fork. Give me a few hours

Of course. Take your time! Just reply here when you are comfortable moving forward.

stellaraccident (approved):

stellaraccident:
I've gone ahead and taken steps to unfork the code from the downstream (https://github.com/openxla/iree/pull/16276). Will take a release cycle to percolate but in the meantime, we're fine with this minor difference.

aartbik:
Thanks for your flexibility moving this forward quickly!

PR #2828, opened 2024-01-30
[CI] Change Roll PyTorch runner

vivekkhandelwal1:

### DIFF
--- a/.github/workflows/RollPyTorch.yml
+++ b/.github/workflows/RollPyTorch.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build_linux:
     name: Manylinux Build
-    runs-on: a100
+    runs-on: torch-mlir-cpubuilder-manylinux-x86-64
 
     # Don't run this in everyone's forks.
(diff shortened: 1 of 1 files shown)


Abhishek-Varma (approved):

vivekkhandelwal1:
@stellaraccident, did I select the correct runner?

stellaraccident:
Yes, but this action will need to be reworked:

* The build packages script doesn't work on this runner anymore. It will be deleted shortly.
* I have revoked my tokens which were being used to make these modifications because I didn't want them leaking.

There isn't much reason this needs to be using custom runners at all: it could just open the pr and then we could have a job that submits it when tests pass. We'll still need to deal with the security situation, though, and I don't want that in this repo. Let me talk to the llvm admin folks about creating a repo for release activities that we can tightly control.

stellaraccident:
https://discourse.llvm.org/t/rfc-create-companion-repo-torch-mlir-release/76643?u=stellaraccident

PR #2830, opened 2024-01-30
[onnx] Convert resources to denseattr for `onnx.constant` to `torch`

rsuderman:
`onnx` explicitly specifies that `raw_data` is stored in `little-endian` layout. While converting
to `torch` we need to convert from a known endian format to an internal format of consistent
layout. This means endianness must be correct during the import of `onnx.Constant`.


renxida (approved):
looks good. fixed minor typo.

PR #2831, opened 2024-01-30
[onnx] Import `onnx` constants as `onnx.Constant` instead of literals

rsuderman:
To handle the conversion from raw bytes to `DenseElementsAttr` we need
to handle the endianness conversion during `torch-onnx-to-torch`.
Therefore when importing `onnx.Constant` it is better to represent using
the `onnx` constant operation so that only one location requires the
endianness correction.


renxida (approved):
Simple enough

PR #2833, opened 2024-01-31
[torch] Supporting `torch.aten.mul.float` lowering to `arith`

rsuderman:
Simple missing scalar operation for multiply floats was missing.


daveliddell (approved):
Ok, I get it :-)

PR #2795, opened 2024-01-24
[onnx] Add testing using the `onnx` compilation using torch tests

rsuderman:
We can route the torch tests via `onnx` using the `torch.onnx.export`
tooling. We can then reimport, lower to torch, and compile to linalg to
validate the onnx path is working correctly.

The current implementation exposes some failures in the `onnx` path so
we cannot enable the onnx test suite yet due to segmentation faults.

PR #2834, opened 2024-01-31
[onnx] Fix `onnx-to-torch` lowering for flatten shape

rsuderman:
The existing `flatten` lowering did not define what the intermediate
shape was. This could result in failures to lower further to linalg as
the intermediate shape was unknown. Added a shape refinement section.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -1421,7 +1421,10 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
 
         // We collapse in the dimensions to the right of the axis.
         for (int i = axis + 1; i < rank; ++i) {
-          shape[axis] *= shape[i];
+          bool dynamic = shape[axis] == Torch::kUnknownSize ||
+                         shape[axis] == Torch::kUnknownSize;
+          if (!dynamic)
+            shape[axis] = shape[axis] * shape[i];
         }
 
         shape.resize(axis + 1, 1);
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -486,7 +486,6 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
           auto message = llvm::formatv("unimplemented support for the given "
                                        "dtype conversion (onnx 'type' = {0})",
                                        dtypeIntOnnx);
-          llvm::errs() << message << "\n";
           auto y = rewriter.notifyMatchFailure(binder.op, message);
 
           return y;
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1455,10 +1462,7 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
 
         // Otherwise, collapse the left range into a single dimension:
         // torch._prims.collapse(cr, 0, axis - 1)
-        Value axisLess1Const = rewriter.create<Torch::ConstantIntOp>(
```
newling, line 1458: With this CollapseOp removed, it seems like now there is at most one collapse op (for the right-most dimensions). How can this result in a rank-2 tensor, do you not still need to collapse the left-most dimensions?

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1412,16 +1411,24 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        int64_t rank = shape.size();
+
         // If axis is negative, count from the right instead of left
-        int64_t rank =
-            cast<Torch::ValueTensorType>(operand.getType()).getSizes().size();
         if (axis < 0)
           axis = rank + axis;
 
-        Value collapsedRight;
-        auto baseType = Torch::ValueTensorType::getWithLeastStaticInformation(
```
newling, line 1422: I didn't notice this in my review of this the first time around, should have. One question about the replacement code is, does it work with dynamic dimensions? Should it be something like 

 for (int i = axis + 1; i < rank; ++i) {
          if (shape[i] == kUnknownSize) { shape[axis] = kUnknownSize; break;  }
          shape[axis] *= shape[i];
}

newling (commented):
Did this come up in a downstream test? Would this be caught with https://github.com/llvm/torch-mlir/pull/2795 landed?

rsuderman:
> Did this come up in a downstream test? Would this be caught with #2795 landed?

#2795 will help catch these tests in the future. It was caught by attempting to push the first model through (resnet) and we found this error, along with a few others.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
rsuderman: I removed this note realizing it was necessary. From the sample cases I was looking at it appeared to be unneeded. It was restored when addressing test failures.

newling (commented):
Just the 2 issues
1) axis == rank
2) handling dynamic

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
rsuderman: If `axis == rank` then the loop never executes meaning we never index into `shape`.

rsuderman:
> Just the 2 issues
> 
> 1. axis == rank
> 2. handling dynamic

If axis == rank then the loop never executes meaning we never index into shape.

newling (changes requested):
The `bool dynamic = .` line needs another look

renxida (changes requested):
It looks like at DefaultDomainAtoF.cpp:1424 there's a redundant or. Did you mean shape[i]?

Code:
```cpp
         bool dynamic = shape[axis] == Torch::kUnknownSize ||
                         shape[axis] == Torch::kUnknownSize;
```
otherwise lgtm

rsuderman:
> The `bool dynamic = .` line needs another look

Woops, fixed it up. Just needed to check whether the value is dynamic.

newling (approved):
looks all good to me now

renxida (approved):

PR #2836, opened 2024-01-31
[torch] Support `!countIncludePad` when unpadded for average pool

rsuderman:
We do not support average pool when `countIncludePad is set to false.
However if the input is unpadded then the setting of the boolean is
unneeded. Extended use by checking if padding is zero before rejecting
the lowering.


renxida (approved):
LGTM

PR #2837, opened 2024-01-31
[onnx] Fix `pool` lowering for non-symmetric padding

rsuderman:
`torch` requires that padding be symmetric for pooling operations. To
support non-symmetric pad we need to separately materialize out the
padding operation.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -262,29 +262,84 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         if (!maybeRank)
           return rewriter.notifyMatchFailure(binder.op,
                                              "Unimplemented: unranked tensor");
-        unsigned rank = *maybeRank;
+        int64_t rank = *maybeRank;
+        int64_t spatial = rank - 2;
 
         SmallVector<int64_t> kernel, padding, strides, dilations;
         if (binder.s64IntegerArrayAttr(kernel, "kernel_shape", {}))
           return rewriter.notifyMatchFailure(binder.op,
                                              "kernel_shape bind failure");
-        if (kernel.size() != rank - 2)
+        if (kernel.size() != spatial)
           return rewriter.notifyMatchFailure(
               binder.op, "kernel list size does not match the number of axes");
-        if (binder.s64IntegerArrayAttr(padding, "pads", {0}))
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -262,30 +262,85 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          }
+          if (equal)
+            padding.resize(spatial);
+        }
+
+        // Torch pool operators require symmetric padding so we materialize the
+        // padding behavior explicitly and set the padding to 0.
+        if (padding.size() == 2 * spatial) {
+          auto operandTy = cast<Torch::ValueTensorType>(operand.getType());
+          llvm::SmallVector<int64_t> shuffledPadding(spatial * 2, 0);
```
newling, line 312:
```suggestion
          llvm::SmallVector<int64_t> shuffledPadding(spatial * 2);
```

nit: to set the expectation that values will be set in loop

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -262,30 +262,85 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        if (padding.size() == 2 * spatial) {
+          bool equal = true;
+          for (int i = 0; i < spatial; ++i) {
+            equal = equal && (padding[i] == padding[i + spatial]);
+          }
+          if (equal)
+            padding.resize(spatial);
+        }
+
+        // Torch pool operators require symmetric padding so we materialize the
```
newling, line 308: nit: It's not clear what 'symmetric padding' is in this comment.  Comment says padding is set to 0, maybe make clear that 0 is -inf/lowest.

newling (approved):
Nice code. The lit test is rather long, but they all are in these onnx->torch tests. The quick numerical tests would be good for this, if/when possible. Especially good for testing complex combinations of padding, dilation and striding (does pooling start top-left of padded image?) in the future 

@renxida might want to take a look.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
rsuderman: Corrected from `0` to lowest. We should have numeric tests on the way but trying to sort out crashed before enabling the  test suite.

PR #2839, opened 2024-01-31
Remove build automation that is no longer used.

stellaraccident:
Most of this can't actually work given the token/security issues, so even if we re-introduce it, we are better off fetching from history vs carrying things that will lead folks astray.


ramiro050 (approved):
LGTM!

stellaraccident:
@sjain-stanford since I authored this PR, you started using some of the bits in the torch-mlir-release repo. Could you let me know if I'm deleting anything in this patch that you need and I'll revert those parts.

sjain-stanford:
> Could you let me know if I'm deleting anything in this patch that you need and I'll revert those parts.

Thanks for checking @stellaraccident . We'll need to keep the `build_tools/python_deploy/build_linux_packages.sh` script and any if its transitive deps (like Dockerfile etc). Let me take a closer look and get back here shortly.

sjain-stanford:
Here's the updated list:
- `build_tools/python_deploy/build_linux_packages.sh`
- `build_tools/docker/Dockerfile` (not being removed in this PR which is ok)
- `test-requirements.txt` (not being removed in this PR which is ok)
- `build-requirements.txt` (not being removed in this PR which is ok)

Reverting these (from being deleted) and landing SGTM. If the nightly release tomorrow breaks, we can send a fix PR as a follow-on.

One small request: Could you paste the stuff being removed from `docs/development.md` into `build_linux_packages.sh` as comments for future reference? Some of this is useful info and carrying it forward makes sense in the script (not in the readme).

docs/development.md
```diff
@@ -225,106 +225,6 @@ is built as part of the LLVM project along with MLIR.
 We mount a ccache and pip cache inside the docker container to speed up iterative builds. Iterative
 builds should be as fast as running without docker.
 
-### In-Tree builds
```
sjain-stanford, line 228: One small request: Could you paste the stuff being removed from `docs/development.md` into `build_linux_packages.sh` (as comments) for future reference? Some of this is useful info and carrying it forward makes sense in the script (not in the readme).

PR #2844, opened 2024-01-31
[MLIR][TORCH] Add OnnxToLinalg lowering pipeline

vivekkhandelwal1:
This commit adds the lowering pipeline for the onnx-torch IR to linalg-on-tensors backend IR. The pipeline comprises of three stages: 1.) OnnxToTorch Lowering Pass
2.) TorchFunctionToTorchBackend Pipeline
3.) TorchBackendToLinalgOnTensorsBackend Pipeline


test/Dialect/Torch/onnx-backend-to-linalg-on-tensors-backend-pipeline.mlir
```diff
@@ -0,0 +1,99 @@
+// RUN: torch-mlir-opt -pass-pipeline='builtin.module(onnx-backend-to-linalg-on-tensors-backend-pipeline)' -split-input-file %s | FileCheck %s
+
```
newling, line 2: Is it necessary to have check resulting IR quite so thoroughly? I would have thought something like 

// CHECK-LABEL: test_softmax_axis_0
// CHECK-NOT:    onnx
// CHECK-NOT:    torch
// CHECK:            return
// CHECK-SAME: tensor<3x4x5xf32>

would be enough.

Motivation: avoiding brittle tests 

I'm also curious how you create long lit tests, isn't it a bit tedious :-D

vivekkhandelwal1: There's a python script to generate the checks, located at `llvm-project/mlir/utils/generate-test-checks.py`.

newling: Thanks for sharing the script link. My feeling is that there is too much being tested here, interested to hear your thoughts.

vivekkhandelwal1: Hi @newling, I'm not sure if there exists any specific guideline anywhere about adding these checks but in my opinion having thorough checks do not have any disadvantage as compared to doing the reverse. Also, for the test where the lowering contains a large set of ops being generated it's really hard to add the checks selectively. This is the reason that I have added the complete checks. I think someone more familiar with the LLVM contribution guidelines can comment better on this.

vivekkhandelwal1: Also, this particular PR adds a pipeline to lower the ONNX Torch IR to Linalg IR, I think the complete checks here are a representative of the IR being generated during the execution of this complete pipeline.

newling: The disadvantage that I see is that an unrelated change in any of the passes in the pipeline can make this test fail. For example if the order in which IR is created changes in `DecomposeAtenLeakyReluOp::matchAndRewrite` then this test will fail. The person who makes the unrelated change then needs to understand what this test is doing, and update it. That's what I call a "brittle" test. 

I understand your concern about "doing the reverse", but in this case I don't think my suggestion above is undertesting what the new functionality adds?

stellaraccident: > There's a python script to generate the checks, located at `llvm-project/mlir/utils/generate-test-checks.py`.

Yes, I know about that. I often have to come back and properly generalize tests when people have wholesale used it.

+1 to a basic check that the pipeline is doing what it should do and not an overly comprehensive integration test.

stellaraccident (commented):
What is this doing above and beyond the normal torch to linalg pipeline?

I'd rather just teach that one what to do vs having another bespoke pipeline.

stellaraccident:
> What is this doing above and beyond the normal torch to linalg pipeline?
> 
> I'd rather just teach that one what to do vs having another bespoke pipeline.

(Or if just adding a pass at the front door onnx, delegate to the regular one)

lib/Dialect/TorchConversion/Transforms/Passes.cpp
```diff
@@ -101,6 +106,14 @@ void TorchConversion::createTorchBackendToLinalgOnTensorsBackendPipeline(
   pm.addPass(TorchConversion::createVerifyLinalgOnTensorsBackendContractPass());
 }
 
+void TorchConversion::createOnnxBackendToLinalgOnTensorsBackendPipeline(
+    OpPassManager &pm) {
+  pm.addNestedPass<func::FuncOp>(onnx_c::createTorchOnnxToTorchPass());
+  Torch::TorchLoweringPipelineOptions options;
+  Torch::createTorchFunctionToTorchBackendPipeline(pm, options);
```
ramiro050, line 113: Is this entire pipeline needed? From skimming the ONNX Python importer, it seems that it has similar guarantees as the FX Python importer (value semantics, shapes, dtypes, etc). If that's the case, then we don't need to perform the entire `TorchFunctionToTorchBackendPipeline`. I think you just need the decompositions pass.

test/Dialect/Torch/onnx-backend-to-linalg-on-tensors-backend-pipeline.mlir
ramiro050: Agreed on keeping the tests as minimal as possible. This was the reason we shifted to not adding lit tests for conversion patterns. They were often not generalized and only caused friction when updating patterns.

Also, there's [a nice MLIR document](https://mlir.llvm.org/getting_started/TestingGuide/#filecheck-best-practices) describing some of the best practices for lit tests.

ramiro050 (changes requested):

vivekkhandelwal1:
This PR is not required anymore therefore closing it.

PR #2846, opened 2024-02-01
Push towards OPT onnx by fixing cumsum and adding canonicalizer for view

renxida:
view canonicalizer by @Shukla-Gaurav , onnx fixes by Rob, cumsum constant propagation fixes by  @rsuderman , some tidying up all around by me.


renxida:
@Shukla-Gaurav i think this should work!

I was able to run
```bash
# pipeline with debug

set -e
NAME=opt-125M.fp32

cd ~/SHARK-TestSuite/e2eshark/test-run/pytorch/models/opt-125M/

# step 1 import onxx to mlir
python3 -m torch_mlir.tools.import_onnx $NAME.onnx -o $NAME.onnx.mlir

# step 2 convert onnx to torch
~/torch-mlir/build/bin/torch-mlir-opt --canonicalize --mlir-print-debuginfo --convert-torch-onnx-to-torch $NAME.onnx.mlir -o ./$NAME.torch.mlir 

# step 3 lower to linalg
~/torch-mlir/build/bin/torch-mlir-opt -pass-pipeline='builtin.module(torch-backend-to-linalg-on-tensors-backend-pipeline)' --mlir-print-debuginfo $NAME.torch.mlir -o opt-125M.fp32.linalg.mlir
```
without problem.

The day's been long and I'm signing off now. When you get online, could you do Kumar's

```bash
# full pass
cd ~/SHARK-TestSuite/e2eshark
python ./run.py --upto inference --mode onnx -c ~/torch-mlir/build -i ~/iree-build --tests pytorch/models/opt-125M --hfhome ~/hfhome --verbose
```

with an IREE that's built against this PR + torch-mlir that's built against this PR?

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -927,6 +927,60 @@ OpFoldResult AtenViewOp::fold(FoldAdaptor adaptor) {
...
+
+    auto resType = op.getType().dyn_cast<BaseTensorType>();
+    if (!resType || !resType.hasSizes())
+      return failure();
+    ArrayRef<int64_t> resShape = resType.getSizes();
+    int64_t resRank = resShape.size();
+    if (inputRank >= resRank)
+      return failure();
+    SmallVector<int64_t> dimList;
+    int64_t inputPtr = 0;
```
newling, line 949: [?,?] -> [?,?,1]  isn't necessarily an unsqueeze (it could be [2,3] -> [3,2,1]). Possible easy fix: check for static shapes.

newling (commented):
What's the motivation for a ViewOp canonicalizer? I'd hope this case lowers to tensor.expand_shape without a problem already. 

p.s. I like this canonicalization (and we can add cases for collapse cases too) I'm just curious

Why are the different changes in the same PR, can item and view be in different PRs?

lib/Dialect/Torch/IR/TorchOps.cpp
renxida: This is a question for @Shukla-Gaurav but we encountered a bunch of problems with constant propagation and i'm guessting it has something to do with that.

Shukla-Gaurav: This is a cumulative patch for making the opt model work. There was an issue with this particular case of view of in the opt model, so added the canonicalizer as a temporary fix. Will raise a separate patch if it is at all necessary.

newling: Ok. I'm just wondering if it's needed, because I don't seem to be hitting the view lowering issue with the latest torch-mlir.

Shukla-Gaurav:
This is resolved now and has separate PRs merged for the fix.

PR #2852, opened 2024-02-02
[MLIR][TORCH] Modify Onnx.Reshape lowering for static shape cases

vivekkhandelwal1:
This commit modifies the OnnxToTorch lowering of Onnx.Reshape op by creating the result shape list for the aten.reshape using the result shape values inferred from the op's result shape.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -1561,6 +1561,29 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             binder.tensorResultType(resultType) ||
             binder.s64IntegerAttr(allowzero, "allowzero", 0))
           return failure();
+
+        // If the result shape is static then we can create a result shape list
+        // directly using the result shape values (integers).
+        if (resultType.hasSizes()) {
+          bool hasStaticShape = resultType.areAllSizesKnown();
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1561,6 +1561,35 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        // directly using the result shape values (integers).
+        if (resultType.hasSizes()) {
+          bool hasStaticShape = true;
+          ArrayRef<int64_t> resultShapeInt = resultType.getSizes();
+          for (int64_t dim : resultShapeInt) {
+            if (dim == Torch::kUnknownSize) {
+              hasStaticShape = false;
+              break;
+            }
+          }
```
Shukla-Gaurav, line 1575: You can use 'resultType.areAllSizesKnown()` to achieve this functionality.

Shukla-Gaurav (commented):
It seems fine to bypass the shape parameter when result type is static, but I am not sure. Maybe @rsuderman @stellaraccident can comment better.

rsuderman:
> It seems fine to bypass the shape parameter when result type is static, but I am not sure. Maybe @rsuderman @stellaraccident can comment better.

Yeah this is probably fine. If the result ty is statically known we can assume that whatever generated the code could traverse the input argument. It is a bit gross from a model construction standpoint but likely fine.

rsuderman (approved):

PR #2863, opened 2024-02-05
[torch-mlir][sparse] add sparsification mini-pipeline to linalg reference backend

aartbik:
This adds a few passes that will ensure linalg with sparse tensors are properly lowered to loops and can run using the ExecutionEngine for testing (a few details on parameter passing from PyTorch still TBD)

Running the tests as

$ ./tools/e2e_test.sh --config linalg

with the extra passes still yields

Summary:
    Passed: 1144
    Expectedly Failed: 8


yinying-lisa-li (approved):

aartbik:
Hmm, it looks like a new fail sneaked in while I was working on this.
Let me rebase and investigate....

      FAIL - "ConvolutionBackwardModule2DStrided_basic"

aartbik:
Very interesting, this case even fails on mainline MLIR! 

So for now,  I will close this PR and resent when we have an upstream fix (or downstream depending on your view ;-)

PR #2860, opened 2024-02-05
[WIP] Onnx llama7b Folders

vivekkhandelwal1:

PR #2864, opened 2024-02-05
AtenSortOp Folder

renxida:
A chunk off

https://github.com/llvm/torch-mlir/pull/2856
https://github.com/llvm/torch-mlir/pull/2860

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -1710,6 +1710,41 @@ void AtenSortIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
   });
 }
 
+
+//===----------------------------------------------------------------------===//
+// AtenSortOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult AtenSortOp::fold(FoldAdaptor adaptor,
+                               SmallVectorImpl<OpFoldResult> &results) {
+  auto operand = getSelf();
+  auto operandTy = dyn_cast<ValueTensorType>(operand.getType());
+  auto iTTy = cast<ValueTensorType>(getResult(1).getType());
+  auto indicesTy = iTTy.toBuiltinTensor().clone(iTTy.getDtype());
+
+  if (!operandTy.hasSizes())
+    return failure();
+  if (!indicesTy.hasStaticShape())
+    return failure();
+
+  bool unaryDim = false;
+  IntegerAttr dimAttr = dyn_cast_or_null<IntegerAttr>(adaptor.getDim());
+  if (dimAttr) {
+    unaryDim = operandTy.getSizes()[dimAttr.getInt()] == 1;
+  }
+
+  OpBuilder b(getContext());
+  if (unaryDim || llvm::all_of(operandTy.getSizes(),
+                               [](int64_t dim) { return dim == 1; })) {
+    results.push_back(operand);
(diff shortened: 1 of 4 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1688,6 +1688,41 @@ void AtenSortIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
 
+
+//===----------------------------------------------------------------------===//
+// AtenSortOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult AtenSortOp::fold(FoldAdaptor adaptor,
+                               SmallVectorImpl<OpFoldResult> &results) {
+  auto operand = getSelf();
+  auto operandTy = dyn_cast_if_present<ValueTensorType>(operand.getType());
```
newling, line 1702: When would operand.getType() be nullptr? I'm wondering why you're using dyn_cast_if_present and not dyn_cast.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1688,6 +1688,41 @@ void AtenSortIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+  if (!operandTy || !operandTy.hasSizes())
+    return failure();
+  
+  auto iTTy = dyn_cast_if_present<ValueTensorType>(getResult(1).getType());
+  auto indicesTy = iTTy ? iTTy.toBuiltinTensor().clone(iTTy.getDtype()) : nullptr;
```
newling: I think it'd be clearer if you check !iTTy and return failure before getting indicesTy. Are you trying to minimize the number of exit points from the method?

newling (commented):
(names) : use full names for variables when possible operandType, indicesType, etc.

lib/Dialect/Torch/IR/TorchOps.cpp
renxida: good point, using dyn_cast for this and the other one

lib/Dialect/Torch/IR/TorchOps.cpp
renxida: was thinking there should be a single line i can read to figure out why this folder could fail, but then realized that for debugging, having different lines probably makes it a lot easier

renxida:
running into a lot of trouble because of the llvm version bumps : /

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1688,6 +1692,49 @@ void AtenSortIntOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
 }
 
+//===----------------------------------------------------------------------===//
+// AtenSortOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult AtenSortOp::fold(FoldAdaptor adaptor,
+                               SmallVectorImpl<OpFoldResult> &results) {
+  auto operand = getSelf();
+  auto operandType = dyn_cast<ValueTensorType>(operand.getType());
```
newling, line 1702: I wasn't sure about asserting that the type is ValueTensorType, I asked a q here: https://discord.com/channels/636084430946959380/742573221882364009/1204378906925342750

renxida: Fixed!

renxida:
Got approval from rob on discord.

PR #2868, opened 2024-02-06
[torch] Aten index_select folder

daveliddell:
Folds aten::index_select ops under the following conditions:

1. If the input and output are the same shape, the indexing operation is a NOP, so just return the input.
2. If the input has shape <1x1x...xNx...x1> (all 1's except for one dim), and the output shape is <1x1x...x1> (all 1's), then there is a single index, so extract the single element value and return a tensor with that value.


daveliddell:
Well, I screwed up that branch royally.  Starting a new one.

PR #2869, opened 2024-02-06
[torch] AtenSliceOp folder that produces splat results

renxida:
Includes `slice` folder and lit tests

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -2501,33 +2501,63 @@ OpFoldResult AtenBroadcastToOp::fold(FoldAdaptor adaptor) {
   return getOperand(0);
 }
 
+
 //===----------------------------------------------------------------------===//
 // AtenSliceTensorOp
 //===----------------------------------------------------------------------===//
 
 OpFoldResult AtenSliceTensorOp::fold(FoldAdaptor adaptor) {
-  int64_t start, end, step;
-  if (matchPattern(getStart(), m_TorchConstantInt(&start)) &&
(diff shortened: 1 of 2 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2506,22 +2506,50 @@ OpFoldResult AtenBroadcastToOp::fold(FoldAdaptor adaptor) {
...
-      matchPattern(getStep(), m_TorchConstantInt(&step)) && step == 1 &&
-      start == 0 && end == std::numeric_limits<int64_t>::max())
+  DenseElementsAttr input =
+      dyn_cast_or_null<DenseElementsAttr>(adaptor.getSelf());
+  IntegerAttr start = dyn_cast_or_null<IntegerAttr>(adaptor.getStart());
+  IntegerAttr end = dyn_cast_or_null<IntegerAttr>(adaptor.getEnd());
+  IntegerAttr step = dyn_cast_or_null<IntegerAttr>(adaptor.getStep());
+  IntegerAttr dim = dyn_cast_or_null<IntegerAttr>(adaptor.getDim());
+
+  if (start && end && step && step.getInt() == 1 && start.getInt() == 0 &&
```
rsuderman, line 2516: We will want to get the values with `getValue()` and `getZExtValue()`. Probably worth saving to a var so we do not need to call it every time.

rsuderman: We should also probably check for negative values and correct.

rsuderman (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
renxida: Done

PR #2871, opened 2024-02-06
[torhc] aten.index_select folder

daveliddell:
Folds aten::index_select ops under the following conditions:

1. If the input and output are the same shape, the indexing operation is a NOP, so just return the input.
2. If the input has shape <1x1x...xNx...x1> (all 1's except for one dim), and the output shape is <1x1x...x1> (all 1's), then there is a single index, so extract the single element value and return a tensor with that value.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
@@ -303,7 +303,11 @@ bool potentiallyMutatesListOperands(Operation *op);
 /// the value as a signed integer, which implies that if the attribute has
 /// a 64-bit unsigned value, it will be converted to an int64_t in the manner
 /// that uint64_t is cast to int64_t in C++.
-int64_t getIntAttrAsSigned(IntegerAttr intAttr);
+inline int64_t getIntAttrAsSigned(IntegerAttr intAttr) {
+  if (intAttr.getType().isUnsignedInteger())
+    return intAttr.getValue().getZExtValue();
+  return intAttr.getValue().getSExtValue();
+}
 
 /// Returns the value from an `IntegerAttr` as an integral index.
 ///
@@ -321,7 +325,12 @@ int64_t getIntAttrAsSigned(IntegerAttr intAttr);
 ///
 /// No bounds checking is performed on the index to ensure that it is within
 /// the legal range for `dimSize`.
-int64_t getIntAttrAsIndex(IntegerAttr intAttr, int dimSize = -1);
+inline int64_t getIntAttrAsIndex(IntegerAttr intAttr, int dimSize = -1) {
+  int64_t signedIndex = getIntAttrAsSigned(intAttr);
+  if (dimSize < 0 || signedIndex > 0)
+    return signedIndex;
+  return dimSize + signedIndex; // count backwards from dimSize
+}
 
 } // namespace Torch
 } // namespace torch
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
@@ -294,6 +294,35 @@ bool isListPotentiallyMutated(Value list);
 /// the list.
 bool potentiallyMutatesListOperands(Operation *op);
 
+/// Returns the value from an `IntegerAttr` as an `int64_t`.
+///
+/// @param intAttr the `IntegerAttr` from which to extract the value
+/// @return the value as an `int64_t`
+///
+/// Regardless of the signed-ness of the attribute, this function returns
+/// the value as a signed integer, which implies that if the attribute has
+/// a 64-bit unsigned value, it will be converted to an int64_t in the manner
+/// that uint64_t is cast to int64_t in C++.
+int64_t getIntAttrAsSigned(IntegerAttr intAttr);
+
+/// Returns the value from an `IntegerAttr` as an integral index.
+///
+/// @param intAttr the `IntegerAttr` from which to extract the index
+/// @param dimSize the size of the dimension that the attribute indexes into
+/// @return the index value
+///
+/// Use this function when the given `IntegerAttr` represents an index into
+/// a range, such as an index into a tensor dimension.  If `dimSize` is given,
+/// negative index values are converted into positive vales by counting
+/// elements from the "right" side of the dimension, as in python, numpy, etc.
+/// For example, an index of -2 and a dimSize of 10 returns 8 because 8 is the
+/// 2nd index from the high end of the range 0 to 9.  If `dimSize` is not
+/// given, any negative indices are returned as negative numbers.
+///
+/// No bounds checking is performed on the index to ensure that it is within
+/// the legal range for `dimSize`.
+int64_t getIntAttrAsIndex(IntegerAttr intAttr, int dimSize = -1);
+
 } // namespace Torch
 } // namespace torch
 } // namespace mlir
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -199,6 +199,25 @@ static Value getScalarFloatValue(Value input, Location loc,
   return nullptr;
 }
 
+int64_t mlir::torch::Torch::getIntAttrAsSigned(IntegerAttr intAttr) {
+  if (intAttr.getType().isSignedInteger())
+    return intAttr.getSInt();
+  if (intAttr.getType().isUnsignedInteger())
+    return int64_t(intAttr.getUInt());
+  if (intAttr.getType().isSignlessInteger())
+    return intAttr.getInt(); // signless returns as int64_t
+  assert(false && "Unhandled integer attribute type");
+  return 0;
+}
+
+int64_t mlir::torch::Torch::getIntAttrAsIndex(IntegerAttr intAttr,
+                                              int dimSize) {
+  int64_t signedIndex = getIntAttrAsSigned(intAttr);
+  if (dimSize < 0 || signedIndex > 0)
+    return signedIndex;
+  return dimSize - (-signedIndex);
+}
+
 //===----------------------------------------------------------------------===//
 // MethodOp
 //===----------------------------------------------------------------------===//
@@ -2895,6 +2914,11 @@ OpFoldResult AtenIndexSelectOp::fold(FoldAdaptor adaptor) {
   // a single element.  Handles float and int types.
 
   int64_t dimInt = dimAttr.getInt();
+  // If the selected dim is negative, count backwards from the last dim
+  if (dimInt < 0)
+    dimInt = selfSizes.size() - (-dimInt);
+  assert(uint64_t(dimInt) < selfSizes.size() &&
+         "Selected dim > number of dims");
 
   bool scalarFold = true;
   for (int i = 0, s = selfSizes.size(); i < s; ++i) {
@@ -2905,17 +2929,12 @@ OpFoldResult AtenIndexSelectOp::fold(FoldAdaptor adaptor) {
   if (!scalarFold)
     return nullptr;
 
+  // Get the single index value for the selected dimension
   auto splatValue = indexAttr.getSplatValue<IntegerAttr>();
-  uint64_t indexInt = 0;
-  if (splatValue.getType().isSignedInteger())
-    indexInt = uint64_t(splatValue.getSInt());
-  else if (splatValue.getType().isUnsignedInteger())
-    indexInt = splatValue.getUInt();
-  else if (splatValue.getType().isSignlessInteger())
-    indexInt = uint64_t(splatValue.getInt());
-  else
-    return nullptr;
+  int64_t indexInt = getIntAttrAsIndex(splatValue, selfSizes[dimInt]);
 
+  // Extract the single constant value from the input tensor and turn the
+  // extracted value into a single-element tensor of the output shape and dtype
   auto splattr = selfAttr.getValues<Attribute>()[indexInt];
 
   auto dty = resultTy.getDtype();
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -2876,7 +2876,17 @@ OpFoldResult AtenIndexSelectOp::fold(FoldAdaptor adaptor) {
   if (!scalarFold)
     return nullptr;
 
-  auto indexInt = indexAttr.getSplatValue<IntegerAttr>().getInt();
+  auto splatValue = indexAttr.getSplatValue<IntegerAttr>();
+  uint64_t indexInt = 0;
+  if (splatValue.getType().isSignedInteger())
+    indexInt = uint64_t(splatValue.getSInt());
+  else if (splatValue.getType().isUnsignedInteger())
+    indexInt = splatValue.getUInt();
+  else if (splatValue.getType().isSignlessInteger())
+    indexInt = uint64_t(splatValue.getInt());
+  else
+    return nullptr;
+
   auto splattr = selfAttr.getValues<Attribute>()[indexInt];
 
   auto dty = resultTy.getDtype();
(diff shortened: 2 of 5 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2835,6 +2835,94 @@ OpFoldResult AtenDivIntOp::fold(FoldAdaptor adaptor) {
...
 }
 
+//===----------------------------------------------------------------------===//
+// AtenIndexSelectOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenIndexSelectOp::fold(FoldAdaptor adaptor) {
+  auto self = getSelf();
+  auto index = getIndex();
+  auto selfTy = cast<ValueTensorType>(self.getType());
```
newling, line 2845: I wasn't sure about asserting that the type is ValueTensorType, I asked a q here: https://discord.com/channels/636084430946959380/742573221882364009/1204378906925342750

daveliddell: You are right!  As it turns out, during a torch.tensor's journey to a shaped, typed torch.vtensor, the folders (and I assume everything else in canonicalization) gets called repeatedly on each step, so it appears that the correct strategy is to keep rejecting any attempts to fold until the tensors' shapes and types are nice.  The new code reflects that behavior.

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
```diff
@@ -294,6 +294,35 @@ bool isListPotentiallyMutated(Value list);
...
+/// @param intAttr the `IntegerAttr` from which to extract the value
+/// @return the value as an `int64_t`
+///
+/// Regardless of the signed-ness of the attribute, this function returns
+/// the value as a signed integer, which implies that if the attribute has
+/// a 64-bit unsigned value, it will be converted to an int64_t in the manner
+/// that uint64_t is cast to int64_t in C++.
+int64_t getIntAttrAsSigned(IntegerAttr intAttr);
+
+/// Returns the value from an `IntegerAttr` as an integral index.
```
newling, line 308: Does torch always allow indices in `[-ndim, ndim)`, does it maybe sometimes allow any value and just modulo it ndim? I don't know, but if that's the case this util might trip a developer up. 

In my PR to this (coming soon) I've slightly simplified these utils and made them private to the .cpp -- use if you like.

newling (commented):
Looks good.  I think a comment at the top summarizing the 2 cases handled, and potential future cases might could in theory be handled would be useful. When I started reviewing this I kind of expected to see a general constant folder, ie for the cases where the output is more than a single element.

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
daveliddell: No, modulo is not the behavior.  It gives an "INDICES element is out of DATA bounds" error.  Of course, the pytorch docs don't specify these things, not even about negative indices.

Thanks for the rewrite.  I can take the simplification of the casting logic.  However, you've collapsed my two levels of abstraction into one for some reason I don't understand.  The lower level, getIntAttrAsSigned, was intended for applications where you know the usage of an IntegerAttr is signed but the MLIR could be written as any signed-ness.  The upper level, getIntAttrAsIndex, is for those cases when you know that your integer is specifically a tensor index.

You've also pushed these utilities down, when I think the opposite should be the case: it should be as public as possible, even in IntegerAttr itself.  However, I think taking on a change to the MLIR API itself might cause this PR to drag on too long, so I'm hoping at least to encourage this kind of abstraction among users of TorchOps.h.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2835,6 +2854,102 @@ OpFoldResult AtenDivIntOp::fold(FoldAdaptor adaptor) {
...
 }
 
+//===----------------------------------------------------------------------===//
+// AtenIndexSelectOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenIndexSelectOp::fold(FoldAdaptor adaptor) {
+  auto self = getSelf();
+  auto index = getIndex();
+  auto selfTy = dyn_cast_or_null<ValueTensorType>(self.getType());
```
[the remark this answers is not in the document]
daveliddell: Oh, thanks for the clarification!  I will change these to dyn_cast.

test/Dialect/Torch/canonicalize.mlir
```diff
@@ -2204,3 +2204,55 @@ func.func @torch.aten.detach$canonicalize(%arg0: !torch.tensor<[1],f32>) -> !tor
...
+// CHECK-NEXT:      return %[[RES]] : !torch.vtensor<[1],si64>
+func.func @torch.aten.index_select$const_si_si() -> !torch.vtensor<[1],si64> {
+  %tensor = torch.vtensor.literal(dense<[10,20,30,40,50,60,70,80,90,100]> : tensor<10xsi64>) : !torch.vtensor<[10],si64>
+  %dim = torch.constant.int 0
+  %index = torch.vtensor.literal(dense<5> : tensor<1xsi64>) : !torch.vtensor<[1],si64>
+  %0 = torch.aten.index_select %tensor, %dim, %index : !torch.vtensor<[10],si64>, !torch.int, !torch.vtensor<[1],si64> -> !torch.vtensor<[1],si64>
+  return %0 : !torch.vtensor<[1],si64>
+}
+
+// CHECK-LABEL:   func.func @torch.aten.index_select$const_si_ui(
```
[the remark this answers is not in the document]
daveliddell: They differ in the dtype of the indices arg, thereby testing the different branches of getIntAttrAsIndex.

daveliddell (commented):
Thanks for your feedback!

daveliddell:
> Looks good. I think a comment at the top summarizing the 2 cases handled, and potential future cases might could in theory be handled would be useful. When I started reviewing this I kind of expected to see a general constant folder, ie for the cases where the output is more than a single element.

I think _all_ the code needs more extensive documentation to describe purpose and intent, at all levels of the MLIR code base, but that doesn't seem to be the way we do things.  I have at least commented in each section what they do, as it took me quite some time to figure out what they were doing.  I also agree that more folding is possible, for this folder and most others, but please keep in mind that we were targeting specific failures in certain models under a tight deadline (which has moved out).  In any case, I'm not sure relying on folders to get us through a lowering without errors is in keeping with the architectural goals of MLIR, as stated in the canonicalization docs.  For now, I intend to leave this folder as is, and perhaps revisit it when we have a more general need.

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
newling: > No, modulo is not the behavior. It gives an "INDICES element is out of DATA bounds" error. Of course, the pytorch docs don't specify these things, not even about negative indices.

 I was trying to ask if ALL torch ops in torch have this behavior, and if not then this API is not useful for the cases where it might by modulo logic. 
 
 > However, you've collapsed my two levels of abstraction into one for some reason I don't understand.
 
Personal taste I suppose, maybe I "inline heuristic" is more aggressive.

test/Dialect/Torch/canonicalize.mlir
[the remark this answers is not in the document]
newling: Oh, I missed that (obviously!)

newling (approved):
Thanks!

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -199,6 +199,25 @@ static Value getScalarFloatValue(Value input, Location loc,
...
+  assert(false && "Unhandled integer attribute type");
+  return 0;
+}
+
+int64_t mlir::torch::Torch::getIntAttrAsIndex(IntegerAttr intAttr,
+                                              int dimSize) {
+  int64_t signedIndex = getIntAttrAsSigned(intAttr);
+  if (dimSize < 0 || signedIndex > 0)
+    return signedIndex;
+  return dimSize - (-signedIndex);
```
rsuderman, line 218: Shouldn't this just be `signedIndex + dimSize`

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2835,6 +2854,102 @@ OpFoldResult AtenDivIntOp::fold(FoldAdaptor adaptor) {
...
+    return nullptr;
+
+  if (!selfTy.hasDtype() || !indexTy.hasDtype() || !resultTy.hasDtype())
+    return nullptr;
+
+  auto selfSizes = selfTy.getSizes();
+  auto indexSizes = indexTy.getSizes();
+  auto resultSizes = resultTy.getSizes();
+
+  if (selfTy.getDtype() != resultTy.getDtype())
```
rsuderman, line 2880: You might as well `||` of these checks together so we save a few lines.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2835,6 +2854,102 @@ OpFoldResult AtenDivIntOp::fold(FoldAdaptor adaptor) {
...
+OpFoldResult AtenIndexSelectOp::fold(FoldAdaptor adaptor) {
+  auto self = getSelf();
+  auto index = getIndex();
+  auto selfTy = dyn_cast_or_null<ValueTensorType>(self.getType());
+  auto indexTy = dyn_cast_or_null<ValueTensorType>(index.getType());
+  auto resultTy = dyn_cast_or_null<ValueTensorType>(getType());
+  if (!selfTy || !indexTy || !resultTy)
+    return nullptr;
+
+  if (!selfTy.hasSizes() || !indexTy.hasSizes() || !resultTy.hasSizes())
```
rsuderman, line 2870: Same with the check below, we might as well `||` them together.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2835,6 +2854,102 @@ OpFoldResult AtenDivIntOp::fold(FoldAdaptor adaptor) {
...
+
+  // If the input's dimensions are all 1 except for one dimension, and if
+  // there is a single index in the index list (as detected by the result
+  // dimension being 1), then fold to a <1x1x...x1> tensor literal containing
+  // a single element.  Handles float and int types.
+
+  int64_t dimInt = dimAttr.getInt();
+  // If the selected dim is negative, count backwards from the last dim
+  if (dimInt < 0)
+    dimInt = selfSizes.size() - (-dimInt);
```
rsuderman, line 2919: Might as well do `dimInt = selfSizes.size() + dimInt;`

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2835,6 +2854,102 @@ OpFoldResult AtenDivIntOp::fold(FoldAdaptor adaptor) {
...
+  assert(uint64_t(dimInt) < selfSizes.size() &&
+         "Selected dim > number of dims");
+
+  bool scalarFold = true;
```
rsuderman, line 2923: Rather than using a variable we can just do:
```
if ((selfSizes[i] != 1 && i != dimInt) || resultSizes[i] != 1) return nullptr;
```

rsuderman (changes requested):

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
daveliddell: > I was trying to ask if ALL torch ops in torch have this behavior, and if not then this API is not useful for the cases where it might by modulo logic.
Oh, I see.  Yeah, pytorch doesn't document at this level, which means someone would have to test all of them, and it could change at any time.  Not an ideal way to design an API.

However, I don't think the value of getIntAttrAsIndex is diminished from not having consistent API behavior, as the function itself has well-defined behavior.  That is, you use the function only when it matches what you want.  If it is discovered that there are multiple behaviors for indexes, it would be better to enhance getIntAttrAsIndex.  For example, if you have an op with modulo behavior, then perhaps a bool usesModulo arg should be added.  

I think the current technique of ad hoc implementations all over the code, some matching the behavior of pytorch and some not, is far worse.  I don't see these implementations even considering corner cases like modulo or negative indexes.  I feel that lowering the effort through the addition of the right abstractions is the best way to deal with that problem.  I don't know that I've come up with the right abstractions, but at least I want to start that conversation.

> Personal taste I suppose, maybe I "inline heuristic" is more aggressive.
Inlining the implemenation would be better; I'll do that.  However, I'm hoping to turn your attention away from the style of implementation and toward my point that there are two levels of abstractions that would be useful.  Perhaps you disagree that the lower level is useful, and in fact, I don't personally have a second use case for it yet.

I'll have to admit that my efforts to encourage others to introduce levels of abstraction for higher quality and better functional coverage have largely failed over the decades.  I introduce the abstractions and write code to demonstrate how using those abstractions raises the level at which one implements, but everyone still continues to code at the level of the underlying API.  It doesn't stop me from trying, though!  I believe my difficulty is in getting others to recognize the opportunity for abstraction in their own code; it takes a lot of mentoring, which I could really only do with my own staff.

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: Yes, they're identical, of course.  I thought I was making it clearer what the intent was.  I was hoping to illustrate how negative indexes work by counting backwards from dimSize by abs(signedIndex).  If you think that's overkill, I can change to a +.  Personally, if I were looking at someone else's code, I'd find it more challenging to figure out where `signedIndex + dimSize` was coming from, especially since we don't comment at this level.

In light of this explanation, please let me know if you'd rather see the + anyway.

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: Ok, will do.

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: Actually, never mind.  Words are always clearer.  I'll rewrite the expression your way and say what it is. :-)

lib/Dialect/Torch/IR/TorchOps.cpp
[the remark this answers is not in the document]
daveliddell: done

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
daveliddell: done according to offline agreement

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: done

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: done

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: done

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: done

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: done

lib/Dialect/Torch/IR/TorchOps.cpp
rsuderman: Did you push these changes? I am not seeing the if statements merged.

rsuderman (approved):

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: My bad!  Got confused by multiple blocks of ifs.  We had a chat about the new fix offline, so resolving now.  Thanks for all your help!

PR #1384, opened 2022-09-19
Add Byte type support

AmosLewis:
[https://github.com/llvm/torch-mlir/issues/1383](https://github.com/llvm/torch-mlir/issues/1383)

PR #2873, opened 2024-02-06
Implement lowering of torch.aten.all.dim

mmakevic:
Lowering of torch.aten.all.dim to linalg.

Per PyTorch documentation:

> This function matches the behaviour of NumPy in returning output of dtype bool for all supported dtypes except uint8. For uint8 the dtype of output is uint8 itself.

Since there is no support for ui8 in torch-mlir currently (https://github.com/llvm/torch-mlir/pull/1384#issuecomment-1260011334) implementation returns failure for that case.


newling (commented):
Looks good but I think the tests can be more targeted (smaller tensor, set values by hand). Can you test the empty tensor case returns 'True' ?

mmakevic:
@newling done :)

lib/Conversion/TorchToLinalg/Reduction.cpp
```diff
@@ -277,6 +277,10 @@ static Value createInitElementForReduceOp(OpBuilder &b, Location loc,
   if (isa<AtenLinalgVectorNormOp>(op) || isa<AtenFrobeniusNormDimOp>(op))
     return b.create<arith::ConstantOp>(loc, b.getZeroAttr(elementType));
 
+  if (auto all = dyn_cast<AtenAllDimOp>(op)) {
```
newling, line 280: nit : use isa<AtenAllDimOp> 

to avoid compiler warnings

lib/Conversion/TorchToLinalg/Reduction.cpp
```diff
@@ -357,6 +361,11 @@ static Value createLinalgPayloadForReduceOp(OpBuilder &b, Location loc,
     auto ord = b.create<arith::ConstantOp>(loc, twoAttr);
     auto pow = b.create<math::PowFOp>(loc, abs, ord);
     return b.create<arith::AddFOp>(loc, pow, result);
+  } else if (auto allOp = dyn_cast<AtenAllDimOp>(op)) {
```
newling, line 364: ` isa<AtenAllDimOp>(op)`

newling (changes requested):
Thanks for updating the tests, just the isa<.> comment remaining

lib/Conversion/TorchToLinalg/Reduction.cpp
mmakevic: Fixed

lib/Conversion/TorchToLinalg/Reduction.cpp
mmakevic: Fixed as well

lib/Conversion/TorchToLinalg/Reduction.cpp
```diff
@@ -535,6 +547,9 @@ class ConvertReductionOp : public ConversionPattern {
         !elemType.isa<mlir::FloatType>())
       return rewriter.notifyMatchFailure(
           op, "only float types are valid for vector norm ops");
+    if((isa<AtenAllDimOp>(op)) && elemType.isa<mlir::IntegerType>() &&
```
newling, line 550: unnecessary outer braces (and formatting looks weird to me)

newling (approved):
Just the little comment about the extra brackets and formatting, but approving now as it looks good to me otherwise.

lib/Conversion/TorchToLinalg/Reduction.cpp
mmakevic: Missed that, fixed

newling:
My view of the PR:

image

@mmakevic  Let me know if this looks correct ("authored by" email look ok?) and I'll merge it if no one else comments in the next day.

mmakevic:
@newling Yeah, everything looks okay. Thanks!

PR #2875, opened 2024-02-06
Add lowering support for math::AbsIOp

aviator19941:
There is no lowering support for math::AbsIOp, so if the operand is an integer type, it will fail to lower to math::AbsFOp since the op operand #0 must be floating-point-like.


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -424,8 +424,11 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
         b.create<arith::ConstantOp>(loc, b.getFloatAttr(floatDtype, 0));
     return createEqual(b, loc, floatDtype, self, zero);
   }
-  if (isa<AtenAbsOp>(op))
+  if (isa<AtenAbsOp>(op)) {
+    if (payloadArgs[0].getType().isa<IntegerType>())
+      return b.create<math::AbsIOp>(loc, payloadArgs[0]);
     return b.create<math::AbsFOp>(loc, payloadArgs[0]);
+  }
```
vivekkhandelwal1, line 431: Are you sure that apart from integer or float it doesn't take any other element type?

aviator19941: I am not sure, but I will check. All I know is that before it was only lowering to AbsFOp.

aviator19941: MLIR only has math::AbsIOp/AbsFOp here https://mlir.llvm.org/docs/Dialects/MathOps/, so it should only take integer or float element types.

newling (commented):
LGTM, just the small test comment

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -2127,14 +2127,36 @@ def forward(self, a):
...
+        None,
+        ([-1, -1, -1], torch.int64, True),
+    ])
+    def forward(self, a):
+        return torch.abs(a)
+
+
+@register_test_case(module_factory=lambda: ElementwiseAbsIntModule())
+def ElementwiseAbsIntModule_basic(module, tu: TestUtils):
+    module.forward(tu.randint(3, 4, 5, low=-1, high=1))
```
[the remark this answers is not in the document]
aviator19941: Yeah that makes sense, I changed the tests to specify one negative value always.

Issue #2866, opened 2024-02-05
Strip torch.onnx_meta attributes after convert-torch-onnx-to-torch


ScottTodd:
Downstream in IREE (https://github.com/openxla/iree/issues/16316) we have some logic expecting that a "converted program" has no remaining input-dialect-specific ops/attributes/etc.

When I run a program like this through `iree-opt --convert-torch-onnx-to-torch onnx_input.mlir` (this should be the same as `torch-opt` or equivalent):

```mlir
func.func @simple_add_onnx(%arg0: !torch.vtensor<[],si64>, %arg1: !torch.vtensor<[],si64>) -> !torch.vtensor<[],si64>
    attributes {
  torch.onnx_meta.ir_version = 8 : si64,
  torch.onnx_meta.opset_version = 17 : si64,
  torch.onnx_meta.producer_name = "pytorch",
  torch.onnx_meta.producer_version = "2.1.0"
}
{
  %0 = torch.operator "onnx.Add"(%arg0, %arg1) : (!torch.vtensor<[],si64>, !torch.vtensor<[],si64>) -> !torch.vtensor<[],si64>
  return %0 : !torch.vtensor<[],si64>
}
```

I get this output:

```mlir
module {
  func.func @simple_add_onnx(%arg0: !torch.vtensor<[],si64>, %arg1: !torch.vtensor<[],si64>) -> !torch.vtensor<[],si64> attributes {torch.onnx_meta.ir_version = 8 : si64, torch.onnx_meta.opset_version = 17 : si64, torch.onnx_meta.producer_name = "pytorch", torch.onnx_meta.producer_version = "2.1.0"} {
    %int1 = torch.constant.int 1
    %0 = torch.aten.add.Tensor %arg0, %arg1, %int1 : !torch.vtensor<[],si64>, !torch.vtensor<[],si64>, !torch.int -> !torch.vtensor<[],si64>
    return %0 : !torch.vtensor<[],si64>
  }
}
```

I would expect for those `torch.onnx_meta.` attributes to be removed.

Are the attributes being used after conversion by any other code / tools / users / etc.? We could add a cleanup pass downstream in IREE, if this isn't just an oversight in torch-mlir.

ScottTodd:
We discussed this a bit over on [IREE's Discord](https://discord.com/channels/689900678990135345/908461811026767952/1204184905920028692).

[...]

PR #2876, opened 2024-02-06
Drop torch attributes at the end of backend conversion.

ScottTodd:
Fixes https://github.com/llvm/torch-mlir/issues/2866

Some backends / downstream projects expect that a "fully converted" program has no remaining ops or attributes from the original dialect(s).


lib/Dialect/TorchConversion/Transforms/BackendTypeConversionPasses.cpp
```diff
@@ -115,6 +115,15 @@ static void setupFinalization(ConversionTarget &target,
   setupFinalization<OpTy2, OpTys...>(target, patterns, typeConverter);
 }
 
+static void stripTorchAttrs(func::FuncOp func) {
+  SmallVector<NamedAttribute> newAttrs;
+  for (auto attr : func->getDialectAttrs()) {
```
ScottTodd, line 120: I'm not very familiar with how torch-mlir uses attributes. This could also look through `getAllArgAttrs()` and `getAllResultAttrs()`.

lib/Dialect/TorchConversion/Transforms/BackendTypeConversionPasses.cpp
```diff
@@ -151,6 +160,9 @@ struct FinalizingBackendTypeConversionPass
 
     if (failed(applyFullConversion(func, target, std::move(patterns))))
       signalPassFailure();
+
+    // Drop attributes that are no longer used after conversion out of Torch.
+    stripTorchAttrs(func);
```
ScottTodd, line 165: Are there users of torch-mlir that would miss these attributes after conversion? I could make this optional (via a separate pass, an option here, etc.)

test/Conversion/TorchToLinalg/basic.mlir
```diff
@@ -30,7 +30,7 @@ func.func @torch.aten.mm$basic(%arg0: !torch.vtensor<[?,?],f32>, %arg1: !torch.v
 // -----
 
 // CHECK-LABEL: func.func @torch.aten.matmul.2d
-func.func @torch.aten.matmul.2d(%arg0: !torch.vtensor<[8,16],f32>, %arg1: !torch.vtensor<[16,8],f32>) -> !torch.vtensor<[8,8],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 13 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
+func.func @torch.aten.matmul.2d(%arg0: !torch.vtensor<[8,16],f32>, %arg1: !torch.vtensor<[16,8],f32>) -> !torch.vtensor<[8,8],f32> {
```
ScottTodd, line 33: This diff isn't directly related to this change, but I felt like including it (the attributes are irrelevant to the test case). Could revert or move to another PR.

lib/Dialect/TorchConversion/Transforms/BackendTypeConversionPasses.cpp
stellaraccident: It is relatively expensive to mutate these attributes. Let's make it only pay that cost if there are attributes to change. Can you carry a `bool modified` outside the loop to note if you dropped any and then only setDialectAttrs if modified?

lib/Dialect/TorchConversion/Transforms/BackendTypeConversionPasses.cpp
stellaraccident: I don't think so. Historically there haven't been any/many torch dialect attributes used. I suspect that we just added some and forgot to make the conversion "full".

test/Conversion/TorchToLinalg/basic.mlir
stellaraccident: Makes sense to me. Fine to include.

stellaraccident (approved):
I'd like to see the conditional modification I note in the comments. Otherwise, lgtm.

lib/Dialect/TorchConversion/Transforms/BackendTypeConversionPasses.cpp
ScottTodd: Added a modified check and just a comment about how "arg" and "result" attrs could also be stripped if needed.

PR #2878, opened 2024-02-07
[torch] Folders for `torch.aten.*.tensor` operators [add, sub, mul]

rsuderman:
Simple folder for limited size aten tensor operations. This is primarily
useful for shape computation folding as they unfortunately can use
`aten` operators. Add, sub, mul are common examples of these folders.


daveliddell (approved):
As mentioned offline, as part of a bigger effort, I wouldn't mind seeing the duplicated handling of dtypes brought under some kind of unified umbrella, not specifically for this PR but in general.

Nice set of test cases for so many different possibilities!

PR #2879, opened 2024-02-07
add support for decomposition

saienduri:
This commit adds decomposition support into the core aten operators before importing the module from torch. 

Also, this commit deals with the lifted tensor constants in torch.export.export(). We don't want to add unnecessary placeholder nodes in the graph (extra args in the block module), and should treat them like the constants that they are. The unnecessary clone is also removed for max efficiency.


python/torch_mlir/extras/decomp_util.py
```diff
@@ -0,0 +1,51 @@
...
+    torch.ops.aten.nll_loss_backward,
+    torch.ops.aten._to_copy,
+    torch.ops.aten._log_softmax_backward_data,
+    torch.ops.aten.lift_fresh_copy.default,
+    torch.ops.aten._unsafe_index.Tensor,
+]
+
+def get_decomposition_table():
+    return get_decompositions(DEFAULT_DECOMPOSITIONS)
+
```
stellaraccident, line 51: nl at end

python/torch_mlir/extras/fx_importer.py
```diff
@@ -368,6 +379,12 @@ def import_frozen_exported_program(self, prog: torch.export.ExportedProgram):
                 replacement = arg_replacements.get(node.name)
                 if replacement is None:
                     continue
+                # remove the clone op (unnecessary extra copy due to torch.tensor())
+                # this is only for efficiency (no functional changes)
+                keys = list(node.users.keys())
```
stellaraccident, line 384: Can I understand why this is needed here? Every ad-hoc thing we add here is debt that we carry forward. I wouldn't expect this to have any impact on the compiled output. What am I not seeing?

saienduri: This is specifically related to the torch.tensor() operation (called a lot in many of our models). When a user calls this torch op, in the torch.export.export stage it is getting lowered into a torch.aten.clone operation. It is simply cloning a tensor that is passed in as a placeholder node. With the other changes in this PR (replacing the tensor placeholder with the tensor constant), we deal with the tensor argument and turn it into a constant torch.literal.vtensor. The clone is an unnecessary operation that is introduced in the export stage. Wherever the cloned tensor is being used, we can simply use our constant. The thought process here was to eliminate the unneccesary clone operation that happens everytime torch.tensor() is called and just use the constant. I thought this was the right place for it, because it is something that happens in the export stage, so we should take care of it here (preproccesing) before continuing with the import. Do you think that this helps in overall efficiency and should we include it?

python/CMakeLists.txt
```diff
@@ -44,6 +44,7 @@ declare_mlir_python_sources(TorchMLIRPythonSources.PublicAPI
   ADD_TO_PARENT TorchMLIRPythonSources
   SOURCES
     fx.py
+    extras/decomp_util.py
```
stellaraccident, line 47: The "extras" namespace is global across a number of compilers. Need to fully qualify things in it. Please rename to `fx_decomp_util.py`.

python/torch_mlir/extras/fx_importer.py
stellaraccident: It's more of a layering thing: The importer should be rendering as faithful a translation as possible with limited "smarts". Almost no one will look at the importer to try to debug a transformation decision, and the importer is aiming for "correct", not "maximally efficient".

Transformations belong in the IR themselves, and I expect that clone ops on vtensors should just have a folder that elides them. Then it helps everyone, not just this special case.

If it isn't actually breaking correctness, let's leave such a transformation out of the importer.

stellaraccident (changes requested):

python/torch_mlir/extras/fx_importer.py
saienduri: Ah okay, that makes sense. Will leave it out the importer then. Especially if it gets elided, that's great :)

stellaraccident (approved):

PR #2883, opened 2024-02-07
GridSamplerOp TorchToLinalg lowering

afalkenberg1:
This is a GridSamplerOp which provides TorchToLinalg lowering. 

It is based on the following python code:

import numpy
import copy

inp = numpy.array([[[[7,1,2,3],[0,9,8,99.],[4.,16.,32.,67.],[8,7,9,11],[13,153,112,19]]]])
grid = numpy.array([[[[-1.0, -1],[-1,-0.5],[-1 ,0],[-1,1],\
                      [-0.5,-1],[-0.5,-0.5],[-0.5, 0],[-0.5,1],[0,-1],\
                      [0,-0.5],[0,0],[0,1],[1,-1],[1,-0.5],[1,0],[1,1]]]]).astype(float)

gridShape = numpy.array(grid.shape)
inpShape  = numpy.array(inp.shape)
values    = numpy.empty((grid.shape[0], grid.shape[1],grid.shape[2],4))
delta     = numpy.empty((grid.shape[0], grid.shape[1],grid.shape[2],4))
innerDim  = inpShape[-2:]
gridPlusOne      = grid + 1.0
innerDimMOne     = innerDim - 1
innerDimDivTwo   = innerDimMOne / 2.0
sGrid = gridPlusOne * innerDimDivTwo
lower = numpy.floor(sGrid)

upper = lower + 1
upperCheck = upper > innerDimMOne
upperCheckX = upperCheck[:,:,:,0]
upperCheckY = upperCheck[:,:,:,1]
upperChecked = numpy.where(upperCheck, lower, upper) 

res00  = inp[:,:,lower[0,0,:,0].astype(int),lower[0,0,:,1].astype(int)]
res01c = inp[:,:,lower[0,0,:,0].astype(int),upperChecked[0,0,:,1].astype(int)]
res01 = numpy.where(upperCheckY, 0  , res01c)
res10c = inp[:,:,upperChecked[0,0,:,0].astype(int),lower[0,0,:,1].astype(int)]
res10 = numpy.where(upperCheckX, 0  , res10c)

res11c = inp[:,:,upperChecked[0,0,:,0].astype(int),upperChecked[0,0,:,1].astype(int)]
res11ca = numpy.where(upperCheckX, 0  , res11c)
res11 = numpy.where(upperCheckY, 0  , res11ca)

values[:,:,:,3] = res00[:,:,:]
values[:,:,:,1] = res01[:,:,:]
values[:,:,:,2] = res10[:,:,:]
values[:,:,:,0] = res11[:,:,:]

d = sGrid - lower
dx = d[:,:,:,0]
dy = d[:,:,:,1]

d_1 = 1 - d
dx_1 = 1 - dx
dy_1 = 1 - dy

delta[:,:,:,0] = dx * dy  #dx * dy
delta[:,:,:,1] = dx_1 * dy # 1-dx * dy
delta[:,:,:,2] = dx * dy_1 #dx * (1-dy)
delta[:,:,:,3] = dx_1 * dy_1 # 1-dx * 1-dy 

resm = values * delta
print(resm)
print(numpy.sum(resm,3))


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2336,6 +2336,305 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+              b.create<arith::SelectOp>(loc, notValid_0, lower_0, upper_0);
+          Value upperValid_1 =
+              b.create<arith::SelectOp>(loc, notValid_1, lower_1, upper_1);
+          b.create<linalg::YieldOp>(loc, ValueRange{result_0, result_1, lower_0,
+                                                    lower_1, upperValid_0,
+                                                    upperValid_1, notValid_0,
+                                                    notValid_1});
+        });
+
+    Value gridResult_0 = rewriter.create<tensor::CastOp>(
```
rsuderman, line 2465: You should not require casts here. If you are something is going wrong with how you are selecting the expected types. Everything between linalg ops should be shape compatible.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2336,6 +2336,305 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+        ValueRange{},
+        ValueRange{deltaEmpty_00, deltaEmpty_01, deltaEmpty_10, deltaEmpty_11},
+        deltaMaps, deltaIterators,
+        [&](OpBuilder &b, Location loc, ValueRange args) {
+          Value N = b.create<linalg::IndexOp>(loc, 0);
+          Value H = b.create<linalg::IndexOp>(loc, 2);
+          Value W = b.create<linalg::IndexOp>(loc, 3);
+          Value zeroIndex = rewriter.create<arith::ConstantIndexOp>(loc, 0);
+          SmallVector<Value> gridIndex{N, H, W, zeroIndex};
+          Value lw_0 = b.create<tensor::ExtractOp>(loc, lower_0, gridIndex);
```
rsuderman, line 2509: Check through how linalg affine maps work. You should not be using extracts for the majority of these cases as you are just dropping a dimension. Also, given this is is purely an elementwise operation you can likely merge this math into the previous / following linalg generic.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2336,6 +2336,305 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+          Value dm1 = b.create<arith::SubFOp>(loc, oneFloat, d1);
+          Value res_00 = b.create<arith::MulFOp>(loc, d0, d1);
+          Value res_01 = b.create<arith::MulFOp>(loc, dm0, d1);
+          Value res_10 = b.create<arith::MulFOp>(loc, d0, dm1);
+          Value res_11 = b.create<arith::MulFOp>(loc, dm0, dm1);
+          b.create<linalg::YieldOp>(loc,
+                                    ValueRange{res_00, res_01, res_10, res_11});
+        });
+
+    Value delta_00 = rewriter.create<tensor::CastOp>(
```
rsuderman, line 2531: Ditto on the casts. If you are inserting them something is wrong.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2336,6 +2336,305 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+                  SmallVector<Value> inputIndex11{N, C, up_0a, up_1a};
+                  Value result_11 = rewriter.create<tensor::ExtractOp>(
+                      loc, input, inputIndex11);
+                  Value result_11a = rewriter.create<arith::SelectOp>(
+                      loc, nv_0, zeroVal, result_11);
+                  Value result_11b = rewriter.create<arith::SelectOp>(
+                      loc, nv_1, zeroVal, result_11a);
+
+                  Value d_00 =
+                      b.create<tensor::ExtractOp>(loc, delta_00, gridIndex_0);
```
rsuderman, line 2603: Again, these extractions should be performed via the `linalg.Generic`'s operands.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2336,6 +2336,305 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenGridSamplerOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    // calculate input dimensions
+    Value input = adaptor.getInput();
+    auto inputType = input.getType().cast<RankedTensorType>();
+    auto inputShape = inputType.getShape();
+    Value innerDim_0a =
```
rsuderman, line 2351: A bunch of the math below is materializing static values then using `arith` do constant propagation. For initially support require that the input types have static shapes (`RankedTensorType` has a `hasStaticShape()` function). Then we can worry about dynamic shapes in the future. this way we can shorten this initially.

This should substantially simplify this block initially.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2336,6 +2336,305 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+    Value lowerEmpty_0 =
+        rewriter.create<tensor::EmptyOp>(loc, gridShapeExtracted, int64type);
+    Value lowerEmpty_1 =
+        rewriter.create<tensor::EmptyOp>(loc, gridShapeExtracted, int64type);
+    Value upperEmpty_0 =
+        rewriter.create<tensor::EmptyOp>(loc, gridShapeExtracted, int64type);
+    Value upperEmpty_1 =
+        rewriter.create<tensor::EmptyOp>(loc, gridShapeExtracted, int64type);
+    Type int1type = mlir::IntegerType::get(op->getContext(), 1);
+    Value notValidEmpty_0 =
```
rsuderman, line 2412: You won't need to materialize the not valid tensor. You can just check the produced values.

rsuderman (changes requested):
You should be able to write this as a single `linalg.generic`. Going to follow up in a comment with a simplified example of what it would look like.

rsuderman:
```
#map0 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2)>
#map1 = affine_map<(d0, d1, d2, d3) -> (d0, d1, d2, d3)>

%c4 = arith.constant 4 : index
%c16 = arith.constant 16 : index
%slicew = tensor.extract_slice %t[%c0, %c0, %c0, %c1] [%c4, %c16, %c16, %c1] [%c1, %c1, %c1, %c1] : tensor<4x16x16x1xf32> to tensor<4x16x16x1xf32>
%sliceh = tensor.extract_slice %t[%c0, %c0, %c0, %c0] [%c4, %c16, %c16, %c1] [%c1, %c1, %c1, %c1] : tensor<4x16x16x1xf32> to tensor<4x16x16x1xf32>
%collapseh = tensor.collapse_shape %sliceh [0, 1, [2,3]] : tensor<4x16x16x1xf32> into tensor<4x16x16x1xf32>
%collapsew = tensor.collapse_shape %slicew [0, 1, [2,3]] : tensor<4x16x16x1xf32> into tensor<4x16x16x1xf32>
%empty = tensor.empty() : tensor<4x16x16x8xf32>
%4, %5 = linalg.generic {
    indexing_maps = [#map0, #map1],
    iterator_types = ["parallel", "parallel", "parallel", "parallel"]
  } ins(%sliceh : tensor<8x16x16xf32>, %slicew : tensor<8x16x16xf32>) 
    outs(%empty : tensor<8x16x16x8xf32>) {
  ^bb0(%arg2 : f32, %arg3 : f32):
  %floorw = math.floor(%arg2)
  %floorh = math.floor(%arg3)
  n = linalg.index(0);
  c = linalg.index(3);
  %extract = tensor.extract(%input, %n, %floorw, %floorh, %c)
  return %extract : f32
}
```

lib/Conversion/TorchToLinalg/Uncategorized.cpp
afalkenberg1: If a coordinate is not valid how do I check if the produced value is valid or not? 
// I guess you are saying : 
a = inp[x,y]
if a is not valid 
    a = 0

lib/Conversion/TorchToLinalg/Uncategorized.cpp
afalkenberg1: done

lib/Conversion/TorchToLinalg/Uncategorized.cpp
afalkenberg1: done

lib/Conversion/TorchToLinalg/Uncategorized.cpp
afalkenberg1: probably need to check this one again

lib/Conversion/TorchToLinalg/Uncategorized.cpp
afalkenberg1: combined this within one larger linalg::genericop

afalkenberg1:
I checked in an update which pretty much resolves all concerns; I have only one generiOp now and therefore the implementation is considerably shorter ; probably I need a but there is still one odd thing happening which I need to resolve.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+    Value sGrid =
+        rewriter
+            .create<linalg::GenericOp>(
+                loc, resultFinalType,
+                ValueRange{gridCollapsed_0, gridCollapsed_1}, resultFinal,
+                gridMaps, gridIterators,
+                [&](OpBuilder &b, Location loc, ValueRange args) {
+                  Value gr_0 = args[0];
+                  Value gr_1 = args[1];
+                  Value oneFloat = b.create<arith::ConstantOp>(
```
rsuderman, line 2462: Probably group above with constants. Tends to clean up the IR a bit.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+                      b.create<arith::FPToSIOp>(loc, int64type, result_0);
+                  Value lower_1 =
+                      b.create<arith::FPToSIOp>(loc, int64type, result_1);
+                  Value oneInt = b.create<arith::ConstantOp>(
+                      loc, b.getIntegerAttr(int64type, 1));
+                  Value upper_0 =
+                      b.create<arith::AddIOp>(loc, int64type, lower_0, oneInt);
+                  Value upper_1 =
+                      b.create<arith::AddIOp>(loc, int64type, lower_1, oneInt);
+                  Value notValid_0 = createGreaterThan(b, loc, int64type,
```
rsuderman, line 2480: You need to pass in either a signed or unsigned `int64Type`, specifically the signedness of the grid sample index element type. This is so that the GT operator knows whether to make a signed or unsigned version. This is why the presubmits are failing.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+                      loc, notValid_1, lower_1, upper_1);
+                  Value lw_0 = b.create<arith::IndexCastOp>(
+                      loc, b.getIndexType(), lower_0);
+                  Value lw_1 = b.create<arith::IndexCastOp>(
+                      loc, b.getIndexType(), lower_1);
+                  Value up_0 = b.create<arith::IndexCastOp>(
+                      loc, b.getIndexType(), upperValid_0);
+                  Value up_1 = b.create<arith::IndexCastOp>(
+                      loc, b.getIndexType(), upperValid_1);
+                  Value N = b.create<linalg::IndexOp>(loc, 0);
```
rsuderman, line 2496: Move these to the top of the linalg.generic body. Anything constant like is good to declare at the beginning.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+                  Value C = b.create<linalg::IndexOp>(loc, 1);
+                  SmallVector<Value> inputIndex00{N, C, lw_0, lw_1};
```
rsuderman, line 2498: Note that you have this pattern of `SmallVectora<int64_t>` and extract. You can make a helper lambda that bundles them both together and invoke it 4 times. It is about the same line count but can conceptually group operations.

rsuderman (changes requested):
This is a lot better. Now we get to focus on cleanness and some overall fixes.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenGridSamplerOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    // calculate input dimensions
+    Value input = adaptor.getInput();
+    auto inputType = input.getType().cast<RankedTensorType>();
+    auto inputShape = inputType.getShape();
+    Value innerDim_0a =
```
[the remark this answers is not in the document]
afalkenberg1: ok done ; but want to still keep the numbers i.e. innerDim0a

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+                      b.create<tensor::ExtractOp>(loc, input, inputIndex11);
+                  Value result_11a = b.create<arith::SelectOp>(
+                      loc, notValid_0, zeroVal, result_11);
+                  Value result_11b = b.create<arith::SelectOp>(
+                      loc, notValid_1, zeroVal, result_11a);
+                  Value lw_0a =
+                      b.create<arith::SIToFPOp>(loc, floatType, lower_0);
+                  Value lw_1a =
+                      b.create<arith::SIToFPOp>(loc, floatType, lower_1);
+                  Value d0 = b.create<arith::SubFOp>(loc, result_0, lw_0a);
```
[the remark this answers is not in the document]
afalkenberg1: yes it is interleaving which I already tried to optimize ; but I think if I leave the original lerp function in tact then I can call it 3 times and be done. interpolation function : x0 * dx + x1 * (1-dx) 
We call it for lerp(lerp(x00, x01),lerp(x10,x11)) something like that.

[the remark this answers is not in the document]
afalkenberg1: actually it is the other way round x0(1-dx)+x1*dx

lib/Conversion/TorchToLinalg/Uncategorized.cpp
afalkenberg1: done

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+    auto gridCollapsed_0 = rewriter.create<tensor::CollapseShapeOp>(
+        loc, gridTypeCollapsed, grid_0, associations);
+    auto gridCollapsed_1 = rewriter.create<tensor::CollapseShapeOp>(
+        loc, gridTypeCollapsed, grid_1, associations);
+    AffineExpr d0 = getAffineDimExpr(0, op->getContext());
+    AffineExpr d2 = getAffineDimExpr(2, op->getContext());
+    AffineExpr d3 = getAffineDimExpr(3, op->getContext());
+    SmallVector<AffineMap> gridMaps{
+        rewriter.getMultiDimIdentityMap(gridRank),
+        AffineMap::get(4, 0, {d0, d2, d3}, op->getContext()),
```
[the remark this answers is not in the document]
afalkenberg1: still trying to understand how to define affine maps correctly

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+                      loc, resultScaled_10, resultScaled_11);
+                  Value resultScaled = b.create<arith::AddFOp>(
+                      loc, resultScaled_0, resultScaled_1);
+                  b.create<linalg::YieldOp>(loc, resultScaled);
+                })
+            .getResult(0);
+
+    RankedTensorType resultType =
+        getTypeConverter()->convertType(op.getType()).cast<RankedTensorType>();
+    sGrid.dump();
```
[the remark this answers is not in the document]
afalkenberg1: yes I kept it in there for now

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+  LogicalResult
+  matchAndRewrite(AtenGridSamplerOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    // calculate input dimensions
+    Value input = adaptor.getInput();
+    auto inputType = input.getType().cast<RankedTensorType>();
+    auto inputShape = inputType.getShape();
+    Value innerDim_0a =
+        rewriter.create<arith::ConstantIndexOp>(loc, inputShape[2]);
```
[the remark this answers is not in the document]
afalkenberg1: used tensor::DimOp

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+    Value innerDim_1a =
+        rewriter.create<arith::ConstantIndexOp>(loc, inputShape[3]);
+    Value oneIndex = rewriter.create<arith::ConstantIndexOp>(loc, 1);
+    Value innerDim_0b =
+        rewriter.create<arith::SubIOp>(loc, innerDim_0a, oneIndex);
+    Value innerDim_1b =
+        rewriter.create<arith::SubIOp>(loc, innerDim_1a, oneIndex);
+    Type int64type = mlir::IntegerType::get(op->getContext(), 64);
```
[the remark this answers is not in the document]
afalkenberg1: done

lib/Conversion/TorchToLinalg/Uncategorized.cpp
afalkenberg1: yes I pushed constants to the top

lib/Conversion/TorchToLinalg/Uncategorized.cpp
afalkenberg1: need some enlightenment here ;  in the moment I dont see what I am missing

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2362,6 +2362,200 @@ class ConvertCastEquivalentOp : public OpConversionPattern<OpTy> {
...
+    Value innerDim_0c =
+        rewriter.create<arith::IndexCastOp>(loc, int64type, innerDim_0b);
+    Value innerDim_1c =
+        rewriter.create<arith::IndexCastOp>(loc, int64type, innerDim_1b);
+    Type floatType = mlir::FloatType::getF32(op->getContext());
```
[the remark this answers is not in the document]
afalkenberg1: done

lib/Conversion/TorchToLinalg/Uncategorized.cpp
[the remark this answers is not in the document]
afalkenberg1: done

test/Conversion/TorchToLinalg/gridsampler.mlir
```diff
@@ -0,0 +1,68 @@
+// RUN: torch-mlir-opt <%s -convert-torch-to-linalg -split-input-file -verify-diagnostics | FileCheck %s
+
+// CHECK: #map
+// CHECK-LABEL: func @grid_sampler
+// CHECK:  %0 = torch_c.to_builtin_tensor %arg0 : !torch.vtensor<[4,10,10,4],f32> -> tensor<4x10x10x4xf32>
```
rsuderman, line 5: You will want to check how to do regex matching. Shortened example below but it lets us do `CHECK-DAG` instead of `CHECK`, removing the ordering restrictions:
```
// CHECK-LABEL: func @grid_sampler
// CHECK:  %[[ARG0:.+]] = torch_c.to_builtin_tensor %arg0 : !torch.vtensor<[4,10,10,4],f32> -> tensor<4x10x10x4xf32>
// CHECK:  %[[C2:.+]] = arith.constant 2 : index
// CHECK:  %[[DIM2:.+]] = tensor.dim %0, %c2_3 : tensor<4x10x10x4xf32>
// CHECK:  %[[C3:.+]] = arith.constant 3 : index
// CHECK:  %[[DIM3:.+]] = tensor.dim %0, %c3 : tensor<4x10x10x4xf32>
```

You can reuse names sometimes if values are only used in limited scope but it avoids MLIR core changes from breaking your tests. E.g. code reordering is technically allowed between versions.

rsuderman (changes requested):
Awesome, this is way better. Only update is for the tests to use regex captures so we can avoid MLIR reordering the IR causing failures.

rsuderman (approved):

PR #2890, opened 2024-02-09
[torch] Folder for torch.aten.select.int for splat cases

rsuderman:
If the input or result is a splat value we can just constant fold the
result. This is common for shape computations and can help with shape
inference.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -1328,6 +1328,41 @@ void AtenSizeOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
 }
 
 //===----------------------------------------------------------------------===//
+// AtenSelectIntOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenSelectIntOp::fold(FoldAdaptor adaptor) {
+  auto self = dyn_cast_or_null<DenseElementsAttr>(adaptor.getSelf());
+  auto ty = cast<ValueTensorType>(getType());
+  if (!self || !ty.hasDtype() || !ty.hasSizes())
+    return nullptr;
+
+  auto selfTy = cast<ShapedType>(self.getType());
+  auto bty = ty.toBuiltinTensor().clone(ty.getDtype());
+  if (!bty.hasStaticShape())
+    return nullptr;
+
+  if (self.isSplat())
+    return DenseElementsAttr::get(bty, self.getSplatValue<Attribute>());
+
+  auto dimAttr = dyn_cast_or_null<IntegerAttr>(adaptor.getDim());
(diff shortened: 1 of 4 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1327,6 +1327,41 @@ void AtenSizeOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
   });
 }
 
+//===----------------------------------------------------------------------===//
+// AtenSelectIntOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenSelectIntOp::fold(FoldAdaptor adaptor) {
+  auto self = dyn_cast_or_null<DenseElementsAttr>(adaptor.getSelf());
```
daveliddell, line 1335: James was just telling me that dyn_cast_or_null is for when you think your arg might be null, and you would use dyn_cast if that's not a possibility.  If that's true, I'll leave it to you whether to change it or not.  :-)

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1327,6 +1327,41 @@ void AtenSizeOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+  auto ty = cast<ValueTensorType>(getType());
```
daveliddell, line 1336: I see that CI failed.  Maybe you're caught in the same canonicalization loop that I was: eventually it will be castable to a ValueTensorType, but not the first time you get called

daveliddell (approved):

lib/Dialect/Torch/IR/TorchOps.cpp
rsuderman: `FoldAdaptor` can return `nullptr` when the attribute is not foldable, `dyn_cast_or_null` allows us to avoid checking type and `nullptr` separately.

lib/Dialect/Torch/IR/TorchOps.cpp
rsuderman: I am surprised it fails (probably related to `BaseTensorType`) but either way, change to `dyn_cast` should fix.

renxida (approved):

lib/Dialect/Torch/IR/TorchOps.cpp
daveliddell: I had the same problem with index_select.  It turns out that the first time the folder is called, the tensor is still `!torch.tensor`, with no shape or size.  It seems the tensor is resolved to `!torch.vtensor` in several steps, so if you were to print out the type of the tensor at this point, you would see your print statement several times (once per call of your folder, of course) with different types, progressing toward a fully shaped and typed vtensor.

PR #2891, opened 2024-02-09
Bump stablehlo to openxla/stablehlo@e191eb4c3c3f3144503a8a117d760de5d…

qingyunqu:
…dcc7e89.  
* to involve `chlo-legalize-to-stablehlo` pass.


sjain-stanford (approved):
LGTM, and this is not a blocker but a request: We've seen issues in the past where build errors were not exposed in the CMake build (which is run as part of per-merge CI). If it's not too much trouble, can you trigger BazelBuildAndTest on your fork just to be sure? 

Here's how:
From your fork of torch-mlir, go to `Actions` -> select `Bazel Build and Test` on the left, then click on `Run workflow` on the right, select your branch then hit `Run workflow`.

Issue #2888, opened 2024-02-08
onnx.Add int8 test case fails to lower to linalg


renxida:
Error message:

https://gist.github.com/renxida/2927572ce08c07a371ce308ce3955cb1

Reproducing mlir:

```
// CHECK-LABEL: @test_add_uint8
func.func @test_add_uint8(%arg0: !torch.vtensor<[3,4,5],ui8>, %arg1: !torch.vtensor<[3,4,5],ui8>) -> !torch.vtensor<[3,4,5],ui8> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 14 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
  // CHECK: %[[INT1:.*]] = torch.constant.int 1
  // CHECK: torch.aten.add.Tensor %arg0, %arg1, %[[INT1]] : !torch.vtensor<[3,4,5],ui8>, !torch.vtensor<[3,4,5],ui8>, !torch.int -> !torch.vtensor<[3,4,5],ui8>
  %0 = torch.operator "onnx.Add"(%arg0, %arg1) : (!torch.vtensor<[3,4,5],ui8>, !torch.vtensor<[3,4,5],ui8>) -> !torch.vtensor<[3,4,5],ui8>
  return %0 : !torch.vtensor<[3,4,5],ui8>
}
```

Relevant stack trace sections:

```
 #4 0x000055c2cbe86210 mlir::torch::Torch::convertScalarToDtype(mlir::OpBuilder&, mlir::Location, mlir::Value, mlir::Type, std::optional<mlir::Type>, std::optional<mlir::Type>) /home/azureuser/torch-mlir/lib/Conversion/Utils/Utils.cpp:269:11
 #5 0x000055c2cb98027c createLinalgPayloadCalculationForElementwiseOp(mlir::OpBuilder&, mlir::Location, mlir::TypeConverter const*, mlir::ValueRange, mlir::Operation*, llvm::ArrayRef<mlir::Value>) /home/azureuser/torch-mlir/lib/Conversion/TorchToLinalg/Uncategorized.cpp:575:19
[...]

PR #2893, opened 2024-02-09
Fix test_add_uint8 failure to lower to linalg

renxida:
By updating convertScalarToDtype invocation pass original source and destination datatypes for the add op. Also fixes a potential problem with the sub op.


renxida:
Fixes #2888

renxida:
Requesting reviews from @PhaneeshB @Shukla-Gaurav because you are working on related type conversion issues.

PhaneeshB (approved):
LGTM!

PR #2894, opened 2024-02-09
[torch-mlir][sparse] implement first sparse_jit end-to-end path

aartbik:
This PR introduces a sparse_jit wrapper that can run simple models with sparse tensor inputs end-to-end. The implementation shows all required components on modifying sparse tensor types with a 1:N relation on the call sites. Two tests shows that the JIT runs end-to-end while computing the correct results.

More details to follow (generalizing to COO and different ranks, as well as support for *output* sparse tensors), but the general concepts are all here now.

**_Update: Thanks to Rob, bump to proper LLVM/MLIR hash is done!_**

_**NOTE that all parameter passing changes are nicely done "downstream" in MLIR, so very little changes are required in torch-mlir code proper**_


python/torch_mlir/extras/fx_importer.py
```diff
@@ -479,14 +483,18 @@ def format_asm_shape(self, shape: torch.Size) -> str:
...
+        shape: torch.Size,
+        dtype: torch.dtype,
+        sparsity: Tuple[torch.layout, int, int],
     ):
         shape_asm = self.format_asm_shape(shape)
         mlir_dtype = str(self.dtype_to_type(dtype))
-        if sparse_layout is not None:
-            sparsity = sparsity_encoding(shape, sparse_layout)
+        if sparsity is not None:
+            sparse_layout, posw, crdw = sparsity
```
PeimingLiu, line 494: Is `nnz` also needed here?

PeimingLiu (commented):
Nice!!

python/torch_mlir/extras/fx_importer.py
aartbik: most likely, also sparsity level eventually (since CSR/CSC can be define at higher dimension in torch.sparse).

but for the 2-dim CSR/CSC, this is all we need (with probably _**two**_ more ints at most in a future extension)

aartbik:
Note that Robert is kindly working on the required LLVM/MLIR "bump":

https://github.com/llvm/torch-mlir/pull/2895

rdadolf:
Seeing the bodies of `test_sparse_sum` and `test_sparse_SpMM` brings some strong emotions to the surface, Aart. 18 months ago, this was a pipedream. To see it (mostly) propagating *properly* with encoding handled as a first-class property is incredible.

As far as this PR specifically, I suspect re-packing output tensors into a form torch can grok will iron out a bit of the clunkiness here and there. I'm assuming you're also planning to matrix the `_sum` and `_SpMM` tests with `SPARSE_LAYOUTS` when the COO form is end-to-end?

rdadolf (approved):

aartbik:
Thank you so much for your very kind words, Bob!

I, too, am thrilled seeing all the pieces of the sparse puzzle come together,  building on the amazing foundation laid by the torch-mlir team, with Stella leading the way to FX importing!

python/torch_mlir/extras/fx_importer.py
```diff
@@ -208,28 +208,33 @@
 }
 
 
-def sparsity_encoding(shape: torch.Size, sparse_layout: torch.layout) -> str:
+def sparsity_encoding(
+    shape: torch.Size, sparsity: Tuple[torch.layout, int, int]
```
stellaraccident, line 212: Use lowercase `tuple` type descriptors in new code. These capitalized versions are deprecated

python/torch_mlir/extras/fx_importer.py
```diff
@@ -479,14 +484,17 @@ def format_asm_shape(self, shape: torch.Size) -> str:
     """Return IrType for !torch.vtensor with the given shape and dtype"""
 
     def get_vtensor_type(
-        self, shape: torch.Size, dtype: torch.dtype, sparse_layout: torch.layout = None
+        self,
+        shape: torch.Size,
+        dtype: torch.dtype,
+        sparsity: Tuple[torch.layout, int, int],
```
stellaraccident, line 490: Here too

stellaraccident: This should be annotated as `Optional` based on usage in the function. Also default it to None.

I'd prefer this was a keyword only arg (add an `, *,` before).

python/torch_mlir/extras/fx_importer.py
```diff
@@ -479,14 +484,17 @@ def format_asm_shape(self, shape: torch.Size) -> str:
...
     ):
         shape_asm = self.format_asm_shape(shape)
         mlir_dtype = str(self.dtype_to_type(dtype))
-        if sparse_layout is not None:
-            sparsity = sparsity_encoding(shape, sparse_layout)
+        if sparsity is not None:
+            encoding = sparsity_encoding(shape, sparsity)
             return IrType.parse(
-                f"!torch.vtensor<[{shape_asm}],{str(mlir_dtype)},{sparsity}>",
+                f"!torch.vtensor<[{shape_asm}],{str(mlir_dtype)},{encoding}>",
```
stellaraccident, line 497: I'm having trouble seeing how this works if sparsity is None. Either the code or the function signature is wrong.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -507,12 +515,12 @@ def node_val_to_type(self, node: torch_fx.Node) -> IrType:
...
                     )
                 else:
-                    return self.tensor_metadata_to_type(tensor_meta, sparse_layout)
+                    return self.tensor_metadata_to_type(tensor_meta, sparsity)
             elif val is not None:
                 # some nodes with symbolic inputs pass a 'val' attribute rather than
                 # tensor_meta
                 if isinstance(val, TorchFakeTensor):
-                    return self.get_vtensor_type(val.size(), val.dtype, sparse_layout)
+                    return self.get_vtensor_type(val.size(), val.dtype, sparsity)
```
stellaraccident, line 523: Will pass None for sparsity to get_vtensor_type, which I think will trigger and UnboundLocalError.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -528,16 +536,16 @@ def node_val_to_type(self, node: torch_fx.Node) -> IrType:
             )
 
     def tensor_metadata_to_type(
-        self, tm: TensorMetadata, sparse_layout: torch.layout = None
+        self, tm: TensorMetadata, sparsity: Tuple[torch.layout, int, int] = None
```
stellaraccident, line 539: s/Tuple/tuple/

stellaraccident: Also per how it is used, should be `Optional[]`

test/python/fx_importer/sparse_test.py
```diff
@@ -22,13 +26,50 @@
...
+            a.layout,
+            sparse_overhead_width(a.ccol_indices().dtype),
+            sparse_overhead_width(a.row_indices().dtype),
+        )
+    else:
+        raise RuntimeError(f"Unsupported sparse layout for {a}")
+
+
+def sparse_export(
+    f: Callable, args: Tuple[Any, ...], kwargs: Optional[Dict[str, Any]] = None
```
stellaraccident, line 71: s/Tuple/tuple/
s/Dict/dict/

stellaraccident (commented):
Some nits and I'm scratching my head about why we don't have tests failing for what I think is some unhandled None cases for sparsity. Could you have a look?

python/torch_mlir/extras/fx_importer.py
aartbik: L494 tests sparsity is not None.
(and since in that case encoding is not None, we are good here)

At L500 we deal with the sparsity (and thus encoding) is None

python/torch_mlir/extras/fx_importer.py
aartbik: Ah, thanks for pointing that out.

python/torch_mlir/extras/fx_importer.py
stellaraccident: huh, not sure how I missed that. Was looking 3-4x for an `encoding =`. Maybe didn't expand something or my eyes were just not looking.

stellaraccident (approved):
Thanks. Fine to land with the nits resolved.

python/torch_mlir/extras/fx_importer.py
aartbik: makes sense, done

python/torch_mlir/extras/fx_importer.py
aartbik: sparsity is None is accounted for?

python/torch_mlir/extras/fx_importer.py
aartbik: done, and done

test/python/fx_importer/sparse_test.py
aartbik: thanks

aartbik:
Thanks for the very informative review. I like how we are moving towards all the modern Python features
(or perhaps that just dates me ;-)

Issue #2843, opened 2024-01-31
Fx importer does not support bfloat16


kumardeepakamd:
**Error seen:**
```
Traceback (most recent call last):
  File "/proj/gdba/kumar/nod/SHARK-TestSuite/e2eshark/t-r-bf16-direct-fx-importer/pytorch/combinations/mlp/runmodel.py", line 131, in <module>
    torch_mlir_model = export_and_import(model, test_input)
  File "/proj/gdba/kumar/nod/SHARK-TestSuite/e2eshark/t-r-bf16-direct-fx-importer/pytorch/combinations/mlp/runmodel.py", line 59, in export_and_import
    fx_importer.import_frozen_exported_program(prog)
  File "/proj/gdba/kumar/anaconda3/envs/e2e/lib/python3.10/site-packages/torch_mlir/extras/fx_importer.py", line 351, in import_frozen_exported_program
    self.import_stateless_graph(g)
  File "/proj/gdba/kumar/anaconda3/envs/e2e/lib/python3.10/site-packages/torch_mlir/extras/fx_importer.py", line 377, in import_stateless_graph
    node_importer.import_nodes(g.nodes)
  File "/proj/gdba/kumar/anaconda3/envs/e2e/lib/python3.10/site-packages/torch_mlir/extras/fx_importer.py", line 620, in import_nodes
    self._import_torch_op_overload(loc, node, target)
  File "/proj/gdba/kumar/anaconda3/envs/e2e/lib/python3.10/site-packages/torch_mlir/extras/fx_importer.py", line 798, in _import_torch_op_overload
    self._import_argument(loc, node.args[i], parameter.type)
  File "/proj/gdba/kumar/anaconda3/envs/e2e/lib/python3.10/site-packages/torch_mlir/extras/fx_importer.py", line 860, in _import_argument
    return self._import_literal(arg)
  File "/proj/gdba/kumar/anaconda3/envs/e2e/lib/python3.10/site-packages/torch_mlir/extras/fx_importer.py", line 877, in _import_literal
    return converter(py_value, self, self._cc)
  File "/proj/gdba/kumar/anaconda3/envs/e2e/lib/python3.10/site-packages/torch_mlir/extras/fx_importer.py", line 1169, in <lambda>
    lambda arg, gni, cc: _make_vtensor_literal_op(
[...]

PR #2896, opened 2024-02-10
Add support for bfloat16 in fximporter

dan-garvey:
this introduces an additional soft dependency on the python ml_dtypes python packages in order to support bfloat16

Addresses #2843

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -1045,10 +1045,12 @@ def _make_vtensor_literal_op(
 ) -> Operation:
     mapping = py_attr_tracker.track(tensor)
     if mapping.is_empty:
+        # check support for bfloat16
+        assert (
+            not (tensor.dtype == torch.bfloat16 and ml_dtypes is None)
+        ), f'torch.bfloat16 requires the mldtypes package, please run:\n\npip install ml_dtypes\n')
         # Resolve the attribute.
         npy_dtype = TORCH_DTYPE_TO_NPY_TYPE.get(tensor.dtype)
-        if tensor.dtype == torch.bfloat16 and ml_dtypes is None:
-            print(f'torch.bfloat16 requires the mldtypes package, please run:\n\npip install ml_dtypes\n')
         assert (
             npy_dtype is not None
         ), f"Can not create literal tensor for unsupported datatype: {tensor.dtype}"
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -1048,7 +1048,7 @@ def _make_vtensor_literal_op(
         # check support for bfloat16
         assert (
             not (tensor.dtype == torch.bfloat16 and ml_dtypes is None)
-        ), f"torch.bfloat16 requires the ml_dtypes package, please run:\n\npip install ml_dtypes\n")
+        ), f"torch.bfloat16 requires the ml_dtypes package, please run:\n\npip install ml_dtypes\n"
         # Resolve the attribute.
         npy_dtype = TORCH_DTYPE_TO_NPY_TYPE.get(tensor.dtype)
         assert (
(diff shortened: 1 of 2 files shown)


dan-garvey:
@stellaraccident 
How do you feel about adding ml_dtypes dependency in torch-mlir and turbine in order to support bfloat16?

dan-garvey:
fyi @kumardeepakamd

stellaraccident:
Yes, that is reasonable

dan-garvey:
> Yes, that is reasonable

Ok, I'll clean this up and add the same in turbine with tests

python/torch_mlir/extras/fx_importer.py
```diff
@@ -136,7 +137,7 @@
     torch.int16: np.int16,
     torch.int32: np.int32,
     torch.int64: np.int64,
-    # torch.bf16: None, there's no equivalent np datatype so this isn't supported right now
+    torch.bfloat16: mld.bfloat16,# , there's no equivalent np datatype so this isn't supported right now
```
stellaraccident, line 140: What do you think about making this a "soft" dependency? Also, let's just spell out ml_dtypes. The abbreviation isn't paying for itself.

I'd recommend at the top:

```
try:
  import ml_dtypes
except ModuleNotFoundError:
  # The third-party ml_dtypes package provides some optional
  # low precision data-types. If used in this file, it is
  # conditional.
  ml_dtypes = None
```

Then instead of doing this inline:

```
if ml_dtypes is not None:
  TORCH_DTYPE_TO_NPY_TYPE[torch.bfloat16] = ml_dtypes.bfloat16
```

stellaraccident (approved):
I'd prefer we make this a conditional dep. Left comments on the review on how to do it.

python/torch_mlir/extras/fx_importer.py
dan-garvey: sounds good

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1047,6 +1047,8 @@ def _make_vtensor_literal_op(
     if mapping.is_empty:
         # Resolve the attribute.
         npy_dtype = TORCH_DTYPE_TO_NPY_TYPE.get(tensor.dtype)
+        if tensor.dtype == torch.bfloat16 and ml_dtypes is None:
+            print(f'torch.bfloat16 requires the mldtypes package, please run:\n\npip install ml_dtypes\n')
```
stellaraccident, line 1051: Please don't do loose prints in API code. An assert here with a formatted message is fine/preferred.

dan-garvey:
can we merge this if a file an issue to add a test once they are migrated here? Or would you prefer I just mirror this on the turbine side for now and add the test there?

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1036,6 +1045,10 @@ def _make_vtensor_literal_op(
 ) -> Operation:
     mapping = py_attr_tracker.track(tensor)
     if mapping.is_empty:
+        # check support for bfloat16
+        assert (
+            not (tensor.dtype == torch.bfloat16 and ml_dtypes is None)
+        ), f'torch.bfloat16 requires the mldtypes package, please run:\n\npip install ml_dtypes\n')
```
stellaraccident, line 1051: Nit: this file exclusively uses double quotes. Be consistent style wise

stellaraccident (approved):
Yes, I would like an issue in torch-mlir outlining the testing upstreaming needs but can be done async. We've still got more to upstream in this area, so I'm fine with it being in progress for the moment.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1048,7 +1048,7 @@ def _make_vtensor_literal_op(
         # check support for bfloat16
         assert (
             not (tensor.dtype == torch.bfloat16 and ml_dtypes is None)
-        ), f'torch.bfloat16 requires the mldtypes package, please run:\n\npip install ml_dtypes\n')
+        ), f"torch.bfloat16 requires the mldtypes package, please run:\n\npip install ml_dtypes\n")
```
stellaraccident, line 1051: Nit: need underscore in mldtypes

PR #2899, opened 2024-02-12
Implement lowering of torch.aten.norm.Scalar

petarTSyrmia:
Closes [nod-ai/SHARK-Turbine#365](https://github.com/nod-ai/SHARK-Turbine/issues/365)


lib/Conversion/TorchToLinalg/Reduction.cpp
```diff
@@ -587,11 +609,22 @@ class ConvertReductionOp : public ConversionPattern {
       return rewriter.notifyMatchFailure(
           op, "failed to create linalg.generic operation for reduction");
 
+    // If this is aten.norm.Scalar op, then we need to generate another
+    // linalg.generic op that references the first linalg.generic op.
+    if (auto normOp = dyn_cast<AtenNormScalarOp>(op)) {
```
newling, line 614: nit `if (isa<AtenNormScalarOp>(op))`

lib/Conversion/TorchToLinalg/Reduction.cpp
```diff
@@ -551,6 +568,11 @@ class ConvertReductionOp : public ConversionPattern {
     if (isa<AtenAllDimOp>(op) && elemType.isa<mlir::IntegerType>() &&
         elemType.getIntOrFloatBitWidth() == 8)
       return rewriter.notifyMatchFailure(op, "uint8 is not supported");
+
+    if (isa<AtenNormScalarOp>(op) &&
+        !(elemType.isa<mlir::FloatType>() || elemType.isa<mlir::ComplexType>()))
+      return rewriter.notifyMatchFailure(
+          op, "only float and complex types are valid for norm op");
```
newling, line 575: Is this a constraint from pytorch? Please include a comment in the code, if so.  You could use an op verifier for this (see for example Torch_AtenPermuteOp)

lib/Conversion/TorchToLinalg/Reduction.cpp
```diff
@@ -484,10 +499,12 @@ class ConvertReductionOp : public ConversionPattern {
     return err ? Value{} : powOp;
   }
 
-  FailureOr<Value> createSecondReductionForVectorNormOp(
-      Location loc, Type elemType, AtenLinalgVectorNormOp op, Value ordOp,
-      Value firstReduction, const torch_to_linalg::ReductionOpInfo &opInfo,
-      ConversionPatternRewriter &rewriter) const {
+  template <typename T>
```
newling, line 502:
```suggestion
  template <typename TOp>
```

nit

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -3881,6 +3884,17 @@ def aten〇linalg_vector_norm〡dtype(self_rank_dtype: Tuple[int, int], ord: Uni
...
     return aten〇std〡dtype(self_rank_dtype)
 
+@check_dtype_function(
+        _check_tensors_with_the_same_dtype(
+            num_of_tensors=1,
+            error_types={torch.bool, torch.uint8, torch.int8, torch.int16, torch.int32, torch.int64}))
+def aten〇norm〇Scalar〡dtype(self_rank_dtype: Tuple[int, int], p: Union[int, float, complex] = 2) -> int:
+    self_rank, self_dtype = self_rank_dtype
+    assert not is_integer_dtype(self_dtype)
+    if self_dtype == torch.complex32:
```
newling, line 3894: It isn't obvious to me what this conditional is doing, please add a comment in the code (I'm not familiar with complex32, it might be obvious to someone else!)

newling (changes requested):
Looks good. Thanks for the code reuse with template. Just a few minor suggestions

petarTSyrmia:
Hi @newling! Thanks for the comments!
I updated the code per your requests. 
As for the types acceptable by norm.Scalar, I have included a verifier as you suggested, which allows only float and complex types for input tensor dtype, but I intend to add linalg lowering support for complex types later, possibly in as a separate issue, if that sounds ok.

ptrifunovic98:
> > As for the types acceptable by norm.Scalar, I have included a verifier as you suggested, which allows only float and complex types for input tensor dtype, but I intend to add linalg lowering support for complex types later, possibly in as a separate issue, if that sounds ok.
> 
> Please check what happens if you run a test with complex32: ensure that a sensible error message is obtained mentioning that there is something unimplemented.

I've updated the lowering with a check for complex types. It produces an error when a test with complex input type is run. It should be ready for another review and workflow checks.

newling:
Thanks @ptrifunovic98

ptrifunovic98:
@newling I updated the xfail_set for ONNX so I think the CI jobs will pass now

newling (approved):

PR #2901, opened 2024-02-12
[torch-mlir][sparse] sparsity metadata refinement

aartbik:
Various improvements on sparsity metadata:

(1) define single data structure for all sparsity related metadata 
(2) handle batched dense dimensions, as well as dense subtensor dimensions
(3) refine sparsity propagation for deeper networks

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -210,7 +210,7 @@ SYMBOLIC_OP_TO_TORCH_OP = {
 
 
 @dataclass(frozen=True)
-class Sparsity:
+class SparsityMeta:
     """Class for keeping track of sparsity meta data."""
 
     layout: torch.layout
@@ -221,7 +221,7 @@ class Sparsity:
     crd_width: int
 
 
-def sparsity_encoding(shape: torch.Size, sparsity: Sparsity) -> str:
+def sparsity_encoding(shape: torch.Size, sparsity: SparsityMeta) -> str:
     """Returns sparse tensor encoding for the given sparse layout as string."""
     assert sparsity is not None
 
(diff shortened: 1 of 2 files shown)


python/torch_mlir/extras/fx_importer.py
```diff
@@ -208,34 +209,58 @@
...
     else:
         # TODO: block format (derive block size!)
         raise RuntimeError(f"Unsupported sparse layout {sparse_layout}")
 
-    return f"#sparse_tensor.encoding<{{map={smap},posWidth={posw},crdWidth={crdw}}}>"
+    if batch_dim > 0:
+        batch = ",".join(f"d{d}:dense" for d in range(0, batch_dim))
+        lvls = f"{batch},{lvls}"
+
+    if dense_dim > 0:
```
PeimingLiu, line 258: nit: Are these condition necessary? `for d in range(0, 0)` seems to work even when batch_dim == 0 as well.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -208,34 +209,58 @@
 }
 
 
-def sparsity_encoding(
-    shape: torch.Size, sparsity: tuple[torch.layout, int, int]
-) -> str:
-    """Returns sparse tensor encoding for the given sparse layout as string.
+@dataclass(frozen=True)
+class Sparsity:
```
PeimingLiu, line 213: nit: How about `SparsityMeta` (analogous to `TensorMeta` in PyTorch).

stellaraccident (approved):

python/torch_mlir/extras/fx_importer.py
aartbik: yeah, except for adding the , in between batch and lvls in that case

python/torch_mlir/extras/fx_importer.py
aartbik: Sure (btw, this may migrate into upstream PyTorch at some point, depending on how our FX export efforts will go)

python/torch_mlir/extras/fx_importer.py
aartbik: (but I find that other people are usually much better in writing very concise python, so let me know if I missed a cleverl trick)

PeimingLiu (approved):

PR #2904, opened 2024-02-13
Fix onnx.Unsqueeze

renxida:
The current implementation breaks Unsqueeze down into multiple AtenUnsqueezeOps.

The intermediate results are not given a shape and that fails to lower to linalg.

This patch pretty much completely rewrites the onnx conversion pass to fetch axes from a onnx.Constant op directly and sort them in a SmallVector<int64_t>, so that the shapes can be statically known & AtenToLinAlg conversion can proceed.


renxida:
#2991 solves this and is already merged

PR #2906, opened 2024-02-13
[torch-mlir][sparse] add JIT test to expose pending issues

aartbik:
This test exposes issues that need fixing
(1) propagate sparsity into the FX graph (over elt-wise) (2) batched dimensions need a new "dense(batch)" format


PeimingLiu (approved):

stellaraccident (approved):
I'm fine landing "TODO tests" like this so long as you are engaged to drive them to completion. If by chance, you end up going a different way, would just ask your help in expunging them (as I don't expect anyone but you/your team has the context to know what is going on here).

aartbik:
Absolutely! We expect this to fix this really soon, since the whole team is now focused on our PyTorch efforts (but in the unlikely event this does not happen, we will help cleaning this up, of course!).

PR #2907, opened 2024-02-13
[torch] Fix tm_tensor.attention for end-to-end

rsuderman:
Some operations include a backend matcher for specialized operations. We
map these back to generics so they appropriately match to the high
performance versions. This is done for the attention operation.

### DIFF
--- a/include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorOps.td
+++ b/include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorOps.td
@@ -314,7 +314,7 @@ def TMTensor_AttentionOp : TMTensor_Op<"attention",
       return getOutputType().getRank();
     }
     int64_t getIterationDomainRank() {
-      return 2;
+      return getKeyRank() - 2;
     };
     // Method to implement for specifying output range for
     // DestinationStyleOpInterface
--- a/include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorOps.td
+++ b/include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorOps.td
@@ -313,9 +313,6 @@ def TMTensor_AttentionOp : TMTensor_Op<"attention",
     int64_t getOutputRank() {
       return getOutputType().getRank();
     }
-    int64_t getIterationDomainRank() {
-      return getKeyRank() - 2;
-    };
     // Method to implement for specifying output range for
     // DestinationStyleOpInterface
     std::pair<int64_t, int64_t> getDpsInitsPositionRange() {
--- a/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
+++ b/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
@@ -1625,8 +1625,6 @@ public:
     auto collapseBatch = [&rewriter, &reassociation,
                           loc](Value value) -> Value {
       auto valueTy = cast<ShapedType>(value.getType());
-      if (valueTy.getRank() < 3)
-        return nullptr;
       if (valueTy.getRank() == 3)
         return value;
 
--- a/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
+++ b/lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
@@ -1600,27 +1600,77 @@ public:
                                            "only default scale supported");
     }
 
+    auto opTy = cast<ValueTensorType>(op.getType()).toBuiltinTensor();
+    auto query = adaptor.getQuery();
+    auto value = adaptor.getValue();
+    auto key = adaptor.getKey();
+
+    auto loc = op.getLoc();
+    auto collapseBatch = [&rewriter, loc](Value value) -> Value {
+      auto valueTy = cast<ShapedType>(value.getType());
+      if (valueTy.getRank() <= 3)
+        return value;
+
+      llvm::SmallVector<int64_t, 3> newShape(3, 1);
+      llvm::SmallVector<ReassociationIndices, 3> reassociation(3);
+      reassociation[1].push_back(valueTy.getRank() - 2);
+      reassociation[2].push_back(valueTy.getRank() - 1);
+      newShape[1] = valueTy.getDimSize(valueTy.getRank() - 2);
+      newShape[2] = valueTy.getDimSize(valueTy.getRank() - 1);
+
+      for (int i = 0, s = valueTy.getRank() - 2; i < s; ++i) {
+        reassociation.front().push_back(i);
+        if (valueTy.isDynamicDim(i) || newShape[0] == ShapedType::kDynamic) {
+          newShape[0] = ShapedType::kDynamic;
+          continue;
+        }
+        newShape[0] = newShape[0] * valueTy.getDimSize(i);
+      }
+
+      auto collapseTy = valueTy.clone(newShape);
+      return rewriter.create<tensor::CollapseShapeOp>(loc, collapseTy, value,
+                                                      reassociation);
+    };
+
+    query = collapseBatch(query);
+    key = collapseBatch(key);
+    value = collapseBatch(value);
+
     SmallVector<int64_t> outSizes(
-        adaptor.getQuery().getType().cast<ShapedType>().getShape());
+        query.getType().cast<ShapedType>().getShape());
     SmallVector<int64_t> valueSizes(
-        adaptor.getValue().getType().cast<ShapedType>().getShape());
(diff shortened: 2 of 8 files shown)


lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1600,27 +1600,77 @@ class ConvertAtenScaledDotProductAttentionOp
...
 
+    auto opTy = cast<ValueTensorType>(op.getType()).toBuiltinTensor();
+    auto query = adaptor.getQuery();
+    auto value = adaptor.getValue();
+    auto key = adaptor.getKey();
+
+    auto loc = op.getLoc();
+    auto collapseBatch = [&rewriter, loc](Value value) -> Value {
+      auto valueTy = cast<ShapedType>(value.getType());
+      if (valueTy.getRank() <= 3)
```
harsh-nod, line 1611: If the rank is less than 3, should we emit a failure because we can't codegen that?

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1600,27 +1600,77 @@ class ConvertAtenScaledDotProductAttentionOp
...
+      newShape[1] = valueTy.getDimSize(valueTy.getRank() - 2);
+      newShape[2] = valueTy.getDimSize(valueTy.getRank() - 1);
+
+      for (int i = 0, s = valueTy.getRank() - 2; i < s; ++i) {
+        reassociation.front().push_back(i);
+        if (valueTy.isDynamicDim(i) || newShape[0] == ShapedType::kDynamic) {
+          newShape[0] = ShapedType::kDynamic;
+          continue;
+        }
+        newShape[0] = newShape[0] * valueTy.getDimSize(i);
```
harsh-nod, line 1627: If we have a 4-d rank tensor, whose shape is <?x4x1024x64> then newShape[0] will be -4 ? Is this intended?

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1600,27 +1600,77 @@ class ConvertAtenScaledDotProductAttentionOp
...
-        SmallVector<Value>{output});
+    Value attention =
+        rewriter
+            .create<AttentionOp>(loc, outType,
+                                 SmallVector<Value>{query, key, value},
+                                 SmallVector<Value>{output})
+            .getResult()[0];
+
+    if (opTy != outType) {
+      llvm::SmallVector<ReassociationIndices, 3> reassociation(3);
```
harsh-nod, line 1661: Can we re-use the reassociation from above? Just wondering when the expand reassociation will be different from the collapse one.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -189,6 +189,78 @@ class ConvertHasValueSemanticsOpsToValueTensors : public RewritePattern {
...
+  }
+
+private:
+  DenseMap<StringAttr, SmallVector<HandlerFn, 1>> namedHandlers;
+};
+
+void TorchMatchSpecializedBackendOp::populateSpecializedConversions(
+    TorchMatchSpecializedBackendOp &matcher) {
+  matcher.populate(
+      "torch.aten._scaled_dot_product_flash_attention_for_cpu",
```
harsh-nod, line 240: Does PyTorch emit separate SDPA ops for CPU and GPU? And is this matcher a way to unify our codegen efforts for both?

test/Dialect/Torch/reduce-op-variants.mlir
```diff
@@ -171,3 +184,22 @@ func.func @torch.aten.bernoulli_.float(%t: !torch.tensor) -> !torch.tensor {
   %ret = torch.aten.bernoulli_.float %t, %p, %generator : !torch.tensor, !torch.float, !torch.none -> !torch.tensor
   return %ret : !torch.tensor
 }
+
+// -----
+
```
harsh-nod, line 189: Not sure if this is possible, but can you add a test that shows hows the 4d SDPA op is converted to a 3d linalg_ext with collapse and expands?

harsh-nod (commented):
Thanks! Overall, looks good. Just had a few minor comments.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
rsuderman: There does not appear to be a `gpu` one. There is a {}_math one, but I don't know where we invoke one or another. We should likely handle them as they appear.

harsh-nod: Sounds good, thanks!

test/Dialect/Torch/reduce-op-variants.mlir
rsuderman: So the easiest way is to make one of the end-to-end tests do 4D. I had to include the dynamic dimensions cleanup (sorry for stacking).

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
rsuderman: Reworked to reuse the reassociation list for the collapses. Included some more dimension checks so operands must be the same rank.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
rsuderman: Nope, check the bailout above. Once it goes dynamic it bails out.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
rsuderman: Notify match error added.

include/torch-mlir-dialects/Dialect/TMTensor/IR/TMTensorOps.td
```diff
@@ -314,7 +314,7 @@ def TMTensor_AttentionOp : TMTensor_Op<"attention",
       return getOutputType().getRank();
     }
     int64_t getIterationDomainRank() {
-      return 2;
+      return getKeyRank() - 2;
```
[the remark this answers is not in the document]
rsuderman: I ripped out the getIterationDomainRank. It turns out its completely irrelevant. Because of the multi-loop behavior of flash attention we actually have multiple adjacent loops with different iteration domains, so this is effectively stubbed out.

I was wondering why it didn't matter what this value was, tests always passed.

test/Dialect/Torch/reduce-op-variants.mlir
harsh-nod: No worries, thank you for accommodating.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1600,27 +1600,84 @@ class ConvertAtenScaledDotProductAttentionOp
...
+        }
+        newShape[0] = newShape[0] * valueTy.getDimSize(i);
+      }
+
+      auto collapseTy = valueTy.clone(newShape);
+      return rewriter.create<tensor::CollapseShapeOp>(loc, collapseTy, value,
+                                                      reassociation);
+    };
+
+    query = collapseBatch(query);
```
harsh-nod, line 1650: Since this can return a nullptr, maybe add a check here?

rsuderman: I actually can remove the nullptr return altogether as the rank is checked before hand.

lib/Dialect/Torch/Transforms/ReduceOpVariants.cpp
```diff
@@ -189,6 +189,78 @@ class ConvertHasValueSemanticsOpsToValueTensors : public RewritePattern {
...
+          auto oldOperands = op->getOperands();
+          llvm::SmallVector<Value> newOperands{
+              oldOperands[0], oldOperands[1], oldOperands[2], oldOperands[5],
+              oldOperands[3], oldOperands[4], oldOperands[6]};
+
+          auto newOp = rewriter.create<Torch::AtenScaledDotProductAttentionOp>(
+              op.getLoc(), op->getResultTypes()[0], newOperands,
+              op->getAttrs());
+          rewriter.replaceAllUsesWith(op.getResult(0), newOp.getResult());
+          rewriter.eraseOp(op);
```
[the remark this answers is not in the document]
rsuderman: Yes but it is better to erase as we know its not longer needed.

PR #2910, opened 2024-02-14
DecomposeComplexOps: update parseEquation to skip space char for AtenEinsumOp op

zezhang:
Just a minor update to skip the space char if included in the equation string


navahgar (commented):
The change LGTM. 

I see some other changes included in the PR under "Unchanged files with check annotations". Can you rebase with the latest on main?

PR #2913, opened 2024-02-15
[onnx] Fix onnx.gather lowering to use torch.aten.index_select

rsuderman:
Onnx's gather maps directly to `torch.aten.index_select`. We should just use that path.


renxida (approved):
lgtm, just need to pass the testcases

PR #2916, opened 2024-02-16
[fx] Support mutation in ExportedProgram.

stellaraccident:
As of https://github.com/pytorch/pytorch/pull/118969, `ExportedProgram` has the long awaited fixes to correctly categorize various things relating to parameters, buffers, mutated inputs and constants.

With this additional modeling, we are finally able to implement (safely/soundly) the mutable semantics that were attempted on the TorchScript path. The difference is that on that path, we had to conservatively treat everything as mutable and run some dodgy heuristics (which have been the cause of many bugs relating to "MaximizeValueSemantics") to try to get back to an immutable state.

The new model supports mutability at the graph edges, allowing both user inputs and buffers to be mutated (there is some more support than that, but that is all I fully tracked through to implementation).

Therefore, when we receive programs like this, we now can selectively enable mutation at the edges. This happens to be the mutability model that IREE supports, which I expect to be a primary beneficiary. However, there is nothing stopping anyone else from handling the `!torch.tensor` types and the existing copy/overwrite ops that will be selectively added.

Since this relies on API changes that will not release until 2.3, I'm being a bit cautious about not refactoring existing facilities.

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -91,11 +91,6 @@ __all__ = [
     "FxImporter",
 ]
 
-# An external callback that, given a Python value and a GraphNodeImporter, may choose
-# to materialize IR to load the value as a vtensor. If it returns None, then default
-# literal resolution proceeds.
-LiteralResolverCallback = Callable[[Any, "GraphNodeImporter"], Optional[Value]]
-
 REQUIRED_DIALCTS = [
     "builtin",
     "func",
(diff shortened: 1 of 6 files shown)


test/python/fx_importer/v2.3/mutation_import.py
```diff
@@ -0,0 +1,163 @@
...
+            return x * y
+
+    m = fx.export_and_import(
+        Basic(),
+        torch.randn(3, 4),
+        torch.randn(3, 4),
+        experimental_support_mutation=True,
+    )
+    print(m)
+    m.operation.verify()
```
sjain-stanford, line 83: If possible, it would be a nice reference to include a negative test (i.e. importing an in-place `mul_` but with `experimental_support_mutation=False`).

python/torch_mlir/extras/fx_importer.py
```diff
@@ -377,6 +610,10 @@ def import_frozen_exported_program(self, prog: torch.export.ExportedProgram):
         As we anticipate more nuanced treatment options in the future, we name this
         method to indicate that it is producing "frozen" modules. Additional top-level
         approaches to handling state can be introduced later as an addition.
+
+        TODO: This mechanism should be eventually replaced by `import_program` with
+        hooks set on the subclass to freeze parameters and buffers. However, that is
+        waiting for the Torch 2.3 release cut.
```
sjain-stanford, line 616: I'm hoping this could be done in a backwards compatible way (i.e. to allow it to work with older Torch releases)?

sjain-stanford (approved):
Thanks for supporting this. LGTM besides two minor comments.

python/torch_mlir/extras/fx_importer.py
stellaraccident: We will have to see. At best, it might be possible to retrofit to 2.2 but I'm not sure it is worth it. Further, the API is still evolving so we need to keep this separated at least until the 2.3 release so we can keep adapting it.

There are limits to what features can be supported on old versions as torch itself was missing the needed sophistication. I suspect that the narrow use case of frozen, stateless graphs is all we will manage for the older versions.

test/python/fx_importer/v2.3/mutation_import.py
stellaraccident: I can try, but before they fixed this, it was undefined and the result was different between versions.

stellaraccident: I'm going to punt on the negative test. I wasn't able to make the old versions behave consistently. So I think this is just unsupported (and it messes up in different, incidental ways I don't want to firefight).

aartbik (commented):
lgtm

PR #2918, opened 2024-02-16
[torch-mlir][sparse] inline sparse helper methods

aartbik:
Even though the reference compiler is not about performance, inlining the generated sparse helper methods has a rather big positive impact on performance, leaving a much better first impression. Therefore, we added this inlining pass (which leaves all other PyTorch modules unaffected, since they tend to be one big main() method to start with).

testing:

$./tools/e2e_test.sh --config linalg

Summary:
    Passed: 1164
    Expectedly Failed: 8

$ python -m e2e_testing.main --config=torchdynamo

Summary:
    Passed: 976
    Expectedly Failed: 162


PeimingLiu (approved):

yinying-lisa-li (approved):

stellaraccident (approved):
Seems ok to me

PR #2921, opened 2024-02-17
[torch] Add folder for `prim.NumToTensor.Scalar`

rsuderman:
Useful for `slice` lowerings that depend on tensors made form scalars.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -3647,7 +3647,7 @@ OpFoldResult PrimMaxIntOp::fold(FoldAdaptor adaptor) {
 
 OpFoldResult PrimNumToTensorScalarOp::fold(FoldAdaptor adaptor) {
   Attribute a = adaptor.getA();
-  auto resultTy = cast<ValueTensorType>(getType());
+  auto resultTy = cast<BaseTensorType>(getType());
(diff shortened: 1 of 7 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -3428,6 +3428,30 @@ OpFoldResult PrimMaxIntOp::fold(FoldAdaptor adaptor) {
...
+    return {};
+
+  auto dty = resultTy.getDtype();
+  if (auto iattr = dyn_cast<IntegerAttr>(a)) {
+    a = IntegerAttr::get(dty, iattr.getInt());
+  } else if (auto fattr = dyn_cast<FloatAttr>(a)) {
+    a = FloatAttr::get(dty, fattr.getValueAsDouble());
+  }
+
+  auto builtin =
```
renxida, line 3450: It might be better to say "llvmTensorType" instead of "builtin" - - just to leave some breadcrumbs for anybody who might use this to learn to work on folders in the future.

renxida (approved):

lib/Dialect/Torch/IR/TorchOps.cpp
rsuderman: Seems reasonable, though I named it `mlirTensorType` as thats more accurate.

PR #2925, opened 2024-02-19
Aten_LocalScalarDense conversion to linalg

Max191:
This is added as a ConvertAtenTensorToScalarLikeOp conversion in torch to linalg, lowering into a tensor.extract.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -10324,7 +10324,7 @@ def Torch_Aten_LocalScalarDenseOp : Torch_Op<"aten._local_scalar_dense", [
   ]> {
   let summary = "Generated op for `aten::_local_scalar_dense : (Tensor) -> (Scalar)`";
   let arguments = (ins
-    AnyTorchTensorType:$self
+    AnyTorchTensorType:$a
   );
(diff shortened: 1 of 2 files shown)


Max191:
I'm not sure how to create an e2e test for this that will actually generate an `aten._local_scalar_dense` operation. Is there a torch function I can use in a test that will generate this? (`item()` does not seem to work).

rsuderman:
Please add a test to the torch test suite

rsuderman (changes requested):
Looks fine however having a test would give me a better idea if it works

Max191:
> Looks fine however having a test would give me a better idea if it works

I tried to add a test, but it didn't seem to generate the _local_scalar_dense op. Do you know what torch op would generate this when we export it? When I ran into the op it was generated by `item()`, but I couldn't get that to generate the right op in my test.

rsuderman:
> calar_dense op. Do you know what torch op would generate this when we export it? When I ran into the op it was generated by

Having trouble generating the test. Seems like it depends on tracing. I'll stamp as it does not seem worth the additional hassle.

rsuderman (approved):

Max191:
rebased

Max191:
@rsuderman I don't understand the build failure here, and I'm unable to reproduce it on my machine. Do you know what the failure is?

PR #2931, opened 2024-02-20
[onnx] Update the importer to create a `none` for missing operands

rsuderman:
Some operands are optional so we require a placeholder for missing operands. We invent an `onnx.None` operation as our placeholder.


stellaraccident:
I assume that !torch.none and torch.constant.none are too hard to use? 

I'm ok with a one off here if needed. Just asking/suggesting

rsuderman:
> I assume that !torch.none and torch.constant.none are too hard to use?
> 
> I'm ok with a one off here if needed. Just asking/suggesting

It should be possible to use `torch.constant.none`. Let me give it a quick try.

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -272,6 +274,20 @@ def import_all(self):
         with InsertionPoint(self._b), Location.unknown():
             func_dialect.ReturnOp(outputs)
 
+    def get_none(self):
+        if '' in self._nv_map:
```
stellaraccident, line 278: Just use None as a key. Better to not share the stringy type namespace with onnx

stellaraccident: Can just change the type of _nv_map to `dict[Optional[str], Value]`

stellaraccident (approved):
Nit suggestion. Others, lgtm

python/torch_mlir/extras/onnx_importer.py
rsuderman: > Just use None as a key. Better to not share the stringy type namespace with onnx

So this may seem weird but thats actually on purpose. `onnx` uses empty string as the marker for a missing parameters. By using that as our key we do not need to do anything special for grabbing the `none` value as a parameter.

PR #2934, opened 2024-02-20
allow tosa.cast to convert from f32 to f16

srinathava:
According to the [official TOSA spec](https://www.mlplatform.org/tosa/tosa_spec.html#_cast), `tosa.cast` allows a cast from `fp32` to `fp16`. We were not previously accounting for this in the `TorchToTosa` lowering.

Also did a tiny bit of cleanup in the code to make it easier to spot which conversions are currently allowed.

### DIFF
--- a/lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
+++ b/lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
@@ -266,23 +266,33 @@ std::optional<Value> getConstTensor<float>(PatternRewriter &rewriter,
 }
 
 static LogicalResult checkValidityOfCast(Type src, Type dest) {
-  if ((src == dest) || (src.isInteger(64) && dest.isInteger(32)) ||
+  if ((src == dest) ||
+      // int64 -> *
+      (src.isInteger(64) && dest.isInteger(32)) ||
       (src.isInteger(64) && dest.isInteger(8)) ||
       (src.isInteger(64) && dest.isInteger(1)) ||
       (src.isInteger(64) && dest.isF32()) ||
+      // int32 -> *
       (src.isInteger(32) && dest.isInteger(64)) ||
       (src.isInteger(32) && dest.isInteger(1)) ||
       (src.isInteger(32) && dest.isF32()) ||
       (src.isInteger(32) && dest.isBF16()) ||
+      // int16 -> *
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
```diff
@@ -266,23 +266,33 @@ std::optional<Value> getConstTensor<float>(PatternRewriter &rewriter,
...
       (src.isInteger(8) && dest.isInteger(1)) ||
       (src.isInteger(8) && dest.isBF16()) ||
+      // int1 -> *
       (src.isInteger(1) && dest.isInteger(64)) ||
-      (src.isInteger(1) && dest.isF32()) || (src.isF32() && dest.isF64()) ||
-      (src.isF32() && dest.isBF16()) || (src.isF64() && dest.isF32()) ||
-      (src.isF64() && dest.isBF16()) || (src.isF32() && dest.isInteger(8)) ||
+      (src.isInteger(1) && dest.isF32()) ||
+      // f64 -> *
+      (src.isF64() && dest.isF32()) || (src.isF64() && dest.isBF16()) ||
```
sjain-stanford, line 289: minor nit: consider breaking this into multiple lines (like the int cases above).

sjain-stanford (commented):
Thanks @srinathava. LGTM besides a minor nit. Let's wait for @sjarus to review.

lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
srinathava: To get these to stay on the same line, I needed to do a `clang-format off/on` around this.

sjarus (approved):
Looks reasonable to me. A possible concern is that AFAIK there's no defined protocol in TOSA as to how to convert [fp32 to fp16](https://www.corsix.org/content/converting-fp32-to-fp16) but this should still be a reasonable abstract expression before that's defined. 

cc: @eric-k256

eric-k256 (approved):
We've gotten better at defining the behavior when casting between floating point types. If you look at https://www.mlplatform.org/tosa/tosa_spec.html#_main_inference_profile_compliance
There is a table that includes the accuracy requirements for CAST. Also at the top it mentions NaNs should be preserved in implementations with FP inputs. 

But overall this looks fine.

PR #2935, opened 2024-02-20
Migrate passes in TorchConversion to use FunctionOpInterface.

stellaraccident:
This enables better re-use in downstreams which use different func implementations and should have no impact on those that don't except in opt pipelines if using the old form. With interfaces, explicit pipelines via `--pass-pipeline=` must be used.


Groverkss (approved):
LGTM

PR #2939, opened 2024-02-21
[torch-mlir][sparse] support e2e sparse kernels with COO inputs.

PeimingLiu:
**DO NOT MERGE!**: Still waiting llvm-project be synced with https://github.com/llvm/llvm-project/pull/82449

### DIFF
--- a/test/python/fx_importer/sparse_test.py
+++ b/test/python/fx_importer/sparse_test.py
@@ -292,7 +292,6 @@ def test_sparse_SpMM():
     print(m)
 
     # Run it with PyTorch torch.sparse and with TORCH-MLIR sparse_jit.
-    sparse_input = dense_input.to_sparse_coo()
     net = MatMulNet()
     res1 = net(sparse_input, dense_input)
     res2 = sparse_jit(net, sparse_input, dense_input)
--- a/test/python/fx_importer/sparse_test.py
+++ b/test/python/fx_importer/sparse_test.py
@@ -161,7 +161,14 @@ def sparse_jit(f, *args, **kwargs):
     for a in args:
         if a.layout is torch.sparse_coo:
             xargs.append(a.values().numpy())
-            xargs.append(a.indices().numpy())
+            # Construct the additional position array required by MLIR with data
+            # array([0, nnz]).
+            nnz = a.indices().shape[1]
+            xargs.append(torch.tensor([0, nnz], dtype=a.indices().dtype).numpy())
+            # Transform a tensor<ndim x nnz> into [tensor<nnz> x ndim] to conform
+            # MLIR SoA COO representation.
+            for idx in a.indices():
+                xargs.append(idx.numpy())
         elif a.layout is torch.sparse_csr or a.layout is torch.sparse_bsr:
             xargs.append(a.values().numpy())
             xargs.append(a.crow_indices().numpy())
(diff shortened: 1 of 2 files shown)


test/python/fx_importer/sparse_test.py
```diff
@@ -158,7 +158,14 @@ def sparse_jit(f, *args, **kwargs):
     for a in args:
         if a.layout is torch.sparse_coo:
             xargs.append(a.values().numpy())
-            xargs.append(a.indices().numpy())
+            # Construct the additional position array required by MLIR with data
+            # array([0, nnz]).
+            nnz = a.indices().shape[1]
```
aartbik, line 163: I have been puzzled why torch.sparse does not just expose the nnz() as a method :-)

test/python/fx_importer/sparse_test.py
```diff
@@ -158,7 +158,14 @@ def sparse_jit(f, *args, **kwargs):
...
+            xargs.append(torch.tensor([0, nnz], dtype=a.indices().dtype).numpy())
```
aartbik, line 164: why not do this directly in numpy?

test/python/fx_importer/sparse_test.py
```diff
@@ -247,14 +258,17 @@ def forward(self, x, y):
     print(m)
 
     # Run it with PyTorch torch.sparse and with TORCH-MLIR sparse_jit.
-    # TODO: run with COO, right now only CSR works
     sparse_input = dense_input.to_sparse_csr()
```
aartbik, line 261: remove this csr() and res3

(that was the original intent of this test, i.e. run it only in the way we CHECK the IR above)

aartbik (commented):
Make sure MLIR bumps to right hash first

test/python/fx_importer/sparse_test.py
```diff
@@ -158,7 +158,14 @@ def sparse_jit(f, *args, **kwargs):
...
             xargs.append(a.values().numpy())
-            xargs.append(a.indices().numpy())
+            # Construct the additional position array required by MLIR with data
+            # array([0, nnz]).
+            nnz = a.indices().shape[1]
+            xargs.append(torch.tensor([0, nnz], dtype=a.indices().dtype).numpy())
+            # Transform a tensor<ndim x nnz> into [tensor<nnz> x ndim] to conform
+            # MLIR SoA COO representation.
+            for idx in a.indices():
+                xargs.append(idx.numpy())
```
aartbik, line 168: can this be done with a view change and then the append? also, does this generalize to n-dim COO?

PeimingLiu: I am hoping that it is already a view change (but I am not sure).

I think it generalize to n-dim COO. Per torch.sparse documentation " the indices of specified elements are collected in indices tensor of size (ndim, nse)"

test/python/fx_importer/sparse_test.py
aartbik: actually _nnz() seems to work ;-)

test/python/fx_importer/sparse_test.py
PeimingLiu: Per offline discussion, we will keep it this way to avoid clumsy conversion between `torch.dtype` to `numpy.dtype`

test/python/fx_importer/sparse_test.py
PeimingLiu: I dumped the `data_ptr()` of `a.indices()` and `idx` used in the loop and confirm that they are alias.

test/python/fx_importer/sparse_test.py
```diff
@@ -247,8 +253,7 @@ def forward(self, x, y):
     print(m)
 
     # Run it with PyTorch torch.sparse and with TORCH-MLIR sparse_jit.
-    # TODO: run with COO, right now only CSR works
-    sparse_input = dense_input.to_sparse_csr()
+    sparse_input = dense_input.to_sparse_coo()
```
aartbik, line 256: remove L256 completely, it is already in L251
(that way, it has the same structure as the first test above)

aartbik:
bumped, please update your commit message before submitting

aartbik (approved):

PR #2941, opened 2024-02-22
[MLIR][TORCH] Add support for tanh approximation for Gelu op

vivekkhandelwal1:
Fixes https://github.com/nod-ai/SHARK-Turbine/issues/461

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -511,11 +511,42 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     }
     // TODO: Take approximation into account.
     std::string approximate;
-    if (!matchPattern(gelu.getApproximate(), m_TorchConstantStr(approximate)) ||
-        approximate != "none")
+    if (!matchPattern(gelu.getApproximate(), m_TorchConstantStr(approximate))) {
+      gelu.emitError(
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -511,11 +511,41 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
     }
     // TODO: Take approximation into account.
     std::string approximate;
-    if (!matchPattern(gelu.getApproximate(), m_TorchConstantStr(approximate)) ||
-        approximate != "none")
+    if (!matchPattern(gelu.getApproximate(), m_TorchConstantStr(approximate))) {
```
rsuderman, line 514: Nit: Given there is only one shared operator (`MulFOp`). We might as well check cases and bail early. It makes the control flow a little easier to read.

```
if (approximate == "none") {
  Value multiplier = buildUnitNormalCdf(b, loc, payloadArgs[0]);
  return b.create<arith::MulFOp>(loc, payloadArgs[0], multiplier);
}

if (approximate == "tanh") {
  ....
  return b.create<arith::MulFOp>(loc, payloadArgs[0], multiplier);
}


gelu.emitError("unimplemented: approximate value should be none or tanh");
return nullptr;
```

rsuderman (approved):

PR #2842, opened 2024-01-31
Rename torch_mlir.compile APIs and introduce FX based analogs

saienduri:
Link to related RFC: https://discourse.llvm.org/t/rfc-rename-torch-mlir-compile-apis-and-introduce-fx-based-analogs/76646
This commit updates the documentation, tests, CMake files, and API for the proposed changes in the RFC. There is a new torch_mlir/fx.py for user level APIs related to importing modules and a corresponding test for this path can be found at test/python/fx_importer/basic_test.py.

PR #2944, opened 2024-02-23
updated examples based on the current init from torch_mlir

VimalWill:
updated the torchscript based examples in the projects/p1/examples


newling:
This doesn't look correct to me, I think we changed to import torchscript recently https://github.com/llvm/torch-mlir/pull/2842

VimalWill:
@newling I checked the init of the torch_mlir which is installed from the commands in the Readme. There it's not from torchscript, so I though of updating it

newling:
Thanks @VimalWill
@saienduri can you please for advise?

PR #2951, opened 2024-02-26
[onnx] Import `onnx` import to pass remaining tests

rsuderman:
Finish supporting importing the vast majority of `onnx` operations. This includes:
- region support
- region value inherentance
- `torch.string` support
- `torch.list` support
- `torch.optional` support


test/python/onnx_importer/import_smoke_test.py
```diff
@@ -44,155 +44,10 @@
...
 
 TEST_CAST_XFAILS = [
-    "light_light_bvlc_alexnet",
-    "light_light_inception_v1",
-    "light_light_squeezenet",
-    "light_light_vgg19",
-    "node_test_affine_grid_2d_align_corners_expanded_model",
-    "node_test_affine_grid_2d_expanded_model",
-    "node_test_affine_grid_3d_align_corners_expanded_model",
-    "node_test_affine_grid_3d_expanded_model",
```
ScottTodd, line 54: Nice! These should also start passing in the downstream test suite I've been setting up for IREE to use: https://github.com/nod-ai/SHARK-TestSuite/blob/main/iree_tests/onnx/node/import_failures.txt

externals/llvm-project
ScottTodd, line 1: Does torch-mlir have a policy for updating the llvm-project submodule? Should that be moved out of this PR?

ScottTodd (approved):

test/python/onnx_importer/import_smoke_test.py
rsuderman: Yup, should be expected. Hopefully this should be the last major changes there.

PR #2954, opened 2024-02-27
[MLIR][TORCH] Fix OnnxToLinalg lowering issue for sub and sum op

vivekkhandelwal1:
This commit adds the support for scalar conversion to byte. 
This commit also fixes the OnnxToLinalg lowering issue for Onnx.Sub and Onnx.Sum op.
Fixes https://github.com/nod-ai/SHARK-Turbine/issues/466 
Fixes https://github.com/nod-ai/SHARK-Turbine/issues/467

### DIFF
--- a/lib/Conversion/Utils/Utils.cpp
+++ b/lib/Conversion/Utils/Utils.cpp
@@ -245,12 +245,20 @@ mlir::RankedTensorType GetTypeFromTensorShape(llvm::ArrayRef<int64_t> shape,
                                      elementType, encoding);
 }
 
+static std::optional<int64_t> getIntegerValue(Value scalar) {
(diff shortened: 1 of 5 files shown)


lib/Conversion/Utils/Utils.cpp
```diff
@@ -245,12 +245,20 @@ mlir::RankedTensorType GetTypeFromTensorShape(llvm::ArrayRef<int64_t> shape,
                                      elementType, encoding);
 }
 
+static int64_t getIntegerValue(Value scalar) {
+  if (auto constOp = scalar.getDefiningOp<Torch::ConstantIntOp>()) {
+    return (int64_t)constOp.getValue();
+  }
+  return -1;
```
rsuderman, line 252: You should return an `std::optional<int64_t>` instead of using a special value. Its safer to unknown values and clearer why you check for -1

PR #2959, opened 2024-02-28
[tosa] Fix TOSA batch matmul lowering to correct transpose ordering

rsuderman:
The corrective transpose at the end is computed incorrectly. Is it actually computin the inverse transpose. Inverting the permutations fixes the issue.

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -1234,7 +1234,7 @@ public:
       return false;
     };
 
-    SmallVector<TensorShape_t> commonElems, lhsSqueezedElems, rhsSqueezedElems;
+    SmallVector<TensorShape_t> batchElems, lhsSqueezedElems, rhsSqueezedElems;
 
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -1234,7 +1234,7 @@ class ConvertAtenMatmulBaseOp : public OpConversionPattern<AtenOpT> {
       return false;
     };
 
-    SmallVector<TensorShape_t> commonElems, lhsSqueezedElems, rhsSqueezedElems;
```
renxida, line 1237: why was this named commonElems originally?

renxida (approved):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
rsuderman: The name was pretty unclear - I think what they meant was dimensions commonly shared between the LHS and RHS but not contracted on. By definition these are the batch dimensions so I just renamed.

PR #2963, opened 2024-02-28
[torch]  Fix folders and `cat` and `view` torch lowerings

rsuderman:
A bunch of small fixes are interlinked and trigger crashes if not addressed as a group. This includes:

- aten view when expand from a rank-0 tensor
- slice folder with negative indices
- `aten._shape_as_tensor` folder on a rank-0 tensor
- `aten.cat` of a tensor with a length-0 tensor

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -2972,8 +2972,10 @@ OpFoldResult AtenSliceTensorOp::fold(FoldAdaptor adaptor) {
       unaryNonDim &= inType.getSizes()[i] == 1 || i == dimInt;
     }
     if (unaryNonDim) {
-      Attribute value =
-          input.getValues<Attribute>()[start.getValue().getSExtValue()];
+      int64_t idx = start.getValue().getSExtValue();
+      if (idx < 0)
+        idx += input.getNumElements();
+      Attribute value = input.getValues<Attribute>()[idx];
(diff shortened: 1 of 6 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -3409,25 +3439,25 @@ OpFoldResult AtenItemOp::fold(FoldAdaptor adaptor) {
 OpFoldResult AtenOnesOp::fold(FoldAdaptor adaptor) {
   SmallVector<int64_t> sizes;
   if (!matchPattern(getSize(), m_TorchListOfConstantInts(sizes))) {
-    LLVM_DEBUG(llvm::dbgs() << "Failing to fold AtenOnesOp: size operand is "
```
renxida, line 3412: I've been wondering. Why do we not see more of these debug statements.

Since it's using the LLVM_DEBUG macro, they doesn't make it into non debug builds anyway.

renxida (approved):

lib/Dialect/Torch/IR/TorchOps.cpp
rsuderman: You want to be strategic where you put them. Too many messages and you fill the debug logs with garbage. In this case every time you run `canonicalize` one of these would print for every matching operation (as you print the bail out). This would make debugging compilation runs of ever moderately sized test cases basically unusable. So while we can include them per pass, its usually bad per rewriter / folder.

PR #2968, opened 2024-02-29
[onnx] Fix type on create_module() in onnx_importer.py.

ScottTodd:
The type returned was changed in https://github.com/llvm/torch-mlir/pull/2795. This led to errors in the downstream IREE project: https://github.com/openxla/iree/pull/16622.

### DIFF
--- a/python/torch_mlir/extras/onnx_importer.py
+++ b/python/torch_mlir/extras/onnx_importer.py
@@ -99,12 +99,12 @@ class ModelInfo:
         assert model_proto.graph, "Model must contain a main Graph"
         self.main_graph = GraphInfo(self, model_proto.graph)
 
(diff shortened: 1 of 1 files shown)


stellaraccident (approved):

stellaraccident:
I think it was better the original way, but I don't care enough to break the API again. We should have a higher battier for API breaks in the future.

ScottTodd:
> I think it was better the original way, but I don't care enough to break the API again. We should have a higher battier for API breaks in the future.

Agreed. I missed it during the review but am getting more familiar with torch-mlir. One thing that could have helped (@rsuderman ) would be splitting the PR into two changes - one for any prerequisite API changes and another for the new tests.

PR #2971, opened 2024-03-02
[torch] Make torch.aten.flatten lower directly to linalg

rsuderman:
Existing lowering via aten.view does not work as well for dynamic shapes as the lowering to tensor.expand must re-infer dynamic shape matching. Better to directly lower.


ScottTodd (approved):

vivekkhandelwal1:
Hi @rsuderman, shouldn't we remove this lowering? https://github.com/vivekkhandelwal1/torch-mlir/blob/a86e89ecb5c7929a39a38743fb7cacadf1ff41bb/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L2465-L2555

Since you've added a direct lowering to Linalg.

PR #2972, opened 2024-03-02
[torch] Additional folders for shape computations

rsuderman:
A handful of operations are commonly used in shape calculations (slice, concat, broadcast). Added these additional folders to better propagate simple shape computations.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -2900,11 +2900,55 @@ OpFoldResult AtenSubIntOp::fold(FoldAdaptor adaptor) {
 
 OpFoldResult AtenCatOp::fold(FoldAdaptor adaptor) {
   auto list = getOperand(0).getDefiningOp<PrimListConstructOp>();
-  if (!list || list.getElements().size() != 1)
+  if (!list)
     return nullptr;
-  if (list.getElements()[0].getType() != getResult().getType())
+
+  auto elements = list.getElements();
+  if (elements.size() == 1 && elements[0].getType() == getResult().getType())
+    return list.getElements()[0];
+
+  auto resultTy = dyn_cast<ValueTensorType>(getType());
+  if (!resultTy || !resultTy.hasSizes() || !resultTy.hasDtype())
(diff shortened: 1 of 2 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2900,11 +2900,55 @@ OpFoldResult AtenSubIntOp::fold(FoldAdaptor adaptor) {
...
+    return list.getElements()[0];
+
+  auto resultTy = dyn_cast<ValueTensorType>(getType());
+  if (!resultTy || !resultTy.hasSizes() || !resultTy.hasDtype())
     return nullptr;
-  return list.getElements()[0];
+
+  auto bResultTy = resultTy.toBuiltinTensor();
+  if (!bResultTy.hasStaticShape() || bResultTy.getNumElements() > 16)
+    return nullptr;
```
ScottTodd, line 2916: Why 16 elements? Is that arbitrary, or is there a specific reason?

test/Dialect/Torch/canonicalize.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: torch-mlir-opt %s -canonicalize | FileCheck %s
+// RUN: torch-mlir-opt %s -canonicalize --split-input-file | FileCheck %s
```
ScottTodd, line 1: Oof.. this file is 2000+ lines and wasn't using `--split-input-file`? Can you add the `// -----` blocks to the rest of the file? Maybe in a separate PR to keep history cleaner.

lib/Dialect/Torch/IR/TorchOps.cpp
rsuderman: Somewhat arbitrary - its a reasonable upper bound for shape computations.

test/Dialect/Torch/canonicalize.mlir
rsuderman: Follow up would make more sense. We have been incrementally inserting them as models grow. Alternatively someone could write a script to quickly add.

ScottTodd (approved):

PR #2974, opened 2024-03-04
[FxImporter] remove dataclass slots to support python3.9

qingyunqu:
* that `dataclass`'s `slots` is supported after python 3.10.

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -315,6 +315,7 @@ def is_builtin_function_or_method(obj: Any) -> bool:
     return isinstance(obj, (BuiltinMethodType, BuiltinFunctionType))
 
 
+# TODO: switch back to `slots=True` when py3.9 support is dropped
 @dataclass(frozen=True)
 class InputInfo:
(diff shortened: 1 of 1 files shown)


python/torch_mlir/extras/fx_importer.py
```diff
@@ -315,15 +315,23 @@ def is_builtin_function_or_method(obj: Any) -> bool:
     return isinstance(obj, (BuiltinMethodType, BuiltinFunctionType))
 
 
-@dataclass(frozen=True, slots=True)
+@dataclass(frozen=True)
 class InputInfo:
     """Provides additional metadata when resolving inputs."""
 
+    __slots__ = [
```
sjain-stanford, line 322: nit: perhaps add a `TODO` here to switch back to using `slots=True` when py-3.9 support is dropped?

sjain-stanford (approved):
LGTM besides a comment request.

python/torch_mlir/extras/fx_importer.py
qingyunqu: done

PR #2976, opened 2024-03-04
[onnx] Handle optional arguments in Clip op pattern.

ScottTodd:
Spec: https://onnx.ai/onnx/operators/onnx__Clip.html


lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -603,34 +603,35 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+
         if (binder.op->getNumOperands() == 1) {
-          Value source;
-          if (binder.tensorOperand(source) ||
-              binder.tensorResultType(resultType))
-            return failure();
           Value cstNone =
               rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
           rewriter.replaceOpWithNewOp<Torch::AtenClampOp>(
               binder.op, resultType, source, /*min=*/cstNone, /*max=*/cstNone);
```
ScottTodd, line 625: Seems like this could use a different op instead of AtenClamp with two empty parameters. Suggestions?

rsuderman: Yeah you might as well just do a replacement with the input and erase the operation

projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -1917,14 +1917,6 @@
...
     # Failure - onnx_lowering: onnx.Clip
-    "ElementwiseClampMaxModule_basic",
-    "ElementwiseClampMinModule_basic",
-    "ElementwiseClampMinTensorFloatModule_basic",
-    "ElementwiseClampMinTensorIntModule_basic",
-    "ElementwiseClampModule_basic",
-    "ElementwiseClampTensorFloatModule_basic",
-    "ElementwiseClampTensorInt8Module_basic",
-    "ElementwiseClampTensorIntModule_basic",
     "NormalizeModule_basic",
```
rsuderman, line 1920: I will try to reprocess the file this week so we know what the new blocker is on this issue.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -603,34 +603,35 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+
+        // Inputs and outputs must be tensors. Min and max are both optional
+        // and if provided must be "tensors of empty shape" (scalars).
+        Value source;
         Torch::ValueTensorType resultType;
+        if (binder.tensorOperandAtIndex(source, 0) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+
```
rsuderman, line 620: I would restructure this to avoid having exit points decided by `getNumOperands`. We know that there are versions of `clip` that include attribute defined values. Do something like the following

```
Value min, max;
if (binder.op->getNumOperands() > 1) {
  if (binder.tensorOperandAtIndex(min, 1)
    return rewriter.notifyMatchFailure(binder.op, "error message");
}

if (binder.op->getNumOperands() > 2) {
  if (binder.tensorOperandAtIndex(max, 2)
    return rewriter.notifyMatchFailure(binder.op, "error message");
}

// Can insert attribute based checking with value formation here.

if (!min && !max) {
  rewriter.replaceOp(binder.op, operand);
}

// Min case

// Max case

// Clamp case
```

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
ScottTodd: >  We know that there are versions of clip that include attribute defined values.

Would you like those to be supported? That would be version 6, when this op lowering is currently using a minimum version of 13. I figure the ONNX Version Converter (https://github.com/onnx/onnx/blob/main/docs/VersionConverter.md) should be used to support older versions, rather than include potentially complicated switching logic both here _and_ in that upstream tool.

ScottTodd: > ```c++
> if (binder.op->getNumOperands() > 1) {
>   if (binder.tensorOperandAtIndex(min, 1)
>     return rewriter.notifyMatchFailure(binder.op, "error message");
> }
> ```
> 

These also have the same bug as before - the operands can be 'none', which is not a tensor type (? I'm not sure exactly what type, should run under a debugger, but the `tensorOperandAtIndex` check was failing)

ScottTodd: Tried restructuring how you suggested but it got messy and loose with validation (https://github.com/ScottTodd/torch-mlir/commit/b6c307b023b52adc3a46e09c33c16298fc6d52e9).

Instead, I added a second pattern, like what is done here: https://github.com/llvm/torch-mlir/blob/933db87a07fb828b17b75f2f5f396a434f8f1a17/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp#L1038-L1040

How does this look now?

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -601,40 +601,75 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+              /*max=*/binder.op->getOperand(2));
           return success();
         }
         return failure();
       });
+  // onnx.Clip with min/max as attributes
+  patterns.onOp(
+      "Clip", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        // https://onnx.ai/onnx/operators/onnx__Clip.html
+        // Note: version 6 has optional attributes with default min/max values.
```
ScottTodd, line 643: (I'd still prefer to drop this - opset 11 is old, and it doesn't seem worth duplicating logic already in the version converter)

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -601,40 +601,75 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
                       binder.op, resultType, operand);
                   return success();
                 });
+  // onnx.Clip with min/max as arguments introduced in opset 11
   patterns.onOp(
-      "Clip", 13, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+      "Clip", 11, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        // https://onnx.ai/onnx/operators/onnx__Clip.html
+        // Note: version 13 added 'bfloat16' support
```
rsuderman, line 608: Anything that is just superset behavior (e.g. adding type support) we can just support without comments. I wouldn't both including comments as even if a type were not support (e.g. `bfloat16`) it wouldn't change how our lowering work.s

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
rsuderman: You don't have to duplicate an additional rewriter, you just insert a check for the attributes and operand creator. Generally we want to avoid recursive invocation of rewriters when possible.

https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp#L1178

I can promise you we need to support this. Some of the models I have debugged have already had the attribute based passing mechanism. Relying on the external tooling to upgrade will not guarantee support. The `ReduceMin` example used attributes for defining the axes despite being version 13. Old != unused.

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
rsuderman: Try to embed similar to ReduceMin. The omni-rewriter is a weird restriction we run in and recursion limits applying a single monolithic rewrite pattern could cause problems.

https://github.com/llvm/torch-mlir/blob/main/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp#L1178

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
ScottTodd: > You don't have to duplicate an additional rewriter, you just insert a check for the attributes and operand creator.

Alright, I can try to clean up https://github.com/ScottTodd/torch-mlir/commit/b6c307b023b52adc3a46e09c33c16298fc6d52e9 ... it's not pretty :/

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
ScottTodd: Refactored a bit. LMK what you think... seems to be getting better.

PR #2977, opened 2024-03-04
[torch-mlir][sparse] support cases where batch_dim != 0

PeimingLiu:
**DO NOT MERGE** until syncing to https://github.com/llvm/llvm-project/pull/83898

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -294,7 +294,7 @@ def sparsity_encoding(shape: torch.Size, sparsity: SparsityMeta) -> str:
         )
 
     if batch_dim > 0:
-        batch = ",".join(f"d{d}:dense" for d in range(0, batch_dim))
+        batch = ",".join(f"d{d}:batch" for d in range(0, batch_dim))
         lvls = f"{batch},{lvls}"
 
     if dense_dim > 0:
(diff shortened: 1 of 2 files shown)


yinying-lisa-li (approved):

test/python/fx_importer/sparse_test.py
```diff
@@ -358,11 +364,15 @@ def forward(self, x):
...
-    #  (2) for dense_dim=0, this will need a dense(batched) property
     sparse_input = dense_input.to_sparse_csr(dense_dim=1)
     res1 = net(sparse_input)
     res2 = sparse_jit(net, sparse_input)
     print("torch.sparse")
     print(res1)
     print("torch.mlir")
     print(res2)
+
+    batch_input = dense_input.to_sparse_csr(dense_dim=0)
```
aartbik, line 375: can you rename L354 to batch_input and use it here
then also, remove L367

that way we nicely have setting up the IR with the same input as used for the JIT

aartbik (approved):
One request, but tears in my eyes so pretty!

(needs MLIR bump)

python/torch_mlir/extras/fx_importer.py
```diff
@@ -294,7 +294,7 @@ def sparsity_encoding(shape: torch.Size, sparsity: SparsityMeta) -> str:
         )
 
     if batch_dim > 0:
-        batch = ",".join(f"d{d}:dense" for d in range(0, batch_dim))
+        batch = ",".join(f"d{d}:batch" for d in range(0, batch_dim))
```
arcturusannamalai, line 297:
```suggestion
        batch = ",".join(f"d{d}:batch" for d in range(batch_dim))
```

PR #2979, opened 2024-03-05
[torch] Add support for `torch.split_with_sizes` via decompose

rsuderman:
Convert to individiual slices and tuple together as a list.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -772,7 +772,7 @@ class DecomposeAtenSplitWithSizesOp
     int64_t rank = selfTy.getSizes().size();
     auto resultTy = dyn_cast<Torch::ListType>(op.getResult().getType());
     if (!resultTy)
-      return rewriter.notifyMatchFailure(op, "Result tpe not a list");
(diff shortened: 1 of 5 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -693,6 +693,131 @@ class DecomposeAtenSelectIntOp : public OpRewritePattern<AtenSelectIntOp> {
...
+      return rewriter.notifyMatchFailure(op, "No split sizes");
+
+    auto selfTy = dyn_cast<BaseTensorType>(self.getType());
+    if (!selfTy || !selfTy.hasSizes())
+      return rewriter.notifyMatchFailure(op, "Self shape unknown");
+
+    int64_t rank = selfTy.getSizes().size();
+    auto resultTy = dyn_cast<Torch::ListType>(op.getResult().getType());
+    if (!resultTy)
+      return rewriter.notifyMatchFailure(op, "Result tpe not a list");
```
ScottTodd, line 775:
```suggestion
    if (!resultTy)
      return rewriter.notifyMatchFailure(op, "Result type not a list");
```

ScottTodd (approved):

PR #2980, opened 2024-03-05
[torch-mlir][sparse] fix for uncoming upstream update

PeimingLiu:
to fix future breakage caused by https://github.com/llvm/llvm-project/pull/84079


aartbik (approved):
Values at the end. The way all sparse engineers would like to see it ;-)
Good to go when bumped.

yinying-lisa-li (approved):

PR #2986, opened 2024-03-06
Implement lowering of torch.aten.linalg_cross

ptrifunovic98:
Closes [nod-ai/SHARK-Turbine#497](https://github.com/nod-ai/SHARK-Turbine/issues/497)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4062,6 +4062,101 @@ LogicalResult AtenPermuteOp::verify() {
...
+  if (selfRank != otherRank) {
+    return emitOpError("input tensors must have the same number of dimensions, "
+                       "but got ")
+           << selfRank << " and " << otherRank;
+  }
+
+  int64_t dim;
+  if (!matchPattern(getDim(), m_TorchConstantInt(&dim))) {
+    return emitOpError("dim must be a constant int");
+  }
```
ramiro050, line 4100: nit: this is not part of the semantics of the op, so the verifier should not produce an error here. Normally we check for this in conversion patterns because it either simplifies the pattern, or because the compile-time information is necessary to lower to the target backend dialect. However, it is valid for ops to have non-constant arguments (sometimes they might get folded away).

One option is to return success if we are unable to get the constant value. This way you're able to keep all the static checks below

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4062,6 +4062,101 @@ LogicalResult AtenPermuteOp::verify() {
...
+    return success();
+  }
+
+  Type selfDtype = selfType.getDtype();
+  Type otherDtype = otherType.getDtype();
+
+  auto selfRankedType = getSelf().getType().cast<RankedTensorType>();
+  auto otherRankedType = getOther().getType().cast<RankedTensorType>();
+
+  auto selfShape = selfRankedType.getShape();
```
ramiro050, line 4084: shapes are also not guaranteed. Needs to check `hasShape()`

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4062,6 +4062,101 @@ LogicalResult AtenPermuteOp::verify() {
...
+  }
+
+  // compensate for possible negative dim value
+  if (dim < 0) {
+    dim += selfRank;
+  }
+
+  // check if the size of the dimensions specified by 'dim' is equal to 3
+  // (required by the operation)
+  if (selfShape[dim] != 3 || otherShape[dim] != 3) {
```
ramiro050, line 4115: nit: must check that it is not `kUnknownSize` (dynamic dimension)

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4062,6 +4062,101 @@ LogicalResult AtenPermuteOp::verify() {
...
+    return emitOpError("inputs dimension ")
+           << dim << " must have length 3, but got " << selfShape[dim]
+           << " and " << otherShape[dim];
+  }
+
+  // Check if any of the input tensors has torch.bool dtype.
+  // The operation does not support this type.
+  // The docs state that only float, double, cfloat and cdouble dtypes are
+  // supported, but, when testing, it fails only for boolean dtype. Update to
+  // fit the docs if necessary.
```
ramiro050, line 4125: nit: i think it would be good to have a link to the docs too

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4062,6 +4062,101 @@ LogicalResult AtenPermuteOp::verify() {
...
+    ++i;
+    ++selfCurrent;
+    ++otherCurrent;
+  }
+
+  // the operation succeeds only if both inputs have the same dtype
+  if (selfDtype != otherDtype) {
+    return emitOpError("input tensors must have the same dtype, but got ")
+           << selfDtype << " and " << otherDtype;
+  }
```
ramiro050, line 4155: nit: I would move this check up to where you get the dtype values. It is good to make checks as soon as the necessary information becomes available and exit early when the check fails to avoid performing unnecessary calculations.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4062,6 +4062,101 @@ LogicalResult AtenPermuteOp::verify() {
...
+
+  auto selfCurrent = selfShape.begin();
+  auto selfEnd = selfShape.end();
+  auto otherCurrent = otherShape.begin();
+  int32_t i = 0;
+
+  // Check if there is a disparity between dimension sizes.
+  // Dimensions at the same index must either have the same size,
+  // or one of them must be equal to 1.
+  while (selfCurrent != selfEnd) {
```
ramiro050, line 4138: nit: make this a `for` loop to make it explicit that the iteration is finite. You can do something like

```c++
for (auto [selfCurrent, otherCurrent] : llvm::zip_equal(selfShape, otherShape)) { ... }
```

projects/pt1/python/torch_mlir_e2e_test/test_suite/matmul.py
```diff
@@ -289,3 +289,94 @@ def forward(self, x, y):
 def AtenMmQuint8_basic(module, tu: TestUtils):
     module.forward(tu.randint(3, 4, low=-128, high=127).to(torch.int8),
                    tu.randint(4, 3, low=-128, high=127).to(torch.int8))
+    
+# ==============================================================================
+
+class AtenLinalgCrossInt(torch.nn.Module):
```
ramiro050, line 295: Does the lowering work with dynamic dimensions? If so, can you add at least one test case with dynamic dims?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1806,6 +1806,118 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+    Value dim = op.getDim();
+
+    auto resType = self.getType().cast<BaseTensorType>();
+    if (!resType.hasDtype()) {
+      return rewriter.notifyMatchFailure(op, "result should have dtype");
+    }
+
+    Type dtype = resType.getDtype();
+    if (dtype.isa<mlir::ComplexType>()) {
+      printf("Is a complex type\n");
```
ramiro050, line 1841: nit: remove print statements

ramiro050 (changes requested):
Looks pretty good, just some minor requests. Thanks!

ptrifunovic98:
@ramiro050, thanks for the review! I have updated everything you mentioned. The CI tests passed, so if everything looks good now, I think it is ready for merging.

ramiro050 (approved):
Thanks!

PR #2988, opened 2024-03-06
[ONNX][MLIR] Add support for onnx.gather_nd

Shukla-Gaurav:


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -558,6 +558,95 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+      "GatherND", 13, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value data, indices;
+        int64_t batch_dims;
+        if (binder.tensorOperandAtIndex(data, 0) ||
+            binder.tensorOperandAtIndex(indices, 1) ||
+            binder.tensorResultType(resultType) ||
+            binder.s64IntegerAttr(batch_dims, "batch_dims", 0))
+          return failure();
+        // TODO: Add support for non zero batch_dims.
```
rsuderman, line 571: I would prefer a plan for how to support batch dimensions before adding this lowering. If it turns out that we cannot back to `onnx.Gather` with batch dims then we might as well not have this lowering for the non-batch dims case.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -558,6 +558,95 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        flattenDataShape.insert(flattenDataShape.end(),
+                                dataShape.begin() + indicesLastDim,
+                                dataShape.end());
+        auto flattenDataTy = Torch::ValueTensorType::get(
+            ctx, llvm::ArrayRef(flattenDataShape), dataTy.getOptionalDtype());
+        Value zero = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(0));
+        Value end = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(indicesLastDim - 1));
+        Value flattenData = rewriter.create<Torch::AtenFlattenUsingIntsOp>(
```
rsuderman, line 639: Generally it is bad to flatten your data tensor from a gather as for embedding cases (where gathers are a common use) this would be reshaping a massive tensor where only a subsection of the data is used. Is there an alternative that does not involve manipulating the `data` type?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -558,6 +558,95 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        Value boolFalse = rewriter.create<Torch::ConstantBoolOp>(loc, false);
+        Type multiplierTy =
+            Torch::ValueTensorType::get(ctx, llvm::ArrayRef({indicesLastDim}),
+                                        indicesTy.getOptionalDtype());
+        Value multiplierTensor = rewriter.create<Torch::AtenTensorOp>(
+            loc, multiplierTy, indexMultiplierList,
+            Torch::getDtypeIntValueForType(rewriter, loc,
+                                           indicesTy.getOptionalDtype()),
+            none, boolFalse);
+        Value updatedIndices = rewriter.create<Torch::AtenMulTensorOp>(
```
rsuderman, line 616: When taking indice values, multiplying, and summing we are better to slice each out explicitly, mul, and ad. We want to avoid using reduction style cases as they form more complex dispatches that often do not work with fusion.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
Shukla-Gaurav: yeah, to support batch_dims we can not map to onnx.gather op. Updated the lowering algorithm. Thanks!

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
Shukla-Gaurav: The best thing I got is to flatten only index_shape[-1] dims of data. Since it involves indexing at multiple axis, I don't see a way without flattening data at least that much. Thanks!
Lmk if there is a better way.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -558,6 +558,95 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          return failure();
+        // TODO: Add support for non zero batch_dims.
+        if (batch_dims != 0)
+          return rewriter.notifyMatchFailure(
+              binder.op, "non zero batch_dims is not supported yet");
+        auto ctx = binder.op->getContext();
+        Location loc = binder.getLoc();
+        auto indicesTy = cast<Torch::ValueTensorType>(indices.getType());
+        auto dataTy = cast<Torch::ValueTensorType>(data.getType());
+        // TODO: Add support for dynamic sizes.
```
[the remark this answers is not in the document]
Shukla-Gaurav: Now it should work for dynamic shapes in data. The only static dim it forces is indices_shape[-1], as it's necessary to have deterministic output rank.

Shukla-Gaurav:
@rsuderman Thanks for reviewing, could you please review it again? Meanwhile I am adding support for 0-1D indices with more test cases. Thanks!

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -558,6 +558,220 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          auto boolTy = rewriter.getType<Torch::ValueTensorType>(
+              indicesSliceShape, rewriter.getI1Type());
+          Value lt = rewriter.create<Torch::AtenLtScalarOp>(
+              loc, boolTy, indicesSlice, constZero);
+          Value add = rewriter.create<Torch::AtenAddScalarOp>(
+              loc, indicesSliceTy, indicesSlice, dataDims[batchDimCount + i],
+              /*alpha=*/constOne);
+          indicesSlice = rewriter.create<Torch::AtenWhereSelfOp>(
+              loc, indicesSliceTy, lt, add, indicesSlice);
+          if (i == 0)
```
rsuderman, line 667: Avoid if/else. In this case

updatedIndices = indicesSlice;
if (i != 0)
  updateIndices = rewriter.create...

rsuderman (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
Shukla-Gaurav: updatedIndices should be initialized with indicesSlice only at i=0. 
I can do something like this:
if(i==0){
  updatedIndices = indicesSlice;
  continue;}
updatedIndices = ....

PR #2989, opened 2024-03-07
[onnx] Fix constant pad for dynamic shape

rsuderman:
The current padding operation was not functional for dynamic shapes.
Updated and enabled tests so that onnx.pad tests pass.

Work TBD for reflection padding.

### DIFF
--- a/lib/Conversion/TorchToLinalg/TensorConstructors.cpp
+++ b/lib/Conversion/TorchToLinalg/TensorConstructors.cpp
@@ -45,36 +45,69 @@ public:
     auto type = self.getType().cast<RankedTensorType>();
     int64_t rank = type.getRank();
 
-    // Pattern match against the op's original operands, because otherwise we
-    // will get the lowered version of the operands which is harder to pattern
-    // match.
-    SmallVector<int64_t> padInts;
-    if (!matchPattern(op.getPad(), m_TorchListOfConstantInts(padInts)))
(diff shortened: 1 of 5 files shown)


lib/Conversion/TorchToLinalg/TensorConstructors.cpp
```diff
@@ -45,36 +45,69 @@ class ConvertAtenConstantPadNdOp
...
 
     // Initialize low/high paddings with the dims that should not be padded.
-    SmallVector<int64_t, 4> lowPadding(/*Size=*/rank - padRank, /*Value=*/0);
-    SmallVector<int64_t, 4> highPadding(/*Size=*/rank - padRank, /*Value=*/0);
+    int64_t noPad = rank - padRank;
+    Attribute zero = rewriter.getIndexAttr(0);
+    SmallVector<int64_t, 4> staticLow(noPad, 0);
+    SmallVector<int64_t, 4> staticHigh(noPad, 0);
+    SmallVector<OpFoldResult, 4> lowPad(noPad, zero);
+    SmallVector<OpFoldResult, 4> highPad(noPad, zero);
```
ScottTodd, line 67: nit: what's the convention for including `, 4>` in torch-mlir? I recall some suggestions to drop the number and let the SmallVector template default figure that out on its own

ScottTodd (approved):

PR #2991, opened 2024-03-07
[onnx] Fix OnnxToLinalg lowering issue for `onnx.Squeeze` and `onnx.Unsqueeze`

vivekkhandelwal1:
This commit also cleans up the OnnxToTorch lowering for the Squeeze and Unsqueeze op and adds the support for handling edge cases.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
@@ -12,6 +12,7 @@
 
 #include "mlir/Bytecode/BytecodeOpInterface.h"
 #include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/DialectResourceBlobManager.h"
 #include "mlir/IR/Matchers.h"
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/IR/OpImplementation.h"
@@ -142,7 +143,7 @@ m_TorchConstantBool(bool *bind_value) {
 }
 
 namespace detail {
-/// Matches the constant integers stored in a `torch.ListConstruct`.
+/// Matches the constant integers stored in a `torch.prim.ListConstruct`.
 struct torch_list_of_constant_ints_op_binder {
   SmallVectorImpl<int64_t> &bind_values;
 
@@ -265,6 +266,41 @@ struct torch_tensor_size_int_op_binder {
 inline detail::torch_tensor_size_int_op_binder
 m_TorchTensorSizeInt(Value tensor, int64_t *dim) {
   return detail::torch_tensor_size_int_op_binder(tensor, dim);
+}
+
+namespace detail {
+/// Matches the constant integers stored in a `onnx.Constant`.
+struct onnx_list_of_constant_ints_op_binder {
+  SmallVectorImpl<int64_t> &bind_values;
+
+  /// Creates a matcher instance that binds the value to bvs if match succeeds.
+  onnx_list_of_constant_ints_op_binder(SmallVectorImpl<int64_t> &bvs)
+      : bind_values(bvs) {}
+
+  bool match(Operation *op) {
+    auto constOp = dyn_cast<Torch::OperatorOp>(op);
+    if (!constOp || !constOp.getName().equals("onnx.Constant"))
+      return false;
+
+    DenseResourceElementsAttr attr =
+        constOp->getAttr("torch.onnx.value")
+            .dyn_cast_or_null<DenseResourceElementsAttr>();
+    auto ty = cast<ShapedType>(attr.getType());
+    ElementsAttr denseAttr;
+    auto ptr = attr.getRawHandle().getBlob()->getData();
+    denseAttr = DenseElementsAttr::getFromRawBuffer(ty, ptr);
+    for (auto axis : denseAttr.getValues<llvm::APInt>()) {
+      bind_values.push_back(axis.getSExtValue());
+    }
+    return true;
+  }
+};
+} // namespace detail
+
+/// Matches the constant integers stored in a `onnx.Constant`.
+inline detail::onnx_list_of_constant_ints_op_binder
+m_OnnxListOfConstantInts(SmallVectorImpl<int64_t> &bind_values) {
+  return detail::onnx_list_of_constant_ints_op_binder(bind_values);
 }
 
 /// Create code to copy `tensor` to type `newType`.
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.h
@@ -12,7 +12,6 @@
 
 #include "mlir/Bytecode/BytecodeOpInterface.h"
 #include "mlir/IR/BuiltinTypes.h"
-#include "mlir/IR/DialectResourceBlobManager.h"
 #include "mlir/IR/Matchers.h"
 #include "mlir/IR/OpDefinition.h"
 #include "mlir/IR/OpImplementation.h"
@@ -266,41 +265,6 @@ struct torch_tensor_size_int_op_binder {
 inline detail::torch_tensor_size_int_op_binder
 m_TorchTensorSizeInt(Value tensor, int64_t *dim) {
   return detail::torch_tensor_size_int_op_binder(tensor, dim);
-}
-
-namespace detail {
-/// Matches the constant integers stored in a `onnx.Constant`.
-struct onnx_list_of_constant_ints_op_binder {
-  SmallVectorImpl<int64_t> &bind_values;
-
-  /// Creates a matcher instance that binds the value to bvs if match succeeds.
-  onnx_list_of_constant_ints_op_binder(SmallVectorImpl<int64_t> &bvs)
-      : bind_values(bvs) {}
-
-  bool match(Operation *op) {
-    auto constOp = dyn_cast<Torch::OperatorOp>(op);
-    if (!constOp || !constOp.getName().equals("onnx.Constant"))
-      return false;
-
-    DenseResourceElementsAttr attr =
-        constOp->getAttr("torch.onnx.value")
-            .dyn_cast_or_null<DenseResourceElementsAttr>();
-    auto ty = cast<ShapedType>(attr.getType());
-    ElementsAttr denseAttr;
-    auto ptr = attr.getRawHandle().getBlob()->getData();
-    denseAttr = DenseElementsAttr::getFromRawBuffer(ty, ptr);
-    for (auto axis : denseAttr.getValues<llvm::APInt>()) {
-      bind_values.push_back(axis.getSExtValue());
-    }
-    return true;
-  }
-};
-} // namespace detail
-
-/// Matches the constant integers stored in a `onnx.Constant`.
-inline detail::onnx_list_of_constant_ints_op_binder
-m_OnnxListOfConstantInts(SmallVectorImpl<int64_t> &bind_values) {
-  return detail::onnx_list_of_constant_ints_op_binder(bind_values);
 }
 
 /// Create code to copy `tensor` to type `newType`.
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -17,20 +17,6 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::onnx_c;
 
-class Endian {
-private:
-  static constexpr uint32_t uint32_ = 0x01020304;
-  static constexpr uint8_t magic_ = (const uint8_t &)uint32_;
-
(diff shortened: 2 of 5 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -540,57 +540,86 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
   patterns.onOp(
       "Squeeze", 13, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
```
renxida, line 541: update version number to 1. the old impl had overly high since-version numbers.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -604,103 +633,102 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
-            result = rewriter.create<Torch::AtenUnsqueezeOp>(
-                binder.getLoc(), resultType, result, dim);
-          } else {
-            result = rewriter.create<Torch::AtenUnsqueezeOp>(
-                binder.getLoc(), baseType, result, dim);
-          }
+              binder.getLoc(), isDimGreaterThanRank, one);
+          Value unsqueezeDim =
+              rewriter.create<Torch::AtenSubIntOp>(loc, dim, finalOffset);
+          auto baseType = Torch::ValueTensorType::getWithLeastStaticInformation(
```
renxida, line 726: This might be the cause of the problem. During my earlier attempt at fixing this, I had to make this static to get it through to linalg.

https://github.com/llvm/torch-mlir/pull/2904

renxida:
https://github.com/llvm/torch-mlir/pull/2904 here's my earlier fix for the unsqueeze issue. That one should fully lower to linalg with no problem.

feel free to cannibalize it to make this work. from what it looks like you've simplified unsqueeze a lot but the intermediate unsqueeze results requrie static typing to lower to linalg when i tried it.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: Hi @renxida, the implementation https://github.com/llvm/torch-mlir/pull/2904/files, is based on the assumption that axes is defined using the `ConstantOp`. I'm not sure if we want to restrict the lowering to that assumption. @rsuderman, would you like to add something here?

renxida: last time i checked with @rsuderman it was a reasonable assumption to make because the models we are targeting would usually have the axes defined using ConstantOp

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: No, it can't be changed to 1 since the version number 1 expects inputs to be in different form. Please refer: https://onnx.ai/onnx/operators/text_diff_Squeeze_1_13.html

renxida:
@vivekkhandelwal1 oh consider adding support for the lower versions of unsqueeze. Here's an IREE issue where somebody ran into a wall with an opset7 onnx file

https://github.com/openxla/iree/issues/16339

vivekkhandelwal1:
> @vivekkhandelwal1 oh consider adding support for the lower versions of unsqueeze. Here's an IREE issue where somebody ran into a wall with an opset7 onnx file
> 
> [openxla/iree#16339](https://github.com/openxla/iree/issues/16339)

Someone can add that support in a follow-up PR.

vivekkhandelwal1:
@rsuderman, can you please approve this PR? Seems like since you added some comments, you need to approve it to be merged.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -604,103 +648,83 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+              inputShapeCounter++;
+            } else {
+              unsqueezeDims.push_back(i);
+            }
+          }
+        } else {
+          auto constOp = axes.getDefiningOp<Torch::OperatorOp>();
+          if (!constOp || !constOp.getName().equals("onnx.Constant"))
+            return failure();
+          DenseResourceElementsAttr attr =
```
rsuderman, line 695: You should not replicate grabbing the constant value by assuming the input is an onnx constant operation. You should be able to use the `m_Constant` to fold. This runs via the dialect conversion framework so I am reasonably certain it should be able to pull the constant values that way.

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: I tried using that but it doesn't work for the `DenseResourceElementsAttr`.

vivekkhandelwal1: Hi @rsuderman, let me know if we can proceed with this patch for now since it has been pending since long.

rsuderman: If thats the case then we should make a constant matcher for the `onnx.Constant` operation instead of embedding the logic within the rewriter. See examples of how constant folding matchers are written for torch constants:

https://github.com/llvm/torch-mlir/blob/ce7d4f1660c210b98978b73a0ed4a3db440e4646/include/torch-mlir/Dialect/Torch/IR/TorchOps.h#L140

We should also guarantee the resource variable matching occurs using the same code for `onnx.Constant`. There is a reasonable chance we will rework constant loading at some point and need to handle alternative methods for loading constants.

zjgarvey: As far as I collected from the onnx documentation, you shouldn't need to match on constant axes, because even with dynamic dims on the input, the onnx shape inference for this op checks if the input axes can be realized (e.g. if they are coming from a constant), and adds the 1's to where they need to be. See, for example, the generated onnx graph for ElementwisePreluModule_basic and [onnx.Unsqueeze](https://github.com/onnx/onnx/blob/366ac64b5af0af2d02b019d84664e6b46b3ceb09/onnx/defs/tensor/defs.cc#L1849). That is, the check on line 676 is not entirely necessary for the logic of that block to work,

zjgarvey (approved):
I still think just a bit more can be squeezed (*chuckle*) out of this (for example, unsqueeze : [?,?,1] -> [?,?]). Is the concern that one of the ? might secretly be a 1? In any case, I think that should just be included in a different patch at this point, so LGTM.

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
```diff
@@ -267,6 +268,41 @@ m_TorchTensorSizeInt(Value tensor, int64_t *dim) {
   return detail::torch_tensor_size_int_op_binder(tensor, dim);
 }
 
+namespace detail {
+/// Matches the constant integers stored in a `onnx.Constant`.
+struct onnx_list_of_constant_ints_op_binder {
```
rsuderman, line 273: This should not be in `TorchOps.h`. You can move it into the `Utils.h` file for `onnx`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -17,6 +17,20 @@ using namespace mlir;
 using namespace mlir::torch;
 using namespace mlir::torch::onnx_c;
 
+class Endian {
```
rsuderman, line 20: I don't see this used anywhere in here.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -661,57 +675,86 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                binder.getLoc(), rewriter.getI64IntegerAttr(i)));
+          }
+        }
+
+        if (dimList.empty()) {
+          Value axes = inputOperands[1];
+          Torch::BaseTensorType axesType =
+              axes.getType().cast<Torch::BaseTensorType>();
+          Type selectResultType = axesType.getWithSizesAndDtype(
+              llvm::ArrayRef<int64_t>{1}, axesType.getOptionalDtype());
```
rsuderman, line 738: ArrayRefs are dangerous to construct with list initializers. The memory allocated for the list may go out of scope.

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
```diff
@@ -267,6 +268,41 @@ m_TorchTensorSizeInt(Value tensor, int64_t *dim) {
...
+
+  /// Creates a matcher instance that binds the value to bvs if match succeeds.
+  onnx_list_of_constant_ints_op_binder(SmallVectorImpl<int64_t> &bvs)
+      : bind_values(bvs) {}
+
+  bool match(Operation *op) {
+    auto constOp = dyn_cast<Torch::OperatorOp>(op);
+    if (!constOp || !constOp.getName().equals("onnx.Constant"))
+      return false;
+
```
rsuderman, line 284: You should include the endianness check here so we don't assume things are fine on a wrong endian system

rsuderman, line 284: You should also do a check on the Attr type with `return false`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -661,57 +675,86 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
-        selectSizes.push_back(1);
-        Type selectResultType = axesType.getWithSizesAndDtype(
-            llvm::ArrayRef(selectSizes), axesType.getOptionalDtype());
-        auto sizes =
-            dyn_cast<Torch::ValueTensorType>(axes.getType()).getSizes();
-        if (sizes.size() == 0) {
+
+        Value data = inputOperands[0];
+        auto inputType = data.getType().dyn_cast<Torch::ValueTensorType>();
+        if (!inputType.hasSizes() || !resultType.hasSizes()) {
```
rsuderman, line 685: Uneeded `{}`

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -725,103 +768,70 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
-        SmallVector<Value> dimList;
-        SmallVector<int64_t> selectSizes;
-        selectSizes.push_back(1);
-        Type selectResultType = axesType.getWithSizesAndDtype(
-            llvm::ArrayRef(selectSizes), axesType.getOptionalDtype());
-        auto sizes =
-            dyn_cast<Torch::ValueTensorType>(axes.getType()).getSizes();
-        if (sizes.size() == 0) {
+        auto inputType = data.getType().dyn_cast<Torch::ValueTensorType>();
+        if (!inputType.hasSizes() || !resultType.hasSizes()) {
```
rsuderman, line 778: More unneeded curly brackets.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -725,103 +768,70 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            }
           }
+        } else {
+          SmallVector<int64_t> unsqueezeDimsInts;
+          if (!matchPattern(axes,
+                            Torch::m_OnnxListOfConstantInts(unsqueezeDimsInts)))
+            return rewriter.notifyMatchFailure(
+                binder.op, "only support constant int axes values");
+
+          for (auto dim : unsqueezeDimsInts) {
```
rsuderman, line 818: Ditto curly braces.

rsuderman (changes requested):

include/torch-mlir/Dialect/Torch/IR/TorchOps.h
```diff
@@ -12,6 +12,7 @@
 
 #include "mlir/Bytecode/BytecodeOpInterface.h"
 #include "mlir/IR/BuiltinTypes.h"
+#include "mlir/IR/DialectResourceBlobManager.h"
```
rsuderman, line 15: No longer needed

rsuderman (approved):

PR #2993, opened 2024-03-07
Added a fix so that Gemm accepts 2 as well as 3 inputs.

afalkenberg1:
Tests show that Gemm only was working with 3 inputs, nevertheless the third input is optional.
A simple check was implemented to check if the 3rd input is available or not. If it is available 
everything stays as is. If it is not available the Value c is not used and code ends appropriately.


rsuderman (changes requested):
Try to rework title and commit message for something more succinct and clear. Phrases like `added a fix` is a little redundant. For this I would do something like the following:

```
[onnx] Add support for `onnx.Gemm` with no bias

Previous version required a bias vector. This provides an alternate path to `Torch::AtenMm` with no bias operation.
```

We should also still have a lit test in onnx-to-torch tests for completeness.

rsuderman (approved):

PR #2994, opened 2024-03-07
Add OnnxToTorch support for EyeLike op

jinchen62:


jinchen62:
@rsuderman Hi, I figured that the CI failure was because of the dynamic shape. I found something like
` %34 = torch.operator "onnx.EyeLike"(%33) {torch.onnx.k = -1 : si64} : (!torch.vtensor<[?,?],f32>) -> !torch.vtensor<[?,?],f32>`
I'm not sure how to handle it so just return failure if it's dynamic. Do you have any suggestions?

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1745,6 +1745,129 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          dtypeIntTorch = onnxDtypeIntToTorchDtypeInt(dtypeIntOnnx);
+          dtypeVal = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(dtypeIntTorch));
+        }
+
+        if (diagonalIndex == 0) {
+          // diagonalIndex = 0 populates the main diagonal
+          Value nVal = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(shapeSizes[0]));
+          if (shapeSizes[0] == shapeSizes[1]) {
```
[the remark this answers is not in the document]
zjgarvey: To clarify, if you are trying to save writing the extra value m into the IR right now, it will get written later on anyway so all that happens is the code is slightly harder to read here.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1745,6 +1745,129 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          // concatenate zeros op and eye op
+          SmallVector<Value> tensors = {zerosOp, eyeOp};
+          Type listElemType =
+              zerosOp.getType()
+                  .cast<Torch::BaseTensorType>()
+                  .getWithSizesAndDtype(/*optionalSizes=*/std::nullopt,
+                                        /*optionalDtype=*/nullptr);
+          Value tensorsList = rewriter.create<Torch::PrimListConstructOp>(
+              binder.getLoc(), Torch::ListType::get(listElemType), tensors);
+          rewriter.replaceOpWithNewOp<Torch::AtenCatOp>(binder.op, resultType,
```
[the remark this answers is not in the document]
zjgarvey: I'm not sure what the difference is performance-wise, so I would ask someone who knows more about that sort of thing.

zjgarvey (commented):
The logic here makes sense, and I can see why it ended up a bit more complicated than one would think. What kind of e2e testing have you tried? Do the lit tests also convert successfully to linalg?

Consider looking into converting to AtenDiagEmbedOp for the dynamic dim case. It appears to allow for a dynamic input (as long as the ints are left unspecified?). Having support for dynamic dims is probably important for this op.

The fact that no e2e tests are Xpassing is a bit odd. It seems like torch eye ops get converted to onnx.constants. It also appears that torch.ops.aten.diagonal gets exported to onnx with an onnx.EyeLike op (it likely multiplies the original matrix elementwise by an EyeLike op to extract the diagonal). Most of those tests use dynamic dims, so that's likely the reason they are still not passing with your patch.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1749,6 +1749,107 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+            binder.getLoc(), rewriter.getI64IntegerAttr(0));
+        Value cst1 = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getI64IntegerAttr(1));
+        Value nVal = rewriter.create<Torch::AtenSizeIntOp>(binder.getLoc(),
+                                                           operand, cst0);
+        Value mVal = rewriter.create<Torch::AtenSizeIntOp>(binder.getLoc(),
+                                                           operand, cst1);
+        Value noneVal = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+        if (dtypeIntOnnx == -1) {
+          dtypeVal = noneVal;
```
rsuderman, line 1782: Declare `dtypeVal` here and remove from the `if`. We only need the `else` part as a conditional as having `if-else` only avoids a single assign.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1749,6 +1749,107 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+              binder.getLoc(), rewriter.getI64IntegerAttr(dtypeIntTorch));
+        }
+
+        if (diagonalIndex == 0) {
+          // diagonalIndex = 0 populates the main diagonal
+          rewriter.replaceOpWithNewOp<Torch::AtenEyeMOp>(
+              binder.op, resultType, nVal, mVal, dtypeVal, noneVal, noneVal,
+              noneVal);
+          return success();
+        } else {
```
rsuderman, line 1795: remove the `else` as we return in the `if` anyway.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1749,6 +1749,107 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+            if (eyeShape[0] != Torch::kUnknownSize) {
+              eyeShape[0] += diagonalIndex;
+            }
+            dimVal = cst0;
+            startVal = nVal;
+          }
+
+          // create main diag eye op
+          auto eyeResultType =
+              Torch::ValueTensorType::get(binder.op->getContext(), eyeShape,
```
rsuderman, line 1828: You can do `rewriter.getType<Torch::ValueTensorTyoe>(...)` so that you do not need to request the context.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1749,6 +1749,107 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+                                              .getOptionalDtype());
+          Value eyeOp = rewriter.create<Torch::AtenEyeMOp>(
+              binder.getLoc(), eyeResultType, newN, newM, dtypeVal, noneVal,
+              noneVal, noneVal);
+          // create zeros op
+          SmallVector<Value> zerosShapeValues = {nVal, mVal};
+          Value zerosShapeList = rewriter.create<Torch::PrimListConstructOp>(
+              binder.getLoc(),
+              Torch::ListType::get(
+                  Torch::IntType::get(binder.op->getContext())),
```
rsuderman, line 1840: `rewriter.getType<Torch::IntType>()`

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1749,6 +1749,98 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
             binder.op, resultType, data, dimValueList);
         return success();
       });
+  patterns.onOp(
+      "EyeLike", 9, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value operand;
+        int64_t dtypeIntOnnx, diagonalIndex;
+        if (binder.tensorOperand(operand) ||
+            binder.s64IntegerAttr(dtypeIntOnnx, "dtype", -1) ||
```
vivekkhandelwal1, line 1758: Instead of `-1` use the dtype value of float.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1749,6 +1749,98 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+            binder.s64IntegerAttr(diagonalIndex, "k", 0) ||
+            binder.tensorResultType(resultType))
+          return failure();
+
+        TensorType tensorType =
+            dyn_cast<Torch::ValueTensorType>(operand.getType())
+                .toBuiltinTensor();
```
vivekkhandelwal1, line 1765: I don't think you need to convert to `BuiltinTensor`. You can just use `getSizes()`.

vivekkhandelwal1 (changes requested):

vivekkhandelwal1:
@jinchen62, I don't see any failing test getting passed because of this patch, also no new tests are added. How are we ensuring the e2e correctness of this?

jinchen62:
@vivekkhandelwal1 The `DiagonalModule` tests in torch-mlir are blocked by the support of onnx.if. The e2e iree tests from shark-testsuite: `test_eyelike_populate_off_main_diagonal` passed, the other two compiled but have different run issues. What new tests should I add?

zjgarvey (approved):
I like the change to AtenSliceScatterOp over AtenCatOp. This looks much cleaner overall. Since the e2e testing for this pattern is hard in torch-mlir, you can consider adding a test to SHARK-Test-Suite's e2eshark/onnx/operators/EyeLike. I'd be especially curious to see what happens on compilation with a dynamic dim case.

vivekkhandelwal1 (approved):
LGTM

rsuderman (approved):

PR #2996, opened 2024-03-08
Align softmax accumulation types with Torch's CUDA implementation

nithinsubbiah:


include/torch-mlir/Dialect/Torch/Utils/Utils.h
```diff
@@ -131,6 +131,15 @@ Value createRank0Tensor(PatternRewriter &rewriter, Location loc,
...
                                 int64_t dimB, Type &transposedType);
 
+// Approximates the heuristic in the torch `acc_type` template for kernels
+// that are defined in terms of it. For now, this just returns accumulators
+// as if for CUDA from that implementation. In the future, this could be
+// extended to look at hints on the `forOp` or its container to better
+// control the behavior. Such support would be done in coordination with
+// the fx_importer and APIs, which could add hints to the IR (based on
+// Torch flags, user options, etc).
+Type getDefaultAccType(PatternRewriter &rewriter, Type inputType);
```
antiagainst, line 141: Actually this should be specific to a certain op like softmax? We don't want to just convert everything unconditionally? That's what I recall what @stellaraccident recommends--`Type getDefaultAccType(Operation *forOp, Type inputType);`. But i'm certainly not experienced enough here to say exactly.

stellaraccident: I think it is fine like this : lowerings should use this if the corresponding torch op uses the torch version of this to get a default accumulator type. Still under control of the specific lowering but let's it line up with how torch kernels are structured.

stellaraccident: (in torch, the soft max kernels defer to this default accumulator heuristic, so we should too)

qedawkins: yeah not everything unconditionally, but the same principle applies to convolution and a few other reduction operators (i.e. variance).

stellaraccident (approved):

lib/Dialect/Torch/Utils/Utils.cpp
```diff
@@ -525,3 +525,33 @@ LogicalResult Torch::getTransposedType(BaseTensorType inType, int64_t dimA,
...
+  if (inputType.isFloat8E5M2())
+    return rewriter.getF32Type();
+  if (inputType.isFloat8E4M3FN())
+    return rewriter.getF32Type();
+  if (inputType.isFloat8E5M2FNUZ())
+    return rewriter.getF32Type();
+  if (inputType.isFloat8E4M3FNUZ())
+    return rewriter.getF32Type();
+  if (inputType.isSignedInteger(8))
+    return rewriter.getI64Type();
```
rsuderman, line 547: This is not correct, `int8` is almost guaranteed to be an `int32_t` accumulator

lib/Dialect/Torch/Utils/Utils.cpp
```diff
@@ -525,3 +525,33 @@ LogicalResult Torch::getTransposedType(BaseTensorType inType, int64_t dimA,
...
+  if (inputType.isUnsignedInteger(8))
+    return rewriter.getI64Type();
```
rsuderman, line 549: Ditto.

rsuderman (changes requested):

lib/Dialect/Torch/Utils/Utils.cpp
nithinsubbiah: I'm following the convention from CUDA - https://github.com/pytorch/pytorch/blob/862b99b571e7544c82232aa6956a8ddd0ab247ed/aten/src/ATen/AccumulateType.h#L118

rsuderman: This is strange but we should match. We will probably want a linalg lowering that truncates the accumulator type when it is guaranteed to fall within bounds.

rsuderman (approved):

PR #2999, opened 2024-03-08
fximporter: support newer torch versions

dan-garvey:
uses version checking since attributes exist in both versions, the only thing that changes is what we're receiving as an fx graph

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -225,9 +225,11 @@ PY_BUILTIN_TO_TORCH_OP = {
 _IS_TORCH_2_1_OR_EARLIER = torch.__version__.split("+")[0] <= "2.1.0"
 
 # The following are maps from symbolic ops to their non symbolic equivalents.
-# In <=2.1.0, imported fx graphs come with torch.ops.aten.sym_size and a number
-# identifying either default or int for the types we include here. Later
-# versions drop this in favor of torch.ops.aten.sym_size.<type>.
+# In <=2.1.0, imported fx graphs come with a type inspecific torch.ops.aten.sym_size
+# We identify it using the number of args in the node, 1 being default, 2 being int
+# In the mapping below (torch.aten.sym_size, 2) indicates len(args)=2 therefore
+# map to torch.aten.size.int.
+# Thankfully, newer versions provide a specific torch.ops.aten.sym_size.<type>.
 # Once we drop support for <2.1.0, we can get rid of the the SYMBOLIC_TORCH_OPS
 # set and just check key existence in SYMBOLIC_OP_TO_TORCH_OP
 
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -220,19 +220,35 @@ PY_BUILTIN_TO_TORCH_OP = {
     "gt": torch.ops.aten.gt,
 }
 
-SYMBOLIC_TORCH_OPS = {
-    torch.ops.aten.sym_size,
-    torch.ops.aten.sym_stride,
-    torch.ops.aten.sym_numel,
-}
+# this allows stripping of cuda versions like: "+cu113"
+if torch.__version__.split("+")[0] <= "2.1.0":
+	SYMBOLIC_TORCH_OPS = {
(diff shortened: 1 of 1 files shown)


python/torch_mlir/extras/fx_importer.py
```diff
@@ -220,19 +220,35 @@
...
-}
-
-SYMBOLIC_OP_TO_TORCH_OP = {
-    (torch.ops.aten.sym_size, 1): torch.ops.aten.size.default,
-    (torch.ops.aten.sym_size, 2): torch.ops.aten.size.int,
-    (torch.ops.aten.sym_stride, 1): torch.ops.aten.stride.default,
-    (torch.ops.aten.sym_stride, 2): torch.ops.aten.stride.int,
-    (torch.ops.aten.sym_numel, 1): torch.ops.aten.numel.default,
-}
+# this allows stripping of cuda versions like: "+cu113"
```
stellaraccident, line 223: I don't understand what this comment means?

dan-garvey:
@stellaraccident I know you had some thoughts on how the tests for fx_importer should be unittestier for torch_mlir. I'm happy to pick that up if you want to give me a pointer to what you'd like changed.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -220,19 +220,35 @@
...
+if torch.__version__.split("+")[0] <= "2.1.0":
```
stellaraccident, line 224: Why can't we just do a direct comparison (what is the split doing?)

python/torch_mlir/extras/fx_importer.py
```diff
@@ -220,19 +220,35 @@
...
+    SYMBOLIC_TORCH_OPS = {
+        torch.ops.aten.sym_size,
+        torch.ops.aten.sym_stride,
+        torch.ops.aten.sym_numel,
+    }
+
+    SYMBOLIC_OP_TO_TORCH_OP = {
```
stellaraccident, line 231: I've thought for quite a while that this whole fragment is hinting at a design flaw, and now we're adding to it. Can you please precede this section with a comment about what these maps are and what their fields mean?

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1212,7 +1230,10 @@ def _import_symbolic_torch_op(
             ), f"Unsupported builtin function for symbolic types: {target} with args {node.args}"
             concrete_target = getattr(torch_op, op_overload)
         else:
-            concrete_target = SYMBOLIC_OP_TO_TORCH_OP.get((target, len(node.args)))
+            if torch.__version__.split("+")[0] <= "2.1.0":
```
stellaraccident, line 1233: I think we should have these version checks at the module level bound to constants like `_IS_TORCH_2_1_OR_EARLIER`. saves some inline work and also makes the code more readable and clear for when we can drop branches.

stellaraccident (changes requested):
Thanks, a few changes I'd like to see as this part was already limping and this adds to it.

python/torch_mlir/extras/fx_importer.py
dan-garvey: when torch installs with cuda,  torch.__version__ = "2.1.0+cu113". In order to support both "2.1.0" and "2.1.0+cu*" split at the plus and then compare. 

"2.1.0+cu113" <= "2.1.0" resolves to False. Maybe there is a fancier way to do it but 🤷

python/torch_mlir/extras/fx_importer.py
```diff
@@ -220,19 +220,45 @@
...
+# set and just check key existence in SYMBOLIC_OP_TO_TORCH_OP
+
+if _IS_TORCH_2_1_OR_EARLIER:
+    SYMBOLIC_TORCH_OPS = {
+        torch.ops.aten.sym_size,
+        torch.ops.aten.sym_stride,
+        torch.ops.aten.sym_numel,
+    }
+
+    SYMBOLIC_OP_TO_TORCH_OP = {
```
stellaraccident, line 241: One more nit: Please say what the number in the tuple key means.

stellaraccident (approved):
One nit then lgtm. Thanks for the comments and ergonomics.

PR #3002, opened 2024-03-09
[onnx] Fix onnx.ReduceMean lowering

rsuderman:
Reduce mean lowerings did not succesfully lower to `linalg` via torched. There were two separate paths that could be consolidated to a single simpler pass. This resulted in a significant improvement in test coverage.


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -847,155 +847,95 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
       });
   // onnx.ReduceMean with axes provided as argument introduced in opset 18
   patterns.onOp(
-      "ReduceMean", 18,
+      "ReduceMean", 1,
```
ScottTodd, line 850: Drop comment about opset 18

ScottTodd (approved):

PR #3003, opened 2024-03-09
[Torch Dialect] emit aten::warn

qingyunqu:
* torch-mlir may not handle `aten.warn`. But it could be handled by custom users' backend which involves torch-mlir.


Vremold (approved):
LGTM. Thanks for the effort.

Issue #3004, opened 2024-03-10
Dynamic shape support needed for flatten using ints in tosa dialect


Abhishek-TyRnT:
`flattenusingints` op currently only supports static shapes. 
https://github.com/llvm/torch-mlir/blob/a3fe130f73f160c4d5b984f8de5b26aa471cfc9a/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L2488-L2491

Due to this tests like argmax fail. Sending a PR for fix

PR #3005, opened 2024-03-10
Dynamic size support for flatten

Abhishek-TyRnT:
Added support for dynamic shapes in `flattenusingints` op in tosa dialect. Due to this some Argmax tests pass 
This PR fixes this issue https://github.com/llvm/torch-mlir/issues/3004

The following tests pass after this PR
 ```
1. "ArgmaxIntModule_basic"
2. "ArgmaxIntModule_multiple_maxs"
3. "ArgmaxModule_basic"
```

### CODE REVIEW

File: lib/Conversion/TorchToTosa/TorchToTosa.cpp [hunk 1]
```diff
@@ -2511,16 +2510,27 @@ LogicalResult ConvertAtenOp<AtenFlattenUsingIntsOp>::matchAndRewrite(
...
+  SmallVector<int64_t> oldShape;
+
+  if (selfType.hasStaticShape()) {
+    for (auto s : selfType.getShape())
+      oldShape.push_back(s);
+  } else {
+    // Pushing unknown sizes
+    for (unsigned i = 0; i < selfRank; i++)
+      oldShape.push_back(ShapedType::kDynamic);
+  }
```
ramiro050, line 2522: I'm a bit confused by this. `selfType` will already contain `ShapedType::kDynamic` for the dimensions that are dynamic. Why do we need to wipe out all the static information?

projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -897,6 +897,9 @@
     "ArangeStartNegativeStepFloatModule_basic",
     "ArangeStartOutDtypeModule_basic",
     "ArangeStartStepFloatModule_basic",
+    "ArgmaxIntModule_basic",
+    "ArgmaxIntModule_multiple_maxs",
+    "ArgmaxModule_basic",
```
ramiro050, line 902: Given that the changes are to `AtenFlattenUsingIntsOp`, I would expect at least one test for `FlattenUsingInts` to also start passing (if all already pass for TOSA, then a new one should be added to test new features)

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -2511,16 +2510,27 @@ LogicalResult ConvertAtenOp<AtenFlattenUsingIntsOp>::matchAndRewrite(
...
+  for (auto s : llvm::enumerate(makeShapeTorchCompatible(oldShape))) {
     int64_t idx = s.index();
     if (idx < start_dim || idx > end_dim) {
       newShape.push_back(s.value());
     } else {
       if (idx == start_dim)
         newShape.push_back(s.value());
-      else
+      // Only updating when the shapes are static
+      else if (s.value() >= 0)
```
ramiro050, line 2533: This does not seem right to me. If one of the dimensions in the range `[start_dim, end_dim]` is dynamic, this will result in the wrong size for the flattened portion.

ramiro050 (changes requested):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
Abhishek-TyRnT: I guess I haven't considered the case I will quickly make the fix.

ramiro050 (changes requested):

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -2520,7 +2519,8 @@ LogicalResult ConvertAtenOp<AtenFlattenUsingIntsOp>::matchAndRewrite(
     } else {
       if (idx == start_dim)
         newShape.push_back(s.value());
-      else
+      // Only updating when the shapes are static
+      else if (s.value() > 0 && newShape.back() > 0)
```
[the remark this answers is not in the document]
Abhishek-TyRnT: I added an else condition which would make the collapsed shape dynamic even if one dim is dynamic

https://github.com/llvm/torch-mlir/blob/8527ab2cf746ed4ee1d6e105531fd3803063ac95/lib/Conversion/TorchToTosa/TorchToTosa.cpp#L2523-L2526

I hope this is upto expectation

ramiro050 (approved):
Just one more small change, but other than that it LGTM. Thanks!

Abhishek-TyRnT:
Completed please check

PR #3007, opened 2024-03-11
[Torch Dialect] emit aten::mul and add folder

qingyunqu:


Vremold (approved):
LGTM. Thanks for the effort.

PR #3013, opened 2024-03-11
OnnxToTorch lowering resize op

aldesilv:


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2589,6 +2589,204 @@ class ConvertAtenGridSamplerOp : public OpConversionPattern<AtenGridSamplerOp> {
...
+    SmallVector<utils::IteratorType> iteratorTypes(
+        inputRank, utils::IteratorType::parallel);
+
+    Value finalRes =
+        rewriter
+            .create<linalg::GenericOp>(
+                loc, outTensor.getType(), ValueRange{}, outTensor,
+                /*indexingMaps=*/idMap,
+                /*iteratorTypes=*/iteratorTypes,
+                [&](OpBuilder &b, Location loc, ValueRange args) {
```
rsuderman, line 2639: Given the length it is better to not use the body builder and do something similar to the following:

https://github.com/openxla/iree/blob/main/compiler/plugins/input/StableHLO/Conversion/StableHLOToLinalg.cpp#L2509

This avoids the multiple identations.

rsuderman (changes requested):
Only one implementation change, the rest looks good. However are we able to use `torchvision` to generate an end-to-end test? If not can you link to a external e2e test used to validate the numerics?

ScottTodd:
> However are we able to use `torchvision` to generate an end-to-end test? If not can you link to a external e2e test used to validate the numerics?

We have these test cases running in SHARK-TestSuite / IREE: https://github.com/openxla/iree/blob/ee32fc7b74b4aad49fc46d67ea6c1f617416c4d4/experimental/regression_suite/external_test_suite/config_cpu_llvm_sync.json#L498-L534

Imported .mlir files for those are found in subfolders matching test case names in https://github.com/nod-ai/SHARK-TestSuite/tree/main/iree_tests/onnx/node/generated

The original ONNX files for that are here: https://github.com/onnx/onnx/tree/main/onnx/backend/test/data/node

---

If you want to fully validate numerics on IREE, you can follow the instructions in https://github.com/nod-ai/SHARK-TestSuite/tree/main/iree_tests to see if those XFAIL tests flip to passing. For staying in torch-mlir, you could look at the source .onnx and imported .mlir files and fit those to how tests are already configured here.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2648,4 +3035,7 @@ void mlir::torch::torch_to_linalg::populateUncategorizedPatternsAndLegality(
   patterns.add<ConvertDequantizePerChannel>(typeConverter, context);
   target.addIllegalOp<AtenGridSamplerOp>();
   patterns.add<ConvertAtenGridSamplerOp>(typeConverter, context);
+  target.addIllegalOp<Aten__InterpolateSizeListScaleListOp>();
+  patterns.add<ConvertBilinearInterpolateOp>(typeConverter, context);
+  patterns.add<ConvertInterpolateNearestOp>(typeConverter, context);
```
vivekkhandelwal1, line 3040: @aldesilv, I think a target is missing here.

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/torch_ods_gen.py
```diff
@@ -502,6 +502,12 @@ def emit_with_mutating_variants(key, **kwargs):
     emit(
         "aten::_log_softmax : (Tensor, int, bool) -> (Tensor)"
     )
+    emit(
+        "aten::__interpolate.scale_list : (Tensor, int?, float[]?, str, bool?, bool?, bool) -> (Tensor)"
+    )
+    emit(
+        "aten::__interpolate.size_list_scale_list : (Tensor, int[]?, float[]?, str, bool?, bool?, bool) -> (Tensor)"
+    )
```
vivekkhandelwal1, line 510: Please add the shape info for both of these ops. Refer https://github.com/llvm/torch-mlir/blob/main/docs/adding_abstract_interpretation_functions.md

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2593,6 +2593,393 @@ class ConvertAtenGridSamplerOp : public OpConversionPattern<AtenGridSamplerOp> {
...
+  int hDimOffset = 2;
+  indices[hDimOffset] = yNearest;
+  indices[hDimOffset + 1] = xNearest;
+  Value retVal = b.create<tensor::ExtractOp>(loc, input, indices);
+  return retVal;
+}
+
+namespace {
+class ConvertInterpolateNearestOp
+    : public OpConversionPattern<Aten__InterpolateSizeListScaleListOp> {
```
vivekkhandelwal1, line 2656: How can we lower one single op from two different lowering paths?

vivekkhandelwal1 (changes requested):
@aldesilv, your patch is quite large. It would be better if you could add some description for the approach followed by you for this lowering. If you have referenced it from somewhere then add the references too. Otherwise, it would be really hard to review the patch.

It would be better if you could add some e2e tests as well.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
aldesilv: It does not lower so it sets failure() and will try the other path

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/torch_ods_gen.py
aldesilv: ok

lib/Conversion/TorchToLinalg/Uncategorized.cpp
aldesilv: one is for the nearest mode the other is for bilinear

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/torch_ods_gen.py
vivekkhandelwal1: I think there's some confusion here. I'm asking to add the shape info here.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
vivekkhandelwal1: You don't need two different conversion patterns for one op. You can have one single pattern and lower it to two distinct paths based on the value of Mode.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2206,4 +2206,210 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          modeStrValue = rewriter.create<Torch::ConstantStrOp>(binder.getLoc(),
+                                                               "bilinear");
+          Value alignCorners;
+          if (coordTfMode == "align_corners") {
+            alignCorners = cstTrue;
+          } else {
+            alignCorners = cstFalse;
+          }
+          if (operands.size() < 4) {
+            Value scalesOperand = operands[2];
```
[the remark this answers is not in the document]
aldesilv: onnx "resize" requires either scales or sizes to be present but not both.

vivekkhandelwal1:
@aldesilv, you should add `QuantizedMLP_basic` to the xfail with the torch version check.

vivekkhandelwal1:
> QuantizedMLP_basic

I have updated the PR. Let's wait for CI to pass.

tdietert:
Great work! I'm looking forward to this getting merged :)

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2637,4 +2637,117 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         rewriter.replaceOp(binder.op, {loss, logProb});
         return success();
       });
+  patterns.onOp(
+      "Resize", 11, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        llvm::SmallVector<Value> operands;
+        std::string mode;
+        std::string coordTfMode;
```
vivekkhandelwal1, line 2645: Define them in the same line.

vivekkhandelwal1 (dismissed):

vivekkhandelwal1:
Hi @aldesilv, after addressing the above comments, the PR can be merged.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2637,4 +2637,117 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        Value noneVal = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+
+        if (binder.tensorOperandsList(operands) ||
+            binder.tensorResultType(resultType) ||
+            binder.customOpNameStringAttr(mode, "mode") ||
+            binder.customOpNameStringAttr(
+                coordTfMode, "coordinate_transformation_mode", "half_pixel"))
```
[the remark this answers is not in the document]
vivekkhandelwal1: Hi @aldesilv, even if we're not supporting them right now. You should add a check saying that this is not yet supported and error out.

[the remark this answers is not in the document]
vivekkhandelwal1: Okay, then please add checks and failure for the same. We can't just ignore them.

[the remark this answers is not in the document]
aldesilv: sometimes those attributes are included but not used, it would error out in those cases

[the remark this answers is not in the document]
vivekkhandelwal1: You can add a check like this: https://github.com/llvm/torch-mlir/blob/0abc5868b5f5cdc0abd123f6dc03853c6bd8c39d/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp#L2533-L2538 for all the unsupported attributes.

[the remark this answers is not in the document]
aldesilv: ok done

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2637,4 +2637,151 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          if (operands.size() < 4) {
+            Value scaleOperand = operands[2];
+            scalesValueList = getValueList(scaleOperand);
+            sizesValueList = noneVal;
+          } else {
+            Value sizesOperand = operands[3];
+            scalesValueList = noneVal;
+            sizesValueList = getValueList(sizesOperand);
+          }
+        } else {
```
rsuderman, line 2775: Try to avoid `if {} else if {} else if{}{ ...` chains. They can be difficult to parse.Instead you can do `if (nearest) { ... }; if(bilinear) {... }` the ncheck the failure of `sizesValueList`. I.e. if its empty return as we know the result is incorrect.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2637,4 +2637,151 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          return failure();
```
rsuderman, line 2776: We should avoid blank `return failure();` Instead we want a `notifyMatchFailure` so that during debug we know _why_ the pass failed to apply.

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2637,4 +2637,152 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          if (operands.size() < 4) {
+            Value scaleOperand = operands[2];
+            scalesValueList = getValueList(scaleOperand);
+            sizesValueList = noneVal;
+          } else {
+            Value sizesOperand = operands[3];
+            scalesValueList = noneVal;
+            sizesValueList = getValueList(sizesOperand);
+          }
+        }
```
rsuderman, line 2778: You are missing the follow 
```
if (scalesValueList.empty()) {
  return rewriter.notifyMatchFailure(op, "unknown scaling mode");
}
```

rsuderman (approved):

ScottTodd:
I'm seeing various downstream failures with this: https://github.com/iree-org/iree/issues/17345.

* A few test cases (like 2/30) are now passing
* A few test cases that were failing are now failing at different places
* A few test cases are compiling but producing incorrect results at runtime
* A few test cases are hanging (either compiling or running, not sure)

PR #3017, opened 2024-03-12
Add accumulator logic for AtenConvOp linalg lowering

nithinsubbiah:


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -943,37 +943,123 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
     // - grouped 1d-3d (quantized)
     // - ungrouped 1d-3d
     if (groupSize == 1 && !inputZp && !weightZp) {
+      // Get default accumulator tensor type for the result tensor and use
+      // it for upcast/downcast if it doesn't match with the result type.
+      Value accumulatorTensor;
+      Type accumulatorTensorType = getDefaultAccType(rewriter, resultDTy);
```
rsuderman, line 949: You should not be calling this `TensorType` as it is an element type of the tensor. `accumulatorDtype` is more correct.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -943,37 +943,123 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+      }
       switch (numSpatialDims) {
       case 1:
-        conv = rewriter
-                   .create<linalg::Conv1DNcwFcwOp>(
-                       loc, outputTensor.getType(),
-                       ValueRange{paddedInput, weight}, outputTensor,
-                       stridesAttr, dilationAttr)
-                   .getResult(0);
+        if (accumulatorTensorType != resultDTy) {
```
rsuderman, line 1002: You can simplify the implementation her by instead always using `accumulatorDType`, then perform truncation at the end if it does not match the result type. This removes the need to branch for all of the cases.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -943,37 +943,123 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+        }
         break;
       default:
         return rewriter.notifyMatchFailure(
             op, "unimplemented: only 1D, 2D, and 3D convolution supported");
       };
       Type newResultType = getTypeConverter()->convertType(op.getType());
-      rewriter.replaceOpWithNewOp<tensor::CastOp>(op, newResultType, conv);
+      if (accumulatorTensorType != resultDTy) {
+        conv = rewriter.create<arith::TruncFOp>(loc, newResultType, conv);
```
rsuderman, line 1058: Not guaranteed to be a floating point type.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -943,37 +943,123 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
     // - grouped 1d-3d (quantized)
     // - ungrouped 1d-3d
     if (groupSize == 1 && !inputZp && !weightZp) {
+      // Get default accumulator tensor type for the result tensor and use
+      // it for upcast/downcast if it doesn't match with the result type.
+      Value accumulatorTensor;
+      Type accumulatorTensorType = getDefaultAccType(rewriter, resultDTy);
+      if (accumulatorTensorType != resultDTy) {
```
rsuderman, line 950: You should be handling this here:
https://github.com/llvm/torch-mlir/pull/3017/files#diff-b9279b005b7ec749040c06d16d460b492e48b69d5d833dc3021454725de59b84R881

You are overriding / throwing away the outputTensor rather than correcting the generated type at creation.

rsuderman (changes requested):

lib/Conversion/TorchToLinalg/Linear.cpp
nithinsubbiah: The reason I did this way is because I didn't check/test if the accumulation logic remains for all other conv logics. I can change to cover other cases too

PR #3018, opened 2024-03-12
[torch] Fix clamp ranges on quantize_per_tensor on unsigned

rsuderman:
SExtValue was used for `int` and `uint` clamp values. This caused the result to always be outputed as `zero`.


zjgarvey:
Hi, with a clang-format, this should be merged right away. Is it possible for me to do this as a reviewer?

zjgarvey (approved):
LGTM

PR #3019, opened 2024-03-13
Add accumulation logic for AtenMmOp linalg lowering

nithinsubbiah:


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -158,10 +160,25 @@ class ConvertAtenMmOp : public OpConversionPattern<AtenMmOp> {
...
+            RankedTensorType::get(zeroFillShapeVector, accumulatorTensorType);
+        zeroFill = rewriter.create<arith::ExtFOp>(
+            loc, accumulatorRankedTensorType, zeroFill);
+      }
       matmul = rewriter
                    .create<linalg::MatmulOp>(loc, zeroFill.getType(),
                                              ValueRange{lhs, rhs}, zeroFill)
                    .getResult(0);
+      if (accumulatorTensorType != resultDTy) {
+        matmul = rewriter.create<arith::TruncFOp>(loc, newResultType, matmul);
```
rsuderman, line 180: This assumes floating point matmul which is not guaranteed. It could be a non-quantized matmul.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -158,10 +160,25 @@ class ConvertAtenMmOp : public OpConversionPattern<AtenMmOp> {
                        loc, zeroFill.getType(), ValueRange{lhs, rhs}, zeroFill)
                    .getResult(0);
     } else {
+      auto accumulatorTensorType = getDefaultAccType(rewriter, resultDTy);
```
rsuderman, line 163: This should not be overwritten. You should be checking the determining the accumulator type above `createZeroInitTensor` then pass it into the zero init work. Doing so will drastically simplify the change as `MatmulUnsignedOp`, `QuantizedMatmulOp`, and `MatmulOp` should all consider accumulator checks.

rsuderman (changes requested):

lib/Conversion/TorchToLinalg/Linear.cpp
nithinsubbiah: That makes sense

lib/Conversion/TorchToLinalg/Linear.cpp
nithinsubbiah: I do agree with you that this could be a non-quantized matmul and if it is then there's no downcast (wouldn't pass the check). I'm not sure I follow

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -127,8 +127,15 @@ class ConvertAtenMmOp : public OpConversionPattern<AtenMmOp> {
               "mismatching contracting dimension for torch.aten.mm"));
     }
 
+    auto resultTy = op.getType().cast<ValueTensorType>();
+    auto resultDTy = resultTy.toBuiltinTensor().getElementType();
     Type newResultType = getTypeConverter()->convertType(op.getType());
     Type elementType = newResultType.cast<TensorType>().getElementType();
+    auto accumulatorDType = getDefaultAccType(rewriter, resultDTy);
+    if (!isa<mlir::NoneType>(accumulatorDType) &&
```
rsuderman, line 135: See comment in other PR - if accumulatorDType is none you should set it to the result type rather than repeating this check everwhere.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -163,6 +170,15 @@ class ConvertAtenMmOp : public OpConversionPattern<AtenMmOp> {
                                              ValueRange{lhs, rhs}, zeroFill)
                    .getResult(0);
     }
+
+    if (!isa<mlir::NoneType>(accumulatorDType) &&
+        accumulatorDType != resultDTy) {
+      Type resultElementType =
+          newResultType.cast<RankedTensorType>().getElementType();
+      matmul = torch_to_linalg::convertTensorToElementType(
```
rsuderman, line 178: See comment in other PR by having convertTensorToElementType be able to handle the noop case.

rsuderman (changes requested):

PR #3023, opened 2024-03-13
Setuptools uses a separate build directory

penguin-wwy:
* setuptools not steal the build directory name https://github.com/llvm/torch-mlir/pull/3021#issuecomment-1994447855
* support pre-built LLVM
* support CMAKE_BUILD_TYPE env

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ externals/pytorch/
 libtorch*
 
 /build/
+/setup_build/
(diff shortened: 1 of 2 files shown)


stellaraccident (approved):
We don't use this in our projects, but it looks ok to me. I haven't touched these files in a long time though. If it's working for you, that's good enough for me.

PR #3024, opened 2024-03-14
[ONNX] LogSoftmax to Torch

zjgarvey:
This PR adds support for onnx.LogSoftmax both for old versions (<13, with axis >=0), and new versions (13).


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -195,6 +195,86 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        Value input;
+        Torch::ValueTensorType resultType;
+        if (binder.tensorOperand(input) || binder.tensorResultType(resultType))
+          return failure();
+
+        int64_t axis;
+        if (binder.s64IntegerAttr(axis, "axis", 1))
+          return rewriter.notifyMatchFailure(binder.op, "axis bind failure");
+        if (axis < 0) {
+          return rewriter.notifyMatchFailure(binder.op, "NYI: negative axis.");
```
renxida, line 227: NYI is an abbreviation i just saw for the first time here. Might be good to just say "Not Impelemented"

renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
zjgarvey: Yeah, can do.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -195,6 +195,87 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        if (!inputTy || !inputTy.hasSizes())
+          return failure();
+        llvm::SmallVector<int64_t> allDims(inputTy.getSizes());
+        llvm::SmallVector<int64_t> leftDims(allDims.begin(),
+                                            allDims.begin() + axis);
+        llvm::SmallVector<int64_t> rightDims(allDims.begin() + axis,
+                                             allDims.end());
+        // flatten post-axis dims
+        int64_t prodRightSizes = 1;
+        for (int64_t n : rightDims) {
```
vivekkhandelwal1, line 255: Instead of creating a new vector, you can just iterate over the allDims vector and use that.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -195,6 +195,87 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        if (binder.tensorOperand(input) || binder.tensorResultType(resultType))
+          return failure();
+
+        int64_t axis;
+        if (binder.s64IntegerAttr(axis, "axis", 1))
+          return rewriter.notifyMatchFailure(binder.op, "axis bind failure");
+        if (axis < 0) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Not yet implemented: negative axis.");
+        }
```
vivekkhandelwal1, line 229: Why have we added this check? Can't we just convert the negative dim to a valid dim by adding rank?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -195,6 +195,87 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          if (n == Torch::kUnknownSize) {
+            prodRightSizes = -1;
+            break;
+          }
+          prodRightSizes *= n;
+        }
+        leftDims.push_back(prodRightSizes);
+        auto flatRightTy = rewriter.getType<Torch::ValueTensorType>(
+            leftDims, inputTy.getOptionalDtype());
+        leftDims.pop_back();
```
vivekkhandelwal1, line 265: I don't think this is required since this value is not used at any instance after this.

vivekkhandelwal1 (changes requested):
@zjgarvey, it would be better if you could add a small description saying why you're doing all this flatten/unflatten thing.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
zjgarvey: Ah, good point. I was originally going to flatten the left too since this would be required for negative dims, but won't be needed here.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
zjgarvey: I think that using a negative dim does the operation on the left instead of the right half, but frankly I'm not sure. I'd need to do some testing since the documentation wasn't clear about what negative axis does.

rsuderman: You should be able to do `axis += inputTy.getSizes().size()`

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -195,6 +195,87 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        Value none = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+        Value cstEnd = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getI64IntegerAttr(rank - axis));
+
+        if (!inputTy || !inputTy.hasSizes())
+          return failure();
+        llvm::SmallVector<int64_t> allDims(inputTy.getSizes());
+        llvm::SmallVector<int64_t> leftDims(allDims.begin(),
+                                            allDims.begin() + axis);
+        llvm::SmallVector<int64_t> rightDims(allDims.begin() + axis,
```
rsuderman, line 251: You can make a `llvm::ArrayRef` for these. It works as a pair of pointers and lets you track it.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
rsuderman: Added a comment above - its not a bad idea to make a separate track of the dims, it just makes more sense to do an `ArrayRef` in this case.

rsuderman (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
zjgarvey: I did use rightDims elsewhere, but I made some changes and let me know if the new work looks better.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
zjgarvey: I made the change you asked about and it looks like, from the e2e tests, that it passes with negative axis as is so I included it. Woops!

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
zjgarvey: Oh, let me try that out right now.

zjgarvey:
I'm going to do what @rsuderman is suggesting with the ArrayRefs and put up a new patch in a short while.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
zjgarvey: I kept leftDims as a vector (because I need to append the product of right dims to the end) but converted the others to ArrayRefs and reorganized for readability.

zjgarvey:
@rsuderman Do you mind checking out the changes I made to see if they make sense? I decided to keep a SmallVector for the left dims because I had to push back the folded dim to make the folded tensor's ValueTensorType.

vivekkhandelwal1 (approved):
LGTM

PR #3025, opened 2024-03-14
Add OnnxToTorch support for Compress op

jinchen62:


lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -704,6 +704,70 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+            numElements *= i;
+          }
+          SmallVector<int64_t> flattenShape = {numElements};
+          auto flattenType = Torch::ValueTensorType::get(
+              binder.op->getContext(), flattenShape, resultType.getDtype());
+          Value flattenTensor = rewriter.create<Torch::AtenFlattenUsingIntsOp>(
+              binder.getLoc(), flattenType, operand, cstZero, cstNegOne);
+          rewriter.replaceOpWithNewOp<Torch::AtenIndexSelectOp>(
+              binder.op, resultType, flattenTensor, cstZero, indexVal);
+        } else {
```
ScottTodd, line 761: Could return early here to reduce nesting.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -704,6 +704,70 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          if (axis < 0)
+            axis += shapeSizes.size();
```
ScottTodd, line 763: Can you link to a spec and/or add a comment here to explain what this is doing? It isn't clear to me what a non-zero axis should mean (maybe this wraps around like a python list index?).

jinchen62: Added a comment in code. Negative axis value means counting dimensions from the back.

ScottTodd (approved):
Seems reasonable to me.

vivekkhandelwal1:
@jinchen62 Did you verify if it's lowering to Linalg or not?

PR #3026, opened 2024-03-14
Bring back `dynamic_shapes` constraints in fx importer API

sjain-stanford:
https://github.com/llvm/torch-mlir/pull/2992 dropped `constraints` from the fx importer API, [breaking](https://github.com/cruise-automation/mlir-tcp/actions/runs/8284385380/job/22669774071) downstream AOT compile tests in `mlir-tcp` that use it. This knob has been soft-deprecated for a while now, replaced by `dynamic_shapes` - a more ergonomic interface. This PR brings back dynamic_shapes constraints in the new supported form. Also added a python lit test with dynamic shaped annotations.

### DIFF
--- a/python/torch_mlir/fx.py
+++ b/python/torch_mlir/fx.py
@@ -3,7 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 # Also available under a BSD-style license. See LICENSE.
 
-from typing import Optional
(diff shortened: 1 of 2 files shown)


stellaraccident (approved):
Oh, thanks. Yeah - I noted that this should have been dropped with replacement but didn't do torch version archaeology. We're using the dynamic_shapes= option in a downstream and it is indeed more ergonomic.

PR #3028, opened 2024-03-15
Normalize type hints to be compatible with multiple Python versions

penguin-wwy:
Although we provide a wheel package for Python 3.8, it may actually throw the following exception:
`TypeError: 'type' object is not subscriptable`


stellaraccident (approved):
There is nothing ensuring that 3.8 doors not regress, and given that the eol date is very close, were unlikely to do anything special for it. But if patches like this get you going, we can accept them.

penguin-wwy:
Our release environment is using Python 3.8, and currently, there are no exceptions when using torchscript compile. In the future, I will try to use the Fx Importer to perform the compilation.

PR #3031, opened 2024-03-16
[FxImporter]: support parsing nodes which return list

Vremold:


stellaraccident (commented):
Thanks for the contribution. If you wouldn't mind, could you add a test? I know the testing in tree is thin for this (we have a bunch of tests out of tree that have not been ported yet), but would like to not keep adding features without tests for structural things.

Existing tests are here: https://github.com/llvm/torch-mlir/tree/main/test/python/fx_importer

Could you create a new op_forms.py and add a case. It should follow how basic_test.py does it. Checks don't need to be super detailed.

Vremold:
> Thanks for the contribution. If you wouldn't mind, could you add a test? I know the testing in tree is thin for this (we have a bunch of tests out of tree that have not been ported yet), but would like to not keep adding features without tests for structural things.
> 
> Existing tests are here: https://github.com/llvm/torch-mlir/tree/main/test/python/fx_importer
> 
> Could you create a new op_forms.py and add a case. It should follow how basic_test.py does it. Checks don't need to be super detailed.

Thanks for the kind advice. A simple test case for list_return_test.py which contains only one case of torch.unbind.int op is added.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -888,6 +888,20 @@ def node_val_to_type(self, node: torch_fx.Node, *, mutable: bool = False) -> IrT
             tensor_meta = node.meta.get("tensor_meta")
             val = node.meta.get("val")
             sparsity = node.meta.get("sparsity", None)
+            # Some nodes returns a list, like torch.ops.aten.unbind.int
+            if isinstance(tensor_meta, List) or isinstance(val, List):
```
stellaraccident, line 892: I am pretty sure that the second clause should be `isinstance(val, list)`: `List` is a type descriptor that is never instantiated. `list` is the runtime type.

stellaraccident (commented):
Thank you for the contribution. Since this stuff gets a bit twisty, do you mind if I take a bit of time and poke at this so that I understand? If I make changes, I'll extend your patch and land as co-author. These things are hard to review in the abstract without actually trying things.

Vremold:
> Thank you for the contribution. Since this stuff gets a bit twisty, do you mind if I take a bit of time and poke at this so that I understand?

OK.

RuiminChen:
@Vremold Thank you for your work! I've encountered a similar issue myself, and this pull request effectively addresses my problem. I noticed that this PR has not yet been merged. Would it be possible to resolve the merge conflicts and proceed with the merge? Your prompt attention would be greatly appreciated. Many thanks!

Vremold:
@rog93 No problem.

rog93:
> @rog93 No problem. Done.

Appreciate it~ Seems like there are code conflicts that need resolving before the merge can proceed.

test/python/fx_importer/list_return_test.py
```diff
@@ -0,0 +1,39 @@
...
+def test_unbind_int_op():
+    class UnbindIntModule(nn.Module):
+        def __init__(self):
+            super().__init__()
+
+        def forward(self, x):
+            return torch.unbind(x, 1)
+
+    m = fx.export_and_import(
+        UnbindIntModule(), torch.randn(3, 4), func_name="test_unbind_int"
```
rog93, line 37: I think `decomposition_table=False` should be added here to pass the CI test.

rog93:
May I ask if this commit is ready to be merged now, or do we still need approvals from other reviewers?

PR #3038, opened 2024-03-19
[MLIR][TORCH] Add lit test support for Onnx op lowering to Linalg

vivekkhandelwal1:
This commit adds the support for the ops which are lowered from OnnxToTorch but are not tested for the path OnnxToLinalg. With this commit all the current OnnxToTorch lit tests are also tested for the OnnxToLinalg path. Also, the tests which are failing are added in a seperate file from where they can be moved to the original test file as and when they are fixed.

This commit will ensure that any new Onnx op lowering added or the existing lowering modified not only lowers to Torch but Linalg.

### DIFF
--- a/test/Conversion/TorchOnnxToTorch/simple_ops_a_to_f.mlir
+++ b/test/Conversion/TorchOnnxToTorch/simple_ops_a_to_f.mlir
@@ -1,4 +1,5 @@
 // RUN: torch-mlir-opt <%s --split-input-file -convert-torch-onnx-to-torch | FileCheck %s
+// RUN: torch-mlir-opt <%s --split-input-file -convert-torch-onnx-to-torch --torch-decompose-complex-ops --canonicalize --torch-backend-to-linalg-on-tensors-backend-pipeline
 // Generally, the test cases accumulated here come from running the importer
 // over all included backend tests that involve simple ops with no model
 // level constants. This is a pragmatic choice which lets us have a lot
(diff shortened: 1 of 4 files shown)


test/Conversion/TorchOnnxToTorch/simple_ops_a_to_f.mlir
```diff
@@ -1,4 +1,5 @@
 // RUN: torch-mlir-opt <%s --split-input-file -convert-torch-onnx-to-torch | FileCheck %s
+// RUN: torch-mlir-opt <%s --split-input-file -convert-torch-onnx-to-torch --torch-decompose-complex-ops --canonicalize --torch-backend-to-linalg-on-tensors-backend-pipeline
```
ScottTodd, line 2: What is this actually testing, without piping to FileCheck?

Rather than splice the files this way and have them tested through two different pipelines, how about forking the test cases into a new `TorchOnnxToLinalg` folder where linalg can be tested directly?

(Alternately keep the test cases in a single file but use `FileCheck -check-prefix=CHECK-LINALG` to differentiate between the different `RUN` commands)

ScottTodd (changes requested):

test/Conversion/TorchOnnxToTorch/simple_ops_a_to_f.mlir
vivekkhandelwal1: @kumardeepakamd @rsuderman Would you like to add something here?

vivekkhandelwal1:
Closing this PR, since I think it's not required right now. We can revisit this if required later.

PR #3040, opened 2024-03-19
Fix SCF Forloop fails to convert to linalg when a tensor argument is supplied to the loop block

renxida:


test/Conversion/TorchToSCF/tensorargs.mlir
```diff
@@ -0,0 +1,25 @@
+// RUN: torch-mlir-opt <%s -convert-torch-to-scf -split-input-file -verify-diagnostics | FileCheck %s
+// CHECK-LABEL:   func.func @test_forloop_tensorargs() -> !torch.vtensor<[2,3],f32> {
+// CHECK:           %[[LOOP_RESULT:.*]] = scf.for %[[VAL_17:.*]] = %[[VAL_13:.*]] to %[[VAL_15:.*]] step %[[VAL_14:.*]] iter_args(%[[LOOP_TENSOR_ARG:.*]] = %[[VAL_11:.*]]) -> (tensor<2x3xf32>) {
```
rsuderman, line 3: Pick better names than `VAL_15` etc. Better to name with meaningful parts.

test/Conversion/TorchToSCF/tensorargs.mlir
rsuderman, line 1: You should be putting this in an existing file not making a new one.

rsuderman (changes requested):
We should included a simple execution test in the e2e example

renxida:
something i'm concerned about the test case is it doesnt work when i give a tensor as the iteration count 

like this:

```
class TorchPrimLoop_ForLike_TensorArg_NoUnroll_Module(torch.nn.Module):
    def __init__(self):
        super().__init__()

    @export
    @annotate_args([
        None,
        ([7,9], torch.float32, True),
        ([1], torch.int64, False)
    ])
    def forward(self, x: torch.Tensor, n) -> torch.Tensor:
        for i in range(int(n[0])):
            x = x + i
        return x

@register_test_case(module_factory=lambda: TorchPrimLoop_ForLike_TensorArg_NoUnroll_Module())
def TorchPrimLoop_ForLike_TensorArg_NoUnroll_Module_basic(module, tu: TestUtils):
    x_test = torch.zeros([7, 9]).float()
    n_test = torch.tensor([50], dtype=torch.int64)

    module.forward(x_test, n_test)

```

I suspect there's something going on with unrolling

renxida:
Error message:
```


        error: unknown: unsupported by backend contract: tensor with unknown rank
        note: unknown: see current operation: %3 = "torch.tensor_static_info_cast"(%arg0) : (!torch.vtensor<[7,9],f32>) -> !torch.vtensor
        note: unknown: this is likely due to a missing transfer function in abstract_interp_lib_gen.py

        For Torch-MLIR developers, the error can be reproduced with:
        $ torch-mlir-opt -pass-pipeline='builtin.module(torchscript-module-to-torch-backend-pipeline{backend-legal-ops=aten.flatten.using_ints,aten.adaptive_avg_pool1d,aten.unflatten.int extra-library=})' /tmp/TorchPrimLoop_ForLike_TensorArg_NoUnroll_Module.mlir
        Add '-mlir-print-ir-after-all -mlir-disable-threading' to get the IR dump for debugging purpose.



Summary:
    Failed: 1
```

renxida:
I banged my head against it a bunch without luck. Current guess is the forloop somehow removes shape information but unrolling somehow preserves it.

rsuderman (approved):

PR #3041, opened 2024-03-19
[ONNX] Reduces onnx.Div sinceVersion to 7

zjgarvey:
The only difference between version 7 and newer versions is support for different data types. We should allow this pattern to match as early as 7. Earlier versions have a more manual broadcast specification through attributes, so I did not include those versions. 

See: [onnx.Div docs](https://onnx.ai/onnx/operators/onnx__Div.html#l-onnx-doc-divl)


renxida (approved):

renxida:
Noice!

zjgarvey:
Thanks!

PR #3045, opened 2024-03-21
[fx] Make the lift_fresh_copy -> clone special form use kwargs.

stellaraccident:
At some point, this op became kwarg-only instead of arg/kwarg. Discovered when upgrading to PyTorch 2.3.

Also adds a test as this was untested in-tree (was caught out of tree).


dan-garvey (approved):
lgtm

dan-garvey:
double checking to see if the test is run by the ci since I don't see it in logs

stellaraccident:
> double checking to see if the test is run by the ci since I don't see it in logs
> 
> edit ok, I see lit tests get done as part of the check after the build.

Yeah, lit is tight lipped. This indicates it visited one of the sub-directories:   [100/103] Running the torch-mlir regression tests
  Enabling Torch v2.3+ tests

PR #3046, opened 2024-03-22
[fx] Use module Operations instead of Module.

stellaraccident:
This was only used in certain advanced uses of the API that want to build into their own module. The MLIR `Module` class is an awkward/restrictive way to require this to go as only some things will have it. Just switch everything to be based on a module `Operation`.


dan-garvey (approved):
looks good

dan-garvey:
Looks like there is some underlying execution engine change needed to support operation versus module?

stellaraccident:
Ugh. There is stilly tech debt in the ExecutionEngine that needs to be excised...

PR #3049, opened 2024-03-22
[MLIR][TORCH] Add OnnxToTorch lowering for ops

vivekkhandelwal1:
This commit adds the OnnxToTorch lowering for the Mish, Softplus, HardSwish, Trilu, ThresholdedRelu op

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2099,4 +2099,67 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                       binder.op, resultType, operand);
                   return success();
                 });
+  patterns.onOp(
+      "Softplus", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value input;
+        if (binder.tensorOperand(input) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+        // out = ln(exp(x) + 1)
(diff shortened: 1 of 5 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2099,4 +2099,67 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+      "Softplus", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value input;
+        if (binder.tensorOperand(input) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+        // out = ln(exp(x) + 1)
+        Value exp = rewriter.create<Torch::AtenExpOp>(binder.getLoc(),
+                                                      resultType, input);
```
zjgarvey, line 2112: Why not use [AtenSoftplusOp](https://github.com/llvm/torch-mlir/blob/50635dd50922c71d52120b8cb38c229291a3c7ba/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L3832)? Are you just trying to avoid multiplying by some ones?

zjgarvey (approved):
Looks great. The Trilu tests are very extensive, nice.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: Yeah. It will create two more multiplies. I think it's better to lower it this way.

PR #3051, opened 2024-03-25
[MLIR][TORCH] Support parallel dimemsions expand/collapse

xinan-jiang:
This PR support `aten.view` with unique unknown dimension both in input shape and output shape while the pass convert-torch-to-linalg that lowing `aten.view` to `tensor.collapse_shape` or `tensor.expand_shape`.

Below is an example
```
func.func @test_reshape(%arg0: !torch.vtensor<[1,?,50,16],f32>) -> !torch.vtensor<[1,?,16],f32> attributes {torch.assume_strict_symbolic_shapes, torch.onnx_meta.ir_version = 9 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
  %int1 = torch.constant.int 1
  %int-1 = torch.constant.int -1
  %int16 = torch.constant.int 16
  %0 = torch.prim.ListConstruct %int1, %int-1, %int16 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %1 = torch.aten.view %arg0, %0 : !torch.vtensor<[1,?,50,16],f32>, !torch.list<int> -> !torch.vtensor<[1,?,16],f32>
  return %1 : !torch.vtensor<[1,?,16],f32>
}
```


xinan-jiang:
@ramiro050 @rsuderman @silvasean Could you help to review this PR?

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -789,6 +792,63 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
     return success();
   }
 
+  // Starting from the beginning of the dims arrays, this helper finds the
+  // smallest set of consecutive dims in each array such that the product of the
+  // dim sizes in the two subsets is equal. The indices arrays are populated
+  // with the indices of the dims arrays that correspond to the subsets found.
+  //
+  // An error is returned if two subsets of dims with total number of elements
+  // equal to each other is not found.
```
ramiro050, line 801: this seems to be copied over from the previous helper. Can you update it to explain what this function does?

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1020,6 +1099,14 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
           /// `mapStaticallyKnownDims` maps the smallest number of
           /// input and output dimensions in the slice statically
           /// known to have the same number of elements.
+        } else if (succeeded(mapParallelUnknownDims(
```
ramiro050, line 1102: Can you add an e2e test showing the newly supported case? Example: https://github.com/llvm/torch-mlir/blob/40008b025a40e8bab1f362edc1ed0ee551b5e7a3/projects/pt1/python/torch_mlir_e2e_test/test_suite/reshape_like.py#L363

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -770,6 +771,8 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
     SmallVector<int64_t> yIndicesResult({0});
     size_t nextXIndex = 1;
     size_t nextYIndex = 1;
+    if (xTotalSize == kUnknownSize)
```
ramiro050, line 774: nit: can you move this to before the vectors `xIndicesResult` and `yIndicesResult` are created, to avoid unnecessary computation? Also, I think we should make the same check for `yTotalSize` and return `failure` if it is unknown at the start.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -844,6 +904,21 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
     return std::make_pair(inputShape, outputShape);
   }
 
+  static std::pair<int64_t, int64_t>
+  getMultiplier(SmallVector<int64_t> inputShape,
```
ramiro050, line 908: nit: can you add a comment explaining what this does and why it is needed?

ramiro050 (changes requested):
Thanks for working on this! A few initial comments. Also, can you paste here what the MLIR looks like after converting to linalg for your example?:

```mlir
func.func @test_reshape(%arg0: !torch.vtensor<[1,?,50,16],f32>) -> !torch.vtensor<[1,?,16],f32> attributes {torch.assume_strict_symbolic_shapes, torch.onnx_meta.ir_version = 9 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
  %int1 = torch.constant.int 1
  %int-1 = torch.constant.int -1
  %int16 = torch.constant.int 16
  %0 = torch.prim.ListConstruct %int1, %int-1, %int16 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %1 = torch.aten.view %arg0, %0 : !torch.vtensor<[1,?,50,16],f32>, !torch.list<int> -> !torch.vtensor<[1,?,16],f32>
  return %1 : !torch.vtensor<[1,?,16],f32>
}
```

xinan-jiang:
> Thanks for working on this! A few initial comments. Also, can you paste here what the MLIR looks like after converting to linalg for your example?:
> 
> ```mlir
> func.func @test_reshape(%arg0: !torch.vtensor<[1,?,50,16],f32>) -> !torch.vtensor<[1,?,16],f32> attributes {torch.assume_strict_symbolic_shapes, torch.onnx_meta.ir_version = 9 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
>   %int1 = torch.constant.int 1
>   %int-1 = torch.constant.int -1
>   %int16 = torch.constant.int 16
>   %0 = torch.prim.ListConstruct %int1, %int-1, %int16 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
>   %1 = torch.aten.view %arg0, %0 : !torch.vtensor<[1,?,50,16],f32>, !torch.list<int> -> !torch.vtensor<[1,?,16],f32>
>   return %1 : !torch.vtensor<[1,?,16],f32>
> }
> ```

the result after convert-torch-to-linalg
```
func.func @test_reshape(%arg0: !torch.vtensor<[1,?,50,16],f32>) -> !torch.vtensor<[1,?,16],f32> attributes {torch.assume_strict_symbolic_shapes, torch.onnx_meta.ir_version = 9 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
  %0 = torch_c.to_builtin_tensor %arg0 : !torch.vtensor<[1,?,50,16],f32> -> tensor<1x?x50x16xf32>
  %int1 = torch.constant.int 1
  %int-1 = torch.constant.int -1
  %int16 = torch.constant.int 16
  %1 = torch.prim.ListConstruct %int1, %int-1, %int16 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
  %c1_0 = arith.constant 1 : index
  %dim = tensor.dim %0, %c1_0 : tensor<1x?x50x16xf32>
  %c2 = arith.constant 2 : index
  %c50 = arith.constant 50 : index
  %c3 = arith.constant 3 : index
  %c16 = arith.constant 16 : index
  %2 = torch_c.to_i64 %int1
  %3 = torch_c.to_i64 %int-1
  %4 = torch_c.to_i64 %int16
  %collapsed = tensor.collapse_shape %0 [[0], [1, 2], [3]] : tensor<1x?x50x16xf32> into tensor<1x?x16xf32>
  %5 = torch_c.from_builtin_tensor %collapsed : tensor<1x?x16xf32> -> !torch.vtensor<[1,?,16],f32>
  return %5 : !torch.vtensor<[1,?,16],f32>
}
```

lib/Conversion/TorchToLinalg/DataMovement.cpp
xinan-jiang: done

lib/Conversion/TorchToLinalg/DataMovement.cpp
xinan-jiang: I have added the e2e test. However, I don't know how to run the e2e test. Could you give me some guide?

lib/Conversion/TorchToLinalg/DataMovement.cpp
xinan-jiang: done

lib/Conversion/TorchToLinalg/DataMovement.cpp
xinan-jiang: done

xinan-jiang:
@ramiro050

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: See: https://github.com/llvm/torch-mlir/blob/main/docs/development.md#running-execution-end-to-end-tests

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -789,6 +792,67 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+      xTotalSize *= xMultiplier;
+    if (yTotalSize == kUnknownSize)
+      yTotalSize *= yMultiplier;
+
+    while (xTotalSize != yTotalSize) {
+      if (xTotalSize < 0 && yTotalSize < 0) {
+        xTotalSize = -xTotalSize;
+        yTotalSize = -yTotalSize;
+      }
+      if (yTotalSize < 0 || (xTotalSize > 0 && xTotalSize < yTotalSize)) {
```
ramiro050, line 834: nit: compare directly with `kUnknownSize`, since the value of `kUnknownSize is an implementation detail that should not be relied on

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -789,6 +792,67 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+        if (nextXIndex == xDims.size())
+          return failure();
+        if (xDims[nextXIndex] == kUnknownSize)
+          xTotalSize *= xMultiplier;
+        xTotalSize *= xDims[nextXIndex];
```
ramiro050, line 839: nit: the `xTotalSize *= xDims[nextXIndex];` should be in an `else` branch to avoid making things negative when the value is equal to `kUnknownSize`

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -789,6 +792,67 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+        xIndicesResult.push_back(nextXIndex++);
+      } else {
+        if (nextYIndex == yDims.size())
+          return failure();
+        if (yDims[nextYIndex] == kUnknownSize)
+          yTotalSize *= yMultiplier;
+        yTotalSize *= yDims[nextYIndex];
```
ramiro050, line 846: Same comment as above: move this to `else` branch

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -844,6 +908,23 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
     return std::make_pair(inputShape, outputShape);
   }
 
+  // Gets the ratio between the unknown dimensions in the input shape and the
+  // output shape. This ratio is used to match parallel unknown dimensions.
+  static std::pair<int64_t, int64_t>
+  getMultiplier(SmallVector<int64_t> inputShape,
+                SmallVector<int64_t> outputShape) {
+    int64_t inputMultiplier = 1;
+    int64_t outputMultiplier = 1;
```
ramiro050, line 917: nit: no need to set these to 1. I would move the initialization down to where the final value is computed

ramiro050 (approved):
Looking good, final set of comments

ramiro050:
> @ramiro050 how to deal with below exception? this test case is added in this PR
> 
> ```
>  ****** Unexpectedly Passed tests - 1 tests
>       XPASS - "ViewDynamicExpandCollapseWithParallelUnknownDimModule_basic"
> ```

You need to add it to the TOSA set: https://github.com/llvm/torch-mlir/blob/ec4cb8be446da8c118675ff7db2ea279ce3effe1/projects/pt1/e2e_testing/xfail_sets.py#L789

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -789,6 +792,67 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+    int64_t xTotalSize = xDims[0];
+    int64_t yTotalSize = yDims[0];
+    SmallVector<int64_t> xIndicesResult({0});
+    SmallVector<int64_t> yIndicesResult({0});
+    size_t nextXIndex = 1;
+    size_t nextYIndex = 1;
+    if (xTotalSize == kUnknownSize)
+      xTotalSize *= xMultiplier;
+    if (yTotalSize == kUnknownSize)
+      yTotalSize *= yMultiplier;
```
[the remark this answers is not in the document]
xinan-jiang: Good suggestion

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -789,6 +792,67 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+
+    while (xTotalSize != yTotalSize) {
+      if (xTotalSize < 0 && yTotalSize < 0) {
+        xTotalSize = -xTotalSize;
+        yTotalSize = -yTotalSize;
+      }
```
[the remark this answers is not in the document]
xinan-jiang: done

lib/Conversion/TorchToLinalg/DataMovement.cpp
xinan-jiang: done

lib/Conversion/TorchToLinalg/DataMovement.cpp
xinan-jiang: done

lib/Conversion/TorchToLinalg/DataMovement.cpp
xinan-jiang: done

lib/Conversion/TorchToLinalg/DataMovement.cpp
xinan-jiang: done

xinan-jiang:
I think this error is not caused by my code.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -789,6 +792,81 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+      return failure();
+
+    int64_t xTotalSize = xDims[0];
+    int64_t yTotalSize = yDims[0];
+    SmallVector<int64_t> xIndicesResult({0});
+    SmallVector<int64_t> yIndicesResult({0});
+    size_t nextXIndex = 1;
+    size_t nextYIndex = 1;
+    bool xHasUnknownSize = false;
+    bool yHasUnknownSize = false;
```
ramiro050, line 825: nit: there is no need for these, since it is already ensured by the `llvm::count` check above that there is at most one unknown size in each array

ramiro050 (changes requested):

projects/pt1/python/torch_mlir_e2e_test/test_suite/reshape_like.py
```diff
@@ -379,6 +379,25 @@ def ViewDynamicExpandCollapseModule_basic(module, tu: TestUtils):
...
+        super().__init__()
+
+    @export
+    @annotate_args([
+        None,
+        ([2, 3, -1, 5], torch.float32, True),
+    ])
+
+    def forward(self, a):
+        return a.view(2, -1, 6)
```
ramiro050, line 393: Is this the right result size? If I understand correctly, the case you're handling here is `a.view(2, -1, 5)`, right?

ramiro050 (changes requested):

projects/pt1/python/torch_mlir_e2e_test/test_suite/reshape_like.py
xinan-jiang: the input size is 2x3x`a`x5,  which `a` = 4, 120 elements totally. the output size is 2x`b`x6， when b is 10, the total elements is 120, too.

lib/Conversion/TorchToLinalg/DataMovement.cpp
xinan-jiang: It is necessary. because the bool flags updated in the while loop.

projects/pt1/python/torch_mlir_e2e_test/test_suite/reshape_like.py
ramiro050: Oh, that's really nice! thanks for the explanation

lib/Conversion/TorchToLinalg/DataMovement.cpp
ramiro050: Makes sense, thanks!

ramiro050:
@rsuderman, do you know what the error in the ONNX test is due to? I think the error in the CI is due to the test crashing, but not completely sure. @xinan-jiang, can you try adding it to the `ONNX_CRASH_SET`?

xinan-jiang:
> @rsuderman, do you know what the error in the ONNX test is due to? I think the error in the CI is due to the test crashing, but not completely sure. @xinan-jiang, can you try adding it to the `ONNX_CRASH_SET`?

I following your suggestion, but it is still failed.

ramiro050:
That's very strange. Let me try to reproduce locally.

ramiro050:
Okay, I figured out the issue. You changes were causing an already `XFAIL` test in ONNX (`FlattenDynamicModule_basic`) to take a slightly different route in the `view` lowering leading to a crash. 

Can you add the following check to prevent inserting `assert inputShape[inputDim] == -1` into the IR?

```diff
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -1121,8 +1121,10 @@ public:
           assumedDynamicDimNotSplit = true;
         } else if (inputShapeSlice[0] == kUnknownSize) {
           // Defer the dynamic shape check to avoid DialectConversion assertion:
-          checkDimPairs.push_back(
-              std::pair<int64_t, int64_t>(inputDim, outputDim));
+          if (outputShapeSlice[0] != kUnknownSize) {
+            checkDimPairs.push_back(
+                std::pair<int64_t, int64_t>(inputDim, outputDim));
+          }

           inputShape[inputDim] = outputShape[outputDim];
           inputSliceIndices.push_back(0);
```

ramiro050:
I've verified that with the change above, the ONNX tests behave properly.

xinan-jiang:
I think it's ok now.

ramiro050 (approved):
Thanks!

PR #3055, opened 2024-03-26
[torch] Improve shape inference for `torch-to-linalg` path for reshapes

rsuderman:
Reshaping tensors depend on directly matching individual dimensions to their corresponding dim in the `torch.view` reshape dimensions. This involves decoupling dynamic dimensions from their static counterparts and support cleanup / canonicalization.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/Transforms/Passes.td
+++ b/include/torch-mlir/Dialect/Torch/Transforms/Passes.td
@@ -235,6 +235,11 @@ def DecomposeComplexOps : Pass<"torch-decompose-complex-ops", "func::FuncOp"> {
   }];
 }
 
+def ScalarizeShapes : Pass<"torch-scalarize-shapes", "func::FuncOp"> {
+  let summary = "Takes common shape computation operations and scalarizes them to";
+  let constructor = "mlir::torch::Torch::createScalarizeShapesPass()";
+}
+
 def RecomposeComplexOps : Pass<"torch-recompose-complex-ops", "func::FuncOp"> {
   let summary = "Recompose torch operations that have been decomposed by TorchScript";
   let constructor = "mlir::torch::Torch::createRecomposeComplexOpsPass()";
--- a/include/torch-mlir/Dialect/Torch/Transforms/Passes.td
+++ b/include/torch-mlir/Dialect/Torch/Transforms/Passes.td
@@ -236,8 +236,14 @@ def DecomposeComplexOps : Pass<"torch-decompose-complex-ops", "func::FuncOp"> {
 }
 
 def ScalarizeShapes : Pass<"torch-scalarize-shapes", "func::FuncOp"> {
-  let summary = "Takes common shape computation operations and scalarizes them to";
+  let summary = "Takes common shape computation operations and scalarizes them.";
   let constructor = "mlir::torch::Torch::createScalarizeShapesPass()";
+  let description = [{
+    Scalarizes shape computations to better propagate static shapes. As some
+    shape operation happen on tensors a single dynamic dimension can prevent
+    propagating static shapes. Scalarization prevents these dynamic
+    dimensions from blocking the statically computable operations.
+  }];
 }
 
 def RecomposeComplexOps : Pass<"torch-recompose-complex-ops", "func::FuncOp"> {
--- a/lib/Conversion/CMakeLists.txt
+++ b/lib/Conversion/CMakeLists.txt
@@ -1,7 +1,7 @@
 add_subdirectory(TorchOnnxToTorch)
+add_subdirectory(TorchToArith)
 add_subdirectory(TorchToLinalg)
 add_subdirectory(TorchToSCF)
-add_subdirectory(TorchToArith)
 add_subdirectory(TorchToTensor)
 add_subdirectory(TorchToTosa)
 if(TORCH_MLIR_ENABLE_STABLEHLO)
@@ -12,9 +12,9 @@ add_subdirectory(TorchConversionToMLProgram)
 add_subdirectory(Utils)
 
 # TODO: Automate this with add_torch_mlir_conversion_library.
-set(linked_libs TorchMLIRTorchToLinalg
+set(linked_libs TorchMLIRTorchToArith
+                TorchMLIRTorchToLinalg
                 TorchMLIRTorchToSCF
-                TorchMLIRTorchToArith
                 TorchMLIRTorchToTensor
                 TorchMLIRTorchToTosa
                 TorchMLIRTorchToTMTensor
--- a/lib/Conversion/TorchToArith/TorchToArith.cpp
+++ b/lib/Conversion/TorchToArith/TorchToArith.cpp
@@ -240,6 +240,23 @@ public:
 } // namespace
 
 namespace {
+class ConvertAtenIntBoolOp : public OpConversionPattern<AtenIntBoolOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenIntBoolOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Type resultType =
+        this->getTypeConverter()->convertType(op->getResult(0).getType());
+    Value result =
+        convertScalarToDtype(rewriter, op.getLoc(), adaptor.getA(), resultType);
+    rewriter.replaceOp(op, result);
+    return success();
+  }
+};
+} // namespace
+
+namespace {
 class ConvertAtenFloatScalarOp : public OpConversionPattern<AtenFloatScalarOp> {
 public:
   using OpConversionPattern::OpConversionPattern;
@@ -432,6 +449,9 @@ public:
 
     target.addIllegalOp<AtenFloatScalarOp>();
     patterns.add<ConvertAtenFloatScalarOp>(typeConverter, context);
+
+    target.addIllegalOp<AtenIntBoolOp>();
+    patterns.add<ConvertAtenIntBoolOp>(typeConverter, context);
 
     target.addIllegalOp<AtenAddOp>();
     patterns.add<ConvertAtenAddOp>(typeConverter, context);
--- a/lib/Conversion/TorchToLinalg/TensorScalarInterop.cpp
+++ b/lib/Conversion/TorchToLinalg/TensorScalarInterop.cpp
@@ -191,6 +191,61 @@ public:
 } // namespace
 
 namespace {
+class ConvertAtenFullOp : public OpConversionPattern<AtenFullOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenFullOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+    Location loc = op.getLoc();
+
+    SmallVector<Value> inShape;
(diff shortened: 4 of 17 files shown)


lib/Conversion/TorchToArith/TorchToArith.cpp
```diff
@@ -239,6 +239,23 @@ class ConvertTorchConstantIntOp
...
+class ConvertAtenIntBoolOp : public OpConversionPattern<AtenIntBoolOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenIntBoolOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Type resultType =
+        this->getTypeConverter()->convertType(op->getResult(0).getType());
+    Value result =
+        convertScalarToDtype(rewriter, op.getLoc(), adaptor.getA(), resultType);
```
renxida, line 252: oh nice. really wish we have a glossary of helper functions.

lib/Conversion/TorchToTensor/TorchToTensor.cpp
```diff
@@ -105,6 +105,43 @@ class ConvertAtenShapeToTensorPatternOp
   }
 };
 
+class ConvertAtenTensorOpPattern : public OpConversionPattern<AtenTensorOp> {
```
renxida, line 108: oh nice this is another wall i had to work around when implementing LSTM

lib/Conversion/TorchToTensor/TorchToTensor.cpp
```diff
@@ -105,6 +105,43 @@ class ConvertAtenShapeToTensorPatternOp
...
+    auto list = op.getData().getDefiningOp<Torch::PrimListConstructOp>();
+    if (!list)
+      return failure();
+
+    auto typeConverter = getTypeConverter();
+    auto resultTy = cast<ShapedType>(typeConverter->convertType(op.getType()));
+    auto resultETy = resultTy.getElementType();
+
+    SmallVector<Value> values;
+    for (Value operand : list.getOperands()) {
```
renxida, line 125: would something like this work with AtenAppendTOp or is the overall strategy just to avoid the mutating ops?

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -715,16 +715,58 @@ OpFoldResult AtenNeBoolOp::fold(FoldAdaptor adaptor) {
...
+  }
+  return nullptr;
+}
+
 //===----------------------------------------------------------------------===//
 // AtenSqueezeOp
 //===----------------------------------------------------------------------===//
 
 OpFoldResult AtenSqueezeOp::fold(FoldAdaptor adaptor) {
-  if (getOperand().getType() != getResult().getType())
```
renxida, line 723: this func is small enough for this to not be confusing but i think i prefer selfTy for consistency with the rest of the codebase

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -0,0 +1,341 @@
+//===----------------------------------------------------------------------===//
```
renxida, line 1: the reason we don't have anything here explaining what ScalarizeShapes does is because we're explaining it in the Summary = line, right?

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -0,0 +1,341 @@
...
+    return success();
+  }
+};
+} // namespace
+
+namespace {
+class PropagateAtenSliceTensorPattern
+    : public OpRewritePattern<AtenSliceTensorOp> {
+public:
+  using OpRewritePattern<AtenSliceTensorOp>::OpRewritePattern;
```
renxida, line 155: this could use a few words of comment explaining when it applies. From my reading it seems to do this:
- find aten.slice.tensor(aten.tensor(x), dims)
- return aten.tensor(x[dims])
- where x is a list of values
- and the tensor has to have shape 1...1, x, 1...1

renxida (approved):

lib/Conversion/TorchToTensor/TorchToTensor.cpp
rsuderman: Yes / no. We could do a canonicalizer that checks whether the input is a list constructor and replace with a new list constructor however this would not solve your issue. Mutative operations (e.g. list append) are generally hard to represent without a native types - especially for canonicalization.

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
rsuderman: Yes, I also flushed out a greater description so there is more justification.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -715,16 +715,58 @@ OpFoldResult AtenNeBoolOp::fold(FoldAdaptor adaptor) {
...
+  if (getSelf().getType() != getResult().getType())
+    return nullptr;
+  if (ty && rty) {
+    if (ty.hasSizes() && rty.hasSizes() &&
+        ty.getSizes().size() == rty.getSizes().size())
+      return getSelf();
+  }
+  return nullptr;
+}
+
```
zjgarvey, line 745: I haven't written a folder before, so I have a few questions:

1. What is the difference between "getSelf()" and "adaptor.getSelf()" I assume "getSelf()" is getting "self"/the input arg from the Unsqueeze op being folded, but I'm not sure about the latter. 
2. I'm similarly confused about getType(). Is this equivalent to "op.getResult().getType()"? 
3. Is the purpose of this folder to convert the UnsqueezeOp to a splat array if possible and to completely ignore the unsqueeze if no dims were actually unsqueezed?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -684,6 +684,13 @@ class DecomposeAtenSelectIntOp : public OpRewritePattern<AtenSelectIntOp> {
                              /*keepDim=*/true),
         op.getSelf(), dim, start, startPlusOne, /*step=*/one);
 
+    auto sliceTy = cast<BaseTensorType>(slice.getType());
+    auto resultTy = cast<BaseTensorType>(op.getResult().getType());
+    if (sliceTy.getSizes().size() == resultTy.getSizes().size()) {
```
zjgarvey, line 689: Okay, so you are checking if the slice should be squeezed or not, correct? Is this sort of like double protecting against the no-op squeezes, since the folder above should get rid of them in that case anyway?

zjgarvey (commented):
I suppose @renxida and I are looking at this at the same time, sorry for the double comments if there are any.

lib/Conversion/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 add_subdirectory(TorchOnnxToTorch)
+add_subdirectory(TorchToArith)
```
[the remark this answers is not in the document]
rsuderman: Its just cleanliness. I noticed this so I quickly cleaned it up.

lib/Conversion/TorchToLinalg/TensorScalarInterop.cpp
```diff
@@ -190,6 +190,61 @@ class ConvertPrimNumToTensorScalarOp
...
+    Value full = adaptor.getFillValue();
+
+    if (full.getType() != resultTy.getElementType()) {
+      if (isa<mlir::FloatType>(full.getType())) {
+        full = rewriter.create<arith::TruncFOp>(loc, resultTy.getElementType(),
+                                                full);
+      } else if (isa<mlir::IntegerType>(full.getType())) {
+        full = rewriter.create<arith::TruncIOp>(loc, resultTy.getElementType(),
+                                                full);
+      }
```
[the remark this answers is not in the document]
rsuderman: For `torch` - yes. there is also a `index` type but that won't be relevant.

lib/Conversion/TorchToArith/TorchToArith.cpp
```diff
@@ -239,6 +239,23 @@ class ConvertTorchConstantIntOp
 };
 } // namespace
 
+namespace {
```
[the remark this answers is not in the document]
rsuderman: We usually generalize once duplicating. It can be good habit to check whether something similar already exists before replicating.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -715,11 +715,43 @@ OpFoldResult AtenNeBoolOp::fold(FoldAdaptor adaptor) {
...
 
+//===----------------------------------------------------------------------===//
+// AtenUnsqueezeOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult AtenUnsqueezeOp::fold(FoldAdaptor adaptor) {
+  if (auto attr = dyn_cast_or_null<DenseElementsAttr>(adaptor.getSelf())) {
+    auto ty = cast<RankedTensorType>(attr.getType());
+    auto rty = cast<BaseTensorType>(getType());
+    if (rty.hasSizes() && rty.areAllSizesKnown() && attr.isSplat()) {
```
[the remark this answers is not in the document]
rsuderman: I am not certain myself. I generally call `hasSizes` to guarantee its safe but I am not totally positive.

lib/Dialect/Torch/IR/TorchOps.cpp
rsuderman: 1. Type conversion - `getSelf()` gets the non-converted version. I.e. the source version. While `adaptor.getSelf()` gets the version to the target operation. In this case it gives either the 'torch' types version vs the `mlir` version of the tensor.

2. `getType()` vs `op.getResult().getType()` is no difference. Most of this is likely force of habit from before `getType()` worked standalone.

3. Thats about it. If something is a splat we can just proxy it through, if the rank doesn't change we can remove the operation.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
rsuderman: Correct - for some reason we were generating an illegal squeeze (it came up during the folder). I corrected this so we would no longer generate an illegal squeeze op.

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
rsuderman: Added

include/torch-mlir/Dialect/Torch/Transforms/Passes.td
```diff
@@ -235,6 +235,11 @@ def DecomposeComplexOps : Pass<"torch-decompose-complex-ops", "func::FuncOp"> {
   }];
 }
 
+def ScalarizeShapes : Pass<"torch-scalarize-shapes", "func::FuncOp"> {
+  let summary = "Takes common shape computation operations and scalarizes them to";
```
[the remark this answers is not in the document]
rsuderman: updated.

lib/Conversion/TorchToArith/TorchToArith.cpp
rsuderman: I just stole this from surrounding code. Useful but not always obvious when to use.

lib/Dialect/Torch/IR/TorchOps.cpp
rsuderman: Done.

rsuderman:
Addressed

PR #3056, opened 2024-03-26
[Torch] support decompose aten.einsum with ellipsis slicing

Xinyu302:
as title

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -171,7 +171,7 @@ rewriteEquationWithEllipsisSlicing(std::string &equation,
   std::string resultStr = equation.substr(arrowPos + 2);
 
   // split input into tokens
-  llvm::SmallVector<std::string> inputTokens;
+  SmallVector<std::string> inputTokens;
   size_t start = 0;
   size_t end = 0;
   std::set<char> usedTokens;
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -21,6 +21,7 @@
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/StringSet.h"
 #include <cstdint>
+#include <set>
 
 using namespace mlir;
 using namespace mlir::torch;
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -158,6 +159,99 @@ static SmallVector<int64_t> computeDimsOrderForMoveDim(int64_t srcDimInt,
...
+    int num = maxEllipsisRank;
+    char lastChar = 'a';
+    while (num--) {
+      while (usedTokens.find(lastChar) != usedTokens.end()) {
+        lastChar++;
+      }
+      token.push_back(lastChar);
+      lastChar++;
+    }
+    return token;
```
ramiro050, line 227: nit: this does not need to be a lambda expression. I think `token` can just be replaced with `ellipsisToken` and the lambda expression can be removed.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -158,6 +159,99 @@ static SmallVector<int64_t> computeDimsOrderForMoveDim(int64_t srcDimInt,
...
+  // tensor can be broadcasted
+  SmallVector<int64_t> ellipsisRanks;
+  int maxEllipsisRank = 0;
+  for (const auto &[token, inputRank] : llvm::zip(inputTokens, inputRanks)) {
+    int explictRank = 0;
+    for (auto c : token) {
+      if (std::isalpha(c)) {
+        usedTokens.insert(c);
+        explictRank++;
+      } else if (c == '.' || c == ' ') {
```
ramiro050, line 201: are blank spaces ever encountered in the equations?

ramiro050 (approved):
Just a couple of small comments, but other than that LGTM!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
Xinyu302: Actually, space is a legal input in einsum equations

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
Xinyu302: I have already made the modifications according to your suggestions.

ramiro050 (approved):
Thanks!

PR #3070, opened 2024-03-28
[MLIR][ONNX] Fix onnx.gather_nd implementation

Shukla-Gaurav:
The indices should be expanded before the torch.gather operation.


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -774,7 +783,7 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
               dataDims[batchDimCount + i]);
         }
 
-        // step4. Compute all the required result types here.
+        // step5. Compute all the required result types here.
```
rsuderman, line 786: If we are already relabelling the numbers on the steps lets insert a space between `step` and the number

rsuderman (approved):

PR #3071, opened 2024-03-28
[Torch Dialect] decompose all index_put-like op to aten.index_put.hacked_twin for stricter semantics

Vremold:
The is a draft PR. Currently in our project, we decompose all index_put like ops (e.g. `aten.index_put.hacked_twin`) to `aten._index_put_impl`. But we have little support for the possible `None` index in the operands in the conversions or transforms.  

Previously, I proposed a PR (#2344 ) that replaces the `None` indices in the `aten.index.Tensor` ops with manually generated indices. Maybe we can duplicate the work here. This will simplifies all backend conversions. And the place I came up to put the implementation is ops' canonicalizers. A draft code is as follows.  

I would like to hear your opinions @ramiro050 @vivekkhandelwal1 @qingyunqu . If it's ok, I will also move the implementation of index.Tensor to the TorchOps.cpp.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -7323,10 +7323,8 @@ public:
 // Torch ops related to indexing tensors, e.g., AtenIndexTensor, AtenIndexPut.
 namespace {
 
-// TODO: It might be better to use aten.view op instead of mulitple
-// aten.unsqueeze. But currently, torch-to-linalg pass has limited support for
-// view on dynamic shapes, such as [?] -> [?,1,1,1]. Using aten.view op will
-// cause relevant e2e tests fail.
+// unsqueeze is more easily optimized than a generic view, and we prefer to
+// enjoy ops with more structure than less in compositions.
 static FailureOr<Value> unsqueezeTensorAtTrailingDim(Operation *op,
                                                      PatternRewriter &rewriter,
                                                      Value input, int count) {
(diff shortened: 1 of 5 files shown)


ramiro050:
This would definitely be nice to have.

> If it's ok, I will also move the implementation of index.Tensor to the TorchOps.cpp.

That's fine with me. The transformations here are very much canonicalizations. Thanks for working on this!

Vremold:
Update at 03.31
When I‘m ready to start the modification, I find that all the index_put-like operators would be decomposed into `aten.index_put_impl`, but this operator still allows the existence of `None` index. So I change my idea and convert all the index_put-like operators to index_put.hacked_twin and reuse the relevant code of index.Tensor's decomposition pattern. Please review for me @ramiro050 @vivekkhandelwal1 . Thanks!

Vremold:
A TOSA backend test fails. The issue appears to be due to the fact that the `index_put` conversion in TOSA does not support broadcast indices. Could you take a look at it? @vivekkhandelwal1 @AmosLewis

stellaraccident:
The tosa test can be marked as fail and the tosa folks can fix it if needed for something like this (op support missing).

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -6885,65 +6830,140 @@ class DecomposeAtenTypeAsOp : public OpRewritePattern<AtenTypeAsOp> {
...
-    Value result = input;
-    while (count--) {
-      auto unsqzTensorInfo =
-          unsqueezeTensor(rewriter, op, result, /*dim=*/constMinusOne);
-      if (failed(unsqzTensorInfo)) {
-        return failure();
-      }
-
-      result = *unsqzTensorInfo;
+// TODO: It might be better to use aten.view op instead of mulitple
```
stellaraccident, line 6836: I would change this from a todo and just say that "unsqueeze is more easily optimized than a generic view, and we prefer to enjoy ops with more structure than less in compositions".

View is a very bad op to be adding in a lowering pipeline because it represents a loss of information.

Vremold:
Could you please review this PR? @stellaraccident @rsuderman

stellaraccident:
I'm on vacation for a couple more days. If Rob doesn't have a look, I will when back. He's working on some other index ops this week so has context. The rationale makes sense to me.

Vremold:
Could you review this PR? @stellaraccident @rsuderman

qingyunqu (approved):
LGTM

PR #3072, opened 2024-03-28
[MLIR][Torch] Do not convert bias tensor to element type if NoneType

tdietert:
The `convertTensorToElementType` function expects it's argument to have a valid tensor type that is not `Torch::NoneType`. This PR checks that the bias tensor is not of type `Torch::NoneType` before calling `convertTensorToElementType` on the bias tensor argument in the `matchAndRewrite` member function of the `ConvertAtenConvolutionOp` class.


vivekkhandelwal1 (approved):
LGTM

PR #3075, opened 2024-03-29
[WIP][Torch] Add decomposition of RepeatInterleaveSelfInt Op

Xinyu302:


zjgarvey:
Hi @Xinyu302. Thanks for the PR!

I'd like to review this soon. I've had a similar issue with a test that passes in nightly. My case was because I upstreamed a bug fix related to the op I was supporting, and it wouldn't get into stable until we get the next stable pytorch release. I assume someone fixed *something* in pytorch between stable and now for your case. I'll try to ask around about what the best practice should be for these kinds of test fails on stable. These kinds of issues are also good to ask in the torch-mlir channel of the LLVM discord server. I think the invite link is [https://discord.gg/xS7Z362](https://discord.gg/xS7Z362).

In the meantime, would you mind including explicit shape and dtype specifications in your decompositions? Some pass pipelines, like IREE's default pass pipeline, don't include the somewhat heavy shape and dtype calculation passes, and it would be nice to have your contribution work for those projects. See, for example, the discussion in issue #3092 .

Xinyu302:
> Hi @Xinyu302. Thanks for the PR!
> 
> I'd like to review this soon. I've had a similar issue with a test that passes in nightly. My case was because I upstreamed a bug fix related to the op I was supporting, and it wouldn't get into stable until we get the next stable pytorch release. I assume someone fixed _something_ in pytorch between stable and now for your case. I'll try to ask around about what the best practice should be for these kinds of test fails on stable. These kinds of issues are also good to ask in the torch-mlir channel of the LLVM discord server. I think the invite link is https://discord.gg/xS7Z362.
> 
> In the meantime, would you mind including explicit shape and dtype specifications in your decompositions? Some pass pipelines, like IREE's default pass pipeline, don't include the somewhat heavy shape and dtype calculation passes, and it would be nice to have your contribution work for those projects. See, for example, the discussion in issue #3092 .

It is hard to give explicit shape and dtype to ops which are generated during some very complex decomposition process. But in the PRelu case, your approach is fine.

zjgarvey:
> It is hard to give explicit shape and dtype to ops which are generated during some very complex decomposition process. But in the PRelu case, your approach is fine.

At least for some of the commonly used ops that can be difficult to get the shape for (like squeeze and unsqueeze) there are helper functions which can take care of this logic for you in [/lib/Dialect/Torch/Utils/Utils.cpp](https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Utils/Utils.cpp#L280).

My limited understanding, gained through informally discussing this topic with a few more experienced people, is that getWithLeastValueSemantics() should be avoided when possible. I'll try asking someone outside of my organization about it and see if that is a more general community sentiment or not.

zjgarvey:
I'm guessing that the old onnx export for this op is bad. See [gist](https://gist.github.com/zjgarvey/32645f9e2ef8743a15fa4db1a6e88d57) for the differing IR. I believe the onnx test also passes with torch>=2.3.0, so you could try an easy fix like adding the following to xfail_sets.py:

```python
#TODO: remove once 2.3.0 is stable
if torch_version_for_comparison() < version.parse('2.3.0.dev'):
    ONNX_XFAIL_SET.add("RepeatInterleaveSelfIntNoDimModule_basic")
 ```

**Comment by Xinyu302** [comment]
> I'm guessing that the old onnx export for this op is bad. See [gist](https://gist.github.com/zjgarvey/32645f9e2ef8743a15fa4db1a6e88d57) for the differing IR. I believe the onnx test also passes with torch>=2.3.0, so you could try an easy fix like adding the following to xfail_sets.py (towards the bottom after ONNX_XFAIL_SET):
> 
> ```python
> #TODO: remove once 2.3.0 is stable
> if torch_version_for_comparison() < version.parse('2.3.0.dev'):
>     ONNX_XFAIL_SET.add("RepeatInterleaveSelfIntNoDimModule_basic")
> ```

Thanks, I have modified xfail_sets.py.

**Comment by qingyunqu** [comment]
> > It is hard to give explicit shape and dtype to ops which are generated during some very complex decomposition process. But in the PRelu case, your approach is fine.
> 
> At least for some of the commonly used ops that can be difficult to get the shape for (like squeeze and unsqueeze) there are helper functions which can take care of this logic for you in [/lib/Dialect/Torch/Utils/Utils.cpp](https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Utils/Utils.cpp#L280).
> 
> My limited understanding, gained through informally discussing this topic with a few more experienced people, is that getWithLeastValueSemantics() should be avoided when possible. I'll try asking someone outside of my organization about it and see if that is a more general community sentiment or not.

I think it's not a good idea to migrate complex shape inference function into decomposition. It will make decomposition unreadable and hard to maintain or fix. I think it's ok when the shape inference function is simple enough.

**Comment by zjgarvey** [comment]
> > > It is hard to give explicit shape and dtype to ops which are generated during some very complex decomposition process. But in the PRelu case, your approach is fine.
> > 
> > 
> > At least for some of the commonly used ops that can be difficult to get the shape for (like squeeze and unsqueeze) there are helper functions which can take care of this logic for you in [/lib/Dialect/Torch/Utils/Utils.cpp](https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Utils/Utils.cpp#L280).
> > My limited understanding, gained through informally discussing this topic with a few more experienced people, is that getWithLeastValueSemantics() should be avoided when possible. I'll try asking someone outside of my organization about it and see if that is a more general community sentiment or not.
> 
> @zjgarvey I think it's not a good idea to migrate complex shape inference function into decomposition. It will make decomposition unreadable and hard to maintain or fix. I think it's ok when the shape inference function is simple enough. @ramiro050 What's your opinion about this?

Thanks for the input. I was briefly chatting with Ramiro about this on discord, so I'll try to reach out and see if he would be willing to discuss it in this context.

File: lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp [hunk 1]
```diff
@@ -2800,6 +2800,105 @@ class DecomposeAtenRepeatOp : public OpRewritePattern<AtenRepeatOp> {
 };
 } // namespace
 
+// decompose aten.repeat_interleave.self_int into following ops:
+// aten.flatten.using_ints, aten.unsqueeze, aten.tile, aten.reshape
+namespace {
+
+class DecomposeAtenRepeatInterleaveSelfIntOp
```
ramiro050, line 2807: I think we can simplify this a bit by doing something like

```python
def my_repeat_interleave(input, repeats, dim=None):
    dim_is_none = dim is None
    if dim is None:
        dim = input.dim() - 1

    repeat_shape = [-1] * (input.dim() + 1)
    repeat_shape[dim + 1] = repeats
    input = input.unsqueeze(dim + 1)
    tiled = input.expand(repeat_shape)

    if dim_is_none:
        return tiled.flatten()
    else:
        return torch.ops.prims.collapse(tiled, dim, dim + 1)
```

If we use this approach with the `unsqueeze` helper mentioned by @zjgarvey, the only op that needs shape calculation is the `expand` op, and the shape for it is already computed as part of the decomposition (`repeat_shape`).

We are currently in a bit of an awkward transition where different users depend on different parts of the old and new stack in torch-mlir. In general I agree that we should only include shape calculations in decompositions if it does not get too in the way of readability or if it is deeply needed by the community. In my opinion, this case with the simplifications above is simple enough to not get in the way of readability.

ramiro050 (commented):
A few comments. Thanks for working on this!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
Xinyu302: Thank you! I've implemented the changes you suggested. Could you please take another look at the code?

ramiro050 (approved):
Thanks for the quick turnaround! LGTM

PR #3076, opened 2024-03-29
[MLIR][Torch] Fix OnnxToLinalg lowering for AvgPool op

vivekkhandelwal1:

### DIFF
--- a/lib/Conversion/TorchToLinalg/Pooling.cpp
+++ b/lib/Conversion/TorchToLinalg/Pooling.cpp
@@ -114,8 +114,22 @@ static Value padInputTensor(Operation *op, ConversionPatternRewriter &rewriter,
                             SmallVectorImpl<int64_t> &paddingInts,
                             Value initValue) {
   SmallVector<int64_t> lowPaddingIncludingNC = {0, 0};
-  lowPaddingIncludingNC.append(paddingInts);
-  SmallVector<int64_t> highPaddingIncludingNC = lowPaddingIncludingNC;
+  SmallVector<int64_t> highPaddingIncludingNC = {0, 0};
+
+  unsigned selfRank = self.getType().cast<RankedTensorType>().getRank();
+  unsigned paddingIntsSize = paddingInts.size();
+
+  if (paddingIntsSize == 2 * (selfRank - 2)) {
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -114,8 +114,22 @@ static Value padInputTensor(Operation *op, ConversionPatternRewriter &rewriter,
...
+  unsigned paddingIntsSize = paddingInts.size();
+
+  if (paddingIntsSize == 2 * (selfRank - 2)) {
+    // This condition being true means that the `paddingInts` contain seperate
+    // values for low padding and high padding.
+    for (unsigned i = 0; i < paddingIntsSize / 2; i++)
+      lowPaddingIncludingNC.push_back(paddingInts[i]);
+    for (unsigned i = paddingIntsSize / 2; i < paddingIntsSize; i++)
+      highPaddingIncludingNC.push_back(paddingInts[i]);
+  } else {
```
zjgarvey, line 129: I didn't know that this op allowed specifying different high and low padding ints. 

My understanding of the torch op is that you can either specify one int or a spatial-dim-sized tuple of ints to specify padding for each spatial dim separately.

I see that the onnx op allows different high and low padding, is this onnx padding list just getting passed to the torch op as-is for the padding ints even though the original torch op doesn't actually support this functionality?

zjgarvey (approved):
Looks good to me, as long as I'm understanding the padding int "hack" of passing the onnx padding ints to the torch op, even though it isn't default behavior for the original torch op. If that is the case, it might be good to add that this is needed specifically for supporting the onnx AveragePool op, and not for the torch op itself.

lib/Conversion/TorchToLinalg/Pooling.cpp
vivekkhandelwal1: Hi @zjgarvey, you are correct that the torch op expects only one value for padding one dimension but during the padding actually 2 values are used one for low padding and one for high padding. In case of torch pooling ops both values are assumed to be same while for the onnx pooling op they can be different. So. it's valid to use the different padding values for low and high padding.

Also, I have made the changes only to the padTensor utility not the original pooling op lowering, so the pooling is still unaffected.

PR #3078, opened 2024-03-29
[build] Specify numpy < 2.0 in build.requirements.txt

tdietert:
When running the e2e tests, specifically `python -m e2e_testing.main`, many tests fail when numpy-2.X.X is installed, and an error message stating numpy-1.X.X is required. We should reflect this in the build.requirements.txt.

### DIFF
--- a/build-requirements.txt
+++ b/build-requirements.txt
@@ -1,4 +1,4 @@
-numpy
+numpy < 2.0
 pybind11
 wheel
 setuptools


tdietert:
Should this be a requirement in the `test.requirements.txt` instead?

vivekkhandelwal1:
Hi @tdietert, I don't see any failures with the current requirements. Can you please elaborate the issue in more detail? If you're not encountering the issue anymore, then please close this PR.

PR #3084, opened 2024-03-30
[Stablehlo] Add lowering of GridSampler Op

Xinyu302:
Inspired by PyTorch decompositions.py.
See https://github.com/pytorch/pytorch/blob/ec58f1f74ebcec744d2ab90ad34abd09c1018e92/torch/_decomp/decompositions.py#L3923-L4086
Only support paddingMode=0 or 1 and interpolationMode=0 or 1


qingyunqu (approved):
LGTM, Thanks for the efforts.

PR #3091, opened 2024-04-01
[torch] Improve shape inference for dynamic shapes

rsuderman:
Shapes can be processed as tensors to represent the set of dimensions. As reshapes take a list of scalars this can result in a single dynamic dimension blocking the adjacent static dimensions.

This pass attempts to de-couple tensor computations related to shapes and propagate values to better support lowering scalar tensor computations.


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1756,19 +1756,32 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                  binder.getLoc(), rewriter.getI64IntegerAttr(i));
+              Value inDim = rewriter.create<Torch::AtenSizeIntOp>(
+                  binder.getLoc(), torchIntTy, data, iv);
+              isZero = rewriter.create<Torch::PrimNumToTensorScalarOp>(
+                  binder.getLoc(), boolTy, isZero);
+              inDim = rewriter.create<Torch::PrimNumToTensorScalarOp>(
+                  binder.getLoc(), dimTy, inDim);
+              dim = rewriter.create<Torch::PrimNumToTensorScalarOp>(
+                  binder.getLoc(), dimTy, dim);
+              Value finalDim = rewriter.create<Torch::AtenWhereSelfOp>(
```
renxida, line 1778: The codebase generally uses atenIntBoolOp to convert the bool into an int 

and then use an AtenMulOp to add it.

This is more efficient than that right?

renxida: Should someone go through the other ones and try to fix them to use AtenWhereSelfOp too?

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1073,6 +1066,23 @@ class ConvertAtenViewOp : public OpConversionPattern<AtenViewOp> {
...
+    SmallVector<Value> inputSize = getTensorSizes(rewriter, loc, input);
+
+    SmallVector<Value> outputSizeInt = getTypeConvertedValues(
+        rewriter, loc, typeConverter, outputSizeTorchInt);
+    if (resultRank != (int64_t)outputSizeInt.size()) {
+      return rewriter.notifyMatchFailure(
+          op, "desired size list length mismatches with the result type rank");
+    }
+
+    for (auto pair : checkDimPairs) {
```
renxida, line 1078: https://en.cppreference.com/w/cpp/language/structured_binding

Can we use

auto [inputDim, outputDim] : checkDimPairs

to make this more readable? Or does it only work for `=` and not `:`

renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
rsuderman: This is less about efficiency and more about folding. I really want an atenSelectScalar op that takes all scalars but sadly it doesn't exist so we have to proxy with the scalar-tensor version. Multiplications with bools and additions is a way to simulate a select operation but many cpus have instructions to make ternary selection easier.

lib/Conversion/TorchToLinalg/DataMovement.cpp
rsuderman: I somehow knew someone would call me out on this. Darn C++ 21. Fixed.

PR #3093, opened 2024-04-02
[stablehlo] Fix test stablehlo e2e test suite

rsuderman:
There is an issue with stablehlo's linalg compilation. Canonicalization appears to cleanup the issues until we can determine what in mlir/stablehlo is the source of the issue.


qingyunqu (approved):
LGTM

PR #3095, opened 2024-04-02
[MLIR][TORCH] Fix Onnx.ReduceSum lowering for failing e2e tests

vivekkhandelwal1:

### DIFF
--- a/include/torch-mlir/Conversion/TorchOnnxToTorch/Utils.h
+++ b/include/torch-mlir/Conversion/TorchOnnxToTorch/Utils.h
@@ -20,6 +20,8 @@ Value createConstantIntList(OpBinder binder,
 
 Type getQTorchTypeFromTorchIntType(Type ty);
 
+bool areAllElementsDistinct(SmallVector<int64_t> array);
+
 } // namespace mlir::torch::onnx_c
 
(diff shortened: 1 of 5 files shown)


zjgarvey (approved):
Looks good to me. I also like how some of it got into a helper function. I think there are a few more Reduce ops, so I'm still curious if more could be factored out of these conversions. I'm not sure how to do it with the way the pass is structured overall, but something like:

```cpp
LogicalResult getAxesListForReduceOps(rewriter, binder.op, loc, axesList, otherArgs)
```

could do all of the work involved in getting the axesList in the Reduce ops. I'm not sure what the tradeoffs would be in an approach like that, but it might be worth considering on the next iteration.

vivekkhandelwal1:
> Looks good to me. I also like how some of it got into a helper function. I think there are a few more Reduce ops, so I'm still curious if more could be factored out of these conversions. I'm not sure how to do it with the way the pass is structured overall, but something like:
> 
> ```c++
> LogicalResult getAxesListForReduceOps(rewriter, binder.op, loc, axesList, otherArgs)
> ```
> 
> could do all of the work involved in getting the axesList in the Reduce ops. I'm not sure what the tradeoffs would be in an approach like that, but it might be worth considering on the next iteration.

Sure, I can do something like this while working on fixing the next reduce op.

PR #3097, opened 2024-04-02
[Stablehlo] add stablehlo-canonicalize-dynamism when lowering

qingyunqu:
so that many stablehlo e2e testcases could pass


Vremold (approved):
LGTM! Thanks for the effort!

Issue #3092, opened 2024-04-02
Regression in onnx.PRelu


stellaraccident:
While integrating to torch-mlir HEAD, IREE's test suite caught a regression:

https://github.com/openxla/iree/actions/runs/8515624855/job/23323678533?pr=16940

```
________ IREE compile and run: test_prelu_broadcast::cpu_llvm_sync_test ________
[gw62] linux -- Python 3.11.8 /home/runner/work/iree/iree/venv/bin/python
Error invoking iree-compile
Error code: 1
Stderr diagnostics:
model.mlir:4:10: error: failed to legalize operation 'torch.aten.mul.Tensor' that was explicitly marked illegal
    %0 = torch.operator "onnx.PRelu"(%arg0, %arg1) : (!torch.vtensor<[3,4,5],f32>, !torch.vtensor<[5],f32>) -> !torch.vtensor<[3,4,5],f32> 
         ^
model.mlir:4:10: note: see current operation: %1 = "torch.aten.mul.Tensor"(%arg0, %arg1) : (!torch.vtensor<[3,4,5],f32>, !torch.vtensor<[5],f32>) -> !torch.vtensor


Invoked with:
  cd /home/runner/work/iree/iree/SHARK-TestSuite/iree_tests/onnx/node/generated/test_prelu_broadcast && iree-compile model.mlir --iree-hal-target-backends=llvm-cpu -o model_cpu_llvm_sync_test.vmfb
```

Can someone please have a look?

zjgarvey:
Might be due to changes in [#3069](https://github.com/llvm/torch-mlir/commit/40008b025a40e8bab1f362edc1ed0ee551b5e7a3) from a few days ago. I'll take a look.

zjgarvey:
I think I have a fix, but I need to figure out how to test locally. For reference: [branch](https://github.com/zjgarvey/torch-mlir/tree/prelufix)

zjgarvey:
So, it might be the case that the pass pipeline used in the iree tests does not include shape and dtype calculations. The result type of the mul.Tensor op should have been inferred before reaching this error. 

Running the following on the mlir file in the corresponding test suite folder for prelu_broadcast:

```
torch-mlir-opt --convert-torch-onnx-to-torch --torch-decompose-complex-ops --torch-backend-to-linalg-on-tensors-backend-pipeline
```

will reproduce the error; however,

```
[...]

PR #3098, opened 2024-04-02
Adds result types to a prelu decomp

zjgarvey:
This adds explicit result types instead of relying on shape/dtype computations. 

Solves a regression issue with IREE: #3092


ScottTodd (approved):

zjgarvey:
I'm satisfied that the changes here will fix the regression issue, so feel free to merge.

Issue #3101, opened 2024-04-03
[ONNX] Unsqueeze -> Torch -> Linalg issue


zjgarvey:
The ElementwisePreluModule_basic test fails in the e2e testing with onnx config. Removing the xfail for this test and running:

```
projects/pt1/tools/e2e_test.sh -f ElementwisePreluModule_basic -v -c onnx
```

gives the error message:

```
****** Failed tests - 1 tests
    FAIL - "ElementwisePreluModule_basic"
        Compilation error: Traceback (most recent call last):
          File "/home/zjgar/code/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/framework.py", line 298, in compile_and_run_test
            compiled = config.compile(test.program_factory())
          File "/home/zjgar/code/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/configs/onnx_backend.py", line 88, in compile
            compiled_module = self.backend.compile(onnx_module)
          File "/home/zjgar/code/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/onnx_backends/linalg_on_tensors.py", line 55, in compile
            run_pipeline_with_repro_report(
          File "/home/zjgar/code/torch-mlir/build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/compiler_utils.py", line 73, in run_pipeline_with_repro_report
            raise TorchMlirCompilerError(trimmed_message) from None
        torch_mlir.compiler_utils.TorchMlirCompilerError: Lowering TorchFX IR -> Torch Backend IR failed with the following diagnostics:


        python exception: Failure while executing pass pipeline:
        error: "/Unsqueeze": unsupported by backend contract: tensor with unknown rank
        note: "/Unsqueeze": see current operation: %12 = "torch.aten.unsqueeze"(%arg1, %11) : (!torch.vtensor<[?],f32>, !torch.int) -> !torch.vtensor<*,f32>
        note: "/Unsqueeze": this is likely due to a missing transfer function in abstract_interp_lib_gen.py

        For Torch-MLIR developers, the error can be reproduced with:
        $ torch-mlir-opt -pass-pipeline='builtin.module(torch-lower-to-backend-contract{backend-legal-ops=aten.flatten.using_ints,aten.adaptive_avg_pool1d,aten.unflatten.int})' /tmp/UnnammedModule.mlir
[...]

PR #3102, opened 2024-04-03
[ONNX] Reworks onnx.Unsqueeze conversion

zjgarvey:
This rework significantly simplifies the logic of the onnx.Unsqueeze conversion and unblocks issue #3101 

Unfortunately, ElementwisePreluModule_basic is still not passing, and seems to crash my testing now because the runtime assertion: [lib/Conversion/TorchToLinalg/Utils.cpp#L301](https://github.com/llvm/torch-mlir/blob/f97cd4893f2b6c2a40bbf54f92fd3820ae4889c2/lib/Conversion/TorchToLinalg/Utils.cpp#L301).

The reason for this issue is that the dynamic dim for the weight tensor needs to either be 1 (i.e., a scalar weight), or must be equal to the second input dim (the number of channels). It is dynamic, so doesn't get treated in the broadcasting as 1, and it doesn't match the second input dim, so fails the above assert. If dynamic dims are used for this op, the e2e test should really provide an example where the weight matches the correct dim in the input (i.e., 4 for this test). Interestingly, making this change indeed will cause the onnx test to pass, but actually makes the linalg route fail since it does not actually account for the unique broadcasting of this op...


zjgarvey:
Please do not merge yet. I'll add a few lit tests after checking CI.

zjgarvey:
Okay. 9 onnx unexpected passes... siiiick.

zjgarvey:
Closing, as I've determined that #2991 contains work that makes this irrelevant.

PR #3105, opened 2024-04-03
Changed setup.py for wheel building to a more stable version

afalkenberg1:
We are using the setup.py to build a wheel as for example in: 

CMAKE_GENERATOR=Ninja python setup.py bdist_wheel --dist-dir ./torch-mlir-wheel -v
pip uninstall torch-mlir
pip install torch-mlir-wheel/torch_mlir-0.0.1-cp310-cp310-linux_x86_64.whl

I found that I needed to change setup.py to use clang, ccache and also turn off LTC by setting 
TORCH_MLIR_ENABLE_LTC_DEFAULT=False. 
This usually led to a version which I can usually build without any additional problems. 
Not sure if this might be of general interest and warrant a PR. 
Thank you

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ def check_env_flag(name: str, default=None) -> bool:
 PACKAGE_VERSION = os.environ.get("TORCH_MLIR_PYTHON_PACKAGE_VERSION") or "0.0.1"
 
 # If true, enable LTC build by default
-TORCH_MLIR_ENABLE_LTC_DEFAULT = True
+TORCH_MLIR_ENABLE_LTC_DEFAULT = False
 TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS = check_env_flag(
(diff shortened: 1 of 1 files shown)


afalkenberg1:
maybe someone can comment on this ? 
@vivekkhandelwal1 
@rsuderman

setup.py
```diff
@@ -49,7 +49,7 @@ def check_env_flag(name: str, default=None) -> bool:
 PACKAGE_VERSION = os.environ.get("TORCH_MLIR_PYTHON_PACKAGE_VERSION") or "0.0.1"
 
 # If true, enable LTC build by default
-TORCH_MLIR_ENABLE_LTC_DEFAULT = True
+TORCH_MLIR_ENABLE_LTC_DEFAULT = False
```
antoniojkim, line 52: Why did you need to disable LTC? What build failures do you encounter using `clang`?

PR #3106, opened 2024-04-03
[torch] Fix aten.squeeze lowering to use result shape

rsuderman:
Squeezes can be ambiguous without the output shape information. For instance (1, 1, 256) squeezed can be either (1, 256) or (256). We need to check the resulting shape to know what the shape should look like.

### DIFF
--- a/projects/pt1/python/torch_mlir_e2e_test/configs/torchdynamo.py
+++ b/projects/pt1/python/torch_mlir_e2e_test/configs/torchdynamo.py
@@ -137,7 +137,6 @@ def jit(
             lambda method, *inputs: method(*inputs))
         dynamo_f(lambda *inputs: model(*[x.clone() for x in inputs]),
                  *example_args)
-        print(mlir_module)
         option_string = ("{backend-legal-ops=" + ",".join(backend_legal_ops) +
                          " extra-library=" + extra_library_file_name + "}")
         print(mlir_module)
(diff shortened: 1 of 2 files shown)


zjgarvey (approved):
Cool, these are nice changes. I assume from context that reassociation[i] is the like the start and end dims of the input tensor that get collapsed to output at dim[i], correct?

rsuderman:
> Cool, these are nice changes. I assume from context that reassociation[i] is the like the start and end dims of the input tensor that get collapsed to output at dim[i], correct?

Pretty much - it basically specifies which input dims collapse into which output dims. Its a little silly because they are required to be 0 ... N-1 in order so the format is a little excessive but good enough.

projects/pt1/python/torch_mlir_e2e_test/configs/torchdynamo.py
```diff
@@ -130,16 +130,17 @@ def my_aot_autograd_backend(gm: torch.fx.GraphModule,
...
                               decompositions=_get_decomposition_table)
-
     with torch.no_grad():
         set_model_name(model.__class__.__name__)
         torch._dynamo.reset()
         dynamo_f = dynamo.optimize(my_backend, nopython=True)(
             lambda method, *inputs: method(*inputs))
         dynamo_f(lambda *inputs: model(*[x.clone() for x in inputs]),
                  *example_args)
+        print(mlir_module)
```
renxida, line 140: 1. possible debug statement left in
2. this is duplicated around something that apparently does not change the mlir module

renxida (approved):

projects/pt1/python/torch_mlir_e2e_test/configs/torchdynamo.py
rsuderman: Debug statement that leaked through. I'll remove.

Issue #3104, opened 2024-04-03
Out of tree build in conda environment


ZzEeKkAa:
Hello there I'm trying to build out-of-tree `torch-mlir` in conda environment, but failing to do so because of numerous problems.
I had to build locally `mlir-python-binding` to have `MLIRPythonExtension*` and `MLIRPythonSources*` targets populated, but then I'm getting this kind of error:
```
CMake Error at /home/yevhenii/.miniconda3/envs/mlir-dev/lib/cmake/mlir/AddMLIRPython.cmake:674 (target_link_libraries):
  Cannot find source file:

    /home/yevhenii/.miniconda3/envs/mlir-dev/src/python/MLIRPythonExtension.Core/MainModule.cpp

  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .mpp .m .M .mm .ixx .cppm
  .ccm .cxxm .c++m .h .hh .h++ .hm .hpp .hxx .in .txx .f .F .for .f77 .f90
  .f95 .f03 .hip .ispc
Call Stack (most recent call first):
  /home/yevhenii/.miniconda3/envs/mlir-dev/lib/cmake/mlir/AddMLIRPython.cmake:228 (add_mlir_python_extension)
  /home/yevhenii/.miniconda3/envs/mlir-dev/lib/cmake/mlir/AddMLIRPython.cmake:248 (_process_target)
  python/CMakeLists.txt:112 (add_mlir_python_modules)


CMake Error at /home/yevhenii/.miniconda3/envs/mlir-dev/share/cmake-3.29/Modules/FindPython/Support.cmake:4078 (add_library):
  No SOURCES given to target: TorchMLIRPythonModules.extension._mlir.dso
Call Stack (most recent call first):
  /home/yevhenii/.miniconda3/envs/mlir-dev/share/cmake-3.29/Modules/FindPython3.cmake:549 (__Python3_add_library)
  /home/yevhenii/.miniconda3/envs/mlir-dev/lib/python3.12/site-packages/pybind11/share/cmake/pybind11/pybind11NewTools.cmake:239 (python3_add_library)
  /home/yevhenii/.miniconda3/envs/mlir-dev/lib/cmake/mlir/AddMLIRPython.cmake:642 (pybind11_add_module)
  /home/yevhenii/.miniconda3/envs/mlir-dev/lib/cmake/mlir/AddMLIRPython.cmake:228 (add_mlir_python_extension)
[...]

PR #3107, opened 2024-04-04
Add missing dependency to TorchMLIRRefBackend target

ZzEeKkAa:
Discovered in https://github.com/llvm/torch-mlir/issues/3104
Most likely when building with stablehlo, while waiting for it missing dependency was generated to location shared with another dependency.

### DIFF
--- a/lib/RefBackend/CMakeLists.txt
+++ b/lib/RefBackend/CMakeLists.txt
@@ -7,6 +7,7 @@ add_mlir_library(TorchMLIRRefBackend
   DEPENDS
   MLIRTorchTypesIncGen
   TorchMLIRRefBackendPassIncGen
+  MLIRTorchConversionOpsIncGen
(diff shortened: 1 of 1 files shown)


vivekkhandelwal1:
@qingyunqu Can you please confirm if this patch is actually needed while building the project for StableHlo?

marbre:
It is correct that `RefBackend.cpp` includes `torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h` which itself includes `torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h.inc`. The latter is created via `MLIRTorchConversionOpsIncGen` but we should rather add a dependency to `TorchMLIRTorchConversionDialect` as this is where the functions are defined.

qingyunqu (approved):
LGTM

PR #3109, opened 2024-04-06
[Stablehlo] add stablehlo-aggressive-simplification in e2e test

qingyunqu:
* so that more stablehlo e2e testcases would pass.


Vremold (approved):
LGTM! Thanks for the effort!

Issue #3120, opened 2024-04-09
LTC build error after PT bump: multiple definition of `torch::lazy::compute_shape_arange`


sjain-stanford:
Just noticed that the nightly releases of torch-mlir have been broken for 5 days (https://github.com/llvm/torch-mlir-release/actions/runs/8537617646/job/23388584293) starting April 2nd. Seems correlated in timing to the PyTorch bump PR: https://github.com/llvm/torch-mlir/pull/3094, and I see this being disabled from CI as well.

Linker error: `multiple definition of torch::lazy::compute_shape_arange`

```
.
.
  [3874/3917] Linking CXX shared library tools/torch-mlir/python_packages/torch_mlir/torch_mlir/_mlir_libs/lib_torch_mlir_ltc.so
  FAILED: tools/torch-mlir/python_packages/torch_mlir/torch_mlir/_mlir_libs/lib_torch_mlir_ltc.so
  : && /opt/rh/devtoolset-10/root/usr/bin/c++ -fPIC -fPIC -fno-semantic-interposition -fvisibility-inlines-hidden -Werror=date-time .
.
.
  shape_inference.cpp:(.text._ZN5torch4lazy20compute_shape_arangeERKN3c106ScalarES4_S4_St8optionalINS1_10ScalarTypeEES5_INS1_6LayoutEES5_INS1_6DeviceEES5_IbE+0x0): multiple definition of `torch::lazy::compute_shape_arange(c10::Scalar const&, c10::Scalar const&, c10::Scalar const&, std::optional<c10::ScalarType>, std::optional<c10::Layout>, std::optional<c10::Device>, std::optional<bool>)'; tools/torch-mlir/projects/ltc/csrc/base_lazy_backend/CMakeFiles/torch_mlir_ltc_backend.dir/generated/shape_inference.cpp.o:shape_inference.cpp:(.text._ZN5torch4lazy20compute_shape_arangeERKN3c106ScalarES4_S4_St8optionalINS1_10ScalarTypeEES5_INS1_6LayoutEES5_INS1_6DeviceEES5_IbE+0x0): first defined here
  collect2: error: ld returned 1 exit status
```

stellaraccident:
I think we've already disabled that in the regular CI. Might need to be disabled for releases.

PR #3121, opened 2024-04-09
Disable LTC from release builds to bypass linker issue

sjain-stanford:
Workaround for https://github.com/llvm/torch-mlir/issues/3120. This should bring prebuilt releases back to green: https://github.com/llvm/torch-mlir-release/actions.


sjain-stanford:
At the moment the only user of `build_tools/python_deploy/build_linux_packages.sh` is the `torch-mlir-release` (which is broken atm), so I'm taking the call to land this without a stamp. @antoniojkim please feel free to revert the change once you've identified a fix.

PR #3122, opened 2024-04-09
Adds Some uint8 Quantization Fixes

zjgarvey:
1. Changes the linalg lowering for dequantization ops to always sign cast to float to prevent misrepresenting uint32 overflow on subtraction with zero point. 
2. Adds a basic quantized model test which only quantizes and dequantizes and now passes with these changes in linalg and onnx configs.
3. Changes the aten.mm lowering to allow mismatched quantized types. 
4. If a quantized matmul arg is uint8, we shift by 128 to faithfully represent the quantization as a signed i8 quantization. This worked fine in the AtenMmOp lowering, but I'd be happy to move it to a rewrite in FuseQuantizedOps.cpp instead if that seems more appropriate. 

With the changes 3 and 4, the QuantizedMLP_basic and QuantizedSingleLayer_basic e2e tests now passes with the onnx config.


zjgarvey:
I need to make a few changes after checking more carefully what extsi/extui do. It seems like the old examples did not run into large enough uint8 values to encounter issues with sign-extending. Instead, should unsign extend, and just always cast the i32 to float as a signed int32. 

For matmul, I'll need to make some slightly more significant fixes to avoid the same potential issue with extsi.

zjgarvey:
Some context that may be helpful for others:

1. Integers are bit vectors (vectors of 0's and 1's).
2. The top bits correspond to the highest powers of 2 (B.E.)
3. If treated as a signed integer, the two's complement rule applies (essentially, -x + x should give all 0's). E.g., a bit vector of all ones == -1. Additionally, all integers with top digit == 1 are considered negative. For example, for si3: 1,0,0 = -4 instead of 4. 
4. arith::extsi takes the top digit from the input and repeats it for the top (outWidth - inWidth) many digits of the output. For example: -2 (i3) = 1,1,0 -> 1,1,1,1,0 = -2 (i5). Also, treating 1,1,0 as a signed i3, this would be "6" as an i3 but would cast to -2. Acting on unsigned ints with arith::extsi will essentially "consider input as equivalent same-width signed integer first, then make a larger bit integer of an equivalent value".
5. arith::extui just puts zero for the first (outWidth - inWidth) many digits. This essentially does the opposite: "consider input as equivalent same-width unsigned integer first, then make a larger bit integer of an equivalent value".
6. arith::sitfp and uitofp are similar in "consider input as si/ui first, then get closest fp representation of that value".
7. given a set of integers S, ordered from smallest to largest *considered as unsigned ints* of bitwidth W, then subtracting or adding 2^{W-1} from each value of S will result in a new set S' which is now ordered from smallest to largest *considered as signed ints* of bitwidth W. We use this fact to convert uint8 quantization to an equivalent int8 quantization to pass to linalg::quantized_matmul (which uses arith::extsi to extend to the accumulator type).

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -152,6 +153,35 @@ class ConvertAtenMmOp : public OpConversionPattern<AtenMmOp> {
...
+        minSIValue = rewriter.create<arith::ConstantIntOp>(loc, minSI, numBits);
+        arg = torch_to_linalg::createElementwiseLinalgGeneric(
+            rewriter, loc, ValueRange{arg},
+            arg.getType().cast<TensorType>().getElementType(),
+            [&](OpBuilder &b, Location loc, ValueRange payloadArgs) {
+              Value result = rewriter.create<arith::AddIOp>(loc, payloadArgs[0],
+                                                            minSIValue);
+              b.create<linalg::YieldOp>(loc, result);
+            });
+        return;
```
rsuderman, line 176: This return is not necessary.

rsuderman (approved):

PR #3124, opened 2024-04-09
Disable LTC from e2eshark builds to bypass linker issue

saienduri:


saienduri:
No longer needed with LTC fix

PR #3125, opened 2024-04-09
[Stablehlo] lowering aten.view to shape.num_elements + stablehlo.comp…

qingyunqu:
…ute_reshape_shape

as that `aten.view` support at most one `-1` in dim list. The original calculation of `numel` is wrong when there is a `-1` in dim list.


Vremold (approved):
LGTM.

PR #3132, opened 2024-04-10
[Torch] Emit and decompose prims.iota op

penguin-wwy:


penguin-wwy:
@stellaraccident This Op will appear in the fx graph due to the torch._dynamo.inductor's decomposition.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -4789,6 +4789,35 @@ class DecomposeAtenArangeStartOp : public OpRewritePattern<AtenArangeStartOp> {
 };
 } // namespace
 
+namespace {
+// The `prims.iota` op is converted to `aten.arange.startStep` op.
```
stellaraccident, line 4793: I expect that if we hadn't started with the aten op, we would have inverted this and decomposed to prims.iota. but I can't think of a good reason to go back and change that now.

stellaraccident (approved):

PR #3133, opened 2024-04-10
[torch] Fix `onnx.MaxPool` lowering

vivekkhandelwal1:
This commit fixes the onnx.MaxPool op lowering which was lacking the indices result support.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -466,15 +466,14 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
           return rewriter.notifyMatchFailure(
               binder.op, "unsupported conversion: auto_pad != NOTSET");
 
-        Torch::ValueTensorType resultType;
+        Torch::ValueTensorType resultTypeOut;
         Value operand;
-        bool ceilMode;
-        int64_t storageOrder;
+        int64_t ceilMode, storageOrder;
         // TODO: Add support for indices output and storage_order
         if (binder.tensorOperand(operand) ||
-            binder.s64BoolAttr(ceilMode, "ceil_mode", false) ||
+            binder.s64IntegerAttr(ceilMode, "ceil_mode", 0) ||
             binder.s64IntegerAttr(storageOrder, "storage_order", 0) ||
-            binder.tensorResultType(resultType))
+            binder.tensorResultTypeAtIndex(resultTypeOut, 0))
           return rewriter.notifyMatchFailure(
               binder.op,
               "operand/ceil_mode/storage_order/resultType bind failure");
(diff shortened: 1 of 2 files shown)


projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -2745,6 +2740,7 @@
 ONNX_CRASHING_SET = { 
     "FakeQuantizePerTensorAffineModule_basic",
     "FakeQuantizePerTensorAffineDynamicShapeModule_basic",
-
     "ViewDynamicExpandCollapseWithParallelUnknownDimModule_basic",
+    "MaxPool2dWithIndicesAllNegativeValuesModule_basic",
+    "MaxPool2dWithIndicesNonDefaultPaddingModule_basic",
```
zjgarvey, line 2745: Why do these crash?

Does the conversion only work in the static case?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -484,12 +483,12 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
           Value shuffledPaddingList =
               createConstantIntList(binder, rewriter, padding);
           Value zero;
-          if (resultType.getDtype().isa<FloatType>()) {
+          if (resultTypeOut.getDtype().isa<FloatType>()) {
             zero = rewriter.create<Torch::ConstantFloatOp>(
                 binder.getLoc(), rewriter.getType<Torch::FloatType>(),
```
zjgarvey, line 488: Why does the ConstantFloatOp builder require Torch::FloatType passed, but ConstantIntOp doesn't require Torch::IntType?

zjgarvey (commented):
I'd really like to know what is causing the crashing. Does the torch-to-linalg lowering have some runtime assert that fails?

renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -403,15 +403,14 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
           return rewriter.notifyMatchFailure(
               binder.op, "unsupported conversion: auto_pad != NOTSET");
 
-        Torch::ValueTensorType resultType;
+        Torch::ValueTensorType resultTypeOut;
```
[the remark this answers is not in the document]
vivekkhandelwal1: Because there are two result types for the MaxPool op. Result type for the output tensor and for the indices tensor.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
vivekkhandelwal1: Actually, I haven't added that code, it was added by @rsuderman. As per my understanding, `Torch::FloatType` is also not required.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -515,17 +514,39 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+                cstCeilMode);
+            return success();
+          }
+          if (rank == 5) {
+            rewriter.replaceOpWithNewOp<Torch::AtenMaxPool3dWithIndicesOp>(
+                binder.op, resultTypeOut, resultTypeIndices, operand,
+                kernelSizeList, stridesList, paddingList, dilationsList,
+                cstCeilMode);
+            return success();
+          }
```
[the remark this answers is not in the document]
vivekkhandelwal1: Sorry, I didn't get this comment. If I return the failure after this if then also I have to check for some cases and there might exist a case where I have to return the failure.

zjgarvey (approved):

vivekkhandelwal1:
Hi @rsuderman, can you please review this PR?

rsuderman (approved):

PR #3134, opened 2024-04-10
[ONNX] Add support for reduceprod with axes given

AmosLewis:
Rewrite the reduceprod to derive the chosen axes by the input data size and result data size.
Add keepdims = 0 new lit test.
Pass keepdims = 0/1 new Shark-TestSuite e2e tests. https://github.com/nod-ai/SHARK-TestSuite/pull/158

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -1895,8 +1895,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
               binder.op,
               "Expected the input and result type to have known sizes");
 
-        auto dataSize = dataTy.getSizes();
-        int64_t rank = dataSize.size();
+        int64_t rank = dataTy.getSizes().size();
         SmallVector<Value> axesList;
 
         // Previous version of the operation had the axes as an attribute:
@@ -1919,9 +1918,9 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             return success();
           }
 
-          // Derive the static axes from the data type and final result type.
-          // This algorithm won't work if the input shape has dupliacte dim:
-          // [2,2,2] -> [2,2], axes=1/2/3
+          // Try to derive the static axes from the data type and final result
+          // type. This algorithm won't work if the input shape has dupliacte
+          // dim: [2,2,2] -> [2,2], axes=1/2/3
           if (areAllElementsDistinct(inputShape)) {
             int resultRank = resultShape.size();
             if (!keepDims) {
@@ -1953,11 +1952,28 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
               }
             }
           }
+
+          // Unable to derive the static axes by the result shape and input
+          // shape. Extract the axes values from the axes operand. This really
+          // shouldn't happen but it helps pass weird tests.
           if (axesList.empty()) {
-            return rewriter.notifyMatchFailure(
-                binder.op,
-                "Unable to derive the static axes by the result shape and "
-                "input shape. Please provide the axes as attributes.");
+            Torch::BaseTensorType axesType =
+                axes.getType().cast<Torch::BaseTensorType>();
+            auto sizes = axesType.getSizes();
+            for (int i = 0; i < sizes[0]; i++) {
+              Value selectIndex = rewriter.create<Torch::ConstantIntOp>(
+                  binder.getLoc(), rewriter.getI64IntegerAttr(i));
+              Value zero = rewriter.create<Torch::ConstantIntOp>(
+                  binder.getLoc(), rewriter.getI64IntegerAttr(0));
+              Value extract = rewriter.create<Torch::AtenSelectIntOp>(
+                  binder.getLoc(),
+                  axesType.getWithSizesAndDtype(llvm::SmallVector<int64_t>{1},
+                                                axesType.getOptionalDtype()),
+                  axes, zero, selectIndex);
+              Value dim = rewriter.create<Torch::AtenItemOp>(
+                  binder.getLoc(), torchIntTy, extract);
+              axesList.push_back(dim);
+            }
           }
         }
 
@@ -1977,7 +1993,6 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             Value dimValue = rewriter.create<Torch::ConstantIntOp>(
                 binder.getLoc(), rewriter.getI64IntegerAttr(i));
             axesList.push_back(dimValue);
-            axesAttr.push_back(i);
           }
         }
 
@@ -1987,7 +2002,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         Value trueVal =
             rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), true);
         Value noneVal = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
-        SmallVector<int64_t> intermediateShape(dataSize);
+        SmallVector<int64_t> intermediateShape(rank, Torch::kUnknownSize);
         Value dataReduceProd = data;
         for (int i = 0, numAxes = axesList.size(); i < numAxes; i++) {
           auto axisValue = axesList[i];
@@ -2000,7 +2015,6 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             rewriter.replaceOp(binder.op, dataReduceProd);
             return success();
           }
-          intermediateShape[axesAttr[i]] = 1;
           Type resultTyReduceProd = dataTy.getWithSizesAndDtype(
               ArrayRef(intermediateShape), dataTy.getOptionalDtype());
           dataReduceProd = rewriter.create<Torch::AtenProdDimIntOp>(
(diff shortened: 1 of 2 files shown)


AmosLewis:
Pass the new e2e tests [ReduceProdKeepdims1/model.py](https://github.com/nod-ai/SHARK-TestSuite/pull/158/files#diff-eaace9eb934652c3b80063e0a565f794afaebb32758bbc4935a414973510a6b6) 
```
module {
  func.func @reduceprodgraph(%arg0: !torch.vtensor<[2,3,4],f32>, %arg1: !torch.vtensor<[1],si64>) -> !torch.vtensor<[2,1,4],f32> attributes {torch.onnx_meta.ir_version = 9 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "", torch.onnx_meta.producer_version = ""} {
    %none = torch.constant.none
    %0 = torch.operator "onnx.ReduceProd"(%arg0, %arg1) {torch.onnx.keepdims = 1 : si64, torch.onnx.noop_with_empty_axes = 0 : si64} : (!torch.vtensor<[2,3,4],f32>, !torch.vtensor<[1],si64>) -> !torch.vtensor<[2,1,4],f32> 
    return %0 : !torch.vtensor<[2,1,4],f32>
  }
}
```
->
```
#map = affine_map<(d0, d1, d2) -> (d0, d1, d2)>
#map1 = affine_map<(d0, d1, d2) -> (d0, 0, d2)>
module {
  ml_program.global private mutable @global_seed(dense<0> : tensor<i64>) : tensor<i64>
  func.func @reduceprodgraph(%arg0: tensor<2x3x4xf32>, %arg1: tensor<1xi64>) -> tensor<2x1x4xf32> {
    %cst = arith.constant 1.000000e+00 : f32
    %0 = tensor.empty() : tensor<2x1x4xf32>
    %1 = linalg.fill ins(%cst : f32) outs(%0 : tensor<2x1x4xf32>) -> tensor<2x1x4xf32>
    %2 = linalg.generic {indexing_maps = [#map, #map1], iterator_types = ["parallel", "reduction", "parallel"]} ins(%arg0 : tensor<2x3x4xf32>) outs(%1 : tensor<2x1x4xf32>) {
    ^bb0(%in: f32, %out: f32):
      %3 = arith.mulf %in, %out : f32
      linalg.yield %3 : f32
    } -> tensor<2x1x4xf32>
    return %2 : tensor<2x1x4xf32>
  }
}
```

AmosLewis:
Get 2nd e2e test pass [e2eshark/onnx/operators/ReduceProdKeepdims0/model.py](https://github.com/nod-ai/SHARK-TestSuite/pull/158/files#diff-5cf40751ac42b9b31b110d99e7a0545607862bccf53feb3aaa9c57154a104978)

AmosLewis:
Rewrite 2 old lit tests and add a new keepdims0 lit tests.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1872,26 +1870,35 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
-                axesType.getWithSizesAndDtype(llvm::SmallVector<int64_t>{1},
-                                              axesType.getOptionalDtype()),
-                axes, zero, selectIndex);
-            Value dim = rewriter.create<Torch::AtenItemOp>(binder.getLoc(),
-                                                           torchIntTy, extract);
-            axesList.push_back(dim);
+        // Derive the chosen axes from the data type and final result type
+        // Fix --convert-torch-to-linalg bug:  `dim` argument must be a
+        // constant int list or None
+        auto resultTypeSizes = resultType.getSizes();
```
rsuderman, line 1876: It is possible there will be more information you can derive by folding on the inputs values. We should not assume the values do not matter.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1872,26 +1870,35 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        int resultRank = resultTypeSizes.size();
+        if (!keepDims) {
+          // Handle the keepDimsBool == False case:
+          // [2,3,4] -> [2,4] when axes=1
+          int j = 0;
+          for (int i = 0; i < rank; i++) {
+            if (resultRank && j < resultRank &&
+                dataSize[i] == resultTypeSizes[j]) {
```
rsuderman, line 1884: How could you identify which dimension the reduction is done over for a `[4, 4, 4] -> [4]`? You can't without looking at the axis. You will need to use that information somehow.

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
AmosLewis: No. I cannot identify this cases. Let me think about it.

AmosLewis: Then this implementation is not able to fix all the cases. Either we add a special check for this kind of case. 
Or we back to the previous implementation. But the previous implementation, after we select the dynamic axis, it will fail when lowering to linalg `dim argument must be a constant int list or None`. I feel kind of stuck what to do next.

AmosLewis: @rsuderman Could you provide some guidance on it?

rsuderman: I honestly wouldn't worry about cases where we can't fold the dim value.  Its a bit nonsensical if we don't know what the value is. We could do a shape determined value, i.e. infer from result types when the value is not foldable, but I don't think there is much value in the added complexity.

AmosLewis: ` Its a bit nonsensical if we don't know what the value is.` Do you mean the operand1/dimtensor is not useless? This is what I am trying to do in this patch to derive the shape from the result types. But in the first comment, you said `We should not assume the values do not matter.`

AmosLewis: Just added check for input shape dims are distinct when dynamic axes given

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1910,26 +1908,72 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
-                axesType.getWithSizesAndDtype(llvm::SmallVector<int64_t>{1},
-                                              axesType.getOptionalDtype()),
-                axes, zero, selectIndex);
-            Value dim = rewriter.create<Torch::AtenItemOp>(binder.getLoc(),
-                                                           torchIntTy, extract);
-            axesList.push_back(dim);
+          SmallVector<int64_t> inputShape{dataTy.getSizes()};
+          SmallVector<int64_t> resultShape{resultType.getSizes()};
+          // if the shapes are equal, none of the dims is reduced
+          if (llvm::equal(inputShape, resultShape)) {
```
rsuderman, line 1915: We cannot guarantee this is true. Dynamic shapes with poor shape inference could result in multiple dynamic dimensions with currently unknown folding.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1952,20 +1996,6 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           }
         }
 
-        // Handle negative axis:
-        Value rankVal = rewriter.create<Torch::AtenDimOp>(binder.getLoc(),
-                                                          torchIntTy, data);
-        for (Value &axes : axesList) {
```
rsuderman, line 1958: You are removing support for negative dimensions specified in the axes list. This is wrong as we cannot guarantee when extracting it falls in the dimensions range.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1910,26 +1908,72 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          if (llvm::equal(inputShape, resultShape)) {
+            // simply fill in the op and return
+            rewriter.replaceOp(binder.op, data);
+            return success();
+          }
+
+          // Try to derive the static axes from the data type and final result
+          // type. This algorithm won't work if the input shape has dupliacte
+          // dim: [2,2,2] -> [2,2], axes=1/2/3
+          if (areAllElementsDistinct(inputShape)) {
```
rsuderman, line 1924: You still are making assumptions that shapes are well known. It is possible have dynamic dimensions resulting in mismatchings due to weirdness in shape inference.

rsuderman (changes requested):
Fixing the failure to lower the cases we described should not be handled at the `onnx-to-torch` level and instead is a `torch-to-linalg` issue, namely related to folders. Attempting to introduce more complexity at the `onnx-to-torch` level is avoiding the issue rather than fixing it.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
AmosLewis: The negative dim will be handled in torch-to-linalg.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
AmosLewis: Yes. This is to pass some static know test case.

AmosLewis:
Drop the patch for now.  If there are some urgent model required this support, we will try to fix it in torch-linalg pass.

PR #3138, opened 2024-04-11
[Stablehlo] fix aten.arange's lowering to stablehlo

qingyunqu:
* promote to f64 to do division, avoid division on i64 (floor div)
* refactor torch-to-stablehlo-pipeline


Vremold (approved):
LGTM. Thanks for the effort!

PR #3141, opened 2024-04-11
[FxImporter] Fix torch.fill value type

penguin-wwy:
'torch.aten.fill.Scalar' op operand 1 must be Any Python numeric type compatible with being the scalar type of a tensor, but got '!torch.bool'


penguin-wwy:
PR on #3163

PR #3146, opened 2024-04-11
[MLIR][TORCH] Add OnnxToTorch lowering for ReduceL1 Op

vinayakdsci:
Adds OnnxToTorch Lowering for the ReduceL1 op.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -39,6 +39,127 @@ Value getItemOp(OpBinder binder, ConversionPatternRewriter &rewriter,
   return rewriter.create<Torch::AtenItemOp>(binder.getLoc(),
                                             rewriter.getType<T>(), ofItem);
 }
+
+// originalOperand is explicitly passed to handle the case where ReduceSum
+// was not the first operation performed on the operand, the original value of
+// which we might need for noop_with_empty_axes handling.
+LogicalResult reducedSumImpl(OpBinder binder,
+                             ConversionPatternRewriter &rewriter, Value &data,
+                             Value &originalOperand,
+                             Torch::ValueTensorType resultType,
+                             Value *storeResult, int64_t keepDims,
+                             int64_t noop_with_empty_axes,
+                             bool isIntermediateOp) {
+
+  SmallVector<Value> axesList;
+  Value axesVal;
+  if (!binder.tensorOperandAtIndex(axesVal, 1)) {
+    auto inputType = data.getType().dyn_cast<Torch::ValueTensorType>();
+    if (!inputType.hasSizes() || !resultType.hasSizes()) {
+      return rewriter.notifyMatchFailure(
+          binder.op, "unimplemented: expected input and result to have shapes");
+    }
+
+    if (inputType.areAllSizesKnown() && resultType.areAllSizesKnown()) {
+      SmallVector<int64_t> inputShape{inputType.getSizes()};
+      SmallVector<int64_t> resultShape{resultType.getSizes()};
+      // if the shapes are equal, none of the dims is reduced
+      if (llvm::equal(inputShape, resultShape)) {
+        // simply fill in the op and return
+        rewriter.replaceOp(binder.op, data);
+        return success();
+      }
+      if (areAllElementsDistinct(inputShape)) {
+        // The check for the input shape elements to be distinct is added
+        // for the cases like:
+        // Input: [3, 2, 2] -> Output: [3, 2]
+        // For the above case, from the input and output shape it can't be
+        // inferred whether the dim:1 is reduced or dim:2. To avoid these
+        // type of cases, the check has been placed.
+        SmallVector<int64_t> reduceDims;
+        unsigned resultShapeCounter = 0;
+        for (unsigned i = 0; i < inputShape.size(); i++) {
+          if (resultShapeCounter < resultShape.size() &&
+              inputShape[i] == resultShape[resultShapeCounter]) {
+            resultShapeCounter++;
+          } else {
+            reduceDims.push_back(i);
+            if (resultShapeCounter < resultShape.size() &&
+                resultShape[resultShapeCounter] == 1)
+              resultShapeCounter++;
+          }
+        }
+        for (auto i : reduceDims) {
+          axesList.push_back(rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(i)));
+        }
+      }
+    }
+    if (axesList.empty()) {
+      Torch::BaseTensorType axesType =
+          axesVal.getType().cast<Torch::BaseTensorType>();
+      auto axesTy = dyn_cast<Torch::ValueTensorType>(axesVal.getType());
+      auto axesShape = axesTy.getSizes();
+      if (axesShape.size() != 1 || axesShape[0] == Torch::kUnknownSize)
+        return failure();
+
+      Value zero = rewriter.create<Torch::ConstantIntOp>(
+          binder.getLoc(), rewriter.getType<Torch::IntType>(),
+          rewriter.getI64IntegerAttr(0));
+      SmallVector<int64_t> selectSizes{1};
+      auto selType = rewriter.getType<Torch::ValueTensorType>(
+          selectSizes, axesType.getOptionalDtype());
+      int64_t numAxes = axesShape[0];
+      for (int64_t i = 0; i < numAxes; ++i) {
+        Value iv = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getI64IntegerAttr(i));
+        Value extract = rewriter.create<Torch::AtenSelectIntOp>(
+            binder.getLoc(), selType, axesVal, zero, iv);
+        Value dim = rewriter.create<Torch::AtenItemOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(), extract);
+        axesList.push_back(dim);
+      }
+    }
+  }
+
+  SmallVector<int64_t> axesInts;
+  if (!binder.s64IntegerArrayAttr(axesInts, "axes", {})) {
+    for (int64_t i = 0, s = axesInts.size(); i < s; ++i) {
+      Value iv = rewriter.create<Torch::ConstantIntOp>(
+          binder.getLoc(), rewriter.getType<Torch::IntType>(),
+          rewriter.getI64IntegerAttr(axesInts[i]));
+      axesList.push_back(iv);
+    }
+  }
+
+  // Do not include absolute value in the noop
+  if (axesList.empty() && noop_with_empty_axes) {
+    rewriter.replaceOp(binder.op, originalOperand);
+    return success();
+  }
+
+  Value dimValueList = rewriter.create<Torch::PrimListConstructOp>(
+      binder.getLoc(),
+      Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+      axesList);
+  Value keepDimBool =
+      rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), keepDims);
+  Value dType = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+  // If we are using the ReducedSum as an intermediate op to be passed into
+  // another operation, we might not want to replace the Op. So we create a new
+  // Op and store the result in a variable.
+  if (!isIntermediateOp) {
+    rewriter.replaceOpWithNewOp<Torch::AtenSumDimIntListOp>(
+        binder.op, resultType, data, dimValueList, keepDimBool,
+        /*dtype=*/dType);
+  } else {
+    *storeResult = rewriter.create<Torch::AtenSumDimIntListOp>(
+        binder.getLoc(), resultType, data, dimValueList, keepDimBool,
+        /*dtype=*/dType);
+  }
+  return success();
+}
 } // namespace
 
 void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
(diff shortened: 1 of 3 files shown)


vinayakdsci:
cc @vivekkhandelwal1 @renxida @rsuderman @zjgarvey

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -758,6 +758,124 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          for (int64_t i = 0, s = axesInts.size(); i < s; ++i) {
+            Value iv = rewriter.create<Torch::ConstantIntOp>(
+                binder.getLoc(), rewriter.getType<Torch::IntType>(),
+                rewriter.getI64IntegerAttr(axesInts[i]));
+            axesList.push_back(iv);
+          }
+        }
+
+        if (axesList.empty() && noop_with_empty_axes) {
+          rewriter.replaceOp(binder.op, data);
```
zjgarvey, line 863: This should be changed to use "operand" instead of "data", since you do not want to include the absolute value op in the noop case.

zjgarvey (changes requested):
Looks pretty good, but should edit the noop case to not include absolute value.

Do you know what is causing the other e2e tests for this op to fail?

vinayakdsci:
> Looks pretty good, but should edit the noop case to not include absolute value.

Thanks! I will fix this and re-push.

> Do you know what is causing the other e2e tests for this op to fail?

The other e2e test `ReduceL1NormWithDTypeModule_basic` fails on my build with `Found value Double but Expected Float`. I tried putting in Dtype conversions to check, but the test fails regardless. Do you have any suspicion on where the test could be failing? If I change dtype from `float64` to `float32` in `build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/test_suite/reduction.py` where the traceback points, I am able to get the test to pass.

vinayakdsci:
Is it possible that there could be a mistake in `projects/pt1/python/torch_mlir_e2e_test/onnx_backends/reductions.py`?

zjgarvey:
> Is it possible that there could be a mistake in `projects/pt1/python/torch_mlir_e2e_test/onnx_backends/reductions.py`?

It is certainly possible. I'll take a look. We should also have someone like @rsuderman or @renxida  start the workflows since I don't think I have permissions to start those. Feel free to resolve the earlier conversations, since you made appropriate changes.

zjgarvey (approved):
Hi, I think it is actually an issue with either the onnx import or onnx export not respecting the other dtypes. So I wouldn't worry about the other tests with onnx config. 

So I'd say this looks good to me. I won't be able to merge the PR, so please reach out to someone with merge permissions. @vivekkhandelwal1 is very familiar with these reduce ops, so he would be a good one to take a look.

vinayakdsci:
@zjgarvey Thanks for the review! I am waiting for a review from @vivekkhandelwal1 too, in the meantime.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -758,6 +758,126 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            binder.tensorResultType(resultType) ||
+            binder.s64IntegerAttr(keepDims, "keepdims", 1) ||
+            binder.s64IntegerAttr(noop_with_empty_axes, "noop_with_empty_axes",
+                                  0))
+          return failure();
+
+        // create an intermediate dummy result type with same shape as operand
+        auto operandType = operand.getType().dyn_cast<Torch::ValueTensorType>();
+        Value data = rewriter.create<Torch::AtenAbsOp>(binder.getLoc(),
+                                                       operandType, operand);
```
vivekkhandelwal1, line 776: I think if you just pass operand.getType(), it should suffice

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -758,6 +758,126 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+
+        // The code till the end is identical to ReduceSum Conversion Op below,
+        // as ReduceL1 is equivalent to ReduceSum(Abs(Operand)).
```
vivekkhandelwal1, line 779: If this is the case then the better approach is to extract this code and put it inside a function so that it can be reused for both these lowering and possible future lowerings.

projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -2035,7 +2035,6 @@
     "RandModule_basic",
 
     # Failure - onnx_lowering: onnx.ReduceL1
-    "ReduceL1NormModule_basic",
     "ReduceL1NormWithDTypeModule_basic",
     "ReduceL1NormComplexModule_basic",
```
vivekkhandelwal1, line 2039: You should try to fix the remaining tests as well.

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
```diff
@@ -863,6 +863,93 @@ func.func @test_reduce_max_attr(%arg0: !torch.vtensor<[4,2],i1>) -> !torch.vtens
...
+  // CHECK: %[[ABS:.+]] = torch.aten.abs %arg0 : !torch.vtensor<[3,2,2],f32> -> !torch.vtensor<[3,2,2],f32>
+  // CHECK: %[[INT0:.+]] = torch.constant.int 0
+  // CHECK: %[[INT0_0:.+]] = torch.constant.int 0
+  // CHECK: %[[SELECT:.+]] = torch.aten.select.int %arg1, %[[INT0]], %[[INT0_0]] : !torch.vtensor<[1],si64>, !torch.int, !torch.int -> !torch.vtensor<[1],si64>
+  // CHECK: %[[ITEM:.+]] = torch.aten.item %[[SELECT]] : !torch.vtensor<[1],si64> -> !torch.int
+  // CHECK: %[[DIMS:.+]] = torch.prim.ListConstruct %[[ITEM]] : (!torch.int) -> !torch.list<int>
+  // CHECK: %[[TRUE:.+]] = torch.constant.bool true
+  // CHECK: %[[NONE:.+]] = torch.constant.none
+  // CHECK: %[[SUM:.+]] = torch.aten.sum.dim_IntList %[[ABS]], %[[DIMS]], %[[TRUE]], %[[NONE]] : !torch.vtensor<[3,2,2],f32>, !torch.list<int>, !torch.bool, !torch.none -> !torch.vtensor<[3,2,1],f32>
+  // CHECK: return %[[SUM]] : !torch.vtensor<[3,2,1],f32>
```
vivekkhandelwal1, line 930: You can remove this test since it's similar to the test above.

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
```diff
@@ -863,6 +863,93 @@ func.func @test_reduce_max_attr(%arg0: !torch.vtensor<[4,2],i1>) -> !torch.vtens
...
+  // CHECK: %[[ABS:.+]] = torch.aten.abs %arg0 : !torch.vtensor<[2,0,4],f32> -> !torch.vtensor<[2,0,4],f32>
+  // CHECK: %[[INT1:.+]] = torch.constant.int 1
+  // CHECK: %[[DIMS:.+]] = torch.prim.ListConstruct %[[INT1]] : (!torch.int) -> !torch.list<int>
+  // CHECK: %[[TRUE:.+]] = torch.constant.bool true
+  // CHECK: %[[NONE:.+]] = torch.constant.none
+  // CHECK: %[[SUM:.+]] = torch.aten.sum.dim_IntList %[[ABS]], %[[DIMS]], %[[TRUE]], %[[NONE]] : !torch.vtensor<[2,0,4],f32>, !torch.list<int>, !torch.bool, !torch.none -> !torch.vtensor<[2,1,4],f32>
+  // CHECK: return %[[SUM]] : !torch.vtensor<[2,1,4],f32>
+  %0 = torch.operator "onnx.ReduceL1"(%arg0, %arg1) {torch.onnx.keepdims = 1 : si64} : (!torch.vtensor<[2,0,4],f32>, !torch.vtensor<[1],si64>) -> !torch.vtensor<[2,1,4],f32>
+  return %0 : !torch.vtensor<[2,1,4],f32>
+}
```
vivekkhandelwal1, line 949: I am not sure if we have added the e2e support for handling the `0-dim` case as mentioned here: https://onnx.ai/onnx/operators/onnx__ReduceL1.html#summary

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vinayakdsci: I extracted it into a common implementation, but it seems a little too complex. Is it OK if I push that for review?

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
vinayakdsci: Understood, Will do.

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
vinayakdsci: Should I remove this test then?

projects/pt1/e2e_testing/xfail_sets.py
vinayakdsci: This is where I am running into a problem, as `ReduceL1NormWithDTypeModule_basic` is failing with `Found dtype Double but expected Float`. No matter In which Order I apply type conversion, the error is not going away. Do you have any suspicion about what could be going on?

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
vivekkhandelwal1: Yeah, you can remove this test. But, I think it's necessary to error out on this as well.

vinayakdsci: So I should add a check in e2e tests?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -39,6 +39,127 @@ Value getItemOp(OpBinder binder, ConversionPatternRewriter &rewriter,
...
+      rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), keepDims);
+  Value dType = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+  // If we are using the ReducedSum as an intermediate op to be passed into
+  // another operation, we might not want to replace the Op. So we create a new
+  // Op and store the result in a variable.
+  if (!isIntermediateOp) {
+    rewriter.replaceOpWithNewOp<Torch::AtenSumDimIntListOp>(
+        binder.op, resultType, data, dimValueList, keepDimBool,
+        /*dtype=*/dType);
+  } else {
```
vivekkhandelwal1, line 156: Why do we need this functionality?

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vinayakdsci: This is to replace binder.op when the ReduceSum op is the last operation performed on the data. If it is an intermediate operation, and its result will be passed into another operation(imagine ReduceL2) We might need to create a new op value, as a replacement will give a segfault here.
That is why I put in  this check, just for future proofing.

vinayakdsci: I am sure it can be leveraged in the implementation of ReduceL2 too.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -39,6 +39,127 @@ Value getItemOp(OpBinder binder, ConversionPatternRewriter &rewriter,
...
 }
+
+// originalOperand is explicitly passed to handle the case where ReduceSum
+// was not the first operation performed on the operand, the original value of
+// which we might need for noop_with_empty_axes handling.
+LogicalResult reducedSumImpl(OpBinder binder,
+                             ConversionPatternRewriter &rewriter, Value &data,
+                             Value &originalOperand,
+                             Torch::ValueTensorType resultType,
+                             Value *storeResult, int64_t keepDims,
```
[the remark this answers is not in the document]
vinayakdsci: But wouldn't this make it necessary to initialize a variable to create a reference even when it is not required?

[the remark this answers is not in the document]
vinayakdsci: Right now I can pass in  a `nullptr`.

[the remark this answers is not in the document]
vinayakdsci: No, sorry, I was wrong here. When it is not required, I can simply pass in  a reference to operand, as it won't be modified anyway.

vinayakdsci:
@vivekkhandelwal1 could `ReduceL1NormComplexModule_basic` be failing because of failure in import/export tests for `LinalgVectorNorm`? `LinalgVectorNormComplexModule_basic` is failing with the same error message when run.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -39,6 +39,125 @@ Value getItemOp(OpBinder binder, ConversionPatternRewriter &rewriter,
...
+          rewriter.getI64IntegerAttr(axesInts[i]));
+      axesList.push_back(iv);
+    }
+  }
+
+  // Do not include absolute value in the noop
+  if (axesList.empty() && noop_with_empty_axes) {
+    rewriter.replaceOp(binder.op, originalOperand);
+    return success();
+  }
```
vivekkhandelwal1, line 138: Instead, you can pass `originalOperand` as an arg corresponding to `storeResult` that way you won't have a need for an extra arg. For this case you can just use `storeResult` and for the other ones you can modify it.

projects/pt1/e2e_testing/xfail_sets.py
vivekkhandelwal1: I can take a look after this patch of yours getting merged.

vivekkhandelwal1 (changes requested):

vivekkhandelwal1:
Hi @vinayakdsci, after a small change your patch is good to go. Also, please mark all those conversations as resolved which are fixed. Also, one more test is passing as can be seen in CI, please remove that from the xfail list. If it's passing only for the stable torch version then you can do something like: https://github.com/llvm/torch-mlir/commit/6e84752c395a828eb612b21be4ab26d9f7e60b22#diff-9e059002448f477a0e7ca75f6d440b9c4d0bfe9b092851afce938fdbe51b7368R2194-R2199

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vinayakdsci: The CI tests for torch-nightly seems to have been cancelled, so I am unable to judge whether this passes only for stable or for both. Is it OK if I remove it from the file altogether?

vinayakdsci:
Hi @vivekkhandelwal1, the CI for torch-nightly appears to have been cancelled, so I am unable to judge whether the test passes only for stable or for both. Is it OK, if I remove it altogether from the file?

vinayakdsci:
Currently I have added a check against nightly, so stable should pass. Thanks!

vivekkhandelwal1 (approved):

PR #3147, opened 2024-04-11
Adds Some Quantization Support for AtenMatmulOp

zjgarvey:
1. onnx.MatMulInteger now converts to aten.matmul instead of aten.mm
2. aten.matmul, for ranks >=2, now allows quantized inputs and will lower to linalg::quantized_matmul or linalg::quantized_batch_matmul.
3. added AtenMatmulOp to the FuseQuantizeOps rewrite patters QuantizeOperands, QuantizeTransposedOperands, and QuantizeAccumulator
4. added several tests, including some to test AtenMmOp with varying quantization signed-ness.
5. a quantized matmul mat-vec test is added to verify the failure to lower to linalg; cleaned of out-of-date code related to common torch-mlir lowering xfails. 
6. in debugging a real model with quantized matmuls, I found a bug on the scalarize-shapes pass which resulted from the aten.full op folder returning an incompatible result type. This is fixed by the small change here to [lib/Dialect/Torch/IR/TorchOps.cpp](https://github.com/llvm/torch-mlir/compare/main...zjgarvey:torch-mlir:MatMulIntegerFix?expand=1#diff-dc8ed165c207918e606490eee3984b1ad51d7034e6aac36fc046bf47f6f03f4f).


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -36,6 +36,26 @@ static void getZeroPoint(Value value, Value &zeropoint) {
   }
 }
 
+// for uint8 types, we shift down by 128 so that we can faithfully
+// represent the quantization with signed i8 types.
+static void signShift(PatternRewriter &rewriter, Location loc, Value &arg,
+                      Value &zp, bool isUnsignedType, int64_t numBits) {
+  if (!isUnsignedType)
+    return;
+  int64_t minSI = -std::pow(2, numBits - 1);
```
rsuderman, line 45: Bit shifting is magic
```
-(1 << (numBits - 1))
```

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -302,10 +302,59 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
     auto lhsType = lhs.getType().cast<RankedTensorType>();
     auto rhsType = rhs.getType().cast<RankedTensorType>();
 
+    auto lhsTorchType = op.getSelf().getType().cast<ValueTensorType>();
```
rsuderman, line 305: We should use the global case

```
        auto lhsTorchType = cast<ValueTensorType>(op.getSelf().getType());
        auto rhsTorchType = cast<ValueTensorType>(op.getOther().getType());
```

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -302,10 +302,59 @@ class ConvertAtenMatmulOp : public OpConversionPattern<AtenMatmulOp> {
...
+
+    if (static_cast<bool>(lhsZeroPoint) != static_cast<bool>(rhsZeroPoint)) {
+      return rewriter.notifyMatchFailure(
+          op, "unsupported: aten.matmul with mixed quantization");
+    }
+
+    bool isUnsigned = torch_to_linalg::isUnsignedTorchType(lhsTorchType);
+    bool isUnsignedR = torch_to_linalg::isUnsignedTorchType(rhsTorchType);
+
+    if (!lhsZeroPoint) {
```
rsuderman, line 324: Split this nested if into an unnested versions. Namely

if (!lhsZeroPoint && lhsTorchType.getDtype() != rhsTorchType.getDtype())
            op, "unsupported: aten.matmul with different input element types");
if (lhsZeroPoint) {
  ....
}

rsuderman (changes requested):

lib/Conversion/TorchToLinalg/Linear.cpp
zjgarvey: Haha, good point.

rsuderman (approved):

Issue #3143, opened 2024-04-11
[TORCH] add support for torch.aten.floor_divide.Scalar


IanWood1:
`torch.aten.floor_divide` is already implemented for 2 tensors but needs to be implemented for scalars as well

PR #3156, opened 2024-04-12
Added 2 Ops: Floor divide scalar and Floor divide scalar mode

IanWood1:
- Added linalg lowering for `AtenFloorDivideScalarOp`
  - Needed `AtenDivScalarModeOp` for the decomp.
- Added linalg lowering for `AtenDivScalarModeOp`
  - Moved linalg payload logic to `createDivModePayload()` since the logic was nearly identical for both `AtenDivScalarModeOp` and `AtenDivTensorModeOp`. Just a template function 
 -  Added `AtenDivScalarModeOp` lowering for stablehlo
 

Pytorch's [`torch.floor_divide()`](https://pytorch.org/docs/stable/generated/torch.floor_divide.html) in a previous version (for a reason unknown to me) preformed a truncation instead of "floor". The already implemented op `AtenFloorDivideTensorOp` was done before this change.  However, this wasn't caught because our testcases only tested positive floor division. I changed this to floor as well as adding a few test cases.


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -25,7 +25,10 @@
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/ADT/APSInt.h"
+#include <cassert>
+#include <iostream>
```
renxida, line 29: apparently including `<iostream>` is one of the cardinal sins in a LLVM project file.

Why is this included and can you do the same with just the llvm::errs etc streams?

IanWood1:
nod-ai/SHARK-Turbine#618

lib/Conversion/TorchToLinalg/Uncategorized.cpp
renxida: https://llvm.org/docs/CodingStandards.html#include-iostream-is-forbidden

renxida: (looking further it looks like you're moving some code from one file to another)

> 
> [Use raw_ostream](https://llvm.org/docs/CodingStandards.html#id52)
> LLVM includes a lightweight, simple, and efficient stream implementation in llvm/Support/raw_ostream.h, which provides all of the common features of std::ostream. All new code should use raw_ostream instead of ostream.
> 
> Unlike std::ostream, raw_ostream is not a template and can be forward declared as class raw_ostream. Public headers should generally not include the raw_ostream header, but use forward declarations and constant references to raw_ostream instances.
-- https://llvm.org/docs/CodingStandards.html#use-raw-ostream

Since we're already touching this code, let's modernize this part

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -213,6 +216,79 @@ createTriangularMatrix(OpBuilder &b, Location loc, ValueRange payloadArgs,
   return success();
 }
 
+template <typename OpT>
+Value createDivModePayload(OpBuilder &b, Location loc,
+                           const TypeConverter *converter,
+                           ValueRange payloadArgs, OpT op,
+                           ArrayRef<Value> operands) {
+  static_assert(std::is_same_v<OpT, AtenDivTensorModeOp> ||
```
renxida, line 224: nice touch

lib/Conversion/TorchToStablehlo/Basic.cpp
```diff
@@ -432,21 +432,25 @@ class ConvertAtenMulDivOp : public OpConversionPattern<AtenOpT> {
...
-    if (!matchPattern(divTensorModeOp.getRoundingMode(),
-                      m_TorchConstantStr(roundingMode)))
+    if (!matchPattern(opRoundingMode, m_TorchConstantStr(roundingMode))) {
       return rewriter.notifyMatchFailure(
           op, "only support constant str rounding mode");
+    }
 
     // if trunc and int, do nothing
-    if (roundingMode == "trunc" && isa<mlir::FloatType>(outElemTy)) {
+    if (roundingMode == "trunc" && outElemTy.isa<mlir::FloatType>()) {
```
renxida, line 453: @rsuderman you mentioned earlier that we want to write casts using the non-member function. Does that apply to `isa<>()`? Or does the opposite apply?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
IanWood1: embarrassing... it's just left over from some "debugging"

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -25,7 +25,10 @@
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/ADT/APSInt.h"
+#include <cassert>
```
rsuderman, line 28: Do you still need `cassert`? I see your static assert below.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -213,6 +216,79 @@ createTriangularMatrix(OpBuilder &b, Location loc, ValueRange payloadArgs,
...
+Value createDivModePayload(OpBuilder &b, Location loc,
+                           const TypeConverter *converter,
+                           ValueRange payloadArgs, OpT op,
+                           ArrayRef<Value> operands) {
+  static_assert(std::is_same_v<OpT, AtenDivTensorModeOp> ||
+                    std::is_same_v<OpT, AtenDivScalarModeOp>,
+                "template type must be a tensor/scalar div mode");
+  typename OpT::Adaptor adaptor(operands);
+  Type dtype = converter->convertType(op.getType())
+                   .template cast<RankedTensorType>()
```
rsuderman, line 229: Typically we do `cast<RankedTensorType>(...)` instead of using the templated member function. It avoids the extract `template` part.

lib/Conversion/TorchToStablehlo/Basic.cpp
rsuderman: I think we want to do the non-member version? IIRC it does a pointer case and some magic but its preferable due to better behavior for templated code. Similar comment at the top.

rsuderman (changes requested):

lib/Conversion/TorchToStablehlo/Basic.cpp
```diff
@@ -432,7 +432,8 @@ class ConvertAtenMulDivOp : public OpConversionPattern<AtenOpT> {
     Value result =
         rewriter.create<ChloOpT>(loc, outType, lhs, rhs, bcastDimensions);
 
-    if (!isa<AtenDivTensorModeOp, AtenDivScalarModeOp>(&op)) {
+    if (!std::is_same<AtenDivTensorModeOp, AtenOpT>() ||
```
IanWood1: Just changed isa to `is_same` since the type can be evaluated against tensor/scalar mode at compile time (if could be marked `constexpr` but idk how important that is)

IanWood1:
made requested changes: use non-member functions and removed includes

renxida (approved):

rsuderman (approved):

IanWood1:
closes #3143

PR #3163, opened 2024-04-15
[FxImporter] Type conversion to resolve the mismatch between Py type and schema type

penguin-wwy:


penguin-wwy:
Some end-to-end (e2e) test cases also have this issue, but fixing it did not increase the number of passing cases.
@stellaraccident

stellaraccident:
Thanks - will review in a bit. Can you walk me through what needs to be done on the CI to get the FxImporter e2e tests going? The reason I'm slow on reviewing fx_importer tests is because I manually do a lot of verification. Would be more efficient if I could rely more on the automation.

penguin-wwy:
> Thanks - will review in a bit. Can you walk me through what needs to be done on the CI to get the FxImporter e2e tests going? The reason I'm slow on reviewing fx_importer tests is because I manually do a lot of verification. Would be more efficient if I could rely more on the automation.

I understand that the current status is passing tests on the nightly torch, and we just need to add `python -m e2e_testing.main --config=torchdynamo -v` to the CI.

There are mainly two issues at present:
* Since the e2e config of fx_importer_backend comes from torchdynamo, there will be some problems with the config itself (I'm fixing the parts I've found), which will lead to the failure of some cases. These cases are currently in the xfail set and need time to identify and fix.
* The reasons for the failure of the cases in the `xfail set` and `crashing set` are not yet understood, and further iteration is required to address them.

stellaraccident (approved):
Thank you!

PR #3169, opened 2024-04-16
[Torch] Canonicalize aten.log

Xinyu302:
This PR did two things:
1. Canonicalize aten.log when this op has scalar input
2. Make AtenScalarImplicitOp::getCanonicalizationPatterns support float input


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -3343,6 +3347,38 @@ void AtenCatOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
   });
 }
 
+//===----------------------------------------------------------------------===//
+// AtenLogOp
+//===----------------------------------------------------------------------===//
+
+void AtenLogOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
```
qingyunqu, line 3354: I think fold implementation is more suitable than canonicalize implementation. Because it should not create new op.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -3343,6 +3347,38 @@ void AtenCatOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+    double floatScalar;
+    int64_t intScalar;
+    double resultScalar;
+    if (matchPattern(self, m_TorchConstantInt(&intScalar)))
+      resultScalar = std::log(intScalar);
+    if (matchPattern(self, m_TorchConstantFloat(&floatScalar)))
+      resultScalar = std::log(floatScalar);
+    Value result = rewriter.create<ConstantFloatOp>(
+        loc, rewriter.getF64FloatAttr(resultScalar));
+    rewriter.replaceOpWithNewOp<PrimNumToTensorScalarOp>(op, outType, result);
```
qingyunqu, line 3377: Use `PrimNumToTensorScalarOp` will cause dtype mismatch when `outType`'s element type is `f32`. Because that `PrimNumToTensorScalarOp (!torch.float)` should infer outType to be `f64`.

qingyunqu (changes requested):

PR #3173, opened 2024-04-16
[rfc] Use logging.getLogger for projects/pt1/e2e_testing

renxida:
see: #3172

This adds a `--print-ir` flag to activate debug printing on 

```
    logger = logging.getLogger("e2e_test")
```

which prints ir after compile calls

right now it only prints after onnx and linalg backend, which are the most commonly used ones, but we should really enforch something like this for every backend.


projects/pt1/e2e_testing/main.py
```diff
@@ -93,6 +98,18 @@ def _get_argparse():
...
+    ir_printer = logging.getLogger("ir_printer")
+    if args.print_ir:
+        print("WARNING: --print-ir is a work in progress feature.")
+        print("print-ir: Setting logging level to DEBUG and enabling IR printing.")
+        print("print-ir: This currently only affects the Linalg-on-Tensors and onnx configs.")
+        print("print-ir: Work in progress. See https://github.com/llvm/torch-mlir/issues/3172")
+        ir_printer.setLevel(logging.DEBUG)
+    else:
+        # disable logging
+        ir_printer.setLevel(logging.CRITICAL+1)
```
rsuderman, line 110: Why is this CRITICAL+1 ?

projects/pt1/python/torch_mlir_e2e_test/configs/onnx_backend.py
```diff
@@ -82,6 +83,9 @@ def __init__(self, backend: OnnxBackend, use_make_fx: bool = False):
     def compile(self, program: torch.nn.Module) -> Any:
         example_args = convert_annotations_to_placeholders(program.forward)
         onnx_module = convert_onnx(program, example_args)
+        ir_printer.debug("OnnxBackendTestConfig imported module:")
```
rsuderman, line 86: I woudn't bothering including the header / footers here

projects/pt1/python/torch_mlir_e2e_test/configs/stablehlo_backend.py
```diff
@@ -31,7 +33,9 @@ def __init__(self, backend: StablehloBackend):
     def compile(self, program: torch.nn.Module) -> Any:
         example_args = convert_annotations_to_placeholders(program.forward)
         module = torchscript.compile(program, example_args, output_type="stablehlo")
-
+        ir_printer.debug("StablehloBackendTestConfig compiled module:")
```
rsuderman, line 36: Ditto

projects/pt1/python/torch_mlir_e2e_test/framework.py
```diff
@@ -143,6 +143,7 @@ def compile(self, program: torch.nn.Module) -> CompiledArtifact:
         """Compile the provided torch.nn.Module into a compiled artifact"""
         pass
 
+
```
rsuderman, line 146: Unneeded

rsuderman (changes requested):

projects/pt1/e2e_testing/main.py
renxida: Was looking for the python idiom for completely disabling logging. This looks like the way to disable it.

projects/pt1/python/torch_mlir_e2e_test/framework.py
renxida: reverted

projects/pt1/e2e_testing/main.py
```diff
@@ -4,6 +4,7 @@
 # Also available under a BSD-style license. See LICENSE.
 
 import argparse
+import logging
```
ScottTodd, line 7: Some other projects to take inspiration from:

* [`shark_turbine/support/logging.py`](https://github.com/nod-ai/SHARK-Turbine/blob/main/core/shark_turbine/support/logging.py) and [`shark_turbine/support/debugging.py`](https://github.com/nod-ai/SHARK-Turbine/blob/main/core/shark_turbine/support/debugging.py)
* https://pytorch.org/docs/stable/logging.html

torch-mlir could share logging configuration code throughout the project or it could be isolated to just these testing folders. One advantage to doing it at the project level is that running both inside and outside of tests could get the same information. Could start with something in just the tests and see how it feels then generalize to the full project.

projects/pt1/e2e_testing/main.py
```diff
@@ -143,7 +153,7 @@ def main():
     if args.crashing_tests_to_not_attempt_to_run_and_a_bug_is_filed is not None:
         for arg in args.crashing_tests_to_not_attempt_to_run_and_a_bug_is_filed:
             if arg not in all_test_unique_names:
-                print(f"ERROR: --crashing_tests_to_not_attempt_to_run_and_a_bug_is_filed argument '{arg}' is not a valid test name")
+                logger.error(f"ERROR: --crashing_tests_to_not_attempt_to_run_and_a_bug_is_filed argument '{arg}' is not a valid test name")
```
ScottTodd, line 156: Is putting "ERROR:" in the message redundant here? (May depend on the logging formatter)

projects/pt1/e2e_testing/main.py
```diff
@@ -152,21 +162,24 @@ def main():
...
 
     # Report the test results.
-    failed = report_results(results, xfail_set, args.verbose, args.config)
+    failed = report_results(results, xfail_set,
+                            verbose=logger.level >= logging.INFO,
+                            config=args.config)
     if args.config == "torchdynamo":
-        print("\033[91mWarning: the TorchScript based dynamo support is deprecated. "
+        logger.warning("\033[91mWarning: the TorchScript based dynamo support is deprecated. "
               "The config for torchdynamo is planned to be removed in the future.\033[0m")
```
ScottTodd, line 183: Similar comment here - the escape code + "Warning:" text may be redundant with the logging level already set to 'warning'.

projects/pt1/python/torch_mlir_e2e_test/configs/linalg_on_tensors_backend.py
```diff
@@ -32,7 +34,7 @@ def compile(self, program: torch.nn.Module) -> Any:
         example_args = convert_annotations_to_placeholders(program.forward)
         module = torchscript.compile(
             program, example_args, output_type="linalg-on-tensors")
-
+        logger.debug(module)
```
ScottTodd, line 37: Could add context to this log, unless there is already some provided by the caller.

projects/pt1/e2e_testing/main.py
renxida: fixd

projects/pt1/e2e_testing/main.py
renxida: Wll add a formatter to do this.

projects/pt1/e2e_testing/main.py
```diff
@@ -71,10 +72,11 @@ def _get_argparse():
     parser.add_argument("-f", "--filter", default=".*", help="""
 Regular expression specifying which tests to include in this run.
 """)
-    parser.add_argument("-v", "--verbose",
-                        default=False,
-                        action="store_true",
-                        help="report test results with additional detail")
+    parser.add_argument("--log_level", default="WARNING", choices=["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"], help="set the log level")
```
ScottTodd, line 75: Found a tip for these choices at https://stackoverflow.com/a/71667675

add `type=str.upper` to the `parser.add_argument` options and then users can pass `--log_level=info` to have it be treated as `--log_level=INFO`

PR #3177, opened 2024-04-16
[torch] Adds Quantization Support for `aten.relu`

zjgarvey:
A choice was made to quantize the return type of Relu with a scale and zero point copied from the input's quantization scheme. With this choice, the torch-to-linalg conversion of quantized Relu essentially computes max(input, zeroPoint) in the elementwise payload.


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -528,10 +534,29 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
   if (auto relu = dyn_cast<AtenReluOp>(op)) {
-    if (!relu.getType()
-             .cast<ValueTensorType>()
-             .getDtype()
-             .isa<mlir::FloatType>()) {
+    Value zeroPoint;
+    getZeroPoint(relu.getSelf(), zeroPoint);
+    Value arg = payloadArgs[0];
+    auto reluTorchType = cast<ValueTensorType>(relu.getType());
+    if (zeroPoint) {
```
rsuderman, line 541: This is better if it is

```
if (!zeroPoint) {
      relu.emitError("unimplemented: non-integer quantized Relu.");
      return nullptr;
}
```

Also move up to where you get the zero point value

rsuderman (changes requested):

zjgarvey:
The recent commit tries to make the logic cleaner. I decided to drop the check for floating point entirely, and just treat zeroPoint as zero ConstantOp of the appropriate type for nonquantized relu.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -528,19 +535,67 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
-                                         payloadArgs[0], constZero);
-    return b.create<arith::SelectOp>(loc, pred, payloadArgs[0], constZero);
+    auto reluTorchType = cast<ValueTensorType>(relu.getType());
+    bool isUnsigned =
+        torch_to_linalg::isUnsignedTorchType(reluTorchType.getDtype());
+    if (zeroPoint) {
+      int64_t zeroPointInt;
+      int64_t width = intType.getWidth();
+      int64_t minForIntType = isUnsigned ? 0 : -(1 << (width - 1));
+      int64_t maxForIntType =
```
rsuderman, line 552: Seems dumb but mind including `assert(width < 64);`. It really shouldn't happen but its easier to have the verification contained.

projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -1794,6 +1800,9 @@
     "AtenMatmulQint8VV_basic",
     "AtenMatmulQint8VM_basic",
     "AtenMatmulQint8_basic",
+    "QuantizedReluInt32_basic",
```
rsuderman, line 1803: Is there a reason these tests are newly added but failing?

rsuderman (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
zjgarvey: Yeah, good idea.

projects/pt1/e2e_testing/xfail_sets.py
zjgarvey: It's just an export error.
```
Exporting the operator 'aten::_make_per_tensor_quantized_tensor' to ONNX opset version 17 is not supported. Please feel free to request support or submit a pull request on PyTorch GitHub: https://github.com/pytorch/pytorch/issues.
```
I could test quantized Relu through onnx with something similar to the quantized_models.py tests if you think it would be worth it to add one. I think the quantized models in this list:

```python
    "NativeGroupNormBackwardModule_basic",
    "QuantizedMLP_basic",
    "QuantizedSingleLayer_basic",
    "QuantizedBatchedInputSingleLayer_basic",
    "ReduceMaxAlongDimUnsignedInt_basic",
    "ReduceMinAlongDimUnsignedInt_basic",
    "ElementwiseToDtypeI64ToUI8Module_basic",
}
```
 
fail due to the error:

 ```
 error: unknown: unsupported by backend contract: module initializers
        note: unknown: see current operation: "torch.initialize.global_slots"(%18, %22) <{slotSymNames = [@layers.0._packed_params._packed_params, @layers.2._packed_params._packed_params]}> : (!torch.LinearParams, !torch.LinearParams) -> ()
        note: unknown: this is likely due to InlineGlobalSlots being unable to inline a global slot
```

are likely only because of the linear layers not getting unpacked. So a Relu-only model would probably work.

rsuderman: Makes sense. Tracing quantized models always seems to be a problem - I hope that dynamo figures that out.

rsuderman (approved):

PR #3179, opened 2024-04-17
[onnx] Add OnnxToTorch support for `onnx.ConvInteger`

jinchen62:
All e2e iree tests compiled, but they have the run issue of mismatch of dtype like the following
```
expected:
1x1x2x2xsi32=[[[12 16][24 28]]]
actual:
1x1x2x2xi32=[[[12 16][24 28]]]
```

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -9,6 +9,7 @@
 
 #include "mlir/IR/DialectResourceBlobManager.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
+#include "torch-mlir/Conversion/TorchOnnxToTorch/Utils.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/Support/FormatVariadic.h"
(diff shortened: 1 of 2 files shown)


zjgarvey:
@jinchen62 

Before I read this PR too carefully: have you looked at the conversion for onnx.MatmulInteger? 

Similar to AtenMatmulOp for onnx.MatmulInteger, we already have quantization support for AtenConvolutionOp in torch-mlir, so this conversion should leverage that if possible. If you pass in Aten_MakePerTensorQuantizedTensorOp's for the convolution operands (like with MatmulInteger), the torch-fuse-quantize-ops pass should appropriately quantize the bias and accumulators for you, and the torch-to-linalg lowering should lower to appropriate linalg named ops for quantized convolution. These quantized convolution ops automatically extend the lower precision inputs to the accumulator type, apply zero-Point corrections, etc. 

The benefits would be 

1. Potentially making this conversion significantly simpler. 
2. The linalg named ops for quantized convolutions might have special optimizations. Especially for matmuls and convs, I think this is probably quite important. 

I would ask @rsuderman to corroborate what I'm saying here, because I'm not 100% familiar with quantized conv support.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1035,6 +1035,181 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        if (binder.s64IntegerArrayAttr(strides, "strides", defaultStrides)) {
+          return failure();
+        }
+        if (strides.size() != rank - 2) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "strides list size does not match the number of axes");
+        }
+
+        Value cstOne;
+        if (binder.op->getNumOperands() > 2) {
```
rsuderman, line 1130: You can remove the `if(binder.op->getNumOperands() > N)` checks Try something like this

```
Value weightZeroPoint;
if (!binder.tensorOperandAtIndex(inputZeroPoint, 2)) {
  cstOne = rewriter.create<Torch::ConstantIntOp>(
    binder.getLoc(), rewriter.getI64IntegerAttr(1));
  auto inputType = input.getType().cast<Torch::ValueTensorType>();
  input = rewriter.create<Torch::AtenSubTensorOp>(
    binder.getLoc(), inputType, input, inputZeroPoint, cstOne);
}
Value weightZeroPoint;
if (!binder.tensorOperandAtIndex(weightZeroPoint, 3)) {
  weight = rewriter.create<Torch::AtenSubTensorOp>(
  binder.getLoc(), weightTensorType, weight, weightZeroPoint,
  cstOne);
}
```

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1035,6 +1035,181 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          cstStrides.push_back(rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(i)));
+        }
+        Value cstZero = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getI64IntegerAttr(0));
+        cstOutputPadding = {cstZero, cstZero};
+
+        Value paddingList = rewriter.create<Torch::PrimListConstructOp>(
+            binder.getLoc(),
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
```
rsuderman, line 1186: `rewriter.getType<Torch::ListType>(rewriter.getType<Torch::IntType>())`

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1035,6 +1035,181 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        Value cstOne;
+        if (binder.op->getNumOperands() > 2) {
+          Value inputZeroPoint;
+          if (binder.tensorOperandAtIndex(inputZeroPoint, 2)) {
+            return failure();
+          }
+          cstOne = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(1));
+          auto inputType = input.getType().cast<Torch::ValueTensorType>();
+          input = rewriter.create<Torch::AtenSubTensorOp>(
```
rsuderman, line 1138: You can't do this as this could cause a wrap around on the values.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1035,6 +1035,181 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+              binder.getLoc(), inputType, input, inputZeroPoint, cstOne);
+        }
+        if (binder.op->getNumOperands() > 3) {
+          Value weightZeroPoint;
+          if (binder.tensorOperandAtIndex(weightZeroPoint, 3)) {
+            return failure();
+          }
+          weight = rewriter.create<Torch::AtenSubTensorOp>(
```
rsuderman, line 1146: Same here

rsuderman:
> @jinchen62
> 
> Before I read this PR too carefully: have you looked at the conversion for onnx.MatmulInteger?
> 
> Similar to AtenMatmulOp for onnx.MatmulInteger, we already have quantization support for AtenConvolutionOp in torch-mlir, so this conversion should leverage that if possible. If you pass in Aten_MakePerTensorQuantizedTensorOp's for the convolution operands (like with MatmulInteger), the torch-fuse-quantize-ops pass should appropriately quantize the bias and accumulators for you, and the torch-to-linalg lowering should lower to appropriate linalg named ops for quantized convolution. These quantized convolution ops automatically extend the lower precision inputs to the accumulator type, apply zero-Point corrections, etc.
> 
> The benefits would be
> 
> 1. Potentially making this conversion significantly simpler.
> 2. The linalg named ops for quantized convolutions might have special optimizations. Especially for matmuls and convs, I think this is probably quite important.
> 
> I would ask @rsuderman to corroborate what I'm saying here, because I'm not 100% familiar with quantized conv support.

They are extremely important. By making sure things plumb via the `linalg` `quantized` operations we are able to redo how the zero-point calculation is performed and post convolution correct the values. Happy to include a link for interested parties.

More important we cannot simply apply the zero point offset before feeding into the convolution as these are small bit numbers (e.g. 8-bit) it would cause wrap around yielding a completely incorrect result.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1035,6 +1035,195 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        }
+        for (int64_t i : strides) {
+          cstStrides.push_back(rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(i)));
+        }
+        Value cstZero = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getI64IntegerAttr(0));
+        cstOutputPadding = {cstZero, cstZero};
+
+        Value paddingList = rewriter.create<Torch::PrimListConstructOp>(
```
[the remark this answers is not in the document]
zjgarvey: Yes : Thanks to @AmosLewis, there is [createConstantIntList from Utils.cpp](https://github.com/llvm/torch-mlir/blob/6c4f7deebb89308b8916648a9d3bd1d0ce4edea6/lib/Conversion/TorchOnnxToTorch/Utils.cpp#L17)

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1035,6 +1035,195 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+            rewriter.getType<Torch::ListType>(
+                rewriter.getType<Torch::IntType>()),
+            cstOutputPadding);
+        Value transposed =
+            rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), false);
+        Value bias = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+        Value cstGroup = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getI64IntegerAttr(group));
+
+        auto q = [&](Type qty) -> Type {
```
zjgarvey, line 1202: Another helper function was made for this purpose. I only know this because I just linked the same file earlier: 

[getQTorchTypeFromTorchIntType in Utils.cpp](https://github.com/llvm/torch-mlir/blob/6c4f7deebb89308b8916648a9d3bd1d0ce4edea6/lib/Conversion/TorchOnnxToTorch/Utils.cpp#L31)

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1035,6 +1035,195 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        }
+
+        auto weightTensorType = weight.getType().cast<Torch::ValueTensorType>();
+        if (!weightTensorType || !weightTensorType.hasSizes())
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected weight type having sizes");
+        ArrayRef<int64_t> weightShape = weightTensorType.getSizes();
+        SmallVector<int64_t> kernelShape;
+        if (binder.s64IntegerArrayAttr(kernelShape, "kernel_shape", {}))
+          return failure();
```
zjgarvey, line 1088: Based on the onnx documentation, the kernel shape could be optional. If so, I'd add a todo comment.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1035,6 +1035,195 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          inputZp = rewriter.create<Torch::AtenItemOp>(
+              binder.getLoc(), rewriter.getType<Torch::IntType>(), inputZp);
+        }
+        if (binder.tensorOperandAtIndex(weightZp, 3))
+          weightZp = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(0));
+        if (auto zpTy = dyn_cast<Torch::ValueTensorType>(weightZp.getType())) {
+          for (auto dim : zpTy.getSizes())
+            if (dim != 1)
+              return failure();
```
zjgarvey, line 1076: Is this what you mentioned earlier about the case where the quantization is per channel? If so, I would make a more descriptive match failure and add a todo. 

On a similar note: does Aten Item Op work on tensors like [1,1,1,1]? If not, you might need a squeeze for ranks > 0 or 1 or whatever works for item op.

zjgarvey (changes requested):
Thanks for the work @jinchen62 ! I've left some comments that I think will help clean up the code a bit.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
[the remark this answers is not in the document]
jinchen62: I made the changes to others but I think `paddingList` could be remained cuz it already loops `padding` above for two cases, so no need to store the int then loop again in the helper function?

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
jinchen62: I think kernel_shape is just for checking if it matches the weight shape here, so if not present then no check is needed? Correct me if I'm wrong.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
jinchen62: aten.item would work with [1,1,1,1] as long as there is only one element in the tensor.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
zjgarvey: Yeah, I think you are probably right, but I don't know what the onnx op actually does with the information. I think you could ignore my comment.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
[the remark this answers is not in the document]
zjgarvey: Yeah, good idea.

zjgarvey (commented):
Before approving, are there any e2e-tests yet for this op? 

Also, have you lowered the test IR to linalg to see if it gets converted to a quantized conv op?

zjgarvey:
> Before approving, are there any e2e-tests yet for this op?
> 
> Also, have you lowered the test IR to linalg to see if it gets converted to a quantized conv op?

Ah, I see the comment about the dtypes at the beginning.

Did the values match even though the dtype issue popped up?

zjgarvey:
> @zjgarvey Yes the values match for both tests `test_convinteger_with_padding` and `test_convinteger_without_padding`.

What about the linalg IR? Does it generate quantized convolutions?

jinchen62:
@zjgarvey I think so, this is the lowering ir https://gist.github.com/jinchen62/804dcfebb609eb321adba61a1ee0eb6b.

zjgarvey (approved):
Thanks for the changes and testing! Make sure to resolve our conversations since I don't have merge privileges.

rsuderman (approved):

PR #3180, opened 2024-04-17
[FxImporter] Replace local_scalar_dense in fx_importer

penguin-wwy:


python/torch_mlir/extras/fx_importer.py
```diff
@@ -1420,6 +1420,16 @@ def _import_torch_op_overload(
                 for key_node in node.users:
                     if key_node.target == torch.ops.aten.baddbmm.default:
                         node.target = target = torch.ops.aten.zeros.default
+        elif target == torch.ops.aten._local_scalar_dense.default:
```
stellaraccident, line 1423: This cascade of ifs is getting a bit out of hand. I don't want to break your stride: maybe a todo to promote it to a table?

stellaraccident (approved):

stellaraccident (approved):

python/torch_mlir/extras/fx_importer.py
penguin-wwy: In a previous version before pr, I had actually extracted a function.

But I'm hesitating about the level of processing needed. For example, do I need to match multiple nodes? Should I normalize all nodes and call `fx.graph.lint` before importing nodes?

Therefore, I did not include in this pr.

stellaraccident: Yeah, I remember when I accepted patches for the first of these. My sense is that each of these would be better as an fx preprocessing pass, but people have been daisy chaining things here for a while. It's good to face test coverage for them: makes it easier to contemplate cleaning up.

penguin-wwy: In our team's project, I have a similar practice of maintaining an fx graph preprocess function, which works independently like a pass. It takes a GraphModule as input, uses pattern matching to replace "bad" operators to make the backend work more easily, and calls `graph.eliminate_dead_code` and `graph.lint` before exiting.

If we reach a consensus on this point, then I will try to modify it at a convenient time.

stellaraccident: Yeah, I think that is the right thing to do. Would be happy with something like this.

PR #3184, opened 2024-04-18
Fix nan issue for fp16 torch.randn/randn_like in ConvertAtenUniformOp

aviator19941:


vivekkhandelwal1 (approved):
LGTM! I'm not sure about the performance regressions arising out of this but still if this results in numerical stability then it's fine.

PR #3187, opened 2024-04-18
[onnx] Fix `onnx.Not` for non-bool inputs

rsuderman:
Need to perform a bool cast to support `onnx.Not` on non-bool inputs.


renxida:
https://onnx.ai/onnx/operators/onnx__Not.html

It sounds like onnx not isn't supposed to accept non bool inputs. Do we have models that use this out of spec?

renxida (approved):

PR #3190, opened 2024-04-19
[StableHLO] Fix aten.clamp.Tensor in FxImporter2StableHLO

penguin-wwy:
The FX importer will pass static shapes to the Torch dialect, so it needs to generate a StableHLO that satisfies shape inference.


lib/Conversion/TorchToStablehlo/Basic.cpp
```diff
@@ -1500,6 +1500,10 @@ LogicalResult ConvertAtenOp<AtenClampTensorOp>::matchAndRewrite(
     }
     maxValue = *maxInfo;
   }
+  if (inputType.hasStaticShape()) {
+    minValue = hlo::promoteAndBroadcast(rewriter, minValue, inputType);
```
qingyunqu, line 1504: `stablehlo.clamp` support min/max to be 0d tensor or shape == input shape, see https://github.com/openxla/stablehlo/blob/main/docs/spec.md#clamp.  
I think it's no need to broadcast when it's 0d tensor.

penguin-wwy: Tanks.
In fact, in similar cases `torch.ops.aten.clamp_min(x, min=2)`, this test passes(case name ElementwiseClampMinModule_basic)
The case that this PR resolves is written `torch.ops.aten.clamp_min(x, min=torch.tensor([-5]))`, that min input is `torch.vtensor<[1],si64>`.
Maybe I should further determine here whether the min value's rank == 0?

qingyunqu: I see. Thanks!

qingyunqu (approved):

PR #3193, opened 2024-04-19
[onnx] Add `onnx-to-torch` lowering for random ops

vivekkhandelwal1:


renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2265,4 +2265,242 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          return failure();
+        }
+
+        dtypeIntTorch = onnxDtypeIntToTorchDtypeInt(dtypeIntOnnx);
+        if (dtypeIntTorch == -1) {
+          return rewriter.notifyMatchFailure(
+              binder.op,
+              "unimplemented support for the given dtype conversion");
+        }
+        Value constDtype = rewriter.create<Torch::ConstantIntOp>(
```
zjgarvey, line 2416: I think it is better to have the shorter version, right?
```cpp
...<Torch::ConstantIntOp>(binder.getLoc(), rewriter.getI64IntegerAttr(dtypeIntTorch);
```

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2265,4 +2265,242 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(64),
+                                    dtypeIntTorch));
+
+        SmallVector<Value> shapeValues;
+        for (auto dim : shape)
+          shapeValues.push_back(rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(dim)));
+
+        Value shapeList = rewriter.create<Torch::PrimListConstructOp>(
```
zjgarvey, line 2426: Earlier comment about the createConstantIntList()

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2265,4 +2265,242 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        }
+        Value constDtype = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(64),
+                                    dtypeIntTorch));
+
+        Value cstNone = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+        Value cstFalse =
+            rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), false);
+        input = rewriter.create<Torch::AtenToDtypeOp>(
```
zjgarvey, line 2489: I'm a bit confused on this, and similarly with the randomNormalLike ops. If you aren't making a copy, does AtenUniformOp (or the similar op in the other conversion) make an empty memory format op at some point of the same shape as input? Or does it overwrite the contents of input and replace them with randomized values?

zjgarvey (changes requested):
Thanks @vivekkhandelwal1 . 

I mostly have some questions and small changes.

A side question: Sometimes you put a lot of lit tests, and sometimes none. Do you only use lit tests when e2e-tests are insufficient?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2265,4 +2265,242 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            binder.op, resultType, input, cstMean, cstStd,
+            /*generator=*/cstNone);
+        return success();
+      });
+  patterns.onOp(
+      "RandomUniform", 1,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        SmallString<64> name("torch.onnx.seed");
+        auto attr = binder.op->getAttr(name);
+        if (attr) {
```
rsuderman, line 2392: Unneeded curly brackets.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2265,4 +2265,242 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            binder.op, resultType, self, cstLow, cstHigh,
+            /*generator=*/cstNone);
+        return success();
+      });
+  patterns.onOp(
+      "RandomUniformLike", 1,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        SmallString<64> name("torch.onnx.seed");
+        auto attr = binder.op->getAttr(name);
+        if (attr) {
```
rsuderman, line 2456: Ditto

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2265,4 +2265,242 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                                    dtypeIntTorch));
+
+        SmallVector<Value> shapeValues;
+        for (auto dim : shape)
+          shapeValues.push_back(rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(dim)));
+
+        Value shapeList = rewriter.create<Torch::PrimListConstructOp>(
+            binder.getLoc(),
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
```
rsuderman, line 2428: `rewriter.getType<Torch::ListType>(rewriter.getType<Torch::IntType>())`

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2265,4 +2265,242 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                       binder.op, resultType, input, cstAlpha, value);
                   return success();
                 });
+  patterns.onOp(
+      "RandomNormal", 1,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        SmallString<64> name("torch.onnx.seed");
+        auto attr = binder.op->getAttr(name);
+        if (attr) {
```
rsuderman, line 2273: You should name it something more specific than `attr` in this case. E.g. `seedAttr`

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2265,4 +2265,242 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        if (binder.tensorOperand(input) ||
+            binder.s64IntegerAttr(dtypeIntOnnx, "dtype", 1) ||
+            binder.f32FloatAttr(high, "high", 1.0) ||
+            binder.f32FloatAttr(low, "low", 0.0) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+
+        dtypeIntTorch = onnxDtypeIntToTorchDtypeInt(dtypeIntOnnx);
+        if (dtypeIntTorch == -1) {
```
rsuderman, line 2476: See comment about `std::optional`

rsuderman (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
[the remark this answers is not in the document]
vivekkhandelwal1: Yeah, the seed can be any value so the only option to check if it's present or not is this way.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: The random ops during Torch->Linalg lowering creates new empty tensors in which results are stored. They don't overwrite the existing input operands. Although, the shape and dtype info is sometimes inferred from the input operand.

rsuderman (approved):

lib/Conversion/TorchOnnxToTorch/Utils.cpp
```diff
@@ -59,3 +59,41 @@ bool mlir::torch::onnx_c::areAllElementsDistinct(SmallVector<int64_t> array) {
   // as array's size.
   return (set.size() == array.size());
 }
+
+std::optional<int64_t>
+mlir::torch::onnx_c::onnxDtypeIntToTorchDtypeInt(int64_t dtypeIntOnnx) {
```
zjgarvey, line 64: What happened to the Endian class?

zjgarvey (approved):
As long as Endian class isn't used anywhere and you are just cleaning it up, I think this looks good.

lib/Conversion/TorchOnnxToTorch/Utils.cpp
vivekkhandelwal1: It's declared in Utils.h

PR #3195, opened 2024-04-19
[onnx] Extend op version number of `onnx.ScatterElements`

rsuderman:
Version number was set too high. Lowered to support more cases allows more tests to pass.


renxida (approved):

stellaraccident:
Reminder: Please set your git email address and check it on reviews.

PR #3196, opened 2024-04-20
[onnx] Support `onnx.OneHot` lowering to `torch`

rsuderman:
[onnx] Support `onnx.OneHot` lowering to `torch`

Leverage the `aten.onehot` implementation along with `aten.transpose` and
`aten.where.scalar`.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1566,6 +1566,99 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
                       binder.op, resultType, input);
                   return success();
                 });
+  patterns.onOp(
+      "OneHot", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        llvm::SmallVector<Value> inputs;
+        Torch::ValueTensorType resultType;
+        if (binder.tensorOperandsList(inputs) ||
+            binder.tensorResultType(resultType))
+          return failure();
+
+        if (inputs.size() != 3)
+          return rewriter.notifyMatchFailure(binder.op, "expected 3 operands");
+
+        int64_t axis;
+        if (binder.s64IntegerAttr(axis, "axis", -1))
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "`axis` attr not found");
+
+        auto loc = binder.getLoc();
+        Value indices = inputs[0];
+        Value depth = inputs[1];
+        Value values = inputs[2];
+
+        auto indicesTy = cast<Torch::ValueTensorType>(indices.getType());
+        auto valuesTy = cast<Torch::ValueTensorType>(values.getType());
+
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1557,6 +1557,99 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        if (inputs.size() != 3)
+          return rewriter.notifyMatchFailure(binder.op, "expected 3 operands");
+
+        int64_t axis;
+        if (binder.s64IntegerAttr(axis, "axis", -1))
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "`axis` attr not found");
+
+        auto loc = binder.getLoc();
+        Value indices = inputs[0];
```
renxida, line 1577: const auto[i, c, d] = tuple;

can we do tuple unpacking with smallvectors?

should we be doing tuple unpacking?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1557,6 +1557,99 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        Value depth = inputs[1];
+        Value values = inputs[2];
+
+        auto indicesTy = cast<Torch::ValueTensorType>(indices.getType());
+        auto valuesTy = cast<Torch::ValueTensorType>(values.getType());
+
+        axis = axis < 0 ? axis + indicesTy.getSizes().size() + 1 : axis;
+
+        depth = rewriter.create<Torch::AtenItemOp>(
```
renxida, line 1586: specs say depth should be cast to inttype if it wasn't an inttype. would item op do that for us?

renxida: depth (heterogeneous) - T2:

Scalar or Rank 1 tensor containing exactly one element, specifying the number of classes in one-hot tensor. This is also the size of the one-hot dimension (specified by ‘axis’ attribute) added on in the output tensor. The values in the ‘indices’ input tensor are expected to be in the range [-depth, depth-1]. In case ‘depth’ is of non-integer type, it will be casted to int64 before use.

renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
rsuderman: Tuple unpacking is used for `std::pair` and `std::tuple`, won't work with small vector as its not a known fixed size at compilation time.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
rsuderman: Fixed.

PR #3197, opened 2024-04-20
[stablehlo] add aten.remainder.Tensor op conversion support

penguin-wwy:


lib/Conversion/TorchToStablehlo/Basic.cpp
```diff
@@ -1817,6 +1817,30 @@ LogicalResult ConvertAtenOp<AtenFlipOp>::matchAndRewrite(
...
+    AtenRemainderTensorOp op, OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+  Value lhs = adaptor.getSelf();
+  Value rhs = adaptor.getOther();
+
+  auto lhs_type = cast<RankedTensorType>(lhs.getType());
+  auto lhs_elem_type = lhs_type.getElementType();
+  auto rhs_type = cast<RankedTensorType>(rhs.getType());
+  auto rhs_elem_type = rhs_type.getElementType();
+  if (isa<mlir::FloatType>(lhs_elem_type) &&
```
qingyunqu, line 1832: I think always promoting `lhs` and `rhs` to `result element type` is more simple.

penguin-wwy: The remainder.Tensor op also supports `(tensor<integer>, tensor<integer>) -> tensor<integer>`. So I think that calculating the result type would not be simpler than this, maybe?

qingyunqu: the case you describe doesn't matter. you can just use `hlo::promoteType` here.

qingyunqu: `stablehlo.convert i64 => i64` will be fold in backend compiler.

penguin-wwy: The result type also need be selected, like this
```
if lhs_ty.isInt && rhs_ty.isInt:
    result_ty = int_ty
else if lhs_ty.isxxx
    ....
```

And this can lead to a bad type conversion operation, making the thrown error difficult to locate. I understand that any type errors that do not meet the requirements should be alerted by the RemOp.

qingyunqu: emmm, I think it will be:
```c++
auto resultTy = cast<RankedTensorType>(op.getType());
auto resultElementTy = resultTy.getElementType();
lhs = hlo::promoteType(rewriter, op->getLoc(), lhs, resultElementTy);
rhs = hlo::promoteType(rewriter, op->getLoc(), rhs, resultElementTy);
rewriter.replaceOpWithNewOp<stablehlo::RemOp>(op, lhs, rhs);
```

penguin-wwy: Indeed, this is more simple, thanks.

stellaraccident (approved):

PR #3198, opened 2024-04-20
[stablehlo] add aten.fmod.Tensor op conversion support

penguin-wwy:


lib/Conversion/TorchToStablehlo/Basic.cpp
```diff
@@ -1833,6 +1833,37 @@ LogicalResult ConvertAtenOp<AtenRemainderTensorOp>::matchAndRewrite(
   return success();
 }
 
+// AtenFmodTensorOp
+// torch.fmod(a, b) == a - a.div(b, rounding_mode="trunc") * b
```
qingyunqu, line 1837: As your description, maybe it's a good idea to write a decompisition on torch dialect.

penguin-wwy: For linalg, it decomposes directly to linalg dialect at a similar position. For ONNX, I'm not sure have any impact. I will try to implement it on torch decompisition to see if there are any issues.

stellaraccident (commented):
Yeah, I don't think we should be unconditionally decomposing such primitives for the benefit of hlo level op sets. Fine to have some notion of optional decompositions that backends can use if helpful.

stellaraccident (approved):
While a value judgment, I think hlo is the outlier here, requiring a decomposition, so I would do it like this. It is a weak opinion, though, and I defer to the folks maintaining this backend.

PR #3200, opened 2024-04-22
Switch to pre-commit for lint checks.

stellaraccident:
Users can run via `pre-commit run` or set up a hook as described in the instructions: https://pre-commit.com/

The CI is set to only run pre-commit on files changed in the patch. We will run with `--all-files` in a separate patch.


penguin-wwy:
Before this, we haven't enabled black lint for Python code, do we need to add the black config file?
https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-format

.pre-commit-config.yaml
```diff
@@ -0,0 +1,21 @@
...
+    hooks:
+    -   id: trailing-whitespace
+    -   id: end-of-file-fixer
+    -   id: check-ast
+    -   id: check-yaml
+    -   id: check-added-large-files
+-   repo: https://github.com/psf/black
+    rev: 22.10.0
+    hooks:
+    -   id: black
```
hameerabbasi, line 16: How about using [Ruff](https://github.com/astral-sh/ruff) for formatting and linting Python?

stellaraccident: I'd prefer to stick with black as we use it on other projects and people/tooling are more familiar with it. I ran it a bunch of times during testing this patch, and performance was not an issue for this project.

Can reevaluate later if it makes a difference.

renxida:
excited excited excited

finally no more passing around precommit gists

stellaraccident:
> Before this, we haven't enabled black lint for Python code, do we need to add the black config file? https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-format

Added.

PR #3202, opened 2024-04-22
[torch] `aten.eye` should use dynamic dims when no static dims are available

renxida:


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1059,44 +1059,44 @@ class DecomposeAtenEyeMOp : public OpRewritePattern<AtenEyeMOp> {
...
+        n == kUnknownSize ? std::nullopt : llvm::ArrayRef(n), si64Type);
     Value rangeN = rewriter.create<AtenArangeOp>(
-        loc, arangeType, op.getN(), /*dtype=*/int64Dtype, /*layout=*/none,
+        loc, rangeNType, op.getN(), /*dtype=*/int64Dtype, /*layout=*/none,
         /*device=*/op.getDevice(), /*pin_memory=*/none);
 
-    auto arangeType1 =
-        outType.getWithSizesAndDtype(llvm::ArrayRef(m), si64Type);
+    Type rangeMType = outType.getWithSizesAndDtype(
+        m == kUnknownSize ? std::nullopt : llvm::ArrayRef(m), si64Type);
```
rsuderman, line 1097: Constant a `SmallVector<int64_t>` before hand. Its always a bit unclear when its safe to construct an `ArrayRef` from a value.

rsuderman (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1059,44 +1059,49 @@ class DecomposeAtenEyeMOp : public OpRewritePattern<AtenEyeMOp> {
...
+    // if output shape is not available, try to get shape from input
+    if (n == kUnknownSize)
+      matchPattern(op.getN(), m_TorchConstantInt(&n));
+    if (m == kUnknownSize)
+      matchPattern(op.getM(), m_TorchConstantInt(&m));
+
+    // prepare two unsqueezed ranges that are equal on and only on the diagonal
+    std::optional<llvm::SmallVector<int64_t, 1>> rangeNSize;
+    if (n != kUnknownSize)
+      rangeNSize = {n};
```
rsuderman, line 1093: `rangNSize.push_back(n);`

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1059,44 +1059,49 @@ class DecomposeAtenEyeMOp : public OpRewritePattern<AtenEyeMOp> {
...
+      m = outType.getSizes().back();
+    }
+    // if output shape is not available, try to get shape from input
+    if (n == kUnknownSize)
+      matchPattern(op.getN(), m_TorchConstantInt(&n));
+    if (m == kUnknownSize)
+      matchPattern(op.getM(), m_TorchConstantInt(&m));
+
+    // prepare two unsqueezed ranges that are equal on and only on the diagonal
+    std::optional<llvm::SmallVector<int64_t, 1>> rangeNSize;
```
rsuderman, line 1091: You can just make this an `llvm::SmallVector`

rsuderman (changes requested):

rsuderman (approved):

PR #3209, opened 2024-04-23
[Stablehlo] support unary ops which promote to floating point

qingyunqu:
* promote input to output element-type when lowering to stablehlo, so that it could satisfy stablehlo's type constraints.
* split promote-to-fp unary ops from fp-only unary ops.


Vremold (approved):
LGTM.

PR #3212, opened 2024-04-23
Support select_last_index attribute of onnx argmin op

jinchen62:
The tests listed in https://github.com/nod-ai/SHARK-Turbine/issues/648 all compiled, and the values of results match, but having runtime issue of dtype mismatch of i/si.


rsuderman (approved):
Ditto comment about end-to-end test

PR #3214, opened 2024-04-23
[onnx] handle dynamic padSize tensor in onnx.Pad

PhaneeshB:
- Fix pad size to data_rank for dynamic paddingSize Tensor.
- This fix is in accordance with [input specification](https://onnx.ai/onnx/operators/onnx__Pad.html#inputs) for onnx.Pad
- Impl will need to be updated for dynamic padSize when support for `axes` is added.


AmosLewis (approved):
LGTM

PR #3215, opened 2024-04-23
[Torch] Support AtenProdOp on linalg and stablehlo

Xinyu302:


qingyunqu (approved):
LGTM. Thanks for the effort.

PR #3218, opened 2024-04-23
Fix reducesum onnx lit test to linalg lowering fails

renxida:
fixes https://github.com/nod-ai/SHARK-Turbine/issues/653


vivekkhandelwal1 (commented):
Is there any reason why we have removed all the detailed checks and replaced them just with the checks containing op names?

renxida:
> Is there any reason why we have removed all the detailed checks and replaced them just with the checks containing op names?

I did so because i was feeling that the original checks were a little too detailed. For example, it made sure that there are two instances of int_0, and running CSE might cause the tests to fail.

I'm operating under the understanding that the lit tests are supposed to make sure the generated IR is valid & the generation doesn't crash.

But given what I learned from @zjgarvey in the standup this morning, we also rely on the tests as a source of understanding for how the lowerings work.

So I'd love your thoughts on how detailed the tests should be @vivekkhandelwal1 @zjgarvey

vivekkhandelwal1:
> > Is there any reason why we have removed all the detailed checks and replaced them just with the checks containing op names?
> 
> I did so because i was feeling that the original checks were a little too detailed. For example, it made sure that there are two instances of int_0, and running CSE might cause the tests to fail.
> 
> I'm operating under the understanding that the lit tests are supposed to make sure the generated IR is valid & the generation doesn't crash.
> 
> But given what I learned from @zjgarvey in the standup this morning, we also rely on the tests as a source of understanding for how the lowerings work.
> 
> So I'd love your thoughts on how detailed the tests should be @vivekkhandelwal1 @zjgarvey

I think the checks which are already present are fine, if you want then you can just remove the constant's checks but that won't we much either. Also, what's the reason that we have removed a lit test?

rsuderman (approved):

PR #3219, opened 2024-04-24
[Torch] support aten.trunc

qingyunqu:
decompose `trunc(x)` to `sign(x) * floor(abs(x))`


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -5886,6 +5886,32 @@ class DecomposeAtenCosineSimilarityOp
...
+                                PatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value self = op.getSelf();
+
+    auto resultTy = dyn_cast<ValueTensorType>(op.getType());
+    if (!resultTy || !resultTy.hasDtype()) {
+      return rewriter.notifyMatchFailure(op, "result must have dtype");
+    }
+
+    if (resultTy.getDtype().isa<mlir::FloatType>()) {
```
penguin-wwy, line 5903: Could you please modify functional style `isa<mlir::FloatType>(resultTy.getDtype())`
We should prefer functional style as the method style is deprecated https://github.com/llvm/mlir-www/blob/main/website/content/deprecation/_index.md#deprecated

qingyunqu: Done

penguin-wwy (approved):
LGTM

PR #3221, opened 2024-04-24
[ONNX] Fix Onnx.Selu lowering and canonicalizer for IntImplicit op

vivekkhandelwal1:


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -847,15 +847,19 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
       "Selu", 6, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        // y = gamma * (alpha * e^x - alpha) for x <= 0, y = gamma * x for x > 0
         Torch::ValueTensorType resultType;
         float alpha, gamma;
         Value operand;
         if (binder.tensorOperand(operand) ||
-            binder.f32FloatAttr(alpha, "alpha") ||
-            binder.f32FloatAttr(gamma, "gamma") ||
+            binder.f32FloatAttr(alpha, "alpha", 1.67326) ||
+            binder.f32FloatAttr(gamma, "gamma", 1.0507) ||
```
renxida, line 856: would be nice to have a name / comment explaining why these particular vales

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -864,12 +868,31 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             binder.getLoc(), rewriter.getType<Torch::FloatType>(),
             rewriter.getFloatAttr(rewriter.getF64Type(), gamma));
 
-        Value vInputScale = rewriter.create<Torch::ConstantFloatOp>(
+        Value cstOne = rewriter.create<Torch::ConstantFloatOp>(
             binder.getLoc(), rewriter.getType<Torch::FloatType>(),
             rewriter.getFloatAttr(rewriter.getF64Type(), 1.0));
 
-        rewriter.replaceOpWithNewOp<Torch::AtenEluOp>(
```
renxida, line 871: what's wrong with just AtenEluOp and relying on DecomposeComplexOps to do the following 20 ish lines? does it cause a failure sohewhere?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: They come from the official lowering: https://onnx.ai/onnx/operators/onnx__Selu.html

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: AtenEluOp is not an equivalent of Selu Op.

renxida: yes, but it looks like we can just add a scaling factor to Elu to get Selu. This way we have a little less code duplication.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
renxida: ah gotcha. may i suggest we link that in a comment?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: I have already tried that implementation but it's not working maybe because of the way Elu is lowered in Torch-MLIR.

renxida: ah gotcha. thanks for letting me know.

renxida (approved):

rsuderman:
@vivekkhandelwal1 We should be reverting this. When there is an equivalent operation in the `torch` dialect we should be relying on it. If the `torch` implementation is failing the correct choice is to fix it not avoid it. See similar comments about asinh

https://github.com/llvm/torch-mlir/pull/3271

PR #3224, opened 2024-04-24
Add `.yamllint` and disable some annoying recurring warnings on every pr

renxida:


.yamllint
```diff
@@ -0,0 +1,9 @@
+---
+
+extends: default
+
+rules:
```
ScottTodd, line 5: I still see a few annotations like `[comments] too few spaces before comment` and `[comments-indentation] comment not indented like content` still

We have some other rules in the IREE repo that could help there:
https://github.com/iree-org/iree/blob/2a3a2f721b0353d7fc11ba23cbfde8fa4a875e7a/.yamllint.yml#L19-L24

ScottTodd (approved):

PR #3228, opened 2024-04-25
[onnx] Update DefaultDomainGtoP.cpp gridsampler

afalkenberg1:
Gridsampler
In onnx the interpolation mode is called 'linear' whereas in pytorch it is called 'bilinear'. This led to the problem that everything other than 'bilinear' was rejected. It needed to be changed to linear.


jinchen62 (approved):

afalkenberg1:
why is no runner picking up this job ?

jinchen62 (approved):

PR #3232, opened 2024-04-25
[MLIR][ONNX] Add OnnxToTorch support for GlobalMaxPool Op

NeverRaR:
https://github.com/nod-ai/SHARK-Turbine/issues/658

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -7,6 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/Support/LogicalResult.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Utils.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
@@ -1303,33 +1304,28 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
             binder.getLoc(),
             Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
             outputSize);
-        Value outputTensor;
+
         Type returnIndicesType = torch::Torch::ValueTensorType::get(
             binder.op->getContext(), resultShape,
             rewriter.getIntegerType(64, true));
         if (inputRank == 3) {
-          outputTensor = rewriter
-                             .create<Torch::AtenAdaptiveMaxPool1dOp>(
-                                 binder.getLoc(), resultType, returnIndicesType,
-                                 operand, outputSizeList)
-                             .getResult(0);
+          rewriter.replaceOpWithNewOp<Torch::AtenAdaptiveMaxPool1dOp>(
+              binder.op, resultType, returnIndicesType, operand,
+              outputSizeList);
+          return success();
         } else if (inputRank == 4) {
-          outputTensor = rewriter
-                             .create<Torch::AtenAdaptiveMaxPool2dOp>(
-                                 binder.getLoc(), resultType, returnIndicesType,
-                                 operand, outputSizeList)
-                             .getResult(0);
+          rewriter.replaceOpWithNewOp<Torch::AtenAdaptiveMaxPool2dOp>(
+              binder.op, resultType, returnIndicesType, operand,
+              outputSizeList);
+          return success();
         } else if (inputRank == 5) {
-          outputTensor = rewriter
-                             .create<Torch::AtenAdaptiveMaxPool3dOp>(
-                                 binder.getLoc(), resultType, returnIndicesType,
-                                 operand, outputSizeList)
-                             .getResult(0);
-        } else {
-          return failure();
+          rewriter.replaceOpWithNewOp<Torch::AtenAdaptiveMaxPool3dOp>(
+              binder.op, resultType, returnIndicesType, operand,
+              outputSizeList);
+          return success();
         }
-        rewriter.replaceOp(binder.op, outputTensor);
-        return success();
+
+        return failure();
       });
   patterns.onOp(
       "LayerNormalization", 17,
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1324,10 +1324,12 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         Value cstCeilMode =
             rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), false);
 
-        if (inputRank == 3)
-          return rewriter.notifyMatchFailure(binder.op,
-                                             "Unimplemented: AtenMaxPool1dOp");
-        else if (inputRank == 4) {
+        if (inputRank == 3) {
+          rewriter.replaceOpWithNewOp<Torch::AtenMaxPool1dOp>(
+              binder.op, resultType, operand, kernelSizeList, stridesList,
+              paddingList, dilationsList, cstCeilMode);
+          return success();
+        } else if (inputRank == 4) {
           rewriter.replaceOpWithNewOp<Torch::AtenMaxPool2dOp>(
               binder.op, resultType, operand, kernelSizeList, stridesList,
               paddingList, dilationsList, cstCeilMode);
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1285,45 +1285,59 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
           return rewriter.notifyMatchFailure(
               binder.op, "Expected result type having sizes");
         }
-        ArrayRef<int64_t> resultShape = resultType.getSizes();
-        unsigned resultRank = resultShape.size();
-
-        SmallVector<Value> outputSize;
+        SmallVector<Value> cstKernel, cstPadding, cstStrides, cstDilations;
+        Value cstZero = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getI64IntegerAttr(0));
         Value cstOne = rewriter.create<Torch::ConstantIntOp>(
             binder.getLoc(), rewriter.getI64IntegerAttr(1));
-        for (unsigned i = 2; i < resultRank; i++) {
-          if (resultShape[i] == Torch::kUnknownSize) {
-            outputSize.push_back(cstOne);
+        for (unsigned i = 2; i < inputRank; i++) {
+          if (inputShape[i] == Torch::kUnknownSize) {
+            Value dim = rewriter.create<Torch::ConstantIntOp>(
+                binder.getLoc(), rewriter.getI64IntegerAttr(i));
+            Value inputDimSize = rewriter.create<Torch::AtenSizeIntOp>(
+                binder.getLoc(), operand, dim);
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -628,42 +628,39 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
                 });
-  patterns.onOp("Not", 1,
-                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-                  Torch::ValueTensorType resultType;
-                  Value operand;
-                  if (binder.tensorOperand(operand) ||
-                      binder.tensorResultType(resultType)) {
-                    return failure();
-                  }
+  patterns.onOp(
```
renxida, line 631: why this change?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1205,6 +1202,72 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected input type having sizes");
+        }
+        ArrayRef<int64_t> inputShape = inputTensorType.getSizes();
+        unsigned inputRank = inputShape.size();
+        if (!resultType || !resultType.hasSizes()) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected result type having sizes");
+        }
+        ArrayRef<int64_t> resultShape = resultType.getSizes();
```
renxida, line 1225: @rsuderman is this one of the situations where we should do a SmallVector rather than an ArrayRef?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -628,42 +628,39 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+  patterns.onOp(
+      "Not", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value operand;
+        if (binder.tensorOperand(operand) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
 
-                  auto loc = binder.getLoc();
```
renxida, line 640: ditto. why does the onnx.Not lowering need to be touched at all?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1205,6 +1202,72 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            binder.getLoc(),
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+            outputSize);
+        Value outputTensor;
+        Type returnIndicesType = torch::Torch::ValueTensorType::get(
+            binder.op->getContext(), resultShape,
+            rewriter.getIntegerType(64, true));
+        if (inputRank == 3) {
+          outputTensor = rewriter
+                             .create<Torch::AtenAdaptiveMaxPool1dOp>(
```
renxida, line 1249: use rewriter.replaceOpWithNewOp instead.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1205,6 +1202,72 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        } else if (inputRank == 5) {
+          outputTensor = rewriter
+                             .create<Torch::AtenAdaptiveMaxPool3dOp>(
+                                 binder.getLoc(), resultType, returnIndicesType,
+                                 operand, outputSizeList)
+                             .getResult(0);
+        } else {
+          return failure();
+        }
+        rewriter.replaceOp(binder.op, outputTensor);
```
renxida, line 1268: wouldn't be necessary if we use rewriter.replaceOpWithNewOp instead

renxida (changes requested):

renxida:
Hello! Welcome!

Nice PR & much appreciated! Left some comments for code quality. I hope you find them helpful.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1205,6 +1202,72 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        }
+        Value outputSizeList = rewriter.create<Torch::PrimListConstructOp>(
+            binder.getLoc(),
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+            outputSize);
+        Value outputTensor;
+        Type returnIndicesType = torch::Torch::ValueTensorType::get(
+            binder.op->getContext(), resultShape,
+            rewriter.getIntegerType(64, true));
+        if (inputRank == 3) {
```
renxida, line 1247: instead of running through the if statement and then returning, use early exits and just return failure at the end of the function

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
rsuderman: Nah this is fine. `resultType` has an array allocated in it that has the reference returned. You only need to be concerned about using `ArrayRef` in situations like `ArrayRef<int64_t> resultShape = {dim0, dim1};`. This is becase we rely on a list initializer that goes out of memory after the constructor.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -7,6 +7,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/Support/LogicalResult.h"
```
vivekkhandelwal1, line 10: This is probably not required.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -628,42 +629,39 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          auto i1ty = rewriter.getI1Type();
+          auto ty = rewriter.getType<Torch::ValueTensorType>(
+              operandTy.getSizes(), i1ty);
+          auto torchqTy = Torch::getScalarTypeForType(i1ty);
+          Value tyConst = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getType<Torch::IntType>(),
+              rewriter.getIntegerAttr(rewriter.getIntegerType(64),
+                                      static_cast<int64_t>(torchqTy)));
+          Value none = rewriter.create<Torch::ConstantNoneOp>(loc);
+          Value cstFalse = rewriter.create<Torch::ConstantBoolOp>(loc, false);
```
vivekkhandelwal1, line 655: If all of this is done by clang-format, then it's strange that it was not caught by CI before.

vivekkhandelwal1 (changes requested):
Hi @NeverRaR, I see many things done wrongly in this lowering. Please correct them.

- First of all, you should not be lowering this op to AdaptiveMaxPooling, instead use "AtenMaxPool<n>dOp". 
- Your lowering returns 2 results while the original Onnx op expects 1 result. 
- The logic for the computation of output shape is also wrong, we can't just assume any dynamic dim to be 1.

NeverRaR:
> Hi @NeverRaR, I see many things done wrongly in this lowering. Please correct them.
> 
> * First of all, you should not be lowering this op to AdaptiveMaxPooling, instead use "AtenMaxPooldOp".
> * Your lowering returns 2 results while the original Onnx op expects 1 result.
> * The logic for the computation of output shape is also wrong, we can't just assume any dynamic dim to be 1.
@vivekkhandelwal1 
*The reason for using AdaptiveMaxPooling is that the current torch-mlir does not support AtenAdaptiveMaxPool1d, and my model happens to require this situation.
*Directly replacing onnx.GlobalMaxPool with AdaptiveMaxPool will still use the result it needs (the first result of AtenAdaptiveMaxPool)
*And why it is  assumed that all dynamic dims are 1. Actually, according to the [onnx specification](https://onnx.ai/onnx/operators/onnx__GlobalMaxPool.html):
>Output data tensor from pooling across the input tensor The output tensor has the same rank as the input The first two dimensions of output shape are the same as the input (N x C), while the other dimensions are all 1

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
NeverRaR: It seems that Lint Check has not been run by CI?

vivekkhandelwal1:
> * The reason for using AdaptiveMaxPooling is that the current torch-mlir does not support AtenAdaptiveMaxPool1d, and my model happens to require this situation.

If that's the case, you could have added a lowering for AtenMaxPool1dOp.

> * Directly replacing onnx.GlobalMaxPool with AdaptiveMaxPool will still use the result it needs (the first result of AtenAdaptiveMaxPool)
No, that's not true. You can't replace an op with one result with an op with different number of results. See the crash in the CI for the same.
```
  torch-mlir-opt: /_work/torch-mlir/torch-mlir/externals/llvm-project/mlir/lib/Transforms/Utils/DialectConversion.cpp:1645: virtual void mlir::ConversionPatternRewriter::replaceOp(Operation *, ValueRange): Assertion `op->getNumResults() == newValues.size() && "incorrect # of replacement values"' failed.
```

> * And why it is  assumed that all dynamic dims are 1. Actually, according to the [onnx specification](https://onnx.ai/onnx/operators/onnx__GlobalMaxPool.html):
> 
> > Output data tensor from pooling across the input tensor The output tensor has the same rank as the input The first two dimensions of output shape are the same as the input (N x C), while the other dimensions are all 1

If that's the case then there's no need to put the check, just assign value 1.

NeverRaR:
> > * The reason for using AdaptiveMaxPooling is that the current torch-mlir does not support AtenAdaptiveMaxPool1d, and my model happens to require this situation.
> 
> If that's the case, you could have added a lowering for AtenMaxPool1dOp.
> 
> > * Directly replacing onnx.GlobalMaxPool with AdaptiveMaxPool will still use the result it needs (the first result of AtenAdaptiveMaxPool)
> 
> No, that's not true. You can't replace an op with one result with an op with different number of results. See the crash in the CI for the same.
> 
> ```
>   torch-mlir-opt: /_work/torch-mlir/torch-mlir/externals/llvm-project/mlir/lib/Transforms/Utils/DialectConversion.cpp:1645: virtual void mlir::ConversionPatternRewriter::replaceOp(Operation *, ValueRange): Assertion `op->getNumResults() == newValues.size() && "incorrect # of replacement values"' failed.
> ```
> 
> > * And why it is  assumed that all dynamic dims are 1. Actually, according to the [onnx specification](https://onnx.ai/onnx/operators/onnx__GlobalMaxPool.html):
> > 
> > > Output data tensor from pooling across the input tensor The output tensor has the same rank as the input The first two dimensions of output shape are the same as the input (N x C), while the other dimensions are all 1
> 
> If that's the case then there's no need to put the check, just assign value 1.

Okay, I think I understand now. All modifications have been completed.I will added a lowering for AtenMaxPool1dOp in next pr.

NeverRaR:
@vivekkhandelwal1 Does the failure of CI seem unrelated to my code?

vivekkhandelwal1:
> @vivekkhandelwal1 Does the failure of CI seem unrelated to my code?

Yes

vivekkhandelwal1 (approved):
LGTM

vivekkhandelwal1:
@NeverRaR, you need approval from other reviewers as well to merge this.

renxida (approved):

PR #3237, opened 2024-04-26
Extract the Python APIs in the pt1 dir back to the root

penguin-wwy:


penguin-wwy:
Peel the Python APIs in the pt1 dir, which without dependency on the JitIR importer, back to the root python dir. Is this what you expected? @stellaraccident 
https://github.com/llvm/torch-mlir/pull/3205#issuecomment-2075427726

stellaraccident:
Thanks. Will find some time to look at this in detail in the next bit (in the middle of some things this minute).

python/torch_mlir/compiler_utils.py
```diff
@@ -0,0 +1,162 @@
...
+    pass
+
+def run_pipeline_with_repro_report(module,
+                                   pipeline: str,
+                                   description: str,
+                                   enable_ir_printing: bool = False):
+    """Runs `pipeline` on `module`, with a nice repro report if it fails."""
+    module_name = get_module_name_for_debug_dump(module)
+    try:
+        original_stderr = sys.stderr
```
stellaraccident, line 36: Nit: Would be better to move this line out of the try, since the finally is restoring it.

python/torch_mlir/compiler_utils.py
```diff
@@ -0,0 +1,162 @@
...
+            """
+        trimmed_message = '\n'.join([m.lstrip() for m in message.split('\n')])
+        raise TorchMlirCompilerError(trimmed_message) from None
+    finally:
+        sys.stderr = original_stderr
+
+
+class OutputType(Enum):
+
+    # This output type consists of `torch` dialect ops that have been converted
```
stellaraccident, line 81: How about we update:

"Output `torch` dialect. When converting from FX, this will be immediately after the import from FX to MLIR. When converting from torchscript, this will come after some cleanup passes which attempt to de-alias, decompose and infer shapes. These should be roughly the same level of abstraction since those steps are done within PyTorch itself when coming directly from Dynamo/FX."

stellaraccident (approved):
Thanks. I know you were just moving code that you didn't write, but I gave it a once over and added a couple of comments. If you've got the time to make those updates, great, otherwise, it is fine.

PR #3239, opened 2024-04-26
[ONNX] Fix Onnx.Hardsigmoid lowering

vivekkhandelwal1:

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -46,7 +46,6 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         Value constAlpha = rewriter.create<Torch::ConstantFloatOp>(
             binder.getLoc(), rewriter.getType<Torch::FloatType>(),
             rewriter.getF64FloatAttr(alpha));
-
         Value constBeta = rewriter.create<Torch::ConstantFloatOp>(
             binder.getLoc(), rewriter.getType<Torch::FloatType>(),
             rewriter.getF64FloatAttr(beta));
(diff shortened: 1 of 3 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -52,17 +52,22 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
-            binder.getLoc(), resultType, tensorOperand, constBeta,
-            /*alpha=*/constAlpha);
+        Value alphaMulX = rewriter.create<Torch::AtenMulScalarOp>(
+            binder.getLoc(), resultType, tensorOperand, constAlpha);
+        Value constOne = rewriter.create<Torch::ConstantFloatOp>(
+            binder.getLoc(), rewriter.getType<Torch::FloatType>(),
+            rewriter.getF64FloatAttr(1.0));
+        Value alphaMulXPlusBeta = rewriter.create<Torch::AtenAddScalarOp>(
+            binder.getLoc(), resultType, alphaMulX, constBeta,
+            /*alpha=*/constOne);
```
zjgarvey, line 62: AtenAddScalarOp has the input signature (tensor, float, float) and performs result = operand[0] + operand[1]*operand[2], correct? I feel like this op would be way more useful if it multiplied the tensor operand[0] with operand[2].

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -52,17 +52,22 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
             rewriter.getF64FloatAttr(beta));
 
         // Expression: alpha * x + beta
-        Value alpha_x_plus_beta = rewriter.create<Torch::AtenAddScalarOp>(
-            binder.getLoc(), resultType, tensorOperand, constBeta,
-            /*alpha=*/constAlpha);
+        Value alphaMulX = rewriter.create<Torch::AtenMulScalarOp>(
+            binder.getLoc(), resultType, tensorOperand, constAlpha);
+        Value constOne = rewriter.create<Torch::ConstantFloatOp>(
```
zjgarvey, line 57: There is an aten minimum op that compares the result of the linear expression with a constant oneTensor a few lines below this, and for some reason it is using a constant int one instead of a constant float one. I'm sure it will get cast to a float at some point, but maybe this could just be cleaned up here to use your constOne.

If you make a change to use constOne for the oneTensor, it might also make sense just to change constantZero below to a float as well.

zjgarvey (commented):
Thanks!

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
renxida: i wonder if this is just to be consistent with the aten.add.Tensor op

where you get tensor1+tensor2*alpha

renxida: but also this leave the question of why we have the alpha in the first place ... for quantization maybe?

renxida (approved):

renxida (approved):

zjgarvey (approved):

PR #3240, opened 2024-04-26
[TorchToLinalg] Adds Quantization Support for ConvTranspose

zjgarvey:
I spent a little while debugging numerics issues with some tests similar to the ones in quantized_models.py, only to find that pytorch's quantized conv transpose is catastrophically inaccurate. I'll upstream the issue and only leave the tests here which are of the form quantize -> dequantize -> op.


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -797,6 +798,7 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
     auto resultTy = cast<ValueTensorType>(op.getType());
 
     Value inputZp, weightZp;
+    bool inputUnsigned, weightUnsigned = false;
```
rsuderman, line 801: Are you expect to initialize both of these to `false`? I usually recommend to do the following for clarity
```
bool inputUnsigned = false;
bool weightUnsigned = false;
```

rsuderman (approved):

PR #3251, opened 2024-04-28
[Torch] decompose AtenLerpTensorOp

Xinyu302:
as title


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2488,6 +2488,35 @@ class DecomposeAtenLerpScalarOp : public OpRewritePattern<AtenLerpScalarOp> {
...
+        rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
+    auto start = op.getSelf();
+    auto inputType = cast<BaseTensorType>(start.getType());
+
+    auto delta = rewriter.create<AtenSubTensorOp>(loc, inputType, op.getEnd(),
+                                                  start, cstOne);
+
+    auto weightedDelta =
+        rewriter.create<AtenMulTensorOp>(loc, inputType, delta, op.getWeight());
+    auto lerp = rewriter.create<AtenAddTensorOp>(loc, inputType, start,
```
qingyunqu, line 2512: `inputType` => `resType`: seems that `inputType` may have shape and `resType` may not have shape.

qingyunqu (approved):

PR #3253, opened 2024-04-29
Fix onnx sinh lowering

jinchen62:
iree tests `test_sinh` and `test_sinh_example` passed


vivekkhandelwal1 (approved):

rsuderman:
@renxida We should revert this and instead fix the `math` dialects lowerings for hyperbolic functions, we should not be decomposing these during the lowering to `torch`. There are better numeric methods for evaluating `sinh`, `cosh`, etc than doing the explicit decomposition which is why the `math` dialect exists.

PR #3271, opened 2024-05-01
Revert hyperbolic trigonometric decompositions

rsuderman:
We should be using the `torch` path and handling decomposition in the `math` dialect.


rsuderman:
You can see these kind of decompositions live downstream in `math` and `linalg`
https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Math/Transforms/ExpandPatterns.cpp
https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Math/Transforms/PolynomialApproximation.cpp

jinchen62 (approved):
Was about to revert these once https://github.com/llvm/llvm-project/pull/90718 is landed. Thanks Rob.

PR #3272, opened 2024-05-01
Folder and Canonicalizer for PrimsConvertElementTypeOp and AtenMaxPool2dWithIndicesOp

zezhang:
While playing with TorchDynamo on ResNet18. I notice following issues:

-  `prims.convert_element_type` can’t  be canonicalized even if the input and the output share the same type

- `aten.max_pool2d_with_indices` is always used instead of `aten.max_pool2d`, even if the second returned output (indices) has no user

This PR fixes above issues by adding a folder to the PrimsConvertElementTypeOp and a canonicalizer to the AtenMaxPool2dWithIndicesOp

Lit test:

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -4716,6 +4716,38 @@ LogicalResult AtenPermuteOp::verify() {
 }
 
 //===----------------------------------------------------------------------===//
+// PrimsConvertElementTypeOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult PrimsConvertElementTypeOp::fold(FoldAdaptor adaptor) {
+  if (getA().getType() == getResult().getType())
+    return getA();
+}
+
+//===----------------------------------------------------------------------===//
+// AtenMaxPool2dWithIndicesOp
+//===----------------------------------------------------------------------===//
+
+void AtenMaxPool2dWithIndicesOp::getCanonicalizationPatterns(
+    RewritePatternSet &patterns, MLIRContext *context) {
+  patterns.add(+[](AtenMaxPool2dWithIndicesOp op, PatternRewriter &rewriter) {
+    if (!op.getResult1().use_empty()) {
+      return rewriter.notifyMatchFailure(
+          op, "result1 of MaxPool2dWithIndices should be unused");
+    }
+
+    Value result = rewriter.create<Torch::AtenMaxPool2dOp>(
+        op->getLoc(), op.getResult0().getType(), op.getSelf(),
+        op.getKernelSize(), op.getStride(), op.getPadding(), op.getDilation(),
+        op.getCeilMode());
+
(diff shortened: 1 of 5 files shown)


zezhang:
Looks like adding a folder for `PrimsConvertElementTypeOp` makes the linalg `PrimsConvertElementTypeModule_basic` test fail. Because the input and output type could be a simple `torch.tensor` (no element dtype is available). We will have to use the `dtype` argument of the op to specify the output type. Any idea on this? @sjain-stanford

sjain-stanford:
> Because the input and output type could be a simple torch.tensor (no element dtype is available). We will have to use the dtype argument of the op to specify the output type. 

I think you want to prevent it from being folded when element dtype is unknown. I haven't validated it but something like this should work:

```c++
OpFoldResult PrimsConvertElementTypeOp::fold(FoldAdaptor adaptor) {
  auto inputType = cast<BaseTensorType>(getA().getType());
  auto outputType = cast<BaseTensorType>(getResult().getType());
  if (inputType != outputType)
    return nullptr;
  if (!inputType.hasDtype() || !outputType.hasDtype())
    return nullptr;
  if (inputType.getDtype() != outputType.getDtype())
    return nullptr;
  return getA();
}
```

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4715,6 +4715,41 @@ LogicalResult AtenPermuteOp::verify() {
...
+  if (inputType.getDtype() == outputType.getDtype())
+    return getA();
+  return nullptr;
+}
+
+//===----------------------------------------------------------------------===//
+// AtenMaxPool2dWithIndicesOp
+//===----------------------------------------------------------------------===//
+
+void AtenMaxPool2dWithIndicesOp::getCanonicalizationPatterns(
```
sjain-stanford, line 4734: please add lit tests for the canonicalizer

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4715,6 +4715,41 @@ LogicalResult AtenPermuteOp::verify() {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// PrimsConvertElementTypeOp
+//===----------------------------------------------------------------------===//
+
+OpFoldResult PrimsConvertElementTypeOp::fold(FoldAdaptor adaptor) {
```
sjain-stanford, line 4722: please add lit test for the folder

sjain-stanford:
Good news is your PR now lets 4 more tests pass, which previously failed. You may update the xfail set.
```
  ****** Unexpectedly Passed tests - 4 tests
      XPASS - "MaxPool2dEmptyStrideStaticModule_basic"
      XPASS - "MaxPool2dStaticCeilModeTrueModule_basic"
      XPASS - "MaxPool2dStaticModule_basic"
      XPASS - "ResNet18StaticModule_basic"
```

sjain-stanford (approved):
Feel free to land once CI is clean.

PR #3280, opened 2024-05-02
[ONNX] Enables data propogation for onnx shape inference

zjgarvey:
This small change seems to dramatically improve shape inference for complex models, and consequently, improves onnx importer reliability.


python/torch_mlir/tools/import_onnx/__main__.py
```diff
@@ -75,6 +75,7 @@ def load_onnx_model(args: argparse.Namespace) -> onnx.ModelProto:
 
     # Load the model, with possible external data coming from the default
     # location, or the location specified on the conmand line.
+    data_prop = False if args.no_data_prop else True
```
rsuderman, line 78: `data_prop = not args.no_data_prop`. Though you really should just name it `data_prop` and change the default

python/torch_mlir/tools/import_onnx/__main__.py
```diff
@@ -138,6 +143,11 @@ def parse_arguments(argv=None) -> argparse.Namespace:
         action="store_true",
         help="Disable verification prior to printing",
     )
+    parser.add_argument(
+        "--no-data-prop",
```
rsuderman, line 147: Typically you want flags to be positive. E.g. `--data-prop` with a value of true, then people can set to false

rsuderman (changes requested):

python/torch_mlir/tools/import_onnx/__main__.py
zjgarvey: Yeah, I was wondering about that. I just figured if data prop was default it would be easier to pass --no-data-prop than --data-prop=False. I'll make the change.

python/torch_mlir/tools/import_onnx/__main__.py
zjgarvey: Yeah, that makes sense.

python/torch_mlir/tools/import_onnx/__main__.py
zjgarvey: I figure the cleanest option is to use argparse.BooleanOptionalAction, which provides both --data-prop and --no-data-prop with default value of True for data prop. Let me know what you think.

rsuderman (approved):

rsuderman:
Looks like the Quantized_MLP test is still being flaky. We may want to disable it just for the CI

PR #3279, opened 2024-05-02
Bump llvm version to 97069a86193a617a9e4cf742a29db6116b2bf449

jpienaar:

PR #3282, opened 2024-05-03
[torch-mlir] Bump LLVM to llvm/llvm-project@fc83eda4

aartbik:
(1) this brings in all "warnings" introduced in https://github.com/llvm/llvm-project/pull/90413 (to be fixed by community)
(2) all errors due to new tosa int_div have been fixed
(3) all errors due to tensor output shape have been fixed

### DIFF
--- a/include/torch-mlir/Conversion/TorchOnnxToTorch/Utils.h
+++ b/include/torch-mlir/Conversion/TorchOnnxToTorch/Utils.h
@@ -65,8 +65,8 @@ struct onnx_list_of_constant_ints_op_binder {
       return false;
 
     if (DenseResourceElementsAttr attr =
-            constOp->getAttr("torch.onnx.value")
-                .dyn_cast_or_null<DenseResourceElementsAttr>()) {
+        dyn_cast_or_null<DenseResourceElementsAttr>(
+            constOp->getAttr("torch.onnx.value"))) {
       // Bytes are stored in little endian order. Big endian support will
       // require swizzling.
       if (!Endian::little) {
--- a/include/torch-mlir/Conversion/TorchOnnxToTorch/Utils.h
+++ b/include/torch-mlir/Conversion/TorchOnnxToTorch/Utils.h
@@ -65,8 +65,8 @@ struct onnx_list_of_constant_ints_op_binder {
       return false;
 
     if (DenseResourceElementsAttr attr =
-        dyn_cast_or_null<DenseResourceElementsAttr>(
-            constOp->getAttr("torch.onnx.value"))) {
+            dyn_cast_or_null<DenseResourceElementsAttr>(
+                constOp->getAttr("torch.onnx.value"))) {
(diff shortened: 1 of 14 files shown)


aartbik:
arghhh, it looks like that (at least some of) those warnings are turned into errors....
so we need to fix those first....

aartbik:
I fixed all the real errors, and fixed a lot of deprecated warning = error in most header files; fixing this in all cpp files seems a bit too daunting right now (and I cannot come up with an easy find/sed solution for this either). So I propose to have the warning != error for a bit, so this can go in, and fixing it can be done in a more manageable way.

bjacob:
I'm surprised that this PR isn't running into the same error as @jpienaar 's #3279 ,
https://github.com/llvm/torch-mlir/actions/runs/8930048121/job/24533271831?pr=3279#step:8:1358

```
  python: /_work/torch-mlir/torch-mlir/externals/llvm-project/mlir/lib/Dialect/Tensor/IR/TensorOps.cpp:1681: static void mlir::tensor::ExpandShapeOp::build(OpBuilder &, OperationState &, Type, Value, ArrayRef<ReassociationIndices>): Assertion `succeeded(outputShape) && "unable to infer output shape"' failed.
```

Maybe it's just because the current warnings-as-errors stop compilation before getting there?

In case you do run into this here, see the explanation: https://github.com/llvm/torch-mlir/pull/3279#issuecomment-2097246585

aartbik:
> I'm surprised that this PR isn't running into the same error as @jpienaar 's #3279 ,

Ah. very well possible that we never reach that point. I did fix some output checks with output shape (see e.g. flatten.mlir), so it definitely is at the same problematic update of MLIR.

Anyway, I am a bit stuck with this bump and could use some help (fixing the deprecations, or temporarily allowing me to remove the warning=error setting [which probably will run into jpienaar's issue first then])

jpienaar:
I think it's great you fixed the deprecated warnings. We could land those even without a llvm rev bump and so make the rev bump simpler.

There is a lowering TOSA to Tensor side that is shown by one of the PyTorch tests, but could be a upstream change. I don't think torch-mlir bumps allow for cherry picks LLVM side, so it's probably up and through ;-)

bjacob:
Started a new integrate https://github.com/llvm/torch-mlir/pull/3300, not yet wrapping the fixes from this PR but maybe we'll need to carry over some of the things you did here. For the deprecated declaration though, this is just disabling the warning.

aartbik:
> Started a new integrate #3300, not yet wrapping the fixes from this PR but maybe we'll need to carry over some of the things you did here. For the deprecated declaration though, this is just disabling the warning.

Great, I will wait for that and see if we still want to salvage this PR (or start a new one)...

aartbik:
Looks like Benoit is already working on a much more recent sha, so closing this PR.
Thanks Benoit!

PR #3283, opened 2024-05-03
[torch] Add OnnxToTorch lowering for `onnx.HammingWindow`

vinayakdsci:
Adds OnnxToTorch lowering for the `onnx.HammingWindow` op.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -2349,4 +2349,35 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
 
         return success();
       });
+
+  patterns.onOp(
+      "HammingWindow", 17,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Value size;
+        Torch::ValueTensorType resultType;
+        int64_t periodic, output_datatype;
+        if (binder.tensorOperand(size) ||
+            binder.s64IntegerAttr(output_datatype, "output_datatype", 1) ||
+            binder.s64IntegerAttr(periodic, "periodic", 1) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+        Value a0 = rewriter.create<Torch::ConstantFloatOp>(
+            binder.getLoc(),
+            rewriter.getFloatAttr(rewriter.getF64Type(), 0.543478));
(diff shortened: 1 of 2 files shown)


vinayakdsci:
cc @renxida.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -2349,4 +2349,35 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        int64_t periodic, output_datatype;
+        if (binder.tensorOperand(size) ||
+            binder.s64IntegerAttr(output_datatype, "output_datatype", 1) ||
+            binder.s64IntegerAttr(periodic, "periodic", 1) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+        Value a0 = rewriter.create<Torch::ConstantFloatOp>(
+            binder.getLoc(),
+            rewriter.getFloatAttr(rewriter.getF64Type(), 0.543478));
```
renxida, line 2367: I've seen @rsuderman use a constant variable at the top of the function to cache binder.getLoc and rewriter.getF646ype. but for something as simple as this, it might be easier to do it your way. i'll leave this a tradeoff for you to decide based on readability.

renxida (approved):
very nice. feel free to merge after merging numerical test.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
vinayakdsci: @renxida Thanks for the approval! I think that the values are repeated only thrice, so I am not sure if it would benefit as much from caching the values, but I would be happy to change the code if you do think that it might seriously affect readability.

Also, I did write the numerical test -> https://github.com/nod-ai/SHARK-TestSuite/pull/222 .
But it requires your approval before merging. I hope you could have a look at that too.

Thanks!

vinayakdsci: No worries, I stored binder.getLoc() in a variable, and switched from getFloatAttr() to getF64FloatAttr(). I think it is much better now, thanks for the suggestion!

vivekkhandelwal1:
@vinayakdsci, it would be helpful if you can leave a comment on https://github.com/llvm/torch-mlir/pull/3283 when you start working on an op so that no two person works on the same op.

renxida:
https://github.com/nod-ai/SHARK-TestSuite/pull/222

Tested here

PR #3287, opened 2024-05-06
[Torch]Emit AtenDetach_Op

Xinyu302:
as title


vivekkhandelwal1 (commented):
@Xinyu302 What do we need this for? Also, it's better if we add an op with shape and dtype information, even if we don't lower it.

PR #3288, opened 2024-05-06
[FxImporter] Add backend lowering to Fx API

penguin-wwy:


python/torch_mlir/fx.py
```diff
@@ -16,18 +16,58 @@
...
+        + " extra-library="
+        + extra_library_file_name
+        + "}"
+    )
+    run_pipeline_with_repro_report(
+        torch_mod,
+        f"builtin.module(torch-function-to-torch-backend-pipeline{option_string})",
+        "Lowering TorchFX IR -> Torch Backend IR",
+        enable_ir_printing=verbose,
+    )
```
penguin-wwy, line 55: To digress a bit. Although the torch dialect module, which is directly from Dynamo/FX, theoretically does not require steps like decomposition and shape inference, there might be a situation where we decompose an operator into one that needs shape inference in the DecomposeComplexOpsPass, and this decomposed operator is not in dynamo's decomposition list. So, we can't temporarily strip off these passes, such as TorchShapeRefinement, in order to pass the corresponding case.
@stellaraccident

stellaraccident: I'm fine adding this stuff for now. We can evaluate/change course once we have better test coverage. Hard to do anything without that.

stellaraccident (approved):

PR #3292, opened 2024-05-06
[Stablehlo] enable Stabelhlo refbackend with Interpreter

qingyunqu:


stellaraccident:
FYI - This patch is causing some kind of crash/timeout of the integration test harness, which gets the CI into a bad state. I'm attempting to cancel it to move on.

PR #3297, opened 2024-05-07
[torch-mlir][sparse] add simple sparsity "propagation" rules

aartbik:
While waiting for the full resolution of feature request https://github.com/pytorch/pytorch/issues/117188
(which will propagate sparsity the right way in upstream PyTorch for all FX Graphs), this minor change allows us to start testing sparsity "within" a network, rather than just the parameters. Feel free to add your own rules for testing (but within reason for what will be done upstream).

Note, two TODOs need to be addressed to work around some pending issues to make the JIT execution work.


PeimingLiu (approved):

aartbik:
The fail is on head (not related to this), but let's wait to see if it gets fixed so I can merge it first....

Issue #3290, opened 2024-05-06
onnx.Transpose version 1 not lowered (but it's essentially the same as version 13)


mgehre-amd:
We support onnx.Transpose version 13, but not version 1, even though their difference is minimal: https://onnx.ai/onnx/operators/text_diff_Transpose_1_13.html#l-onnx-op-transpose-d1-13

Could we please also support version 1?
This is relevant for older models, e.g. for running the ONNX conformance suite.

PR #3302, opened 2024-05-07
[ONNX] Reduces Transpose Opset Version

zjgarvey:
As mentioned in issue #3290 , the difference between onnx.Transpose in versions 1 and 13 is minimal, and therefore should be supported with the same conversion pattern.


mgehre-amd:
Thanks!

vivekkhandelwal1 (approved):

PR #3306, opened 2024-05-09
[Linalg] [Stablehlo] Promote type for compare scalar op

penguin-wwy:


penguin-wwy:
@stellaraccident This PR also requires a review

PR #3309, opened 2024-05-09
[FxImporter] Eliminate the dependency on the refinement pass

penguin-wwy:


penguin-wwy:
We can make the FxImporter eliminate its dependency on the shape and dtype refinement pass. Only need to decompose three ops to pass e2e test, one of which depends on #3306.
@stellaraccident

stellaraccident (approved):

PR #3311, opened 2024-05-09
[torch-mlir][sparse] minor tweaks in sparse tests

aartbik:
(1) test full pytorch output for eltwise
(2) use "random" input for LIF, to get general sparse tensor 
(3) introduce way to get true sparsity into network (needs backend fix first)

### DIFF
--- a/test/python/fx_importer/sparse_test.py
+++ b/test/python/fx_importer/sparse_test.py
@@ -128,6 +128,10 @@ def sparse_export(
                 node.meta["sparsity"] = SparsityMeta(
                     torch.sparse_coo, 0, dim, 0, None, torch.int64, torch.int64
                 )
+            # Uncomment this to hack sparsity into the network.
+            # elif node.name == "_to_dense":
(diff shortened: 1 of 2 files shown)


test/python/fx_importer/sparse_test.py
```diff
@@ -567,9 +575,8 @@ def forward(self, X):
                 mem = mem * (1.0 - spike)
                 spike_pot.append(spike)
             spike_pot = torch.stack(spike_pot, dim=-1)
-            # TODO: we would like to see something like
-            #       return spike_pot.to_sparse()
-            return spike_pot
+            # Hack to get to_sparse() through propagation.
+            return spike_pot.to_sparse().to_dense()
```
PeimingLiu, line 579: can we put this before `torch.stack`?

aartbik: Oh, sure, I keep forgetting what way is best. Let me change.

PeimingLiu (approved):

Issue #3310, opened 2024-05-09
Update version pin to onnx 1.16


stellaraccident:
onnx 1.16 includes support for Python 3.12. We should upgrade. Based on a user report, there may need to be some cleanup.

zjgarvey:
Just to verify, you would like me to

1. pin onnx==1.16.0 in test-requirements.txt
2. check for e2e test regressions with config onnx and fix them
3. setup building nightly python wheels for python version 3.12
4. anything else?

stellaraccident:
That's it.

zjgarvey:
Sounds good. Thanks.

zjgarvey:
Closing since we updated to 1.16.1.

PR #3317, opened 2024-05-09
Updates Onnx Version to 1.16

zjgarvey:
- Some export fails are now moving along to crashes, so I moved those to the appropriate onnx crashing set. 
- new data types need to be supported (uint4/int4)
- uint8 tensor protos now supported


zjgarvey:
Addresses issue #3310

zjgarvey:
@stellaraccident Should I add cp312-cp312 to [build_linux_packages.sh](https://github.com/llvm/torch-mlir/blob/main/build_tools/python_deploy/build_linux_packages.sh) as well? I wasn't sure if I should do that or put in a request to update in [torch-mlir-release](https://github.com/llvm/torch-mlir-release/).

mgehre-amd:
From what I understand, new ONNX versions bring new operator versions. How do we make sure that the new operator versions are not matched by the existing matchers in OnnxToTorch, which are unaware of changes in those new op versions (e.g. additional optional attributes/operands)?
If we don't make them aware, it seems to me that we would silently miscompile newer operator versions.

zjgarvey:
> From what I understand, new ONNX versions bring new operator versions. How do we make sure that the new operator versions are not matched by the existing matchers in OnnxToTorch, which are unaware of changes in those new op versions (e.g. additional optional attributes/operands)?
> If we don't make them aware, it seems to me that we would silently miscompile newer operator versions.

As always, @mgehre-amd, thanks for the insightful comments. That's definitely a valid concern. If it's any reassurance, we are writing conversions based on the onnx documentation (which is currently based on version 1.17) matching the newest opset version signature, and setting the minimum opset version to the oldest version with a compatible signature. So the opsets introduced between 1.15 and 1.16 should hopefully cause minimal impact. 

I think the best long-term solution would be adding a maximum version number to the `onOp` method. I'd be happy to add this functionality, and look through the onnx docs for newer incompatible signatures.

PR #3318, opened 2024-05-09
[sparse] propagate sparsity properly when decompose torch operations.

PeimingLiu:


include/torch-mlir/Dialect/Torch/IR/TorchTypes.h
```diff
@@ -53,6 +53,9 @@ class BaseTensorType : public Type {
   /// convenient API.
   Type getOptionalDtype() const;
 
+  /// Get the raw optional sparse tensor encoding.
+  Attribute getOptionalSparsity() const;
```
yinying-lisa-li, line 57: It could be my personal preference, but I feel like `getOptionalSparseEncoding` sounds clearer to me than `getOptionalSparsity`. When I saw the function call later, I wasn't sure what it was getting.

yinying-lisa-li (approved):

include/torch-mlir/Dialect/Torch/IR/TorchTypes.h
PeimingLiu: I am reusing the existing API so I think it is better to keep the original name.

aartbik (approved):
LGTM and approved, but since we are touching a bit more core torch here, please wait for Stella to approve as well

also, please double check if we need to fix the bazel build (as a courtesy), since we are touching a CMake file

PeimingLiu:
@stellaraccident

ramiro050 (approved):
LGTM!

PR #3319, opened 2024-05-10
[linalg] Implement strict mode lowering for aten.view.

stellaraccident:
* Enables assume_strict_symbolic_shapes on fx_importer imported programs, indicating strict shape semantics.
* Reworks the view->reshape lowering to take advantage of strict mode and do one of:
  * Collapse to 0D
  * Flatten/Unflatten when there is an inferred dim.
  * Fallback to tensor.reshape
* Splits some test cases up and adds an attribute to control the old pattern (so new corners can be tested in strict mode in isolation).
* Dynamic inferred mode needs upstream work to generalize expand_shape (so that case is suppressed here).
* Deletes the assert from the existing tensor.reshape lowering if strict shape mode is enabled (since the condition it is dynamically asserting cannot happen).

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -940,6 +940,9 @@ public:
   LogicalResult
   matchAndRewrite(AtenViewOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
+    if (op->getParentOp()->hasAttr("torch.disable_legacy_view"))
+      return rewriter.notifyMatchFailure(op.getLoc(),
+                                         "legacy view lowering diabled");
     if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
       return failure();
     Location loc = op.getLoc();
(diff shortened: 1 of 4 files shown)


lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -1356,6 +1366,165 @@ class ConvertAtenViewOpToReshape : public OpConversionPattern<AtenViewOp> {
...
+      Value selfProduct = one;
+      for (int i = 0, e = selfTy.getRank(); i < e; ++i) {
+        Value index = rewriter.create<arith::ConstantIndexOp>(loc, i);
+        Value dim = rewriter.create<tensor::DimOp>(loc, self, index);
+        dim = rewriter.create<arith::IndexCastOp>(loc, sizeTy, dim);
+        selfProduct = rewriter.create<arith::MulIOp>(loc, selfProduct, dim);
+      }
+
+      Value inferredSize =
+          rewriter.create<arith::DivSIOp>(loc, selfProduct, sizeProduct);
```
rsuderman, line 1499: Probably a bit pedantic but this should probably be DivUIOp. This could be presented as 32-bit integers and potentially hit the upper bound on those platforms.

rsuderman (approved):

lib/Conversion/TorchToLinalg/DataMovement.cpp
stellaraccident: Not pedantic at all. Actually meant to do that - thanks for the catch.

PR #3327, opened 2024-05-10
Generalize Operand Quantization in FuseQuantizeOps

zjgarvey:
This change enables more customization with operand quantization, and generalizes the patterns QuantizeOperands and QuantizeTransposeOperands to QuantizeOperandsPastCommutingOps. 

This allows for passing quantization through operations which are functionally unaffected by quantization, such as view-like ops. The purpose of this change is to address a myriad of quantization issues seen in quantized onnx models that have some reshape-like operations sandwiched in between a dequant and something like a matmul (whose other operand is immediately quantizable).

If other model failures of this type occur in the future, other authors can simply add the problematic ops to the qCommutingOps function defined here to enable the quantization to pass through. 

For each op, I have set the depth of the search to the largest depth I have seen in practice:

For aten mm op, see [this issue](https://github.com/nod-ai/SHARK-Turbine/issues/556)
For aten matmul op, see [this issue](https://github.com/nod-ai/SHARK-Turbine/issues/666)

If longer patterns exist, the depth can easily be increased.


zjgarvey:
Ah, I see the issue. Debugging currently.

rsuderman (approved):

PR #3328, opened 2024-05-11
[Stablehlo] lowering aten.randn & aten.normal_functional to mhlo.rng …

qingyunqu:
…NORMAL

* split lowering of uniform, randn, normal from Basic.cpp into Rng.cpp


Vremold (approved):
LGTM. Thanks for the effort!

PR #3329, opened 2024-05-11
[Stablehlo] fix aten.randn's lowering with f32 element type

qingyunqu:


Vremold (approved):
LGTM. Thanks for the effort!

PR #3333, opened 2024-05-13
[ONNX] Fix incorrect result type for `onnx.OneHot` operator

angelz913:

PR #3334, opened 2024-05-13
[ONNX] Fix bugs for the `onnx.OneHot` operator

angelz913:
This commit fixes the bugs for the `onnx.OneHot` operator by:

1) Converting negative indices to non-negative indices
2) Handling both `int` and `float` types for `off` and `on` values
3) Using the correct result type

It also includes a new unit test.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1761,23 +1761,32 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
           depth = rewriter.create<Torch::AtenIntScalarOp>(
               loc, rewriter.getType<Torch::IntType>(), depth);
 
-        auto selectTy = rewriter.getType<Torch::ValueTensorType>(
-            llvm::SmallVector<int64_t>{1}, valuesTy.getDtype());
-
+        Type boolTy = rewriter.getType<Torch::ValueTensorType>(
+            indicesTy.getSizes(), rewriter.getI1Type());
         Value zero = rewriter.create<Torch::ConstantIntOp>(
             loc, rewriter.getI64IntegerAttr(0));
         Value one = rewriter.create<Torch::ConstantIntOp>(
             loc, rewriter.getI64IntegerAttr(1));
+        Value lt =
+            rewriter.create<Torch::AtenLtScalarOp>(loc, boolTy, indices, zero);
+        Value add = rewriter.create<Torch::AtenAddScalarOp>(
+            loc, indicesTy, indices, depth, one);
+        indices = rewriter.create<Torch::AtenWhereSelfOp>(loc, indicesTy, lt,
+                                                          add, indices);
+
+        auto selectTy = rewriter.getType<Torch::ValueTensorType>(
+            llvm::SmallVector<int64_t>{1}, valuesTy.getDtype());
+
+        bool valuesAreInt = isa<IntegerType>(valuesTy.getDtype());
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1654,13 +1654,22 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
           depth = rewriter.create<Torch::AtenIntScalarOp>(
               loc, rewriter.getType<Torch::IntType>(), depth);
 
-        auto selectTy = rewriter.getType<Torch::ValueTensorType>(
-            llvm::SmallVector<int64_t>{1}, valuesTy.getDtype());
-
+        // Apply bounds checking on the indices
```
renxida, line 1657: Would be easier to understand if this comment is either more concise or more specific. E.g. I personally would have written something like `// If indices < 0, indices += depth`

another perhaps better option is to remove this comment entirely - - the code is decently self-explanatory

renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1654,13 +1654,22 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        Type boolTy = rewriter.getType<Torch::ValueTensorType>(
+            indicesTy.getSizes(), rewriter.getI1Type());
         Value zero = rewriter.create<Torch::ConstantIntOp>(
-            loc, rewriter.getI64IntegerAttr(0));
+            loc, intTy, rewriter.getI64IntegerAttr(0));
         Value one = rewriter.create<Torch::ConstantIntOp>(
-            loc, rewriter.getI64IntegerAttr(1));
+            loc, intTy, rewriter.getI64IntegerAttr(1));
```
vivekkhandelwal1, line 1663: You don't need to specify the `intTy` here.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1712,7 +1721,7 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
 
         onehotTy = rewriter.getType<Torch::ValueTensorType>(
             onehotShape, resultType.getDtype());
-        onehot = rewriter.create<Torch::AtenWhereScalarOp>(loc, onehotTy,
+        onehot = rewriter.create<Torch::AtenWhereScalarOp>(loc, resultType,
```
vivekkhandelwal1, line 1724: I'm not sure why the `onehotTy` was created, so can't comment on removing it either. It's better if you confirm this with @rsuderman before merging.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
angelz913: This is addressed in PR #3333

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
angelz913: Removed this comment

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
angelz913: Removed `intTy`

vivekkhandelwal1 (approved):
LGTM

PR #3347, opened 2024-05-15
Don't fold `aten.clone` if result isn't same type as input

AaronStGeorge:
Similar to https://github.com/llvm/torch-mlir/pull/2824, we were seeing some assertion failures after the addition checks around folders were tightened up in LLVM: https://github.com/llvm/llvm-project/pull/75887 . This PR essentially moves the logic that used to be applied at the LLVM level into the folder, which seems to be the suggested fix.


qingyunqu (approved):
LGTM

PR #3348, opened 2024-05-15
[Stablehlo] refactor amax, max, max.dim's lowering to stablehlo

qingyunqu:
* not to decompose `aten.amax` on `stablehlo` backend. Because it could be lowering to `stablehlo.reduce` directly.
* lowering `aten.max.dim` to `stablehlo.reduce apply max` when `AtenMaxDimOp.getIndices()` doesn't have users. It's more simple.


Vremold (approved):
LGTM. Thanks for the effort!

PR #3349, opened 2024-05-15
[FxImporter][TOSA] Enable FxImporter to TOSA e2e tests

sjarus:


sjarus:
The e2e tests yield 
Summary:
    Passed: 573
    Expectedly Failed: 743

This is significantly higher than tosa and make_fx_tosa configs: 

tosa
Summary:
    Passed: 433
    Expectedly Failed: 883

make_fx_tosa
Summary:
    Passed: 428
    Expectedly Failed: 888

stellaraccident (approved):
Makes sense to me. I'd really like to remove at least one of the other tosa test suites from the CI when adding this one. I don't think tosa needs *three* test suites.

sjarus:
Yeah I'd ideally prefer to jettison the make_fx_tosa one now that this is in place. @mgehre-amd had prior feedback on wanting that pathway but perhaps with this one available, it's no longer a concern ?

mgehre-amd:
Until we migrate to `fx_importer_tosa` downstream, we are mainly interested in `make_fx_tosa`; and we would be fine with dropping the plain `tosa` config.

sjain-stanford:
> @sjain-stanford / @navahgar - do you still have a dependency on the plain tosa path, or would fx_importer_tosa be ideal now ?

We're migrating much of our load-bearing pieces (that include TOSA and TCP based flows) to use the TorchDynamo frontend, so keeping the fx_importer_tosa e2e configs seems reasonable to me. We might still send fixes to TOSA based lowerings that depend on the TS frontend, but it may not be all that frequent.

sjarus:
Given that tosa and make_fx_tosa both use the same pathway with the use_make_fx flag distinguishing them, perhaps the best option is to keep make_fx_tosa and fx_importer_tosa short term, and longer term just fx_importer_tosa ? 

If any legacy TS requirement mandates it, there's still just the torchscript target in CI for testing the TS ingress path, so that should be enough ? Arm has an internal mirror of the repo where we'll probably keep all three targets running, but will focus on fx_importer_tosa. 

If all this sounds ok, then we have an immediate answer for @stellaraccident's question on which targets to keep in CI. 

cc: @eric-k256 @stellaraccident

stellaraccident:
Sounds reasonable. As long as we are moving forward at this point. Looking forward a few of months, if this and LTC are the only things keeping the legacy build required, I'll probably ask interested parties to bring their own CI to test those parts. We're not there yet but getting close to having things pried apart to eliminate the torch and llvm backend build dep.

(Not trying to be a jerk but am keeping us on a path to progress... The CI is currently coming from my budget)

PR #3354, opened 2024-05-16
[Stablehlo]index type use i64

Xinyu302:


qingyunqu (approved):
nit: should add testcase for this later.

PR #3358, opened 2024-05-17
[ONNX][TOSA] Adds ONNX to TOSA e2e tests

sjarus:
- Refactors OnnxBackend to be generic and consume any Torch backend.


sjarus:
The e2e tests report the following results: 

Summary:
    Passed: 340
    Expectedly Failed: 976

There's no change in the standard onnx target result, since there's no functional impact.

rsuderman (approved):

PR #3360, opened 2024-05-17
Implement lowering of torch.aten.kthvalue

ptrifunovic98:
Closes [nod-ai/SHARK-Turbine#620](https://github.com/nod-ai/SHARK-Turbine/issues/620)


ptrifunovic98:
This is an implementation of the torch.aten.kthvalue operation. The operation first finds min K values in the input, and then takes the max of those values.
For min K values, the TMTensorTopkOp was added, which is the same as IREE Linalg_ext TopK op (https://iree.dev/reference/mlir-dialects/LinalgExt/#iree_linalg_exttopk-linalgexttopkop), or, rather the MinK version of it. The max value out of those was found using linalg generic ops.

I'm open to suggestions about improving this implementation, or adding clarification for certain parts of it.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -260,6 +260,41 @@ static Value createTMTensorScanOp(
...
+    arith::CmpIPredicate g =
+        isEqual ? arith::CmpIPredicate::sge : arith::CmpIPredicate::sgt;
+    arith::CmpIPredicate l =
+        isEqual ? arith::CmpIPredicate::sle : arith::CmpIPredicate::slt;
+    if (intType.isUnsignedInteger()) {
+      g = isEqual ? arith::CmpIPredicate::uge : arith::CmpIPredicate::ugt;
+      l = isEqual ? arith::CmpIPredicate::ule : arith::CmpIPredicate::ult;
+    }
+    arith::CmpIPredicate predicate = isDescending ? g : l;
+    compareOp = rewriter.create<arith::CmpIOp>(loc, predicate, lhs, rhs);
```
rsuderman, line 281: Just return `compareOp` in each if condition then remove the elses. It provides simpler control flow.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -260,6 +260,41 @@ static Value createTMTensorScanOp(
...
+  } else if (elementType.isa<mlir::FloatType>()) {
+    // Case for using arith::CmpFOp.
+    arith::CmpFPredicate g =
+        isEqual ? arith::CmpFPredicate::OGE : arith::CmpFPredicate::OGT;
+    arith::CmpFPredicate l =
+        isEqual ? arith::CmpFPredicate::OLE : arith::CmpFPredicate::OLT;
+
+    arith::CmpFPredicate predicate = isDescending ? g : l;
+    compareOp = rewriter.create<arith::CmpFOp>(loc, predicate, lhs, rhs);
```
rsuderman, line 290: Ditto above

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1578,6 +1639,459 @@ class ConvertAtenScaledDotProductAttentionOp
...
+             makeShapeTorchCompatible(extractedIdxType.getShape()))) {
+      castedIdxMapExprs.push_back(rewriter.getAffineDimExpr(size.index()));
+    }
+
+    auto castedIdxMaps = AffineMap::inferFromExprList(
+        {castedIdxMapExprs, castedIdxMapExprs}, rewriter.getContext());
+
+    // Linalg generic op for casting topk idx output tensor elements from i32 to
+    // result idx tensor element type.
+    auto castedIdxLinalg = rewriter.create<linalg::GenericOp>(
```
rsuderman, line 2000: Technically you can just simplify this by not bothering with `linalg::GenericOp`. If I remember correctly you can just do `arith::IndexCastOp` on tensors.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -7951,6 +7954,7 @@ class DecomposeComplexOpsPass
     addPatternIfTargetOpIsIllegal<DecomposeAtenMatmulOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenMvOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenLinalgCrossOp>(patterns);
+    // addPatternIfTargetOpIsIllegal<DecomposeAtenKthvalueOp>(patterns);
```
rsuderman, line 7957: Is this intended  to be commented out? Typically we control which are enabled via the pass settings for DecomposeComplexOps.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1622,6 +1622,9 @@ class DecomposeAtenAMinMaxOp : public OpRewritePattern<Torch::AtenAminOp> {
       auto idxTy = rewriter.getType<Torch::ValueTensorType>(
           reductionShape, rewriter.getIntegerType(32, /*is_signed*/ true));
       llvm::SmallVector<Type, 2> types{reductionTy, idxTy};
+
+      reduction.print(llvm::outs());
```
rsuderman, line 1626: Debug code - remove

lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -404,6 +404,7 @@ static void markDecomposedOpsAsIllegal(MLIRContext *context,
   target.addIllegalOp<AtenSelectIntOp>();
   target.addIllegalOp<AtenMvOp>();
   target.addIllegalOp<AtenLinalgCrossOp>();
+  // target.addIllegalOp<AtenKthvalueOp>();
```
rsuderman, line 407: Ditto above

rsuderman (changes requested):

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
ptrifunovic98: Hey @rsuderman, thank for the comments!

I tried using `arith::IndexCastOp` directly on tensors, but I get the following error:

```
error: failed to legalize operation 'arith.index_cast'
note: see current operation: %14 = "arith.index_cast"(%13) : (tensor<3xi32>) -> tensor<3xindex>
error: Module does not conform to the linalg-on-tensors backend contract. See dialect conversion legality information above.
```

Basically, casting from a tensor of i32 element type to a tensor with the same shape and index element type fails. The docs state that the operation can be used for conversion between scalar or vector integers/indexes, but the operation is marked as Tensorizable (as all arith operations are), so it should pass for tensors too. Do you have an idea what could have produced this error?

ptrifunovic98: @rsuderman The problem here is that operations from the `arith` dialect will not be legalized by the `torch-verify-linalg-on-tensors-backend-contract` pass. The implementation of this pass adds `arith` as a legal dialect only for scalar ops:

```
auto isLegalScalarOp = [&](Operation *op) {
      // We recognize basic scalar ops by them having the trait "Elementwise",
      // even though we don't expect them to operate on tensors.
      return scalarConverter.isLegal(op) &&
             op->hasTrait<OpTrait::Elementwise>();
};

...

target.addDynamicallyLegalDialect<arith::ArithDialect>(isLegalScalarOp);
```

When I add the following to the pass:
```
target.addDynamicallyLegalDialect<arith::ArithDialect>(opHasLegalTypes);
```
the error dissapears. `opHasLegalTypes` is used in this pass for every dialect that works on tensors. So I guess the only way to not use linalg.generic is to modify the pass, but I don't know what other side-effects this can have, so I don't think it's the best idea for me to do this.

Is it alright if I stick with linalg.generic here?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
ptrifunovic98: I accidentally left this here from when I tried to implement the Kthvalue lowering as a decomposition. It serves no purpose, so I have removed it.

lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -1570,6 +1634,456 @@ class ConvertAtenScaledDotProductAttentionOp
...
+    // get k, check it is a constant int
+    int64_t k;
+    if (!matchPattern(op.getK(), m_TorchConstantInt(&k)))
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only constant k value is supported");
+
+    // check if element type is float, int, or unsigned
+    bool isUnsigned = false;
+    if (!isa<mlir::FloatType>(inputElementType)) {
+      if (isa<mlir::IntegerType>(inputElementType)) {
```
rsuderman, line 1691: It makes more sense to do
```
if (!isa<mlir::IntegerType>(inputElementType) {
}

// else condition
```

rsuderman (approved):

PR #3361, opened 2024-05-17
[Torch] emit aten.dot and canonicalize it to aten.matmul

qingyunqu:
* canonicalize `aten.dot` to `aten.matmul`


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -548,6 +548,24 @@ void PrimIfOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
 }
 
+//===----------------------------------------------------------------------===//
+// AtenDotOp
+//===----------------------------------------------------------------------===//
+
+void AtenDotOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                            MLIRContext *context) {
+  patterns.add(+[](AtenDotOp op, PatternRewriter &rewriter) {
+    auto ty = dyn_cast<ValueTensorType>(op.getResult().getType());
```
Vremold, line 558: Are these checks necessary?

Vremold (commented):
LGTM. A few comments.

lib/Dialect/Torch/IR/TorchOps.cpp
qingyunqu: not necessary, but more canonical

Issue #3352, opened 2024-05-15
Crash with TorchBackendToLinalgOnTensorsBackendPipeline


zjgarvey:
## The Issue

Running ```--torch-mlir-opt --torch-backend-to-linalg-on-tensors-backend-pipeline``` on some (rather large IR) causes a crash during the pass --resolve-shaped-type-result-dims (from memref). 

Here is some IR that can reproduce the issue:

```MLIR
func.func @dim_of_pad(%in : tensor<?x?xi8>) -> (index , tensor<?x?xi8>) {
    %c1 = arith.constant 1 : index
    %c0_i8 = arith.constant 0 : i8
    %padded = tensor.pad %in low[1, 1] high[1, 1] {
    ^bb0(%arg1: index, %arg2: index):
      tensor.yield %c0_i8 : i8
    } : tensor<?x?xi8> to tensor<?x?xi8>
    %dim = tensor.dim %padded, %c1 : tensor<?x?xi8>
    return %dim, %padded : index, tensor<?x?xi8>
}
```

- If adding the pass ```resolve-shaped-type-result-dims``` to torch-mlir-opt.cpp, you can reproduce the error with ```torch-mlir-opt --resolve-shaped-type-result-dims``` applied to the above IR.
- ```torch-mlir-opt --torch-backend-to-linalg-on-tensors-backend-pipeline``` applied to the above IR also reproduces the crash.
- However, ```mlir-opt --resolve-shaped-type-result-dims``` does NOT reproduce the crash, but instead generates the valid IR:

```MLIR
affine_map<()[s0] -> (s0 + 2)>
module {
  func.func @dim_of_pad(%arg0: tensor<?x?xi8>) -> (index, tensor<?x?xi8>) {
    %c1 = arith.constant 1 : index
    %c0_i8 = arith.constant 0 : i8
    %padded = tensor.pad %arg0 low[1, 1] high[1, 1] {
    ^bb0(%arg1: index, %arg2: index):
      tensor.yield %c0_i8 : i8
    } : tensor<?x?xi8> to tensor<?x?xi8>
    %dim = tensor.dim %arg0, %c1 : tensor<?x?xi8>
    %0 = affine.apply #map()[%dim]
    return %0, %padded : index, tensor<?x?xi8>
  }
}
```

## Questions:

[...]

PR #3364, opened 2024-05-17
Link necessary op interface implementations

zjgarvey:
This patch adds two `memref` passes to `torch-mlir-opt`, which already occur in the pass pipeline `torch-backend-to-linalg-on-tensors-backend-pipeline`. Additionally, necessary op interface external models are included to address issue #3352.


lib/InitAll.cpp
```diff
@@ -48,6 +49,7 @@ void mlir::torch::registerOptionalInputDialects(
   registry.insert<complex::ComplexDialect, linalg::LinalgDialect,
                   memref::MemRefDialect, ml_program::MLProgramDialect,
                   scf::SCFDialect, tensor::TensorDialect, tosa::TosaDialect>();
+  tensor::registerInferTypeOpInterfaceExternalModels(registry);
```
ramiro050, line 52: nit: this is [an extension](https://github.com/llvm/llvm-project/blob/bba5ee47e63298d61f6ea441a140144ce370ba92/mlir/lib/Dialect/Tensor/IR/TensorInferTypeOpInterfaceImpl.cpp#L204). We should add in [InitAll.h](https://github.com/llvm/torch-mlir/blob/main/include/torch-mlir/InitAll.h)

```cpp
void registerAllExtensions(mlir::DialectRegistry &registry);
```
and define the function in this file, `InitAll.cpp`.

Then you can do

```diff
 int main(int argc, char **argv) {
+  DialectRegistry registry;
   mlir::torch::registerAllPasses();
+  mlir::torch::registerAllExtensions(registry);
```

in `torch-mlir-opt.cpp`.

ramiro050 (changes requested):
I've also run into similar issues in the past. We have not been very strict about ensuring `torch-mlir-opt` has the right dependencies. Thanks for this fix!

ramiro050 (approved):
LGTM!

PR #3367, opened 2024-05-20
[Stablehlo] support uint8

qingyunqu:
Support lowering unsigned integer type to stablehlo as discussed in https://github.com/llvm/torch-mlir/pull/2184.

The things I do in this PR:
1. create `setupBackendTypeConversionForStablehlo()`, `createFuncBackendTypeConversionForStablehloPass` and `createFinalizingBackendTypeConversionForStablehloPass`.
2. remove `InferTypeOpInterface` from `torch_c.to_builtin_tensor`, because it's different result type between linalg backend and stablehlo backend:
```
// linalg backend
func.func @forward(%arg0: !torch.vtensor<[3],ui8>) -> tensor<3xf32> {
    %c = torch_c.to_builtin_tensor %arg0 : (!torch.vtensor<[3], ui8> -> tensor<3xi8>
    %0 = tensor.empty() : tensor<3xf32>
    %1 = linalg.generic {indexing_maps = [#map, #map], iterator_types = ["parallel"]} ins(%arg0 : tensor<3xi8>) outs(%0 : tensor<3xf32>) {
    ^bb0(%in: i8, %out: f32):
      %2 = arith.uitofp %in : i8 to f32
      linalg.yield %2 : f32
    } -> tensor<3xf32>
    return %1 : tensor<3xf32>
}
// stablehlo backend
func.func @forward(%arg0: !torch.vtensor<[3],ui8>) -> tensor<3xf32> {
    %c = torch_c.to_builtin_tensor %arg0 : (!torch.vtensor<[3], ui8> -> tensor<3xui8>
    %0 = stablehlo.convert %arg0 : (tensor<3xui8> -> tensor<3xf32>
    return %0 : tensor<3xf32>
}
```
3. fix stablehlo and linalg's conversion

### DIFF
--- a/lib/Dialect/TorchConversion/IR/TorchConversionOps.cpp
+++ b/lib/Dialect/TorchConversion/IR/TorchConversionOps.cpp
@@ -24,6 +24,10 @@ static bool haveSameSizeAndElementType(TensorType lhs, TensorType rhs) {
     return false;
   bool sameSize = lhs.hasRank() ? lhs.getShape().equals(rhs.getShape()) : true;
   bool sameElementType = false;
+  // Namely, it is worth mentioning that the backends can have different
+  // expectations for signedness when converting from and to the builtin MLIR
+  // types. Therefore, the verifier cannot expect the input and output types to
+  // match in their signedness.
   if (isa<IntegerType>(lhs.getElementType()) &&
       isa<IntegerType>(rhs.getElementType())) {
     sameElementType = lhs.getElementType().getIntOrFloatBitWidth() ==
--- a/lib/Dialect/TorchConversion/IR/TorchConversionOps.cpp
+++ b/lib/Dialect/TorchConversion/IR/TorchConversionOps.cpp
@@ -23,7 +23,14 @@ static bool haveSameSizeAndElementType(TensorType lhs, TensorType rhs) {
   if (lhs.hasRank() != rhs.hasRank())
     return false;
   bool sameSize = lhs.hasRank() ? lhs.getShape().equals(rhs.getShape()) : true;
-  bool sameElementType = lhs.getElementType() == rhs.getElementType();
+  bool sameElementType = false;
+  if (isa<IntegerType>(lhs.getElementType()) &&
+      isa<IntegerType>(rhs.getElementType())) {
+    sameElementType = lhs.getElementType().getIntOrFloatBitWidth() ==
+                      rhs.getElementType().getIntOrFloatBitWidth();
+  } else {
+    sameElementType = lhs.getElementType() == rhs.getElementType();
+  }
   return sameElementType && sameSize;
 }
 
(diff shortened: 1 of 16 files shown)


lib/Dialect/TorchConversion/Transforms/BackendTypeConversion.cpp
```diff
@@ -51,6 +51,39 @@ setupValueTensorToBuiltinTensorConversion(ConversionTarget &target,
...
+    ConversionTarget &target, TypeConverter &typeConverter) {
+  target.addLegalOp<TorchConversion::ToBuiltinTensorOp,
+                    TorchConversion::FromBuiltinTensorOp>();
+  typeConverter.addConversion(
+      [](Torch::ValueTensorType type) -> std::optional<Type> {
+        auto builtinType = type.toBuiltinTensor();
+        if (type.getDtype().isUnsignedInteger()) {
+          return builtinType.clone(type.getDtype());
+        }
+        return builtinType;
```
ramiro050, line 65: The functions `setupValueTensorToBuiltinTensorConversionForStablehlo` and `setupValueTensorToBuiltinTensorConversion` only differ here. Let's create a helper function that takes in a `function_ref<std::optional<Type>>(ValueTensorType)` so that the rest of the logic can be shared.

Also, this feels a bit backwards. I think it would be better to change `toBuiltinTensor` to always output types with the same signedness -- this is aligned with the [documentation for this function](https://github.com/llvm/torch-mlir/blob/617b00b983dec0fca0a7e13224d04a5862ffab05/include/torch-mlir/Dialect/Torch/IR/TorchTypes.td#L195-L196), which says that the result type has the same static information as the input type. Removing signedness is a loss of information. Backends should then choose if they want to drop the signedness or not.

lib/Dialect/TorchConversion/IR/TorchConversionOps.cpp
```diff
@@ -25,7 +25,14 @@ static bool haveSameSizeAndElementType(TensorType lhs, TensorType rhs) {
   if (lhs.hasRank() != rhs.hasRank())
     return false;
   bool sameSize = lhs.hasRank() ? lhs.getShape().equals(rhs.getShape()) : true;
-  bool sameElementType = lhs.getElementType() == rhs.getElementType();
+  bool sameElementType = false;
+  if (isa<IntegerType>(lhs.getElementType()) &&
```
ramiro050, line 29: nit: I would add a comment explaining this. Namely, it is worth mentioning that the backends can have different expectations for signedness when converting from and to the builtin MLIR types. Therefore, the verifier cannot expect the input and output types to match in their signedness.

ramiro050 (changes requested):
Looks mostly good. Thanks for this PR! A couple of comments

lib/Dialect/TorchConversion/Transforms/BackendTypeConversion.cpp
qingyunqu: Yeah, I agree it, I will fix `toBuiltinTensor` in next PR. 
`toBuiltinTensor` was used widely, it deserves a new PR to review.

lib/Dialect/TorchConversion/IR/TorchConversionOps.cpp
qingyunqu: Done

lib/Dialect/TorchConversion/Transforms/BackendTypeConversion.cpp
ramiro050: Thanks!

ramiro050 (approved):

PR #3371, opened 2024-05-21
[ONNX] Add OnnxToTorch lowering for Onnx.Upsample Op

vivekkhandelwal1:


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2875,4 +2875,64 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                 /*Torch_BoolType:$antialias*/ cstFalse);
         return success();
       });
+  patterns.onOp(
+      "Upsample", 9, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        std::string mode;
+        Value input, scales;
+        if (binder.tensorOperands(input, scales) ||
```
zjgarvey, line 2883: You don't seem to be using the scales. 

At least for bilinear upsampling, the scale information is important to compute the weights in the average over nearby pixels. For nearest mode, do you know if they use round prefer floor (in onnx and/or pytorch), or just floor to determine nearest pixel? If it's round prefer floor, the scales might still be relevant. 

Also, if it doesn't matter in the end, would it be better to write this conversion by just passing the scales to the aten.upsample_nearest2d op instead of getting the sizes? That's not a review requirement, but an honest question. 

Either way, I feel like it should be documented in a comment how this op does rounding. These resizing ops are too varied in their implementations between each framework, so any commentary you can add would be really valuable to others who need to add functionality (like different modes) in the future.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2875,4 +2875,64 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          for (uint32_t i = 2; i < resultShape.size(); i++)
+            outputSizeIntList.push_back(rewriter.create<Torch::ConstantIntOp>(
+                binder.getLoc(), rewriter.getI64IntegerAttr(resultShape[i])));
+        } else {
+          Torch::BaseTensorType scalesType =
+              cast<Torch::BaseTensorType>(scales.getType());
+          SmallVector<int64_t> selectSizes{1};
+          Type selectResultType = scalesType.getWithSizesAndDtype(
+              selectSizes, scalesType.getOptionalDtype());
+          Value zero = rewriter.create<Torch::ConstantIntOp>(
```
zjgarvey, line 2910: I'd prefer the use of the shorter build form for this constant int op, since specifying rewriter.getType<Torch::IntType>() is superfluous.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2875,4 +2875,64 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+              binder.getLoc(), rewriter.getType<Torch::IntType>(),
+              rewriter.getI64IntegerAttr(0));
+          for (int i = 2; i < resultRank; i++) {
+            // Go through the scales list and get output sizes in the list
+            Value selectIndex = rewriter.create<Torch::ConstantIntOp>(
```
zjgarvey, line 2915: Same as above.

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
```diff
@@ -2163,3 +2163,16 @@ f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_ve
     %0 = torch.operator "onnx.Resize"(%arg0, %none, %none, %arg1) {torch.onnx.mode = "linear"} : (!torch.vtensor<[1,1,2,4],f32>, !torch.none, !torch.none, !torch.vtensor<[4],si64>) -> !torch.vtensor<[?,?,?,?],f32>
     return %0 : !torch.vtensor<[?,?,?,?],f32>
   }
+
+// -----
+
+func.func @test_upsample_nearest(%arg0: !torch.vtensor<[1,1,2,2],f32>, %arg1: !torch.vtensor<[4],f32>) -> !torch.vtensor<[1,1,4,6],f32> attributes {torch.onnx_meta.ir_version = 4 : si64, torch.onnx_meta.opset_version = 9 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
```
zjgarvey, line 2169: Have you run any e2e tests? I'm concerned about the scale issue etc., so it would be good to double check some corresponding iree_tests, for example.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2875,4 +2875,64 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+
+        int64_t resultRank = resultType.getSizes().size();
+        if (resultRank != 4)
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "supports 2d sampling only");
+
+        SmallVector<Value> outputSizeIntList;
+        if (resultType.areAllSizesKnown()) {
+          SmallVector<int64_t> resultShape(resultType.getSizes());
+          for (uint32_t i = 2; i < resultShape.size(); i++)
```
zjgarvey, line 2901: Question: is `unsigned` the same as `uint32_t`?

zjgarvey (changes requested):
Just some questions and nits. 

I'd like to hear about the e2e tests, and would like to request some comments to be included to make it easier for others to contribute more functionality in the future.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2875,4 +2875,64 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                 /*Torch_BoolType:$antialias*/ cstFalse);
         return success();
       });
+  patterns.onOp(
+      "Upsample", 9, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
```
renxida, line 2879: https://onnx.ai/onnx/operators/text_diff_Upsample_1_10.html

I'm not sure but it looks like this is general enough to support as early as Upsample 1.

renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: They expect the scales arg to be of different type.

vivekkhandelwal1:
Hi @zjgarvey, the PR is updated, can you please review it now?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -3258,4 +3257,42 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        Torch::ValueTensorType resultType;
+        std::string mode;
+        Value input, scales;
+        if (binder.tensorOperands(input, scales) ||
+            binder.customOpNameStringAttr(mode, "mode", "nearest") ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+
+        if (mode != "nearest" && mode != "linear" && mode != "bilinear" &&
```
zjgarvey, line 3271: I think the onnx op only has options "nearest" and "linear".

To lower appropriately to interpolate, you might have to convert the modeStringValue from "linear" to "bilinar" in 2d, "linear" to "trilinear" in 3d, etc.

zjgarvey (commented):
A small change related to the mode string should be made.

vivekkhandelwal1:
Hi @zjgarvey, I have addressed the comment. Could you please take a look now?

zjgarvey (approved):

PR #3372, opened 2024-05-21
Representing Symbolic Shape Expressions in Torch Dialect

sjain-stanford:
Torch Dialect with symbolic shape expressions:
```ll
module {                                                                                                                                                                                                     
  func.func @main(%arg0: !torch.vtensor<[?,?,3],f32>, %arg1: !torch.vtensor<[?,?,3],f32>) -> !torch.vtensor<[?,?,3],f32> {                                                                                   
    %0 = torch.symbolic_int "s0" {min_val = 5, max_val = 10} : !torch.int                                                                                                                                    
    %1 = torch.symbolic_int "s1" {min_val = 0, max_val = 100} : !torch.int                                                                                                                                   
    %2 = torch.symbolic_int "s3" {min_val = 0, max_val = 50} : !torch.int                                                                                                                                    
    
    torch.bind_symbolic_shape %arg0, [%0, %1], #affine_map<()[s0, s1] -> (s0, s1, 3)> : !torch.vtensor<[?,?,3],f32>                                                                                          
    torch.bind_symbolic_shape %arg1, [%0, %2], #affine_map<()[s0, s1] -> (s0, s1, 3)> : !torch.vtensor<[?,?,3],f32>                                                                                          
    
    %3 = torch.aten.tanh %arg0 : !torch.vtensor<[?,?,3],f32> -> !torch.vtensor<[?,?,3],f32>                                                                                                                  
    torch.bind_symbolic_shape %3, [%0, %1], #affine_map<()[s0, s1] -> (s0, s1, 3)> : !torch.vtensor<[?,?,3],f32>                                                                                             
    
    %4 = torch.aten.sigmoid %arg1 : !torch.vtensor<[?,?,3],f32> -> !torch.vtensor<[?,?,3],f32>                                                                                                               
    torch.bind_symbolic_shape %4, [%0, %2], #affine_map<()[s0, s1] -> (s0, s1, 3)> : !torch.vtensor<[?,?,3],f32>                                                                                             
    
    %5 = torch.prim.ListConstruct %3, %3, %4 : (!torch.vtensor<[?,?,3],f32>, !torch.vtensor<[?,?,3],f32>, !torch.vtensor<[?,?,3],f32>) -> !torch.list<vtensor>                                               
    %int1 = torch.constant.int 1                                                                                                                                                                             
    %6 = torch.aten.cat %5, %int1 : !torch.list<vtensor>, !torch.int -> !torch.vtensor<[?,?,3],f32>                                                                                                          
    torch.bind_symbolic_shape %6, [%0, %1, %2], #affine_map<()[s0, s1, s2] -> (s0, s1 * 2 + s2, 3)> : !torch.vtensor<[?,?,3],f32>                                                                            
    
    return %6 : !torch.vtensor<[?,?,3],f32>                                                                                                                                                                  
  }                                                                                                                                                                                                          
}              
```

For reference, this is the TorchDynamo exported program with symbolic shape expressions that the above Torch dialect program is imported from:
```py
ExportedProgram:                                                                                                                                                                                             
    class GraphModule(torch.nn.Module):                                                                                                                                                                      
        def forward(self, x: "f32[s0, s1, 3]", y: "f32[s0, s3, 3]"):                                                                                                                                         
            # File: /home/sambhav.jain/workspaces/cruise/src/3p/torch-mlir/test/python/fx_importer/symbolic_shape_expr_test.py:31 in forward, code: a = torch.tanh(x)                                        
            tanh: "f32[s0, s1, 3]" = torch.ops.aten.tanh.default(x);  x = None                                                                                                                               
                                                                                                                                                                                                             
            # File: /home/sambhav.jain/workspaces/cruise/src/3p/torch-mlir/test/python/fx_importer/symbolic_shape_expr_test.py:32 in forward, code: b = torch.sigmoid(y)                                     
            sigmoid: "f32[s0, s3, 3]" = torch.ops.aten.sigmoid.default(y);  y = None                                                                                                                         
                                                                                                                                                                                                             
            # File: /home/sambhav.jain/workspaces/cruise/src/3p/torch-mlir/test/python/fx_importer/symbolic_shape_expr_test.py:33 in forward, code: return torch.cat((a, a, b), dim=1)                       
            cat: "f32[s0, 2*s1 + s3, 3]" = torch.ops.aten.cat.default([tanh, tanh, sigmoid], 1);  tanh = sigmoid = None                                                                                      
            return (cat,)                                                                                                                                                                                    
                                                                                                                                                                                                             
Graph signature: ExportGraphSignature(input_specs=[InputSpec(kind=<InputKind.USER_INPUT: 1>, arg=TensorArgument(name='x'), target=None, persistent=None), InputSpec(kind=<InputKind.USER_INPUT: 1>, arg=TensorArgument(name='y'), target=None, persistent=None)], output_specs=[OutputSpec(kind=<OutputKind.USER_OUTPUT: 1>, arg=TensorArgument(name='cat'), target=None)])                                               
Range constraints: {s0: ValueRanges(lower=5, upper=10, is_bool=False), s1: ValueRanges(lower=0, upper=100, is_bool=False), s3: ValueRanges(lower=0, upper=50, is_bool=False)} 
```

Huge credit to @stellaraccident for the inputs that helped evaluate the various design options and arrive at the representation of choice.


- [x] Op definitions for symbolic_int and bind_symbolic_shape ops
- [x] fx_importer updates to import range constraints + create symbolic_int ops
- [x] fx_importer changes for AffineMapAttr building + adding bind_symbolic_shape ops
- [x] custom printer/parser for inlined AffineMap expressions in mlir assembly
- [x] Dialect lit test
- [x] fx_importer python lit tests
- [ ] Cleanup pass to remove these ops (can add in a follow-on)


include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -1337,4 +1338,67 @@ def Torch_DtypeCalculateYieldDtypesOp : Torch_Op<"dtype.calculate.yield.dtypes",
   let hasVerifier = 1;
 }
 
+//===----------------------------------------------------------------------===//
+// Symbolic shape modeling ops for TorchDynamo frontend.
+//===----------------------------------------------------------------------===//
+
+def Torch_SymbolicIntOp : Torch_Op<"symbolic_int", [Pure]> {
```
sjarus, line 1345: Personally, I find 'symbolic_int' to be an odd name. Does that mean there are other symbolic types envisioned - a float perhaps ? Or a non-symbolic int ? 

If we reasonably assume dim shapes are only ever going to be constrained i64 ranges, would something like symbolic_dim,  dymbolic_shape or symbolic_dim_shape be better? A name that fits more closely with the realted op bind_symbolic_shape would be ideal.

test/Dialect/Torch/canonicalize.mlir
```diff
@@ -3026,3 +3026,35 @@ func.func @torch.aten.clone$no_fold(%arg0: !torch.vtensor<[1,2,50,4],f32>) -> (!
...
 }
+
+
+// -----
+
+// CHECK-LABEL:   @torch.symbolic_int$canonicalize(
+// CHECK-SAME:        %[[ARG0:.*]]: !torch.vtensor<[?],f32>,
+// CHECK-SAME:        %[[ARG1:.*]]: !torch.vtensor<[?],f32>) -> !torch.vtensor<[?],f32> {
+// CHECK:               %[[S0:.*]] = torch.symbolic_int "s0" {min_val = 3, max_val = 6} : !torch.int
+// CHECK-NOT:           %[[S1:.*]] = torch.symbolic_int "s0 + 1" {min_val = 4, max_val = 7} : !torch.int
```
stellaraccident, line 3037: Is this just verifying that symbolic_int is a pure op and DCE's if not referenced?

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1485,6 +1585,74 @@ def _import_torch_op_overload(
...
+                affine_map = AffineMap.get(0, len(symbols_set), affine_exprs)
+
+                # Build operand list
+                operand_list = []
+                operand_list.append(self.resolve_node_value(node))
+                for symbol in symbol_table.keys():
+                    operand_list.append(self.resolve_symbol_value(symbol))
+
+                # Create torch.bind_symbolic_shape ops
+                Operation.create(
```
stellaraccident, line 1649: I think we need to make the emission of these ops optional/configurable, even if just as a transitional measure while phased in: without a cleanup pass wired into every pipeline, these impure ops will persist and block a number of optimizations if added. Probably best to add a knob to FxImporter to enable symbolic shape guards and then gate anything that calls this function on it.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1485,6 +1585,74 @@ def _import_torch_op_overload(
...
+                }
+
+                # Convert symbolic shape expressions into affine expressions
+                affine_exprs = [
+                    (
+                        AffineConstantExpr.get(expr)
+                        if isinstance(expr, int)
+                        # Use `symbol_table.copy()` here to avoid `eval` from
+                        # adding the `__builtins__` entry to the map.
+                        else eval(str(expr), symbol_table.copy())
```
stellaraccident, line 1636: Using eval like this isn't great. Just going to leave this comment for a second while considering.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -1337,4 +1338,67 @@ def Torch_DtypeCalculateYieldDtypesOp : Torch_Op<"dtype.calculate.yield.dtypes",
...
+
+    Example:
+    ```
+    torch.bind_symbolic_shape %arg0, [%0, %1], affine_map<()[s0, s1] -> (s0, s1, 3)> : !torch.vtensor<[?,?,3],f32>
+    torch.bind_symbolic_shape %out0, [%0, %1, %2], affine_map<()[s0, s1, s2] -> (s0, s1 * 2 + s2, 3)> : !torch.vtensor<[?,?,3],f32>
+    ```
+  }];
+  let arguments = (ins
+    Torch_ValueTensorType:$operand,
+    Variadic<Torch_IntType>:$shape_symbols,
```
sjarus, line 1396: What happens if the shape symbols are not purely generated from torch.symbolic_int or scalar literals ? How about if they are values read from a (not necessarily const) tensor ? 

While I cannot yet find a relevant example in PyTorch, this is seen during the decomposition of tf/tfl.batch_to_space_nd and similar ops. 

This does fall within the class of data dependent dynamism, and it's ok to say this PR itself doesn't go so far as as to address it and that's fine - I'm just pattern matching against several dynamic cases of interest I'm looking at too.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -14,6 +14,8 @@
 import logging
 import operator
 import re
+import sympy
```
sjarus, line 17: How does this relate to the recent sympy PR in PyTorch for symbolic shape expression: https://github.com/pytorch/pytorch/pull/94985 ?

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1485,6 +1585,74 @@ def _import_torch_op_overload(
...
+                results=[self._cc.torch_int_type],
+                loc=loc,
+            )
+            self.bind_symbol_value(symbol, operation.result)
+
+    def _create_bind_symbolic_shape_ops(self, loc: Location, node: torch_fx.Node):
+        node_val = node.meta.get("val")
+        if (node_val is not None) and isinstance(node_val, TorchFakeTensor):
+            # Only create bind ops if the shapes contain symbolic sizes
+            if node_val._has_symbolic_sizes_strides:
```
stellaraccident, line 1609: Could you add some commentary either pointing to docs or having an inline explanation about the types involved in _has_symbolic_sizes_strides? I think I roughly understand it but would have trouble modifying it without a lot of digging.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1021,6 +1041,54 @@ def get_node_location(self, node: torch_fx.Node) -> Optional[Location]:
...
+        for val in prog.range_constraints.values():
+            if (
+                isinstance(val.lower, sympy.Integer)
+                and isinstance(val.upper, sympy.Integer)
+                and not val.is_bool
+            ):
+                contains_symbolic_ints = True
+                break
+        if contains_symbolic_ints:
+            self._symbolic_guards = {
```
stellaraccident, line 1081: Can you please add some commentary about the transformation this is doing? I'm a little bit fuzzy on it.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1021,6 +1041,54 @@ def get_node_location(self, node: torch_fx.Node) -> Optional[Location]:
...
+                type(val),
+                val,
+                adjust,
+            )
+            if adjust == "floor":
+                return math.floor(val)
+            elif adjust == "ceil":
+                return math.ceil(val)
+            else:
+                raise RuntimeError(f"Got invalid adjustment {adjust}")
```
stellaraccident, line 1069: You can remove this if cascade and the potential for a runtime error if you make the `adjust` just be a unary function. Invoke it with `_sympy_int_to_int(v.lower, math.ceil)` below, for example.

test/python/fx_importer/basic_test.py
```diff
@@ -108,6 +113,12 @@ def forward(self, x):
...
 # CHECK:     func.func @test_net(%[[ARG0:[a-zA-Z0-9]+]]: !torch.vtensor<[1,2],f32>, %[[ARG1:[a-zA-Z0-9]+]]: !torch.vtensor<[?],f32>) -> !torch.vtensor<[?,2],f32>
+# CHECK:     %[[S0:.*]] = torch.symbolic_int "s0" {min_val = {{[0-9]+}}, max_val = {{[0-9]+}}} : !torch.int
+# CHECK:     torch.bind_symbolic_shape %[[ARG1]], [%[[S0]]], affine_map<()[s0] -> (s0)> : !torch.vtensor<[?],f32>
+# CHECK:     torch.aten.size.int
+# CHECK:     torch.prim.ListConstruct
+# CHECK:     %[[EXPAND:.*]] = torch.aten.expand
+# CHECK:     torch.bind_symbolic_shape %[[EXPAND]], [%[[S0]]], affine_map<()[s0] -> (s0, 2)> : !torch.vtensor<[?,2],f32>
 def test_broadcast_with_dynamic_shapes():
     class Basic(nn.Module):
         def __init__(self):
```
sjarus, line 124: Would it be possible to add a couple of more examples of broadcasting, especially the potential multiple options around binary broadcasting against a non-unit dim e.g. tensor<3x3xf32> vs tensor<?x3xf32> or where both tensors have unknown dim shapes ? 

Reduction ops would also be a great example to see, especially with multiple unknown axial dims with keepdim=False.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -1485,6 +1585,74 @@ def _import_torch_op_overload(
...
+                    "symbol_name": StringAttr.get(symbol),
+                    "min_val": self._cc.integer_attr(constraints.min_val, 64),
+                    "max_val": self._cc.integer_attr(constraints.max_val, 64),
+                },
+                results=[self._cc.torch_int_type],
+                loc=loc,
+            )
+            self.bind_symbol_value(symbol, operation.result)
+
+    def _create_bind_symbolic_shape_ops(self, loc: Location, node: torch_fx.Node):
```
stellaraccident, line 1605: So if I understand correctly what this function is doing, it seems like the core of it is doing a kind of abstract evaluation by substituting the SymInt terminals with a corresponding AffineSymbolExpr, and then evaluating the expression tree as Python with the substitution. This then relies on AffineExpr having operator overloads and enough correspondence to be able to create an equivalent AffineExpr from the sympy expression tree. Do I have that right?

A couple thing I would wish for (assuming I have that right):

* Could we separate out the core of the abstract evaluation/expression transform into a dedicated top level helper function in the module. 
* This is the kind of thing that tends to have a number of corners to it that are hard to cover with e2e tests, and I would like to see a classic Python unit test that attempts to cover the weird cases, exercising the sympy_to_affine_expr_transform() helper.
* I'm not sure I trust the `affine_exprs = []` list comprehension yet. Does that only work for expression nodes that map 1:1 with Python operators and are overloaded on the AffineExpr?

There are some docs here https://docs.sympy.org/latest/tutorials/intro-tutorial/manipulation.html on how to do advanced expression manipulation and re-interpretation (example showing how to convert to latex, for example). I'd need to fiddle with it, but I would expect more of a walk that used a table of supported expressions that mapped explicitly to AffineExpr combiners, versus relying just on the accidental convergence of the two when evaluated as Python.

stellaraccident (commented):
Neat! I've left a number of first pass comments. I think I get the overall approach. If you'd like to break this up while continuing to work on the importer, feel free to land the dialect changes and corresponding tests while we work on the rest. Up to you.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
stellaraccident: I think Torch refers to these everywhere as SymInt (there are also SymFloat and SymBool). Not strictly what is being referenced here. This is more of a torch.sym_int_constant for SymInts that exist purely within the program, but I'm ok leaving it how it is (i.e. it makes sense coming from the Torch view of things).

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
stellaraccident: I don't think Torch allows that and the cases where it comes up are special cased. Something to watch but I don't think needs to be modeled here.

python/torch_mlir/extras/fx_importer.py
stellaraccident: That abstract interpreter is very similar to what I expected the translation from sympy expr -> AffineExpr to be implemented in terms of (versus punting to eval).

test/python/fx_importer/basic_test.py
stellaraccident: I know I've seen more complicated cases in the wild. They tend to come up when symbols are related in program arguments by some non trivial expression. Struggling to come up with one now.

python/torch_mlir/extras/fx_importer.py
sjain-stanford: Good callout. I'm revamping this to actually recursively traverse the sympy AST and translate it to AffineExpr depending on the type of sympy expr (Symbol vs Add vs Pow etc). Encountered a case of `s0**2` (sympy.Pow) which AffineExpr don't support natively (i.e. `s0**2` isn't but `s0 * s0` is supported). I think most such cases should be handled with the `mul`, `floordiv` , `ceildiv` and `mod` from semi-affine maps : https://mlir.llvm.org/docs/Dialects/Affine/#semi-affine-maps.

python/torch_mlir/extras/fx_importer.py
sjain-stanford: Good callout. I'm revamping this to actually recursively traverse the sympy AST and translate it to AffineExpr depending on the type of sympy expr (Symbol vs Add vs Pow etc). Encountered a case of s0**2 (sympy.Pow) which AffineExpr don't support natively (i.e. s0**2 isn't but s0 * s0 is supported). I think most such cases should be handled with the mul, floordiv , ceildiv and mod from semi-affine maps : https://mlir.llvm.org/docs/Dialects/Affine/#semi-affine-maps, but as you said, needs a more systematic transation with the `sympy_to_affine_expr_transform` helper. Let me rework this part and get back to you.

python/torch_mlir/extras/fx_importer.py
sjain-stanford: Thanks for sharing, taking a look.

test/Dialect/Torch/canonicalize.mlir
sjain-stanford: That's correct.

python/torch_mlir/extras/fx_importer.py
sjain-stanford: Took a first stab at `sympy_expr_to_semi_affine_expr()`. I've added the cases I could think of at the moment as being valid shape expressions (includes Add, Sub, Mul, Pow, Mod besides ConstantExpr / SymbolExpr). Needless to say, this will need to be massaged over time as we encounter fancier expressions that don't fit these categories. But glad that this seems much more extensible than its predecessor (python evaluations and operator overloads). 

I think this should address bullets 1 and 3 above (LMK if not). Bullet 2 is a reasonable ask - I will add python lit tests that unit test this function on some sympy expressions in a standalone manner, shortly.

sjain-stanford:
> Neat! I've left a number of first pass comments. I think I get the overall approach. If you'd like to break this up while continuing to work on the importer, feel free to land the dialect changes and corresponding tests while we work on the rest. Up to you.

Thanks for checking. While this is in flux, keeping the changes in one PR is a bit more convenient but I do want to be mindful of the review overhead, so happy to stack them into smaller PRs if you prefer.

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
sjain-stanford: Yeah precisely ⬆️ , didn't want to call it `symbolic_dim` to allow for future expansions where we might support importing `symbolic_float` and `symbolic_bool`s as well. Right now we insert these strictly for the SymInt case.

python/torch_mlir/extras/fx_importer.py
sjain-stanford: I've refactored this to use a structured translation from sympy expr -> AffineExpr, and the function itself looks very similar to `sympy_interp()` indeed.

python/torch_mlir/extras/fx_importer.py
sjain-stanford: Done

python/torch_mlir/extras/fx_importer.py
sjain-stanford: Just noticed you explicitly mentioned gating emission of the impure ops, but I've gated the insertion of both the pure `symbolic_int` ops as well as these impure `bind_symbolic_shape_ops`, just to provide a way to completely disable this feature if needed (we can relax this once this becomes a bit more load bearing).

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
sjain-stanford: Marking as resolved for now, feel free to reopen if you disagree.

test/python/fx_importer/basic_test.py
sjarus: We have a collection of more complex cases but primarily within the TF ecosystem. We're keen on seeing how things look in the torch world for at least some of the simpler forms, but incremental support for these constructs is fine - I don't want to block @sjain-stanford in the quest of immediate perfection!

sjain-stanford: Haha, no I'd be happy to include more cases to the testsuite, to help uncover rough edges. @sjarus LMK if you have any pytorch snippets you'd like me to try (either in this PR or as a follow-on), and I'll be sure to include.

python/torch_mlir/extras/fx_importer.py
sjain-stanford: Added the python unit test for validating the `sympy_expr_to_semi_affine_expr` helper, addressing bullet 2 above.

test/python/fx_importer/basic_test.py
sjain-stanford: With an eye to add support for more complicated cases as we come across them in the future, marking as resolved.

python/torch_mlir/extras/fx_importer.py
sjain-stanford: `_has_symbolic_sizes_strides` is a bool attribute. I've added an inline comment (couldn't find a doc upstream).

python/torch_mlir/extras/fx_importer.py
sjain-stanford: Done.

sjain-stanford:
@stellaraccident Since we already gate the insertion of these ops under a flag on fx_importer, and to keep this PR from becoming too large, in a follow-on PR I'll add a Torch transform pass that removes these symbolic guard and shape bind ops for the individual backend pipelines to include once they've translated/discarded the shape ops during conversion.

stellaraccident:
Thanks for the cleanups. Will re review properly shortly. Landing plan sgtm

test/python/fx_importer/basic_test.py
sjarus: Sorry I didn't see the followon comment in time! Doing it in follow-on PR sounds reasonable at this point!

sjarus (approved):

python/torch_mlir/extras/fx_importer.py
```diff
@@ -258,6 +268,71 @@
...
+    Recursively traverse the sympy expr AST and build the affine expr.
+    This is not a perfect translation. Sympy expressions are much more
+    expressive and not as constrained as affine (linear) expressions are.
+    However, for the most part, we don't need to support all of sympy.
+    PyTorch only uses a subset of sympy for capturing and expressing
+    symbolic shapes, and among what's supported, we expect the semi-affine
+    expressions (https://mlir.llvm.org/docs/Dialects/Affine/#semi-affine-maps)
+    to be sufficient.
+    """
+    if isinstance(expr, sympy.Symbol):
```
stellaraccident, line 291: Doesn't need to be for this patch, but I usually turn lookups like this into a dict keyed by type with a callback. I think there's a utility in this file that does a proper subclass map. Fine like you have it for now (I just usually try to avoid long if/else isinstance cascades)

stellaraccident (approved):
Thanks! Let's land and see where it goes next.

python/torch_mlir/extras/fx_importer.py
sjain-stanford: Thanks, that SGTM.

PR #3375, opened 2024-05-22
[FxImporter] Fix constant bool tensor

penguin-wwy:


python/torch_mlir/extras/fx_importer.py
```diff
@@ -1749,6 +1755,10 @@ def _make_vtensor_literal_op(
             )
         else:
             bytes_view = np_tensor.view(npy_dtype)
+            if tensor.dtype == torch.bool:
+                # The np.bool_ type has a length of 1 byte, which needs to be compressed into 1 bit,
+                # and specify the bitorder to ensure that the memory layout is consistent with DenseResourceElementsAttr.
+                bytes_view = np.packbits(bytes_view, bitorder="little")
```
stellaraccident, line 1761: This is surprising to me. I could have sworn that this is not how DRE stores bool tensors. Am almost positive these materialize as one value per byte.

stellaraccident: I expect that this needs to be an i8 DRE followed by a cast to an i1 tensor. Bool tensors are a mess in MLIR.

penguin-wwy: Based on my current observations, DRE actually does not support i1 tensors, but torch.bool must be converted to i1 type (otherwise, it would cause even greater trouble). Therefore, the DRE here will be cast to DenseElementsAttr during TorchToArith. I understand that there is confusion here, essentially, I used DRE to store the memory of DE.

stellaraccident: Yeah, I'm just trying to think of how hard this is going to be to unwind. It is important that our encodings of types matches the layout that torch uses (some integrations go so far as to jit these from device memory, for example). With that principle in mind, I think that i8 is the right element type. It can be followed by a tensor.cast to i1. Doing it that way makes the encoding change an internal detail of the compiler vs an "ABI" change at the boundary.

penguin-wwy: I modified it to generate an i8 constant tensor and added `convert_element_type` op to convert it to i1.
Please review this.

stellaraccident (approved):
This is probably the best we can do for now and at least doesn't change the layout of the data at rest.

PR #3381, opened 2024-05-22
Modifies onnx resize lowering to fix numerical issues

zjgarvey:
Updates:

- some unsupported modes are now going to report a match failure for unsupported coordinate transformation modes.
- fixes a bug that was introduced in the last patch for resize (my bad...)
- uses actual x and y coordinates for computing weights in bilinear interpolation (rather than eps modified values)
- slightly simplifies the bilinear interpolation payload for readability and performance
- passes coordinate transformation mode information from an onnx.Resize op to the mode string for the aten._interpolate op. This allows us to perform custom logic in the torch->linalg lowering to support onnx.Resize options without losing the default behaviors of the interpolate op.


aldesilv:
LGTM

renxida (approved):

PR #3385, opened 2024-05-23
[Onnx] Add Onnx->Torch lowering for Onnx.Shrink Op

vivekkhandelwal1:

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -3050,4 +3050,72 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             binder.op, resultType, permutedInput, reshapeSizesList);
         return success();
       });
+  patterns.onOp(
+      "Shrink", 9, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Location loc = binder.getLoc();
+        Torch::ValueTensorType resultType;
+        Value input;
+        float bias, lambd;
+        if (binder.tensorOperand(input) ||
+            binder.f32FloatAttr(bias, "bias", 0.0) ||
+            binder.f32FloatAttr(lambd, "lambd", 0.5) ||
+            binder.tensorResultType(resultType)) {
+          return failure();
+        }
+
+        Torch::ValueTensorType inputType =
+            cast<Torch::ValueTensorType>(input.getType());
+        if (!isa<mlir::FloatType>(inputType.getDtype()))
+          return rewriter.notifyMatchFailure(
+              binder.op, "unimplemented: non-floating point dtype");
+
+        // The formula of this operator is: If x < -lambd, y = x + bias; If x >
+        // lambd, y = x - bias; Otherwise, y = 0.
+        // The implementation is based on the following algorithm:
+        // Shrink <bias,lambd>(input) => (output)
+        // {
+        //    Lambd = Constant <value_float: float = @lambd> ()
+        //    LambdCast = CastLike (Lambd, input)
+        //    Bias = Constant <value_float: float = @bias> ()
+        //    BiasCast = CastLike (Bias, input)
+        //    Zero = Constant <value: tensor = float {0}> ()
+        //    ZeroCast = CastLike (Zero, input)
+        //    NegLmbda = Neg (LambdCast)
+        //    InputLessThanNegLambda = Less (input, NegLmbda)
+        //    InputAddBias = Add (input, BiasCast)
+        //    InputSubBias = Sub (input, BiasCast)
+        //    LambdaLessThanInput = Less (LambdCast, input)
+        //    InputSubBiasOrZero = Where (LambdaLessThanInput, InputSubBias,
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2897,4 +2897,66 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        //    InputLessThanNegLambda = Less (input, NegLmbda)
+        //    InputAddBias = Add (input, BiasCast)
+        //    InputSubBias = Sub (input, BiasCast)
+        //    LambdaLessThanInput = Less (LambdCast, input)
+        //    InputSubBiasOrZero = Where (LambdaLessThanInput, InputSubBias,
+        //    ZeroCast) output = Where (InputLessThanNegLambda, InputAddBias,
+        //    InputSubBiasOrZero)
+        // }
+        Torch::ValueTensorType inputType =
+            cast<Torch::ValueTensorType>(input.getType());
```
zjgarvey, line 2932: Is this conversion assuming the input has dtype f64? 

I would either add support for the cases where the input dtype is not a float, or put in a match failure.

zjgarvey (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: Yeah, forgot to add that check. Added now.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2895,4 +2895,70 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        Value constLambd = rewriter.create<Torch::ConstantFloatOp>(
+            loc, rewriter.getFloatAttr(rewriter.getF64Type(), lambd));
+        Value constBias = rewriter.create<Torch::ConstantFloatOp>(
+            loc, rewriter.getFloatAttr(rewriter.getF64Type(), bias));
+        Value constZero = rewriter.create<Torch::ConstantFloatOp>(
+            loc, rewriter.getFloatAttr(rewriter.getF64Type(), 0.0));
+        Value constOne = rewriter.create<Torch::ConstantFloatOp>(
+            loc, rewriter.getFloatAttr(rewriter.getF64Type(), 1.0));
+        Value constNegLambd = rewriter.create<Torch::ConstantFloatOp>(
+            loc, rewriter.getFloatAttr(rewriter.getF64Type(), lambd));
```
zjgarvey, line 2944: is this supposed to be `-lambd` instead of `lambd`?

zjgarvey: If this is a mistake and it wasn't caught by a test somewhere, then we really need to do more e2e testing.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2895,4 +2895,70 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          return failure();
+        }
+
+        Torch::ValueTensorType inputType =
+            cast<Torch::ValueTensorType>(input.getType());
+        if (!inputType.getDtype().isa<mlir::FloatType>())
+          return rewriter.notifyMatchFailure(
+              binder.op, "unimplemented: non-floating point dtype");
+
+        // The implementation is based on the following algorithm:
```
renxida, line 2917: Including this might make this easier to understand:

```
 The formula of this operator is: If x < -lambd, y = x + bias; If x > lambd, y = x - bias; Otherwise, y = 0.
```

Also, there's a torch analogue to this called Softshrink. Might be a good idea to move the impl to torch: https://pytorch.org/docs/stable/generated/torch.nn.Softshrink.html

andfau-amd:
Ah, but there's only a function available for it in newer versions of the ONNX operator set, so this lowering might be useful for the old ones?

vivekkhandelwal1:
> Ah, but there's only a function available for it in newer versions of the ONNX operator set, so this lowering might be useful for the old ones?

I just need to know, if we need this lowering or not. If yes, then I can address the comments and get this merged, otherwise I will close this PR.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: Sorry, that was a typo.

vivekkhandelwal1:
> > Ah, but there's only a function available for it in newer versions of the ONNX operator set, so this lowering might be useful for the old ones?
> 
> I just need to know, if we need this lowering or not. If yes, then I can address the comments and get this merged, otherwise I will close this PR.

Since, we don't have clarity on this, I think we can take a conservative approach and get this lowering in (there's no harm in that). We can remove this later, if it will become obsolete.

zjgarvey (approved):

PR #3387, opened 2024-05-24
[FxImporter] Fix embedding bag

penguin-wwy:


penguin-wwy:
@stellaraccident please review

stellaraccident (approved):
These daisy chained conditional op replacements are getting a bit out of hand. I'm approving this one, but I'd prefer these were moved to some kind of lookup table/callback. And would also prefer a comment describing the motivation for doing this lightweight "canonicalization" here.

penguin-wwy:
> These daisy chained conditional op replacements are getting a bit out of hand. I'm approving this one, but I'd prefer these were moved to some kind of lookup table/callback. And would also prefer a comment describing the motivation for doing this lightweight "canonicalization" here.

I will refactor this part as soon as possible to make it more efficient and readable.

PR #3388, opened 2024-05-24
Implement lowering of torch.aten.renorm

BaneTrifa:
Closes [nod-ai/SHARK-Turbine/issues/689](https://github.com/nod-ai/SHARK-Turbine/issues/689)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2088,6 +2088,151 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+                         maxnorm, sizeList);
+
+    // Divide maxnorm and normPlusEps
+    auto divideMaxnormAndNorm = rewriter.create<AtenDivTensorOp>(
+        loc, vectorNormOutType, maxnormTensorValue, normPlusEps);
+
+    // Next few lines corespond to this pythorch code: norm_factor =
+    // torch.where(norm > maxnorm, maxnorm / (norm + eps), 1.0)
+    auto boolTensorType = rewriter.getType<ValueTensorType>(
+        cast<BaseTensorType>(vectorNormOutType).getOptionalSizes(),
```
zjgarvey, line 2205: Same comment as above

zjgarvey: There's another below, too.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2088,6 +2088,151 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+      inputSizeValue.push_back(
+          rewriter.create<Torch::ConstantIntOp>(loc, inputSize[i]));
+    }
+
+    // Prepare arguments for linalg.vector_norm
+    Value dtypeValue;
+    Type vectorNormOutType;
+
+    if (dtype.isa<mlir::Float16Type, mlir::BFloat16Type>()) {
+      dtype = cast<Type>(rewriter.getF32Type());
```
zjgarvey, line 2171: I don't think the cast is necessary here.

projects/pt1/python/torch_mlir_e2e_test/test_suite/norm_like.py
```diff
@@ -633,3 +633,73 @@ def forward(self, x, w, b):
...
+    def forward(self, x):
+        return torch.ops.aten.renorm(x, self.p, self.dim, self.maxnorm)
+
+
+@register_test_case(module_factory=lambda: RenormModuleFloat16())
+def RenormModuleFloat16_basic(module, tu: TestUtils):
+    module.forward(tu.rand(3, 4, 5).to(torch.float16))
+
+
+class RenormModuleFloat32NegativeDim(torch.nn.Module):
```
zjgarvey, line 685: Does the conversion work with dynamic dims? I imagine so, but it would be nice if one of the tests could verify that.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4641,6 +4641,104 @@ LogicalResult AtenNormScalarOp::verify() {
...
+
+  auto inShape = selfType.getSizes();
+  int64_t selfRank = inShape.size();
+  auto selfDtype = selfType.getDtype();
+
+  // Check if dtype is one of those supported by renorm operation.
+  // ComplexType will match any torch complex types, but each float must be
+  // checked individually.
+  if (selfDtype.isa<mlir::ComplexType>()) {
+    return emitOpError(
```
zjgarvey, line 4664: If you include this verification here, do you also need to check for the corresponding match failure in the decompose complex ops pattern?

zjgarvey: Or rather, since this is just something which is unsupported in the lowering, but not for the original op, should it be included in the verifier after all?

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4641,6 +4641,104 @@ LogicalResult AtenNormScalarOp::verify() {
...
+    return emitOpError("dim must be a constant int");
+  }
+
+  // check if is dim is in the correct range
+  if (dim >= selfRank || dim < -selfRank) {
+    return emitOpError("Dimension out of range (expected to be in range of [")
+           << -selfRank << ", " << selfRank - 1 << "], but got " << dim;
+  }
+
+  // Canonicalize dimension if necessary
```
zjgarvey, line 4734: does this do anything?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2088,6 +2088,151 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+
+    Value reduceDimsList = rewriter.create<Torch::PrimListConstructOp>(
+        loc,
+        rewriter.getType<Torch::ListType>(rewriter.getType<Torch::IntType>()),
+        reduceDimsVector);
+
+    // Make output shape for linalg.vector_norm operation
+    SmallVector<Value> inputSizeValue;
+    for (u_int64_t i = 0; i < inputSize.size(); i++) {
+      if (i != (u_int64_t)dimInt) {
```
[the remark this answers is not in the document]
zjgarvey: There are a few other places like this throughout.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4641,6 +4641,104 @@ LogicalResult AtenNormScalarOp::verify() {
...
+//===----------------------------------------------------------------------===//
+// AtenRenormOp
+//===----------------------------------------------------------------------===//
+
+LogicalResult AtenRenormOp::verify() {
+
+  auto selfType = cast<BaseTensorType>(getSelf().getType());
+
+  if (!selfType.hasDtype() || !selfType.hasSizes()) {
+    return success();
```
zjgarvey, line 4653: Should this be a failure?

zjgarvey (commented):
Hey, thanks for the PR!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2088,6 +2088,151 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+
+    Value normPlusEps = rewriter.create<AtenAddScalarOp>(
+        loc, vectorNormOutType, norm, epsValue, cstOne);
+
+    Value sizeList = rewriter.create<PrimListConstructOp>(
+        loc, Torch::ListType::get(Torch::IntType::get(op.getContext())),
+        inputSizeValue);
+
+    Value maxnormTensorValue =
+        createInitTensor(rewriter, loc, cast<BaseTensorType>(vectorNormOutType),
```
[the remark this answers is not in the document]
BaneTrifa: I believe we need to cast vectorNormOutType to BaseTensorType since createInitTensor() expects BaseTensorType, while vectorNormOutType is an mlir::Type

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
BaneTrifa: I think that cast is necessary here, vectorNormOutType is mlir::Type but we need BaseTensorType because getOptionalSizes() can only be called on instances of BaseTensorType.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
BaneTrifa: Even though getF32Type() returns a value of type mlir::FloatType, which inherits from mlir::Type, we still need the cast because otherwise it produces an error.

projects/pt1/python/torch_mlir_e2e_test/test_suite/norm_like.py
BaneTrifa: For this operator, dynamic dimensions do not work because the decomposition of AtenRenormOp involves the use of other operators, for example, AtenLinalgVectorNormOp. Ffor AtenLinalgVectorNormOp I need to calculate return Type (shape and dtype). To calculate the return shape, I need the input shape, which is just a list of -1 in case of dynamic dimensions. Additionally, at some point, I need to create a tensor of a certain size, and I calculate that size based on the input shape, which I don't have (just list of -1).
If is there some other solutions feel free to share your thoughts.

lib/Dialect/Torch/IR/TorchOps.cpp
BaneTrifa: You're right. This is unsupported in lowering so it needs to be checked in Decompose only.

lib/Dialect/Torch/IR/TorchOps.cpp
BaneTrifa: No, I forgot to delete that part.

lib/Dialect/Torch/IR/TorchOps.cpp
BaneTrifa: Recently, I read in one of the PRs that one of the reviewers mentioned that verify() only checks the semantics of the operation itself, such as examining aspects like value ranges, whether dimensions are within the range, whether parameters are of correct types, etc. However, HasDtype and HasSizes are framework matters, meaning that for some reason, information about the data type and sizes of your tensor is not provided, even though all requirements of the operation are met when writing the Python command. In that case, reviewer said that you can simply return success() from verify(), and in the lowering (DecomposeComplexOps.cpp in my case) return failure if necessary.

projects/pt1/python/torch_mlir_e2e_test/test_suite/norm_like.py
BaneTrifa: @zjgarvey I added support for dynamic dims, can you take a look at the changes?

zjgarvey (approved):
This looks good to me. 

Before merging, it might be good to get someone's opinion about adding verifiers, since I haven't seen one added in torch-mlir before (perhaps since any torch code which would result in a violation likely throws an error when importing in the first place).

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4655,6 +4655,80 @@ LogicalResult AtenNormScalarOp::verify() {
...
+    return emitOpError("renorm: maxnorm must be real-valued");
+
+  // The argument 'maxnorm' can be either an integer or a floating-point number,
+  // so we need to consider both options and check if 'maxnorm' is within the
+  // correct range
+  int64_t maxnormInt = 0;
+  double_t maxnormDouble = 0;
+  if (!matchPattern(getMaxnorm(), m_TorchConstantInt(&maxnormInt)) &&
+      !matchPattern(getMaxnorm(), m_TorchConstantFloat(&maxnormDouble)))
+    return emitOpError("renorm: maxnorm must be real-valued");
```
ramiro050, line 4714: same comment here: failure of `matchPattern` is a `success` case in a verifier

ramiro050 (commented):
Thanks for this! Below is a review of the verifier

ramiro050 (approved):
LGTM!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -2068,6 +2068,145 @@ class DecomposeAtenMvOp : public OpRewritePattern<AtenMvOp> {
...
+    // Define all constants
+    Value cstTrue = rewriter.create<ConstantBoolOp>(loc, true);
+    Value cstZero = rewriter.create<Torch::ConstantIntOp>(loc, 0);
+    Value cstOne = rewriter.create<Torch::ConstantIntOp>(loc, 1);
+    Value cstNone = rewriter.create<ConstantNoneOp>(loc);
+
+    // Arragne reduce_dims tensor (vector), [0, 1, ... , dim-1, dim+1, ... ,
+    // ndim-1]
+    llvm::SmallVector<Value> reduceDimsVector;
+    for (u_int64_t i = 0; i < ndim; i++) {
```
ScottTodd, line 2119: This does not build on MSVC. The portable type name (also used in other parts of this file) is `uint64_t`.

`torch-mlir\lib\Dialect\Torch\Transforms\DecomposeComplexOps.cpp(2420): error C2065: 'u_int64_t': undeclared identifier`

Sent a fix: https://github.com/llvm/torch-mlir/pull/3519

PR #3391, opened 2024-05-27
[Torch] fix aten.linear's decomposition

qingyunqu:
* support aten.linear with more rank.


Vremold (approved):
LGTM. Thanks for the effort!

PR #3392, opened 2024-05-27
[Torch] move DenseResource handling into torch.vtensor's canonicalize

qingyunqu:
so that the conversion of `DenseResourceElementsAttr` to `DenseElementsAttr` could be shared by both onnx_importer and fx_importer.

### DIFF
--- a/include/torch-mlir/Conversion/TorchOnnxToTorch/Utils.h
+++ b/include/torch-mlir/Conversion/TorchOnnxToTorch/Utils.h
@@ -15,20 +15,7 @@
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
-
-class Endian {
-private:
-  static constexpr uint32_t uint32_ = 0x01020304;
-  static constexpr uint8_t magic_ = (const uint8_t &)uint32_;
-
-public:
-  static constexpr bool little = magic_ == 0x04;
-  static constexpr bool big = magic_ == 0x01;
-  static_assert(little || big, "Cannot determine endianness!");
-
-private:
-  Endian() = delete;
-};
+#include "torch-mlir/Utils.h"
 
 namespace mlir::torch::onnx_c {
 
--- a/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/TorchOps.td
@@ -966,6 +966,7 @@ def Torch_ValueTensorLiteralOp : Torch_Op<"vtensor.literal", [
   }];
 
   let hasFolder = 1;
+  let hasCanonicalizer = 1;
 }
 
 def Torch_TensorStaticInfoCastOp : Torch_Op<"tensor_static_info_cast", [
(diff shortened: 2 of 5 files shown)


stellaraccident (changes requested):
An I reading correctly that this will canonicalize all DenseResourceElementsAttr -> DenseElementsAttr?

Please don't do this. DenseElementsAttr is deprecated for such usage and is responsible for massive compile time and memory overhead.

qingyunqu:
> An I reading correctly that this will canonicalize all DenseResourceElementsAttr -> DenseElementsAttr?
> 
> Please don't do this. DenseElementsAttr is deprecated for such usage and is responsible for massive compile time and memory overhead.

So what is better way to handle DenseResource from fx_importer?

stellaraccident:
Your compiler backend must handle it. We don't have general purpose tooling in torch-mlir for folding resources. Since these are typically very large, general mlir folding is insufficient.

In IREE, for example, we inline them if small (by some arbitrary rule) and we have a global analysis that transforms them to a module level init time computation if profitable, then optionally bakes them out as be constants (also if profitable).

You can implement a simpler strategy in a backend (ie. Inline everything and pay the cost) but the frontend currently does not have the policy infra to make a good decision. Therefore it must make no decision. 

There are some other options. @jpienaar

qingyunqu:
> Your compiler backend must handle it. We don't have general purpose tooling in torch-mlir for folding resources. Since these are typically very large, general mlir folding is insufficient.
> 
> In IREE, for example, we inline them if small (by some arbitrary rule) and we have a global analysis that transforms them to a module level init time computation if profitable, then optionally bakes them out as new constants (also if profitable).
> 
> You can implement a simpler strategy in a backend (ie. Inline everything and pay the cost) but the frontend currently does not have the policy infra to make a good decision. Therefore it must make no decision.
> 
> There are some other options. @jpienaar

How about just change fx_importer(https://github.com/llvm/torch-mlir/blob/main/python/torch_mlir/extras/fx_importer.py#L1800) to just use DenseElementAttrs?

stellaraccident:
I'm sorry - we can't do that. Believe it or not, this choice is responsible for 10-100x compiler performance and memory use regressions.

There is a hook in fx_importer that lets you do whatever you want, but that needs to be something done by a user, not the default for the framework.

Why do you need this? Switching to DRE (and a more pluggable way of dealing with constants) was done a year ago and was a primary motivation for creating the fx importer at all.

qingyunqu:
> I'm sorry - we can't do that. Believe it or not, this choice is responsible for 10-100x compiler performance and memory use regressions.
> 
> There is a hook in fx_importer that lets you do whatever you want, but that needs to be something done by a user, not the default for the framework.
> 
> Why do you need this? Switching to DRE (and a more pluggable way of dealing with constants) was done a year ago and was a primary motivation for creating the fx importer at all.

Got it, thanks for detailed explaination! I totally agree it base on the feature of DRE.

stellaraccident:
We can talk about other ways to get to what you want, but we need to start from DRE. Feel free to start a thread if you're having an issue you can't solve another was and we can discuss.

qingyunqu:
> We can talk about other ways to get to what you want, but we need to start from DRE. Feel free to start a thread if you're having an issue you can't solve another way and we can discuss.

How could I hook fx_importer? Is there any example?

stellaraccident:
Here is where iree does it: https://github.com/iree-org/iree-turbine/blob/c4db712b85bf6ec79ac269fc412d52543806f857/shark_turbine/aot/support/procedural/exported_program.py#L344

(Iree has APIs for promoting literals to global parameters but the hook lets you do anything)

PR #3397, opened 2024-05-29
[ONNX] Add OnnxToTorch Lowering for LpNormalization op

vivekkhandelwal1:

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1897,7 +1897,7 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         return success();
       });
   patterns.onOp(
-      "LpNormalization", 22,
+      "LpNormalization", 1,
       [](OpBinder binder, ConversionPatternRewriter &rewriter) {
         Torch::ValueTensorType resultType;
         int64_t axis, p;
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1896,4 +1896,34 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
 
         return success();
       });
+  patterns.onOp(
+      "LpNormalization", 22,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        int64_t axis, p;
+        Value input;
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1894,6 +1894,36 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         rewriter.replaceOpWithNewOp<Torch::AtenOneHotOp>(binder.op, resultType,
                                                          argmax, oneInt);
 
+        return success();
+      });
+  patterns.onOp(
+      "LpNormalization", 22,
```
renxida, line 1900: I think we can lower since version to 1 because the only difference between 1 and 22 is that 22 supports bfloat16

https://onnx.ai/onnx/operators/text_diff_LpNormalization_1_22.html

zjgarvey, line 1900: Can you reduce the opset version to 1? I don't think there is a significant diff between these versions from [the docs](https://onnx.ai/onnx/operators/onnx__LpNormalization.html)

zjgarvey (approved):
Looks good with change to opset version.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1894,6 +1894,36 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        if (binder.tensorOperand(input) ||
+            binder.s64IntegerAttr(axis, "axis", -1) ||
+            binder.s64IntegerAttr(p, "p", 2) ||
+            binder.tensorResultType(resultType))
+          return failure();
+
+        auto loc = binder.getLoc();
+        Value cstAxis = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(axis));
+        Value cstP = rewriter.create<Torch::ConstantIntOp>(
```
renxida, line 1914: Do we know if AtenNormScalarOptDimOp handles negative dims?

renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
vivekkhandelwal1: Yes, it's handled here: https://github.com/llvm/torch-mlir/blob/e4be197efd85916cd378ef8e7f21ca3de13b5903/lib/Conversion/TorchToLinalg/Reduction.cpp#L480-L485

zjgarvey (approved):
LGTM

PR #3401, opened 2024-05-30
[Torch] support recompose of aten.split.with_sizes and aten.tensor_sp…

qingyunqu:
…lit.sections

* support recompose to aten.split.with_sizes and aten.tensor_split.sections
* fix recompose of aten.chunk


Vremold (approved):

qingyunqu:
@rsuderman I would like to migrate these recompose patterns to decompose in the future, but should carefully handle shape infer & value semantics and support either from jit_importer or onnx_importer.
In the short term, I would want to merge this PR temporarily for model landing.

PR #3404, opened 2024-05-30
[E2E][ONNX] torch.multinomial

123epsilon:


lib/Conversion/TorchToLinalg/Random.cpp
```diff
@@ -12,6 +12,7 @@
 #include "PopulatePatterns.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/Linalg/IR/Linalg.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
```
zjgarvey, line 15: If I understand correctly, it is not acceptable to use scf dialect as a target for a torchtolinalg pass. I think many of your scf loops can be re-expressed as linalg generic ops, so please try to rewrite this pass without scf (and change to camelCase as well).

zjgarvey (changes requested):
Thanks for the PR! 

Before reviewing in more detail, I'd like to discuss whether scf is necessary to implement this conversion. As far as I am aware, lowering to scf is not acceptable for torch to linalg conversions, so I'd like you to investigate whether a similar linalg.generic would suffice as a replacement for the scf.for ops.

lib/Conversion/TorchToLinalg/Random.cpp
123epsilon: Got it, I'll try to refactor as linalg generics instead

lib/Conversion/TorchToLinalg/Random.cpp
```diff
@@ -192,6 +199,301 @@ class ConvertAtenUniformOp : public OpConversionPattern<AtenUniformOp> {
...
+                               Value midPointer =
+                                   b.create<arith::AddIOp>(loc, left, diffMid);
+                               Type indexTy = b.getIndexType();
+                               Value midIndex = b.create<arith::IndexCastOp>(
+                                   loc, indexTy, midPointer);
+
+                               // branch and update search indices
+                               auto thenBlock = [&](OpBuilder &b,
+                                                    Location loc) {
+                                 // left += 1
```
zjgarvey, line 420: Why did you opt to only increment the left endpoint by one in this case?

zjgarvey (commented):
I took a closer look and I have a few questions. 

It might be a good idea to have someone else take a look. too.

lib/Conversion/TorchToLinalg/Random.cpp
```diff
@@ -192,6 +199,301 @@ class ConvertAtenUniformOp : public OpConversionPattern<AtenUniformOp> {
...
+    auto sumBody = [&](OpBuilder &b, Location loc, ValueRange payloadArgs) {
+      Value input = payloadArgs[0];
+      Value result = payloadArgs[1];
+      Value nextSum = b.create<arith::AddFOp>(loc, input, result);
+      b.create<linalg::YieldOp>(loc, nextSum);
+    };
+    Value sumWeights = torch_to_linalg::createReductionLinalgGeneric(
+        rewriter, loc, opInfo, initSum, sumBody);
+
+    // Get multinomial samples for each weight vector
```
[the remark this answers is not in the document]
123epsilon: Yes that pseudo code looks correct to me (except, we're finding the FIRST category where `cdf(category) < randF`), I'll add some comments that explain this logic

lib/Conversion/TorchToLinalg/Random.cpp
```diff
@@ -192,6 +199,301 @@ class ConvertAtenUniformOp : public OpConversionPattern<AtenUniformOp> {
...
+                         .getResult(0);
+
+                 Value updatedCdf = b.create<tensor::InsertOp>(
+                     loc, currCum, distribution, ValueRange(iIndex));
+                 b.create<scf::YieldOp>(loc, ValueRange(updatedCdf));
+               })
+              .getResult(0);
+
+      // Get key, min and max used by RNG.
+      Value key = b.create<TorchConversion::GetNextSeedOp>(loc);
```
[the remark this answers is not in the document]
123epsilon: `GetNextSeedOp` essentially generates a new global seed for our RNG process. The reason that we recompute this seed at the batch level, but not at the sample level is because the RNG algorithm we are using takes both a key and an incrementing counter as input when generating a sequence of samples - each of which contribute to the randomness of subsequently generated elements. In our case, the key is the random seed from this op and the incrementing counter is the sample index we're currently on in the inner sample loop. 

For more clarity on this, you can see that this is how it is called when lowering the `AtenUniformOp` as well: https://github.com/llvm/torch-mlir/blob/7e7dadf819473e49d5287230ad738abc213cf57f/lib/Conversion/TorchToLinalg/Random.cpp#L156

Also, the RNG algorithm is implemented in this file with a link to the paper: https://github.com/llvm/torch-mlir/blob/7e7dadf819473e49d5287230ad738abc213cf57f/lib/Conversion/TorchToLinalg/Random.cpp#L74

lib/Conversion/TorchToLinalg/Random.cpp
123epsilon: Whoops, this is a logic error - I should've done newLeft = mid_pointer + 1

lib/Conversion/TorchToLinalg/Random.cpp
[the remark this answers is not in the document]
zjgarvey: Thanks for the info!

zjgarvey (approved):
With the logic change to the binary search, this looks good to me. 

Since this is a pretty heavy bit of work, it might be a good idea to grab another pair of eyes to look it over. Maybe @renxida or @vivekkhandelwal1 could take a look?

As a small nit suggestion for readability: the indentation is a bit intense at parts of this code, so it might be a good idea to define body lambdas outside the builder for some of these ops.

projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -2736,6 +2736,8 @@
     "ElementwiseUnaryIntModule_basic",
     "ElementwiseFloatTensorGtIntTensorModule_basic",
     "MaskedFillTensorFloatValueModule_basic",
+    "MultinomialModule_basic",
+    "MultinomialModule2D_basic",
```
vivekkhandelwal1, line 2740: Shouldn't these tests be passing, since you've added the Linalg lowering too?

vivekkhandelwal1 (approved):
LGTM

projects/pt1/e2e_testing/xfail_sets.py
123epsilon: @vivekkhandelwal1 This confuses me too, this doesn't happen in any other backend but in ONNX this throws an error `RuntimeError: invalid multinomial distribution (sum of probabilities <= 0)`, which I believe is coming from the [CPU kernel](https://github.com/pytorch/pytorch/blob/eea4ece256d74c6f25c1f4eab37b3f2f4aeefd4d/aten/src/ATen/native/cpu/MultinomialKernel.cpp#L183) for multinomial. 

My implementation works in the other backends, and it seems that this error occurs when computing a reference value using pytorch so I would think it is an issue with tracing or with my onnx to torch lowering, however, in my onnx to torch lowering I'm simply just passing the input tensor through to torch.

Do you have any idea why this could be happening?

vivekkhandelwal1: This might be a tracing issue. We can go ahead with this PR for now, since you've added it as a Xfail.

PR #3405, opened 2024-05-31
[NFC] Change to *cast instead of .*cast variants

rsuderman:
Member casts have been deprecated. Changing over a bunch of the member cast calls to the global templated variants to remove deprecation warnings.


stellaraccident (approved):
Thank you. Been annoyed by those for weeks.

kuhar (approved):

PR #3406, opened 2024-05-31
[Stablehlo] Support AtenScaledDotProductAttentionOp

william0021224:
1. support aten.scaled_dot_product_attention that can handle various arguments(attn_mask, is_causal, scale)
2. add related unit test cases

### DIFF
--- a/lib/Conversion/TorchToStablehlo/Basic.cpp
+++ b/lib/Conversion/TorchToStablehlo/Basic.cpp
@@ -2091,6 +2091,206 @@ LogicalResult ConvertAtenOp<AtenTrilOp>::matchAndRewrite(
   return success();
 }
 
+template<>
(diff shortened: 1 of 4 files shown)


william0021224:
@Vremold Could you please help to review this PR?

lib/Conversion/TorchToStablehlo/Basic.cpp
```diff
@@ -2091,6 +2091,206 @@ LogicalResult ConvertAtenOp<AtenTrilOp>::matchAndRewrite(
   return success();
 }
 
+template <>
+LogicalResult ConvertAtenOp<AtenScaledDotProductAttentionOp>::matchAndRewrite(
```
qingyunqu, line 2095: How about writing it as a decomposition? So it could be reused by linalg/tosa backend?

william0021224: Sounds a good idea~,
 I will check how to implement it

PR #3407, opened 2024-05-31
[NFC] Fix member cast change to global for landing collision

rsuderman:
A PR landed when moving away from a deprecated cast function. Updated the corresponding lines to pass.


renxida (approved):
Diff is a lil messed up because of the format changes but LGTM

rsuderman:
> Diff is a lil messed up because of the format changes but LGTM

Yeah it looks like the original commit bypassed the commit hooks. We were failing on pre-commit hooks for formatting.

zjgarvey (commented):
Yeah, let's get this merged.

PR #3408, opened 2024-05-31
add onnx loop support

PhaneeshB:
- Adds limited support for lowering onnx.Loop to primLoopOp
- lower in the pipeline`torch-to-scf` there is a check to see if loop is for like. A primLoopOp is for like when the input condition is a `trueBoolConstant`. To adapt the onnx to torch lowering to take advantage of it, the implementation checks for specific op patterns in the loodBody region and decides if loop is for like and uses the right input condition op.
- to adapt the onnxLoopBody to torchLoopBody, we need to adapt the input block arguments and set the correct output condition variable in the loop body.
- scanOutput variables are currently not supported.


zjgarvey:
@PhaneeshB There's a new patch which will fix the CI for you.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -257,6 +258,152 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        // results - n loop carreid dependencies and k scan outputs
+        // Fail when there are scanOutputs in onnxLoop (K>0);
+        // unsupported for now
+        if (terminator->getNumOperands() !=
+            loop.getRegion().getNumArguments() - 1) {
+          return failure();
+        }
+
+        auto terminatorOperands = terminator->getOperands();
+        llvm::SmallVector<Value> restOps(terminatorOperands.begin() + 1,
```
renxida, line 384: would be nice for this to get a more informative name. What exactly do these values do?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -257,6 +258,152 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        PatternRewriter::InsertionGuard guard(rewriter);
+        Operation *terminator = loop.getRegion().front().getTerminator();
+        rewriter.setInsertionPoint(terminator);
+
+        // results - n loop carreid dependencies and k scan outputs
+        // Fail when there are scanOutputs in onnxLoop (K>0);
+        // unsupported for now
+        if (terminator->getNumOperands() !=
+            loop.getRegion().getNumArguments() - 1) {
+          return failure();
```
renxida, line 380: instead of failure, useing rewriter.notifyMatchFailure(op, "failure_reason") would make debugging much easier in the future.

renxida (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
PhaneeshB: It would make debugging easier; Added a failure reason for this and other places too. Thanks

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
PhaneeshB: Yes, this is essentially to get all remaining operands from the loop body's terminator Op and passing it to PrimLoopCondition Op
renamed and added a comment for the same.

As I understand, primLoop Op can't take give out more results number of values than iter_args and fails at an assertion when replacing the Op, hence this is unsupported for now

vivekkhandelwal1 (changes requested):
Hi @PhaneeshB, I don't see any lit test for this lowering. Can you please add atleast one such test?

PhaneeshB:
DLRM model is using scanouputs in its usage of Loop op, 
making it a draft PR while adding support for it.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -257,6 +258,158 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+              .getValue()
+              .str();
+        };
+
+        // PrimLoop Op expectes inputCondition to be boolConstantTrue
+        // to decide if the loopOp is `forlike`. Use loopIsForLike to
+        // ensure appropriate inputCondition is set
+        // Case 1 : loopCondInp -> identity -> terminator(loopCondOut)
+        bool loopIsForLike = false;
+        auto case1ForLike = [&getOpName](Region *loopBody) -> bool {
```
renxida, line 322: Are there other cases?

I see a case 1 but nothing else.

Also, why identity?

I would expect onnx forlikes to have loopCondInp to feed directly to the terminator. Where does the identity come from?

renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
PhaneeshB: Yes, there are other cases. 
Each case is essentially a way of identifying the ways in which one can pass a bool_true as the input condition and how is it handled inside before passing it on to terminator.
Another case is using the same onnx_constant_bool_true, as passed to the onnx.Loop Op,
Another way could be casting an int to bool inside the loop-body and then passing to terminator, and so on..
(did see another way in DLRM's for-loop can't recollect now though)

I've just added a mechanism to catch all such case as we see them in usecases and replace with a single const_bool_true in torch-mlir, hoping that the other part will be removed in DCE in later passes.

> Also, why Identity?

Turns out this was just part of the onnx.Loop example in the documentation. It also seems something like an unsaid rule in onnx.Loop examples to clone all the inputs of Loop Body before usage, hence just picked them up so make sure we cover the (sort of) "native" cases :) Maybe they have their reasons for it which I'm unaware of (clearly)

In this particular [example - loop_11](https://github.com/onnx/onnx/blob/main/docs/Operators.md#examples-84), they clone the condition before passing it on to the terminator op.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -259,6 +260,158 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        }
+
+        // MaxTripCount - tensor int64 scalar (or empty)
+        Value maxTripCountTensor = operands[0];
+        auto maxTripCountInt = rewriter.create<Torch::AtenItemOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            maxTripCountTensor);
+
+        // Condition - tensor bool scalar (or empty)
+        Value conditionTensor = operands[1];
```
vivekkhandelwal1, line 292: Before doing this, you should check if this value exists or not. Since, the condition above only make sure that there should be > 0 operands, this operand could still be missing.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -259,6 +260,158 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        auto condArg = loop.getRegion().front().getArgument(1);
+        if (!condArg.use_empty())
+          condArg.replaceAllUsesWith(conditionTensor);
+
+        // replace terminator
+        PatternRewriter::InsertionGuard guard(rewriter);
+        Operation *terminator = loop.getRegion().front().getTerminator();
+        rewriter.setInsertionPoint(terminator);
+
+        // results - n loop carreid dependencies and k scan outputs
```
vivekkhandelwal1, line 380: Nit: `Carried`

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -259,6 +260,158 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        // Fail when there are scanOutputs in onnxLoop (K>0);
+        // unsupported for now
+        if (terminator->getNumOperands() !=
+            loop.getRegion().getNumArguments() - 1) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "scanOutputs in loop body unsupported");
+        }
+
+        // Get remaining operands from onnxLoopBody's terminator Op
+        // these are all the loop carried denpendencies in the loop body
```
vivekkhandelwal1, line 390: Nit: `dependencies`

vivekkhandelwal1 (changes requested):
Some small changes need to be done.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -9,6 +9,7 @@
 
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Utils.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
```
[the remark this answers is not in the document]
PhaneeshB: Oops! Artefacts of auto-include 😅

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
vivekkhandelwal1: > Actually onnx.Loop does have this constraint on the number of inputs
> 
> > Inputs
> > Between 2 and 2147483647 inputs.
> 
> But one can't be too careful these days, adding an extra check :)

Yes, there's a constraint on the number of inputs but that's for the `v_initial` arg, not the `cond` arg (it's optional and can be absent) which you're defining here: `Value conditionTensor = operands[1]`.  My comment is for this specific arg.

vivekkhandelwal1 (approved):

PR #3380, opened 2024-05-22
[ONNX] Add OnnxToTorch lowering for Onnx.NegativeLogLikelihoodLoss Op

123epsilon:
This implements the Onnx.NegativeLogLikelihoodLoss op using the signature provided [here](https://onnx.ai/onnx/operators/onnx__NegativeLogLikelihoodLoss.html) by replacing it with a `NLLLossForward` op.

Additionally, I included a helper function `get_loss_reduction_enum` to convert from a string `reduction` parameter to the corresponding intended integer value since this is an operation that will be reused for any loss function module. This differs from `get_reduction_enum` in `TorchUpstream.cpp` which handles the `reduce` parameter from `scatter_reduce` type operations.

PR #3409, opened 2024-05-31
Create MLIR functions for ONNX operators that are functions

### LINKED ISSUE #3384: [ONNX] Systematically expand ONNX functions before conversion to Torch to avoid needing bespoke conversions
Currently there are [a large number of operators](https://github.com/nod-ai/SHARK-Turbine/issues/215) in ONNX that aren't supported by TorchOnnxToTorch. As a beginner, I was tasked with adding a conversion for one of those, namely [MeanVarianceNormalization](https://onnx.ai/onnx/operators/onnx__MeanVarianceNormalization.html), in https://github.com/nod-ai/SHARK-Turbine/issues/697. But I noticed that this operator is actually a function, and operators that are functions are defined by [a series of other ONNX operators](https://onnx.ai/onnx/intro/concepts.html#functions). In this case, all those other operators are ones we already have conversions for. So I wondered why we don't expand these operator functions before converting, to avoid needing bespoke conversions for all of them.

I discussed a bit with some colleagues (@ScottTodd, @rsuderman) and it doesn't seem like there's a good reason not to do this, so I'm going to make an attempt at it. Here's some things we've learned that might be relevant:

- The ONNX API has a function called [`FunctionExpandHelper`](https://github.com/onnx/onnx/blob/093a8d335a66ea136eb1f16b3a1ce6237ee353ab/onnx/defs/function.cc#L18) ~~(in Python:  `function_expand_helper`)~~ that seems to be intended for exactly this kind of expansion. Here are some places it has been used:
  - ONNX [uses _a similar Python function_ for generating expanded versions of tests](https://github.com/onnx/onnx/blob/b90e252da11dea9bdc191d6b9b8d01511ef3e3bd/onnx/backend/test/case/node/__init__.py#L154-L164). For the example of MeanVarianceNormalization you can follow the flows of:
    -  [original Python](https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/meanvariancenormalization.py) -> (no expansion) -> [generated ONNX](https://github.com/onnx/onnx/blob/main/onnx/backend/test/data/node/test_mvn/model.onnx) -> [MLIR](https://github.com/nod-ai/SHARK-TestSuite/blob/main/iree_tests/onnx/node/generated/test_mvn/model.mlir)
    -  [original Python](https://github.com/onnx/onnx/blob/main/onnx/backend/test/case/node/meanvariancenormalization.py) -> (expansion) -> [generated ONNX](https://github.com/onnx/onnx/blob/main/onnx/backend/test/data/node/test_mvn_expanded/model.onnx) -> [MLIR](https://github.com/nod-ai/SHARK-TestSuite/blob/main/iree_tests/onnx/node/generated/test_mvn_expanded/model.mlir)
  -  ONNX also has a [basic test of it](https://github.com/onnx/onnx/blob/b90e252da11dea9bdc191d6b9b8d01511ef3e3bd/onnx/test/cpp/function_verify_test.cc#L342)
  - OpenVINO [uses it in its ONNX frontend](https://github.com/openvinotoolkit/openvino/blob/00856cdcb21de2a966b8e48cf64cf26041cbe06c/src/frontends/onnx/frontend/src/core/transform.cpp#L60)
-  ONNX also has an [inlining API](https://github.com/onnx/onnx/blob/093a8d335a66ea136eb1f16b3a1ce6237ee353ab/onnx/inliner/inliner.h#L41-L51) but it's not clear to me if it's suitable for this use.
- Trivia: MeanVarianceNormalization seems to be the poster child for an ONNX operator that is a function! The [guide on adding new ops](https://onnx.ai/onnx/repo-docs/AddNewOp.html) gives it as an example of something that should be a function rather than a primitive operator; it predates functions actually being operators and then got [promoted](https://github.com/onnx/onnx/commit/3a9a87871021a02d6b970c1e26ffa4c0dd720677#diff-edac4191612f111be4f5d6f83d690f8d15236c8068161c281f6ebe70cec81d93R8894); and it is even used for [testing function expansion](https://github.com/onnx/onnx/blob/b90e252da11dea9bdc191d6b9b8d01511ef3e3bd/onnx/test/cpp/function_verify_test.cc#L342). :)

Regarding actually implementing an expansion in our importer or converter, some things to keep track of:

- Once we have expansion working, there should be an assertion somewhere to catch cases where operators we expect to be expanded are somehow still present in the IR.
- In case we want to have a bespoke conversion for some particular operator, there should be some allowlist/denylist setup to prevent some operators being expanded, so they can be converted later.
- There should be some instrumentation on the expansion, so it's easy to see which operators are getting expanded, enabling decision-making about possible future special-casing.

If we do implement an expansion, then:

- Should we get rid of any existing bespoke lowerings that are now redundant? I know there's at least one. Maybe something for follow-up work.

andfau-amd:
Resolves https://github.com/llvm/torch-mlir/issues/3384.

Please read the commit message for more information!

I'd really appreciate any feedback! I am completely new to this. :)


andfau-amd:
Marking this as ready for review now. `check-torch-mlir`, `check-torch-mlir-python`, and the CI (including e2e tests) work now. I'm reasonably confident in the quality of the code, and I would appreciate some feedback.

However, I probably need to add some kind of regression testing for this before it can be merged. I'm also going to run some external testing.

This might be a disruptive change, as discussed in the commit message. A different denylisting or allowlisting strategy might be worth discussing?

projects/pt1/python/torch_mlir_e2e_test/configs/onnx_backend.py
```diff
@@ -92,7 +92,7 @@ def _module_lowering(
     # Lower from ONNX to Torch
     run_pipeline_with_repro_report(
         torch_mod,
-        f"builtin.module(func.func({ONNX_TO_TORCH_FUNC_PIPELINE}))",
+        f"builtin.module(inline, func.func({ONNX_TO_TORCH_FUNC_PIPELINE}))",
```
stellaraccident, line 95: This needs a comment. Something like:

```
# ONNX import produces multiple functions (for custom operators, etc), but the reference backend does not support calls currently. Therefore, since this is just for testing, we unconditionally inline.
```

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -631,6 +758,218 @@ def tensor_proto_to_attr(self, tp: onnx.TensorProto) -> Attribute:
...
+                specific_version
+            )
+        if not function_proto_str:
+            raise OnnxImportError(
+                f"Function lookup for {op_name}/{op_domain}/{specific_version}/{is_context_dependent} failed unexpectedly. This probably indicates a bug."
+            )
+        function_proto = onnx.onnx_pb.FunctionProto()
+        function_proto.ParseFromString(function_proto_str)
+
+        # An ONNX function may be polymorphic, parameterized over the types of
```
stellaraccident, line 877: I'd need to study the case further, but it feels like there should be a more unified way of doing this without an ONNX proto rewrite. But I get the issue and why this might be desired/obtuse to do the other way.

Perhaps more of a mechanical ask: Can you separate out the `convert_function_to_model_proto` vs inlining such a complicated transform like this?

stellaraccident (commented):
First pass review. I feel like something complicated like this should also have a lit test that verifies at least a simple structural case case more explicitly.

See for example: https://github.com/llvm/torch-mlir/blob/main/test/python/onnx_importer/import_onnx_tool.runlit#L3

This can be annoyingly hard in ONNX, but if you can identify at least one simple example of function outlining like this, then it would be good to back it up with a lit test like that as it makes it obvious what the corresponding IR structure imports as. Will also make it easier to port to the C version.

projects/pt1/python/torch_mlir_e2e_test/configs/onnx_backend.py
andfau-amd: Right, I was wondering about adding a comment here. How does this sound?

> The importer may produce MLIR functions corresponding to ONNX functions. These need to be inlined before conversion so it won't choke on them.

python/torch_mlir/extras/onnx_importer.py
andfau-amd: > I'd need to study the case further, but it feels like there should be a more unified way of doing this without an ONNX proto rewrite. But I get the issue and why this might be desired/obtuse to do the other way.

Are you imagining a way that doesn't involve creating a whole new model, or something else?

> Perhaps more of a mechanical ask: Can you separate out the convert_function_to_model_proto vs inlining such a complicated transform like this?

Ah, so then `get_operator_function` would only do the caching, schema lookup, and final conversion parts? I think I can do that.

andfau-amd:
> I feel like something complicated like this should also have a lit test that verifies at least a simple structural case case more explicitly.
> 
> See for example: https://github.com/llvm/torch-mlir/blob/main/test/python/onnx_importer/import_onnx_tool.runlit#L3
> 
> This can be annoyingly hard in ONNX, but if you can identify at least one simple example of function outlining like this, then it would be good to back it up with a lit test like that as it makes it obvious what the corresponding IR structure imports as.

Oh, somehow I hadn't realised I could use LIT tests here. I can definitely add a few of those!

> Will also make it easier to port to the C version.

Ah, right, do the two importers need to stay in sync? Ideally I would have wanted to implement a lot of this in C++, and then the two importers could perhaps share it, but I don't know if that's possible without putting that C++ into ONNX itself or something.

stellaraccident:
> > I feel like something complicated like this should also have a lit test that verifies at least a simple structural case case more explicitly.
> > See for example: https://github.com/llvm/torch-mlir/blob/main/test/python/onnx_importer/import_onnx_tool.runlit#L3
> > This can be annoyingly hard in ONNX, but if you can identify at least one simple example of function outlining like this, then it would be good to back it up with a lit test like that as it makes it obvious what the corresponding IR structure imports as.
> 
> Oh, somehow I hadn't realised I could use LIT tests here. I can definitely add a few of those!
> 
> > Will also make it easier to port to the C version.
> 
> Ah, right, do the two importers need to stay in sync? Ideally I would have wanted to implement a lot of this in C++, and then the two importers could perhaps share it, but I don't know if that's possible without putting that C++ into ONNX itself or something.

The C one gets used for ORT integration and is just starting point code for that use case. I've just aimed at keeping the importers relatively structural and eating the cost of keeping them in sync. 

The python version can go a lot of places the C one can't and vica-versa. As an example, the python one doesn't infect every consuming project with a native protobuf/absl dependency. That itself pays for the code duplication.

andfau-amd:
Hmm. As an aside, there's quite a few things in this code that are effectively working around deficiencies in the ONNX Python APIs. We could potentially simplify it a bit and make it easier to keep it in sync with the C++ importers if some improvements and additions to ONNX could be upstreamed.

python/torch_mlir/extras/onnx_importer.py
ScottTodd, line 1: <sub>(commenting here for threading)</sub>

> However, I probably need to add some kind of regression testing for this before it can be merged. I'm also going to run some external testing.

I regenerated some test cases using this PR in SHARK-TestSuite here: https://github.com/nod-ai/SHARK-TestSuite/pull/251

The process was sorta tricky so I wanted to try it myself before asking that of you:

* Patch this commit into IREE's torch-mlir submodule
* Build IREE's Python bindings by following https://iree.dev/building-from-source/getting-started/#python-bindings and ensure the PYTHONPATH points to the IREE build
* Update the import tool to run from the source build instead of an install
* Run the test case importer
* Commit changes
* Push to test on CI (just CPU there, but we also test on various GPUs in IREE itself)

Some example test cases to compare
* `NegativeLogLikelihoodLoss`, e.g. [before](https://github.com/nod-ai/SHARK-TestSuite/blob/main/iree_tests/onnx/node/generated/test_nllloss_NCd1_mean_weight_negative_ii/model.mlir) --> [after](https://github.com/ScottTodd/SHARK-TestSuite/blob/test-3409/iree_tests/onnx/node/generated/test_nllloss_NCd1_mean_weight_negative_ii/model.mlir) then note that `test_nllloss_NCd1_mean_weight_negative_ii` was in the `expected_compile_failures` test group but now it passes

ScottTodd: If you want to run one of those tests yourself step by step or are just curious how they are setup,

Take https://github.com/ScottTodd/SHARK-TestSuite/tree/test-3409/iree_tests/onnx/node/generated/test_nllloss_NCd1_mean_weight_negative_ii for example.

* Import (this PR affects this step directly):
  `iree-import-onnx [onnx repo]/onnx/backend/test/data/node/test_nllloss_NCd1_mean_weight_negative_ii/model.onnx -o /tmp/model.mlir`
* Compile:
  `iree-compile --iree-hal-target-backends=llvm-cpu /tmp/model.mlir -o /tmp/model_cpu.vmfb`
* Run, checking outputs against the reference values from the onnx test suite:
  `iree-run-module --device=local-sync --module=/tmp/model_cpu.vmfb --flagfile=[test suite repo]/iree_tests/onnx/node/generated/test_nllloss_NCd1_mean_weight_negative_ii/test_data_flags.txt`

The iree_tests folder in the test suite repo just wraps those steps in a test harness (pytest): https://github.com/nod-ai/SHARK-TestSuite/tree/main/iree_tests#running-tests

stellaraccident:
> Hmm. As an aside, there's quite a few things in this code that are effectively working around deficiencies in the ONNX Python APIs. We could potentially simplify this Python importer a bit and make it easier to keep it in sync with the C++ importer if some improvements and additions to ONNX could be upstreamed.

Example? Most of the messiness I've seen is just silly twenty year old protobuf-is-really-not-a-great-ir issues. Most of that stuff is just never going to be very good.

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -631,6 +758,218 @@ def tensor_proto_to_attr(self, tp: onnx.TensorProto) -> Attribute:
...
+            # No relevant function definition
+            return None
+        elif ncd_function_version is not None and (
+            cd_function_version is None or cd_function_version < ncd_function_version
+        ):
+            specific_version = ncd_function_version
+            is_context_dependent = False
+        else:
+            specific_version = cd_function_version
+            is_context_dependent = True
```
andfau-amd, line 826: This is one example of something that's working around a deficiency in ONNX's APIs. There are APIs for testing whether an operator has a context-dependent or non-context-dependent function, and in C++ there's even one that lets you test if there's a function for a specific version, but none of these APIs take account of the fact that if version n has a function, version n+1 also has it, nor did they handle the possibility an operator goes from non-context-dependent to context-dependent, so I had to write this thing.

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -603,6 +723,13 @@ def type_proto_to_type(self, tp: onnx.TypeProto) -> IrType:
             element_type = self.get_optional_element_type(ot.elem_type)
             return self.get_optional_type(element_type)
 
+        # Check if TypeProto is empty (sometimes happens for unused function
+        # arguments)
+        if tp.SerializeToString(
+            deterministic=True
+        ) == onnx.TypeProto().SerializeToString(deterministic=True):
```
andfau-amd, line 730: This one is awkward but that might just be Protobuf's fault. Alternatively maybe I've missed something. It really feels like there should be a better way.

python/torch_mlir/extras/onnx_importer.py
andfau-amd: My gut feeling is that this is also somewhere the ONNX APIs could be improved.

- If shape inference could be run on a function and return the types of the interior nodes, not just the return types, we wouldn't have to turn the function into a model.
- If the Python API exposed shape inference on graphs (like the C++ does), not just on functions, we could skip the creating a model, albeit that step is trivial (it's making the graph that's a lot of work).
- I don't know if anyone else has needed to turn functions into graphs, but such a transform does feel like it belongs in ONNX itself if it has to go somewhere. It's easier for it to share code and prevent breakage.
- If the Python API exposed functions for attribute binding, we wouldn't have to reimplement it. See e.g. https://github.com/onnx/onnx/blob/88f8ef15cfaa3138d336f3502aed5018d802bf43/onnx/shape_inference/attribute_binder.h

andfau-amd: I outlined the transform to a new function, `_specialize_function_and_create_model`. That's definitely helped the readability of the code and let me give some things better name, thanks for that suggestion!

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -631,6 +758,240 @@ def tensor_proto_to_attr(self, tp: onnx.TensorProto) -> Attribute:
...
+                if call_attribute.name == ref_name:
+                    new_attribute = onnx.AttributeProto()
+                    new_attribute.CopyFrom(call_attribute)
+                    new_attribute.name = node_attribute.name
+                    new_node.attribute.append(new_attribute)
+                    break
+            else:
+                # The default value is sometimes empty for optional attributes
+                # that don't have a default, in which case it is dropped.
+                # cf. https://github.com/onnx/onnx/blob/88f8ef15cfaa3138d336f3502aed5018d802bf43/onnx/shape_inference/attribute_binder.h#L21-L23
```
andfau-amd, line 825: Ah, when I looked again at this reference, I discovered that attribute binding needs to apply to subgraphs! I hadn't considered that. That might explain some of the test failures, I'll need to fix that. (Yet another reason I wish I could use ONNX's existing C++ code for this!)

stellaraccident:
I hear you. But I'm various downstreams, we are basically never going to take a dep on protobuf or onnx in the c++ code... My experience with Onnx on this stuff is limited, but it is essentially a cone of TF GraphDef, which I have a lot of sad history with. I hate to say it, but at some point you lose the "this could be better upstream" aspect and just try to ensure local sanity. For that, I just try to keep the mechanics contained to mostly write once importers that have reasonable tests. 

But if you see a chance to make the actual onnx python API better, feel free to send them a patch. The protobuf API hasn't changed much in a decade and is what it is, though. 

Onnx only releases quarterly, so it could be a while before you can actually use any contributions.

andfau-amd:
Wait, how does the importer avoid depending on ONNX?

andfau-amd:
Oh, well, the Python importer doesn't depend on C++ _directly_, but it does depend on the ONNX Python library, and that depends on ONNX's C++ code. But I think I understand: we can't add C++ code of our own to the importer, and adding stuff to ONNX would take a long time.

andfau-amd:
Uh, but more importantly, we might be able to make the C++ version of this code a bit simpler than the Python one since there's more stuff exposed by ONNX for C++, I think.

python/torch_mlir/extras/onnx_importer.py
andfau-amd: WIP in https://github.com/llvm/torch-mlir/pull/3409/commits/9426f0f6e2424a0d79deae3a4e22c0e30dd9a89d. I hate how this makes the code explosion worse.

python/torch_mlir/extras/onnx_importer.py
andfau-amd: > The process was sorta tricky so I wanted to try it myself before asking that of you

Thanks a lot for this! It was very helpful. I've managed to reproduce what you did: https://github.com/nod-ai/SHARK-TestSuite/pull/257

andfau-amd: Ah, something I'm not sure how to deal with: there are lots of failures on `main`, so I had to run CI on a revert of my changes to get a proper reference for the status quo ([config_onnx_cpu_llvm_sync-old.json](https://github.com/user-attachments/files/15788709/config_onnx_cpu_llvm_sync-old.json)) to compare with my changes (
[config_onnx_cpu_llvm_sync-new.json](https://github.com/user-attachments/files/15788711/config_onnx_cpu_llvm_sync-new.json)).

Here's some notes on what's changed. I've tried manually compiling things with `iree-compile` to see why they fail.

REMOVED `expected_compile_failures`:

- `test_mvn`. Looks like https://github.com/nod-ai/SHARK-Turbine/issues/697 is solved.
- All the `test_nllloss_*` non-`_expanded` variants. Obsoletes https://github.com/llvm/torch-mlir/pull/3380?
- Lots of `test_sce_` variants, but interestingly _only_ the `_expanded` ones. The expanded ones work now because NegativeLogLikelihoodLoss can be expanded, but it's strange that their unexpanded counterparts (that contain just a SoftmaxCrossEntropyLoss) would fail. One of them failed to compile for this reason:
  ```
  <unknown>:0: error: failed to legalize operation 'torch.constant.float'
  <unknown>:0: note: see current operation: %0 = "torch.constant.float"() <{value = 0.000000e+00 : f64}> : () -> !torch.float
  ```

REMOVED `expected_run_failures`:

- `test_clip_default_int8_inbounds`, `test_reduce_log_sum_exp_default_axes_keepdims_example`, `test_reduce_log_sum_exp_default_axes_keepdims_random`

ADDED `expected_compile_failures`:

- `test_clip_default_inbounds`, `test_clip_default_int8_inbounds` expand to `onnx.Identity`, which has "no matching versioned converter" (according to `--debug-only=dialect-conversion`). Identity seems like a trivial operation, I guess we just don't support this version of it?
- `test_dynamicquantizelinear`, `test_dynamicquantizelinear_max_adjusted`, `test_dynamicquantizelinear_min_adjusted` expand to (among other things) `onnx.ReduceMin`, "no matching versioned converter"
- Various reduction tests (`test_reduce_l2_default_axes_keepdims_example`, `test_reduce_l2_default_axes_keepdims_random`, `test_reduce_log_sum_asc_axes`, `test_reduce_log_sum_default`, `test_reduce_log_sum_desc_axes`, `test_reduce_log_sum_exp_default_axes_keepdims_example`, `test_reduce_log_sum_exp_default_axes_keepdims_random`, `test_reduce_log_sum_negative_axes`) fail with cast errors:
  ```
  0: error: 'tensor.cast' op operand type 'tensor<1x1x1xf32>' and result type 'tensor<f32>' are cast incompatible
      %1 = torch.operator "onnx.ReduceSum"(%0, %arg1) {torch.onnx.keepdims = 1 : si64} : (!torch.vtensor<[3,2,2],f32>, !torch.vtensor<[0],si64>) -> !torch.vtensor<[],f32>
           ^
  onnx/node/generated/test_reduce_l2_default_axes_keepdims_example/model.mlir:10:10: note: see current operation: %33 = "tensor.cast"(%32) : (tensor<1x1x1xf32>) -> tensor<f32>
  ```
  ```
  onnx/node/generated/test_reduce_log_sum_asc_axes/model.mlir:10:10: error: 'tensor.cast' op operand type 'tensor<f32>' and result type 'tensor<5xf32>' are cast incompatible
      %1 = torch.operator "onnx.Log"(%0) : (!torch.vtensor<[],f32>) -> !torch.vtensor<[5],f32>
           ^
  onnx/node/generated/test_reduce_log_sum_asc_axes/model.mlir:10:10: note: see current operation: %22 = "tensor.cast"(%21) : (tensor<f32>) -> tensor<5xf32>
  ```
  But it seems like their `_expanded` counterparts already failed this way, so this might not be an issue with my code?

andfau-amd: Oh, the weird error message for the non-`_expanded` SoftmaxCrossEntropyLoss test is a red herring, I forgot that I'd disabled expansion of SoftmaxCrossEntropyLoss.

andfau-amd: If I un-disable that expansion it just gets worse. Haha. More failures!
[config_onnx_cpu_llvm_sync-new-undisable-sce.json](https://github.com/user-attachments/files/15790909/config_onnx_cpu_llvm_sync-new-undisable-sce.json)

ScottTodd: We don't necessarily need to boil the ocean here, just verify that this change is a step in the right direction (and doesn't regress any critical cases). Once the initial work is landed we can continue to run down the remaining failure cases.

> * Lots of `test_sce_` variants, but interestingly _only_ the `_expanded` ones. The expanded ones work now because NegativeLogLikelihoodLoss can be expanded, but it's strange that their unexpanded counterparts (that contain just a SoftmaxCrossEntropyLoss) would fail. One of them failed to compile for this reason:
>   ```
>   <unknown>:0: error: failed to legalize operation 'torch.constant.float'
>   <unknown>:0: note: see current operation: %0 = "torch.constant.float"() <{value = 0.000000e+00 : f64}> : () -> !torch.float
>   ```

that looks like some op lowering isn't handling the f64 type. Not totally unexpected.

> * `test_clip_default_inbounds`, `test_clip_default_int8_inbounds` expand to `onnx.Identity`, which has "no matching versioned converter" (according to `--debug-only=dialect-conversion`). Identity seems like a trivial operation, I guess we just don't support this version of it?

Agreed that identity should be trivial, I ran into similar troubles before. The lowering uses a min version of 14: https://github.com/llvm/torch-mlir/blob/e07a0bfc5464c3f2cd3f3a7e2a581f55ea99e176/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp#L1578-L1590 when the op has versions for 1, 13, 14, 16, 19, 21: https://onnx.ai/onnx/operators/onnx__Identity.html

andfau-amd: How would I tell what a regression on critical cases is?

I might not be the best person to try to merge this, I don't yet have the necessary context to make such judgments. If anything is broken it's easy to denylist expansion of additional operations.

andfau-amd:
@stellaraccident I added some LIT tests in https://github.com/llvm/torch-mlir/pull/3409/commits/79ea75b91003fb9d790e5278cddacf6ba3184e11, do those look good? Could you take another look?

test/python/onnx_importer/function_expansion/GreaterOrEqual.runlit
rsuderman, line 1: Try to reduce to 1-2 new tests if possible. We should avoid excessive redundancy across tests to make supporting simpler.

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -631,6 +758,318 @@ def tensor_proto_to_attr(self, tp: onnx.TensorProto) -> Attribute:
...
+            default=None,
+        )
+        cd_function_version = max(
+            filter(f, op_schema.context_dependent_function_opset_versions),
+            default=None,
+        )
+        if ncd_function_version is None and cd_function_version is None:
+            # No relevant function definition
+            return None
+        elif ncd_function_version is not None and (
```
rsuderman, line 994: Replace `elif` with `if`. The previous returns so not point to wrap in an else.

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -149,8 +199,18 @@ def find_type_proto_for_name(self, name: str) -> onnx.TypeProto:
...
         # will associate them in the value_info. If not there, it may be a
         # graph output, which must have type information.
-        value_info = self.value_info_map.get(name) or self.output_map.get(name)
-        if value_info is not None:
+        value_info = (
+            self.value_info_map.get(name)
+            or self.output_map.get(name)
+            or self.declared_input_map.get(name)
+        )
+        if value_info is None:
```
rsuderman, line 207: Reverse this around

```
if value_info is not None:
  return value_info
  
 <tensor_proto work>
 ```

File: python/torch_mlir/extras/onnx_importer.py [hunk 8]
```diff
@@ -631,6 +758,318 @@ def tensor_proto_to_attr(self, tp: onnx.TensorProto) -> Attribute:
...
+
+            # Recurse to bind attributes on subgraphs. ONNX's implementation of
+            # attribute binding only does this for subgraphs that didn't come
+            # from a referenced attribute value, so this code doesn't either.
+            new_attribute = onnx.AttributeProto()
+            new_attribute.CopyFrom(old_attribute)
+            if new_attribute.g:
+                new_attribute.g = _bind_attributes_in_subgraph(
+                    new_attribute.g, caller_node, op_schema
+                )
```
rsuderman, line 827: end with `return new_attribute` and drop the `elif`

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -631,6 +758,318 @@ def tensor_proto_to_attr(self, tp: onnx.TensorProto) -> Attribute:
...
+        for call_attribute in caller_node.attribute:
+            if call_attribute.name == ref_name:
+                new_attribute = onnx.AttributeProto()
+                new_attribute.CopyFrom(call_attribute)
+                new_attribute.name = old_attribute.name
+                return new_attribute
+
+        # The default value is sometimes empty for optional attributes
+        # that don't have a default, in which case it is dropped.
+        if (
```
rsuderman, line 849: Check about the multi line ifs with spill over. This should be much simpler by request `op_schema.attributes[ref_name].default_value` before the `if` statement entirely and makes the conditional much simpler.

python/torch_mlir/extras/onnx_importer.py
```diff
@@ -209,11 +285,23 @@ def define_function(
                 for out in graph_info.output_map.values()
             ]
             ftype = FunctionType.get(input_types, output_types)
-            func_op = func_dialect.FuncOp(func_name, ftype, ip=InsertionPoint(body))
+            func_op = func_dialect.FuncOp(
+                func_name,
+                ftype,
+                ip=InsertionPoint(body),
+                visibility="public" if public else "private",
```
rsuderman, line 292: Lets avoid changing all imports with explicit public vs private. Keep `visibility` set to the same default before and set private in only the private cases. I looked at the testsuite run and this appears to have caused an uneeded change on all function signatures.

python/torch_mlir/extras/onnx_importer.py
rsuderman: `onnx.TypeProto().WhichOneof("value") is None`

https://protobuf.dev/reference/python/python-generated/

andfau-amd: Aha, thanks for the pointer. I'd only looked at the generated documentation for Protobuf's classes so far. I'll try this out.

test/python/onnx_importer/function_expansion/GreaterOrEqual.runlit
andfau-amd: Would it be an improvement for me to put the attribute examples into a single lit test, using the lit prefix options so i can test several models in one file?

python/torch_mlir/extras/onnx_importer.py
andfau-amd: I don't like the unnecessary change either, but the problem is that the default visibility is effectively private, and so these new functions don't get thrown away after inlining. At least, that's what happens with `torch-mlir-opt --inline`. Is there a better solution? Or, is it maybe acceptable that inlining won't clean things up? I don't know enough to answer these.

python/torch_mlir/extras/onnx_importer.py
andfau-amd: Are you suggesting an early-return style, like this?

```python
if not op_schema.attributes[ref_name].default_value:
    return None

if not op_schema.attributes[ref_name].default_value.type:
    return None
```

python/torch_mlir/extras/onnx_importer.py
andfau-amd: We discussed this privately and it turns out that default visibility is actually like _public_, and so this can be avoided, yay. :)

test/python/onnx_importer/function_expansion/GreaterOrEqual.runlit
andfau-amd: I've deleted one of the tests. Is the current number better?

andfau-amd:
> I am a little worried about seeing multiple compounding changes on top of the general supporting multiple function imports. I may not have full context but it would surprise me if we need to tweak the decompositions due to wanting additional function inclusion.

We discussed this further privately. As I understand it, Rob's concern was about disruption being caused by the function visibility change, and by the function expansion applying to all operations by default (so, it would affect operations we already support and significantly change their lowering).

We already fixed the function visibility thing.

To solve the other problem we've decided to switch to an allowlisting approach: only a tiny list of operations will be expanded by this way by default. In this patch I make that be just `MeanVarianceNormalization` (resolves https://github.com/nod-ai/SHARK-Turbine/issues/697) and `NegativeLogLikelihoodLoss` (might remove this, see https://github.com/llvm/torch-mlir/pull/3380 discussion). This way there's no immediate disruption.

I guess there can be some tracking issue for gradually expanding the allowlist.

python/torch_mlir/extras/onnx_importer.py
andfau-amd: Oh sorry, I see what you mean now.

python/torch_mlir/extras/onnx_importer.py
andfau-amd: Hmm, since I've changed this to use allowlisting, there should be no risk of regressions now, so I guess I can resolve this.

projects/pt1/python/torch_mlir_e2e_test/configs/onnx_backend.py
andfau-amd: I added a comment, I hope it's satisfactory.

rsuderman (approved):

stellaraccident (approved):

PR #3421, opened 2024-06-05
add resize nearest mode round_prefer_floor, round_prefer_ceil, ceil

aldesilv:


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2876,9 +2896,11 @@ class ConvertInterpolateOp
                 [&](OpBuilder &b, Location loc, ValueRange args) {
                   Value retVal;
                   if (mode.substr(0, 7) == "nearest") {
+                    std::string coordTfMode = mode.substr(7, mode.find(","));
+                    std::string nearestMode = mode.substr(mode.find(",") + 1);
```
zjgarvey, line 2900: If this op was coming from native torch, it would not have a comma in the mode string. Would this cause an error?

zjgarvey: I think you should change it to only have a comma if the onnx.Resize op has a nearest_mode besides floor, and then have the nearestMode defined here be `""` or the substring, conditionally on whether the comma is found.

zjgarvey (changes requested):
The only change I want to make is that the default torch path should still work with the aten_interpolate op, but because you are looking for a comma it might error out as is.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2876,9 +2896,14 @@ class ConvertInterpolateOp
                 [&](OpBuilder &b, Location loc, ValueRange args) {
                   Value retVal;
                   if (mode.substr(0, 7) == "nearest") {
+                    std::string coordTfMode = mode.substr(7, mode.find(","));
```
zjgarvey, line 2899: I think the second argument to substr should be the length of the substring rather than the end position. This might need to be something like `mode.find(",") - 7`.

zjgarvey (commented):
One last change, then lets get this merged.

zjgarvey (approved):
LGTM

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2876,9 +2896,15 @@ class ConvertInterpolateOp
                 [&](OpBuilder &b, Location loc, ValueRange args) {
                   Value retVal;
                   if (mode.substr(0, 7) == "nearest") {
+                    std::string coordTfMode =
+                        mode.substr(7, mode.find(",") - 7);
```
mgehre-amd, line 2900: The coordTfMode  is separated by mode via `_`, so shouldn't this be 
```suggestion
                        mode.substr(8, mode.find(",") - 8);
```

zjgarvey: If the coordinate tf mode isn't specified, then there won't be an underscore at the end. Starting the substring at or after npos would cause an error. We could make this a conditional or something, but I opted to just pass the underscore in with the coordTfMode (see the bilinear interpolate section above, e.g. `if (coordStr == "_asymmetric") { ...`

zjgarvey: @mgehre-amd Does my comment address your concern sufficiently? If so, I'd like to merge this PR.

mgehre-amd: Ok, looks good!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2927,6 +2929,8 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           // apparently asymmetric
           if (coordTfMode != "asymmetric" && coordTfMode != "align_corners")
             modeStr = (modeStr + "_") + coordTfMode;
+          if (nearest_mode != "floor" && nearest_mode != "")
```
mgehre-amd, line 2932: If nearest_mode is empty, I think that means it's the ONNX default (i.e. [round_prefer_floor](https://onnx.ai/onnx/operators/onnx__Resize.html)). But if the nearest_mode of the `torch.aten.__interpolate` op is empty, TorchToLinalg interprets it as floor. Should we change
`binder.customOpNameStringAttr(nearest_mode, "nearest_mode", "")` to `binder.customOpNameStringAttr(nearest_mode, "nearest_mode", "round_prefer_floor")?

mgehre-amd:
Is there a way to test onnx.Resize end2end for the modes that are not supported by pytorch?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
zjgarvey: Thanks for the catch @mgehre-amd !

zjgarvey:
> Is there a way to test onnx.Resize end2end for the modes that are not supported by pytorch?

Yeah, not in torch-mlir but we have some external test suites for this. 

@aldesilv Interestingly, it looks like the e2e tests are failing now for this op. I'm curious if this is a bad torch.onnx.export for the op (maybe they don't specify nearest_mode=floor) or if something else is going on.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2633,9 +2633,29 @@ static Value NearestInterpolate(OpBuilder &b, Location loc,
     Value outFP = b.create<arith::SIToFPOp>(loc, b.getF32Type(), outInt);
     Value proj = b.create<arith::DivFOp>(loc, outFP, scale);
 
+    Value nearestFP;
     // get nearest pixel using floor
-    Value nearestFP = b.create<math::FloorOp>(loc, proj);
-
+    if (nearestMode == "floor") {
```
zjgarvey, line 2638: this should still include the case `nearestMode = ""` since this is default behavior.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2927,6 +2929,8 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           // apparently asymmetric
           if (coordTfMode != "asymmetric" && coordTfMode != "align_corners")
             modeStr = (modeStr + "_") + coordTfMode;
+          if (nearest_mode != "round_prefer_floor" && nearest_mode != "")
```
zjgarvey, line 2932: This should still be `floor`, since that is default pytorch behavior.

zjgarvey: change `binder.customOpNameStringAttr(nearest_mode, "nearest_mode", ""))` from above to 
`binder.customOpNameStringAttr(nearest_mode, "nearest_mode", "round_prefer_floor"))`

zjgarvey (changes requested):

zjgarvey (approved):

PR #3424, opened 2024-06-06
[torch] Add OnnxToTorch lowering for the Col2Im op

vinayakdsci:
Adds OnnxToTorch lowering for the `onnx.Col2Im` op.


vinayakdsci:
cc @vivekkhandelwal1.

vinayakdsci:
@vivekkhandelwal1 gentle ping!

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -941,6 +941,144 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        auto blockShapeTy = cast<Torch::ValueTensorType>(blockShape.getType());
+        auto blockShapeSizes = blockShapeTy.getSizes();
+
+        // TODO: Add support for 5D input tensors.
+        if (imageShapeSizes[0] != 2)
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected length of imageShape to be equal to 2");
+        if (blockShapeSizes[0] != 2)
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected length of blockShape to be equal to 2");
```
vivekkhandelwal1, line 979: Before adding these checks, you need to be sure whether these sizes are statically known or not. Otherwise, in case, when they are dynamic these checks will fail.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -941,6 +941,144 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        for (int64_t i : padOnEachAxis) {
+          Value x = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(i));
+          cstPadding.push_back(x);
+        }
+        for (int64_t i : strides) {
+          Value x = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(i));
+          cstStrides.push_back(x);
+        }
```
vivekkhandelwal1, line 1019: You can do this in 1 single loop since all will run for same number of iterations.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -941,6 +941,144 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+            cstPadding);
+        Value dilationsList = rewriter.create<Torch::PrimListConstructOp>(
+            binder.getLoc(),
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+            cstDilations);
+        Value stridesList = rewriter.create<Torch::PrimListConstructOp>(
+            binder.getLoc(),
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+            cstStrides);
```
vivekkhandelwal1, line 1075: Instead of doing this, and creating the constants above you can use `createConstantList` utility.

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1501,6 +1501,46 @@ def aten〇addcmul〡shape(self: List[int], tensor1: List[int], tensor2: List[in
...
+    assert len(stride) == 2, "stride is expected to have length 2"
+    assert len(padding) == 2, "padding is expected to have length 2"
+
+    assert kernel_size[0] > 0 and kernel_size[1] > 0, "kernel size should be greater than 0"
+    assert dilation[0] > 0 and dilation[1] > 0, "dilation should be greater than 0"
+    assert padding[0] >= 0 and padding[1] >= 0, "padding should be non negative"
+    assert stride[0] > 0 and stride[1] > 0, "stride must be greater than 0"
+
+    ndim = len(self)
+    assert (ndim == 2 and self[0] != 0 and self[1] != 0) or (ndim == 3 and self[1] != 0 and self[2] != 0), "Expected 2D or 3D (batch mode) tensor for input with possibly 0 batch size and non zero dimensions for input"
```
vivekkhandelwal1, line 1524: I think this check should be at the top since it checks for the supported input size.

vinayakdsci: Fixed!

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
vinayakdsci: Ah, I did not know that this utility existed. The code is much shorter now. Thanks!

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
vinayakdsci: Completely removed by the use of `createConstantIntList()`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
vinayakdsci: I actually intended these checks to fail in that case. I hope you mean that when one of these two sizes is dynamic? It is required that the sizes be explicitly provided, especially in `torch.nn.Fold`. In Onnx, these values are tensors, so a single element tuple cannot be passed in as it can be in PyTorch.

vinayakdsci:
> Hi @vinayakdsci, please address the comments. 

Addressed in the latest commit.

>This is a well-written code with sufficient comments. Nice contribution!

Thanks so much @vivekkhandelwal1! I really appreciate it.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
vivekkhandelwal1: I meant the value `imageShapeSizes[0]` should be statically known, since it's a shape of the tensor. And the shape being dynamic mean that something like `[?,?...]` would cause an error here. So, the shape should be static not the values. Since the image and block shape are 1-d tensor you before comparing their shape to be not equal to 2, you can just check whether that is not equal to `Torch::kUnknownSize`, if that's the case then the shape is dynamic and you can add `RuntimeAssert` in that case or fail depending on how much support you want to add.

vinayakdsci: I have added a check that fails if either of the values `imageShapeSizes[0]` or `blockShapeSizes[0]` is equal to `Torch::kUnknownSize`. I hope the PR looks fine now. Thanks!

vivekkhandelwal1 (approved):
LGTM

PR #3431, opened 2024-06-07
Test custom op import with symbolic shapes

sjain-stanford:
Tests the basic constructs of registering a custom op and its abstract implementations (with FakeTensors) in python, going through TorchDynamo export, followed by importing the shape expressions in the Torch dialect.

Also fixes the importer were previously the symbolic bind op insertion was not gated in one place.


sjain-stanford:
@stellaraccident, @sjarus I'm proceeding to land this one since it is a one-line fix that would otherwise impact downstream flows that don't need symbolic shapes. The rest is just a test for custom ops with shapes. Hope that is OK, and please feel free to review post-merge.

stellaraccident (approved):

PR #3432, opened 2024-06-07
Add onnx op LRN lowering

manupak:
This commit adds support for lowering
Onnx LRN op to aten.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1446,6 +1446,121 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
                   return success();
                 });
   patterns.onOp(
+      "LRN", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value operand;
+        int64_t size;
+        float alpha, beta, bias;
+        if (binder.tensorOperand(operand) ||
+            binder.tensorResultType(resultType) ||
+            binder.s64IntegerAttr(size, "size", 2) ||
+            binder.f32FloatAttr(alpha, "alpha", 0.0001f) ||
+            binder.f32FloatAttr(beta, "beta", 0.75f) ||
+            binder.f32FloatAttr(bias, "bias", 1.0f))
+          return failure();
+        Type dtype = resultType.getOptionalDtype();
+        Value constAlpha = rewriter.create<Torch::ConstantFloatOp>(
+            binder.getLoc(), rewriter.getType<Torch::FloatType>(),
+            rewriter.getF64FloatAttr(alpha));
+        Value constBeta = rewriter.create<Torch::ConstantFloatOp>(
+            binder.getLoc(), rewriter.getType<Torch::FloatType>(),
+            rewriter.getF64FloatAttr(beta));
+        Value constBias = rewriter.create<Torch::ConstantFloatOp>(
+            binder.getLoc(), rewriter.getType<Torch::FloatType>(),
+            rewriter.getF64FloatAttr(bias));
+        // Please refer to the operator description
+        // for more info on the lowering
+        // https://onnx.ai/onnx/operators/onnx__LRN.html
+
+        // squared = operand^2
+        Location loc = binder.getLoc();
+        Torch::ValueTensorType inTy =
+            cast<Torch::ValueTensorType>(operand.getType());
+        Value sqOperand = rewriter.create<Torch::AtenMulTensorOp>(
+            loc, inTy, operand, operand);
+        // view it as n x 1 x c x d0 x d..
+        if (!inTy.hasSizes()) {
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "Expected input to have sizes");
+        }
+        ArrayRef<int64_t> inTyShape = inTy.getSizes();
+        if (inTyShape.size() < 3) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Unsupported: the input dimensions should be >= 3");
+        }
+        if (inTyShape[1] == Torch::kUnknownSize) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Unsupported: the second dimension size must be "
+                         "statically known");
+        }
+        SmallVector<int64_t, 5> viewShapeInt{inTyShape[0], 1, inTyShape[1],
+                                             inTyShape[2], Torch::kUnknownSize};
+        Torch::ValueTensorType reshapeType =
+            rewriter.getType<Torch::ValueTensorType>(viewShapeInt, dtype);
+        Value viewShapeListVal =
+            createConstantIntList(binder, rewriter, viewShapeInt);
+        auto view = rewriter.create<Torch::AtenViewOp>(
+            loc, reshapeType, sqOperand, viewShapeListVal);
+        // padding
+        int64_t highPad = (size - 1) / 2;
+        int64_t lowPad = (size - 1) - highPad;
+        SmallVector<int64_t> paddingInt{0, 0, 0, 0, lowPad, highPad};
+        auto constPadVal = rewriter.create<Torch::ConstantFloatOp>(
+            loc, rewriter.getType<Torch::FloatType>(),
+            rewriter.getF64FloatAttr(0.0));
+        Value paddingListVal =
+            createConstantIntList(binder, rewriter, paddingInt);
+        SmallVector<int64_t, 5> paddedShapeInt = viewShapeInt;
+        paddedShapeInt[2] += size - 1;
+        Torch::ValueTensorType paddedType =
+            rewriter.getType<Torch::ValueTensorType>(paddedShapeInt, dtype);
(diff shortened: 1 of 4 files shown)


manupak:
cc: @vivekkhandelwal1

lib/Dialect/Torch/Utils/Utils.cpp
```diff
@@ -617,3 +617,19 @@ Type Torch::getDefaultAccType(PatternRewriter &rewriter, Type inputType) {
...
+  });
+  return ret;
+}
+
+PrimListConstructOp Torch::getList(OpBuilder &builder, Location loc,
+                                   ArrayRef<int64_t> intVector) {
+  SmallVector<Value> vals = getValueVector(builder, loc, intVector);
+  return builder.create<Torch::PrimListConstructOp>(
+      loc, Torch::ListType::get(builder.getType<Torch::IntType>()), vals);
+}
```
vivekkhandelwal1, line 635: This is not required. There already exists a utility which does the same, here: https://github.com/llvm/torch-mlir/blob/7e0e23c66820d1db548103acbdf1337f701dc5a3/lib/Conversion/TorchOnnxToTorch/Utils.cpp#L17-L29

include/torch-mlir/Dialect/Torch/Utils/Utils.h
```diff
@@ -150,6 +152,13 @@ Type getDefaultAccType(PatternRewriter &rewriter, Type inputType);
 LogicalResult getPermutedType(BaseTensorType inType,
                               SmallVector<int64_t> permuteDims,
                               Type &permutedType);
+// A helper function to get a constant value vector from an int64_t ArrayRef
+SmallVector<Value> getValueVector(OpBuilder &builder, Location loc,
+                                  ArrayRef<int64_t> intVector);
+
+// A helper function to get a list Type from an int64_t ArrayRef
+PrimListConstructOp getList(OpBuilder &builder, Location loc,
+                            ArrayRef<int64_t> intVector);
```
vivekkhandelwal1, line 161: Remove these changes, see the comment below.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1445,6 +1445,127 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
                   return success();
                 });
+  patterns.onOp(
+      "LRN", 13, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value operand;
+        int64_t size;
+        float alpha;
+        float beta;
+        float bias;
```
vivekkhandelwal1, line 1455: They can be declared in the same line.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1445,6 +1445,127 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
                       binder.op, resultType, operand, constAlpha);
                   return success();
                 });
+  patterns.onOp(
+      "LRN", 13, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
```
vivekkhandelwal1, line 1449: You can specify the `sinceVersion` to be 1. 1 vs 13 just have the allowed dtype difference.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1445,6 +1445,127 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            rewriter.getF64FloatAttr(alpha));
+        Value constBeta = rewriter.create<Torch::ConstantFloatOp>(
+            binder.getLoc(), rewriter.getType<Torch::FloatType>(),
+            rewriter.getF64FloatAttr(beta));
+        Value constBias = rewriter.create<Torch::ConstantFloatOp>(
+            binder.getLoc(), rewriter.getType<Torch::FloatType>(),
+            rewriter.getF64FloatAttr(bias));
+        // Please refer to the operator description
+        // for more info on the lowering
+        // https://github.com/onnx/onnx/blob/main/docs/Operators.md#LRN
```
vivekkhandelwal1, line 1475: Instead of this link, you can add: https://onnx.ai/onnx/operators/onnx__LRN.html

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1445,6 +1445,127 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        if (inTyShape.size() < 3) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Unsupported: the input dimensions should be >= 3");
+        }
+        SmallVector<int64_t, 5> viewShapeInt{inTyShape[0], 1, inTyShape[1],
+                                             inTyShape[2]};
+        if (inTyShape.size() > 3) {
+          int64_t tailFlatDimSize = 1;
+          for (int64_t dimSize : inTyShape.slice(3)) {
+            tailFlatDimSize *= dimSize;
```
vivekkhandelwal1, line 1498: Before doing this, you need to make sure whether the shape values are statically known or not, otherwise this would result in issue because of unknown dim size.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1445,6 +1445,127 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          }
+          viewShapeInt.push_back(tailFlatDimSize);
+        }
+        Torch::ValueTensorType reshapeType =
+            rewriter.getType<Torch::ValueTensorType>(viewShapeInt, dtype);
+        Torch::PrimListConstructOp viewShapeListVal =
+            Torch::getList(rewriter, loc, viewShapeInt);
```
vivekkhandelwal1, line 1505: Replace `getList` with `createConstantIntList`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1445,6 +1445,127 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        auto view = rewriter.create<Torch::AtenViewOp>(
+            loc, reshapeType, sqOperand, viewShapeListVal);
```
vivekkhandelwal1, line 1507: Instead of this, you can create `AtenFlattenUsingIntsOp`.

vivekkhandelwal1 (changes requested):

lib/Dialect/Torch/Utils/Utils.cpp
manupak: Thanks!, Now I know that exist, but I changed the util to accept an ArrayRef which is more useful here.
Also, why does it live under OnnxToTorch util ? This looks to be useful util be inside the dialect (not just the conversion)

include/torch-mlir/Dialect/Torch/Utils/Utils.h
manupak: Done.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
manupak: Done.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
manupak: Done.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
manupak: Done.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
manupak: Good point. I dont think all the dims needed to be statically known (only the one being padded).
I changed the operation view to take -1 instead. It d do the right thing w/o requiring all dims to be statically known unless Im missing something here.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
manupak: Done.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
manupak: No, I dont think so.
This adds a unit dim as well as flattening. Thus, I d prefer a view op here.

lib/Dialect/Torch/Utils/Utils.cpp
vivekkhandelwal1: I think the original author might not have think about it in that way, maybe we can move it there later when required.

vivekkhandelwal1 (approved):
LGTM

Issue #3418, opened 2024-06-04
Re-enable the disabled sparse tests due to roll PyTorch


vivekkhandelwal1:
Some of the sparse tensor tests have been disabled https://github.com/llvm/torch-mlir/pull/3340 in order to get the Roll PyTorch done. It would be better if we can work on the actual issues related to these tests and get them re-enabled.

CC: @stellaraccident @aartbik

aartbik:
Note that since there was some obvious confusion on this, I added a bit more documentation to the pass description.
https://github.com/llvm/llvm-project/pull/96331

PR #3444, opened 2024-06-10
[torch-mlir][sparse] re-enable all sparse tests

aartbik:
this fixes the following issue:

https://github.com/llvm/torch-mlir/issues/3418

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -2579,6 +2579,8 @@ private:
 SmallVector<StringRef> ConvertSparseOperatorOp::legalizedNames = {
     "torch.aten._to_dense", "torch.aten._to_sparse", "torch.aten._to_csr",
     "torch.aten._to_csc",   "torch.aten._to_bsr",    "torch.aten._to_bsc",
+    "torch.aten.to_dense", "torch.aten.to_sparse", "torch.aten.to_csr",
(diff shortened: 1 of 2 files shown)


yinying-lisa-li (approved):

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -2579,6 +2579,8 @@ class ConvertSparseOperatorOp : public OpConversionPattern<OperatorOp> {
 SmallVector<StringRef> ConvertSparseOperatorOp::legalizedNames = {
     "torch.aten._to_dense", "torch.aten._to_sparse", "torch.aten._to_csr",
     "torch.aten._to_csc",   "torch.aten._to_bsr",    "torch.aten._to_bsc",
+    "torch.aten.to_dense",  "torch.aten.to_sparse",  "torch.aten.to_csr",
+    "torch.aten.to_csc",    "torch.aten.to_bsr",     "torch.aten.to_bsc",
```
PeimingLiu, line 2583: Seems to be a merge mistake?

PeimingLiu: NVM, did not see the `_`-prefix

aartbik: yeah, both in-place and other should simply feed into our SSA based IR

PR #3445, opened 2024-06-11
[ONNX] add some args to the onnx importer to assist shape_inference

zjgarvey:
Adds the following arguments:
- "--clear-domain": enabling this flag (default False) will delete the domain attribute from each node in the onnx model before importing. Shape inference does not seem to work for onnx ops in custom domains. In the rare case when these ops have a corresponding counterpart in base onnx, enabling this flag might allow shape inference to work properly. 
- "--opset-version": allows setting the opset version manually. This will cause the importer to attempt to update the opset_version of the onnx model before importing. Newer opset versions sometimes have more robust shape inference patterns.


vivekkhandelwal1 (approved):
LGTM

PR #3447, opened 2024-06-11
Support onnx.GRU and onnx.RNN

renxida:


renxida:
https://gist.github.com/renxida/22ef1b304559a47d49dfeac4602359f3

temp impl

renxida:
Just ran RNN and GRU through SHARK-TestSuite. We have numerical consistency.

lib/Conversion/TorchOnnxToTorch/OnnxGruExpander.cpp
rsuderman, line 1: We may want to merge these two files in with the LSTM expander and rename to RNN expanders. Fewer files means faster linking time.

lib/Conversion/TorchOnnxToTorch/OnnxGruExpander.cpp
```diff
@@ -0,0 +1,432 @@
...
+      SmallVector<int64_t>{seq_len, batch_size, hidden_size}, hTy.getDtype());
+
+  auto YShapeList = b.create<PrimListConstructOp>(
+      b.getType<ListType>(intType),
+      ValueRange({cstSeqLen, cstBatchSize, cstHiddenSize}));
+
+  int64_t hDtypeInt =
+      static_cast<int64_t>(getScalarTypeForType(hTy.getDtype()));
+  Value hDtypeIntVal =
+      b.create<ConstantIntOp>(loc, b.getI64IntegerAttr(hDtypeInt));
```
rsuderman, line 135: Are you sure you need the `loc` here?

lib/Conversion/TorchOnnxToTorch/OnnxRnnExpander.cpp
```diff
@@ -0,0 +1,312 @@
...
+    }
+  }
+
+  if (!binder.customOpNameStringAttr(direction, "direction", "forward") &&
+      direction != "forward")
+    return rewriter.notifyMatchFailure(binder.op,
+                                       "Unsupported direction attribute value. "
+                                       "Only 'forward' is supported but '" +
+                                           direction + "' is provided.");
+  int64_t num_directions = 1 + (direction == "bidirectional");
```
rsuderman, line 202: Better to do `direction == "bidirectional" ? 2 : 1;` There can be a lack of consistency whether `==` returns `1` or `-1`

rsuderman (approved):

lib/Conversion/TorchOnnxToTorch/OnnxGruExpander.cpp
renxida: testing without it

lib/Conversion/TorchOnnxToTorch/OnnxRnnExpander.cpp
renxida: done

lib/Conversion/TorchOnnxToTorch/OnnxGruExpander.cpp
renxida: will do

lib/Conversion/TorchOnnxToTorch/OnnxRnnExpander.cpp
```diff
@@ -0,0 +1,312 @@
...
+  int64_t hidden_size;
+  if (binder.s64IntegerAttr(hidden_size, "hidden_size"))
+    return rewriter.notifyMatchFailure(
+        binder.op, "Missing required attribute hidden_size");
+
+  auto xTy = cast<ValueTensorType>(X.getType());
+  auto wTy = cast<ValueTensorType>(W.getType());
+  Value B;
+  if (binder.tensorOperandAtIndex(B, 3)) {
+    B = b.create<AtenZerosOp>(W.getType(), W);
```
zjgarvey, line 172: From the docs, B has shape `[num_directions, 2*hidden_size]` and W has shape `[num_directions, hidden_size, input_size]`, so I'm not sure I understand why `W` is being used to create `B` here.

lib/Conversion/TorchOnnxToTorch/Utils.cpp
```diff
@@ -130,3 +131,14 @@ LogicalResult mlir::torch::onnx_c::createTorchPermuteOp(
                                                    permuteDimsList);
   return success();
 }
+
+Value mlir::torch::onnx_c::createActivationByName(ImplicitLocOpBuilder &b,
```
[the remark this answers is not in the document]
zjgarvey: Maybe it would be best to have a list of supported activations and just check the provided one is in that list at the beginning of the expander.

lib/Conversion/TorchOnnxToTorch/OnnxRnnExpander.cpp
```diff
@@ -0,0 +1,312 @@
...
+      direction != "forward")
+    return rewriter.notifyMatchFailure(binder.op,
+                                       "Unsupported direction attribute value. "
+                                       "Only 'forward' is supported but '" +
+                                           direction + "' is provided.");
+  int64_t num_directions = (direction == "bidirectional") ? 2 : 1;
+
+  auto XShape = xTy.getSizes();
+  int64_t batch_size = XShape[1];
+  int64_t input_size = XShape[2];
```
zjgarvey, line 206: Does this conversion require static dims in order to work?

zjgarvey (commented):
Hey Xida, I just had time to look over RNN today. I'll check out the GRU stuff tomorrow.

lib/Conversion/TorchOnnxToTorch/OnnxRnnExpander.cpp
renxida: Good catch. Fixing now.

lib/Conversion/TorchOnnxToTorch/OnnxRnnExpander.cpp
renxida: I think so. I need static dims to construct the big result tensor storing the intermediate hidden states of each cell.

lib/Conversion/TorchOnnxToTorch/OnnxRnnExpander.cpp
```diff
@@ -0,0 +1,312 @@
...
+        binder.op, "hidden_size (" + std::to_string(hidden_size) +
+                       ") does not match the second dimension of wTy (" +
+                       std::to_string(wTy.getSizes()[1]) + ")");
+  if (wTy.getSizes()[2] != input_size)
+    return rewriter.notifyMatchFailure(
+        binder.op,
+        "The third dimension of wTy (" + std::to_string(wTy.getSizes()[2]) +
+            ") does not match input_size (" + std::to_string(input_size) + ")");
+
+  auto getDirection = [&](int64_t direction, Value input) {
```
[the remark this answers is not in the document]
renxida: Really, the only reason is that I decided to only implement mono-directional half-way through.

Avoiding duplication sounds like a great idea.

lib/Conversion/TorchOnnxToTorch/OnnxRnnExpander.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+  Value initial_h;
+  if (binder.tensorOperandAtIndex(initial_h, 5)) {
+    initial_h =
+        b.create<AtenZerosOp>(hTy, hShape, cstDtype, cstNone, cstNone, cstNone);
+  }
+
+  Value initial_h_forward = getDirection(0, initial_h);
+
+  if (num_directions != 1) {
```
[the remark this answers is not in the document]
renxida: done!

lib/Conversion/TorchOnnxToTorch/OnnxGruExpander.cpp
```diff
@@ -0,0 +1,431 @@
...
+  }
+
+  Value X, W, R, B, initial_h, sequence_lens;
+  if (binder.tensorOperandAtIndex(X, 0) || binder.tensorOperandAtIndex(W, 1) ||
+      binder.tensorOperandAtIndex(R, 2))
+    return rewriter.notifyMatchFailure(binder.op,
+                                       "Missing required input tensor");
+
+  if (binder.tensorOperandAtIndex(B, 3))
+    B = b.create<AtenZerosOp>(W.getType(), W);
```
zjgarvey, line 207: This needs a fix similar to the RNN expander. From the docs:

"The bias tensor for the gates. Concatenation of `[Wb[zrh], Rb[zrh]]` and `[WBb[zrh], RBb[zrh]]` (if bidirectional) along dimension 0. This tensor has shape `[num_directions, 6*hidden_size]`. "

lib/Conversion/TorchOnnxToTorch/OnnxGruExpander.cpp
```diff
@@ -0,0 +1,431 @@
...
+    return rewriter.notifyMatchFailure(binder.op,
+                                       "Unsupported layout attribute type.");
+
+  std::string direction;
+  if (!binder.customOpNameStringAttr(direction, "direction", "forward") &&
+      direction != "forward")
+    return rewriter.notifyMatchFailure(binder.op,
+                                       "Unsupported direction attribute value");
+
+  int64_t num_directions = 1 + (direction == "bidirectional");
```
zjgarvey, line 244: I think Rob had a comment about lines like this. Maybe change to a ternary expression.

lib/Conversion/TorchOnnxToTorch/OnnxGruExpander.cpp
renxida: Fixed!

lib/Conversion/TorchOnnxToTorch/OnnxGruExpander.cpp
renxida: Looked through them a bit. Merging them may be annoying because there are a lot of similar-but-not-quite things going on with the slicing and such.

Might be a good idea to not couple them together.

lib/Conversion/TorchOnnxToTorch/Utils.cpp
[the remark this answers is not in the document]
renxida: I think that's a good idea, but i'm not sure how to do a "return rewriter.notifyMatchFailure" with an implicitLocOpBuilder and that may require me to re-do a lot of things if i wanna expose a rewriter to createActivationByName...

renxida:
Thinking about the remaining problems with this PR.

- the 3 ops GRU, LSTM, RNN are in different files and slows down linking a bit
- there is some code duplication but with minor differences between the three

I think it's probably possible to use 3 different xxx_cell functions
and 3 different OnnxXxxExpander functions
but share the xxx_layer functions.

But I don't expect anybody to come in later and try to make the same changes to all 3 functions. I expect somebody coming in to try to fix a problem with e.g. LSTM and I would like for them to only have to look at one file.

There are many minor differences between the Onnx specs for all three functions. I think it might be best to keep the three decoupled and not try to merge them into a single file / refactor them to share functions.

Plus we don't quite have a test infrastructure and it's a little annoying to validate the changes.

I think steps forward probably look like:
- get this PR in since we do have correct numeric test results for the simple case
- update testing infrastructure to have reliable testing in CI
- add e2e, numeric test cases for the various features (bias, direction, etc)
- xfail them, move on to models.
- if we have future models failing that use LSTM / RNN / GRU, come back and fix them.
@rsuderman what do you think

renxida:
https://github.com/nod-ai/SHARK-TestSuite/pull/267/files

test cases here. we have numerical correctness at least for the basic cases.

PR #3448, opened 2024-06-11
Bump LLVM to llvm/llvm-project@0ee71124f4eb47f1f8d

aartbik:


aartbik:
I fixed the in-source uses of applySignatureConversion() to adhere to the new signature, but StableHLO needs to do the same, so I am waiting for a fix upstream before I can bump StableHLO into the next stable situation.

stellaraccident:
Looks like someone else carried forward with both. Closing this one: https://github.com/llvm/torch-mlir/pull/3454

(They asked for review and was going to deduct, but accidentally landed it, so we'll just roll with that one)

PR #3435, opened 2024-06-07
[ONNX] add onnx lowering for global lp pool operator

umangyadav:
Solves https://github.com/nod-ai/SHARK-Turbine/issues/727

Uses AvgPool to implement GlobalLpPool similar to this https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_lp_pool.py

cc: @vivekkhandelwal1

PR #3449, opened 2024-06-12
[ONNX] Add onnx parser for LpPool operator

umangyadav:
Similar to https://github.com/llvm/torch-mlir/pull/3435

Solves https://github.com/nod-ai/SHARK-Turbine/issues/728

CC: @vivekkhandelwal1  @zjgarvey


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1363,6 +1363,147 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            cstPadding);
+        Value stridesList = rewriter.create<Torch::PrimListConstructOp>(
+            binder.getLoc(),
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+            cstStrides);
+        Value cstCeilMode =
+            rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), ceilMode);
+        // onnx lp pool doesn't have countIncludePad attribute
+        Value cstCountIncludePad =
+            rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), false);
```
zjgarvey, line 1472: I believe this should this be set to `true` so that the avg pool op uses `divisor = numElements` along the border in the 1D padded case.

zjgarvey (commented):
Nice. This looks good to me except the comment below. How are the e2e tests going?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1363,6 +1363,147 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        }
+        for (int64_t i : strides) {
+          cstStrides.push_back(rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(i)));
+        }
+        Value kernelSizeList = rewriter.create<Torch::PrimListConstructOp>(
+            binder.getLoc(),
+            Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
+            cstKernel);
+        Value paddingList = rewriter.create<Torch::PrimListConstructOp>(
```
vivekkhandelwal1, line 1460: Instead of doing this all here, you can use `createConstantList` utility.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1363,6 +1363,147 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+              "padding list size does not match twice the number of axes");
+        }
+        if (binder.s64IntegerArrayAttr(
+                strides, "strides", llvm::SmallVector<int64_t>(rank - 2, 1))) {
+          return failure();
+        }
+        if (strides.size() != rank - 2) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "strides list size does not match the number of axes");
+        }
```
vivekkhandelwal1, line 1425: Also, you can combine all the binders and fail separately for each arg.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1363,6 +1363,147 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+              paddingList, cstCeilMode, cstCountIncludePad,
+              /*divisor_override=*/cstOne);
+        } else if (rank == 5) {
+          avgPool = rewriter.create<Torch::AtenAvgPool3dOp>(
+              binder.getLoc(), resultType, pow, kernelSizeList, stridesList,
+              paddingList, cstCeilMode, cstCountIncludePad,
+              /*divisor_override=*/cstOne);
+        } else {
+          return failure();
+        }
```
vivekkhandelwal1, line 1498: You can place this condition at the top after the rank of the input is known following the LLVM standard code practice. https://www.llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
umangyadav: good catch. made changes. thanks.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
umangyadav: Used it for lists except kernelSizeList since i need to calculate 'numElements`

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
umangyadav: Done.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
umangyadav: moved this earlier. Thanks.

zjgarvey (approved):
Thanks for adding the abs!

vivekkhandelwal1 (approved):

PR #3450, opened 2024-06-12
[onnx] Resize supports default-valued attributes

sjarus:
Handles onnx exporters emitting default-valued attributes.


zjgarvey (approved):
Good idea, thanks for the change.

PR #3451, opened 2024-06-12
Implement lowering of torch.aten.triu_indices

BaneTrifa:
Closes [nod-ai/SHARK-Turbine/issues/709](https://github.com/nod-ai/SHARK-Turbine/issues/709)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -5088,3 +5088,38 @@ LogicalResult BindSymbolicShapeOp::verify() {
...
+    return emitOpError("row must be non-negative, got ") << row;
+
+  if (col < 0)
+    return emitOpError("col must be non-negative, got ") << col;
+
+  // Check if dtype is valid
+  int64_t dtype;
+  if (!matchPattern(getDtype(), m_TorchConstantInt(&dtype)))
+    return success();
+  if (dtype != 3 && dtype != 4)
```
ramiro050, line 5120: nit: compare to `torch_upstream::ScalarType::Int` and `torch_upstream::ScalarType::Long` to make it clear what dtypes are being considered

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,306 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
...
+             Value layout, Value device, Value pin_memory, Type type) {
+
+  auto arrangeType = getTensorTypeFromShapeValues({size}, type);
+  return rewriter.create<AtenArangeOp>(loc, arrangeType, size,
+                                       /*dtype=*/dtype, /*layout=*/layout,
+                                       /*device=*/device,
+                                       /*pin_memory=*/pin_memory);
+}
+
+void _getTrilSizes(PatternRewriter &rewriter, Location loc, int64_t row,
```
ramiro050, line 745: nit: we don't follow the Python "underscore helper functions" convention here. `getTrilSizes` is fine. We do, however, set all helper functions as `static` functions

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,306 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
 };
 } // namespace
 
+Value arange(PatternRewriter &rewriter, Location loc, Value size, Value dtype,
```
ramiro050, line 735: nit: I don't think this is needed as a helper function since there is not much going on other than creating the `AtenArangeOp`. I think it is better to keep the `rewriter.create<AtenArangeOp>` explicit every time it is needed in the code

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,306 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
...
+  return rewriter.create<AtenArangeOp>(loc, arrangeType, size,
+                                       /*dtype=*/dtype, /*layout=*/layout,
+                                       /*device=*/device,
+                                       /*pin_memory=*/pin_memory);
+}
+
+void _getTrilSizes(PatternRewriter &rewriter, Location loc, int64_t row,
+                   int64_t col, int64_t offset,
+                   /*return values*/ int64_t &trapezoidSize,
+                   int64_t &rectangleSize) {
```
ramiro050, line 748: nit: passing a reference to the result values is usually useful when the function returns an error message, but that is not the case here. I find it more readable to return a `std::pair` of `int`s, and avoid having implicit mutation of inputs to the function.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,306 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
...
+
+    // Convert to dtype
+    Type int64Type = rewriter.getIntegerType(/*width=*/64, /*isSigned=*/true);
+
+    auto rowInds1Type = cast<BaseTensorType>(rowInds1.getType());
+    ArrayRef<int64_t> sizes = rowInds1Type.getSizes();
+    Type finalRowType = rowInds1Type.getWithSizesAndDtype(sizes, int64Type);
+    rowInds1 = rewriter.create<AtenToDtypeOp>(
+        loc, finalRowType, rowInds1, dtype,
+        /*non_blocking*/ cstFalse, /*copy*/ cstFalse,
```
ramiro050, line 976: nit: the convention for these comments is `/*var_name=*/var` (add `=` and remove whitespace)

projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -1746,6 +1746,36 @@ def aten〇_embedding_bag〡shape(weight: List[int], indices: List[int], offsets
      return _embedding_bag_helper(weight, indices, offsets, include_last_offset,
                                  mode, per_sample_weights, padding_idx)
 
+def aten〇triu_indices〡shape(row: int, col: int, offset: int = 0, dtype: Optional[int] = 4, layout: Optional[int] = None, device: Optional[device] = None, pin_memory: Optional[bool] = None) -> List[int]:
```
ramiro050, line 1749: this function has some complicated logic. Can you add tests to it? See the function below it for an example of how to use `@check_shape_function`. These tests are run whenever you run the script to generate the MLIR for the functions

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,306 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
...
+  int64_t triuSize = row * col - (trapezoidSizeTril + rectangleSizeTril);
+  int64_t trapezoidSizeInt = triuSize - rectangleSizeInt;
+
+  // Create Value from int
+  trapezoidSize = rewriter.create<Torch::ConstantIntOp>(
+      loc, rewriter.getI64IntegerAttr(trapezoidSizeInt));
+  rectangleSize = rewriter.create<Torch::ConstantIntOp>(
+      loc, rewriter.getI64IntegerAttr(rectangleSizeInt));
+  mFirstRow = rewriter.create<Torch::ConstantIntOp>(
+      loc, rewriter.getI64IntegerAttr(mFirstRowInt));
```
ramiro050, line 824: nit: I would remove this wrapping in MLIR `Value`s to outside the function for two reasons:

1. The current use of this helper is followed by a `matchPattern` to unwrap the `Value`s, so it makes sense to keep the values as C++ `int`s until they are no longer needed in such a form
2. Returning just C++ `int`s means that both helpers (`getTrilSizes` and `getTriuSizes`) have consistent signatures

ramiro050 (changes requested):
Thanks for this! Some initial, mostly stylistic comments. Have yet to think about the algorithm used

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
BaneTrifa: Instead of `std::pair` I used `std::tuple`, because functions `getTriuSizes `and `getTrilSizes `have 3 return values.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,293 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
...
+    Value cstMinusTwoFloat = rewriter.create<Torch::ConstantFloatOp>(
+        loc, rewriter.getF64FloatAttr(-2.0));
+
+    // Calculte trapezoidSize, rectangleSize and mFirstRow
+    std::tuple<int64_t, int64_t, int64_t> triuSizes =
+        getTriuSizes(rewriter, loc, rowInt, colInt, offsetInt);
+
+    int64_t trapezoidSizeInt = std::get<0>(triuSizes);
+    int64_t rectangleSizeInt = std::get<1>(triuSizes);
+    ;
```
ramiro050, line 869: nit: remove extra semicolon. there's another one two lines down

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,293 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
...
+  // Number of elements in top rectangle - calculate rectangle size
+  int64_t minimum = (row < -offset) ? row : -offset;
+  int64_t rectangleSize = (minimum * col > 0) ? minimum * col : 0;
+
+  // Number of elements in bottom trapezoid - calculte trapezoid size
+  std::tuple<int64_t, int64_t, int64_t> trilSizes =
+      getTrilSizes(rewriter, loc, row, col, offset - 1);
+  int64_t trapezoidSizeTril = std::get<0>(trilSizes);
+  int64_t rectangleSizeTril = std::get<1>(trilSizes);
+  ;
```
ramiro050, line 794: nit: extra semi-colon

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,293 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
...
+  // Number of elements in bottom rectangle - rectangleSize
+  int64_t diffRow = nRowAll - nRowTrapezoid;
+  int64_t rectangleSize = (diffRow * col > 0) ? diffRow * col : 0;
+
+  // Create return value
+  return std::make_tuple(trapezoidSize, rectangleSize, mFirstRow);
+}
+
+static std::tuple<int64_t, int64_t, int64_t>
+getTriuSizes(PatternRewriter &rewriter, Location loc, int64_t row, int64_t col,
```
ramiro050, line 774: `rewriter` and `loc` are not used

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,293 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
 };
 } // namespace
 
+static std::tuple<int64_t, int64_t, int64_t>
+getTrilSizes(PatternRewriter &rewriter, Location loc, int64_t row, int64_t col,
```
ramiro050, line 736: This function is not needed. It is only used to calculate `getTriuSizes`, and it is more concise to implement `getTriuSizes` directly compared to having both helper functions.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -732,6 +732,293 @@ class DecomposeAtenTriuOp : public OpRewritePattern<AtenTriuOp> {
...
+
+  // Number of elements in bottom rectangle - rectangleSize
+  int64_t diffRow = nRowAll - nRowTrapezoid;
+  int64_t rectangleSize = (diffRow * col > 0) ? diffRow * col : 0;
+
+  // Create return value
+  return std::make_tuple(trapezoidSize, rectangleSize, mFirstRow);
+}
+
+static std::tuple<int64_t, int64_t, int64_t>
```
ramiro050, line 773: Can you add a small description of what this function does? In particular, what do the outputs represent?

ramiro050 (changes requested):

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
BaneTrifa: I was thinking about that but then I saw that there is another operator torch.tril_indices, which needs to be implemented, uses the function getTrilSizes. So I decided to separate getTrilSizes because it'll be used during lowering of torch.tril_indices.

ramiro050: I see, if the lowering for `tril_indices` is future work, then this is fine.

ramiro050 (approved):
Thanks!

PR #3456, opened 2024-06-13
Add aten._scaled_dot_product_flash_attention_for_cpu to default decomposition table

william0021224:
decomposition table for fx/dynamo

Let torch do the decomposition for
aten._scaled_dot_product_flash_attention_for_cpu when importing through Fx/Dynamo. By doing so, some SDPA related e2e tests can be marked as expectly success


william0021224:
@stellaraccident Could you please help to review this PR? Thanks!

stellaraccident (approved):
Should be fine since downstreams manage their decompositions themselves (and will have their own approach to this op).

william0021224:
> Should be fine since downstreams manage their decompositions themselves (and will have their own approach to this op).

@stellaraccident Thanks for your review. BTW, Could you please help to merge this PR, because I don't have write access?

PR #3458, opened 2024-06-13
[ONNX] Add OnnxToTorch Lowering for GroupNormalization op

vivekkhandelwal1:

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1696,6 +1696,28 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
             binder.f32FloatAttr(epsilon, "epsilon", 0.00001f) ||
             binder.s64IntegerAttr(stashType, "stash_type", 1))
           return failure();
+
+        // Since the support for `stash_type` arg does not exist in
+        // the torch op so we just check for the stash_type to be same
+        // as the input dtype since that won't require us to do any
+        // input type conversion and hence can be supported.
+        auto xType = cast<Torch::ValueTensorType>(x.getType());
+        std::optional<int64_t> stashTypeIntTorch =
+            onnxDtypeIntToTorchDtypeInt(stashType);
+        if (!stashTypeIntTorch.has_value())
+          return rewriter.notifyMatchFailure(
+              binder.op, "unimplemented support for the given stash_type");
+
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -2322,4 +2343,54 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            binder.op->getContext(),
+            (torch_upstream::ScalarType)stashTypeIntTorch.value());
+        if (failed(stashDtype))
+          return failure();
+        auto inputDtype =
+            cast<Torch::ValueTensorType>(input.getType()).getOptionalDtype();
+        if (*stashDtype != inputDtype)
+          return rewriter.notifyMatchFailure(
+              binder.op, "unimplemented: stash_type should be same "
+                         "as the input dtype");
```
zjgarvey, line 2382: nit:
```suggestion
        if (*stashDtype != inputDtype)
          return rewriter.notifyMatchFailure(
              binder.op, "unimplemented: stash_type != input dtype");
```

zjgarvey (approved):
Besides a nit suggestion, this looks good to me. 

On a side note:

It looks like adding support for `stash_type` is feasible. It could be implemented with the following steps:

1. cast `input` to `stash_type`
2. do `AtenGroupNormOp` on the casted input with the modifications: `/*scale*/=cstOne`, `/*bias*/=zerosLikeBias`
3. cast the result of the `AtenGroupNormOp` back to `inputDtype`
4. do `scale*casted_result + bias`

If you feel like this should just be done in a separate patch, that's totally fine with me.

vivekkhandelwal1:
> Besides a nit suggestion, this looks good to me.
> 
> On a side note:
> 
> It looks like adding support for `stash_type` is feasible. It could be implemented with the following steps:
> 
> 1. cast `input` to `stash_type`
> 2. do `AtenGroupNormOp` on the casted input with the modifications: `/*scale*/=cstOne`, `/*bias*/=zerosLikeBias`
> 3. cast the result of the `AtenGroupNormOp` back to `inputDtype`
> 4. do `scale*casted_result + bias`
> 
> If you feel like this should just be done in a separate patch, that's totally fine with me.

Sure, it can be done this way but I thought this might not be considered an optimal approach. Also, there would be rarely any instances where the stash_type would be different from the input dtype. Still, we can do it this way, I can add this in a follow-up patch.

PR #3459, opened 2024-06-13
[TorchToLinalg] Fix Quantized Convolution Accumulator Type

zjgarvey:
1. truncates zero-points to i32
2. modifies the default accumulator type for i8 from i64 to i32. 
3. now uses the input dtype to infer accumulator dtype.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -803,6 +803,8 @@ public:
       inputZp = typeConverter->materializeTargetConversion(
           rewriter, loc, typeConverter->convertType(inputZp.getType()),
           inputZp);
+      inputZp =
+          rewriter.create<arith::TruncIOp>(loc, rewriter.getI32Type(), inputZp);
       auto torchDtype = cast<ValueTensorType>(make.getType()).getDtype();
       inputUnsigned = torch_to_linalg::isUnsignedTorchType(torchDtype);
     }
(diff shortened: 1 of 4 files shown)


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -803,6 +803,8 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
       inputZp = typeConverter->materializeTargetConversion(
           rewriter, loc, typeConverter->convertType(inputZp.getType()),
           inputZp);
+      inputZp =
+          rewriter.create<arith::TruncIOp>(loc, rewriter.getI32Type(), inputZp);
```
vivekkhandelwal1, line 807: How did we decide on whether to keep it i32 or i64?

zjgarvey: This is the accumulator type for i8 quantized convolutions. I haven't seen any other kinds of quantized convs yet.

zjgarvey: Here's a link to the linalg op generating code:

<https://github.com/llvm/llvm-project/blob/c6b6e18c4d25305ab98b6eab752de99ea4e15344/mlir/python/mlir/dialects/linalg/opdsl/ops/core_named_ops.py#L859>

vivekkhandelwal1 (approved):

PR #3460, opened 2024-06-13
[ONNX] Add missing "Abs" in GlobalLpPool 

umangyadav:
Taking `abs` is required to mimic same logic as onnx/onnxruntime. 
Without `abs`, it wouldn't produce correct results for negative values. 

Reference code : 
https://github.com/microsoft/onnxruntime/blob/f5b6f6dc26a55ddf7523d832ac5dc56930225264/onnxruntime/core/providers/cpu/nn/pool_functors.h#L604

https://github.com/onnx/onnx/blob/375c161c67855fea9612c15b83ebff40fca838a4/onnx/reference/ops/op_lp_pool.py#L31

CC : @zjgarvey  @vivekkhandelwal1

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -9,6 +9,7 @@
 
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Utils.h"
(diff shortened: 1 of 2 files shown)


zjgarvey (approved):
I assume the change to opset version 2 is because that's when p changed from float to int, correct?

umangyadav:
> I assume the change to opset version 2 is because that's when p changed from float to int, correct?

Yes

zjgarvey:
Would you like this to get merged right away?

umangyadav:
> Would you like this to get merged right away?

I am not sure what it is the approval process. If this PR satisfy all merging criteria then yes i would like to get this merged.

PR #3461, opened 2024-06-14
[Torch]Decompose AtenScaledDotProductAttentionOp

william0021224:
Decompose "aten.scaled_dot_product_attention" according to pytorch document at:
https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html#torch.nn.functional.scaled_dot_product_attention

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -21,6 +21,7 @@
 #include "llvm/ADT/StringExtras.h"
 #include "llvm/ADT/StringSet.h"
 #include <cstdint>
+#include <optional>
 #include <set>
 
 using namespace mlir;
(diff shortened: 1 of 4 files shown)


william0021224:
@stellaraccident Could you please help to review this PR? Thanks!

stellaraccident (changes requested):
I don't think we can accept this patch as a default pattern. Most production backends benefit from the op not being decomposed, and only if a backend needs it should it be opted in.

This needs to be in some kind of optional pass that the ref backend activates as part of its pipeline.

I'll need to defer to others for an opinion on the right way to organize it though. @rsuderman, @ramiro050

william0021224:
> I don't think we can accept this patch as a default pattern. Most production backends benefit from the op not being decomposed, and only if a backend needs it should it be opted in.
> 
> This needs to be in some kind of optional pass that the ref backend activates as part of its pipeline.
> 
> I'll need to defer to others for an opinion on the right way to organize it though. @rsuderman, @ramiro050

No Problem

PR #3463, opened 2024-06-14
onnx.DFT basic support

PhaneeshB:
- adds support for DFT v20 on the FFT and IFFT path
- adds required skeleton code for IFFT ops to be recognised in TMlir


lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -2570,4 +2575,71 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          ftOp = rewriter.create<Torch::AtenFftFftOp>(
+              binder.getLoc(),
+              Torch::ValueTensorType::get(
+                  binder.op->getContext(), resultType.getSizes(),
+                  mlir::ComplexType::get(inType.getDtype())),
+              inComplexTensor, /*n = */ dftLength, /*dim = */ axis,
+              /*norm = */ norm);
+        } else {
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "Inverse FFT unimplemented");
```
vivekkhandelwal1, line 2639: You can add the support for inverse as well by lowering it to https://pytorch.org/docs/stable/generated/torch.fft.ifft.html.

vivekkhandelwal1 (changes requested):
Please add atleast one lit test corresponding to each lowering path of the op with the lowering that's a convention which we follow for all the Onnx op lowerings.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -7,12 +7,20 @@
...
+#include "mlir/IR/ValueRange.h"
+#include "mlir/Support/LLVM.h"
+#include "mlir/Transforms/DialectConversion.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Utils.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchTypes.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/Support/FormatVariadic.h"
+#include <cstdint>
```
vivekkhandelwal1, line 23: All these includes are not required.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -2570,4 +2578,98 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          // Default in torch is -1 and onnx is -2 (since -1 is for real / img)
+          axis = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(-2));
+        }
+        if (binder.s64IntegerAttr(inverse, "inverse", 0) ||
+            binder.s64IntegerAttr(onesided, "onesided", 0) ||
+            binder.tensorResultType(resultType)) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "bind failed for attrs / resultType");
+        }
```
vivekkhandelwal1, line 2613: These binds can be combined with the bind at the top for the input.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
PhaneeshB: Oops! Artefacts of auto-include 😅

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
PhaneeshB: Makes sense to make the code a tiny bit more compact
Although as it stands now the order of binding inputs / attributes is the same as they appear in the onnx DFT documentation.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -7,6 +7,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+#include "mlir/IR/Attributes.h"
+#include "mlir/IR/BuiltinTypes.h"
```
vivekkhandelwal1, line 11: Are you sure that these are needed? Or, you just missed removing them?

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -2583,4 +2585,95 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        Value norm = rewriter.create<Torch::ConstantStrOp>(
+            binder.getLoc(), rewriter.getStringAttr(Twine("backward")));
+        // Convert from [....., 2] complex number repr for fft consumption.
+        Torch::ValueTensorType inType =
+            binder.toValidTensorType(inTensor.getType());
+        int64_t lastIndex = inType.getSizes().back();
+        if (lastIndex != 1 && lastIndex != 2)
+          return rewriter.notifyMatchFailure(
+              binder.op,
+              "Expected input tensor to have dims [..., 1] or [..., 2]");
```
vivekkhandelwal1, line 2632: This check would not behave as expected when the corresponding input shape dim is dynamic. Before this check, or along with this add a check for that dim to be static.

vivekkhandelwal1 (approved):
Except two comments, everything else looks fine.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
PhaneeshB: Thanks for the review
So the idea here is that if the last dim is neither of 1 or 2, then we exit with a failure as 
there is a requirement for [torch.view_as_complex](https://pytorch.org/docs/stable/generated/torch.view_as_complex.html#torch-view-as-complex) for the last dim to be exactly 2 (neither 1, nor dynamic)

What this check achieves is that it will error out on last_dim not 2
we are allowing last_dim = 1 because later we pad it with zeros and make the last dim 2.

Hope that clears it up.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
PhaneeshB: Auto-include crept back in it seems 😓

PR #3073, opened 2024-03-28
[README] update links to snapshot packages

vmiheer:
Source: https://github.com/llvm/torch-mlir/issues/3068#issuecomment-2024109412 Verified commands locally on Ubuntu 22.04 with pyenv virtualenv created for python 3.11.

PR #3466, opened 2024-06-17
Fixed installation command in README.md

Hacker1337:
Current pip installation command raises error
```
ERROR: Could not find a version that satisfies the requirement torch-mlir (from versions: none)
ERROR: No matching distribution found for torch-mlir
```
(checked on Ubuntu 22.04.2 LTS with `venv` and with `conda`)

Because it is trying to install torch-mlir from pytorch repository. The installation command was wrongly split into 2 in #3073. I just merged them back to 1 installation command with both pytorch and llvm/torch-mlir channels.


marbre (approved):
Thanks!

stellaraccident (approved):

PR #3469, opened 2024-06-18
[RFC][ONNX] Basic Support for DeformConv

zjgarvey:
This adds a torchvision op to torch-mlir and a path from onnx.DeformConv to torchvision.deform_conv2d. 

I'm not implementing the torch->linalg lowering for the torchvision op yet, but posting this PR to get feedback on some of the choices being made here and to flesh out the onnx frontend a bit.


zjgarvey:
I'll add a lit test for onnx->torch soon!

zjgarvey:
A few things for reviewers to check out in-particular:

1. I wrote a custom torch->onnx export next to the e2e test which relies on it. Since the op is not in native torch, I don't think it makes sense to upstream this? If this should be removed, or moved to a different file, lets discuss. 
2. I updated the opset version for torch.onnx.export to the maximum allowable value for the given torch install. This was to accomodate newer opset versions (like 19) which support deformconv and apparently some bitwise ops which are now version-conditionally passing.

rsuderman (approved):

vivekkhandelwal1:
Hi @zjgarvey, can you please resolve the conflicts?

PR #3472, opened 2024-06-19
[ONNX] Add OnnxToTorch lowering for OptionalHasElement op

vivekkhandelwal1:

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -2731,4 +2731,41 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
                       binder.op, tensorListResultType, input);
                   return success();
                 });
+  patterns.onOp(
+      "OptionalHasElement", 15,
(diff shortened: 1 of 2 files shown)


renxida:
Looks like this is an op with no corresponding torch op.

Is that why we don't have an e2e test for this?

otherwise, looks good to me.

renxida:
Looks like iree has a test case for this.

https://github.com/iree-org/iree/blob/6f178692f11d356a284eba915f2cf1d067d92a69/build_tools/pkgci/external_test_suite/onnx_cpu_llvm_sync.json#L250

When this gets merged & torch-mlir gets bumped, might be a good idea to make sure that test case is passing

renxida (approved):

vivekkhandelwal1:
> Looks like iree has a test case for this.
> 
> https://github.com/iree-org/iree/blob/6f178692f11d356a284eba915f2cf1d067d92a69/build_tools/pkgci/external_test_suite/onnx_cpu_llvm_sync.json#L250
> 
> When this gets merged & torch-mlir gets bumped, might be a good idea to make sure that test case is passing

Sure, I will do that.

PR #3476, opened 2024-06-20
[Torch] Fix torch.constant.int operation parsing

cmcgirr-amd:
Due to the custom operation parser, the print and parser were expecting two different forms. 

One having the dictionary before the value and the other after. Following the format of the other constants ops, the constant.int will follow the `value attr-dict` format. Updated the parser accordingly.


ramiro050 (approved):
Nice catch!

PR #3477, opened 2024-06-20
Support fake_quantize_per_tensor_affine_cachemask

mgehre-amd:
Add a new op with shape/dtypes and decompose into `fake_quantize_per_tensor_affine` when the second result is unused.

The xfail_set change is on ONNX because torch cannot export this op to ONNX.


zjgarvey (approved):
This looks good to me! I didn't know you could use the `replaceAllUsesWith` and then just erase the op since the other results aren't being used. I'm going to steal this and use it in another PR right now, so thanks for the lesson.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -8375,6 +8400,8 @@ class DecomposeComplexOpsPass
     addPatternIfTargetOpIsIllegal<DecomposeAtenReshapeAsOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenTriuOp>(patterns);
     addPatternIfTargetOpIsIllegal<DecomposeAtenLinalgNormOp>(patterns);
+    addPatternIfTargetOpIsIllegal<
+        DecomposeAtenFakeQuantizePerTensorAffineCachemaskOp>(patterns);
```
zjgarvey, line 8404: Actually, on a second glance, does the op need to be made illegal in [/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp](https://github.com/llvm/torch-mlir/blob/694210f429e44977c35fb5d8890f63b89aaad0ed/lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp#L369)?

mgehre-amd: done

PR #3480, opened 2024-06-20
TorchOnnxToTorch: Fix stack-use-after-free

mgehre-amd:
We used to move the SmallVector into an ArrayRef and then the SmallVector left the scope.

Found by asan.


stellaraccident (approved):
Thanks

PR #3483, opened 2024-06-20
[TorchToLinalg] remove `extract_slice` grid_sample lowering

IanWood1:
Instead of using extract_slice for grid sampler, use affine constants to access the X and Y values in the generic op's region.


AmosLewis:
https://github.com/iree-org/iree/pull/17692

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2486,25 +2454,30 @@ class ConvertAtenGridSamplerOp : public OpConversionPattern<AtenGridSamplerOp> {
...
+        loc, TypeRange{resultType}, ValueRange(), ValueRange(emptyOp), gridMaps,
+        gridIterators, [&](OpBuilder &b, Location loc, ValueRange args) {
+          Value iterIdxZero = b.create<linalg::IndexOp>(loc, 0);
+          Value iterIdxTwo = b.create<linalg::IndexOp>(loc, 2);
+          Value iterIdxThree = b.create<linalg::IndexOp>(loc, 3);
+
+          Value gr1 = createExtract(b, loc, grid, iterIdxZero, iterIdxTwo,
+                                    iterIdxThree, zeroIndex);
+          Value gr0 = createExtract(b, loc, grid, iterIdxZero, iterIdxTwo,
+                                    iterIdxThree, oneIndex);
```
zjgarvey, line 2480: Is it possible to have the same input twice in a linalg generic op but with different indexing maps? Or does that kind of thing cause the same problem as before?

```
gr1_indexing_map = <(d0,d1,d2,d3) -> (d0,d2,d3,1)>
gr0_indexing_map = <(d0,d1,d2,d3) -> (d0,d2,d3,0)>
```
then the ins value range would be `ValueRange{grid, grid}`?

It just seems weird to me to have tensor.extract ops inside a linalg generic, when the accessing is fairly standard.

IanWood1: Both work to solve the issue. I think this might be better to do it this way vs using affine constants. `grid_sample` is very similar to `gather`, which also uses `linalg.index`. See: https://github.com/llvm/torch-mlir/blob/3225f20ab19db12e532e51fe60a9ff78b48be880/lib/Conversion/TorchToLinalg/IndirectDataMovement.cpp#L28

vivekkhandelwal1:
Hi @IanWood1, can you please address the comment and update the PR?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
zjgarvey: Admittedly, I'm not sure how exactly iree treats generic ops, I think it will be much better to use the affine map than `tensor.extract`. The only time I would ever use `tensor.extract` in a linalg generic op payload would be if the indexing of a tensor is not affine with static striding/offsetting. E.g., something like interpolation where the extraction point needs to be computed in a more complicated way, or something like adaptive pooling where the strides and window sizes are dynamically computed from the tensor shapes.

IanWood1: I would agree that using affine maps makes more sense. That's how i first tried to solve this, so I added that commit to this branch.

zjgarvey (approved):

zjgarvey:
@IanWood1 If you don't have merge permissions, let me know and I'll squash and merge this when you are ready.

IanWood1:
@zjgarvey I just got them last week, but thank you. and thanks for the review

PR #3486, opened 2024-06-21
[fx importer] support fx importer with lower version torch

qingyunqu:


python/torch_mlir/extras/fx_importer.py
```diff
@@ -151,11 +151,16 @@
     torch.complex32: "complex<f16>",
     torch.complex64: "complex<f32>",
     torch.complex128: "complex<f64>",
-    torch.float8_e5m2: "f8E5M2",
-    torch.float8_e4m3fn: "f8E4M3FN",
-    torch.float8_e5m2fnuz: "f8E5M2FNUZ",
-    torch.float8_e4m3fnuz: "f8E4M3FNUZ",
 }
+HIGH_VERSION_TORCH_DTYPE_TO_MLIR_TYPE_ASM = {
```
stellaraccident, line 155: In another code base, I called this something like OPTIONAL_TORCH_DTYPE_...

and had a comment that these are entries added only in later versions of torch.

stellaraccident (approved):
Makes sense. Suggested a naming refinement.

Issue #3428, opened 2024-06-07
Failed to lower 'torch.operator "onnx.AveragePool"' to torch (dilation)


josel-amd:
Hi all,

I need to convert 'torch.operator "onnx.AveragePool"' to torch but it fails with `dilation is not supported by torch.aten.avgpool op`

Command to reproduce:

```
torch-mlir-opt --allow-unregistered-dialect --convert-torch-onnx-to-torch dilation.mlir -debug;
```

Observed behavior
See [0]

It might be useful: I tracked the implementation back to this [PR](https://github.com/llvm/torch-mlir/pull/2672).

[0]

```
//===-------------------------------------------===//
Legalizing operation : 'torch.operator'(0x560591ccb250) {
  %1 = "torch.operator"(%arg0) <{name = "onnx.AveragePool"}> {torch.onnx.ceil_mode = 1 : si64, torch.onnx.dilations = [2 : si64, 2 : si64], torch.onnx.kernel_shape = [2 : si64, 2 : si64], torch.onnx.strides = [1 : si64, 1 : si64]} : (!torch.vtensor<[1,1,4,4],f32>) -> !torch.vtensor<[1,1,2,2],f32>

  * Fold {
  } -> FAILURE : unable to fold

  * Pattern : 'torch.operator -> ()' {
Trying to match ""
    ** Failure : dilation is not supported by torch.aten.avgpool op
: conversion failed to apply: "onnx.AveragePool", sinceVersion=11
    ** Failure : no matching versioned converter
"" result 0
  } -> FAILURE : pattern failed to match
} -> FAILURE : no matched legalization pattern
//===-------------------------------------------===//
dilation.mlir:4:10: error: failed to legalize operation 'torch.operator' that was explicitly marked illegal
    %0 = torch.operator "onnx.AveragePool"(%arg0) {torch.onnx.ceil_mode = 1 : si64, torch.onnx.dilations = [2 : si64, 2 : si64], torch.onnx.kernel_shape = [2 : si64, 2 : si64], torch.onnx.strides = [1 : si64, 1 : si64]} : (!torch.vtensor<[1,1,4,4],f32>) -> !torch.vtensor<[1,1,2,2],f32>
         ^
[...]

PR #3490, opened 2024-06-21
[ONNX] Add averagepool dilations support

AmosLewis:
To fix dilations issue: https://github.com/llvm/torch-mlir/issues/3428

Test by: https://github.com/nod-ai/SHARK-TestSuite/pull/268
Status report for run: test-run using mode:onnx todtype:default backend:llvm-cpu
| tests                      | model-run   | onnx-import   | torch-mlir   | iree-compile   | inference   |
|:---------------------------|:------------|:--------------|:-------------|:---------------|:------------|
| onnx/operators/AveragePool | passed      | passed        | passed       | passed         | passed      |

### DIFF
--- a/lib/Conversion/TorchToLinalg/Pooling.cpp
+++ b/lib/Conversion/TorchToLinalg/Pooling.cpp
@@ -612,6 +612,16 @@ public:
                                                   strideInts, paddingInts)))
       return rewriter.notifyMatchFailure(op, "invalid pooling parameters");
 
+    // Decode strideInts into strideInts and dilation
+    if (strideInts.size() == 2 * Dim) {
+      for (int i = 0; i < Dim; i++) {
+        dilationInts[i] = strideInts[Dim + i];
+      }
+      for (int i = 0; i < Dim; i++) {
+        strideInts.pop_back();
+      }
+    }
+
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -612,6 +612,16 @@ class ConvertAtenAvgPoolOp : public OpConversionPattern<OpTy> {
                                                   strideInts, paddingInts)))
       return rewriter.notifyMatchFailure(op, "invalid pooling parameters");
 
+    // Decode strideInts into strideInts and dilation
+    if (strideInts.size() > Dim) {
```
zjgarvey, line 616:
```suggestion
    if (strideInts.size() == 2*Dim) {
```
Since you are extracting up to `2*Dim -1`, I think this would be better.

zjgarvey (commented):
Nice. Since the linalg ops support dilations anyway, I think this is the best way to go.

lib/Conversion/TorchToLinalg/Pooling.cpp
AmosLewis: Done

zjgarvey (approved):

PR #3498, opened 2024-06-25
Implement lowering of torch.aten.atleast_1d

pkapris-syrmia:
This operator is necessary in order to implement torch.aten.vstack. Which will be added in a future PR.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1516,6 +1516,44 @@ class DecomposeAtenReshapeOp : public OpRewritePattern<AtenReshapeOp> {
...
+//     return res if len(res) > 1 else res[0]
+class DecomposeAtenAtleast1dOp : public OpRewritePattern<AtenAtleast1dOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenAtleast1dOp op,
+                                PatternRewriter &rewriter) const override {
+    Value input = op.getSelf();
+    Location loc = op.getLoc();
+    Type opType = op.getType();
+    auto resType = cast<BaseTensorType>(input.getType());
```
vivekkhandelwal1, line 1542: Instead of naming it as `resType, name it as `inpType`.

vivekkhandelwal1 (approved):
Except one small comment, everything else looks good.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
pkapris-syrmia: Just fixed it, thank you very much for the review!

PR #3501, opened 2024-06-27
[ONNX] Add OnnxToTorch lowering for NonMaxSuppression op

vivekkhandelwal1:

https://github.com/nod-ai/SHARK-Turbine/issues/650


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -2923,4 +2923,23 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+  patterns.onOp("NonMaxSuppression", 10,
+                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+                  Torch::ValueTensorType resultType;
+                  SmallVector<Value> operands;
+                  if (binder.tensorOperandsList(operands) ||
+                      binder.tensorResultType(resultType))
+                    return failure();
+
+                  // TODO: Add support for handling max_output_boxes_per_class
+                  // and score_threshold arg.
```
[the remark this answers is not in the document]
zjgarvey: Ah, they are optional inputs. These actually don't get added as operands if they are left unspecified. 

e.g., with all of the optional arguments for this op left blank, here's a sample:

```MLIR
module {
  func.func @nms_graph(%arg0: !torch.vtensor<[1,6,4],f32>, %arg1: !torch.vtensor<[1,1,6],f32>) -> !torch.vtensor<[3,3],si64> attributes {torch.onnx_meta.ir_version = 10 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "", torch.onnx_meta.producer_version = ""} {
    %none = torch.constant.none
    %0 = torch.operator "onnx.NonMaxSuppression"(%arg0, %arg1) : (!torch.vtensor<[1,6,4],f32>, !torch.vtensor<[1,1,6],f32>) -> !torch.vtensor<[3,3],si64> 
    return %0 : !torch.vtensor<[3,3],si64>
  }
}
```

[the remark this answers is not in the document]
zjgarvey: This is actually somewhat annoying for this op, since hoping that `iouThreshold = operands[3]` is not reliable (it might even cause a crash when used on the example above).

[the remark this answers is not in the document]
zjgarvey: I'd say that we need a few match failures:

1. if there are three operands, and the last operand has a float dtype, we cannot infer which argument it is.
2. if there are four operands, and there is an int and a float optional, then we cannot infer which argument is missing.
3. if center_point_box is not zero (default) we should indicate a match failure since this is not handled. 

I'm not sure what to do about the other optional arguments. This is a rather confusing onnx op.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -2923,4 +2923,23 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+                  Type scalarTy = rewriter.getType<Torch::FloatType>();
+                  Value iouThreshold = rewriter.create<Torch::AtenItemOp>(
+                      binder.getLoc(), scalarTy, operands[3]);
+
+                  rewriter.replaceOpWithNewOp<Torch::TorchvisionNmsOp>(
+                      binder.op, resultType, /*boxes=*/operands[0],
```
mgehre-amd, line 2941: `boxes` and `scores` have shapes Nx4 and N, respectively. for torchvision.nms but shapes BxNx4 and BxCxN in ONNX. Shouldn't we need some handling here to align those?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -2926,4 +2926,96 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            binder.getLoc(), maxScores, scoreThreshold);
+        rewriter.create<Torch::RuntimeAssertOp>(
+            binder.getLoc(), scoresCond,
+            rewriter.getStringAttr(
+                "unimplemented: score_threshold should be >= max(scores)"));
+
+        Value iouThreshold = rewriter.create<Torch::AtenItemOp>(
+            binder.getLoc(), rewriter.getType<Torch::FloatType>(), operands[3]);
+        rewriter.replaceOpWithNewOp<Torch::TorchvisionNmsOp>(
+            binder.op, resultType, boxes, scores, iouThreshold);
```
zjgarvey, line 3018: The result type of onnx.NonMaxSuppression doesn't match TorchvisionNmsOp. It looks like nms has output shape `[n]` where `n` is the number of selected indices out of `N` (which I assume corresponds to onnx's `spatial_dimension`). The onnx op has output shape `[n, 3]` (the three is the tuple `batch_index`, `channel_index` , and `box_index` -- or I assume `spatial_index` by their docs), so we might need to concatenate the torchvision result with a `torch.zeroes([n,2])`, since this conversion will only work when the batches and channels are both 1.

I'm wondering if these two ops are sufficiently different to consider something else entirely. I know there is a [batched_nms](https://pytorch.org/vision/main/generated/torchvision.ops.batched_nms.html), but it's in a completely different format. For onnx boxes with shape `[B,N]`, we would need to flatten it to shape `[B*N]`, and make a torchvision `idxs` tensor which is like `N` copies of `arange(0,B)` flattened to shape `[B*N]`. Then, in the onnx `scores` tensor of shape `[B, C, N]`, extract each channel one at a time and perform batched_nms. Here is some pseudocode:

```python
flattened_boxes = boxes.flatten()
idxs = arange(0, B).unsqueeze(1).repeat(1, N).flatten()
# or maybe easier:
idxs = floor(arange(0, B, 1/N))
onnx_result = tensor.empty(resultShape)
count = 0
for c in range(C):
   channelscores = scores[ : , c , : ]
   flattenedscores = channelscores.flatten() # shape = [B*N]
   result = torchvision.batched_nms(flattenedboxes, flattenedscores, idxs, iou_threshold)
   #result contains the values indexing into [B*N] corresponding to the selected boxes
   for res_index in result:
      batch_idx = idxs[res_index]
      channel_ix = c
      box_idx = res_index % batch_idx
      onnx_result[count, 0] = batch_idx
      onnx_result[count, 1] = channel_idx
      onnx_result[count, 2] = box_idx
      count++
return onnx_result
```

Frankly, I'm just guessing how these ops line up since the documentation is a bit limited. It might be good to compare something like the python pseudocode implementation above with the onnx runtime result and see if this is even a viable approach.

zjgarvey: I'd generate an onnx model with the helpers and build an onnxruntime inference session (like we do in the shark-testsuite model.py files).

If you have the python impl to compare with, see if it gives the same results.

vivekkhandelwal1: Hi @zjgarvey, I have modified the original lowering since it's already implicitly checking for B,C to be 1. Hence, we can support this op now for limited cases and can extend it in future, if required.

zjgarvey (changes requested):
I took another look at the runtime asserts and I think they need modification. Sorry for not catching this earlier.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -2984,4 +2984,145 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            rewriter, binder.op, binder.getLoc(), 0, squeezedScores.value());
+        if (failed(squeezedScores))
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "failed to squeeze scores tensor");
+
+        boxes = squeezedBoxes.value();
+        scores = squeezedScores.value();
+
+        // TODO: Add support for handling score_threshold arg.
+        // If score_threshold < min(scores) then the op can't be lowered since
```
zjgarvey, line 3036: nit:
```suggestion
        // If score_threshold > min(scores) then the op can't be lowered since
```

zjgarvey (approved):
This looks correct to me now, thanks for the changes.

PR #3502, opened 2024-06-27
Implement lowering for torch.aten.hann_window.periodic

pkapris-syrmia:


projects/pt1/python/torch_mlir_e2e_test/test_suite/spectral.py
```diff
@@ -0,0 +1,31 @@
...
+        super().__init__()
+
+    @export
+    @annotate_args(
+        [
+            None,
+        ]
+    )
+    def forward(self):
+        return torch.ops.aten.hann_window(20, False)
```
vivekkhandelwal1, line 26: Does the support for `periodic=True` is absent?

vivekkhandelwal1: If not, then please add a test for the same.

vivekkhandelwal1 (changes requested):

projects/pt1/python/torch_mlir_e2e_test/test_suite/spectral.py
pkapris-syrmia: There is support for periodic=True. I just added the relevant test, thank you very much for the comment

vivekkhandelwal1 (approved):

Issue #3499, opened 2024-06-26
support torch.aten.diag


aartbik:
Take something like

```
func.func @forward(%arg0: !torch.vtensor<[1,1,2],f16>) -> !torch.vtensor<[1,1,2],f16> {
  %q_rhs = torch.vtensor.literal(dense<[[0, 1], [2, 3]]> : tensor<2x2xui8>) : !torch.vtensor<[2,2],ui8>
  %scales = torch.vtensor.literal(dense<1.0> : tensor<2x1x1xf16>) : !torch.vtensor<[2,1,1],f16>
  %zps = torch.vtensor.literal(dense<0.0> : tensor<2x1x1xf16>) : !torch.vtensor<[2,1,1],f16>
  %bit_width = torch.constant.int 8
  %group_size = torch.constant.int 2
  %output = torch.operator "quant.matmul_rhs_group_quant"(%arg0, %q_rhs, %scales, %zps, %bit_width, %group_size) : (!torch.vtensor<[1,1,2],f16>, !torch.vtensor<[2,2],ui8>, !torch.vtensor<[2,1,1],f16>, !torch.vtensor<[2,1,1],f16>, !torch.int, !torch.int) -> !torch.vtensor<[1,1,2],f16>
  return %output : !torch.vtensor<[1,1,2],f16>
}
```

then, running this with `--torch-backend-to-linalg-on-tensors-backend-pipeline` pipeline fails to legalize the torch.constants into the torch.operator?

```
<unknown>:0: error: failed to legalize operation 'torch.constant.int'
<unknown>:0: note: see current operation: %3 = "torch.constant.int"() <{value = 8 : i64}> : () -> !torch.int
...
func.func @forward(%arg0: tensor<1x1x2xf16>) -> tensor<1x1x2xf16> {
  %cst = arith.constant dense<[[0, 1], [2, 3]]> : tensor<2x2xi8>
  %cst_0 = arith.constant dense<1.000000e+00> : tensor<2x1x1xf16>
  %cst_1 = arith.constant dense<0.000000e+00> : tensor<2x1x1xf16>
  %int8 = torch.constant.int 8
  %int2 = torch.constant.int 2
  %0 = torch_c.from_builtin_tensor %arg0 : tensor<1x1x2xf16> -> !torch.vtensor<[1,1,2],f16>
  %1 = torch_c.from_builtin_tensor %cst : tensor<2x2xi8> -> !torch.vtensor<[2,2],ui8>
[...]

PR #2885, opened 2024-02-07
Implement linalg lowering of diag_embed torch op

schnkmwt:
This PR adds lowering of diag_embed to linalg dilect.
Tracked in https://github.com/nod-ai/SHARK-Turbine/issues/288

PR #3505, opened 2024-06-27
[torch-mlir][sparse] add decomposition features to sparse compiler

aartbik:
Fixes https://github.com/llvm/torch-mlir/issues/3499

### DIFF
--- a/python/torch_mlir/extras/fx_decomp_util.py
+++ b/python/torch_mlir/extras/fx_decomp_util.py
@@ -49,6 +49,7 @@ DEFAULT_DECOMPOSITIONS = [
     torch.ops.aten.nan_to_num.default,
     torch.ops.aten.unbind,
     torch.ops.aten._scaled_dot_product_flash_attention_for_cpu,
+    torch.ops.aten.diag,
 ]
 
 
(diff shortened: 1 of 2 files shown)


python/torch_mlir/extras/fx_decomp_util.py
```diff
@@ -49,6 +49,7 @@
     torch.ops.aten.nan_to_num.default,
     torch.ops.aten.unbind,
     torch.ops.aten._scaled_dot_product_flash_attention_for_cpu,
+    torch.ops.aten.diag,
```
ramiro050, line 52: I was confused by why this worked because there is no decomposition for this op in [torch/_decomp/decompositions.py](https://github.com/pytorch/pytorch/blob/main/torch/_decomp/decompositions.py). Turns out that PyTorch has _a lot_ of decompositions in [another file](https://github.com/pytorch/pytorch/blob/main/torch/_refs/__init__.py). Pretty nice.

Also, FWIW, I just realized that Torch-MLIR [does support `torch.ops.aten.diag_embed`](https://github.com/llvm/torch-mlir/pull/2885), which behaves exactly like `torch.diag` for 1D inputs.

ramiro050 (approved):

yinying-lisa-li (approved):

PR #3511, opened 2024-06-28
support both option -v and TORCH_MLIR_TEST_VERBOSE

qingyunqu:
so that we could run `python3 -m e2e_testing.main -v` to specify `verbose=True`


projects/pt1/python/torch_mlir_e2e_test/framework.py
```diff
@@ -361,7 +361,7 @@ def run_tests(
     try:
         env_verbose = os.getenv("TORCH_MLIR_TEST_VERBOSE", "0")
         if env_verbose is not None:
-            verbose = bool(int(env_verbose))
+            verbose = verbose | bool(int(env_verbose))
```
aartbik, line 364: some may prefer "or" here

aartbik (approved):

PR #3512, opened 2024-06-28
Fix unused variable warning from assertion variable

Max191:
Inlines a variable into an assertion that is not used elsewhere to fix build warnings.


qedawkins (commented):
you can put `[[maybe_unused]]` in front of the variable declaration instead, or just inline the `getRank()` to the assert.

PR #3513, opened 2024-06-29
[Torch] support decomposition of aten.aminmax

qingyunqu:
* unify decompisition of `aten.amax` and `aten.amin`
* support `aten.amax` with `dim=()`


Vremold (approved):
LGTM.

PR #3514, opened 2024-06-30
[Torch Dialect] improve argmax/argmin's decomposition to support keep…

qingyunqu:
…dim=True when dim=None


Vremold (approved):
LGTM, thanks for the effort.

PR #3521, opened 2024-07-03
[ONNX] Fix LpNormalization Lowering

zjgarvey:
The LpNormalization lowering was previously just computing the norm, which is incorrect. This computes the norm then divides the input tensor by it's norm.

I've tested this against some simple onnx models locally. I'll look into adding a test case for this in an external test suite.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -2673,36 +2673,45 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
 
         return success();
       });
-  patterns.onOp(
-      "LpNormalization", 1,
-      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-        Torch::ValueTensorType resultType;
-        int64_t axis, p;
-        Value input;
-        if (binder.tensorOperand(input) ||
-            binder.s64IntegerAttr(axis, "axis", -1) ||
-            binder.s64IntegerAttr(p, "p", 2) ||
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -2673,36 +2673,45 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
-            llvm::ArrayRef<Value>{cstAxis});
-
-        rewriter.replaceOpWithNewOp<Torch::AtenNormScalarOptDimOp>(
-            binder.op, resultType, input, cstP, axisPrimList, cstKeepDim);
+  patterns.onOp("LpNormalization", 1,
+                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+                  Torch::ValueTensorType resultType;
+                  int64_t axis, p;
+                  Value input;
+                  if (binder.tensorOperand(input) ||
```
renxida, line 2681: This is more a question for my benefit than yours.

I sometimes separate this into individual checks & give a different error message on each.

But also, this code looks cleaner.

Wonder what your thoughts on the tradeoffs are?

renxida (approved):
LGTM

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
zjgarvey: I think the benefits to writing individual checks is minimal unless you have some reason to expect them to fail due to variation in versioning etc., but even then most of the binder methods don't actually fail since they fall back to a default value.

Issue #3457, opened 2024-06-13
aten.pad with mode = reflect is lowered wrong (numeric mismatch)


mgehre-amd:
Example IR
```
func.func @torch.aten.pad.reflect(%input: !torch.tensor<[2],f32>, %pads: !torch.vtensor<[2],si64>) -> !torch.tensor<[4],f32> {
  %int0 = torch.constant.int 0
  %float0.000000e00 = torch.constant.float 0.000000e+00
  %1 = torch.aten.select.int %pads, %int0, %int0 : !torch.vtensor<[2],si64>, !torch.int, !torch.int -> !torch.vtensor<[],si64>
  %2 = torch.aten.item %1 : !torch.vtensor<[],si64> -> !torch.int
  %pad = torch.prim.ListConstruct %2 : (!torch.int) -> !torch.list<int>
  %str = torch.constant.str "reflect"
  %ret = torch.aten.pad %input, %pad, %str, %float0.000000e00 : !torch.tensor<[2],f32>, !torch.list<int>, !torch.str, !torch.float -> !torch.tensor<[4],f32>
  return %ret : !torch.tensor<[4],f32>
}
```

This is because the decomposition doesn't check the mode and unconditionally lowers to ConstantPad:
https://github.com/llvm/torch-mlir/blob/919b599ebe57b1402b1aca21fa54799cc1e0cd91/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L5864

This can be triggered via ONNX.Pad with mode = reflect.

zjgarvey:
I believe this is resolved now with #3528 , so I'm going to close this issue. 

There are still replication/reflection pad issues that might arise (e.g. <https://github.com/llvm/torch-mlir/issues/3537>), but we can make new issues for those as they appear.

Issue #3442, opened 2024-06-10
Failed to convert "onnx.Pad" into torch


josel-amd:
Hi all,

I'm trying to import a very simple onnx model that only contains an `onnx.Pad` operator see [0] and then convert it into torch. For that purpose I perform the commands listed below

```
$ PYTHONPATH=tools/torch-mlir/python_packages/torch_mlir python -m torch_mlir.tools.import_onnx model.onnx -o onnx.mlir
$ torch-mlir-opt --convert-torch-onnx-to-torch onnx.mlir -debug
```

The onnx model can be found here [onnxmodel.zip](https://github.com/user-attachments/files/15774209/onnxmodel.zip) and the textual version is shown below. The generated IR is listed in [1].

```
<
   ir_version: 3,
   opset_import: ["" : 6],
   producer_name: "pytorch",
   producer_version: "0.3"
>
torch-jit-export (float[1,1,2,4] 0) => (float[1,1,3,9] 1) {
   1 = Pad <mode = "reflect", pads = [0, 0, 0, 2, 0, 0, 1, 3]> (0)
}
```

The importer works (i.e.: it does not fail or crash) and it produces the IR in [1]. When I try to convert the IR produced by the importer to torch, I get the following output:

```
//===-------------------------------------------===//
Legalizing operation : 'torch.operator'(0x559fbc3406b0) {
  %1 = "torch.operator"(%arg0) <{name = "onnx.Pad"}> {torch.onnx.mode = "reflect", torch.onnx.pads = [0 : si64, 0 : si64, 0 : si64, 2 : si64, 0 : si64, 0 : si64, 1 : si64, 3 : si64]} : (!torch.vtensor<[1,1,2,4],f32>) -> !torch.vtensor<[1,1,3,9],f32>

  * Fold {
  } -> FAILURE : unable to fold

  * Pattern : 'torch.operator -> ()' {
Trying to match ""
: conversion failed to apply: "onnx.Pad", sinceVersion=1
    ** Failure : no matching versioned converter
"" result 0
  } -> FAILURE : pattern failed to match
} -> FAILURE : no matched legalization pattern
//===-------------------------------------------===//
onnx.mlir:4:10: error: failed to legalize operation 'torch.operator' that was explicitly marked illegal
[...]

PR #3528, opened 2024-07-08
Adds misc fixes for some padding related issues

zjgarvey:
This patch adds a few misc pad op related changes:

1. Addresses issue <https://github.com/llvm/torch-mlir/issues/3457>
2. Addresses issue <https://github.com/llvm/torch-mlir/issues/3442>
3. Fixes the padding order for asymmetrically padded onnx.Conv ops
4. Enables passing quantization through those onnx.Conv op pre-paddings
5. Modifies the torch-to-linalg lowering of AtenReplicationPad2d op to enable support for input rank != 4

Unfortunately, even with all of these changes, the e2e tests for the ReplicationPad2d still fail the onnx config, since the torch export procedure for rearranging the pad order is complicated enough that the padding ints end up not being able to fold back to constants.


renxida (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -2286,34 +2328,20 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
           if (isa<IntegerType>(dataTensorType.getDtype()))
```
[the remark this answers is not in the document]
zjgarvey: Ah, that's a good point. It looks like we are currently just putting zero for the pad value when the tensor has an int dtype. I'll take a look into this, and I think I'll break up the quantization stuff out of this PR in the meantime.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -2333,6 +2365,21 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
                     Torch::ListType::get(rewriter.getType<Torch::IntType>()),
                     padsRearrange)
                 .getResult();
+
+        // lowering to AtenConstantPadNdOp directly allows passing any torch
+        // scalar type for the value, whereas AtenPadOp takes an optional float
+        // type.
+        if (mode == "constant" &&
```
rsuderman, line 2372: Could is insert a `mode == "constant"` check at the copy and store as a `bool`? Makes it easiest for checking models. We can do ones for `edge` and `wrap` as well.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -6366,17 +6366,92 @@ class DecomposeAtenPadOp : public OpRewritePattern<AtenPadOp> {
...
-    return success();
+    // try to reduce the number of padding dims if possible
+    if (matchPattern(op.getPad(), m_TorchListOfConstantInts(padInts))) {
+      if ((padInts.size() % 2) == 1)
+        return rewriter.notifyMatchFailure(op,
+                                           "expected an even number of pads");
+
+      for (unsigned i = padInts.size() - 1; i > 0; i -= 2) {
+        if (padInts[i] == 0 && padInts[i - 1] == 0)
+          usefulPadIndexEnd = i - 1;
```
rsuderman, line 6401: I usually do instead of this if/else
```
if (!cond) {
  break;
}
```

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -6366,17 +6366,92 @@ class DecomposeAtenPadOp : public OpRewritePattern<AtenPadOp> {
...
+    if (usefulPadIndexEnd < padValues.size()) {
+      ArrayRef<Value> usefulPadValues(padValues.begin(),
+                                      padValues.begin() + usefulPadIndexEnd);
+      usefulPads = rewriter.create<PrimListConstructOp>(
+          op.getLoc(),
+          rewriter.getType<Torch::ListType>(rewriter.getType<Torch::IntType>()),
+          usefulPadValues);
+    }
+
+    unsigned numPadDims = usefulPadIndexEnd / 2;
```
rsuderman, line 6428: Typically you want to use something like `uint32_t` so the bit depth is known.

rsuderman (approved):

zjgarvey:
I'm finding that there are some issues e2e with replication pad 2d when the input actually has non-trivial batches/channels. 

I have a plan for a fix but this might take a minute. It might be better to merge this, submit an issue, and then submit another patch with a fix later on. 

Looking into Rob's comments now.

zjgarvey:
Tracking the replication pad issue in <https://github.com/llvm/torch-mlir/issues/3537>. Merging this PR for now, and I'll post a fix for the issue soon.

PR #3529, opened 2024-07-09
[Stablehlo] enable stablehlo's python extension binding

qingyunqu:


penguin-wwy:
Do we need equivalent modifications for the bazel?

Vremold (approved):
LGTM.

qingyunqu:
> Do we need equivalent modifications for the bazel?

Seems that bazel doesn't build python binding. I don't find rules for python in bazel.

Issue #3312, opened 2024-05-09
Include Shape Information Into DecomposeComplexOps Patterns


zjgarvey:
Many pass pipelines rely on --torch-decompose-complex-ops, but do not include shape inference passes afterwards. For large models, the complex computations performed in shape inference passes take a significant amount of time to perform and are not as reliable as explicitly specifying the result types by hand. The only downside to adding these shapes by hand is the minor expense of code complexity, but with the development of appropriate helper functions such as [Torch::unsqueezeTensor](https://github.com/llvm/torch-mlir/blob/main/lib/Dialect/Torch/Utils/Utils.cpp#L315) as a replacement for ```rewriter.create<AtenUnsqueezeOp>(loc, baseType,...)``` as seen [here](https://github.com/llvm/torch-mlir/blob/a033bbfe6c2c3a7bc66102a03ae760136d0112aa/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L710), the potential downside to code readability should rendered completely irrelevant. 

As a design decision for the short term (before larger changes like fx decompositions are added), I'd like to come to a consensus with the community to avoid using 

```cpp
baseType = ValueType::getWithLeastStaticInformation(context);
```
in decompose complex ops conversions, and to request authors of recent conversions to go back and add shape information into their code. 

Here is the very short list of decompose complex ops passes which currently rely on shape inference:

- [x] "DecomposeAtenTriuOp" [here](https://github.com/llvm/torch-mlir/blob/a033bbfe6c2c3a7bc66102a03ae760136d0112aa/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L672) (see also [issue #573](https://github.com/nod-ai/SHARK-Turbine/issues/573)) @Mi-Jiazhi @qingyunqu 
- [x] "DecomposeAtenNanToNumOp" [here](https://github.com/llvm/torch-mlir/blob/a033bbfe6c2c3a7bc66102a03ae760136d0112aa/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L3320) @linuxlonelyeagle 
- [x] "DecomposeAtenGroupNormOp" [here](https://github.com/llvm/torch-mlir/blob/a033bbfe6c2c3a7bc66102a03ae760136d0112aa/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp#L5029) @JianzheXiao @qingyunqu @vivekkhandelwal1 
[...]

PR #3530, opened 2024-07-10
[Torch Dialect] fix aten.nan_to_num's decomposition when inf=None

qingyunqu:
also add shape infer in decomposition, see https://github.com/llvm/torch-mlir/issues/3312


zjgarvey (approved):
Thanks! This looks good to me.

PR #3531, opened 2024-07-10
Allow custom ops to return an array of tensors

matthewfl:
This PR adds support to `fx_importer.py` for handling custom ops that return an array of tensors.  As long as the length of the array is consistent across runs (determined statically), then this patch will work.  This does not require that the number of tensors returned is determined by the op's definition.

CC @sjain-stanford

### DIFF
--- a/test/python/fx_importer/custom_op_test.py
+++ b/test/python/fx_importer/custom_op_test.py
@@ -88,11 +88,16 @@ def test_tanh_sigmoid_cat_custom_op():
 
 @run
 # CHECK-LABEL: test_custom_op_array_output
-# CHECK:  func.func @main(%[[ARG0:[a-zA-Z0-9]+]]: !torch.vtensor<[2,3],f32>)
+# CHECK:  func.func @main(%[[ARG0:[a-zA-Z0-9]+]]: !torch.vtensor<[?,3],f32>)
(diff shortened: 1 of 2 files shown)


test/python/fx_importer/custom_op_test.py
```diff
@@ -84,3 +83,39 @@ def forward(self, x, y, z):
...
+            super().__init__()
+
+        def forward(self, a):
+            return torch.ops.my_custom_library.array_output_op(4, a)
+
+    a = torch.rand(2,3)
+    m = fx.export_and_import(
+        ArrayOutputCustomOp(),
+        a,
+        import_symbolic_shape_expressions=True,
```
sjain-stanford, line 119: Could you also add `dynamic_shapes` on this for testing shape expressions and custom ops with multiple outputs?

sjain-stanford (commented):
Looping in @stellaraccident for review as this touches fx_importer and might need further validation in the downstream suites Stella's team uses.

test/python/fx_importer/custom_op_test.py
matthewfl: done

stellaraccident (approved):
Trickier than I would have expected but lgtm.

sjain-stanford:
Thanks Stella. Going to land this as @matthewfl may not have write access.

PR #3542, opened 2024-07-14
[Symbolic Shapes] Test coverage for unbacked symint from data dependent ops

sjain-stanford:
We do have support for translating unbacked symbolic_ints that arise from data-dependent ops like `aten.nonzero`. This PR adds the python lit test coverage for the same.


stellaraccident (approved):
Lgtm when you get tests to pass

sjain-stanford:
> Lgtm when you get tests to pass

Thanks Stella. There is some stable-nightly mismatch due to bugs in 2.3 that got fixed in the 2.4 nightlies. I've temporarily `CHECK-DISABLED` some parts, with the TODO to re-enable once we move stable to 2.4.

PR #3546, opened 2024-07-17
Implement lowering of aten.atleast_2d

pkapris-syrmia:
This operator is needed to implement aten.vstack, which will be submitted in a subsequent PR


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1514,6 +1514,54 @@ class DecomposeAtenAtleast1dOp : public OpRewritePattern<AtenAtleast1dOp> {
...
+          rewriter.create<AtenAtleast1dOp>(loc, atleast1dResType, input);
+      Value zero = rewriter.create<Torch::ConstantIntOp>(
+          loc, rewriter.getI64IntegerAttr(0));
+      rewriter.replaceOpWithNewOp<AtenUnsqueezeOp>(op, opType, atleast1dRes,
+                                                   zero);
+      return success();
+    }
+
+    rewriter.replaceOp(op, input);
+    return success();
```
vivekkhandelwal1, line 1560: This code should be placed inside if by reversing the condition as per LLVM's https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code

vivekkhandelwal1 (commented):
Except one small comment, everything else looks fine.

vivekkhandelwal1 (approved):

PR #3550, opened 2024-07-18
Fix an error with OOB extract in interpolate lowering

zjgarvey:
The value '1.000001' seems to get truncated to `1.0`, which was causing `nan`'s to appear along the bottom/right border in some cases in the interpolate lowering. 

This just makes the value bigger so that it actually gets treated properly.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -2698,7 +2698,7 @@ static Value BilinearInterpolate(OpBuilder &b,
   auto inputRank = inputType.getRank();
 
   Value cstOneEps =
-      b.create<arith::ConstantOp>(loc, b.getF32FloatAttr(1.000001));
+      b.create<arith::ConstantOp>(loc, b.getF32FloatAttr(1.00001));
   Value cstOneFloat = b.create<arith::ConstantOp>(loc, b.getF32FloatAttr(1.0));
   Value cstHalf = b.create<arith::ConstantOp>(loc, b.getF32FloatAttr(0.5));
   Value zero = b.create<arith::ConstantOp>(loc, b.getF32FloatAttr(0.0));


mgehre-amd:
It's not super obvious why this is fixing the issue. Are we using the `eps` to force rounding into a certain direction? Is there a more direct way to do this?
And is it possible to add a test for this fix?

zjgarvey:
> It's not super obvious why this is fixing the issue. Are we using the `eps` to force rounding into a certain direction? Is there a more direct way to do this? And is it possible to add a test for this fix?

Sorry for the late reply. I was looking into this a bit more, and I unfortunately I can't seem to reproduce the original issue. I'm going to draft this PR, see if I can find a minimal reproducer for the issue, and come up with a better fix.

mgehre-amd:
I had seen an out of bounds access on 
```
<
   ir_version: 8,
   opset_import: ["" : 15]
>
repro (float[1,64,20,36] in) => (float[1,64,40,72] out) 
   <float[4] scales__117 =  {1,1,2,2}, float[0] roi__99 =  {}>
{
   out = Resize <coordinate_transformation_mode: string = "half_pixel", exclude_outside: int = 0, mode: string = "linear", nearest_mode: string = "floor"> (in, roi__99, scales__117)
}
```
when adding the `generate-runtime-verification` pass, see https://discord.com/channels/636084430946959380/742573221882364009/1264897029088350289

Is there a way to run an onnx model through the torch-mlir importer + linalg lowering + refbackend + execution engine from the command line to turn this into an easy reproducer?

zjgarvey:
> I had seen an out of bounds access on
> 
> ```
> <
>    ir_version: 8,
>    opset_import: ["" : 15]
> >
> repro (float[1,64,20,36] in) => (float[1,64,40,72] out) 
>    <float[4] scales__117 =  {1,1,2,2}, float[0] roi__99 =  {}>
> {
>    out = Resize <coordinate_transformation_mode: string = "half_pixel", exclude_outside: int = 0, mode: string = "linear", nearest_mode: string = "floor"> (in, roi__99, scales__117)
> }
> ```
> 
> when adding the `generate-runtime-verification` pass, see https://discord.com/channels/636084430946959380/742573221882364009/1264897029088350289
> 
> Is there a way to run an onnx model through the torch-mlir importer + linalg lowering + refbackend + execution engine from the command line to turn this into an easy reproducer?

We need to make something like this for torch-mlir, but it isn't currently set up. It's very high on my priority list, since the native onnx testing in torch-mlir is rather insufficient. For now, we have been using an external test-suite that runs with IREE as the backend/execution. If you want to take a look, I've been working on it here: <https://github.com/nod-ai/SHARK-TestSuite/tree/main/alt_e2eshark>.

PR #3553, opened 2024-07-19
onnxtotorch tfidfvectorizer 1-d

aldesilv:
1-d/2-d input and output
implemented based on the description and example test cases in
https://github.com/onnx/onnx/blob/main/docs/Operators.md#TfIdfVectorizer
and some notes from
https://github.com/onnx/onnx/blob/main/onnx/reference/ops/op_tfidf_vectorizer.py#L128

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -4272,7 +4272,10 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             dyn_cast<Torch::ValueTensorType>(input.getType()).getSizes();
         SmallVector<int64_t> inputShape(inputSizes);
         bool is_2d = (inputShape.size() > 1) ? true : false;
-        int batch_size = (inputShape.size() > 1) ? inputShape[0] : 1;
+        if (is_2d && inputShape[0] == ShapedType::kDynamic)
+          return rewriter.notifyMatchFailure(
+              binder.op, "input batch dimension cannot be dynamic");
+        int batch_size = (is_2d) ? inputShape[0] : 1;
 
         Value none = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
         Value zero = rewriter.create<Torch::ConstantIntOp>(
@@ -4317,8 +4320,9 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
               TypeRange({loopIndexType, output.getType()}),
               {binder.getLoc(), binder.getLoc()});
           Value batchValue = batchLoopBody->getArgument(0);
-          output = batchLoopBody->getArgument(1);
-          Value inputSequence;
+          Value output = batchLoopBody->getArgument(1);
+          Value outputForBatch = output;
+          Value inputSequence = input;
           if (is_2d) {
             // get input sequence from input (ex: [[0,1],[2,3]] -> [[0,1]] ->
             // [0,1])
@@ -4338,12 +4342,28 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                                             ArrayRef<int64_t>{inputShape[1]},
                                             inputType.getOptionalDtype()),
                 inputSequence, zero);
-          } else {
-            inputSequence = input;
+
+            SmallVector<int64_t> outputForBatchShape;
+            outputForBatchShape.push_back(1);
+            outputForBatchShape.push_back(resultShape[1]);
+            auto outputForBatchType = rewriter.getType<Torch::ValueTensorType>(
+                outputForBatchShape, resultType.getOptionalDtype());
+            outputForBatch = rewriter.create<Torch::AtenSliceTensorOp>(
+                binder.getLoc(), outputForBatchType, output,
+                /*dim=*/zero, batchValue, batchPlusOne, one);
+            outputForBatch = rewriter.create<Torch::AtenSqueezeDimOp>(
+                binder.getLoc(),
+                Torch::ValueTensorType::get(binder.op->getContext(),
+                                            ArrayRef<int64_t>{resultShape[1]},
+                                            inputType.getOptionalDtype()),
+                outputForBatch, zero);
           }
           // ngram_counts[j] records the starting position of ngrams within the
           // pool_int64's of length j+1. The loop below is iterating through the
           // different n-gram sizes
+          // ngram_i keeps track of which ngram we are looking at in the pool.
+          // The frequency of this ngram will be stored in the output tensor at
+          // the position ngram_indexes[ngram_i]
           int ngram_i = 0;
           for (int j = 0; j < (int)ngram_counts.size(); j++) {
             if (j + 1 < min_gram_length || j + 1 > max_gram_length)
@@ -4357,7 +4377,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             Value ngramLength = rewriter.create<Torch::ConstantIntOp>(
                 binder.getLoc(), rewriter.getI64IntegerAttr(ngram_length));
             for (int start = start_idx; start < end_idx;
-                 start += ngram_length) {
+                 start += ngram_length, ngram_i++) {
               Value count = rewriter.create<Torch::ConstantIntOp>(
                   binder.getLoc(), rewriter.getI64IntegerAttr(0));
               // for 1-grams, there is no skipping (skip = gap between
@@ -4417,45 +4437,34 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                   Value startInputIdx = countLoopBody->getArgument(0);
                   count = countLoopBody->getArgument(1);
 
-                  // extract a ngram from input
-                  Value dimValue = zero;
-                  Value offset = rewriter.create<Torch::AtenMulIntOp>(
-                      binder.getLoc(), skipCountPlusOne, ngramLength);
-                  Value end = rewriter.create<Torch::AtenAddIntOp>(
-                      binder.getLoc(), offset, startInputIdx);
-                  SmallVector<int64_t> shape;
-                  shape.push_back(ngram_length);
-                  auto ngramType = rewriter.getType<Torch::ValueTensorType>(
-                      shape, inputType.getOptionalDtype());
-                  Value inputNgram = rewriter.create<Torch::AtenSliceTensorOp>(
-                      binder.getLoc(), ngramType, inputSequence, dimValue,
-                      startInputIdx, end, skipCountPlusOne);
-
-                  Torch::BaseTensorType inputNgramType =
-                      cast<Torch::BaseTensorType>(inputNgram.getType());
+                  // extract input ngram and compare to pool ngram
+                  Torch::BaseTensorType inputSequenceType =
+                      cast<Torch::BaseTensorType>(inputSequence.getType());
                   SmallVector<int64_t> selectSizes;
                   selectSizes.push_back(1);
-                  Type selectResultType = inputNgramType.getWithSizesAndDtype(
-                      llvm::ArrayRef(selectSizes),
-                      inputNgramType.getOptionalDtype());
-                  auto sizes =
-                      dyn_cast<Torch::ValueTensorType>(inputNgram.getType())
-                          .getSizes();
+                  Type selectResultType =
+                      inputSequenceType.getWithSizesAndDtype(
+                          llvm::ArrayRef(selectSizes),
+                          inputSequenceType.getOptionalDtype());
                   Value foundNgram = rewriter.create<Torch::ConstantIntOp>(
                       binder.getLoc(), rewriter.getI64IntegerAttr(1));
-                  for (int i = 0; i < sizes[0]; i++) {
+                  for (int i = 0; i < ngram_length; i++) {
                     Value selectIndex = rewriter.create<Torch::ConstantIntOp>(
                         binder.getLoc(), rewriter.getType<Torch::IntType>(),
                         rewriter.getIntegerAttr(rewriter.getIntegerType(64),
                                                 i));
+                    selectIndex = rewriter.create<Torch::AtenMulIntOp>(
+                        binder.getLoc(), selectIndex, skipCountPlusOne);
+                    selectIndex = rewriter.create<Torch::AtenAddIntOp>(
+                        binder.getLoc(), selectIndex, startInputIdx);
                     Value inputExtract =
                         rewriter.create<Torch::AtenSelectIntOp>(
-                            binder.getLoc(), selectResultType, inputNgram, zero,
-                            selectIndex);
+                            binder.getLoc(), selectResultType, inputSequence,
+                            zero, selectIndex);
                     Value inputNgram_i = rewriter.create<Torch::AtenItemOp>(
                         binder.getLoc(), rewriter.getType<Torch::IntType>(),
                         inputExtract);
-                    // compare inputNgram to poolNgram
+
                     Value poolNgram_i = rewriter.create<Torch::ConstantIntOp>(
                         binder.getLoc(),
                         rewriter.getI64IntegerAttr(pool_int64s[start + i]));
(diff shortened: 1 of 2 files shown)


test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
```diff
@@ -2990,3 +2990,9 @@ func.func @test_scatternd_min(%arg0: !torch.vtensor<[4,4,4],f32>, %arg1: !torch.
   %0 = torch.operator "onnx.ScatterND"(%arg0, %arg1, %arg2) {torch.onnx.reduction = "min"} : (!torch.vtensor<[4,4,4],f32>, !torch.vtensor<[2,1],si64>, !torch.vtensor<[2,4,4],f32>) -> !torch.vtensor<[4,4,4],f32>
     return %0 : !torch.vtensor<[4,4,4],f32>
 }
+
+// CHECK-LABEL: @test_tfidfvectorizer_tf_onlybigrams_levelempty
```
renxida, line 2994: the test case looks like it doesn't have any filecheck comments, just the label.

vivekkhandelwal1:
Hi @aldesilv, can you please add some context about the implementation? I mean from where have you adopted this implementation, maybe add a link or description?

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
```diff
@@ -2990,3 +2990,9 @@ func.func @test_scatternd_min(%arg0: !torch.vtensor<[4,4,4],f32>, %arg1: !torch.
...
+func.func @test_tfidfvectorizer_tf_onlybigrams_levelempty(%arg0: !torch.vtensor<[12],si32>) -> !torch.vtensor<[3],f32> attributes {torch.onnx_meta.ir_version = 4 : si64, torch.onnx_meta.opset_version = 9 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
```
renxida, line 2995: have you run this on some test inputs & compared with onnxruntime? Wanna make sure this runs & generates the actual numerical results.

renxida: SHARK-TestSuite has some op tests you could reference.

renxida (dismissed):

zjgarvey:
We had to make a few changes, but Alex and I ran about 7 different "TF" mode tests in the test suite Friday and they all passed. We used the node tests from the onnx repo. 

I'll try to give a detailed review early next week. Although it gives numerically correct results in the cases we tested, it generates an incredible amount of IR (the IR coming from these small node tests are up to 1000 lines long). It would be good to go through and make sure nothing overly redundant is happening, and it would be good to see if we can convert the loops in IR generation into prim loops in the IR.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4213,4 +4213,301 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                    &skipLoop.getRegion(), skipLoop.getRegion().begin(),
+                    TypeRange({loopIndexType, count.getType()}),
+                    {binder.getLoc(), binder.getLoc()});
+                Value skipCount = skipLoopBody->getArgument(0);
+                Value skipCountPlusOne = rewriter.create<Torch::AtenAddIntOp>(
+                    binder.getLoc(), skipCount, one);
+                count = skipLoopBody->getArgument(1);
+
+                // max_start_index =
+                // inputSizes.back() - ((ngram_length - 1) * (skip_count + 1));
```
zjgarvey, line 4351: The final index of an n-gram (with stride = skip_count + 1) equals input_size - 1 when:
```
final_index = start_idx + (ngram_length - 1)*stride  = input_size - 1
```
or equivalently,
```
start_idx = (input_size - 1) - (ngram_length - 1)*(skip_count + 1)
```
Is my understanding correct? If so, the max_start_index in your code might be one lower than it should be.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4213,4 +4213,301 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                    binder.getLoc(), intType,
+                    rewriter.getI64IntegerAttr(inputSizes.back()));
+                Value ngramLengthMinusOne =
+                    rewriter.create<Torch::AtenSubIntOp>(binder.getLoc(),
+                                                         ngramLength, one);
+                Value ngramSkipLength = rewriter.create<Torch::AtenMulIntOp>(
+                    binder.getLoc(), ngramLengthMinusOne, skipCountPlusOne);
+                Value maxStartIndex = rewriter.create<Torch::AtenSubIntOp>(
+                    binder.getLoc(), seqLen, ngramSkipLength);
+                auto loop = rewriter.create<Torch::PrimLoopOp>(
```
zjgarvey, line 4362: Another comment or different naming might help here. E.g., `countLoop` instead of `loop`, or "This loop will extract each n-gram with the given skip_count from the input sequence, and increment the count if the n-gram matches the one gotten from the pool_int64s"

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4213,4 +4213,301 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                  PatternRewriter::InsertionGuard guard(rewriter);
+                  Block *loopBody = rewriter.createBlock(
+                      &loop.getRegion(), loop.getRegion().begin(),
+                      TypeRange({loopIndexType, count.getType()}),
+                      {binder.getLoc(), binder.getLoc()});
+
+                  Value startInputIdx = loopBody->getArgument(0);
+                  count = loopBody->getArgument(1);
+
+                  if (ngram_length >= min_gram_length &&
```
zjgarvey, line 4375: This check should probably happen earlier. We shouldn't even write the skipLoop, or the countLoop if that particular ngram_length isn't being used, right? 

So in the `j` loop, you could probably just include something like this at the beginning: 
```cpp
if (j+1 < min_gram_length || j+1 > max_gram_length)
   continue;
```

zjgarvey: Stopping for now here. I'll continue reviewing later today.

zjgarvey (commented):
Halfway through. Here are some comments for now!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
aldesilv: inputSizes.back() - ((ngram_length - 1) * (skip_count + 1)) is the index one higher than the last possible start index without the input ngram going out of bounds

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4213,4 +4213,317 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                    rewriter.getI64IntegerAttr(ngram_indexes[ngram_i]));
+                Value insertEnd = rewriter.create<Torch::AtenAddIntOp>(
+                    binder.getLoc(), insertStart, one);
+                outputSequence = rewriter.create<Torch::AtenSliceScatterOp>(
+                    binder.getLoc(), outputSequenceType, outputSequence,
+                    countUnsqueezed,
+                    /*dim=*/one, insertStart, insertEnd, /*step=*/one);
+                output = rewriter.create<Torch::AtenSliceScatterOp>(
+                    binder.getLoc(), resultType, output, outputSequence,
+                    /*dim=*/zero, batchValue, batchPlusOne, /*step=*/one);
```
zjgarvey, line 4507: When batches exist (2d), then for a given batch, you should only need to insert it once into the output tensor right at the end of the batch loop. There shouldn't be any need to repeatedly slice it from the batched output, update the slice, and rewrite the whole slice back into the batched output tensor for each ngram.

1. At the beginning of the batch loop, set up an initial (rank 1) tensor for that individual batch (call it `outputForBatch` or something). For 2-D, this can be a new rank-1 tensor of zeros (or a slice from the original output init), for 1-D this could be the whole output (`outputForBatch = output`). 
2. in this section where you are writing (slice scattering) to the `output` tensor, instead write to `outputForBatch` tensor. The logic will all match the "else" case here, since `outputForBatch` is rank-1. 
3. At the end of the batch loop, insert the `outputForBatch` slice back into the `output` for 2d, and just set `output=outputForBatch` in 1-D.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4213,4 +4213,317 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                  SmallVector<Value>{countFloat});
+              Value cstDtype = rewriter.create<Torch::ConstantIntOp>(
+                  binder.getLoc(), rewriter.getI64IntegerAttr(
+                                       (int)torch_upstream::ScalarType::Float));
+              SmallVector<int64_t> countShape{1};
+              auto countType = rewriter.getType<Torch::ValueTensorType>(
+                  countShape, resultType.getOptionalDtype());
+              Value countTensor = rewriter.create<Torch::AtenTensorOp>(
+                  binder.getLoc(), countType, dataList, /*dtype=*/cstDtype,
+                  /*layout=*/none, /*requires_grad=*/cstFalse);
```
zjgarvey, line 4471: This logic + using slice scatter to insert a single scalar element into a tensor, is performed in the decomposition pattern for `AtenScatterValueOp`. For readability, it might be better to use this op instead. 

Another option is to try and store all of the count values into (something bigger than a SmallVector), then make a primlist out of that value vector directly to generate the output for the given batch, instead of needing to insert each element individually. I'm not 100% sure about this approach, since I don't know what happens if the output size is too big.

zjgarvey (commented):
I added a few more comments. Let me know what you think!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4213,4 +4213,317 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                        rewriter.create<Torch::AtenSelectIntOp>(
+                            binder.getLoc(), selectResultType, inputNgram, zero,
+                            selectIndex);
+                    Value inputNgram_i = rewriter.create<Torch::AtenItemOp>(
+                        binder.getLoc(), rewriter.getType<Torch::IntType>(),
+                        inputExtract);
+                    // compare inputNgram to poolNgram
+                    Value poolNgram_i = rewriter.create<Torch::ConstantIntOp>(
+                        binder.getLoc(),
+                        rewriter.getI64IntegerAttr(pool_int64s[start + i]));
```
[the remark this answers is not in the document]
zjgarvey: Oh, maybe this could be too big to fit into a SmallVector? Not sure how big the pools can get.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4213,4 +4213,317 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                        binder.getLoc(), rewriter.getType<Torch::IntType>(),
+                        rewriter.getIntegerAttr(rewriter.getIntegerType(64),
+                                                i));
+                    Value inputExtract =
+                        rewriter.create<Torch::AtenSelectIntOp>(
+                            binder.getLoc(), selectResultType, inputNgram, zero,
+                            selectIndex);
+                    Value inputNgram_i = rewriter.create<Torch::AtenItemOp>(
+                        binder.getLoc(), rewriter.getType<Torch::IntType>(),
+                        inputExtract);
```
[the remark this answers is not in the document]
aldesilv: yeah

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
aldesilv: I'm not clear how to use the index arg in AtenScatterValueOp (it is a tensor). the second option is a bit like using outputforbatch.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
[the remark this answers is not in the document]
aldesilv: I left it as is for now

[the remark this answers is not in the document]
zjgarvey: Ok

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4238,6 +4238,318 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                SmallVector<int64_t> outputForBatchShape;
+                outputForBatchShape.push_back(1);
+                outputForBatchShape.push_back(resultShape[1]);
+                auto outputForBatchType =
+                    rewriter.getType<Torch::ValueTensorType>(
+                        outputForBatchShape, resultType.getOptionalDtype());
+                outputForBatch = rewriter.create<Torch::AtenSliceScatterOp>(
+                    binder.getLoc(), outputForBatchType, outputForBatch,
+                    countUnsqueezed,
+                    /*dim=*/one, insertStart, insertEnd, /*step=*/one);
```
zjgarvey, line 4526: I don't think this last change really simplified as much as it should have.

The point in my comment earlier is that `outputForBatch` doesn't need to be 1-d or 2-d dependent. It's just one batch regardless of how many batches there are, so it should just be a tensor of shape [number of n-grams]. Once all is said and done, you can insert `outputForBatch` into the actual output tensor at the very end of the batch loop, but I feel like that (and maybe the initialization of `outputForBatch`, since it might be better to get it from the loop arg in 1-d rather than making a new tensor) should be the only place where 1/2d needs to affect the control flow inside any of these loop bodies.

aldesilv: Ok made some changes

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4238,6 +4238,303 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                                             "TF mode supported only");
+        if (pool_int64s.size() == 0)
+          return rewriter.notifyMatchFailure(
+              binder.op, "pool_int64s empty, only integers supported");
+        auto inputType = dyn_cast<Torch::ValueTensorType>(input.getType());
+        auto inputSizes =
+            dyn_cast<Torch::ValueTensorType>(input.getType()).getSizes();
+        SmallVector<int64_t> inputShape(inputSizes);
+        bool is_2d = (inputShape.size() > 1) ? true : false;
+        int batch_size = (inputShape.size() > 1) ? inputShape[0] : 1;
```
zjgarvey, line 4275: What if input has dynamic dims?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4238,6 +4238,303 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            binder.getLoc(), TypeRange({output.getType()}), batchSize,
+            loopConditionTrue, ValueRange({output}));
+        {
+          PatternRewriter::InsertionGuard guard(rewriter);
+          Block *batchLoopBody = rewriter.createBlock(
+              &batchLoop.getRegion(), batchLoop.getRegion().begin(),
+              TypeRange({loopIndexType, output.getType()}),
+              {binder.getLoc(), binder.getLoc()});
+          Value batchValue = batchLoopBody->getArgument(0);
+          Value outputForBatch = batchLoopBody->getArgument(1);
```
zjgarvey, line 4320:
```suggestion
          output = batchLoopBody->getArgument(1);
          Value outputForBatch = output;
```

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4238,6 +4238,303 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          Value inputSequence;
```
zjgarvey, line 4321:
```suggestion
          Value inputSequence = input;
```

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4238,6 +4238,303 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+                /*dim=*/zero, batchValue, batchPlusOne, one);
+            outputForBatch = rewriter.create<Torch::AtenSqueezeDimOp>(
+                binder.getLoc(),
+                Torch::ValueTensorType::get(binder.op->getContext(),
+                                            ArrayRef<int64_t>{resultShape[1]},
+                                            inputType.getOptionalDtype()),
+                outputForBatch, zero);
+          } else {
+            inputSequence = input;
+          }
```
zjgarvey, line 4358: With the change above, you can delete the `else` block here.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
zjgarvey: Because you reference `output` a few times below, I think it is best to keep this as the original loop arg Value. `outputForBatch = output` in 1-d case, and gets modified as below in 2d-case.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
zjgarvey: Maybe add a match failure if batch_size is dynamic? (I think something like `kUnknownSize`?).

zjgarvey (commented):
Just a few more small changes and I'll give it a stamp.

zjgarvey:
Sorry it took a while to get around to this. When I tested e2e again this morning, it had regressed. I just decided to submit a PR against your branch again @aldesilv . The pr fixes the regression and adds a decent lit test.

vivekkhandelwal1:
@aldesilv @zjgarvey Can you please let me know about the status of this op?

zjgarvey (approved):

PR #3556, opened 2024-07-23
Register fake_quantize_cachemask ops and add their decompose patterns

zezhang:
Test:


sjain-stanford (approved):
LGTM

PR #3557, opened 2024-07-23
[Torch] enhance fold of aten.slice.Tensor

qingyunqu:
so that it could support folding slice with any static shape.


Vremold (approved):
LGTM. Thanks for the effort.

PR #3558, opened 2024-07-24
[Torch] enhance fold of aten.squeeze.dim

qingyunqu:


Vremold (approved):
LGTM. Thanks for the effort.

PR #3559, opened 2024-07-24
[Torch] enhance naryFolderHelper to support mixed dtypes

qingyunqu:
* so that it could support like `i64 + f64 => f64`.
* also unify `aten.log`'s folder code to use `naryFolderHelper`.


Vremold (approved):
LGTM. Thanks for the effort.

PR #3562, opened 2024-07-25
Fix macOS package build

dbabokin:
Without `--no-build-isolation` pip invokes `setup.py` in fresh environment, which doesn't have `torch` installed. But `setup.py` does `import torch` to check PyTorch version, so the build crashes. At the same time the script creates a disposable virtual environment with all required dependencies specifically to run wheel build. Note that Linux package build also runs with this option.

https://github.com/llvm/torch-mlir/blob/15cf7106c423019f30fef3cffefc4b4cf064934a/setup.py#L230

This was introduced by this commit: https://github.com/llvm/torch-mlir/commit/74f7a0c9d6ea5b3a6d37dd61d0a83557a90b1d03

And looks like macOS builds were not running in CI ever since.

I also updated Python versions in `install_macos_deps.sh`.

### DIFF
--- a/build_tools/python_deploy/build_macos_packages.sh
+++ b/build_tools/python_deploy/build_macos_packages.sh
@@ -88,7 +88,7 @@ function build_torch_mlir() {
   TORCH_MLIR_PYTHON_PACKAGE_VERSION=${TORCH_MLIR_PYTHON_PACKAGE_VERSION} \
   MACOSX_DEPLOYMENT_TARGET=$MACOSX_DEPLOYMENT_TARGET \
   CMAKE_OSX_ARCHITECTURES=$CMAKE_OSX_ARCHITECTURES \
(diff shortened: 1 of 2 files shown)


vivekkhandelwal1:
Since I'm not familiar with this I can't review this PR. @stellaraccident Can you please take a look?

stellaraccident (approved):
I haven't used macos in a while but I think this is a pip version issue. At some point pip defaulted to non build isolation mode, and most instructions recommend upgrading pip. Probably a `python -m pip install --upgrade pip` before makes it better and work the same on all pytgon versions. 

But if you confirm this patch fixes your issue, go ahead and land.

dbabokin:
My goal is to revive macOS builds at https://github.com/llvm/torch-mlir-release. The changes in this PR are necessary, but not sufficient for that. I tried building with `torch-llvm` trunk and something is broken again in the build.

I suggest proceeding with merging this PR and I'll submit another one for the remaining problems.

PR #3563, opened 2024-07-25
Implement lowering of torch.aten.hstack

BaneTrifa:


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3648,6 +3648,55 @@ class DecomposeAtenStackOp : public OpRewritePattern<AtenStackOp> {
...
+    auto one =
+        rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
+
+    // Execute AtenAtleast1dOp on every tensor inside tensors.
+    SmallVector<Value> atleast1dTensors;
+    for (auto tensor : tensors) {
+      auto atleast1dTensor =
+          rewriter.create<AtenAtleast1dOp>(loc, tensor.getType(), tensor);
+      atleast1dTensors.push_back(atleast1dTensor);
+    }
```
vivekkhandelwal1, line 3679: Before doing this, shouldn't we check whether the tensors are already of rank >= 1 or not? Otherwise, this would result in a lot of unusable code.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3648,6 +3648,55 @@ class DecomposeAtenStackOp : public OpRewritePattern<AtenStackOp> {
...
+    SmallVector<Value> tensors;
+    if (!getListConstructElements(op.getTensors(), tensors))
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: the tensor list is not from list construct");
+
+    // Create constants.
+    auto zero =
+        rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(0));
+    auto one =
+        rewriter.create<ConstantIntOp>(loc, rewriter.getI64IntegerAttr(1));
```
vivekkhandelwal1, line 3671: Since, either of these constants would be used, I would suggest to combine them with the code where the Cat op is created inside if-else.

vivekkhandelwal1 (changes requested):

vivekkhandelwal1 (approved):

BaneTrifa:
Can you please merge this PR? @vivekkhandelwal1

PR #3566, opened 2024-07-27
[Torch] Add support for static uneven divisible AdaptiveAvgPool2d

yyp0:
The static uneven divisible AdaptiveAvgPool2d means that although the input size is not an integer multiple of ouput size, but the kernel and stride size can also be fixed (not dynamic). The derivation logic of kernel and stride size is consistent with torch/_decomp/decomposations.py:adaptive_avg_pool2d as described in the following: 

1. Stride Size
Firstly , derive the start index in each reduce operation according to the output size (`n`), `start_index = ([0, 1, ..., n - 1] * input_size) // output_size`. For each index `k`, if `k * (input_size % output_size) < output_size`, then the current and previous stride keeps the same as `input_size // output_size`. So suppose `(n-1) * (input_size % output_size) < output_size`, the stride in the whole AdaptiveAvgPool2d process keeps static, as `input_size // output_size`.

2. Kernel Size
torch/_decomp/decomposations.py:adaptive_avg_pool2d calculates a static kernel size when the input/output sizes satisfy either of the two conditions, `input_size % output_size == 0` or `output_size % (input_size % output_size) == 0`. Here if `input_size % output_size == 0`, then the kernel size equals `input_size // output_size`, otherwise `input_size // output_size + 1.`


lib/Dialect/Torch/Transforms/LowerToBackendContract.cpp
```diff
@@ -500,6 +500,7 @@ static void markDecomposedOpsAsIllegal(MLIRContext *context,
   target.addIllegalOp<AtenToPrimDeviceOp>();
   target.addIllegalOp<AtenAdaptiveAvgPool1dOp>();
   target.addIllegalOp<AtenAdaptiveAvgPool2dOp>();
+  target.addIllegalOp<Aten_AdaptiveAvgPool2dOp>();
```
qingyunqu, line 503: remove this useless code?

yyp0: Done

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -7038,32 +7038,84 @@ class DecomposeAtenAdaptiveAvgPool2dOp
...
-      Value cond = rewriter.create<AtenEqIntOp>(loc, remainder, constantZero);
-      rewriter.create<RuntimeAssertOp>(loc, cond,
-                                       "unimplemented: only support cases "
-                                       "input size is an integer multiple of "
-                                       "output size");
-      Value stride = rewriter.create<AtenFloordivIntOp>(
+
+      // Filter cases with fixed stride size.
+      Value cond1 = rewriter.create<Torch::AtenGtIntOp>(
+          loc, rewriter.getType<Torch::BoolType>(), outputShapeSizesTorchInt[i],
```
qingyunqu, line 7063: it's no need to write `rewriter.getType<Torch::BoolType>()`, same as other code below.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -7038,32 +7038,84 @@ class DecomposeAtenAdaptiveAvgPool2dOp
...
+      Value cond = rewriter.create<Torch::AtenBoolIntOp>(
+          loc, rewriter.create<Torch::AtenAddIntOp>(loc, cond2, cond3));
+
+      Value condAnd = rewriter.create<Torch::AtenAddIntOp>(
+          loc, rewriter.create<Torch::AtenIntBoolOp>(loc, cond1),
+          rewriter.create<Torch::AtenIntBoolOp>(loc, cond));
+      Value finalCond =
+          rewriter.create<Torch::AtenEqIntOp>(loc, condAnd, constantTwo);
+
+      rewriter.create<RuntimeAssertOp>(
```
qingyunqu, line 7093: It's a little coupled. I think it's more clear to split `RuntimeAssert`, one for kernel and another for stride.
BTW, the code above is hard for human reading. Could you make it closer to the formula described?

qingyunqu (approved):
LGTM. Thanks for the efforts!

PR #3579, opened 2024-07-31
Add extf-trunc f32-f64-f32 ellision

rsuderman:
Torch has all scalars represented as i64 and f64 types which results in extraneous trunc-extf commands. We can rework this by elliding widen-narrow cases away.


zjgarvey (approved):
I don't personally see a problem with this, but I'm not sure where it most appropriately belongs. 

For example, would it make more sense to add this as a Fold pattern for arith::truncf? For example, in <https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Arith/IR/ArithOps.cpp#L1491>

Weirdly enough, there is already something similar (but situationally and the other way around) with <https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Arith/IR/ArithOps.cpp#L1394>

rsuderman:
> I don't personally see a problem with this, but I'm not sure where it most appropriately belongs.
> 
> For example, would it make more sense to add this as a Fold pattern for arith::truncf? For example, in [llvm/llvm-project@`main`/mlir/lib/Dialect/Arith/IR/ArithOps.cpp#L1491](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Arith/IR/ArithOps.cpp?rgh-link-date=2024-07-31T22%3A32%3A13Z#L1491)
> 
> Weirdly enough, there is already something similar (but situationally and the other way around) with [llvm/llvm-project@`main`/mlir/lib/Dialect/Arith/IR/ArithOps.cpp#L1394](https://github.com/llvm/llvm-project/blob/main/mlir/lib/Dialect/Arith/IR/ArithOps.cpp?rgh-link-date=2024-07-31T22%3A32%3A13Z#L1394)

I honestly agree that it should be an `arith.truncf` folder but historically there has been a lot of pushback over these types of folders. Hoping to avoid the controversy of putting this in a core dialect :/

zjgarvey:
> I honestly agree that it should be an `arith.truncf` folder but historically there has been a lot of pushback over these types of folders. Hoping to avoid the controversy of putting this in a core dialect :/

Yeah, I'm okay with adding this here. Maybe it would be good to include a comment like "#todo upstream this to an arith.truncf folder"?

rsuderman:
> > I honestly agree that it should be an `arith.truncf` folder but historically there has been a lot of pushback over these types of folders. Hoping to avoid the controversy of putting this in a core dialect :/
> 
> Yeah, I'm okay with adding this here. Maybe it would be good to include a comment like "#todo upstream this to an arith.truncf folder"?

Done.

PR #3580, opened 2024-07-31
[onnx] Fix `onnx.Shape` to include `start` and `end` processing

rsuderman:
`onnx.Shape` can select only a subset of indices using attributes. Add support for these attributes.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -1611,12 +1611,6 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         Value shape = rewriter.create<Torch::Aten_ShapeAsTensorOp>(
             binder.getLoc(), shapeType, operand);
 
-        int64_t start = 0;
-        int64_t end = -1;
-        if (binder.optionalS64IntegerAttr(start, "start") ||
-            binder.optionalS64IntegerAttr(end, "end"))
-          return failure();
-
         if (start == 0 && end == -1) {
           rewriter.replaceOp(binder.op, shape);
           return success();
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -1597,8 +1597,51 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                   if (binder.tensorOperand(operand) ||
                       binder.tensorResultType(resultType))
                     return failure();
-                  rewriter.replaceOpWithNewOp<Torch::Aten_ShapeAsTensorOp>(
-                      binder.op, resultType, operand);
+
+                  auto inputType =
+                      dyn_cast<Torch::ValueTensorType>(operand.getType());
+                  int64_t inputRank = inputType.getSizes().size();
+
+                  auto shapeType = Torch::ValueTensorType::get(
+                      binder.op->getContext(), SmallVector<int64_t>{inputRank},
+                      resultType.getOptionalDtype());
+
+                  Value shape = rewriter.create<Torch::Aten_ShapeAsTensorOp>(
+                      binder.getLoc(), shapeType, operand);
+
+                  int64_t start = 0;
+                  int64_t end = -1;
+                  if (binder.s64IntegerAttr(start, "start", {}) &&
+                      binder.s64IntegerAttr(end, "end", {})) {
+                    rewriter.replaceOp(binder.op, shape);
+                    return success();
+                  }
+
+                  if (start < 0)
+                    start += inputRank;
+                  if (end < 0)
+                    end += inputRank;
+
+                  Value sv = rewriter.create<Torch::ConstantIntOp>(
+                      binder.getLoc(), rewriter.getType<Torch::IntType>(),
+                      rewriter.getI64IntegerAttr(start));
+
+                  Value ev = rewriter.create<Torch::ConstantIntOp>(
+                      binder.getLoc(), rewriter.getType<Torch::IntType>(),
+                      rewriter.getI64IntegerAttr(end));
+
+                  Value step = rewriter.create<Torch::ConstantIntOp>(
+                      binder.getLoc(), rewriter.getType<Torch::IntType>(),
+                      rewriter.getI64IntegerAttr(1));
+
+                  Value dim = rewriter.create<Torch::ConstantIntOp>(
+                      binder.getLoc(), rewriter.getType<Torch::IntType>(),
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1590,17 +1590,64 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getI64IntegerAttr(end));
+
+        Value step = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getI64IntegerAttr(1));
+
+        Value dim = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getI64IntegerAttr(0));
```
zjgarvey, line 1643: It might be cleaner to use a simpler build method for `Torch::ConstantIntOp`:

```cpp
  static void build(::mlir::OpBuilder &odsBuilder, ::mlir::OperationState &odsState, ::mlir::IntegerAttr value);
```
or
```cpp
  static void build(::mlir::OpBuilder &odsBuilder, ::mlir::OperationState &odsState, uint64_t value);
```
```suggestion
        Value sv = rewriter.create<Torch::ConstantIntOp>(
            binder.getLoc(), rewriter.getI64IntegerAttr(start));

        Value ev = rewriter.create<Torch::ConstantIntOp>(
            binder.getLoc(), rewriter.getI64IntegerAttr(end));

        Value step = rewriter.create<Torch::ConstantIntOp>(
            binder.getLoc(), 1);

        Value dim = rewriter.create<Torch::ConstantIntOp>(
            binder.getLoc(), 0);
```

zjgarvey (commented):
Some small nits

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1590,17 +1590,64 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+
+        if (start == 0 && end == -1) {
+          rewriter.replaceOp(binder.op, shape);
+          return success();
+        }
+
+        if (start < 0)
+          start += inputRank;
+        if (end < 0)
+          end += inputRank;
```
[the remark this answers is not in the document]
rsuderman: Fixed.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1590,17 +1590,64 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            binder.op->getContext(), SmallVector<int64_t>{inputRank},
+            resultType.getOptionalDtype());
+
+        Value shape = rewriter.create<Torch::Aten_ShapeAsTensorOp>(
+            binder.getLoc(), shapeType, operand);
+
+        int64_t start = 0;
+        int64_t end = -1;
+        if (binder.s64IntegerAttr(start, "start", {}) &&
+            binder.s64IntegerAttr(end, "end", {})) {
```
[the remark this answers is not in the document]
rsuderman: Fixed.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1590,17 +1590,53 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            binder.op->getContext(), SmallVector<int64_t>{inputRank},
+            resultType.getOptionalDtype());
+
+        Value shape = rewriter.create<Torch::Aten_ShapeAsTensorOp>(
+            binder.getLoc(), shapeType, operand);
+
+        int64_t start = 0;
+        int64_t end = -1;
+        if (binder.optionalS64IntegerAttr(start, "start", 0) &&
+            binder.optionalS64IntegerAttr(end, "end", -1)) {
```
zjgarvey, line 1614: I suppose it's one or the other. the `optionalS64IntegerAttr` doesn't take a default and returns a failure if the op is missing the optional attr. `s64IntegerAttr` does take a default, but might fail for other reasons. 

Now that I think about it more, it might make sense to add `s64IntegerAttr` with the default values at the beginning (seperated by `||`), and remove this case because it would be covered below by the 
```cpp
if (start == 0 && end == -1) {
```

zjgarvey: My concern is that if these are separated by `&&`, then if it succeeds in binding the start, wouldn't it ignore the second check/get?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1590,17 +1590,51 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
-                      binder.op, resultType, operand);
-                  return success();
-                });
+  patterns.onOp(
+      "Shape", 9, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value operand;
+        if (binder.tensorOperand(operand) ||
+            binder.tensorResultType(resultType))
+          return failure();
```
zjgarvey, line 1599:
```suggestion
        int64_t start, end;
        if (binder.tensorOperand(operand) ||
            binder.tensorResultType(resultType) ||
            binder.s64IntegerAttr(start, "start", 0) ||
            binder.s64IntegerAttr(end, "end", -1))
          return failure();
```

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1590,17 +1590,51 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            resultType.getOptionalDtype());
+
+        Value shape = rewriter.create<Torch::Aten_ShapeAsTensorOp>(
+            binder.getLoc(), shapeType, operand);
+
+        int64_t start = 0;
+        int64_t end = -1;
+        if (binder.optionalS64IntegerAttr(start, "start") ||
+            binder.optionalS64IntegerAttr(end, "end"))
+          return failure();
```
zjgarvey, line 1615: I think the optionalS64IntegerAttr method fails when the attr isn't present. I'd remove this section and use the suggested change above.

rsuderman: Commit integrated

zjgarvey (approved):

PR #3585, opened 2024-08-01
[onnx][torch] Fix `onnx.SoftmaxCrossEntropyLoss` for ignore index

rsuderman:
There were two issues related to `ignore_index` being set

(1) the onnx-to-linalg pass as not reading the value correctly (2) the mean pass was not considering the `ignore_index` value

For (2) when taking the mean we need to know how many of the values were considered in the sum and therefore we cannot divide by the total number of elements. Adding a summation across the total number should correct this issue.


zjgarvey (approved):
This definitely makes sense, but I'm a bit confused why the change to Uncategorized.cpp doesn't have an effect on any e2e tests (e.g., "NllLossModule_mean_basic"). Is the numerical difference so small that the mean reduction examples still pass without the fix?

rsuderman:
> This definitely makes sense, but I'm a bit confused why the change to Uncategorized.cpp doesn't have an effect on any e2e tests (e.g., "NllLossModule_mean_basic"). Is the numerical difference so small that the mean reduction examples still pass without the fix?

It depends on the test but presumably it depends on the number of classes. The more the number of classes the smaller the error. I would have to look at the existing test to know exactly why it was going wrong.

PR #3588, opened 2024-08-02
Avoid warnings-as-errors build failure

rsuderman:
Lambda needs a return value to avoid a build failure.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -1335,9 +1335,10 @@ llvm::SmallVector<APInt> getFoldValueAtIndexInt(llvm::ArrayRef<Attribute> attrs,
 using NAryFoldFpOperator = std::function<double(ArrayRef<double>)>;
 using NAryFoldIntOperator = std::function<APInt(ArrayRef<APInt>)>;
 
-static OpFoldResult naryFolderHelper(ArrayRef<Attribute> operands, Type ty,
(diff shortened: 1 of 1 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1868,6 +1868,7 @@ OpFoldResult AtenLogOp::fold(FoldAdaptor adaptor) {
   };
   auto intFold = [](llvm::ArrayRef<APInt> inputs) -> APInt {
     assert(false && "should not reach here");
+    return APInt(inputs[0].getBitWidth(), log(inputs[0].getLimitedValue()));
   };
 
   return naryFolderHelper(adaptor.getOperands(), resultType, fpFold, intFold);
```
ScottTodd, line 1874: Can `naryFolderHelper` take `std::optional<NAryFoldIntOperator>` instead, and avoid calling this entirely? This is sketchy if asserts are disabled.

These call sites would be safer:
```c++
return naryFolderHelper(adaptor.getOperands(), resultType, fpFold);
return naryFolderHelper(adaptor.getOperands(), resultType, fpFold, nullptr);
```

Could assert (or fail with an error message, if this case isn't impossible by construction) in `naryFolderHelper` in here if `intFold` is not provided: https://github.com/llvm/torch-mlir/blob/d0933b0eb6c94c38132d5d80fd82323cda80a159/lib/Dialect/Torch/IR/TorchOps.cpp#L1400-L1410

ScottTodd (approved):

PR #3591, opened 2024-08-04
[e2e_testing] check process exitcode early in e2e

qingyunqu:
It will exit immediately. So it doesn't need to wait 6 min.


stellaraccident (approved):
Thanks

PR #3595, opened 2024-08-06
[LINALG] Added support for conversion from float to complex.

BaneTrifa:


lib/Conversion/Utils/Utils.cpp
```diff
@@ -364,6 +366,25 @@ Value convertScalarToDtype(OpBuilder &b, Location loc, Value scalar, Type dtype,
 
       return b.create<complex::CreateOp>(loc, dtypeComplex, realVal, imgVal);
     }
+
+    // Float to complex type.
+    if (auto dtypeFloat = dyn_cast<mlir::FloatType>(scalarType)) {
+
```
vivekkhandelwal1, line 372: You can remove this extra line.

lib/Conversion/Utils/Utils.cpp
```diff
@@ -364,6 +366,25 @@ Value convertScalarToDtype(OpBuilder &b, Location loc, Value scalar, Type dtype,
...
+          cast<mlir::FloatType>(dtypeComplex.getElementType());
+      Value realVal;
+      Value imgVal =
+          b.create<arith::ConstantOp>(loc, b.getZeroAttr(complexElementType));
+
+      if (complexElementType.getWidth() > dtypeFloat.getWidth()) {
+        realVal = b.create<arith::ExtFOp>(loc, complexElementType, scalar);
+      } else {
+        realVal = scalar;
+      }
```
vivekkhandelwal1, line 383: Truncation not required in else case?

BaneTrifa: You can truncate floating-point values, but I don't think it's necessary because the `convertScalarToDtype` function will be called when an operator has operands of different types (e.g., one f64 and the other complex<f32>). Even though float64 has a wider bitwidth than complex<f32>, both types will be promoted and converted to complex<f64>, which has the same bitwidth as f64. Truncation is only required if you explicitly want to convert f64 to complex<f32> without promoting types, so I added this to my code in case someone want to use this .

vivekkhandelwal1 (approved):
LGTM

PR #3597, opened 2024-08-06
[ONNX] Add training mode support for BatchNormalization op

vivekkhandelwal1:
This commit extends the OnnxToTorch lowering for BatchNormalization op for supporting the case when training=True.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -339,43 +339,123 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
                       binder.op, resultType, operand);
                   return success();
                 });
-  patterns.onOp("BatchNormalization", 15,
-                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-                  Torch::ValueTensorType resultType;
-                  Value input, weight, bias, runningMean, runningVar;
-                  bool training;
-                  float momentum, eps;
-                  if (binder.s64BoolAttr(training, "training_mode", 0))
-                    return failure();
-                  if (training) {
-                    // TODO: Add support for training = true
-                    return rewriter.notifyMatchFailure(
-                        binder.op, "unsupported conversion: training = true");
-                  }
+  patterns.onOp(
+      "BatchNormalization", 15,
+      [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        Value input, weight, bias, inputMean, inputVar;
+        bool training;
+        float momentum, eps;
+        if (binder.tensorOperandAtIndex(input, 0) ||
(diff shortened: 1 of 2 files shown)


vivekkhandelwal1:
Fixes https://github.com/nod-ai/SHARK-Turbine/issues/285

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -339,43 +339,123 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+
+        // When training_mode=False, the op outputs only Y, where
+        // Y = (X - input_mean) / sqrt(input_var + epsilon) * scale +
+        // B
+        if (!training) {
+          rewriter.replaceOpWithNewOp<Torch::AtenBatchNormOp>(
+              binder.op, resultType, input, weight, bias, inputMean, inputVar,
+              /*training=*/cstFalse, cstMomentum, cstEps,
+              /*cudnn_enabled=*/cstFalse);
+          return success();
```
zjgarvey, line 375: If this op has a training variable, why not implement the logic below for the pytorch op and just lower with `/*training=*/training`?

vivekkhandelwal1: Because the PyTorch op returns only one result, while the OnnxOp expects 3 results.

vivekkhandelwal1: And, the computation for `currentMean/Var` is anyway required for computing other results, hence no extra work is done here.

zjgarvey (changes requested):
This does not pass e2e testing for the following onnx node tests due to a significant numerics mismatch:

"test_batchnorm_epsilon_training_mode"
"test_batchnorm_example_training_mode"

Please try to debug the numerics mismatching when you get the chance. You can either use this branch of torch-mlir in your iree-build and run the iree_tests, or you can use the alt_e2eshark and run `python run.py --torchtolinalg -t test_batch`.

zjgarvey:
From a glance, it looks like the numerics for `Y` and `running_var` are off, but `running_mean` seems to match pretty well.

vivekkhandelwal1:
> This does not pass e2e testing for the following onnx node tests due to a significant numerics mismatch:
> 
> "test_batchnorm_epsilon_training_mode" "test_batchnorm_example_training_mode"
> 
> Please try to debug the numerics mismatching when you get the chance. You can either use this branch of torch-mlir in your iree-build and run the iree_tests, or you can use the alt_e2eshark and run `python run.py --torchtolinalg -t test_batch`.

Hi @zjgarvey, the accuracy issue is fixed. After a lot of debugging, it turned out that the `unbiased` has to be set to False, instead of True.

vivekkhandelwal1:
@zjgarvey Can you please review this PR, today?

zjgarvey (approved):
As long as this is passing numerics, the implementation is clear and well-commented on. Thanks, Vivek.

PR #3598, opened 2024-08-06
[onnx] Fix edge condition for `onnx.ReduceMax`

rsuderman:
For length-0 on `onnx.ReduceMax` the length 0 case was incorrect due to a copy paste error.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -1345,14 +1345,14 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           }
 
           if (IntegerType intTy = dyn_cast<IntegerType>(dty)) {
-            auto mx =
+            auto minInt =
                 intTy.isSigned()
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -1346,8 +1347,8 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           if (IntegerType intTy = dyn_cast<IntegerType>(dty)) {
             auto mx =
```
zjgarvey, line 1348: can you rename this variable to something like `minInt` instead of `mx`?

zjgarvey (approved):
Makes sense to use negative infinity. Kind of a weird case though, when do we ever get a 0 dim size? Nit/optional suggestion included, but looks good to me.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
rsuderman: Donezo

zjgarvey (approved):

PR #3599, opened 2024-08-06
[onnx] Handle negative indices for `onnx.GatherElements`

rsuderman:
Add a check for negative indices and offset appropriately for `onnx.GatherElements`.


zjgarvey (approved):
Yeah, makes sense to me. Please rename the PR to reflect that this edits `onnx.GatherElements`. 

I just tested with "test_gather_elements_negative_indices" (which went from a numerics mismatch to a pass, so thanks).

PR #3600, opened 2024-08-06
Add RestructureNonConstantAxes pass to address reduce op tests failing on non constant axes

renxida:


renxida:
This fixes a whole bunch of `SHARK-TestSuite/iree_tests/onnx/node` test cases: https://gist.github.com/renxida/5d83b9cb9b810fbb07cb93484235911d

renxida:
Had an infinite loop problem. Just merged #3650 to fix that.

lib/Dialect/Torch/Transforms/RestructureNonConstantAxes.cpp
```diff
@@ -0,0 +1,276 @@
+//===- LowerToBackendContract.cpp --------------------------------*- C++-*-===//
```
zjgarvey, line 1: copypasta relic?

lib/Dialect/Torch/Transforms/RestructureNonConstantAxes.cpp
```diff
@@ -0,0 +1,276 @@
...
+    bool oldDimIsList = isa<Torch::ListType>(dim.getType());
+    if (oldDimIsList) {
+      Value len = rewriter.create<Torch::AtenLenTOp>(loc, intType, dim);
+      Value dimListIsLengthOne =
+          rewriter.create<Torch::AtenEqIntOp>(loc, boolType, len, one);
+      rewriter.create<Torch::RuntimeAssertOp>(
+          loc, dimListIsLengthOne,
+          rewriter.getStringAttr("RestructureNonConstantAxes does not support "
+                                 "dim lists with more than one element"));
+      dim = rewriter.create<Torch::Aten__Getitem__TOp>(loc, intType, dim, zero);
```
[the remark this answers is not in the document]
renxida: yup they're sometimes dynamic & passed in as an arg in the test cases

lib/Dialect/Torch/Transforms/RestructureNonConstantAxes.cpp
renxida: a ty for catching that

vivekkhandelwal1:
Hi @renxida, I don't see this PR being approved by any of the reviewers added by you. I have mentioned it before as well, we should avoid merging any PR until it's reviewed and "APPROVED" by a reviewer. If the PR contains trivial changes and they are very urgent, only in that case we might go for merging the PR without any approval (although that is also not good). 

This PR definitely does not contain trivial changes, hence before merging this you should've waited for the reviewers to approve. If the review is taking time then ping reviewers to do it, instead of directly merging. I hope these things would be remembered from the next time in order to maintain the sanctity of this project.

PR #3606, opened 2024-08-07
[onnx] Fix transposition code for `onnx.OneHot`

rsuderman:
The post onehot transposition code was unexercised. Fixed the test and transformation to check use.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -8301,6 +8301,7 @@ class DecomposeAtenOneHotOp : public OpRewritePattern<AtenOneHotOp> {
           op, "input tensor should have known sizes.");
     int64_t inputRank = inputType.getSizes().size();
     int64_t numClasses = Torch::kUnknownSize;
+    auto resultType = cast<ValueTensorType>(op.getType());
     matchPattern(op.getNumClasses(), m_TorchConstantInt(&numClasses));
     Value none = rewriter.create<ConstantNoneOp>(loc);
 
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -8316,11 +8317,13 @@ class DecomposeAtenOneHotOp : public OpRewritePattern<AtenOneHotOp> {
     llvm::SmallVector<int64_t> unsqueezeShape(inputType.getSizes());
     unsqueezeShape.push_back(1);
     auto unsqueezeType =
-        ValueTensorType::get(context, unsqueezeShape, si64Type);
+        ValueTensorType::get(context, unsqueezeShape, inputType.getDtype());
     Value unsqueezeTensor = rewriter.create<AtenUnsqueezeOp>(
         loc, unsqueezeType, input,
         rewriter.create<ConstantIntOp>(loc,
                                        rewriter.getI64IntegerAttr(inputRank)));
```
zjgarvey, line 8324: Optional: what would you think about replacing this highlighted section with a util function? That's like 7 lines of boilerplate code just vanishing into thin air!
```cpp
Value unsqueezeTensor = Torch::unsqueezeTensor(rewriter, op, input, -1);
```

zjgarvey (approved):
Those were some bugs you caught there, Rob.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
rsuderman: Coolio, didn't know that existed

PR #3608, opened 2024-08-07
[onnx] `onnx.Split` may not have `num_outputs` which can be inferred

rsuderman:
The attribute does not exist in all variants of the operation. It can be inferred from the number of results so we should just do that.


zjgarvey:
There seem to be two conversions for split. Should this change be added to both?

rsuderman:
> There seem to be two conversions for split. Should this change be added to both?

It only appears to apply to one of them. I initially was drafting for the first but it wasn't relevant.

zjgarvey (approved):

PR #3610, opened 2024-08-08
[torch] Add integer support for pooling operations

rsuderman:
If we pass an integer type to the pooling operation we incorrectly pad with an integer value with causes downstream compilation failures.


zjgarvey (approved):
Do we have any examples of integer max pool in the torch-mlir e2e tests? It might be good to add one with this patch.

rsuderman:
> Do we have any examples of integer max pool in the torch-mlir e2e tests? It might be good to add one with this patch.

This is mostly a quirk of our `onnx` lowerings. I am not certain that `torch` technically supports them via the python front end.

PR #3617, opened 2024-08-08
[onnx] Add support for `fp8` `onnx.DequantizeLinear`

rsuderman:
Fp8 needs a slightly different path for dequantization as the `torch` dequantize operation does not support `fp8` types.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -2117,41 +2117,73 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
             binder.tensorResultType(resultType))
           return failure();
 
+        auto loc = binder.getLoc();
         Value operand = operands[0];
         Value scale = operands[1];
         Value zeropoint = operands[2];
 
         auto operandTy = cast<Torch::ValueTensorType>(operand.getType());
 
+        auto operandETy = operandTy.getDtype();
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -2117,41 +2117,73 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+              loc, resultType, operand, tyConst,
+              /*non_blocking=*/cstFalse, /*copy=*/cstFalse,
+              /*memory_format=*/none);
+
+          Value one = rewriter.create<Torch::ConstantFloatOp>(
+              loc, rewriter.getF64FloatAttr(1.0));
+          Value sub = rewriter.create<Torch::AtenSubScalarOp>(
+              loc, resultType, toDtype, zeropoint, one);
+          rewriter.replaceOpWithNewOp<Torch::AtenMulScalarOp>(
+              binder.op, resultType, sub, scale);
```
zjgarvey, line 2177: With fp8 I assume there isn't any QDQ fusion stuff, right? It just uses normal fp ops with the lower bitwidth?

zjgarvey (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
rsuderman: From an onnx standpoint it does not appear to be as relevant. Overall `fp8` you can basically just treat as a scale and offset. There is a chance this will change over time but given how new `fp8` is it will likely change over time.

PR #3619, opened 2024-08-08
[onnx] Support `fp8` for `onnx.QuantizeLinear`

rsuderman:
We need to directly decompose quantize linear for `fp8` types as the equivalent torch operations do not support the operation.


zjgarvey (approved):
I swear I just reviewed this patch a second ago... *searches browser history* ah, right this is the other way.

rsuderman:
> I swear I just reviewed this patch a second ago... _searches browser history_ ah, right this is the other way.

No worries it happens.

PR #3622, opened 2024-08-09
[torch] Fix lowering for `onnx.Split` equal parts variant

vinayakdsci:
The TorchOnnxToTorch lowering for `onnx.Split`, equal parts variant, hardcodes the value `2` for `num_outputs`. When the `num_outputs` attr is not passed in, and the resultType, which returns variadic outputs, has number of outputs not equal to 2, the op does not successfully lower to `linalg`.

This patch replaces the hard-coded value of 2 with `binder.op->getResults().size()`, so that the difference in the number of outputs which causes assertions failures inside of LLVM, is fixed.


vinayakdsci:
Closing this PR, to check more into the errors, as this patch does not seem to be resolving the actual problem.

PR #3623, opened 2024-08-09
[torch] Add basic support for per-channel quantized graphs

ubfx:
This patch adds basic support for lowering graphs with per-channel quantization. Per-channel quantized ops have to be excluded from `FuseQuantizedOps` for now but can be used in QDQ quantized form.

Using this patch, we're able to import and execute (on the linalg backend) graphs with per-channel quantization applied using the "new" PyTorch 2.0 Export Quantization.


lib/Dialect/Torch/Transforms/FuseQuantizedOps.cpp
```diff
@@ -175,16 +190,16 @@ class QuantizeOperandsPastCommutingOps : public OpRewritePattern<SrcOp> {
...
-          cast<ValueTensorType>(dequantOpd.getType()).getOptionalDtype();
+          cast<ValueTensorType>(chain.dequantOpd.getType()).getOptionalDtype();
       auto newMPTQTType = rewriter.getType<ValueTensorType>(
           cast<ValueTensorType>(operands[i].getType()).getSizes(), qTorchType);
       operands[i] = rewriter.create<Aten_MakePerTensorQuantizedTensorOp>(
           loc, newMPTQTType, oldOpd, MPTQTOperands[1], MPTQTOperands[2]);
     }
 
-    if (!dequanted) {
+    if (operandChains.empty()) {
```
zjgarvey, line 202: I suppose this could get moved above the rewriting loop. 

Speaking of, how do you generally feel about returning a match failure if 
```cpp
operandChains.size() != QuantInfo<SrcOp>OperandsToQuantize.size()
```
?
This way, we won't get "half-quantized" ops, which generally cause issues lower in the pipeline. 
Pros of adding this match failure: more reliable compile
Cons of adding this match failure: less likely that people would notice their ops aren't getting QDQ'd properly.

Let me know what you think.

lib/Dialect/Torch/Transforms/FuseQuantizedOps.cpp
```diff
@@ -72,51 +76,61 @@ class QuantizeOperandsPastCommutingOps : public OpRewritePattern<SrcOp> {
...
 
       // move to next operand if this trace was unsuccessful
-      if (!MPTQTOpd)
+      if (!chain.MPTQTOpd)
         continue;
 
       // a successful trace occured, so set dequant to true
-      dequanted = true;
+      operandChains.push_back(std::move(chain));
+    }
```
zjgarvey, line 122: Probably don't need the early continue now that there is only one line of code aftewards. I'd like it if you either update the comment about the absent `dequant` bool (if you'd like to keep it like this), or make a change like:
```cpp
  // if trace was successful, add it to operandChains
  if (chain.MPTQTOpd)
    operandChains.push_back(std:move(chain));
}
```

zjgarvey (approved):
Nice! Overall, these changes look great. The QuantizedChain struct makes this code easier to read and handle.

lib/Dialect/Torch/Transforms/FuseQuantizedOps.cpp
ubfx: I think this is a good idea. The problem with half-quantized ops is also that there is little chance of salvaging them in downstream passes, while leaving them in the QDQ form still allows for later handling in specialized patterns.

I also agree with the Cons, but I think currently users will have to make sure that their IR gets quantized sensibly anyways. As our implementation becomes more comprehensive, e.g. also covering fusion for per-channel quantization, we could throw an error if an op remains in QDQ form until the end. At this point the user still has the option to apply his own transformations or run on a backend which supports the float point operator.

I've implemented this suggestion and have added a warning message emitted if at least one but not all operands were successfully traced to a supported per-tensor dequant Op. Let me know if you think this is too verbose.

zjgarvey (approved):
I think emitting the warning is a good compromise. 

This looks good to me!

lib/Dialect/Torch/Transforms/FuseQuantizedOps.cpp
zjgarvey: I'm totally happy with the op warning as a compromise. Thanks for the changes.

ubfx:
Thank you for the review!

PR #3624, opened 2024-08-09
[onnx] Fix `onnx.Hardmax` lowering to torch

rsuderman:
The lowering to torch makes assumption about the dimensions / types of reduce max and onehot. We need to correct for expected torch behavior.


lialan (approved):
Thanks for making it complete.

PR #3627, opened 2024-08-12

vivekkhandelwal1:
Set PyTorch and TorchVision version to nightly release 2024-08-12.
This commit also updates the `scaled_dot_product_attention` op. 
A new attribute `enable_gqa` has been added. As of now, only the
default value for the same is supported.


zjgarvey (approved):

qingyunqu:
I meet some problem when I intergrate this PR with old version torch. It will fail at `scaled_dot_product_attention` op's signature. Do we have any measure to provide compatibility on different version torch?

PR #3629, opened 2024-08-13
[Torch] emit upsample_nearest1d/2d/vec, and add shape/dtype functions

qingyunqu:


Vremold (approved):
LGTM. Thanks for the effort.

Issue #3628, opened 2024-08-12
aten.Multinomial tries to add f32 and f64 together


zjgarvey:
Trying out some e2e tests for onnx and the generated ir causes a crash when running torch-mlir to convert the torch ir to linalg due to mismatched types on an arith.addf op coming from the aten multinomial conversion.

I'll try to post a reproducer, but it should be an obvious fix (e.g., somewhere something with result element type is being added to something with f64 type).

PR #3630, opened 2024-08-13
[TorchToLinalg] address a dtype mismatch in `aten.multinomial` lowering

zjgarvey:
Resolves <https://github.com/llvm/torch-mlir/issues/3628>
Unblocks a compile failure for one of the MiGraphx models (`AgentModel`).


zjgarvey:
Huh, I don't even know what to say about a torchscript numerics mismatch...

zjgarvey:
This is catastrophically frustrating. Perhaps we should be using the TestUtils-generated sample args for the torchscript onnx export?? Neither zeros nor ones work uniformly for all tests.

pashu123 (approved):
One nit. LGTM!

PR #3631, opened 2024-08-14
[TorchToLinalg] Support torch.isclose lower to linalg

lingzhiz1998:
torch.isclose has tosa lowering but no linalg lowering. Add linalg lowering to fix the failure of isclose related tests on linalg backend.


zjgarvey (approved):
Great, thanks! Looking at the non-camel case op name makes me cringe a bit, but it is what it is.

PR #3639, opened 2024-08-15
[onnx] Fix `torch` lowering for determinant

rsuderman:
The determinant lowering had some extract / insert shape mismatches. Replumbed shape manipulations to correctly implement the determinant operation.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -3032,11 +3032,28 @@ public:
     Value cstZeroF = getConstant(rewriter, loc, 0, elemTy);
     // get some shapes
     SmallVector<int64_t> inputShape(inputType.getShape());
+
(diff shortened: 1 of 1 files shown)


zjgarvey (approved):
Ah, I see that iree-compile crashes on this op with the original implementation. Since it passed on the ref-backend (e.g. the e2e tests in torch-mlir) I didn't think to test further. 

Thanks for the fix, Rob. In general, do inserts and extracts need to be the same rank in order for iree to properly handle them?

rsuderman:
> Ah, I see that iree-compile crashes on this op with the original implementation. Since it passed on the ref-backend (e.g. the e2e tests in torch-mlir) I didn't think to test further.
> 
> Thanks for the fix, Rob. In general, do inserts and extracts need to be the same rank in order for iree to properly handle them?

So I assumed that they needed to be but apparently not? The failures happened super far into iree compilation so I would guess this is a delta between how generic codegen and iree code gen function.

PR #3643, opened 2024-08-16
[onnx] Decompose `onnx.Adam` into elementwise operations

rsuderman:
We can support `onnx.Adam` by decomposing into elementwise operations. The numerics slightly differ from `onnxruntime` however they appear to be within margins.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -175,7 +175,7 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
         SmallVector<Value> accSqGrads;
 
         size_t inputCount = (operands.size() - 2) / 4;
-        for (int i = 0; i < inputCount; ++i) {
+        for (size_t i = 0; i < inputCount; ++i) {
           inputs.push_back(operands[2 + i]);
           grads.push_back(operands[2 + i + inputCount * 1]);
           accGrads.push_back(operands[2 + i + inputCount * 2]);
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -147,6 +147,170 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+      "Adam", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        llvm::SmallVector<Value> operands;
+        if (binder.tensorOperandsList(operands))
+          return failure();
+
+        float alpha = 0.9;
+        float beta = 0.999;
+        float epsilon = 0.0;
+        float norm = 0.0;
+        float postNorm = 0.0;
```
vivekkhandelwal1, line 160: Is there a need to initialize all of them since the default value would be assigned during the binding itself? Also, can we declare them all in the same line?

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -147,6 +147,170 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          return failure();
+
+        auto loc = binder.getLoc();
+        Value r = operands[0];
+        Value t = operands[1];
+
+        SmallVector<Value> inputs;
+        SmallVector<Value> grads;
+        SmallVector<Value> accGrads;
+        SmallVector<Value> accSqGrads;
```
vivekkhandelwal1, line 175: Same here for declaration.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -147,6 +147,170 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+        SmallVector<Value> newX;
+        SmallVector<Value> newG;
+        SmallVector<Value> newSG;
+
+        for (int i = 0, s = inputs.size(); i < s; ++i) {
+
+          Value x = inputs[i];
+          Value g = grads[i];
+          Value v = accGrads[i];
+          Value h = accSqGrads[i];
```
vivekkhandelwal1, line 251: Same here as above.

vivekkhandelwal1 (changes requested):

Issue #3417, opened 2024-06-04
Failed to lower torch.aten.slice.Tensor to linalg


josel-amd:
Hi all,

I need to convert an instance of the `torch.aten.slice.Tensor` operator that has a dynamic `axis`/`dim`. The torch-mlir that I see and that can be used to reproduce the issue is shown in [0].

This results in the error messages shown below:

```
<unknown>:0: error: unimplemented: dim is not constant
<unknown>:0: note: see current operation: %59 = "torch.aten.slice.Tensor"(%6, %44, %24, %34, %54) : (!torch.vtensor<[20,10,5],f32>, !torch.int, !torch.int, !torch.int, !torch.int) -> !torch.vtensor<[?,10,5],f32>
<unknown>:0: error: failed to legalize operation 'torch.aten.slice.Tensor' that was explicitly marked illegal
<unknown>:0: note: see current operation: %51 = "torch.aten.slice.Tensor"(%6, %41, %23, %32, %50) : (!torch.vtensor<[20,10,5],f32>, !torch.int, !torch.int, !torch.int, !torch.int) -> !torch.vtensor<[?,10,5],f32>
```

The command to reproduce this error is `torch-mlir-opt --convert-torch-to-linalg <insert_filename_here>.mlir;`

[0]

```
#loc = loc(unknown)
module attributes {
  llvm.data_layout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-i128:128-f80:128-n8:16:32:64-S128",
  llvm.target_triple = "x86_64-unknown-linux-gnu",
  "onnx-mlir.symbol-postfix" = "onnxmodel.onnx.mlir"} {
  func.func @forward(%arg0: tensor<20x10x5xf32> {onnx.name = "x"} loc(unknown), %arg1: tensor<2xi64> {onnx.name = "starts"} loc(unknown), %arg2: tensor<2xi64> {onnx.name = "ends"} loc(unknown), %arg3: tensor<2xi64> {onnx.name = "axes"} loc(unknown), %arg4: tensor<2xi64> {onnx.name = "steps"} loc(unknown)) -> (!torch.vtensor<[?,10,5],f32> {onnx.name = "y"}) attributes {
    torch.onnx_meta.opset_version = 19 : si64} {
    %0 = torch.vtensor.literal(dense<1> : tensor<1xsi64>) : !torch.vtensor<[1],si64> loc(#loc)
[...]

PR #3644, opened 2024-08-16
[onnx] Fix `onnx.Slice` for dynamic axis

rsuderman:
If the axis value is dynamic we need to iterate through the options and match to the correct dim.


renxida (approved):
Seem to have missed this from a long time ago. My bad.

LGTM

TinaAMD:
Does this resolve https://github.com/llvm/torch-mlir/issues/3417?

PR #3648, opened 2024-08-19
[torch-mlir][sparse] replace ad-hoc mechanism with proper FX export

aartbik:
Now that the PyDev feature request pytorch/pytorch#117188 has been completed, we can remove all the ad-hoc code that propagates sparsity metadata and replace it with the built-int PyDev metadata for sparse tensors. This removes a lot of code and also ensures sparsity is consistent with the torch.sparse package for all cases.


aartbik:
@vivekkhandelwal1 is there anything special needed to tell the bots to update pytorch to your most recent version. I keep getting the error

  torch._dynamo.exc.Unsupported: torch.compile does not support sparse Tensor with torch.sparse_csr layout

at the bots, but when running locally with the right pip requirements, it all passes (as expected).

aartbik:
I added some debug on the torch version, and the bots see

`2.4.0+cpu`

but the requirements are torch==2.5.0.dev20240818.
When I pip install requirement locally and run, I get

`2.5.0.dev20240818+cpu`

in the debugging output as expected. Any way to fix the bot setup?

@jpienaar @stellaraccident @vivekkhandelwal1

aartbik:
Talked to @ramiro050 and made the test depend on 2.5.0+ (where sparsity prop became active)

PeimingLiu (approved):
nice！

PR #3650, opened 2024-08-19
Add per-test timeouts to catch infinite loops

renxida:
Previously we only had full suite timeouts, making it impossible to identify
which specific tests were hanging. This patch adds:

1. Per-test timeout support in the test framework
2. A default 600s timeout for all tests
3. A deliberately slow test to verify the timeout mechanism works

The timeout is implemented using Python's signal module. Tests that exceed
their timeout are marked as failures with an appropriate error message.

This should help catch and isolate problematic tests that enter infinite 
loops, without needing to re-run the entire suite multiple times.

### DIFF
--- a/projects/pt1/python/torch_mlir_e2e_test/registry.py
+++ b/projects/pt1/python/torch_mlir_e2e_test/registry.py
@@ -40,7 +40,7 @@ def register_test_case(
                 unique_name=f.__name__,
                 program_factory=module_factory,
                 program_invoker=f,
-                timeout=timeout_seconds,
+                timeout_seconds=timeout_seconds,
             )
         )
         return f
--- a/projects/pt1/python/torch_mlir_e2e_test/registry.py
+++ b/projects/pt1/python/torch_mlir_e2e_test/registry.py
@@ -16,7 +16,7 @@ _SEEN_UNIQUE_NAMES = set()
 
 
 def register_test_case(
-    module_factory: Callable[[], torch.nn.Module], timeout_seconds: int = 300
+    module_factory: Callable[[], torch.nn.Module], timeout_seconds: int = 120
 ):
     """Convenient decorator-based test registration.
 
--- a/projects/pt1/python/torch_mlir_e2e_test/registry.py
+++ b/projects/pt1/python/torch_mlir_e2e_test/registry.py
@@ -16,7 +16,7 @@ _SEEN_UNIQUE_NAMES = set()
 
 
 def register_test_case(
-    module_factory: Callable[[], torch.nn.Module], timeout: int = 600
+    module_factory: Callable[[], torch.nn.Module], timeout_seconds: int = 600
 ):
     """Convenient decorator-based test registration.
 
@@ -40,7 +40,7 @@ def register_test_case(
                 unique_name=f.__name__,
                 program_factory=module_factory,
                 program_invoker=f,
-                timeout=timeout,
+                timeout=timeout_seconds,
             )
         )
         return f
--- a/projects/pt1/python/torch_mlir_e2e_test/registry.py
+++ b/projects/pt1/python/torch_mlir_e2e_test/registry.py
@@ -16,7 +16,7 @@ _SEEN_UNIQUE_NAMES = set()
 
 
 def register_test_case(
-    module_factory: Callable[[], torch.nn.Module], timeout_seconds: int = 600
+    module_factory: Callable[[], torch.nn.Module], timeout_seconds: int = 300
 ):
     """Convenient decorator-based test registration.
 
--- a/projects/pt1/python/torch_mlir_e2e_test/registry.py
+++ b/projects/pt1/python/torch_mlir_e2e_test/registry.py
@@ -15,7 +15,7 @@ GLOBAL_TEST_REGISTRY = []
 _SEEN_UNIQUE_NAMES = set()
 
 
-def register_test_case(module_factory: Callable[[], torch.nn.Module]):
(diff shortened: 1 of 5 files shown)


projects/pt1/python/torch_mlir_e2e_test/registry.py
```diff
@@ -38,6 +40,7 @@ def decorator(f):
                 unique_name=f.__name__,
                 program_factory=module_factory,
                 program_invoker=f,
+                timeout=timeout,
```
ScottTodd, line 43: This is error prone without units. Minimally, include units, so `timeout_s` or `timeout_seconds`.

Alternately, use a unitless object that represents a duration like `timedelta`: https://docs.python.org/3/library/datetime.html#timedelta-objects. That can be converted to any units that you want at the location that units are needed. If you need to convert to different units or perform math with the duration, the object can handle that for you with no risk of using the wrong number of decimal places.

See also https://testing.googleblog.com/2017/11/

projects/pt1/python/torch_mlir_e2e_test/registry.py
```diff
@@ -15,7 +15,9 @@
 _SEEN_UNIQUE_NAMES = set()
 
 
-def register_test_case(module_factory: Callable[[], torch.nn.Module]):
+def register_test_case(
+    module_factory: Callable[[], torch.nn.Module], timeout: int = 600
+):
```
ScottTodd, line 20: > A default 600s timeout for all tests

10 minutes for an individual unit test seems like a _very_ long timeout. How about just 1 minute? or 3 minutes? What are the longest test times today (on typical hardware)?

(see also my other comment about units - without explicit units one might think this was measured in milliseconds, nanoseconds, or minutes)

projects/pt1/python/torch_mlir_e2e_test/test_suite/timeout.py
```diff
@@ -0,0 +1,30 @@
...
+        super().__init__()
+
+    @export
+    @annotate_args([None, ([-1, -1], torch.int64, True)])
+    def forward(self, x):
+        x_val = x.size(0)  # this is going to be 2
+        sum = 100
+        while x_val < sum:  # sum will always > 2
+            sum += 1
+        return sum
```
ScottTodd, line 25: I take it the intent here is to have this code get ingested by the framework / compiler / etc. then loop infinitely? That should be expressed in a comment.

Ideally that waiting behavior would be expressed directly, with a `sleep()`, `wait(never_signalled)`, or other code more directly related to time/synchronization, but it looks like this is working within the constraints of pytorch and the supported torch-mlir backends.

projects/pt1/python/torch_mlir_e2e_test/test_suite/timeout.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+@register_test_case(module_factory=lambda: TimeOutModule(), timeout=10)
+def TimeOutModule_basic(module, tu: TestUtils):
+    module.forward(tu.randint(6, 8, high=10))
```
ScottTodd, line 30: Is the `(tu.randint(6, 8,` part here load bearing to the test? Seems like that could be simplified

projects/pt1/python/torch_mlir_e2e_test/registry.py
renxida: great point. using timeout_seconds.

I personally love the timedelta idea but also i think in this case simplicity trumps elegance / adaptability

projects/pt1/python/torch_mlir_e2e_test/registry.py
renxida: fixed the naming too. keeping 600s but will try halving it until cases start failing.

projects/pt1/python/torch_mlir_e2e_test/test_suite/timeout.py
renxida: added comments & renamed variables

projects/pt1/python/torch_mlir_e2e_test/test_suite/timeout.py
renxida: nope. used "ones" instead.

projects/pt1/python/torch_mlir_e2e_test/registry.py
ScottTodd: Most test runners have a way to print the duration of each test case, sorted from longest to shortest. That's what I'd do if it were available here, instead of that sort of bisection search.

projects/pt1/python/torch_mlir_e2e_test/test_suite/timeout.py
```diff
@@ -0,0 +1,47 @@
...
+        # constant because the compiler will
+        # attempt to get rid of the whole loop
+        input_arg_2 = x.size(0)
+        sum = 100
+        while input_arg_2 < sum:  # sum will always > 2
+            sum += 1
+        return sum
+
+
+@register_test_case(module_factory=lambda: TimeOutModule(), timeout=10)
```
ScottTodd, line 45:
```suggestion
@register_test_case(module_factory=lambda: TimeOutModule(), timeout_seconds=10)
```

ScottTodd (approved):

renxida:
Shrinking timeout down to a reasonable ammount and merging.

We should get in another PR to make the tests print how much time they took / potentially rewrite the e2etest stuff to use a testing framework like pytest.

Issue #3651, opened 2024-08-19
Missing dimensionality information in torch


MaheshRavishankar:
As a follow up from https://github.com/iree-org/iree/issues/18229 it seems like there is some dimension information that is not being captured correctly in IR and recovering that in the program is pretty involved. This is the IR after `torch-finalizing-backend-type-conversion`

```
util.func public @torch_jit$async(%arg0: !hal.buffer_view, %arg1: !hal.fence, %arg2: !hal.fence) -> !hal.buffer_view attributes {inlining_policy = #util.inline.never, iree.abi.model = "coarse-fences", iree.abi.stub} {
  %c3 = arith.constant 3 : index
  %c2 = arith.constant 2 : index
  %c64_i64 = arith.constant 64 : i64
  %c12_i64 = arith.constant 12 : i64
  %c768_i64 = arith.constant 768 : i64
  %c512_i64 = arith.constant 512 : i64
  %c30522_i64 = arith.constant 30522 : i64
  %cst = arith.constant dense<9.99999996E-13> : tensor<f32>
  %cst_0 = arith.constant dense<2.000000e+00> : tensor<f32>
  %c2_i64 = arith.constant 2 : i64
  %cst_1 = arith.constant dense_resource<__onnx_constant_not_found_possibly_due_to_being_elided___8> : tensor<1x512xi64>
  %c0_i64 = arith.constant 0 : i64
  %cst_2 = arith.constant dense_resource<__onnx_constant_not_found_possibly_due_to_being_elided__> : tensor<1x512xi64>
  %cst_3 = arith.constant dense_resource<__onnx_constant_not_found_possibly_due_to_being_elided___1> : tensor<30522x768xf32>
  %cst_4 = arith.constant dense_resource<__onnx_constant_not_found_possibly_due_to_being_elided___2> : tensor<512x768xf32>
  %cst_5 = arith.constant dense_resource<__onnx_constant_not_found_possibly_due_to_being_elided___3> : tensor<2x768xf32>
  %cst_6 = arith.constant dense_resource<__onnx_constant_not_found_possibly_due_to_being_elided___4> : tensor<768xf32>
  %cst_7 = arith.constant dense_resource<__onnx_constant_not_found_possibly_due_to_being_elided___5> : tensor<768xf32>
  %cst_8 = arith.constant dense_resource<__onnx_constant_not_found_possibly_due_to_being_elided___6> : tensor<768xf32>
[...]

PR #3656, opened 2024-08-21
Add a canonicalization pattern for `aten.unflatten.int`

zjgarvey:
Addresses an issue in <https://github.com/llvm/torch-mlir/issues/3651> where some unflatten ops generated from onnx models weren't propagating static shape information. It may be necessary to add further optimizations for the more general case when some static information is present in the unflatten (or possibly reshape/view) op's `sizes` list, but not reflected in the output shape. These ops will only successfully infer shapes if the `sizes` list is gotten from a list of constant ints (with possibly one -1). A common example where this fails is when some of the `sizes` are determined from `aten.size.int` ops on dynamic tensors, and other `sizes` are known statically. 

This PR includes:
- a canonicalizer for `aten.unflatten.int` which converts to `aten.unsqueeze` when it is expanding one dim to two, and one of the new dims is statically 1. 
- an improvement to the folder for `aten.__or__.bool` which does not rely on *both* operands being static.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -2193,6 +2193,7 @@ void AtenUnflattenIntOp::getCanonicalizationPatterns(
     Value cstMOne = rewriter.create<Torch::ConstantIntOp>(op.getLoc(), -1);
     // the runtime asserts below are introduced to catch malformed unflatten ops
     // possibly generated from onnx IR.
+    FailureOr<Value> maybeUnsqueeze;
     if (dim0 == 1) {
       // check if the remaining size value is either -1 or equal to original
       // size at dim
@@ -2208,9 +2209,7 @@ void AtenUnflattenIntOp::getCanonicalizationPatterns(
       rewriter.create<Torch::RuntimeAssertOp>(
           op.getLoc(), isMOneOrSameSize,
           rewriter.getStringAttr("unflatten sizes must be compatible"));
-      rewriter.replaceOpWithNewOp<AtenUnsqueezeOp>(op, op.getType(), self,
-                                                   unflattenDim);
-      return success();
+      maybeUnsqueeze = Torch::unsqueezeTensor(rewriter, op, self, unflattenDim);
     }
     if (dim1 == 1) {
       // check if the remaining size value is either -1 or equal to original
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2162,6 +2166,79 @@ void AtenSizeOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+      rewriter.create<Torch::RuntimeAssertOp>(
+          op.getLoc(), isMOneOrSameSize,
+          rewriter.getStringAttr("unflatten sizes must be compatible"));
+      // unsqueeze at dim + 1
+      Value cstOne = rewriter.create<Torch::ConstantIntOp>(op.getLoc(), 1);
+      Value dimPlusOne =
+          rewriter.create<AtenAddIntOp>(op.getLoc(), unflattenDim, cstOne);
+      maybeUnsqueeze = Torch::unsqueezeTensor(rewriter, op, self, dimPlusOne);
+    }
+    if (failed(maybeUnsqueeze))
```
rsuderman, line 2234: You will want to move this check before constructing operators. Generally creating and immediately destroying operations can cause some weird loops.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2162,6 +2166,79 @@ void AtenSizeOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+          op.getLoc(), selfSizeAtDim, sizeValues[1]);
+      Value isMinusOne =
+          rewriter.create<AtenEqIntOp>(op.getLoc(), cstMOne, sizeValues[1]);
+      Value isMOneOrSameSize = rewriter.create<Aten__Or__BoolOp>(
+          op.getLoc(), isMinusOne, isSameSize);
+      // unsqueeze at dim
+      rewriter.create<Torch::RuntimeAssertOp>(
+          op.getLoc(), isMOneOrSameSize,
+          rewriter.getStringAttr("unflatten sizes must be compatible"));
+      maybeUnsqueeze = Torch::unsqueezeTensor(rewriter, op, self, unflattenDim);
```
rsuderman, line 2212: We should know whether we can unsqueeze before attempting to do some in this case.

rsuderman (changes requested):

lib/Dialect/Torch/IR/TorchOps.cpp
zjgarvey: With the prior checks, this unsqueeze operation really shouldn't fail. Regardless, I moved it to the earliest possible position, just in case.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -739,12 +739,16 @@ OpFoldResult Aten__Not__Op::fold(FoldAdaptor adaptor) {
...
-  if (!valueA || !valueB) {
+  if (!valueA && !valueB)
     return nullptr;
-  }
-
-  return IntegerAttr::get(IntegerType::get(getContext(), 1),
-                          valueA.getValue() | valueB.getValue());
+  if ((valueA && valueA.getValue() == 1) || (valueB && valueB.getValue() == 1))
+    return IntegerAttr::get(IntegerType::get(getContext(), 1), 1);
+  if (valueA.getValue() == 0)
```
qingyunqu, line 746: `valueA` may be nullptr here. This will cause `segmentation fault`.

qingyunqu: The specific case is that `valueA == nullptr` and `valueB == 0`.

rsuderman (approved):

PR #3657, opened 2024-08-22
[FxImporter] Fix sympy_int_to_int utility

patel-vimal:
New sympy type is introduced to represent integer infinity in upstream PyTorch repo. Subsequently, sympy.oo is no longer used to represent infinity upper bound for dynamic dimensions where the upper bound is unknown. Instead `int_oo` is used to represent integer infinity. This commit updates the `_sympy_int_to_int` utility in light of this change.

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -78,6 +78,16 @@ except ModuleNotFoundError:
     # conditional.
     ml_dtypes = None
 
+try:
+    from torch.utils._sympy.numbers import int_oo, IntInfinity, NegativeIntInfinity
+except ModuleNotFoundError:
+    # This commit on PyTorch repo introduced IntInfinity and NegativeIntInfinity:
+    # https://github.com/pytorch/pytorch/commit/2229884102ac95c9dda0aeadbded1b04295d892e
+    # Required module may not be present in the stable version of PyTorch.
+    int_oo = None
+    IntInfinity = None
+    NegativeIntInfinity = None
+
 from torch.fx.node import (
     Argument as NodeArgument,
 )
@@ -124,6 +134,8 @@ from ..ir import (
 from ..dialects import (
     func as func_dialect,
 )
+
+from .._mlir_libs._torchMlir import get_int64_max, get_int64_min
 
 __all__ = [
     "FxImporter",
@@ -1165,22 +1177,32 @@ class ContextCache:
         self, prog: torch.export.ExportedProgram
     ) -> Dict[str, RangeConstraint]:
 
+        # Recent PyTorch versions use `int_oo` to represent integer infinity.
+        # Older PyTorch versions like PyTorch stable version may not have
+        # `int_oo` defined just yet.
+        infs = (sympy.oo, int_oo) if int_oo is not None else (sympy.oo,)
+
         def _sympy_int_to_int(val: sympy.Expr, adjust_func: Callable):
             # Convert simple sympy Integers into concrete int
-            if val == sympy.oo:
-                return math.inf
-            if val == -sympy.oo:
-                return -math.inf
+            if val in infs:
+                return get_int64_max()
+            if val in tuple(-inf for inf in infs):
+                return get_int64_min()
             if isinstance(val, sympy.Integer):
                 return int(val)
             # TODO: Remove this adjustment when fractional ranges are removed
             return adjust_func(val)
 
         contains_symbolic_ints = False
+        sym_int_types = (
+            (sympy.Integer, IntInfinity, NegativeIntInfinity)
+            if IntInfinity is not None
+            else sympy.Integer
+        )
         for val in prog.range_constraints.values():
             if (
-                isinstance(val.lower, sympy.Integer)
-                and isinstance(val.upper, sympy.Integer)
+                isinstance(val.lower, sym_int_types)
+                and isinstance(val.upper, sym_int_types)
                 and not val.is_bool
             ):
                 contains_symbolic_ints = True
--- a/test/python/fx_importer/basic_test.py
+++ b/test/python/fx_importer/basic_test.py
@@ -88,12 +88,13 @@ def test_import_frozen_exported_program_with_func_name():
 
 @run
 # CHECK-LABEL: test_import_frozen_exported_program_with_dynamic_shapes
-# CHECK:     func.func @test_net(%[[ARG0:[a-zA-Z0-9]+]]: !torch.vtensor<[?,4],f32>) -> !torch.vtensor<[?,4],f32>
+# CHECK:     func.func @test_net(%[[ARG0:[a-zA-Z0-9]+]]: !torch.vtensor<[?,?,5],f32>) -> !torch.vtensor<[?,?,5],f32>
 # CHECK:     %[[S0:.*]] = torch.symbolic_int "s0" {min_val = {{[0-9]+}}, max_val = {{[0-9]+}}} : !torch.int
-# CHECK:     torch.bind_symbolic_shape %[[ARG0]], [%[[S0]]], affine_map<()[s0] -> (s0, 4)> : !torch.vtensor<[?,4],f32>
-# CHECK:     %[[TANH:.*]] = torch.aten.tanh %[[ARG0]] : !torch.vtensor<[?,4],f32> -> !torch.vtensor<[?,4],f32>
-# CHECK:     torch.bind_symbolic_shape %[[TANH]], [%[[S0]]], affine_map<()[s0] -> (s0, 4)> : !torch.vtensor<[?,4],f32>
-# CHECK:     return %[[TANH]] : !torch.vtensor<[?,4],f32>
+# CHECK:     %[[S1:.*]] = torch.symbolic_int "s1" {min_val = 2, max_val = {{[0-9]+}}} : !torch.int
+# CHECK:     torch.bind_symbolic_shape %[[ARG0]], [%[[S0]], %[[S1]]], affine_map<()[s0, s1] -> (s0, s1, 5)> : !torch.vtensor<[?,?,5],f32>
+# CHECK:     %[[TANH:.*]] = torch.aten.tanh %[[ARG0]] : !torch.vtensor<[?,?,5],f32> -> !torch.vtensor<[?,?,5],f32>
+# CHECK:     torch.bind_symbolic_shape %[[TANH]], [%[[S0]], %[[S1]]], affine_map<()[s0, s1] -> (s0, s1, 5)> : !torch.vtensor<[?,?,5],f32>
+# CHECK:     return %[[TANH]] : !torch.vtensor<[?,?,5],f32>
 def test_import_frozen_exported_program_with_dynamic_shapes():
     class Basic(nn.Module):
         def __init__(self):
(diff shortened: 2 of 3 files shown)


test/python/fx_importer/basic_test.py
```diff
@@ -88,12 +88,13 @@ def forward(self, x):
...
 @run
 # CHECK-LABEL: test_import_frozen_exported_program_with_dynamic_shapes
-# CHECK:     func.func @test_net(%[[ARG0:[a-zA-Z0-9]+]]: !torch.vtensor<[?,4],f32>) -> !torch.vtensor<[?,4],f32>
+# CHECK:     func.func @test_net(%[[ARG0:[a-zA-Z0-9]+]]: !torch.vtensor<[?,?,5],f32>) -> !torch.vtensor<[?,?,5],f32>
 # CHECK:     %[[S0:.*]] = torch.symbolic_int "s0" {min_val = {{[0-9]+}}, max_val = {{[0-9]+}}} : !torch.int
-# CHECK:     torch.bind_symbolic_shape %[[ARG0]], [%[[S0]]], affine_map<()[s0] -> (s0, 4)> : !torch.vtensor<[?,4],f32>
-# CHECK:     %[[TANH:.*]] = torch.aten.tanh %[[ARG0]] : !torch.vtensor<[?,4],f32> -> !torch.vtensor<[?,4],f32>
-# CHECK:     torch.bind_symbolic_shape %[[TANH]], [%[[S0]]], affine_map<()[s0] -> (s0, 4)> : !torch.vtensor<[?,4],f32>
-# CHECK:     return %[[TANH]] : !torch.vtensor<[?,4],f32>
+# CHECK:     %[[S1:.*]] = torch.symbolic_int "s1" {min_val = 2, max_val = 9223372036854775806} : !torch.int
```
sjain-stanford, line 93: JFYI we kept `min_val` check open (`min_val = {{[0-9]+}}`) because when not set it is 0 with the nightlies and 2 with stable; this is short-lived - once we bump PT stable we can make it explicit `min_val = 0` everywhere. You're fine since you're setting it in for channel dim to 2.

Also, max_val should actually be `9223372036854775807` (off by 1) once you fix the bug above.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -273,6 +282,11 @@
 # so split by + and 0 index will always give the base version
 _IS_TORCH_2_1_OR_EARLIER = torch.__version__.split("+")[0] <= "2.1.0"
 
+# The value here is chosen to match that of sys.maxint (or sys.maxsize in case
+# of Python3.x). This is to maintain compatibility with older version of PyTorch.
+MLIR_INT_INFINITY = (2**63 - 1) - 1
```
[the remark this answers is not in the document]
sjain-stanford: To avoid errors like these, why not use `INT64_MAX` and `INT64_MIN` from LLVM/stdint.h directly?
https://sourcegraph.com/search?q=repo:%5Egithub%5C.com/llvm/llvm-project%24+INT64_MAX&patternType=keyword&sm=0

sjain-stanford (approved):
LGTM.

python/torch_mlir/extras/fx_importer.py
```diff
@@ -125,6 +135,8 @@
     func as func_dialect,
 )
 
+from .._mlir_libs._torchMlir import get_int64_max, get_int64_min
```
ScottTodd, line 138: We're running into issues with this line of code downstream: https://github.com/iree-org/iree/pull/18388#discussion_r1736600593

The relative path (into a directory that isn't a full package with a `__init__.py`? I'm not sure) is showing up as missing: 
```
wheelhouse/test.venv/lib/python3.11/site-packages/shark_turbine/aot/support/ir_utils.py:16: in <module>
    from iree.compiler.extras.fx_importer import (
wheelhouse/test.venv/lib/python3.11/site-packages/iree/compiler/extras/fx_importer.py:138: in <module>
    from .._mlir_libs._torchMlir import get_int64_max, get_int64_min
E   ModuleNotFoundError: No module named 'iree.compiler._mlir_libs._torchMlir'
```

I'm not sure if we're missing some setup downstream to make this work, but do you really need to jump over to C++ to get two integer values? Can you use [`sys.maxsize`](https://docs.python.org/3/library/sys.html#sys.maxsize) or `2**63 - 1` for the max and something equivalent for the min? Or some other sentinel value as needed.

sjain-stanford: One issue with `sys.maxsize` is it'd be platform dependent (`2**31 - 1` on 32-bit vs `2**63 - 1` on 64-bit), possibly causing spurious behavior (depending on where it is run), and/or deviation from PT. Hardcoding it sounds OK to unblock (unless there's a better way to address the relative path issue downstream).

ScottTodd: I think our downstream use is missing this: https://github.com/llvm/torch-mlir/blob/98e08023bbf71e00ab81e980eac9f7c96f1f24b4/setup.py#L223-L225

(here's our current list: https://github.com/iree-org/iree/blob/3ceffb60490107ae76a9c20b8a1588d560be0218/compiler/setup.py#L432-L441)

I'm not sure if we want to take on that dependency though, especially for two integers. @stellaraccident any ideas?

zjgarvey: How about using something like `torch.iinfo(torch.int64)`<https://pytorch.org/docs/stable/type_info.html>?

sjain-stanford: That would work. Thanks for sending the fix!

vivekkhandelwal1:
I'm not sure if this commit is the reason for this https://github.com/llvm/torch-mlir-release/actions/runs/10791102951/job/29927793865 failure or not but it seems that this is one of the commit between the passing and failing release build.

@sjain-stanford Do you think this might be a reason for failure?

stellaraccident:
I'm sorry I missed this. Please revert this commit: the dependencies of this code are tightly controlled by design and we need to find another way to do this.

python/torch_mlir/extras/fx_importer.py
stellaraccident: For the future: this file must be self contained. If I'd caught this, I would have commented when the patch was out up. If this has already been fixed forward, fine. Otherwise, please revert and reapply without the native code changes.

stellaraccident:
> I'm not sure if this commit is the reason for this https://github.com/llvm/torch-mlir-release/actions/runs/10791102951/job/29927793865 failure or not but it seems that this is one of the commit between the passing and failing release build.
> 
> @sjain-stanford Do you think this might be a reason for failure?

That looks unlikely but related to the code path chi was just working on. Should not have affected CI though. We really need to disable that completely soon.

vivekkhandelwal1:
@patel-vimal I'm not able to revert this patch, you have to do it manually. Please revert the patch for now, and add this contribution later in an acceptable way.

sjain-stanford:
Sorry for chiming in late. I believe we just need [these two lines](https://github.com/llvm/torch-mlir/blob/main/python/TorchMLIRModule.cpp#L32-L34) removed (it's not being used after https://github.com/llvm/torch-mlir/pull/3685). WDYT @ScottTodd , @patel-vimal , @stellaraccident ?

ScottTodd:
> Sorry for chiming in late. I believe we just need [these two lines](https://github.com/llvm/torch-mlir/blob/main/python/TorchMLIRModule.cpp#L32-L34) removed (it's not being used after #3685). WDYT @ScottTodd , @patel-vimal , @stellaraccident ?

Those two lines are unused and could certainly be removed. I'm not sure if they would be causing the failure though. Can anyone debug what that workflow runs locally to check?

sjain-stanford:
> I'm not sure if they would be causing the failure though. Can anyone debug what that workflow runs locally to check?

If you're referring to the release build failure, I am giving it a shot locally. One sec.

sjain-stanford:
Build still fails with the same error, so it does seem unlikely that the release failure is related to changes from this PR or the follow-on fixes.
```
  [4049/4049] Linking CXX shared module tools/torch-mlir/python_packages/torch_mlir/torch_mlir/_mlir_libs/_mlir.cpython-310-x86_64-linux-gnu.so
  Removing _mlir_mlibs dir to force rebuild: /main_checkout/torch-mlir/setup_build/cmake_build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir/_mlir_libs
  cmake config: cmake -DCMAKE_BUILD_TYPE=Release -DPython3_EXECUTABLE=/opt/python/cp310-cp310/bin/python -DPython3_FIND_VIRTUALENV=ONLY -DMLIR_ENABLE_BINDINGS_PYTHON=ON -DLLVM_TARGETS_TO_BUILD=host -DLLVM_ENABLE_ZSTD=OFF -DCMAKE_VISIBILITY_INLINES_HIDDEN=ON -DCMAKE_C_VISIBILITY_PRESET=hidden -DCMAKE_CXX_VISIBILITY_PRESET=hidden -DTORCH_MLIR_ENABLE_LTC=OFF -DTORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS=ON -DLLVM_ENABLE_PROJECTS=mlir -DLLVM_EXTERNAL_PROJECTS='torch-mlir' -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR=/main_checkout/torch-mlir /main_checkout/torch-mlir/externals/llvm-project/llvm
  cmake build: cmake --build . --config Release --target TorchMLIRPythonModules -- -j16
  cmake workspace: /main_checkout/torch-mlir/setup_build/cmake_build
  error: [('/main_checkout/torch-mlir/setup_build/cmake_build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/onnx_backends/abc.py', 'setup_build/lib.linux-x86_64-cpython-310/torch_mlir_e2e_test/onnx_backends/abc.py', "[Errno 2] No such file or directory: '/main_checkout/torch-mlir/setup_build/cmake_build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/onnx_backends/abc.py'"), ('/main_checkout/torch-mlir/setup_build/cmake_build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/onnx_backends/linalg_on_tensors.py', 
  'setup_build/lib.linux-x86_64-cpython-310/torch_mlir_e2e_test/onnx_backends/linalg_on_tensors.py', 
  "[Errno 2] No such file or directory: '/main_checkout/torch-mlir/setup_build/cmake_build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/onnx_backends/linalg_on_tensors.py'"), 
  ('/main_checkout/torch-mlir/setup_build/cmake_build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/onnx_backends/__init__.py', 'setup_build/lib.linux-x86_64-cpython-310/torch_mlir_e2e_test/onnx_backends/__init__.py', 
  "[Errno 2] No such file or directory: '/main_checkout/torch-mlir/setup_build/cmake_build/tools/torch-mlir/python_packages/torch_mlir/torch_mlir_e2e_test/onnx_backends/__init__.py'")]
  error: subprocess-exited-with-error
  
  × Building wheel for torch-mlir (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> See above for output.
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  full command: /opt/python/cp310-cp310/bin/python /opt/python/cp310-cp310/lib/python3.10/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py build_wheel /tmp/tmppac3ayn_
  cwd: /main_checkout/torch-mlir
  Building wheel for torch-mlir (pyproject.toml): finished with status 'error'
  ERROR: Failed building wheel for torch-mlir
Failed to build torch-mlir
ERROR: Failed to build one or more wheels
```

PR #3660, opened 2024-08-22
[TorchToLinalg] Use `linalg.transpose` instead of `generic` when lowering `aten.T`

ubfx:
The lowering pattern for `aten.T` uses transposition implemented via `linalg.generic`. For downstream passes it is advantageous to use named ops wherever possible, so this patch changes the lowering to use `linalg.transpose` instead.


CoTinker:
Can you add some testcases.

zjgarvey (approved):
I'm not sure why we weren't using the named op to begin with. Thanks!

A lit test might be nice to add like the other person mentioned, but it passes e2e tests, so I'll give the approval and leave it up to you whether to add it or not.

PR #3661, opened 2024-08-22
[TorchToLinalg] Use `linalg.broadcast` instead of `generic` for conv bias

ubfx:
The current implementation uses a `linalg.generic` to broadcast the bias tensor for the lowering of convolutions. This is suboptimal for later pattern matching. This patch changes it to use the respective named op, `linalg.broadcast`, instead.


zjgarvey (approved):
Nice. LGTM.

PR #3668, opened 2024-08-27
[Torch] Extract TensorPlaceholder to a common interface

penguin-wwy:
TensorPlaceholder does not depend on torchscript interfaces. After extraction, e2etest will not introduce torchscript because of TensorPlaceholder


stellaraccident (approved):
Thanks

PR #3671, opened 2024-08-27
[MLIR][TORCH] Add E2E support for aten.polar op

vivekkhandelwal1:


projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py
```diff
@@ -2591,6 +2594,15 @@ def aten〇softshrink〡dtype(self_rank_dtype: Tuple[int, int], lambd: Union[int
     self_rank, self_dtype = self_rank_dtype
     return _get_dtype_of_floating_point_op(self_dtype)
 
+
+def aten〇polar〡dtype(abs_rank_dtype: Tuple[int, int], angle_rank_dtype: Tuple[int, int]) -> int:
+    abs_rank, abs_dtype = abs_rank_dtype
```
zjgarvey, line 2599: Maybe add an assert that the `abs_dtype` matches the `angle_dtype`?

zjgarvey (commented):
The conversion looks good to me. I was wondering if you had considered writing a decomposition for this op instead of a direct conversion to linalg? If so, what were your reasons for choosing one over the other?

vivekkhandelwal1:
> The conversion looks good to me. I was wondering if you had considered writing a decomposition for this op instead of a direct conversion to linalg? If so, what were your reasons for choosing one over the other?

I was thinking of doing the same but the issue was that I didn't find any op that can take the real and imag values and create a complex value tensor. There might exist an op for the same but the lowering for the same has to be added.

vivekkhandelwal1:
Hi @zjgarvey, can you please review this PR?

zjgarvey (approved):

PR #3675, opened 2024-08-29
Fix onnx.Gather lowering with dynamic shapes

jinchen62:
Supports the result with dynamic shape and scalar indices like
```
func.func @test_gather_scalar(%arg0: !torch.vtensor<[3,4,5],f32>, %arg1: !torch.vtensor<[], si64>) -> !torch.vtensor<[?,?],f32> attributes {torch.onnx_meta.opset_version = 13 : si64} {
  %0 = torch.operator "onnx.Gather"(%arg0, %arg1) {torch.onnx.axis = 0 : si64} : (!torch.vtensor<[3,4,5],f32>, !torch.vtensor<[], si64>) -> !torch.vtensor<[?,?],f32>
}
```

`Torch::AtenSqueezeOp` is referring to the result shape, so it will failed on lowering if the result shape is dynamic.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1941,7 +1941,6 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
             indicesCt = Torch::kUnknownSize;
             break;
           }
-
           indicesCt *= sz;
         }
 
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1976,8 +1975,16 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        // 0 indicesRank will always squeeze the axis dim
+        // Use PrimsSqueezeOp for the case of result with dynamic shape
+        SmallVector<Value> dimList({index});
+        Value dimValueList = rewriter.create<Torch::PrimListConstructOp>(
+            loc,
+            rewriter.getType<Torch::ListType>(
+                rewriter.getType<Torch::IntType>()),
+            dimList);
+        rewriter.replaceOpWithNewOp<Torch::PrimsSqueezeOp>(
+            binder.op, resultType, gather, dimValueList);
```
zjgarvey, line 1987: Why use `PrimsSqueezeOp` when the dim list only has one element? You should be able to use the util function `Torch::squeezeTensor(rewriter, binder.op, axis, gather)`, which is what the decomposition for `PrimsSqueezeOp` ends up using anyway.

jinchen62: I didn't know the util function. So when the indices is a scalar, it will select 1 from the axis dim and squeeze it. The result rand is `q + (r - 1)` where `q` is input rand and `r` is indices rank. I changed it `AtenSqueezeDimOp` instead.

zjgarvey (approved):

PR #3676, opened 2024-08-29
[onnx] Relax constraints on input tensors in `onnx.STFT` conversion to torch dialect

giacs-epic:
- When the signal tensor is real, onnx allows its shape to be `[batch][length]` as well as `[batch][length][1]`.
- Onnx also allows to specify `frame_length` together with `window` (not empty), given that it matches the window size.
- Adding checks on signal and result shapes.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -3561,15 +3561,31 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         Value signal = operands[0];
         Value frameStep = operands[1];
         auto signalTy = cast<Torch::ValueTensorType>(signal.getType());
+        if (!signalTy || !signalTy.hasSizes()) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected signal type having sizes");
+        }
         auto signalShape = signalTy.getSizes();
+        if (signalShape.size() != 2 && signalShape.size() != 3) {
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "signal has invalid shape.");
+        }
+        if (!resultType || !resultType.hasSizes()) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected result type having sizes");
+        }
         auto resultShape = resultType.getSizes();
+        if (resultShape.size() != 4) {
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "result has invalid shape.");
+        }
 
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -3561,15 +3561,31 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
         Value frameStep = operands[1];
         auto signalTy = cast<Torch::ValueTensorType>(signal.getType());
+        if (!signalTy || !signalTy.hasSizes()) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected signal type having sizes");
+        }
         auto signalShape = signalTy.getSizes();
+        if (signalShape.size() != 2 && signalShape.size() != 3) {
+          return rewriter.notifyMatchFailure(binder.op,
+                                             "signal has invalid shape.");
```
vivekkhandelwal1, line 3571: Hi @giacs-epic, I don't see it written anywhere in the definition of STFT op (https://onnx.ai/onnx/operators/onnx__STFT.html), that the signal tensor can be of rank 2. Can you please add any references to it?

giacs-epic: Hi @vivekkhandelwal1, thank you for reviewing this PR!

You are right that in the ONNX docs there is no reference to the signal tensor possibly being of rank 2.
We encountered however one instance of a valid model that use a rank-2 signal and it turns out that the infrastructure of ONNX and onnxruntime supports such case.
See for instance the shape inference implementation of STFT: https://github.com/onnx/onnx/blob/060589cb81dfb081ed912c9e722b15fe1dbc1a14/onnx/defs/math/defs.cc#L3475-L3477
Also, here is a reproducible that runs such case with onnxruntime: https://gist.github.com/giacs-epic/330a73e147b487417f32f07d05d07e66

Please let me know if you think this kind of flexibility can be allowed and, in case, how it should be documented in the changes.

vivekkhandelwal1: Hi @giacs-epic, thanks for the explanation. You should add this reference with the short explanation as a comment in the code itself. Rest all, looks good to me.

vivekkhandelwal1 (approved):

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
giacs-epic: Done.

PR #3679, opened 2024-08-30
[TorchToLinalg][test] Add test for ConvertAtenConvolutionOp

CoTinker:
This patch add a test for 638ef14, which use `linalg.broadcast` instead of `generic` for convolution bias.


ubfx:
I think this won't work because it depends on my PR so it will fail the pre-merge checks. But if you want, I can pull this commit into my PR, then it should pass.

CoTinker:
I mean.

CoTinker:
Oh, Maybe my verison is too old. I'll fix it

CoTinker:
I accidentally clicked the `enabled auto-merge`, my fault.

PR #3680, opened 2024-08-30
Add canonicalize pattern for aten.mul.int and aten.floordiv.int

zezhang:
This PR add `floordiv` to the `PY_BUILTIN_TO_TORCH_OP`. For `aten.mul.int` and `aten.floordiv.int` ops, we add new Canonicalization Patterns as follow:

```
%1 = torch.aten.mul.int %input, %const-5
%2 = torch.aten.mul.int %1, %const-6
```

Will be replaced by

`torch.aten.mul.int %input, %const-30`

And 

```
%1 = torch.aten.mul.int %input, %const-5
%2 = torch.aten.floordiv.int %1, %const-5
```
Will directly return `%input`

This PR also relaxes the `float` type constraint in TorchToTosa for the `AtenRsubScalarOp` conversion.

To test:


sjain-stanford (approved):
LGTM

PR #3681, opened 2024-08-31
emit Torch_AtenAnyDimsOp

duck7216:
as title


vivekkhandelwal1 (commented):
Hi @duck7216, why do we just want to emit this op without adding any support for it?

PR #3684, opened 2024-09-03
[MLIR][ONNX] Add suport for zero ranked input for Onnx.Transpose op

vivekkhandelwal1:

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -1886,6 +1886,13 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
 
         // Default permutation is to reverse orders:
         int64_t rank = tensorType.getRank();
+
+        // Case: Rank-0 Input
+        if (rank == 0) {
+          rewriter.replaceOp(binder.op, operand);
+          return success();
+        }
+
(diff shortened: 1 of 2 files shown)


zjgarvey:
Can you provide some context? Without this patch, I am already able to successfully compile the lit test example you provided.

My concern is that this will simply cause poorly formed IR to compile incorrectly. For example, if the importer misses shape information for a transpose with non-trivial perms, and imports it as a rank0 tensor, this will allow the imported IR to compile incorrectly to an identity map. 

ONNX does not allow the perm list to be longer than the rank, even for rank 0 tensors. Trying to load such a model in onnxruntime throws an error, and so should we.

vivekkhandelwal1:
> Can you provide some context? Without this patch, I am already able to successfully compile the lit test example you provided.
> 
> My concern is that this will simply cause poorly formed IR to compile incorrectly. For example, if the importer misses shape information for a transpose with non-trivial perms, and imports it as a rank0 tensor, this will allow the imported IR to compile incorrectly to an identity map.
> 
> ONNX does not allow the perm list to be longer than the rank, even for rank 0 tensors. Trying to load such a model in onnxruntime throws an error, and so should we.

I thought this issue was genuine and the support was missing even though there's no point of transposing a zero-ranked tensor. But it seems the issue is with the Onnx IR itself. I have posted the same on the slack channel.

vivekkhandelwal1:
This is not required now, since:
1.) The issue was false positive for which the support was needed.
2.) Adding this support won't be semantically correct.

PR #3685, opened 2024-09-03
[fximporter] Avoid importing from `_torchMlir`

zjgarvey:
Downstream projects don't necessarily register this C++ module. This package removes the dependency and uses `torch.iinfo` to access the max and min values instead.

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -135,7 +135,6 @@ from ..dialects import (
     func as func_dialect,
 )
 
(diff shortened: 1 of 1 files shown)


ScottTodd (approved):

python/torch_mlir/extras/fx_importer.py
```diff
@@ -135,7 +135,6 @@
     func as func_dialect,
 )
 
-from .._mlir_libs._torchMlir import get_int64_max, get_int64_min
```
ScottTodd, line 138: Oh, these functions can also be removed from `python/TorchMLIRModule.cpp` (they were only added there to be used from this file)

PR #3686, opened 2024-09-04
Implement lowering of torch.column_stack

BaneTrifa:


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -3813,6 +3813,24 @@ class DecomposeAtenStackOp : public OpRewritePattern<AtenStackOp> {
...
+class DecomposeAtenColumnStackOp : public OpRewritePattern<AtenColumnStackOp> {
+public:
+  using OpRewritePattern::OpRewritePattern;
+  LogicalResult matchAndRewrite(AtenColumnStackOp op,
+                                PatternRewriter &rewriter) const override {
+
+    rewriter.replaceOpWithNewOp<AtenCatOp>(
+        op, op.getType(), op.getTensors(),
+        rewriter.create<Torch::ConstantIntOp>(op.getLoc(),
+                                              rewriter.getI64IntegerAttr(1)));
```
vivekkhandelwal1, line 3827: Hi @BaneTrifa, you can't directly replace the `column_stack` op with the `cat` op, since there might be some tensors which have a rank < 2, and in that case this would result in generating a wrong IR. See here: https://github.com/pytorch/pytorch/blob/b5be4d8c053e22672719b9a33386b071daf9860d/torch/_refs/__init__.py#L2839-L2842

PR #3687, opened 2024-09-05
[tmtensor] Add support for i64 index type for tm_tensor.scatter

rsuderman:
This restriction is a bit needless and could be wrong for larger memory blocks.


lib/Conversion/TorchToTMTensor/TorchToTMTensor.cpp
```diff
@@ -915,9 +914,6 @@ class ConvertAtenIndexPutHackedTwinOp
         rewriter.create<AtenViewOp>(loc, valuesType, values, valuesDimsList);
 
     // `TMTensor::ScatterOp` expects indices of element type i32.
```
marbre, line 916:
```suggestion
```

Seems this comment should be dropped as well.

marbre (approved):

PR #3688, opened 2024-09-05
[FxImporter] Add InputInfo to Resolve Literal Hook

chrsmcgrr:
Fixes iree-org/iree-turbine#85

iree-turbine uses the FxImporter hooks to materialize buffers as global constants. In this particular use case we have a buffer that is modified which requires the mutable modifier.

However, this information is not available in the current API as the InputInfo structure is not passed to the literal resolution.

The following change passes the information to the literal API of the FxImporterHook.

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -470,7 +470,7 @@ class FxImporterHooks:
         ...
 
     def resolve_literal(
-        self, gni: "GraphNodeImporter", literal: Any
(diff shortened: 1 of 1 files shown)


penguin-wwy (approved):

stellaraccident (approved):
Thanks. Since this will be an API break, we'll need to do an interlocked update of iree-turbine. I'll land I change there today which accepts this argument as optional so we have an easier compatibility window (so that the people who upgrade iree next aren't in a break the world situation).

PR #3689, opened 2024-09-05
Decompose aten.fmod into aten.mul,sub,div etc.

srinathava:
As titled, create a new decomposition for `aten.fmod.Tensor` to `aten.div`, `aten.trunc`, `aten.mul` and `aten.sub`. Note that we only use `aten.trunc` for floating point operations. This further gets decomposed to `aten.where` etc. by other existing decompositions. 

This decomposition now makes TOSA pass for a simple model with `aten.fmod` while it makes `stablehlo` fail. For now, we disallow this decomposition for `stablehlo`


sjain-stanford (approved):
LGTM

srinathava:
> @srinathava Since this commit adds the decomposition for the `AtenFmodTensorOp`, the existing lowering for the same should be removed from here:
> 
> https://github.com/llvm/torch-mlir/blob/04740824ae9d6e0efe95f940819dba7e6c612a52/lib/Conversion/TorchToLinalg/Uncategorized.cpp#L1285-L1307

@vivekkhandelwal1, thanks for pointing that out. I'll send out a follow-up PR for the cleanup shortly.

PR #3654, opened 2024-08-20
Disable TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS

AmosLewis:
Related discussion: https://discourse.llvm.org/t/drastically-reducing-documented-scope-of-project/80484

To fix the no module named 'torch_mlir._mlir_libs._jit_ir_importer' bug when test pytorch related models like `pytorch/models/mit-b0`

```
python ./run.py  --tolerance 0.001 0.001 --cachedir /proj/gdba/shark/cache --ireebuild ../../iree-build -f pytorch -g models --mode onnx --report --tests  pytorch/models/mit-b0 
Starting e2eshark tests. Using 4 processes
Cache Directory: /proj/gdba/shark/cache
Tolerance for comparing floating point (atol, rtol) = (0.001, 0.001)
Note: No Torch MLIR build provided using --torchmlirbuild. iree-import-onnx will be used to convert onnx to torch onnx mlir
IREE build: /proj/gdba/shark/chi/src/iree-build
Test run directory: /proj/gdba/shark/chi/src/SHARK-TestSuite/e2eshark/test-run
Since --tests or --testsfile was specified, --groups ignored
Framework:pytorch mode=onnx backend=llvm-cpu runfrom=model-run runupto=inference
Test list: ['pytorch/models/mit-b0']
Test pytorch/models/mit-b0 failed [model-run]
Generated status report /proj/gdba/shark/chi/src/SHARK-TestSuite/e2eshark/test-run/statusreport.md
Generated time report /proj/gdba/shark/chi/src/SHARK-TestSuite/e2eshark/test-run/timereport.md
Generated summary report /proj/gdba/shark/chi/src/SHARK-TestSuite/e2eshark/test-run/summaryreport.md
```
If I use :
```
pip install \                                                                     
            --find-links https://github.com/llvm/torch-mlir-release/releases/expanded_assets/dev-wheels \
            --upgrade \
            torch-mlir
```
pip list:
```
torch-mlir         20240820.189
```
Error in model-run.log
```
python runmodel.py  --torchmlirimport fximport --todtype default --mode onnx --outfileprefix mit-b0 1> model-run.log 2>&1
...
Traceback (most recent call last):
  File "/proj/gdba/shark/chi/src/SHARK-TestSuite/e2eshark/test-run/pytorch/models/mit-b0/runmodel.py", line 78, in <module>
    from torch_mlir import torchscript
[...]

PR #3693, opened 2024-09-09
Disable TORCH_MLIR_ENABLE_JIT_IR_IMPORTER by default

AmosLewis:
Temporary fix for Disable TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS #3654

Disable TORCH_MLIR_ENABLE_JIT_IR_IMPORTER by default
- Only enable it in CI for update_abstract_interp_lib.sh and update_torch_ods.sh usage.

TODO:
- Deeper doc change  https://github.com/llvm/torch-mlir/issues/3696
- Pure python verion of torch_ods_gen.py and abstract_interp_lib_gen.py https://github.com/llvm/torch-mlir/issues/3695

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,8 +48,14 @@ option(TORCH_MLIR_OUT_OF_TREE_BUILD "Specifies an out of tree build" OFF)
 
 # PyTorch native extension gate. If OFF, then no features which depend on
 # native extensions will be built.
+# TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS will be disabled by default shortly,
+# once some reorganizations of the testing codebase makes that possible.
 option(TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS "Enables PyTorch native extension features" ON)
-cmake_dependent_option(TORCH_MLIR_ENABLE_JIT_IR_IMPORTER "Enables JIT IR Importer" ON TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS OFF)
+# NOTE: The JIT_IR_IMPORTER paths have become unsupportable due to age and lack of maintainers.
+# Turning this off disables the old TorchScript path, leaving FX based import as the current supported option.
+# The option will be retained for a time, and if a maintainer is interested in setting up testing for it,
+# please reach out on the list and speak up for it. It will only be enabled in CI for test usage.
+cmake_dependent_option(TORCH_MLIR_ENABLE_JIT_IR_IMPORTER "Enables JIT IR Importer" OFF TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS OFF)
 cmake_dependent_option(TORCH_MLIR_ENABLE_LTC "Enables LTC backend" OFF TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS OFF)
 
 option(TORCH_MLIR_ENABLE_ONNX_C_IMPORTER "Enables the ONNX C importer" OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,15 +47,15 @@ endif()
 option(TORCH_MLIR_OUT_OF_TREE_BUILD "Specifies an out of tree build" OFF)
 
 # PyTorch native extension gate. If OFF, then no features which depend on
-# native extensions will be built.
-# TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS will be disabled by default shortly,
-# once some reorganizations of the testing codebase makes that possible.
-option(TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS "Enables PyTorch native extension features" ON)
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -48,8 +48,14 @@ option(TORCH_MLIR_OUT_OF_TREE_BUILD "Specifies an out of tree build" OFF)
 
 # PyTorch native extension gate. If OFF, then no features which depend on
 # native extensions will be built.
+# TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS will be disabled by default shortly,
+# once some reorganizations of the testing codebase makes that possible.
 option(TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS "Enables PyTorch native extension features" ON)
```
stellaraccident, line 53: I assume that we still can't turn TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS OFF, right? If not, ok to land this as is and work towards that point. But if we can turn this off and still function, let's do so: it makes torch-mlir PyTorch version invariant.

docs/add_ops.md
```diff
@@ -38,7 +38,8 @@ PS: IREE is pronounced Eerie, and hence the ghost icon.
 
 ### How to TorchToLinalg
 
-You will need to do 4 things:
+You will need to do 5 things:
+- make sure -DTORCH_MLIR_ENABLE_JIT_IR_IMPORTER=ON is added during build. This is to enable the python file used in `build_tools/update_torch_ods.sh` and `build_tools/update_abstract_interp_lib.sh`
```
stellaraccident, line 42: Can you file an issue and reference it here for having a pure Python version of these?

stellaraccident (approved):
We also need to revise some docs:

https://github.com/llvm/torch-mlir?tab=readme-ov-file#all-the-roads-from-pytorch-to-torch-mlir-dialect : The graphic can either be updated or removed. We should reference FX and ONNX as the entry points. MHLO -> StableHLO

Remove the Demos section (https://github.com/llvm/torch-mlir?tab=readme-ov-file#demos) as it refers only to demos that are now disabled by default. In its place we can do something like:

```
## Using torch-mlir

Torch-MLIR is primarily a project that is integrated into compilers to bridge them to PyTorch and ONNX. If contemplating a new integration, it may be helpful to refer to existing downstreams:

* IREE (link)
* Blade (link)
* Reference CPU runner (local docs)

While most of the project is exercised via testing paths, there are some ways that an end user can directly use the APIs without further integration:

* (see example of FX import)
* (see example of ONNX import)
* etc
```

While editing this, we should also drop the Meetings section as I don't think those are running anymore.

There are some deeper doc changes that would be useful to scrub but we can do those once fixing the front door README.

Fine to make the doc changes in a separate PR, but let's not submit this patch without at least a link to the PR with pending doc changes.

docs/add_ops.md
AmosLewis: https://github.com/llvm/torch-mlir/issues/3695

CMakeLists.txt
AmosLewis: TORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS is ok to turn off. Has turned it off.

stellaraccident (approved):
Updates lgtm

qingyunqu:
I have two points of concern on this PR:
1. It seems that ci doesn't test fx_importer with tosa or stablehlo backend.
2. `GeneratedTorchOps.td` was generated depend on jit_ir_importer now. What's the future plan?
@AmosLewis @stellaraccident

qingyunqu:
Sorry, I see that https://github.com/llvm/torch-mlir/pull/3698 add test on stablehlo backend.

stellaraccident:
The stablehlo test diffs were an omission, which I think now is resolved. I don't have anyone working with me who uses the TOSA side and need contributors there to configure things properly. 

Regarding the ODS issues, Chi is working on a pure python solution for this. The original idea was to do that first but the situation with the c++ extension failing for users (and breaking the whole project) was becoming critical: we had a comparative spike of issue reports over the last week and my team was not able to identify any fixes. In addition, these code paths were implicated in blocking a recent llvm upgrade, consuming multiple engineers on a tangent for a week (https://github.com/llvm/llvm-project/pull/105656#issuecomment-2342171491).

We just don't have any contributors who can maintain this code anymore (and it is the single flakiest thing in the project), and we've been saying that it is deprecated since the beginning of the year/the day will come we can't keep it running. My team is at that point, and we can't spend any more time keeping investing in it. We will continue with a fix to make the ODS generation pure python.

I see you've posted on the discourse thread. Let's discuss options there.

sjarus:
Sorry for the late comment @stellaraccident - I didnt see this as I wasn't tagged. Arm can step in around TOSA - we are actively working on it and test all targets locally, with our priority being the fx_importer_tosa one that has a significantly better pass rate and rising. 

PyTorch 1.0 JIT import isn't really priority, and ideally we'd focus on fx_importer_tosa and onnx_tosa . @mgehre-amd was a user of the PT1 TOSA import path - has that changed ?

PR #3698, opened 2024-09-10
[ci] enable fx_importer2stablehlo ci test

penguin-wwy:


penguin-wwy:
The stablehlo backend is used in many downstream projects, so the corresponding CI tests should be restored immediately. @stellaraccident

stellaraccident (approved):
Thanks. I wasn't aware there were still a bunch of these to move. Definitely not intentional.

PR #3701, opened 2024-09-11
[TorchToLinalg] Merge indices and pooling computation into one linalg generic op

lingzhiz1998:
As discussed in https://github.com/llvm/torch-mlir/pull/3652, we should use one linalg generic op to represent maxpool with indices op.


test/Conversion/TorchToLinalg/pooling.mlir
```diff
@@ -90,8 +90,13 @@ func.func @forward_max_pool3d(%arg0: !torch.vtensor<[?,?,?,?,?],f32>) -> !torch.
...
   // CHECK-NEXT:  ^bb0(%[[CURRENT_VALUE:.*]]: f32, %[[KERNEL:.*]]: f32, %[[ACC_OUT:.*]]: f32):
-  // CHECK-NEXT:    %[[MAXF:.*]] = arith.maximumf %[[CURRENT_VALUE:.*]], %[[ACC_OUT:.*]] : f32
-  // CHECK-NEXT:    linalg.yield %[[MAXF:.*]] : f32
+  // CHECK:         %[[MAX_PRED:.*]] = arith.cmpf ugt, %[[CURRENT_VALUE]], %[[ACC_OUT]] : f32
+  // CHECK:         %[[MAXF:.*]] = scf.if %[[MAX_PRED]] -> (f32) {
+  // CHECK:           scf.yield %[[CURRENT_VALUE]] : f32
+  // CHECK:         } else {
+  // CHECK:           scf.yield %[[ACC_OUT]] : f32
+  // CHECK:         }
+  // CHECK-NEXT:    linalg.yield %[[MAXF]] : f32
```
vivekkhandelwal1, line 99: @lingzhiz1998 I haven't gone through the code in this PR, but this IR getting generated instead of the earlier one, is not good for the downstream pipeline, and I don't see a point in doing that.

lingzhiz1998: This PR aims to change the IR of max_pool _with_indices op to eliminate extract op. Generating different IRs for max_pool op is not necessary. The reason of why I did this is making the lowering logic of these ops as identical as possible. If you worry about this, I can keep the IR of max_pool op as same as previous.

lingzhiz1998:
@vivekkhandelwal1 any comments? Thanks!

vivekkhandelwal1:
> @vivekkhandelwal1 any comments? Thanks!

Let me review it thoroughly today. I will respond after that.

vivekkhandelwal1 (commented):
Hi @zjgarvey, can you please review this PR when you get time? Since for me it's really hard to understand what are we trying to achieve through these changes. I think you asked for this (probably?) here https://github.com/llvm/torch-mlir/pull/3652#pullrequestreview-2257428928. I'm slightly concerned about introducing `scf.if` op and whether that's actually required or not.

Excluding this, there are only 2 Torch->Linalg lowering which contains op from the `scf` dialect, one of which is multinomial which Mahesh was saying that should not have been lowered that way.

PR #3706, opened 2024-09-12
[Torch] add AtenAsStridedOp in torch dialect

yyp0:


qingyunqu (approved):

arnab-oss:
@yyp0 do you plan to add any lowering for this in near future or does one already exist?

qingyunqu:
> @yyp0 do you plan to add any lowering for this in near future or does one already exist? CC @qingyunqu

Sorry, `aten.as_strided` op doesn't have `Value Sematics`. It's hard to lowering it.  
Maybe we will trying to lower it in a very limited situation, but we wouldn't contribute the hacked lowering to `torch-mlir`.

arnab-oss:
Okay. Thanks for the quick reply:)

PR #3707, opened 2024-09-12
Delete unnecessary linalg conversion for aten.fmod

srinathava:
Follow up cleanup for [this PR](https://github.com/llvm/torch-mlir/pull/3689), which introduced a decomposition for `aten.fmod.Tensor`. This means that the lowering for this operator in linalg is no longer needed.

Thanks to @vivekkhandelwal1 for pointing this out.


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -1282,29 +1282,6 @@ static Value createLinalgPayloadCalculationForElementwiseOp(
...
-      result = b.create<arith::SubFOp>(loc, self, n_y);
-    } else if (isa<mlir::IntegerType>(newResultType)) {
-      Value n = b.create<arith::DivSIOp>(loc, self, other);
-      Value n_y = b.create<arith::MulIOp>(loc, n, other);
-      result = b.create<arith::SubIOp>(loc, self, n_y);
-    } else {
-      fmod.emitError("Unsupported type encountered for AtenFmodTensorOp.");
-    }
-    return result;
-  }
```
vivekkhandelwal1, line 1307: You need to remove it from certain other places in the same file. Please search for the op name and remove it.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchToLinalg/Uncategorized.cpp
srinathava: Oops. take two :) Can you take another look?

vivekkhandelwal1 (approved):

PR #3713, opened 2024-09-16
[ONNX] Fix issue with absent value in onnx.ConstantOfShape

zjgarvey:
Previously, if the value was absent, this conversion was creating a dense resource of value 0 with shape equal to the result shape, then later re-extracting a splat value. This only works if the shape is statically known, and even when the shape is known, this is completely unnecessary since the value's shape should be `[1]` and not the result shape. 

This patch simply sets the `splatvalue` to a `torch.constant.float 0.0` when the onnx op's `value` attr is absent, and adds `nullptr` checks to the subsequent conditionals to avoid them in the case where an `attr` is not given.

Addresses <https://github.com/nod-ai/SHARK-Turbine/issues/831>.


jinchen62 (approved):
LGTM!

PR #3714, opened 2024-09-16
[torch] Fix attention on linalg for dynamic shapes

rsuderman:
Current version does not work for a mixture of dynamic and static shaped batch dimensions. Rework to grab the correct dynamic shapes.


zjgarvey (approved):
Makes sense to me. Are we testing this e2e? If so, presumably the pass rate is unchanged, so should we add a test for the mixed static/dynamic query/key?

Where did this issue come up?

PR #3730, opened 2024-09-24
[torch] Fix unsqueezed output shape in canonicalization of AtenUnflattenIntOp

vinayakdsci:
Fixes https://github.com/iree-org/iree/issues/18562.

During canonicalization pass on `AtenUnflattenIntOp`, if the second dim was statically equal to one, we would create an `AtenAddIntOp` to add one to the dimension obtained from `op.getDim()`. This, when passed into `Torch::unsqueezeTensor()`, would make it get interpreted as non-constant, which would lead to MLIR failing an assertion when `UnsqueezeOp` would later get lowered into `ExpandShapeOp`, as the output of the `UnsqueezeOp` would consist of only dynamic dims.

This patch fixes this behavior, by extracting the integer value from the dim if it was constant, and then emitting a `ConstantIntOp` from (dim+1). This creates an output with static shape.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -2189,6 +2189,9 @@ void AtenUnflattenIntOp::getCanonicalizationPatterns(
     if (dim0 != 1 && dim1 != 1)
       return failure();
     Value unflattenDim = op.getDim();
+    int64_t dimAsInt;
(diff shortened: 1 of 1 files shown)


vivekkhandelwal1 (approved):
LGTM

zjgarvey (approved):
Excellent. Thanks, Vinayak.

PR #3732, opened 2024-09-25
[TorchToLinalg]Lower torch.gcd to linalg and scf

bratislavSyrmia:
Add verify() method to check if tensors are of
integer type. Also check if tensors are of same shape, or if the second tensor is a single element tensor.

Add e2e tests. Put them into onnx and stablehlo
xfailed sets.


bratislavSyrmia:
force push: added math::cttz instead of counting trailing zeros manually

projects/pt1/python/torch_mlir_e2e_test/test_suite/elementwise.py
```diff
@@ -6845,3 +6845,52 @@ def forward(self):
...
+    @export
+    @annotate_args([None, [(4, 4), torch.int32, True], [(4, 4), torch.int32, True]])
+    def forward(self, A, B):
+        return torch.gcd(A, B)
+
+
+@register_test_case(module_factory=lambda: GCDModule())
+def GCDModule_I32(module, tu: TestUtils):
+    A = tu.rand(4, 4).to(dtype=torch.int32)
+    B = tu.rand(4, 4).to(dtype=torch.int32)
```
IanWood1, line 6863: I think this only checks elements in the range 0-1 because the defaults to high and low are 0 and 1. I was looking at this before the force push and changing it to [0, 100] or [-100, 100] were causing test failures (but looks like it works now).

bratislavSyrmia: Okay, I have changed the tests to generate values between -100 and 100. 
Also yes, in first version there was an error so it didn't calculate the correct results, it's fixed now.

IanWood1:
> Hi @bratislavSyrmia, I would like you to explore a non-loop path for this lowering, since these kind of lowerings usually causes issues in the downstream pipeline especially the code-generation part.

@vivekkhandelwal1 out of curiosity, do you have an algorithm/solution in mind? The best I could think of was to use `linalg.generic`'s `library_call` attr and define a GCD function but maybe that runs into the same problem.

vivekkhandelwal1:
> > Hi @bratislavSyrmia, I would like you to explore a non-loop path for this lowering, since these kind of lowerings usually causes issues in the downstream pipeline especially the code-generation part.
> 
> @vivekkhandelwal1 out of curiosity, do you have an algorithm/solution in mind? The best I could think of was to use `linalg.generic`'s `library_call` attr and define a GCD function but maybe that runs into the same problem.

Hi @IanWood1, I did not spend time on thinking about it that's why I asked @bratislavSyrmia to explore the possibility of any such solution.

But if there exists a solution based on `linalg.generic` then it would still be a better approach then the current one.

bratislavSyrmia:
> > > Hi @bratislavSyrmia, I would like you to explore a non-loop path for this lowering, since these kind of lowerings usually causes issues in the downstream pipeline especially the code-generation part.
> > 
> > 
> > @vivekkhandelwal1 out of curiosity, do you have an algorithm/solution in mind? The best I could think of was to use `linalg.generic`'s `library_call` attr and define a GCD function but maybe that runs into the same problem.
> 
> Hi @IanWood1, I did not spend time on thinking about it that's why I asked @bratislavSyrmia to explore the possibility of any such solution.
> 
> But if there exists a solution based on `linalg.generic` then it would still be a better approach then the current one.

I have thought about it but I have no idea how I would find the greatest common divisor between two numbers without using loops

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -213,6 +215,83 @@ class ConvertAtenMmOp : public OpConversionPattern<AtenMmOp> {
...
+  using OpConversionPattern::OpConversionPattern;
+
+  LogicalResult
+  matchAndRewrite(torch::Torch::AtenGcdOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto self = adaptor.getSelf();   // tensor A
+    auto other = adaptor.getOther(); // tensor B of the same size
+    auto loc = op.getLoc();
+
+    TensorType resultType =
```
bondhugula, line 230: `auto`

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -213,6 +215,83 @@ class ConvertAtenMmOp : public OpConversionPattern<AtenMmOp> {
...
+      gcdResult =
+          b.create<mlir::arith::ShLIOp>(loc, gcdResult, smalerZerosCount);
+
+      b.create<linalg::YieldOp>(loc, gcdResult);
+    };
+
+    other = torch_to_linalg::createElementwiseLinalgGeneric(
+        rewriter, loc, ValueRange{self, other},
+        cast<TensorType>(self.getType()).getElementType(), gcdPayloadBody);
+
```
bondhugula, line 288: Missing code comments for all the major blocks and a high-level description on the top for the lowering.

lib/Dialect/Torch/Transforms/AbstractInterpLibrary.cpp
```diff
@@ -6639,6 +6639,72 @@ StringRef mlir::torch::Torch::getAbstractInterpLibrary() {
...
+"    }\n"
+"    torch.prim.If %3 -> () {\n"
+"      torch.prim.If.yield\n"
+"    } else {\n"
+"      torch.prim.RaiseException %str, %none : !torch.str, !torch.none\n"
+"      torch.prim.If.yield\n"
+"    }\n"
+"    return %0#1 : !torch.int\n"
+"  }\n"
+"  func.func @__torch__.torch_mlir.jit_ir_importer.build_tools.library_generator.is_integer_dtype(%arg0: !torch.int) -> !torch.bool {\n"
```
bondhugula, line 6693: Why is this part moved?

bondhugula:
> > > > Hi @bratislavSyrmia, I would like you to explore a non-loop path for this lowering, since these kind of lowerings usually causes issues in the downstream pipeline especially the code-generation part.
> > > 
> > > 
> > > @vivekkhandelwal1 out of curiosity, do you have an algorithm/solution in mind? The best I could think of was to use `linalg.generic`'s `library_call` attr and define a GCD function but maybe that runs into the same problem.
> > 
> > 
> > Hi @IanWood1, I did not spend time on thinking about it that's why I asked @bratislavSyrmia to explore the possibility of any such solution.
> > But if there exists a solution based on `linalg.generic` then it would still be a better approach then the current one.
> 
> I have thought about it but I have no idea how I would find the greatest common divisor between two numbers without using loops

This PR is acceptable and reviewable as is and I don't think it should be blocked because of a downstream's inability to deal with it. The lowering output is perfectly valid IR and it's contributing an otherwise missing lowering. Using loops itself, with a `while` loop (Euclid's GCD), it'd take `log n` steps while one could use an alternative approach with a countable `for` loop that would take `O(n)` steps. Downstream users could potentially add a lowering that uses a "GCD" intrinsic call to link with a library if really needed.

lib/Dialect/Torch/Transforms/AbstractInterpLibrary.cpp
IanWood1: The file is auto-generated so I think it's a quirk of `update_abstract_interp_lib.sh`. Its also checked by ci here: https://github.com/llvm/torch-mlir/blob/99115dcdc8cff8ce07bd027a12b001ddd7e957f3/.github/workflows/ci.yml#L77

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -213,6 +215,83 @@ class ConvertAtenMmOp : public OpConversionPattern<AtenMmOp> {
...
+          loc, TypeRange{shiftedA.getType(), shiftedB.getType()},
+          ValueRange{shiftedA, shiftedB}, findGcdConditionBlock,
+          findGcdBodyBlock);
+
+      Value gcdResult = findGcdWhileOp.getResult(1);
+      gcdResult =
+          b.create<mlir::arith::ShLIOp>(loc, gcdResult, smalerZerosCount);
+
+      b.create<linalg::YieldOp>(loc, gcdResult);
+    };
```
bondhugula, line 283: You don't need `mlir::` anywhere here since there is a namespace `using` for it at the top.

PR #3736, opened 2024-09-25
Try pytorch workflow

Hanumanth04:
We would consider `tm_tensor` as the legal dialect after the `torch-to-tosaLinalg` conversion. However, any dialect other than the MLIR core dialects should be illegal after `torch-to-tosaLinalg` pass. This change updates `VerifyTosaLinalgBackendContract to remove the `tm_tensor` dialect from the legal dialect list.

For the `hf_Roberta_base` model, without this change, `torch-to-tosaLinAlg` would succeed; now, it will fail with the error below, as expected. I am looking into enabling the `tm-tensor-to-loops` pass in a separate task. This pass has some issues when operands to the operations are `tensor` type.

`
%230:2 = "tm_tensor.scan"(%225, %227, %229) <{dimension = 1 : i64, inclusive = true, operandSegmentSizes = array<i32: 1, 2>}> ({
^bb0(%arg2: i64, %arg3: i64):
  %1885 = "arith.addi"(%arg2, %arg3) <{overflowFlags = #arith.overflow<none>}> : (i64, i64) -> i64
  "tm_tensor.yield"(%1885) : (i64) -> ()
}) : (tensor<1x512xi64>, tensor<1x512xi64>, tensor<1xi64>) -> (tensor<1x512xi64>, tensor<1xi64>)
/local-ssd/hhanuman/25awork/Pytorch/modelConversion/pytorch-scripts/codegen/hf_Roberta_base.torch.mlir:1:1: error: Module does not conform to the linalg-on-tensors backend contract. See dialect conversion legality information above.
module {
`


stellaraccident:
Since just a draft, I didn't give a detailed review, and what is being done on the tosa side is outside of what I understand the strategy on. But I just wanted to signal that I don't think the torch-to-linalg changes are mergeable (and I don't understand why tosa changes require large scale changes to those pipelines).

It is pretty crucial that the tm_tensor ops not be automatically decomposed into loopy, generic operations by default. It is fine to have pipelines that do that for consumers who need it, but it can't happen by default. This PR has a lot going on in it so it wasn't completely clear to me what the scope/intent there was. Just signalling because if making big cross pipeline changes like this, the approach may need to be different, and the PR should be split in some fashion.

Hanumanth04:
Hi Stella,
Please disregard this pull request; I created it by accident. The original intention was to create a pull request from our own fork, which we are using to try out a few things first. We are still working on some logistics before we upstream some of these changes. And, yes, we will be splitting this PR so that it is manageable to review. I will close this draft pull request to avoid any confusion.

stellaraccident:
No worries - that's what draft's are for. It just looked like a substantial body of work and I noticed that there may be an approach issue. Always better to know that early vs later so I spoke up.

PR #3737, opened 2024-09-25
AtenCumprodOp

renxida:


zjgarvey (approved):
Personally, I think it would be better choice to use a templated conversion class for all the cum ops (there are 4 in total- cummin and cummax NYI), but I know there is some disagreement about code duplication being good/bad. 

I'll leave it up to you. If you want to keep it as-is, then maybe put in a TODO comment about refactoring + add support for cummin/cummax. We could have a refactor + support project added to turbine camp as a starter project.

renxida:
Agreed. This is a super hacky copy+paste solution so I get to test my Nonzero impl.

This should totally be a todo.

PR #3751, opened 2024-10-01
[torch.bind_symbolic_shape] Fix verifier for shapeSymbol detection

meshtag:
The op can be valid with no attached shape symbols if they are not required by the corresponding affine map. Fix the verifier to consider number of arguments for both.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -5405,8 +5405,11 @@ void BindSymbolicShapeOp::print(OpAsmPrinter &p) {
 }
 
 LogicalResult BindSymbolicShapeOp::verify() {
-  if (getShapeSymbols().empty())
-    return emitOpError() << "requires non-empty shapeSymbols";
+  if (getShapeSymbols().size() !=
+      getShapeExpressions().getValue().getNumSymbols())
+    return emitOpError()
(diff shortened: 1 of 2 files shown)


stellaraccident (approved):
Good catch. Thanks. Do you happen to have a minimal example that produces this behavior. It's not obvious to me and we may want to add a test.

meshtag:
Thanks for the review.

> Do you happen to have a minimal example that produces this behavior. It's not obvious to me and we may want to add a test.

I got this error in graph 2 (as partitioned by dynamo) of llama3.2 1B instruct model (model spec [here](https://huggingface.co/meta-llama/Llama-3.2-1B-Instruct)). At the moment, I do not have a python spec which would directly lead to this (it is not obvious to me as well, tbh).

PS: I would require your help to land this (if we want to merge in this state). Thanks.

stellaraccident:
> Thanks for the review.
> 
> > Do you happen to have a minimal example that produces this behavior. It's not obvious to me and we may want to add a test.
> 
> I got this error in graph 2 (as partitioned by dynamo) of nano_llava model (model spec [here](https://huggingface.co/qnguyen3/nanoLLaVA)). At the moment, I do not have a python spec which would directly lead to this (it is not obvious to me as well, tbh).
> 
> PS: I would require your help to land this (if we want to merge in this state). Thanks.

Thanks. It's a reasonable thing to fix -- I just try to keep a mental tally on where things come up.

PR #3752, opened 2024-10-01
[Release] Fix binary name for downstream compatibility

sjain-stanford:
As of Sep 14, the torch-mlir binary [wheels](https://github.com/llvm/torch-mlir-release/releases/tag/dev-wheels) got renamed to `torch-mlir-core` from `torch-mlir`: image 

This was an unintended side-effect of the recent change of `TORCH_MLIR_ENABLE_ONLY_MLIR_PYTHON_BINDINGS=True` (https://github.com/llvm/torch-mlir/pull/3711) which skips setting `NAME = "torch-mlir"` in [setup.py](https://github.com/llvm/torch-mlir/blob/main/setup.py#L226-L232).

To avoid having multiple downstreams fix their pip deps, this change allows using the same `torch-mlir` name for binaries, and reserves a separate `torch-mlir-ext` name for the (less popular) binaries with extensions enabled.

### DIFF
--- a/build_tools/python_deploy/build_linux_packages.sh
+++ b/build_tools/python_deploy/build_linux_packages.sh
@@ -50,7 +50,7 @@ TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp38-cp38 cp310-cp310 cp311-cp311}"
 # Location to store Release wheels
 TM_OUTPUT_DIR="${TM_OUTPUT_DIR:-${this_dir}/wheelhouse}"
 # What "packages to build"
-TM_PACKAGES="${TM_PACKAGES:-torch-mlir torch-mlir-core}"
+TM_PACKAGES="${TM_PACKAGES:-torch-mlir torch-mlir-ext}"
 # Use pre-built Pytorch
 TM_USE_PYTORCH_BINARY="${TM_USE_PYTORCH_BINARY:-ON}"
 # Skip running tests if you want quick iteration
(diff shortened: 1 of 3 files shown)


saienduri:
Fine by me, as `torch-mlir-core` seems to be the main thing that we support and test right now so makes sense to change to `torch-mlir`. Can you also change all the other occurrences of these package names in the repo so we have uniform naming convention? Let's also wait for Stella's opinion (she's probably the one that named them initially lol)

stellaraccident (approved):

sjain-stanford:
> Can you also change all the other occurrences of these package names in the repo to torch-mlir and torch-mlir-ext so we have uniform naming convention?

@saienduri I have done this now, but just for the record, until now we were really calling the release builds with PT extensions (`TM_PACKAGES=torch-mlir`) [here](https://github.com/llvm/torch-mlir-release/blob/main/.github/workflows/buildReleaseAndPublish.yml#L48) - which ran [build_torch_mlir](https://sourcegraph.com/github.com/llvm/torch-mlir/-/blob/build_tools/python_deploy/build_linux_packages.sh?L161).


After this change, we will actually be building the "core" (non-PT-extension) version. It is unlikely this breaks anything downstream for y'all but if it does, we could simply switch this line in the GHA workflow to go back: 
```
- TM_PYTHON_VERSIONS=cp310-cp310 TM_PACKAGES=torch-mlir ./build_tools/python_deploy/build_linux_packages.sh
+ TM_PYTHON_VERSIONS=cp310-cp310 TM_PACKAGES=torch-mlir-ext ./build_tools/python_deploy/build_linux_packages.sh
```

PR #3758, opened 2024-10-03
[TorchToLinalg] Fix AtenReflectionPad2dOp lowering to not assert when dimensions unknown

giacs-epic:
The asserts make the compiler crash (with a misleading error) also when vertical or horizontal dimensions are unknown (which is a valid state at this stage).

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -404,13 +404,13 @@ public:
     Value hDimSize = inputShape[hDim];
     Value vDimSize = inputShape[vDim];
 
-    if (inputType.getShape()[hDim] != kUnknownSize) {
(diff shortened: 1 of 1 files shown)


zjgarvey (changes requested):
Makes sense! Does the conversion even work if the input shape is dynamic? Can we add a dynamic test (either a lit test or preferably an e2e test) to make sure that path actually works?

giacs-epic:
@zjgarvey I just added a dynamic test, it seems to work fine! I also took the chance to add dynamic asserts.

giacs-epic:
> @zjgarvey I just added a dynamic test, it seems to work fine! I also took the chance to add dynamic asserts.

I struggle to understand the CI error. It seems like when exporting to onnx it interprets the input sizes as `[1, 1, 1]` instead of `[1, -1, -1]`. A PyTorch bug?

PR #3770, opened 2024-10-07
[MLIR][TORCH] Add support for 1-d group convolution

vivekkhandelwal1:
This commit adds the support for the 1-d depthwise convolution as a special case of 1-d group convolution.


lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -1236,21 +1232,45 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+                         loc, outputTensor.getType(),
+                         ValueRange{paddedInput, collapsedWeight}, outputTensor,
+                         stridesAttr, dilationAttr)
+                     .getResult(0);
+          break;
+        default:
+          return rewriter.notifyMatchFailure(
+              op, "unimplemented: only 1D and 2D depthwise convolution "
+                  "supported for special case of group convolution");
+        };
```
kumardeepakamd, line 1267: return success()" here too?

kumardeepakamd (approved):
It is covering only the case that can be depthwise 1d conv, and using preexisting function. Good to go.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -1297,6 +1317,10 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
       return success();
     }
 
+    if (numSpatialDims != 2)
+      return rewriter.notifyMatchFailure(
+          op, "unimplemented: only 2D grouped convolution supported");
+
```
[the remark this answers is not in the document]
vivekkhandelwal1: Yes

lib/Conversion/TorchToLinalg/Linear.cpp
vivekkhandelwal1: No, there's a common return for both the if and else path.

PR #3771, opened 2024-10-07
[WIP] onnx.LSTM - support bidirectional and layout attr

PhaneeshB:
- Support Bidirectional LSTM (utilising the forward LSTM layer with flipped Inputs and Outputs)
- Support layout 1 
- fixes for alt_e2e_tests lstm tests (1,2,3)


lib/Conversion/TorchOnnxToTorch/OnnxRecurrentLayerOpExpanders.cpp
```diff
@@ -661,8 +661,8 @@ LogicalResult OnnxLstmExpander(OpBinder binder,
   std::string direction;
 
   ValueTensorType yTy, Y_hType, Y_cType;
-  if (binder.tensorResultTypeAtIndex(yTy, 0) ||
-      binder.tensorResultTypeAtIndex(Y_hType, 1) ||
+  if (binder.tensorResultTypeAtIndex(yTy, 0) &&
+      binder.tensorResultTypeAtIndex(Y_hType, 1) &&
       binder.tensorResultTypeAtIndex(Y_cType, 2)) {
```
vinayakdsci, line 666: @PhaneeshB this change appears to be a little confusing. Possibly it's just me being slow, but it'd be great if you could add a tiny explanation for the need of this. And, to be even more explicit (in case you are checking for failure here), I'd just wrap it in a call to `failed()`.

renxida: i think the change is correct & idiomatic per our codebase - - I got it wrong when i originally wrote it.

But i personally would so like to see us use failed(binder.tensorResultTypeAtIndex()) and succeeded(...)  more. it might be longer but i think would make so many things so much easier to read & would have prevented me from making this mistake in the first place if i had the habit.

renxida (approved):

lib/Conversion/TorchOnnxToTorch/OnnxRecurrentLayerOpExpanders.cpp
vinayakdsci: I agree :). And this is part of the reason why I was having difficulty what this was actually doing!

PR #3774, opened 2024-10-08
Added support for Maxpool (Autopad)

sriram-siloai:
Added autopad. and passed 3 tests 

test_maxpool_2d_precomputed_same_upper
test_maxpool_2d_same_lower'
test_maxpool_2d_same_upper

Address : https://github.com/nod-ai/SHARK-ModelDev/issues/843 

2 attributes yet to complete : storage_order, indices output

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1141,12 +1141,6 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         if (dilations.empty())
           dilations.resize(spatial, 1);
 
-        if (padding.size() != static_cast<size_t>(spatial) &&
-            padding.size() != static_cast<size_t>(2 * spatial)) {
-          return rewriter.notifyMatchFailure(
-              binder.op, "padding list size does not match the number of axes");
-        }
-
         auto inputTensorType = cast<Torch::ValueTensorType>(operand.getType());
 
         // Padding for the beginning and ending along each spatial axis, it can
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1087,9 +1087,6 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         if (binder.customOpNameStringAttr(autoPad, "auto_pad", "NOTSET"))
           return rewriter.notifyMatchFailure(binder.op,
                                              "auto_pad bind failure");
-        // if (autoPad != "NOTSET")
-        //   return rewriter.notifyMatchFailure(
-        //       binder.op, "unsupported conversion: auto_pad != NOTSET");
 
         Torch::ValueTensorType resultTypeOut;
         Value operand;
@@ -1153,11 +1150,7 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         // x2_begin…x1_end, x2_end,…], where xi_begin the number of pixels added
         // at the beginning of axis i and xi_end, the number of pixels added at
         // the end of axis i.
-        if (autoPad == "NOTSET") {
-          // keep previous value
-        } else if (autoPad == "VALID") {
-          // keep previous value
-        } else {
+        if (autoPad != "NOTSET" && autoPad != "VALID") {
           const bool isSameLower = autoPad == "SAME_LOWER";
           const unsigned spatial = rank - 2;
           ArrayRef<int64_t> inputShape = inputTensorType.getSizes();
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1087,9 +1087,9 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         if (binder.customOpNameStringAttr(autoPad, "auto_pad", "NOTSET"))
           return rewriter.notifyMatchFailure(binder.op,
                                              "auto_pad bind failure");
-        if (autoPad != "NOTSET")
-          return rewriter.notifyMatchFailure(
-              binder.op, "unsupported conversion: auto_pad != NOTSET");
+        // if (autoPad != "NOTSET")
+        //   return rewriter.notifyMatchFailure(
+        //       binder.op, "unsupported conversion: auto_pad != NOTSET");
```
vivekkhandelwal1, line 1092: Please remove this code.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1136,13 +1136,51 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        // number of pixels added to the beginning and end part of the
+        // corresponding axis. pads format should be as follow [x1_begin,
+        // x2_begin…x1_end, x2_end,…], where xi_begin the number of pixels added
+        // at the beginning of axis i and xi_end, the number of pixels added at
+        // the end of axis i.
+        if (autoPad == "NOTSET") {
+          // keep previous value
+        } else if (autoPad == "VALID") {
+          // keep previous value
+        } else {
```
vivekkhandelwal1, line 1160: Instead, you can do:
```suggestion
        if (autoPad != "NOTSET" && autoPad != "VALID") {
        ......
        }
```

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1136,13 +1136,51 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+          const bool isSameLower = autoPad == "SAME_LOWER";
+          const unsigned spatial = rank - 2;
```
vivekkhandelwal1, line 1162: This value is already present in the code above as `spatial`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1136,13 +1136,51 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            totalPad = totalPad >= 0 ? totalPad : 0;
+            padding[dimIdx] =
+                isSameLower ? ((totalPad + 1) / 2) : (totalPad / 2);
+            padding[spatial + dimIdx] = totalPad - padding[dimIdx];
+          }
+        }
+        if (padding.size() != spatial && padding.size() != 2 * spatial) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "padding list size does not match the number of axes");
+        }
```
vivekkhandelwal1, line 1182: Shouldn't this condition be placed somewhere above in the code? It does not make sense to me to keep it here.

vivekkhandelwal1 (changes requested):

sriram-siloai:
@vivekkhandelwal1 : I have made the necessary changes and pushed in the latest commit. Could you review it again. Thanks.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
vivekkhandelwal1: This comment is still unresolved.

vivekkhandelwal1: Why do you need to define this value again. Just used the above one which is already defined here: https://github.com/llvm/torch-mlir/blob/dc7a1ff7d9134758128a637dca976f72c2366e59/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp#L1114

vivekkhandelwal1 (changes requested):
Except that one comment, I would also like you to add a lit test here https://github.com/llvm/torch-mlir/blob/main/test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir for the given feature to be tested. Rest all, looks good to me.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1136,13 +1133,48 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
         if (strides.empty())
           strides.resize(spatial, 1);
         if (dilations.empty())
           dilations.resize(spatial, 1);
 
+        if (padding.size() != static_cast<size_t>(spatial) &&
+            padding.size() != static_cast<size_t>(2 * spatial)) {
+          return rewriter.notifyMatchFailure(
+              binder.op, "padding list size does not match the number of axes");
+        }
```
vivekkhandelwal1, line 1148: Hi @sriram-siloai, the overall patch looks good to me. I think we don't need this condition since, the `padding` tensor will satisfy this constraint no matter what the code flow is. I think these changes were adopted by the `Conv` op lowering and the case there's different, also to me it seemed a conservative check not a necessary check. Please remove this code.

vivekkhandelwal1 (approved):
Hi @sriram-siloai, thanks for the patch. This patch can be merged once you address that last comment.

I hope the padding is also numerically correct. If you find some time, then please try to add an E2E test here: https://github.com/nod-ai/SHARK-TestSuite/ to test the robustness of your patch.

PR #3779, opened 2024-10-09
adds a few common patterns to scalarize shapes pass

zjgarvey:
This patch adds two things:

1. support for folding scalar patterns like [1]---squeeze--->[] ---unsqueeze--->[1]. 
2. a canonicalizer for aten.view that applies when we can statically or dynamically (through the scalarized view shapes) infer that it is a flatten or unflatten op in the last dim. 

I'm not sure if this is the right place to be adding such a view canonicalizer. Catastrophically, there is a decomposition from flatten and unflatten into aten.view. Until this gets deleted (and it definitely should be deleted), I felt like this would be an appropriate temporary home. We run scalarize shapes after lowering to the backend contract (i.e., decomposing), and scalarize shapes is required to be able to infer dynamic dims coming from size int ops.

### DIFF
--- a/lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
+++ b/lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
@@ -630,7 +630,8 @@ public:
       // flatten
       Value start =
           rewriter.create<Torch::ConstantIntOp>(op.getLoc(), endMatchingDim);
-      Value end = rewriter.create<Torch::ConstantIntOp>(op.getLoc(), inRank);
+      Value end =
+          rewriter.create<Torch::ConstantIntOp>(op.getLoc(), inRank - 1);
       rewriter.replaceOpWithNewOp<AtenFlattenUsingIntsOp>(
           op, resultTy, op.getSelf(), start, end);
       return success();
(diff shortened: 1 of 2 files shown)


lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -530,11 +530,139 @@ class FoldAtenUnsqueezePattern : public OpRewritePattern<AtenUnsqueezeOp> {
                                                      none, none, none, none);
       return success();
     }
+    auto squeezeOp = op.getSelf().getDefiningOp<AtenSqueezeDimOp>();
+    if (squeezeOp && resultTy.getSizes().size() == 1) {
```
rsuderman, line 534: Shouldn't we include a check that the input / output rank are the same as well?

rsuderman (approved):

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
zjgarvey: If the result rank of unsqueeze is 1, then it can only have shape [1]. Therefore, if the input came from a squeeze.dim op, it must have been a squeeze from [1] to [].

PR #3784, opened 2024-10-11
support `aten._trilinear` and improve `einsum` decomposition

stbaione:
# Tracking
[Issue](https://github.com/nod-ai/SHARK-ModelDev/issues/848)
[TorchToLinalg Op Support](https://github.com/nod-ai/SHARK-ModelDev/issues/347)

# Description

 Aten_TrilinearOp is an implementation of a "trilinear einstein sum". Essentially, just an einsum across 3 tensors.

There are a few inputs:
## Tensor Inputs
- i1, i2, i3 - The three input tensors for the _trilinear op.
## Expands 
These inputs allow you to unsqueeze an input tensor at the specified dims as a pre-processing step to make the shapes compatible for the rest of the op:
- expand1: List[int], expand2: List[int], expand3: List[int]

## sumdim
- sumdim: List[int] - After applying element wise multiplication, the values in sumdim denote where to collapse a dimension by summing over it

## unroll_dim
- unroll_dim: int - In the PyTorch implementation, this specifies a dimension where you could slice the input tensors, multiply and sum them, then concatenate the results in an output tensor. This complicates the implementation significantly, but doesn't change the result, so I opted against it. Along with that, a previously accepted path for solving this involved reusing the AtenEinsumOp, which also would also ignore this input.

# Solution

After trying a bunch of more complicated approaches for it, this op actually ended up being quite simple: [See _trilinear](https://dev-discuss.pytorch.org/t/defining-the-core-aten-opset/1464)

`_trilinear = (i1.unsqueeze(expand1) * i2.unsqueeze(expand2) * i3.unsqueeze(expand3)).sum(sumdim)`

Wish I saw this earlier, but watcha gonna do: :upside_down_face:

## Not Reusing AtenEinsumOp
Frankly, I found multiple cases where valid inputs would have numerical mismatches for EinsumOp, even when running tests against EinsumOp directly. I think it has something to do with the singleton dimensions. Will need to look into this further, but once I realized the simplified approach, it appeared to be more reliable and much simpler.

Either way (credit to @zjgarvey), there are improvements to the einsum op here. When I was originally trying to use the op, intermediate tensors were being flattened properly, but then its 0th dimension was being cast from a static dim to a dynamic dim due to integers not folding correctly in the MLIR. Figured it's worth keeping these improvements for future reusers of EinsumOp.

# The zero'd out dim "bug"

For some reason, if you specify a dimension in all `expands`,

```i.e. 
[expand1=[0], expand2=[0], expand3=[0]],
[expand1=[1], expand2=[1], expand3=[1]]
```

The _trilinear op would specify `0` for that dimension in the output shape, unless it was also included in `sumdim`. This goes against the implementation of torch.einsum:

```
>>> a, b, c = [torch.rand(1, 3, 3, 3) for i in range(3)] # Simulate expand at dim=0 for all input tensors
>>> torch.einsum('abcd,abcd,abcd->abcd', a, b, c).shape
torch.Size([1, 3, 3, 3])
```

And is just straight up incorrect mathematically. I considered "replacing" singleton dims with zeroed out dims, but that seemed like carrying over a bug. Instead, I included a test for the case, verified that the singleton dimensions were handled the way that torch.einsum handles it, instead of torch._trilinear, and xfailed it with a note as to why.


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1973,6 +1996,95 @@ class DecomposeAtenEinsumOp : public OpRewritePattern<AtenEinsumOp> {
...
+    }
+
+    // Apply unsqueeze to respective input tensors at the specified dimensions
+    SmallVector<int64_t> sortedExpand1 = expand1;
+    std::sort(sortedExpand1.begin(), sortedExpand1.end());
+    for (auto expand : sortedExpand1) {
+      Value expandDim = rewriter.create<Torch::ConstantIntOp>(
+          loc, rewriter.getI64IntegerAttr(expand));
+      input1 = *unsqueezeTensor(rewriter, op, input1, expandDim);
+    }
```
zjgarvey, line 2047: We should be careful here. 

expand_dims = [2,0] : Does this do...

1. [d0,d1,d2] -> [d0,d1,1,d2] -> [1,d0,d1,1,d2]
2. [d0,d1,d2] -> [1,d0,d1,d2] -> [1,d0,1,d1,d2] (since dim 2 was shifted)
3. Something else?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1973,6 +1996,95 @@ class DecomposeAtenEinsumOp : public OpRewritePattern<AtenEinsumOp> {
...
+    // dimensions being removed.
+    Value result = mul2;
+    SmallVector<int64_t> sortedSumDims = sumDim;
+    std::sort(sortedSumDims.rbegin(), sortedSumDims.rend());
+    for (int64_t dim : sortedSumDims) {
+      Value dimValue = rewriter.create<Torch::ConstantIntOp>(
+          loc, rewriter.getI64IntegerAttr(dim));
+      result =
+          createSumAlongDimension(rewriter, loc, op, result, dimValue, false);
+    }
```
zjgarvey, line 2080: Ah, good. So this **is** actually just elementwise muls and then reduce sums. I suppose this does generalize matmul with the right options and transposes.

I think it is okay to leave like this for now, but our team would be much better off doing literally everything about this op in a single linalg generic, since:

1. We don't need to explicitly unsqueeze anything. Just have the correct output shape and iterator maps.
2. We can do all the mul ops in one generic
3. We can even do the reduce summing in the same generic.

This way we don't have to worry about whether or not IREE can fuse all these ops together. It's already neatly packaged in one op. 


This is a low priority op, so let's leave the approach as a decomposition.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1973,6 +1996,95 @@ class DecomposeAtenEinsumOp : public OpRewritePattern<AtenEinsumOp> {
...
+
+    SmallVector<int64_t> sumDim;
+    if (!matchPattern(op.getSumdim(), m_TorchListOfConstantInts(sumDim))) {
+      return rewriter.notifyMatchFailure(op, "sumDim should be constant");
+    }
+
+    int64_t unrollDim;
+    if (!matchPattern(op.getUnrollDim(), m_TorchConstantInt(&unrollDim))) {
+      return rewriter.notifyMatchFailure(op, "unrollDim should be constant");
+    }
```
zjgarvey, line 2038: This unrollDim is not used in the logic below...

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -1973,6 +1996,95 @@ class DecomposeAtenEinsumOp : public OpRewritePattern<AtenEinsumOp> {
...
+    SmallVector<int64_t> expand3;
+    if (!matchPattern(op.getExpand1(), m_TorchListOfConstantInts(expand1))) {
+      return rewriter.notifyMatchFailure(op, "expand1 should be constant");
+    }
+    if (!matchPattern(op.getExpand2(), m_TorchListOfConstantInts(expand2))) {
+      return rewriter.notifyMatchFailure(op, "expand2 should be constant");
+    }
+    if (!matchPattern(op.getExpand3(), m_TorchListOfConstantInts(expand3))) {
+      return rewriter.notifyMatchFailure(op, "expand3 should be constant");
+    }
```
zjgarvey, line 2028: What about the case where something lies in the triple intersecton of the expand sets? I thought we were going to handle that case.

zjgarvey (changes requested):
Glad you got something working Stephen! The major review points here:

1. We need to fail the conversion in the cases we don't support. It's not sufficient to xfail the tests for unsupported cases, because a downstream user of the tool isn't going to run a big model and say "Oh this random shape is messed up, it must be that one esoteric e2e test for this one op that I saw one day". We need to report a match failure so that the op actually doesn't get converted, so we don't have model support people spending days debugging a silently failing conversion.
2. Related to 1. What does unroll dim do? It needs to be included, or if unrolldim !=0 we also need to report an "unimplemented" match failure. 
3. Not really major, but glad we don't have to use einsum. I think the einsum changes are generally good, but it might be better to move them into a different patch. I'm fine leaving them in here, but the commit messaging will seem odd if anyone wants to trace back the history.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
stbaione: Yeah, that's the case where it inserts the dimension of zero. I was thinking of leaving it as-is since the actual einsum op doesn't zero out that dimension:

```
>>> a, b, c = [torch.rand(1, 3, 3, 3) for i in range(3)] # Simulate expand at dim=0 for all input tensors
>>> torch.einsum('abcd,abcd,abcd->abcd', a, b, c).shape
torch.Size([1, 3, 3, 3])
```

But, I could see how that'd be confusing downstream. Should I go back to using "Empty tensor or match failure" in that case?

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
stbaione: The unrollDim allows slicing along a dimension across all tensors. Then you can do (slice1 * slice2 * slice3).sum(sumdim), and concat the result to the output tensor. It doesn't change the output of the function, and wasn't used in the EinsumOp approach, but its intent is to save space by processing the tensors in batches instead of the entire tensor at once.

I can look into extending the solution to use this

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
stbaione: The latter:

```
>>> a = torch.rand(3, 4, 5)
>>> expands = sorted([2, 0])
>>> for e in expands:
...     a = a.unsqueeze(e)
... 
>>> a.shape
torch.Size([1, 3, 1, 4, 5])
```

It follows the order in PyTorch implementation,, where you can also see where the zeroed out dim comes from:

```
  auto expand1 = at::dim_list_to_bitset(expand1_, total_dim);
  auto expand2 = at::dim_list_to_bitset(expand2_, total_dim);
  auto expand3 = at::dim_list_to_bitset(expand3_, total_dim);
  auto sumdim  = at::dim_list_to_bitset(sumdim_,  total_dim);
  Tensor i1 = i1_;
  Tensor i2 = i2_;
  Tensor i3 = i3_;
  std::vector<c10::SymInt> output_size;
  std::vector<int64_t> sum_dims_12, sum_dims_23;
  int64_t unroll_size = -1;
  // asserts...
  for (const auto i : c10::irange(total_dim)) {
    c10::SymInt s = 0;
    if (expand1[i]) {
      i1 = i1.unsqueeze(i);
    } else  {
      s = i1.sym_size(i);
    }
    if (expand2[i]) {
      i2 = i2.unsqueeze(i);
    } else  {
      s = i2.sym_size(i);
    }
    if (expand3[i]) {
      i3 = i3.unsqueeze(i);
      if (sumdim[i] && (i != unroll_dim))
        sum_dims_12.push_back(i);
    } else  {
      s = i3.sym_size(i);
      if (sumdim[i] && (i != unroll_dim))
        sum_dims_23.push_back(i);
    }
    output_size.push_back(sumdim[i] ? 1 : s);
    if (i == unroll_dim)
      unroll_size = s.guard_int(__FILE__, __LINE__);
  }
  ```

**Comment by stbaione** [comment]
> Glad you got something working Stephen! The major review points here:
> 
> 1. We need to fail the conversion in the cases we don't support. It's not sufficient to xfail the tests for unsupported cases, because a downstream user of the tool isn't going to run a big model and say "Oh this random shape is messed up, it must be that one esoteric e2e test for this one op that I saw one day". We need to report a match failure so that the op actually doesn't get converted, so we don't have model support people spending days debugging a silently failing conversion.
> 2. Related to 1. What does unroll dim do? It needs to be included, or if unrolldim !=0 we also need to report an "unimplemented" match failure.
> 3. Not really major, but glad we don't have to use einsum. I think the einsum changes are generally good, but it might be better to move them into a different patch. I'm fine leaving them in here, but the commit messaging will seem odd if anyone wants to trace back the history.

@zjgarvey 
1. Is this referring to: "What about the case where something lies in the triple intersecton of the expand sets? I thought we were going to handle that case."? If so, I can go ahead and add a match failure.
2. Copying reply from above to make it easier to track responses: 

```text
The unrollDim allows slicing along a dimension across all tensors. Then you can do (slice1 * slice2 * slice3).sum(sumdim), and concat the result to the output tensor. It doesn't change the output of the function, and wasn't used in the EinsumOp approach, but its intent is to save space by processing the tensors in batches instead of the entire tensor at once.

I can look into extending the solution to use this
```

3. I agree, it's way more straightforward doing it this way. Maybe I should edit my PR title to include the changes for einsum op? After merging main, the changes actually seem to have fixed 5 tests that were xfailed in `fx_importer_stablehlo` pipeline.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
zjgarvey: If you genuinely think there is a bug in the pytorch implementation, we could simply do something like 

```cpp
if (sharedExpand) {
   // pytorch issue filed: <link>
  // TODO: remove warning when issue gets resolved. 
   op->EmitWarning("aten::_trilinear implementation in this case is non-functional (returns an empty tensor). We will intentionally deviate from this behavior.");
```

zjgarvey:
>     1. Is this referring to: "What about the case where something lies in the triple intersecton of the expand sets? I thought we were going to handle that case."? If so, I can go ahead and add a match failure. Explained the reasoning for why I left it as-is above, but makes sense to be more explicit about that case as to not cause downstream confusion.
> 

If it is a genuine bug, let's at least file an issue in pytorch and emit a warning.
If it is not a genuine bug, then we need to mimic the pytorch behavior.

>     2. Copying reply from above to make it easier to track responses:
> 
> ```
> The unrollDim allows slicing along a dimension across all tensors. 
> Then you can do (slice1 * slice2 * slice3).sum(sumdim), and concat the result to the output tensor. 
> It doesn't change the output of the function, and wasn't used in the EinsumOp approach, 
> but its intent is to save space by processing the tensors in batches instead of the entire tensors at once.
> 
> I can look into extending the solution to use this
> ```

I don't think we will need to implement this, but no point in reporting a match failure if the unrollDim is non-constant. Just make a comment somewhere in the conversion that the unrollDim does not change the result of the operation, so we do not use it in the conversion.

> 
>     3. I agree, it's way more straightforward doing it this way. Maybe I should edit my PR title to include the changes for einsum op? After merging main, the changes actually seem to have fixed 5 tests that were xfailed in `fx_importer_stablehlo` pipeline.
>

Ah, good that it resolves some failing tests. We should rename the title to something like "support `aten._trilinear` and improve einsum decomposition". No TorchToLinalg flag, since this is a decomposition that affects other backends too.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
stbaione: PyTorch issue filed [here](https://github.com/pytorch/pytorch/issues/138353), and emitWarning added for this case

stbaione:
> > ```
> > 1. Is this referring to: "What about the case where something lies in the triple intersecton of the expand sets? I thought we were going to handle that case."? If so, I can go ahead and add a match failure. Explained the reasoning for why I left it as-is above, but makes sense to be more explicit about that case as to not cause downstream confusion.
> > ```
> 
> If it is a genuine bug, let's at least file an issue in pytorch and emit a warning. If it is not a genuine bug, then we need to mimic the pytorch behavior.

PyTorch bug filed [here](https://github.com/pytorch/pytorch/issues/138353) and `emitWarning` added.

> 
> > ```
> > 2. Copying reply from above to make it easier to track responses:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > The unrollDim allows slicing along a dimension across all tensors. 
> > Then you can do (slice1 * slice2 * slice3).sum(sumdim), and concat the result to the output tensor. 
> > It doesn't change the output of the function, and wasn't used in the EinsumOp approach, 
> > but its intent is to save space by processing the tensors in batches instead of the entire tensors at once.
> > 
> > I can look into extending the solution to use this
> > ```
> 
> I don't think we will need to implement this, but no point in reporting a match failure if the unrollDim is non-constant. Just make a comment somewhere in the conversion that the unrollDim does not change the result of the operation, so we do not use it in the conversion.

Comment that unrollDim does not impact output and is unused included in function description.

> 
> > ```
> > 3. I agree, it's way more straightforward doing it this way. Maybe I should edit my PR title to include the changes for einsum op? After merging main, the changes actually seem to have fixed 5 tests that were xfailed in `fx_importer_stablehlo` pipeline.
> > ```
> 
> Ah, good that it resolves some failing tests. We should rename the title to something like "support `aten._trilinear` and improve einsum decomposition". No TorchToLinalg flag, since this is a decomposition that affects other backends too.

Updated title of PR to: "support aten._trilinear and improve einsum decomposition"

zjgarvey (approved):
Looks good!

PR #3785, opened 2024-10-11
[Torch Dialect] Allow simplification of shape calculations of aten.tile, col2im, aten.stft

giacs-epic:
- Add `aten.mul.left_t` (+ canonicalizer) to allow simplification of aten.tile.
- Change syntax of the computation of col2im shape to allow the use of an already existing canonicalization pattern (for `aten.add.t`) for its simplification.
- Add `aten.eq.bool` ( + folder) to allow simplification of aten.stft.


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1111,6 +1127,36 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
   });
 }
 
+//===----------------------------------------------------------------------===//
+// AtenMulLeftTOp
+//===----------------------------------------------------------------------===//
+
+void AtenMulLeftTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
+                                                 MLIRContext *context) {
+  // `[3] * 5` -> `[3,3,3,3,3]`, if it is not mutated.
```
zjgarvey, line 1136: It might be better to have an example like `[1,2]*2 = [1,2,1,2]` in this comment.

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -1111,6 +1127,36 @@ void AtenLenTOp::getCanonicalizationPatterns(RewritePatternSet &patterns,
...
+
+    SmallVector<Value> newListElements;
+    for (int rep = 0; rep < numReps; ++rep) {
+      for (auto operand : listLiteral.getOperands()) {
+        newListElements.push_back(operand);
+      }
+    }
+
+    rewriter.replaceOpWithNewOp<PrimListConstructOp>(
+        op, Torch::ListType::get(newListElements[0].getType()),
```
zjgarvey, line 1154: I think it would be more reliable to do 
```suggestion
        op, op.getL().getType(),
```
E.g., when the list has heterogeneous tensor types, the list type will become more specific to the first element: `!torch.list<vtensor>` might become `!torch.list<vtensor<[1,2,3],f32>>` which causes a few issues, both with the updating of types for subsequent consumer ops, and for not being a valid type for other elements of the list with different shapes.

zjgarvey: Actually, better yet `op.getType()` will be the most reliable, since this op also returns a list.

zjgarvey (changes requested):
Thanks! A small change and I think this is good to go.

lib/Dialect/Torch/IR/TorchOps.cpp
giacs-epic: Thanks, I didn't think about lists with heterogeneous types.
Changed in last commit.

zjgarvey (approved):
A small nitpick if you think it would look better. If you like the folder as-is, I don't have any complaints and we can merge as-is.

zjgarvey:
Might also need to re-run the `build_tools/update_abstract_interp_lib.sh`

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -751,6 +751,22 @@ OpFoldResult Aten__Or__BoolOp::fold(FoldAdaptor adaptor) {
...
+OpFoldResult AtenEqBoolOp::fold(FoldAdaptor adaptor) {
+  if (getOperand(0) == getOperand(1))
+    return IntegerAttr::get(IntegerType::get(getContext(), 1), true);
+
+  bool a, b;
+  if (!matchPattern(getOperand(0), m_TorchConstantBool(&a)))
+    return nullptr;
+  if (!matchPattern(getOperand(1), m_TorchConstantBool(&b)))
+    return nullptr;
+  return IntegerAttr::get(IntegerType::get(getContext(), 1), a == b);
```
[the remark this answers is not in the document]
giacs-epic: Changed. Also updated `AtenNeBoolOp` for symmetry.

zjgarvey (approved):
LGTM

PR #3791, opened 2024-10-14
Add AtenSliceTOp Canonicalization to SimplifyShapeCalculations pass

zjgarvey:
Some ops were failing to infer the static component of partially dynamic shapes, and the cause was a missing aten.slice.t pattern.

The lit test included here is an IR dump created before DropAbstractInterpCalculations for an unflatten op that was failing to infer shapes before the change.

### DIFF
--- a/test/Dialect/Torch/simplify-shape-calculations.mlir
+++ b/test/Dialect/Torch/simplify-shape-calculations.mlir
@@ -489,3 +489,42 @@ func.func @shape_calc_with_two_uses(%arg0: !torch.vtensor<[2],f32>) -> !torch.vt
 
   return %arg0 : !torch.vtensor<[2],f32>
 }
+
+// CHECK-LABEL: func.func @unflat_shape_partial_dyn
+func.func @unflat_shape_partial_dyn(%arg0: !torch.vtensor<[?,?,3072],f32>) -> !torch.vtensor<[?,?,4,?],f32> {
+  %int768 = torch.constant.int 768
+  // CHECK-DAG: %[[INT768:.*]] = torch.constant.int 768
+  %int3072 = torch.constant.int 3072
+  %int0 = torch.constant.int 0
+  // CHECK-DAG: %[[INT0:.*]] = torch.constant.int 0
+  %int3 = torch.constant.int 3
+  %int1 = torch.constant.int 1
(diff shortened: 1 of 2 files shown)


test/Dialect/Torch/simplify-shape-calculations.mlir
```diff
@@ -489,3 +489,42 @@ func.func @shape_calc_with_two_uses(%arg0: !torch.vtensor<[2],f32>) -> !torch.vt
...
+    %8 = torch.aten.slice.t %4, %int3, %none, %int1 : !torch.list<int>, !torch.int, !torch.none, !torch.int -> !torch.list<int>
+    %9 = torch.aten.add.t %7, %8 : !torch.list<int>, !torch.list<int> -> !torch.list<int>
+    torch.shape.calculate.yield.shapes %9 : !torch.list<int>
+  } : !torch.vtensor<[?,?,4,?],f32>
+  // CHECK :     } shapes {
+  // CHECK :       %[[SZE0:.*]] = torch.aten.size.int %arg0, %[[INT0]] : !torch.vtensor<[?,?,3072],f32>, !torch.int -> !torch.int
+  // CHECK :       %[[SZE1:.*]] = torch.aten.size.int %arg0, %[[INT1]] : !torch.vtensor<[?,?,3072],f32>, !torch.int -> !torch.int
+  // CHECK :       %[[LIST:.*]] = torch.prim.ListConstruct %[[SZE0]], %[[SZE1]], %[[INT4]], %[[INT768]] : (!torch.int, !torch.int, !torch.int, !torch.int) -> !torch.list<int>
+  // CHECK :       torch.shape.calculate.yield.shapes %[[LIST]] : !torch.list<int>
+  // CHECK :     } : !torch.vtensor<[?,?,4,768],f32>
```
qedawkins, line 528: nit: move all in-line CHECK lines after test to be consistent with the style of the rest of this file

qedawkins (approved):

test/Dialect/Torch/simplify-shape-calculations.mlir
zjgarvey: Will do!

kumardeepakamd (approved):
Seems like a safe and simple change. approved.

PR #3795, opened 2024-10-15
[FxImporter] Added FxImporter test method to be executed via torch.co…

penguin-wwy:


stellaraccident (approved):
Looks good to me.

(as a side note, we may be able to move torch_mlir_e2e_test to the top level python directory at some point now)

PR #3797, opened 2024-10-15
Adds onnx ConvTranspose support for autopadding.

guacamoleo:
Adds onnx ConvTranspose support for autopadding (https://github.com/nod-ai/SHARK-ModelDev/issues/839).

- Adds support for attribute auto_pad="SAME_UPPER" or "SAME_LOWER" which will automatically calculate padding of input based on output shape.
- Adds support, during auto-padding, for output_shape=[H,W] which overrides the default output shape of input_shape[i]*stride[i] (for spatial dimensions only).
- Adds lit test for auto-padding.
- Tests are added by https://github.com/nod-ai/SHARK-TestSuite/pull/370

NOTE: ConvTranspose still doesn't support asymmetric padding, therefore multiple original onnx tests still won't pass.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -1736,7 +1736,8 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
                                              "Unimplemented: unranked tensor");
         unsigned rank = *maybeRank;
 
-        SmallVector<int64_t> padding, strides, dilations, outputPadding, outputShape;
+        SmallVector<int64_t> padding, strides, dilations, outputPadding,
+            outputShape;
         SmallVector<int64_t> defaultPadding, defaultStrides, defaultDilations,
             defaultOutputPadding;
         for (unsigned i = 0; i < rank - 2; i++) {
@@ -1777,7 +1778,7 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
               binder.op,
               "output_padding list size does not match the number of axes");
         }
-	        auto inputTensorType = cast<Torch::ValueTensorType>(input.getType());
+        auto inputTensorType = cast<Torch::ValueTensorType>(input.getType());
         if (!inputTensorType || !inputTensorType.hasSizes()) {
           return rewriter.notifyMatchFailure(
               binder.op, "Expected input type having sizes");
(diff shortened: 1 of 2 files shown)


renxida (approved):
looks good! before you merge, please confirm numerical correctness via e2etest or via shark-testsuite / iree-tests.

otherwise good job!

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1794,7 +1778,47 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+          return rewriter.notifyMatchFailure(
+              binder.op, "Expected input type having sizes");
+        }
+        ArrayRef<int64_t> inputShape = inputTensorType.getSizes();
 
+        if (autoPad == "NOTSET") {
+          // Explicit padding; read pads with defaults.
+          if (binder.s64IntegerArrayAttr(padding, "pads", defaultPadding))
+            return failure();
+        } else {
```
zjgarvey, line 1792: Although more verbose, I'd prefer if it were transparent that the cases considered here are: "SAME_LOWER" "SAME_UPPER" and "VALID". 

Maybe with the additional comment that, since we don't currently support asymmetric padding, this conversion will only apply when "SAME_LOWER" and "SAME_UPPER" would provide symmetric padding.

zjgarvey (commented):
Thanks for the PR! Do you know if this path supports "VALID" autopad? I can't infer what this option does through the onnx docs.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -1794,7 +1778,60 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
+            if (totalPadding % 2) {
+              // TODO: Add support for different padding values for the
+              // beginning and ending along each spatial axis.
+              return rewriter.notifyMatchFailure(
+                  binder.op,
+                  "unsupported conversion: the combination of stride, "
+                  "input_shape, kernel_shape, dilation, output_padding and "
+                  "output_shape caused auto-padding to produce asymmetric "
+                  "padding which isn't currently supported.");
+            }
```
zjgarvey, line 1818: It looks like we already have a match failure for asymmetric padding below, but I think this error messaging is much better for the specific auto-pad cases that are failing, so let's keep this here.

zjgarvey (approved):
Very good. Let's get this merged.

Thanks!

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
guacamoleo: Agreed.

PR #3798, opened 2024-10-16
[Fx Importer] fix mutation importer with non persistent buffer

dixinzhou:
A non-persistent buffer will not be a part of this module’s `state_dict`. Hence when setting `experimental_support_mutation=True` and have non-persistent buffer, the current fx importer will fail to retrieve a value from `state_dict` and produce `torch.constant.none` to represent the buffer. This fix get value of non-persistent buffer from the module's `constants`.

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -723,10 +723,17 @@ class FxImporter:
                 # on a symbolic or other non-SSA association. As such, they
                 # are not modeled with mutable IR but will trigger an output
                 # store hook when the final value is produced.
-                value = prog.state_dict.get(input_spec.target)
-                assert (
-                    not input_spec.persistent or value is not None
-                ), "Expected state_dict value for persistent value"
+                if input_spec.persistent == True:
+                    value = prog.state_dict.get(input_spec.target)
+                    assert (
+                        value is not None
+                    ), "Expected state_dict value for persistent buffer"
+                elif input_spec.persistent == False:
+                    value = prog.constants.get(input_spec.target)
+                    assert (
+                        value is not None
+                    ), "Expected constants value for non-persistent buffer"
+
                 node = placeholder_nodes[arg.name]
                 mutable_producer_node_name = mutable_buffer_target_producers.get(
                     input_spec.target
--- a/test/python/fx_importer/v2.3/mutation_import.py
+++ b/test/python/fx_importer/v2.3/mutation_import.py
@@ -105,6 +105,27 @@ def test_frozen_buffer():
     m.operation.verify()
 
 
+@run
+# CHECK-LABEL: test_frozen_buffer_non_persistent
+# CHECK: %[[buffer_literal:.+]] = torch.vtensor.literal
+# CHECK: %[[mul:.+]] = torch.aten.mul.Tensor %arg0, %0
+# CHECK: return %[[mul]]
+def test_frozen_buffer_non_persistent():
+    class Basic(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.register_buffer("buffer", torch.randn(3, 4), persistent=False)
+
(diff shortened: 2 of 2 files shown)


dixinzhou:
Hi @stellaraccident . I don't have permission to request review yet. I am tagging you here as you were the original author for support mutation in Exported Program. Do you mind taking a look at this change? Thanks!

stellaraccident:
Thanks, yes I'll review. It might take me a bit as it takes me a minute to context switch and make sure I understand what is going on in this part.

dixinzhou:
Hi @stellaraccident, a gentle reminder to review this pull request when you get a chance. Thank you!

python/torch_mlir/extras/fx_importer.py
```diff
@@ -723,10 +723,17 @@ def import_program(
                 # on a symbolic or other non-SSA association. As such, they
                 # are not modeled with mutable IR but will trigger an output
                 # store hook when the final value is produced.
-                value = prog.state_dict.get(input_spec.target)
-                assert (
-                    not input_spec.persistent or value is not None
-                ), "Expected state_dict value for persistent value"
+                if input_spec.persistent == True:
```
penguin-wwy, line 726: Please use `is True` instead of equality operators.
The same applies to `is False`

stellaraccident:
Sorry - have been busy. I will review tomorrow

python/torch_mlir/extras/fx_importer.py
stellaraccident: Similar comment: what are you trying to say with branches for == True and elif == False? Is this tri-state logic (i.e. None, True, False)? If so, please structure it as:

```
if input_spec.persistent is None:
  ...
elif input_spec.persistent:
  ...
else:
  ...
```

And leave a comment on why this is tri-state. If you are just trying to branch on boolean, please use normal truthy/falsey:

```
if input_spec.persistent:
  ...
else:
  ...
```

Having either tri-state or very specific checks for boolean values basically screams to the reader that something very weird is going on here. If something weird is going on, leave a comment about that. If not, write it with normal truthy control flow.

test/python/fx_importer/v2.3/mutation_import.py
```diff
@@ -105,6 +105,27 @@ def forward(self, x):
...
+@run
+# CHECK-LABEL: test_frozen_buffer_non_persistent
+# CHECK: %[[buffer_literal:.+]] = torch.vtensor.literal
+# CHECK: %[[mul:.+]] = torch.aten.mul.Tensor %arg0, %0
+# CHECK: return %[[mul]]
+def test_frozen_buffer_non_persistent():
+    class Basic(nn.Module):
+        def __init__(self):
+            super().__init__()
+            self.register_buffer("buffer", torch.randn(3, 4), persistent=False)
```
stellaraccident, line 117: Can you think of any way to test persistent=True? Looks like that code path doesn't get tested.

dixinzhou: When `register_buffer`, the `persistent` is by default set to be true. `register_buffer(name, tensor, persistent=True)`. So the `test_frozen_buffer()` test that run before the added `test_frozen_buffer_non_persistent()` already covered the `persistent=True` case. But if you prefer, I can make `persistent=True` explicit in the `test_frozen_buffer()` test.

stellaraccident: Oh, thanks. That works then. I wasn't looking at all of the tests in context.

stellaraccident (approved):

python/torch_mlir/extras/fx_importer.py
dixinzhou: Thanks @penguin-wwy @stellaraccident for your review! I have updated the PR to address your feedback.

penguin-wwy (approved):

PR #3799, opened 2024-10-16
Rework Scalarize Shapes Pass

zjgarvey:
This is a first step towards reworking the scalarize-shapes pass which has been integral to our ONNX frontend path detangling shape computations. 

## Purpose:

1. Restrict the scope of the pass to only apply to op sequences which are used to compute shapes.
2. Make the pass more efficient by applying patterns in an appropriate order for scalarization propagation.
3. Report failed scalarization patterns for easier debugging (Not yet implemented). I can't seem to find a good path for this right now to capture the right diagnostics. I'd like to defer this addition to a later patch so we can add some high-value patterns to this pass in the meantime. 

With these changes, some reworking of the conversions themselves will be necessary. 

1. The removal of the SqueezeDim fold pattern was an appropriate fix to avoid folding a pattern that may be needed to propagate further. The reversal of pattern application order uncovered this bug. The addition of rank 0 item logic was added to replace the functionality needed from the squeeze dim pattern.
2. Rework getListFromTensor to modify a `SmallVector<OpFoldResult>` to allow processing value tensor literals without immediately materializing the ints. This should factor out a significant portion of code that was used in specific cases to handle constants. 

## RFC 1:

Currently, we are going to add all prim list of int ops to the worklist. Can anyone identify problems with uniformly anchoring on prim lists of ints? E.g. Does there exist a Torch Op satisfying all of the following conditions:

1. Accepts a list of constant ints, LIST, as an input
2. The role of LIST is **not** shape related. All the examples I can think of are indeed shape related: padding ints passed to a pad op, kernel size ints passed to a conv op, size ints passed to a view op, etc.
4. The LIST is not gotten entirely from scalars already. 

If there does not exist a torch op satisfying all three of those conditions, I think it will be safe to "anchor" on prim lists of ints.

### Conclusion for RFC 1: 

I just scanned through the `GeneratedTorchOps.td` and `TorchOps.td` for all references of `AnyTorchListOfTorchIntType` and verified this will not be problematic to apply in any of those cases. 

## RFC 2:

What should I use to report failed scalarization?

Like my dumb idea was just to walk back through the func op after applying the passes and check if anything in the worklist is still a tensor. If so, emit/log a warning. It certainly works, since you can just look at the warnings and start debugging from the last printed warning upwards, but there has to be a better way to handle this without walking back through the func.func op. 

### Conclusion for RFC 2:

I tried a few things without much success. The fundamental problem is that identifying the cause of a failed scalarization could be myriad: 

1. We could be missing a pattern for an op entirely: E.g., a pattern we need is scalarizing rank0 arithmetic ops (e.g. AtenMulTensorOp -> AtenMulIntOp). 
2. We could fail a scalarization pattern because it should fold instead. This is specifically the case for rank0 where.self ops. These ops MUST fold, or we need to have custom lowering logic for the rank 0 case. 
3. Walking through the func op a second time and emiting a warning for ops that have tensor result types seems to give locations that are inconsistent or hard to track in the converted IR. Doing this on IR that doesn't apply any patterns seems to give decent information, but it's still dramatically insufficient considering how complex these patterns can get, and still takes manually reading IR to try and figure out what is really blocking the simplification.

I'd like to skip out on fleshing out the error reporting for now and come back to it after iterating a few time on the patterns.


zjgarvey:
Okay, scalarizing value tensor literals does not work, since they get folded back to value tensor literals.

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -898,33 +981,55 @@ class ScalarizeShapesPass : public ScalarizeShapesBase<ScalarizeShapesPass> {
...
+                !isSourceOpForShapeScalarization(userOp) &&
+                !isa<AtenViewOp>(userOp)) {
+              shapeCalculationOps.insert(op);
+              return;
+            }
+          }
+        });
+
+    GreedyRewriteConfig config;
+    config.strictMode = GreedyRewriteStrictness::ExistingAndNewOps;
```
stellaraccident, line 1024: Ok, so this is actually the key thing to this algorithm and needs some explanation: You are relying on patterns that, given some anchor, always rewrite to produce some new ops that (if there are further ops above) introduce an op that is a new anchor that goes on the worklist. So this is still a bottom up rewrite where each step is producing a locally correct graph transformation (at the cost of each step needing to create a number of new ops that will just be thrown away on the next pattern match).

This is a bit different from what I had described: do a first traversal to identify all dominators that end with an anchor. Then transform top-down, where each pattern produces an invalid graph transformation that does not need to materialize throw away ops at each step. In this approach, you would make one pass top down through the IR doing transformations and noting errors for anything that could not be converted.

Given the code that already exists for this and that the top-down way is definitely harder to write, doing it this way is fine. The cost will be more garbage/allocations that scale with the number of changes you have to make. And it is harder to do precise error handling. The up-side, though, is that if a pattern fails to fully scalarize, the graph is in a consistent but suboptimal state and can still be used.

So to answer the two questions you raised in the PR description:

1. PrimListOfInts may be an ok anchor. In my mind, that was an anchor but only if it was consumed as an index-oriented operand of an op. It may be a distinction without a difference in the end. I'm just paranoid because I know that in these kinds of graphs, there are definitely weird bottlenecks of such things that need more discrimination.
2. For error handling, if you don't want to do another walk, create your "Propagate*" pattern classes to take a reference to some PropagationState&. For things your patterns encounter that indicate a failure to convert that you want to report, add then to some set in there. Then at the end of your pass, emit one warning with multiple notes indicating the ops that could not be converted. The reason you want your patterns themselves to be silent is that patterns can be run multiple times during iteration and it is impossible to have precise control of your diagnostic behavior as a result.

zjgarvey: Thanks for the comments and offline discussion yesterday, Stella.

I added some more descriptive comments about the walk to collect the `shapeCalculationOps`, so hopefully this resolves some concerns mentioned in your first paragraph. I'd like to keep the graph transformations legal for now, since this will take some time to nail down all of the patterns we need to implement. Reading the debug printouts for the pass as it is, I think we traverse the IR and apply patterns in a safe and reasonably efficient way without having to invent a bespoke PatternRewriteDriver. 

1. I'll scan through the op registry for prim list of int inputs for ops and look for problematic examples. Should be able to remove those cases manually if they aren't common.
2. Sounds like a good plan for error handling.

zjgarvey: I updated the original comment with where I am at with this, so I'm going to resolve this conversation for now.

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -99,14 +82,62 @@ LogicalResult getListFromTensor(Value value, SmallVector<Value> &vals) {
...
+  // materialize
+  auto ty = cast<ValueTensorType>(literalOp.getType());
+  if (!ty.hasSizes() || ty.getSizes().size() > 1)
+    return failure();
+  int64_t listSize = ty.getSizes().size() == 1 ? ty.getSizes().front() : 1;
+  auto intTy = dyn_cast_or_null<IntegerType>(ty.getDtype());
+  if (!intTy || intTy.isUnsigned())
+    return failure();
+
+  if (auto splattr =
```
rsuderman, line 106: You can probably restructure this a bit for clarity:

```
auto splattr = dyn_cast_or_null<SplatElementsAttr>(literalOp.getValue());
auto denseAttr = dyn_cast_or_null<DenseIntElementsAttr>(literalOp.getValue());
if (!splattr && !denseAttr) return false;
if (splattr) ...
if (denseAttr) ...
```

It just handles the escape case first and makes it clear what the block is trying to express

rsuderman (approved):

PR #3801, opened 2024-10-17
[MLIR][TORCH] Add torch-onnx-to-torch-backend pipeline

vivekkhandelwal1:
This commit adds the torch-onnx-to-torch-backend pipeline which
converts the Torch Onnx IR to Torch Backend IR.

This commit also moves the `ScalarizeShapes` pass from the
`torch-backend-to-linalg-on-tensors-backend-pipeline` to the
`torch-onnx-to-torch-backend` pipeline since the primary goal of
this pass is to scalarize the shapes in the IR coming from the
Onnx models.

### DIFF
--- a/lib/Dialect/Torch/Transforms/Passes.cpp
+++ b/lib/Dialect/Torch/Transforms/Passes.cpp
@@ -102,9 +102,10 @@ void mlir::torch::Torch::createTorchOnnxToTorchBackendPipeline(
         Torch::createDecomposeComplexOpsPass(options.backendLegalOps));
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   }
-  // TODO: Move the below two passes i.e., ScalarizeShapes and
+  // TODO: Move the combination of two passes i.e., ScalarizeShapes and
   // TorchShapeRefinementPipeline out of here and create an onnx shape
   // refinement pipeline which runs iteratively over the IR.
+  createTorchShapeRefinementPipeline(pm, options);
   // This pass scalarizes the tensor shape computations.
   pm.addNestedPass<mlir::func::FuncOp>(
       mlir::torch::Torch::createScalarizeShapesPass());
--- a/lib/Dialect/Torch/Transforms/Passes.cpp
+++ b/lib/Dialect/Torch/Transforms/Passes.cpp
@@ -91,15 +91,6 @@ void mlir::torch::Torch::createTorchFunctionToTorchBackendPipeline(
       options.backendLegalOps, options.extraLibrary));
 }
 
-static void createOnnxShapeRefinementPipeline(
-    mlir::OpPassManager &pm,
-    const mlir::torch::Torch::TorchLoweringPipelineOptions &options) {
-  // This pass scalarizes the tensor shape computations.
-  pm.addNestedPass<mlir::func::FuncOp>(
-      mlir::torch::Torch::createScalarizeShapesPass());
-  createTorchShapeRefinementPipeline(pm, options);
-}
-
 void mlir::torch::Torch::createTorchOnnxToTorchBackendPipeline(
     OpPassManager &pm, const TorchLoweringPipelineOptions &options) {
   pm.addNestedPass<func::FuncOp>(onnx_c::createTorchOnnxToTorchPass());
@@ -111,9 +102,18 @@ void mlir::torch::Torch::createTorchOnnxToTorchBackendPipeline(
         Torch::createDecomposeComplexOpsPass(options.backendLegalOps));
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   }
-  createOnnxShapeRefinementPipeline(pm, options);
+  // TODO: Move the below two passes i.e., ScalarizeShapes and
+  // TorchShapeRefinementPipeline out of here and create an onnx shape
+  // refinement pipeline which runs iteratively over the IR.
+  // This pass scalarizes the tensor shape computations.
+  pm.addNestedPass<mlir::func::FuncOp>(
+      mlir::torch::Torch::createScalarizeShapesPass());
+  createTorchShapeRefinementPipeline(pm, options);
   pm.addPass(Torch::createRefinePublicReturnPass());
   pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+  // The decompose pass is run again here since the scalarize shapes pass and
+  // shape refinement pipeline might create some ops for which decomposition
+  // exists.
   if (options.decompose) {
     pm.addNestedPass<func::FuncOp>(
         Torch::createDecomposeComplexOpsPass(options.backendLegalOps));
--- a/lib/Dialect/Torch/Transforms/Passes.cpp
+++ b/lib/Dialect/Torch/Transforms/Passes.cpp
@@ -10,6 +10,7 @@
 #include "torch-mlir/Dialect/Torch/Transforms/Passes.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Transforms/Passes.h"
+#include "torch-mlir/Conversion/TorchOnnxToTorch/Passes.h"
 
 void mlir::torch::registerTorchPasses() {
   mlir::torch::registerPasses();
@@ -25,6 +26,10 @@ void mlir::torch::registerTorchPasses() {
       "torch-function-to-torch-backend-pipeline",
       "Pipeline lowering a Torch function to Torch backend form.",
       mlir::torch::Torch::createTorchFunctionToTorchBackendPipeline);
+  mlir::PassPipelineRegistration<Torch::TorchLoweringPipelineOptions>(
+      "torch-onnx-to-torch-backend-pipeline",
+      "Pipeline lowering Torch Onnx IR to Torch backend form.",
+      mlir::torch::Torch::createTorchOnnxToTorchBackendPipeline);
   mlir::PassPipelineRegistration<Torch::TorchLoweringPipelineOptions>(
       "torch-simplification-pipeline",
       "Pipeline simplifying computations in the program.",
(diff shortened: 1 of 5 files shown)


lib/Dialect/Torch/Transforms/Passes.cpp
```diff
@@ -86,6 +91,15 @@ void mlir::torch::Torch::createTorchFunctionToTorchBackendPipeline(
       options.backendLegalOps, options.extraLibrary));
 }
 
+void mlir::torch::Torch::createTorchOnnxToTorchBackendPipeline(
+    OpPassManager &pm, const TorchLoweringPipelineOptions &options) {
+  pm.addPass(onnx_c::createTorchOnnxToTorchPass());
+  // The above pass just converts the torch onnx IR to torch, hence the given
+  // pipeline will make sure that the IR is transformed such that it satisfies
+  // the backend contract.
+  createTorchFunctionToTorchBackendPipeline(pm, options);
```
zjgarvey, line 100: We don't need most of the passes in this pipeline. I'd like to start with an initial `--torch-decompose-complex-ops`, then let's make a separate `--onnx-shape-refinement-pipeline` to put here that iteratively performs the following: 

`--torch-scalarize-shapes` and `--torch-shape-refinement-pipeline`. 

Then we will follow up with:

`--torch-refine-public-return`
`--canonicalize`
`--torch-decompose-complex-ops`
`--canonicalize`

zjgarvey (changes requested):
Like we were discussing in the sync, I think since scalarize shapes is really onnx specific right now, we should only perform this pass in the onnx-to-torch pipeline.

test/Dialect/Torch/torch-onnx-to-torch-backend-pipeline.mlir
```diff
@@ -0,0 +1,30 @@
...
+  // CHECK: %[[UNSQUEEZE:.+]] = torch.aten.unsqueeze %[[ARANGE]], %[[INT1]] : !torch.vtensor<[4],si64>, !torch.int -> !torch.vtensor<[4,1],si64>
+  // CHECK: %[[UNSQUEEZE_0:.+]] = torch.aten.unsqueeze %[[ARANGE_0]], %[[INT0]] : !torch.vtensor<[5],si64>, !torch.int -> !torch.vtensor<[1,5],si64>
+  // CHECK: %[[ADD:.+]] = torch.aten.add.Scalar %[[UNSQUEEZE]], %[[INT0]], %[[INT1]] : !torch.vtensor<[4,1],si64>, !torch.int, !torch.int -> !torch.vtensor<[4,1],si64>
+  // CHECK: %[[COND:.+]] = torch.aten.ge.Tensor %[[UNSQUEEZE_0]], %[[ADD]] : !torch.vtensor<[1,5],si64>, !torch.vtensor<[4,1],si64> -> !torch.vtensor<[4,5],i1>
+  // CHECK: %[[RESULT:.+]] = torch.aten.where.self %[[COND]], %arg0, %[[ZERO_TENSOR]] : !torch.vtensor<[4,5],i1>, !torch.vtensor<[4,5],si64>, !torch.vtensor<[],si64> -> !torch.vtensor<[4,5],si64>
+  %0 = torch.operator "onnx.Trilu"(%arg0) : (!torch.vtensor<[4,5],si64>) -> !torch.vtensor<[4,5],si64>
+  return %0 : !torch.vtensor<[4,5],si64>
+}
```
zjgarvey, line 30: It would be a better test of functionality to modify <https://github.com/llvm/torch-mlir/blob/9c7067649b9b8373b78d2332101d7211f5aeddb3/projects/pt1/python/torch_mlir_e2e_test/configs/onnx_backend.py#L104-L128> and see if there are any regressions/progressions in the e2e tests.

If we are adding lit tests, it would be better to add a test that actually uses scalarize shapes in a useful way, E.g.:

```mlir
module {
  func.func @"torch-jit-export"(%arg0: !torch.vtensor<[?,?,16,64],f32>) -> !torch.vtensor<[?,?,?],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 21 : si64, torch.onnx_meta.producer_name = "pytorch", torch.onnx_meta.producer_version = "1.11.0"} {
    %0 = torch.operator "onnx.Shape"(%arg0) : (!torch.vtensor<[?,?,16,64],f32>) -> !torch.vtensor<[4],si64> 
    %1 = torch.operator "onnx.Constant"() {torch.onnx.value = dense_resource<__21> : tensor<si64>} : () -> !torch.vtensor<[],si64> 
    %2 = torch.operator "onnx.Gather"(%0, %1) {torch.onnx.axis = 0 : si64} : (!torch.vtensor<[4],si64>, !torch.vtensor<[],si64>) -> !torch.vtensor<[],si64> 
    %3 = torch.operator "onnx.Shape"(%arg0) : (!torch.vtensor<[?,?,16,64],f32>) -> !torch.vtensor<[4],si64> 
    %4 = torch.operator "onnx.Constant"() {torch.onnx.value = dense_resource<__22> : tensor<si64>} : () -> !torch.vtensor<[],si64> 
    %5 = torch.operator "onnx.Gather"(%3, %4) {torch.onnx.axis = 0 : si64} : (!torch.vtensor<[4],si64>, !torch.vtensor<[],si64>) -> !torch.vtensor<[],si64> 
    %6 = torch.operator "onnx.Constant"() {torch.onnx.value = dense<0> : tensor<1xsi64>} : () -> !torch.vtensor<[1],si64> 
    %7 = torch.operator "onnx.Unsqueeze"(%2, %6) : (!torch.vtensor<[],si64>, !torch.vtensor<[1],si64>) -> !torch.vtensor<[1],si64> 
    %8 = torch.operator "onnx.Constant"() {torch.onnx.value = dense<0> : tensor<1xsi64>} : () -> !torch.vtensor<[1],si64> 
    %9 = torch.operator "onnx.Unsqueeze"(%5, %8) : (!torch.vtensor<[],si64>, !torch.vtensor<[1],si64>) -> !torch.vtensor<[1],si64> 
    %10 = torch.operator "onnx.Constant"() {torch.onnx.value = dense_resource<_onnx__Concat_3209> : tensor<1xsi64>} : () -> !torch.vtensor<[1],si64> 
    %11 = torch.operator "onnx.Concat"(%7, %9, %10) {torch.onnx.axis = 0 : si64} : (!torch.vtensor<[1],si64>, !torch.vtensor<[1],si64>, !torch.vtensor<[1],si64>) -> !torch.vtensor<[3],si64> 
    %12 = torch.operator "onnx.Reshape"(%arg0, %11) : (!torch.vtensor<[?,?,16,64],f32>, !torch.vtensor<[3],si64>) -> !torch.vtensor<[?,?,?],f32> 
    return %12 : !torch.vtensor<[?,?,?],f32>
  }
}

{-#
  dialect_resources: {
    builtin: {
      __21: "0x080000000000000000000000",
      __22: "0x080000000100000000000000",
      _onnx__Concat_3209: "0x080000000004000000000000"
    }
  }
#-}
```

Should get converted to:

```mlir
module {
  func.func @"torch-jit-export"(%arg0: !torch.vtensor<[?,?,16,64],f32>) -> !torch.vtensor<[?,?,1024],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 21 : si64, torch.onnx_meta.producer_name = "pytorch", torch.onnx_meta.producer_version = "1.11.0"} {
    %int2 = torch.constant.int 2
    %int3 = torch.constant.int 3
    %0 = torch.aten.flatten.using_ints %arg0, %int2, %int3 : !torch.vtensor<[?,?,16,64],f32>, !torch.int, !torch.int -> !torch.vtensor<[?,?,1024],f32>
    return %0 : !torch.vtensor<[?,?,1024],f32>
  }
}
```

lib/Dialect/Torch/Transforms/Passes.cpp
```diff
@@ -86,6 +91,36 @@ void mlir::torch::Torch::createTorchFunctionToTorchBackendPipeline(
       options.backendLegalOps, options.extraLibrary));
 }
 
+static void createOnnxShapeRefinementPipeline(
```
zjgarvey, line 94: We should either have this as an actual pass that runs iteratively, or we should just put those two in the other pipeline as-is. 

For the sake of getting this in sooner rather than later, let's just remove this function and add the two passes to the onnx-to-torch-backend-pipeline.

lib/Dialect/Torch/Transforms/Passes.cpp
```diff
@@ -86,6 +91,36 @@ void mlir::torch::Torch::createTorchFunctionToTorchBackendPipeline(
...
+  if (options.decompose) {
+    pm.addNestedPass<func::FuncOp>(
+        Torch::createDecomposeComplexOpsPass(options.backendLegalOps));
+    pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+  }
+  createOnnxShapeRefinementPipeline(pm, options);
+  pm.addPass(Torch::createRefinePublicReturnPass());
+  pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
+  if (options.decompose) {
+    pm.addNestedPass<func::FuncOp>(
```
zjgarvey, line 118: I would add a comment explaining why we run this again.

zjgarvey (changes requested):
Nice, a few changes and I think we should be good to go.

lib/Dialect/Torch/Transforms/Passes.cpp
```diff
@@ -111,9 +102,18 @@ void mlir::torch::Torch::createTorchOnnxToTorchBackendPipeline(
         Torch::createDecomposeComplexOpsPass(options.backendLegalOps));
     pm.addNestedPass<func::FuncOp>(createCanonicalizerPass());
   }
-  createOnnxShapeRefinementPipeline(pm, options);
+  // TODO: Move the below two passes i.e., ScalarizeShapes and
+  // TorchShapeRefinementPipeline out of here and create an onnx shape
+  // refinement pipeline which runs iteratively over the IR.
+  // This pass scalarizes the tensor shape computations.
```
zjgarvey, line 108: Now that I think about it, at least until we perform the shape refinement/scalarization iteratively, let's include one pass of shape refinement before scalarize shapes, too.

Shape scalarization leans pretty heavily on folding `aten.where.self` operations, and sometimes this is only possible when a `size.int` op is foldable to a constant.

zjgarvey (changes requested):
One last quick change (sorry for the back and forth), then I think this is good to go. Glad to see the CI green on this change.

zjgarvey (approved):
I think this is a good starting point. Let's get this in and iterate on the pipeline as the need arises.

PR #3802, opened 2024-10-17
OnnxToTorch bicubic interpolation

aldesilv:
(https://github.com/nod-ai/SHARK-TestSuite/pull/391)
Repro (using SHARK TestSuite):
1. `python run.py --torchtolinalg -m cl-onnx-iree -t cubic_test`


lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2895,6 +2920,168 @@ static Value BilinearInterpolate(OpBuilder &b,
   return b.create<arith::AddFOp>(loc, left, right);
 }
 
+static Value WeightFunction(OpBuilder &b, Location loc, Value xDistance) {
```
zjgarvey, line 2923: We shouldn't need to determine which branch of the piecewise definition to use at runtime. `x_1` and `x1` should only compute `lessEqualOne`, and `x_2` and `x2` should only use `lessThanTwo`. Nothing is going to be outside that range, so we shouldn't need anything else. This might also be the source of lingering numerical inaccuracy on the boundary, since the `xDistance` you are passing here is based on the clamped values, but it is possible that the OOB elements still use the part of the piecewise function that corresponds to the OOB position, while using the input pixel intensity from the nearest in-bounds pixel. 

Also, it will probably be better to have this as a lambda function inside the bicubic interpolate, so we can capture constants instead of rewriting them 16 times.

zjgarvey (commented):
Hey, Alex. Looking this over, I have an idea that might resolve the numerical issues. Do you mind taking a look?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
zjgarvey: Oh, nevermind, the clamping happens after. Oh well. Should still make the changes mentioned here.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2895,6 +2920,168 @@ static Value BilinearInterpolate(OpBuilder &b,
...
+
+  for (int j = 0; j < 4; j++) {
+    Value wy = WeightFunction(b, loc, yDistance[j]);
+    Value xInterpy = xInterp[j];
+    for (int i = 0; i < 4; i++) {
+      Value wx = WeightFunction(b, loc, xDistance[i]);
+
+      Value cmp =
+          b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGE, y[j], zero);
+      y[j] = b.create<arith::SelectOp>(loc, cmp, y[j], zero);
```
zjgarvey, line 3050: I think using a max here is cleaner.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2895,6 +2920,168 @@ static Value BilinearInterpolate(OpBuilder &b,
...
+
+      Value inputWSubOne = b.create<arith::SubFOp>(loc, inputFPW, cstOneFloat);
+      Value inputHSubOne = b.create<arith::SubFOp>(loc, inputFPH, cstOneFloat);
+      cmp = b.create<arith::CmpFOp>(loc, arith::CmpFPredicate::UGT, y[j],
+                                    inputHSubOne);
+      y[j] = b.create<arith::SelectOp>(loc, cmp, inputHSubOne, y[j]);
```
zjgarvey, line 3056: similarly, here might be cleaner with a min

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2895,6 +2920,168 @@ static Value BilinearInterpolate(OpBuilder &b,
...
+  Value x_2Distance = b.create<arith::SubFOp>(loc, proj[1], x_2);
+  x_2Distance = b.create<math::AbsFOp>(loc, x_2Distance);
+
+  SmallVector<Value> y{y_2, y_1, y1, y2};
+  SmallVector<Value> x{x_2, x_1, x1, x2};
+  SmallVector<Value> yDistance{y_2Distance, y_1Distance, y1Distance,
+                               y2Distance};
+  SmallVector<Value> xDistance{x_2Distance, x_1Distance, x1Distance,
+                               x2Distance};
+  SmallVector<Value> xInterp{zero, zero, zero, zero};
```
zjgarvey, line 3036: I don't think we need this, just set `Value xInterpy = zero;` in the loop below?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2895,6 +2920,168 @@ static Value BilinearInterpolate(OpBuilder &b,
...
+
+  // f(x_orig, y_orig) = Sum_y Sum_x W(x_original - x)*input[x,y]
+  //                     * W(y_original - y)
+  Value fxy = zero;
+
+  for (int j = 0; j < 4; j++) {
+    Value wy = WeightFunction(b, loc, yDistance[j]);
+    Value xInterpy = xInterp[j];
+    for (int i = 0; i < 4; i++) {
```
zjgarvey, line 3045: Everything in this loop is being computed 16 times. I think we should compute the weights and clamped index values for the four x and y points in parallel loops (rather than nested loops),

zjgarvey: So we should have four `SmallVector<Value>`

1. "xWeights"
2. "xIndices"
3. "yWeights"
4. "yIndices"

The first two get computed in a loop over x neighbors, the latter over y neighbors. Then you can perform the extracts, products, and accumulation in a nested loop without having to re-write the same IR multiple times.

zjgarvey (commented):
More comments!

lib/Conversion/TorchToLinalg/Uncategorized.cpp
zjgarvey: Or better yet, they can just go in the same loop, since you can do them in parallel...

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2926,6 +2951,158 @@ static Value BilinearInterpolate(OpBuilder &b,
...
+
+    Value yInt = b.create<arith::FPToSIOp>(loc, b.getI64Type(), y[j]);
+    Value yIndex = b.create<arith::IndexCastOp>(loc, b.getIndexType(), yInt);
+    indices[dimOffset] = yIndex;
+
+    for (int i = 0; i < 4; i++) {
+      Value wx = wxs[i];
+
+      x[i] = b.create<arith::MaximumFOp>(loc, x[i], zero);
+      Value inputWSubOne = b.create<arith::SubFOp>(loc, inputFPW, cstOneFloat);
```
zjgarvey, line 3087: I'd move this out of the loop.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2926,6 +2951,158 @@ static Value BilinearInterpolate(OpBuilder &b,
...
+    y[j] = b.create<arith::MinimumFOp>(loc, y[j], inputHSubOne);
+
+    Value yInt = b.create<arith::FPToSIOp>(loc, b.getI64Type(), y[j]);
+    Value yIndex = b.create<arith::IndexCastOp>(loc, b.getIndexType(), yInt);
+    indices[dimOffset] = yIndex;
+
+    for (int i = 0; i < 4; i++) {
+      Value wx = wxs[i];
+
+      x[i] = b.create<arith::MaximumFOp>(loc, x[i], zero);
```
zjgarvey, line 3086: This is happening too many times. We should also compute these clips outside the loop body.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2926,6 +2951,158 @@ static Value BilinearInterpolate(OpBuilder &b,
...
+  x2Distance = b.create<math::AbsFOp>(loc, x2Distance);
+  Value x1Distance = b.create<arith::SubFOp>(loc, proj[1], x1);
+  x1Distance = b.create<math::AbsFOp>(loc, x1Distance);
+  Value x_1Distance = b.create<arith::SubFOp>(loc, proj[1], x_1);
+  x_1Distance = b.create<math::AbsFOp>(loc, x_1Distance);
+  Value x_2Distance = b.create<arith::SubFOp>(loc, proj[1], x_2);
+  x_2Distance = b.create<math::AbsFOp>(loc, x_2Distance);
+
+  SmallVector<Value> y{y_2, y_1, y1, y2};
+  SmallVector<Value> x{x_2, x_1, x1, x2};
```
zjgarvey, line 3055: Since we are no longer using the fp x and y values, we should clip them to a valid index and store **those** in a small vector (before getting to the loop below).

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2926,6 +2951,158 @@ static Value BilinearInterpolate(OpBuilder &b,
...
+  SmallVector<Value> yDistance{y_2Distance, y_1Distance, y1Distance,
+                               y2Distance};
+  SmallVector<Value> xDistance{x_2Distance, x_1Distance, x1Distance,
+                               x2Distance};
```
zjgarvey, line 3059: These are unused.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2947,7 +2947,8 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             binder.f32FloatAttr(extrapolation_value, "extrapolation_value",
                                 0.0) ||
             binder.customOpNameStringAttr(nearest_mode, "nearest_mode",
-                                          "round_prefer_floor"))
+                                          "round_prefer_floor") ||
+            binder.f32FloatAttr(cubic_coeff_a, "cubic_coeff_a", -0.75))
```
zjgarvey, line 2951: Please report a match failure if `cubic_coeff_a != -0.75`.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2771,12 +2771,13 @@ static Value NearestInterpolate(OpBuilder &b, Location loc,
...
 }
 
-static Value BilinearInterpolate(OpBuilder &b,
-                                 Aten__InterpolateSizeListScaleListOp op,
-                                 Location loc, SmallVector<Value> outputSizes,
-                                 Value input, SmallVector<Value> inputSizes,
-                                 SmallVector<Value> scaleValues,
-                                 std::string coordStr) {
+static SmallVector<Value>
+CoordinateTransform(OpBuilder &b, Aten__InterpolateSizeListScaleListOp op,
```
zjgarvey, line 2775: nit: I just noticed this, but all these functions should be camel case. Class names start with a capital, but function names do not.

zjgarvey (changes requested):
Before making changes, check out the suggestions for fixing the numerics in <https://github.com/aldesilv/torch-mlir/pull/3>.

Besides the suggestions in that PR, and the comments in this review, I'd generally like more comments in this code. Where does this implementation come from? What assumptions did we make? 

To anyone who doesn't have context on cubic interpolation, this code would be completely inscrutable.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
aldesilv: removed

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2926,6 +2955,176 @@ static Value BilinearInterpolate(OpBuilder &b,
...
+
+    Value yInt = b.create<arith::FPToSIOp>(loc, b.getI64Type(), y[j]);
+    Value yIndex = b.create<arith::IndexCastOp>(loc, b.getIndexType(), yInt);
+    indices[dimOffset] = yIndex;
+
+    for (int i = 0; i < 4; i++) {
+      Value wx = wxs[i];
+
+      Value xInt = b.create<arith::FPToSIOp>(loc, b.getI64Type(), x[i]);
+      Value xIndex = b.create<arith::IndexCastOp>(loc, b.getIndexType(), xInt);
```
zjgarvey, line 3113: I'd do these in the previous loop. `x[i] = b.create<arith::FPToSIOp> ...` then just set `indices[dimOffset + 1] = x[j]` here.

zjgarvey (commented):
Great! One small change and should be good to go. 

It might also be a good idea to rework some of the function signatures to take const references to avoid making unnecessary copies.

zjgarvey (approved):
LGTM

PR #3805, opened 2024-10-18
Disable building STABLEHLO and specify USE_MATH_DEFINES for windows builds.

sahas3:
I'm trying to build python wheel for windows similar to as done for linux in https://github.com/llvm/torch-mlir-release/ however turns out the build process on windows is broken without the following fixes:

1. Building stablehlo for windows fails due to https://github.com/openxla/stablehlo/issues/1772 -- so disabling stablehlo in `build_windows_ci.sh` that will be used for building the python wheels. 

2. Add `USE_MATH_DEFINES` to resolve `torch-mlir\lib\Conversion\TorchOnnxToTorch\DefaultDomainGtoP.cpp(709): error C2065: 'M_LOG10E': undeclared identifier`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,10 @@ macro(torch_mlir_enable_werror)
   endif()
 endmacro()
 
+if(MSVC)
+  add_definitions(-D_USE_MATH_DEFINES)
+endif()
+
 #-------------------------------------------------------------------------------
 # Configure out-of-tree vs in-tree build
 #-------------------------------------------------------------------------------
--- a/build_tools/python_deploy/build_windows_ci.sh
+++ b/build_tools/python_deploy/build_windows_ci.sh
@@ -14,6 +14,7 @@ cmake -GNinja -Bbuild \
   -DLLVM_EXTERNAL_PROJECTS="torch-mlir" \
   -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR="$PWD" \
   -DPython3_EXECUTABLE="$(which python)" \
+  -DTORCH_MLIR_ENABLE_STABLEHLO=OFF \
   $GITHUB_WORKSPACE/externals/llvm-project/llvm
 
 cmake --build build --config Release


sahas3:
I don't have access to request reviews yet, so tagging @ScottTodd and @qingyunqu as potential reviewers. If you can take a look at this PR or suggest other reviewers, it'll be great. Thanks!

CMakeLists.txt
```diff
@@ -71,6 +71,10 @@ macro(torch_mlir_enable_werror)
   endif()
 endmacro()
 
+if(MSVC)
+  add_definitions(-D_USE_MATH_DEFINES)
+endif()
```
ScottTodd, line 76: > 2\. Add `USE_MATH_DEFINES` to resolve `torch-mlir\lib\Conversion\TorchOnnxToTorch\DefaultDomainGtoP.cpp(709): error C2065: 'M_LOG10E': undeclared identifier`

Thanks, I can reproduce that error locally on Windows without this patch. With the patch, I can build `torch-mlir-opt` again.

My team's downstream usage of torch-mlir (https://github.com/iree-org/iree/tree/main/compiler/plugins/input/Torch/torch-mlir) also still works with this since we use our own standalone CMake file. Sometimes these top level options in projects overwrite each other (or one project wants it but another doesn't).

build_tools/python_deploy/build_windows_ci.sh
```diff
@@ -14,6 +14,7 @@ cmake -GNinja -Bbuild \
   -DLLVM_EXTERNAL_PROJECTS="torch-mlir" \
   -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR="$PWD" \
   -DPython3_EXECUTABLE="$(which python)" \
+  -DTORCH_MLIR_ENABLE_STABLEHLO=OFF \
```
ScottTodd, line 17: It would be nice to get https://github.com/llvm/torch-mlir/blob/main/.github/workflows/ci.yml running Windows (and macOS) builds, given sufficiently powerful runners. (We're struggling with that in my project right now too). If the `pull_request` trigger is out of reach (due to slow builds / low runner capacity), nightly builds with the `schedule` trigger would be better than nothing.

I see Windows builds in https://github.com/llvm/torch-mlir/blob/main/.github/workflows/buildRelease.yml but that is only run manually and it hasn't been run in 9 months.

ScottTodd (approved):

ScottTodd:
Want me to merge this for you?

build_tools/python_deploy/build_windows_ci.sh
sahas3: I'm planning to submit a PR in https://github.com/llvm/torch-mlir-release to have Windows build running once this change is merged. The linux build is run nightly and I am planning to run the Windows on the same cadence as well. Eventually we want to have wheels for macOS too in the same fashion.

sahas3:
> Want me to merge this for you?

Yes, that'll be great. Thank you.

PR #3807, opened 2024-10-20
[TorchToLinalg] Use Op with native channel order for quantized conv2d

ubfx:
I've upstreamed the necessary quantized linalg Op with the "channel-first" ordering used by torch (https://github.com/llvm/llvm-project/pull/107740) for 2d convolution.

This patch changes the lowering for the quantized 2d case of `aten.convolution` accordingly, which saves three transpositions per convolution (input, weights, result) and therefore removes the requirement to try to optimize these away in downstream passes.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Linear.cpp
+++ b/lib/Conversion/TorchToLinalg/Linear.cpp
@@ -1125,53 +1125,56 @@ public:
     }
 
     if (numGroups == 1 && inputZp) {
-      // The quantized version uses a different channel ordering so we need to
-      // permute the tensors in order to use the existing path. We should
-      // eventually directly support this channel ordering.
-      llvm::SmallVector<int64_t> inPerms, weightPerms;
-      inPerms.push_back(0); // N stays at the front for input.
-      // Then we expect the spatial dimensions
-      for (size_t i = 0; i < numSpatialDims; ++i) {
-        inPerms.push_back(i + 2);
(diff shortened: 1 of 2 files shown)


zjgarvey (approved):
Nice.

ubfx:
I'm not merging this yet because I noticed that all of the Quantized Conv2D E2E tests are currently not properly evaluated (i.e. they're all in the XFAIL set), because the fx_importer frontend can't deal with it.

zjgarvey:
Are we no longer running the old importer path in the CI? You should at least be able to test locally with that path.

zjgarvey:
Yeah, try running this locally:

```
projects/pt1/tools/e2e_test.sh -f Conv2dQInt8Module_basic -v
```

It should print off IR for that example.

Then I'd check all the Conv2dQ tests with.

```
projects/pt1/tools/e2e_test.sh -f Conv2dQ -v
```

ubfx:
> Are we no longer running the old importer path in the CI?

No seems like only onnx and fx_importer are now run on CI.

> You should at least be able to test locally with that path.

Yeah I was able to test locally but it seems dangerous to not have the quantized tests done on CI at all. By now, I also retired the `jit_ir_importer` locally, though. Seems like this is a good opportunity to move the quantized tests over to the DQ/Q paradigm which we use everywhere else: https://github.com/llvm/torch-mlir/pull/3809

PR #3809, opened 2024-10-21
[fx] Fix importing and tests for quantized conv

ubfx:
The fx tracer does not support tracing "real" quantized tensors currently. A "real" quantized tensor here means a tensor that is created using a method like `torch.quantize_per_tensor()` and carries the quantization parameters (scale, zero_point, scheme) in the object.
However, it seems like the DQ-Q type fake quantizatation is now commonly used as a high level representation of quantized operators and is only lowered to native quantized ops (if available) in the respective hardware backend. Quantization of floating point modules in PyTorch is recently also performed as a graph transformation after exporting/tracing the original module.

```python
# Examples of "real"/native quantization
tens = torch.randint(-127, 127, (1,), dtype=torch.int8)
torch._make_per_tensor_quantized_tensor(tens, 1, 0)
# tensor([90.], size=(1,), dtype=torch.qint8,
#       quantization_scheme=torch.per_tensor_affine, scale=1.0, zero_point=0)

tens = torch.rand((1,))
torch.quantize_per_tensor(tens, 1, 0, torch.qint8)
# tensor([1.], size=(1,), dtype=torch.qint8,
#       quantization_scheme=torch.per_tensor_affine, scale=1.0, zero_point=0)

# Example of DQ/Q quantization
import torch.ao.quantization.fx._decomposed
tens = torch.rand((1,))
torch.ops.quantized_decomposed.quantize_per_tensor.default(tens, 1, 0, -128, 127, torch.int8)
# tensor([1], dtype=torch.int8)
```

This means that a typical import flow for a quantized network into/through torch-mlir would look like this:
`torch.export() -> quantization transformations on fx graph -> fx_importer` Where the tensors in the graph are normal float/int tensors and the quantization parameters are carried by the DQ/Q ops. These kinds of graphs can be traced without issues.

Currently, our quantized convolution tests use the "real" quantized tensors. This means that with the retirement of the `jit_ir_importer`, these tests cannot be imported any longer. In summary, I see no reason to stick to the "real" quantization in these tests, as both PyTorch 2.0 is using DQ/Q quantization and our linalg backend is also using it.

This patch updates our quantized convolution tests to use the DQ-Q quantization with the ops from `torch.ops.quantized_decomposed`.

Note: For future reference, there seems to be an ongoing consolidation of the ops for the DQ/Q scheme on the PyTorch side (https://github.com/pytorch/ao/issues/986#issuecomment-2390296826).


projects/pt1/python/torch_mlir_e2e_test/test_suite/conv.py
```diff
@@ -1183,23 +1183,26 @@ def ConvTbcModule_basic(module, tu: TestUtils):
...
-        bias = torch.quantize_per_tensor(bias, 0.0001, 0, torch.qint32)
-        bias = torch.dequantize(bias)
+    def _forward(self, input, weight, bias):
+        input = torch.ops.quantized_decomposed.dequantize_per_tensor.default(
+            input, 0.01, 7, -128, 127, torch.int8
+        )
+        weight = torch.ops.quantized_decomposed.dequantize_per_tensor.default(
+            weight, 0.01, 3, -128, 127, torch.int8
+        )
+        bias /= 0.01 * 0.01  # compensate for iact scale * weight scale
```
zjgarvey, line 1202: Since the bias is coming in non-quantized to begin with, I'm not sure I understand why it needs to be normalized by the accumulator scale.

zjgarvey (commented):
Maybe to make these tests most representative of models, would it make the most sense to have quantized bias and weight, and a f32 input tensor that gets quantized/dequantized?

projects/pt1/python/torch_mlir_e2e_test/test_suite/conv.py
ubfx: I found the bias issue: The problem is the size of the relative rounding error when choosing a float bias between 0 and 1 via `torch.rand()`. The issue is fixed when doing the `/= 0.01 * 0.01` because that increases the bias into the range of 1e4, such that the potential rounding error falls below the necessary threshold.

For this PR, I'm going with your suggestion of just using a quantized bias which sidesteps the problem. I might add some additional tests with an f32 bias in another PR, but that would require taking a closer look at the rounding mode and/or comparison tolerance.

zjgarvey: Sorry for the late reply. 

The bias rounding error makes sense.

The fake quantized models I've seen are usually stand in replacements for non-quantized versions, so the boundary types are non-quantized (inputs and outputs). Regardless, I'm fine with the tests being a true sample unit of `DQ -> op -> Q`.

zjgarvey:
Do you know if the quantization ops work with the pass `--torch-fuse-quantized-ops`? I'd like to see an IR dump of the linalg IR to make sure we are hitting the fusion (i.e., we should see a q conv linalg op).

projects/pt1/python/torch_mlir_e2e_test/test_suite/conv.py
ubfx: Ah I see what you mean. I implicitly assumed that our tests represent a single layer from a partitioned graph, because all our tests currently only contain a single operator. There, we would have integer boundary types between the layer partitions, but when assembling the whole model, there would be an additional DQ-Q pair around the composed layer partitions, such that the boundary types of the entire thing would be non-quantized again, as you said.

zjgarvey (approved):
LGTM. We need to be testing quantization path in CI. We should probably make the appropriate changes for quantized matmul variants as well in a follow-up patch.

projects/pt1/python/torch_mlir_e2e_test/test_suite/conv.py
zjgarvey: I think single layer is fine for these e2e tests.

PR #3811, opened 2024-10-22
[uncategorized_lowerings] Add lowering for torch.aten.round.decimals

meshtag:
Implement missing lowering for the op in a similar fashion as done by torch inductor. Also fix data movement and reduce op variants patterns to correctly handle explicitly declared legal ops.

Inductor decomposition ref: https://github.com/pytorch/pytorch/blob/main/torch/_inductor/decomposition.py#L223.


vivekkhandelwal1 (commented):
I am not able to understand why the lowering for the op `torch.aten.round.decimals` is added as a custom op lowering, why don't do it the right way?

Please follow the steps specified here https://github.com/llvm/torch-mlir/blob/main/docs/Torch-ops-E2E-implementation.md to add the op lowering.

vivekkhandelwal1:
> I am not able to understand why the lowering for the op `torch.aten.round.decimals` is added as a custom op lowering, why don't do it the right way?
> 
> Please follow the steps specified here https://github.com/llvm/torch-mlir/blob/main/docs/Torch-ops-E2E-implementation.md to add the op lowering.

Hi @meshtag, is this PR still of interest to you?

meshtag:
> Hi @meshtag, is this PR still of interest to you?

Hey @vivekkhandelwal1, looks like I missed this PR. Apologies for that. 

> I am not able to understand why the lowering for the op torch.aten.round.decimals is added as a custom op lowering, why don't do it the right way?

AFAICT, we have two options here: 
1. Get this op directly in the mlir-world and then lower it from there. 
2. Register this op for decomposition (it already exists in torch, as pointed above) and then decompose it at a higher level (ideally before the fx->mlir translation layer) and then directly handle the decomposed op (which is already supported).

This PR tries to go via path 1. I am not sure if we should discard path 2 though, can you please share your thoughts on this. 

> Please follow the steps specified here https://github.com/llvm/torch-mlir/blob/main/docs/Torch-ops-E2E-implementation.md to add the op lowering.

Sure, I can do this (if we want to go via this path). Thanks for pointing it out.

vivekkhandelwal1:
> > I am not able to understand why the lowering for the op torch.aten.round.decimals is added as a custom op lowering, why don't do it the right way?
> 
> AFAICT, we have two options here:
> 
> 1. Get this op directly in the mlir-world and then lower it from there.
> 2. Register this op for decomposition (it already exists in torch, as pointed above) and then decompose it at a higher level (ideally before the fx->mlir translation layer) and then directly handle the decomposed op (which is already supported).
> 
> This PR tries to go via path 1. I am not sure if we should discard path 2 though, can you please share your thoughts on this.

If you want to go thorough the path 1, then I don't think you have to do much here. Just [register](https://github.com/llvm/torch-mlir/blob/main/projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/torch_ods_gen.py )this op in Torch-MLIR, and the same lowering which you have written can be used. Also, you would be able to test the correctness of your lowering e2e.

PR #3812, opened 2024-10-22
Only unroll prim loop-like ops within a `torch.shape.calculate` region

zjgarvey:
Trying this out in the CI to see if this affects any e2e tests.


vivekkhandelwal1:
Hi @zjgarvey, I'm merging this patch since it's needed for the Torch-MLIR bump in IREE.

PR #3819, opened 2024-10-25
Add onnx GridSample support for border padding mode

Ax9D:
* Adds support for padding_mode="border" 

* Clamps the grid coordinates between 0 and size - 1 like this when using this padding mode:

```
x_result = min(max(0, x), W - 1)
y_result = min(max(0, y), H - 1)
```
* Added Lit tests for both TorchToLinalg and OnnxToTorch lowerings


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -157,7 +164,8 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
 
         Value paddingMode = rewriter.create<Torch::ConstantIntOp>(
             binder.getLoc(), rewriter.getType<Torch::IntType>(),
-            rewriter.getIntegerAttr(rewriter.getIntegerType(64), 0));
+            rewriter.getIntegerAttr(rewriter.getIntegerType(64),
```
zjgarvey, line 167: I think it is simpler as:

```cpp
Value paddingMode = rewriter.create<Torch::ConstantIntOp>(binder.getLoc(), paddingModeInt);
```

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2565,10 +2565,29 @@ class ConvertAtenGridSamplerOp : public OpConversionPattern<AtenGridSamplerOp> {
...
+                            Value SizeSubOne) -> Value {
+      Value xMaxZero = b.create<arith::MaximumFOp>(loc, x, zeroFloat);
+      return b.create<arith::MinimumFOp>(loc, xMaxZero, SizeSubOne);
+    };
+
+    auto lambdaPadding = [&](OpBuilder &b, Location loc, Value paddingMode,
+                             Value x, Value SizeSubOne) -> Value {
+      Value border = lambdaBorder(b, loc, x, SizeSubOne);
+      Value zeroInt =
+          b.create<arith::ConstantOp>(loc, b.getIntegerAttr(int64type, 0));
```
zjgarvey, line 2578: It might be better to keep this constant outside the lambda function, and just get it from the capture (otherwise you will re-write the constant multiple times in the IR).

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2565,10 +2565,29 @@ class ConvertAtenGridSamplerOp : public OpConversionPattern<AtenGridSamplerOp> {
...
+      Value isZero = b.create<arith::CmpIOp>(loc, arith::CmpIPredicate::eq,
+                                             paddingMode, zeroInt);
```
zjgarvey, line 2580: The padding mode is always a constant (coming from ONNX, at least). We should be able to get the `paddingModeInt` by `matchPattern(paddingMode, m_TorchConstantInt(&paddingModeInt))`. This way, we can avoid writing IR to check conditions that can be determined statically at compile time. Since the select op and cmp op will likely get folded anyway, this is not hugely problematic, but on principal I'd prefer to generate the simplest possible IR from the get-go.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -2565,10 +2565,29 @@ class ConvertAtenGridSamplerOp : public OpConversionPattern<AtenGridSamplerOp> {
...
     };
 
+    auto lambdaBorder = [&](OpBuilder &b, Location loc, Value x,
+                            Value SizeSubOne) -> Value {
+      Value xMaxZero = b.create<arith::MaximumFOp>(loc, x, zeroFloat);
+      return b.create<arith::MinimumFOp>(loc, xMaxZero, SizeSubOne);
+    };
+
+    auto lambdaPadding = [&](OpBuilder &b, Location loc, Value paddingMode,
+                             Value x, Value SizeSubOne) -> Value {
```
zjgarvey, line 2575: `x` and `sizeSubOne` should be the only inputs for both of these lambda functions. The others can be gotten through the default capture. Just be sure to define these functions after you introduce the other arguments.

zjgarvey (changes requested):
Thanks for adding this support!

I have a couple of changes to request, and I'd also like you to sync your branch with head of main to avoid the failing CI.

Ax9D:
Thanks for the feedback! I'll work on the requested changes.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
Ax9D: If we want to capture the other arguments, the lambdas will have to be defined inside the GenericOp lambda later on the code, but I would want to avoid that as it would clutter the implementation there. There are a few other helper lambdas such as `lambdaInterpolate` and `lambdaExtract` which take arguments like this, so I decided to make it similar to those.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
Ax9D: Yes, I'm evaluating it at compile time now.

vivekkhandelwal1:
@Ax9D Why did you close this PR and open a new one here: https://github.com/llvm/torch-mlir/pull/4186?

The changes are not so large that you face an issue in rebasing this PR. Please reopen this and close the other one. It becomes hard to follow the review comments.

Ax9D:
I completely agree with you, but I accidentally removed the branch and the PR was automatically closed, and I can't seem to open it again, even though the branch is restored...

vivekkhandelwal1:
> I completely agree with you, but I accidentally removed the branch and the PR was automatically closed, and I can't seem to open it again, even though the branch is restored.... Is it due to the new one being open?

I think that's happening probably because the link between your past and the restored one is lost.

Ax9D:
Hey, I reopened the other one. Sorry for the inconvenience.

PR #3820, opened 2024-10-28
[linalg] Fix torch.aten.add of `torch.bool`

rsuderman:
Addition of bools saturate which equates to an `or` operator. Updated to avoid some noticed downstream failures.


zjgarvey (approved):
Makes sense.

zjgarvey:
Should mark the test as an xfail for `fx_importer_stablehlo`.

PR #3825, opened 2024-10-29
Augmented calls to yaml.load to use the safe loader.

TimAtGoogle:


TimAtGoogle:
Accidentally merged with other commits that aren't mine...  I'll just start over.

PR #3826, opened 2024-10-29
[bazel] Fix missing dependency in the build

rsuderman:


dan-garvey (approved):

sjain-stanford:
Thanks for sending this @rsuderman ! I was about to pick this up and was pleasantly surprised see it fixed already!

PR #3829, opened 2024-10-30
[stablehlo] fix: enhance torch's index-like op lowering to stablehlo's gather/scatter

Vremold:
In torch.index_put like ops, `values` is only required to be broadcastable to `input[indices]`, rather than exact dimension match. This patch fixes the problem by add additional stablehlo.dynamic_broadcast_in_dim before creating stablehlo.scatter op. BTW, this patch also enhance the `getBroadcastResultShape` utility in hlo namespace.


qingyunqu (approved):
LGTM

PR #3830, opened 2024-10-30
build: manually update PyTorch version and fix CI failure

vivekkhandelwal1:
This commit sets the PyTorch and TorchVision version to nightly release 2024-10-29.

This commit also fixes the CI failure after this commit https://github.com/llvm/torch-mlir/commit/54d9e2401376e7eb2c6c219e3b3555f45f8b2635 got merged. The issue was that the CI checks in the PR were run before the previous roll pytorch update but the PR was actually merged after the roll pytorch update. Hence, the failure was not caught before merging the PR. 

While exporting the fx_graph through fx_importer for `rrelu` and `rrelu_with_noise` op for train mode, it decomposes the `aten.rrelu_with_noise` op based on the PyTorch decomposition which is the default behavior. However, the decomposition contains an input mutation specifically here https://github.com/pytorch/pytorch/blob/9bbe4a67ad137032add6a3b0b74bda66f5ef83d2/torch/_decomp/decompositions.py#L325, resulting in the runtime failure. This issue would probably be fixed by https://github.com/pytorch/pytorch/pull/138503. Until then, the failing tests are added to the xfail set.

Also, after the roll pytorch update following tests started passing for fx_importer, and fx_importer_stablehlo config.

- "ElementwiseRreluTrainModule_basic"
- "ElementwiseRreluTrainStaticModule_basic"
- "ElementwiseRreluWithNoiseTrainModule_basic"
- "ElementwiseRreluWithNoiseTrainStaticModule_basic"

This commit also updates the dtype check for the `aten.linear` op since the op now expects both the input tensors to have the same dtype.


pashu123 (approved):
LGTM!

Max191 (approved):
Awesome, thanks Vivek!

PR #3833, opened 2024-10-30
[Torch] support AtenNllLossForwardOp decomposition

yyp0:


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -9080,6 +9095,226 @@ class DecomposeAtenCrossEntropyLossOp
 };
 } // namespace
 
+namespace {
+class DecomposeAtenNllLossForwardOp
```
qingyunqu, line 9099: add some comments to describe what the decomposition do?

qingyunqu (approved):
LGTM

PR #3838, opened 2024-10-30
add scalarization patterns to support dynamic pytorch pad exports

zjgarvey:
1. Adds case handling for `aten.slice.tensor` shape inference with negative strides. This is not technically allowed by native pytorch, but it is useful for ONNX ingest. We were getting some incorrect shapes for these negative strided slice ops. 
2. Adds scalarization support for ops seen in pytorch pad exports to ONNX. These are typically `aten.view` `aten.transpose.int` and `aten.slice.Tensor` with negative strides (and rank 2). 
3. Allows view op `self` to be added to the worklist conditionally, based on whether the view op actually occurs as a middle point in a shape computation.


lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -1065,6 +1215,34 @@ bool isAnchorOp(Operation *op) {
...
+  // function call.
+  if (!isa<mlir::IntegerType>(resultTy.getDtype()))
+    return true;
+  // check uses of the view op.
+  // If the view op has a use in our worklist, then it needs to be scalarized.
+  for (OpOperand &use : op->getUses()) {
+    Operation *userOp = use.getOwner();
+    if (workList.contains(userOp))
+      return false;
+  }
```
IanWood1, line 1241: I'm a bit confused why this is needed, shouldn't the `root` `AtenViewOp` be omitted from the worklist instead?

test/Dialect/Torch/scalarize-shapes.mlir
```diff
@@ -374,3 +374,172 @@ func.func @squeeze_dim_full_fold(%arg0: !torch.vtensor<[?,?],si64>) -> !torch.li
...
+    %int2 = torch.constant.int 2
+    %int1 = torch.constant.int 1
+    %int0 = torch.constant.int 0
+    %3 = torch.aten._shape_as_tensor %arg0 : !torch.vtensor<[?,144,?,?],f32> -> !torch.vtensor<[4],si64>
+    %4 = torch.prim.ListConstruct %3, %0 : (!torch.vtensor<[4],si64>, !torch.vtensor<[4],si64>) -> !torch.list<vtensor>
+    %5 = torch.aten.cat %4, %int0 : !torch.list<vtensor>, !torch.int -> !torch.vtensor<[8],si64>
+    %6 = torch.prim.ListConstruct %int-1, %int2 : (!torch.int, !torch.int) -> !torch.list<int>
+    %7 = torch.aten.view %5, %6 : !torch.vtensor<[8],si64>, !torch.list<int> -> !torch.vtensor<[4,2],si64>
+    %8 = torch.aten.slice.Tensor %7, %int0, %int-1, %int-9223372036854775807, %int-1 : !torch.vtensor<[4,2],si64>, !torch.int, !torch.int, !torch.int, !torch.int -> !torch.vtensor<[4,2],si64>
+    %9 = torch.aten.transpose.int %8, %int0, %int1 : !torch.vtensor<[4,2],si64>, !torch.int, !torch.int -> !torch.vtensor<[2,4],si64>
```
IanWood1, line 412: It might be useful to add some smaller unit tests for the future if anything breaks. Especially for the transposes, since the `torch.aten.transpose`s here only test a small subset of the logic

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -424,6 +457,102 @@ class PropagateAtenSliceTensorPattern
...
+    int64_t rank = selfSizes.size();
+
+    int64_t frontDP = 1, dim0L = 1, midDP = 1, dim1L = 1, backDP = 1;
+    for (int64_t i = 0; i < rank; i++) {
+      if (i < dim0) {
+        frontDP *= selfSizes[i];
+        continue;
+      }
+      if (i == dim0) {
+        dim0L *= selfSizes[i];
```
IanWood1, line 495:
```suggestion
        dim0L = selfSizes[i];
```

& could also be moved out of the loop. Makes it more clear that dim0L isnt accumulating

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -424,6 +457,102 @@ class PropagateAtenSliceTensorPattern
...
+
+    ArrayRef<int64_t> selfSizes = selfTy.getSizes();
+    int64_t rank = selfSizes.size();
+
+    int64_t frontDP = 1, dim0L = 1, midDP = 1, dim1L = 1, backDP = 1;
+    for (int64_t i = 0; i < rank; i++) {
+      if (i < dim0) {
+        frontDP *= selfSizes[i];
+        continue;
+      }
```
IanWood1, line 493: `else if` instead of `continue` is probably more clear here

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
zjgarvey: I need to add view ops in general since I want to apply the canonicalization pattern to all view ops in the IR. In the near future, I think we should move the canonicalization pattern out of this pass, since handling these somewhat unrelated view ops is annoying.

test/Dialect/Torch/scalarize-shapes.mlir
zjgarvey: Yeah, because of the anchor condition, the lit tests are annoying to write, but I totally agree. I'll add a few more tests, specifically for transpose and slice.

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
zjgarvey: Yeah, I felt like this whole section was a bit ugly. I'll clean it up.

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -379,39 +385,67 @@ class PropagateAtenSliceTensorPattern
...
-    auto selfTy = cast<BaseTensorType>(op.getSelf().getType());
     auto selfShape = selfTy.getSizes();
+    auto resultShape = resultTy.getSizes();
     int64_t selfRank = selfShape.size();
 
     // Correct for negative indexing:
     dim = dim < 0 ? dim + selfRank : dim;
 
-    int64_t dimLength = elements.size();
+    int64_t dimLength = selfShape[dim];
```
IanWood1, line 395: Could you use `toPositiveDim` + `isValidDim` here as well (also similar problem with PropagateIndexSelectPattern)? 

Since this pass is matching against constants not known during verification, a lot of UB in the input MLIR will go undetected and could cause compiler crashes. So, it probably makes sense to have strict checks whenever there is a `matchPattern`

test/Dialect/Torch/scalarize-shapes.mlir
IanWood1: Thank you. Also, nice cleanup for the transpose pattern, much more readable 😄

IanWood1 (approved):

PR #3839, opened 2024-10-31
[Torch] fix cross_entropy_loss decomposition

yyp0:


lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -8925,10 +8931,19 @@ class DecomposeAtenCrossEntropyLossOp
...
         loc, self.getType(), self, dim, /*dtype=*/noneVal);
+
+    Type secondType;
+    if (reductionInt == 0) {
+      secondType = target.getType();
+    } else {
+      auto targetType = dyn_cast<BaseTensorType>(target.getType());
+      secondType = targetType.getWithSizesAndDtype({}, targetType.getDtype());
+    }
+
```
vivekkhandelwal1, line 8942:
```suggestion

    Type secondType = target.getType();
    if (reductionInt != 0) {
      auto targetType = dyn_cast<BaseTensorType>(target.getType());
      secondType = targetType.getWithSizesAndDtype({}, targetType.getDtype());
    }
```

vivekkhandelwal1 (changes requested):

PR #3843, opened 2024-10-31
Fix Slice Folder OOB Crash and onnx.Shape lowering

zjgarvey:
1. Clamps OOB start index to 0 in slice folder
2. Adds a more descriptive `emitError` in slice folder if the creation of the `DenseElementsAttr` would fail due to a bad result shape.
3. Fixes the `onnx.Shape` lowering to default to `inputRank` for `end` instead of `-1`. When `end==-1` the last element was missing when slicing.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -3998,6 +3998,7 @@ OpFoldResult AtenSliceTensorOp::fold(FoldAdaptor adaptor) {
     int64_t limit = end.getValue().getSExtValue();
     int64_t stride = step.getValue().getSExtValue();
     begin = begin < 0 ? begin + inType.getSizes()[dimInt] : begin;
+    begin = std::max(begin, (int64_t)0);
     limit = limit < 0 ? limit + inType.getSizes()[dimInt] : limit;
     limit = limit < 0 ? -1 : limit;
     limit = std::min(limit, inType.getSizes()[dimInt]);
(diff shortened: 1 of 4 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -3998,6 +3998,7 @@ OpFoldResult AtenSliceTensorOp::fold(FoldAdaptor adaptor) {
     int64_t limit = end.getValue().getSExtValue();
     int64_t stride = step.getValue().getSExtValue();
     begin = begin < 0 ? begin + inType.getSizes()[dimInt] : begin;
+    begin = std::max(begin, (int64_t)0);
```
rsuderman, line 4001: Nit: you can do `std::max<int64_t>(begin, 0)` I think

lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -4038,6 +4039,14 @@ OpFoldResult AtenSliceTensorOp::fold(FoldAdaptor adaptor) {
       }
     };
     recursiveIter(recursiveIter, 0, 0);
+    if ((int64_t)values.size() != count) {
```
rsuderman, line 4042: Try to use `static_cast` so it should be

```
    if (static_cast<int64_t>(values.size()) != count) {
```

rsuderman (approved):

vinayakdsci (approved):
LGTM!

lib/Dialect/Torch/IR/TorchOps.cpp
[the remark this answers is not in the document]
vinayakdsci: This is resolved, thanks!

PR #3846, opened 2024-11-01
Fix onnx.If lowering with scalar condition tensor

jinchen62:
Fixes https://github.com/nod-ai/SHARK-ModelDev/issues/696#issuecomment-2442016530


zjgarvey (approved):
Might be good to add a simple lit test, but this is pretty self-explanatory so I'm fine if we omit it.

PR #3848, opened 2024-11-01
[TOSA] Add legalization for aten.as_strided

justin-ngo-arm:
- Add Torch to TOSA legalization for aten.as_strided op
- Update xfail_sets with the following:
  + New aten.as_strided results
  + Changes from this commit: https://github.com/llvm/torch-mlir/commit/7f9f99c6f8c84323d896b47fcd67c4bc668f6577
  + Failed tests from new PyTorch version update
- Add new LIT test to basic.mlir

Change-Id: I6f471ea116ca47f2bf9537b62950fce75a2c624f

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -6778,6 +6778,108 @@ LogicalResult ConvertAtenOp<AtenThresholdBackwardOp>::matchAndRewrite(
   return success();
 }
 
+// Legalization for aten.as_strided
+template <>
+LogicalResult ConvertAtenOp<AtenAsStridedOp>::matchAndRewrite(
+    AtenAsStridedOp op, OpAdaptor adaptor,
+    ConversionPatternRewriter &rewriter) const {
+  // To lower aten.as_strided to TOSA, we will first reshape the input tensor to
+  // an 1-D tensor, then calculate the indices of result elements based on the
+  // output size, stride and storage offset. With the reshaped 1-D tensor and
+  // the indices, we can apply Gather to extract the required elements into a
+  // new tensor and then reshape it back to the desired output shape.
+  auto self = adaptor.getSelf();
+
+  // Not a tensor type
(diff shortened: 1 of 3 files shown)


sjarus (approved):
Looks good. I'll wait a few hours for @sahas3 and then land if there's no comment.

sahas3 (approved):
LGTM.

arnab-oss:
@justin-ngo-arm in this lowering I see  aten.as_strided being lowered to tosa.gather among other ops. My question is as_strided is a view-like op, but tosa.gather is not. 
So if 
```
%2 = torch.aten.as_strided %arg0, %0, %1, %none : !torch.vtensor<[5,5],f32>, !torch.list<int>, !torch.list<int>, !torch.none -> !torch.vtensor<[3,3],f32>
```
%2 and %arg0 would use the same memory (they are aliases in other words). But the lowering violates that I think. 
Can you please let me know if I am missing anything here?
CC @sjarus @sahas3

sjarus:
> @justin-ngo-arm in this lowering I see aten.as_strided being lowered to tosa.gather among other ops. My question is as_strided is a view-like op, but tosa.gather is not. So if
> 
> ```
> %2 = torch.aten.as_strided %arg0, %0, %1, %none : !torch.vtensor<[5,5],f32>, !torch.list<int>, !torch.list<int>, !torch.none -> !torch.vtensor<[3,3],f32>
> ```
> 
> %2 and %arg0 would use the same memory (they are aliases in other words). But the lowering violates that I think. Can you please let me know if I am missing anything here? CC @sjarus @sahas3 @ramiro050

That's a good question. PyTorch leverages a number of in-place and view semantic constructs that don't necessarily have exact equivalents in TOSA, which does not have the rich interpretive view semantics that PyTorch often leverages. 

It would fall upon the backend to recognize and analyze such behaviors out. Another such example is aten.broadcast_to that is materialized as a tosa.tile. 

We're looking at ways to better express these post the 1.0 release of TOSA that is in the process of being executed now.

PR #3849, opened 2024-11-01
Revert "[MLIR][TORCH] Only unroll prim loop-like ops within a torch.shape.calculate region"

AmosLewis:
Issue found when bump torch-mlir in iree https://github.com/iree-org/iree/pull/18992. This revert can fix the issue https://github.com/iree-org/iree/pull/18992#issuecomment-2452628846 when bump.
This reverts commit 55ff110dc29cab7e2495ccdbec9a60512c29c665.
 @Max191  has done this in last bump https://github.com/iree-org/torch-mlir/commits/integrates/llvm-20241024/ , but just in iree-org/torch-mlir, the revert patch has not been added in llvm/torch-mlir.

### DIFF
--- a/lib/Dialect/Torch/Transforms/SimplifyAbstractInterpCalculationsUtils.cpp
+++ b/lib/Dialect/Torch/Transforms/SimplifyAbstractInterpCalculationsUtils.cpp
@@ -32,6 +32,9 @@ public:
 } // namespace
 
 namespace {
+// TODO: Only unroll inside the shape calculation region.
+// Maybe do this by only applying patterns and folding greedily on the ops
+// inside the region + the shape.calculate op itself?
 class FullyUnrollPrimLoopOp : public OpRewritePattern<PrimLoopOp> {
 public:
   using OpRewritePattern::OpRewritePattern;
@@ -39,12 +42,6 @@ public:
                                 PatternRewriter &rewriter) const override {
     Location loc = op->getLoc();
     MLIRContext *context = op->getContext();
-    // Only unroll loops if they are contained in a shape calculate region.
-    Region *region = op->getParentRegion();
-    Operation *parentOp = region->getParentOp();
-    if (!parentOp || !isa<Torch::ShapeCalculateOp>(parentOp))
-      return rewriter.notifyMatchFailure(
-          op, "Loop is not contained in a shape calculation region.");
     if (!op.isForLike())
       return rewriter.notifyMatchFailure(op, "Loop is not for-like");
     int64_t maxTripCount;
--- a/test/Dialect/Torch/simplify-shape-calculations.mlir
+++ b/test/Dialect/Torch/simplify-shape-calculations.mlir
@@ -152,23 +152,6 @@ func.func @fully_unroll_prim_loop$no_unroll(%arg0: !torch.vtensor, %arg1: !torch
   return %0 : !torch.vtensor
 }
 
-// CHECK-LABEL:   func.func @fully_unroll_prim_loop$outside_region(
-// CHECK:         %[[LOOP:.*]] = torch.prim.Loop
-func.func @fully_unroll_prim_loop$outside_region(%arg0: !torch.vtensor, %arg1: !torch.list<int>, %arg2: !torch.int) -> !torch.vtensor {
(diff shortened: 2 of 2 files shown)


zjgarvey:
What is the issue?

jinchen62 (approved):
LGTM. Might need @zjgarvey to take a look.

AmosLewis:
> What is the issue?

```
model.mlir:4:10: error: failed to legalize operation 'arith.sitofp' that was explicitly marked illegal
    %0 = torch.operator "onnx.TfIdfVectorizer"(%arg0) {torch.onnx.max_gram_length = 2 : si64, torch.onnx.max_skip_count = 0 : si64, torch.onnx.min_gram_length = 2 : si64, torch.onnx.mode = "TF", torch.onnx.ngram_counts = [0 : si64, 4 : si64], torch.onnx.ngram_indexes = [0 : si64, 1 : si64, 2 : si64, 3 : si64, 4 : si64, 5 : si64, 6 : si64], torch.onnx.pool_int64s = [2 : si64, 3 : si64, 5 : si64, 4 : si64, 5 : si64, 6 : si64, 7 : si64, 8 : si64, 6 : si64, 7 : si64]} : (!torch.vtensor<[2,6],si32>) -> !torch.vtensor<[2,7],f32> 
         ^
model.mlir:4:10: note: see current operation: %327 = "arith.sitofp"(<<UNKNOWN SSA VALUE>>) : (i64) -> f64
model.mlir:1:1: error: conversion to vm.module failed
```

zjgarvey:
Yeah let me take a closer look. It doesn't make sense to revert this patch.

zjgarvey:
I'm going to say we should not be reverting this. 

If IREE is unable to manage something about scalar `scf.for` loops, then we should make a change in IREE to address this, rather than going back to indiscriminately unrolling IR loops during a pass that is supposed to perform shape inference in torch-mlir.

Alternatively, if something is wrong with TFIDF, we should be resolving that instead of reverting this patch.

Max posted an issue in IREE to track <https://github.com/iree-org/iree/issues/18961>. I suggest we mark those three tests batched TFIDF tests as failures (regressions) in IREE. Then we can actually get someone to resolve it. Or not, since I've yet to see TFIDF show up in a single model. 

@ScottTodd @Max191 Do you guys have any strong opinions on this?

Max191:
> I'm going to say we should not be reverting this.
> 
> If IREE is unable to manage something about scalar `scf.for` loops, then we should make a change in IREE to address this, rather than going back to indiscriminately unrolling IR loops during a pass that is supposed to perform shape inference in torch-mlir.

I would agree with this. Indiscriminately unrolling all loops does not seem like something we want to be doing. I haven't looked much into the issue, but my guess is that it needs to be fixed in IREE.

> Alternatively, if something is wrong with TFIDF, we should be resolving that instead of reverting this patch.
> 
> Max posted an issue in IREE to track [iree-org/iree#18961](https://github.com/iree-org/iree/issues/18961). I suggest we mark those three tests batched TFIDF tests as failures (regressions) in IREE. Then we can actually get someone to resolve it. Or not, since I've yet to see TFIDF show up in a single model.
> 
> @ScottTodd @Max191 Do you guys have any strong opinions on this?

As far as reverting the PR goes, I don't think we should revert it here in torch-mlir. We can carry the revert in IREE until the bug is resolved. If it ends up turning into a lot of work, and we are worried about carrying the revert for too long, then we can always add some flags/options in torch-mlir, and turn it off in IREE until the issue is resolved. But I doubt we will need to do that, and we can probably just carry the revert until it is fixed.

Max191:
We will need someone to pick up that issue, though. I'll try to signal and see if anyone can pick it up.

ScottTodd:
Sorry it took so long to chime in here.

In general I'd look at the guidance set forth in https://llvm.org/docs/DeveloperPolicy.html#patch-reversion-policy.

Acknowledging that many of those on this comment thread contribute both here and in the downstream IREE project... The burden is typically on downstream users to demonstrate that an upstream patch is unsound. Upstream should focus on what makes sense upstream, then downstream adapts. It sounds like there is consensus here that IREE should adapt in some way, and carrying a local revert while that work is done is reasonable.

PR #3856, opened 2024-11-07
Generalize `aten.view` pattern in scalarize shapes

zjgarvey:
Extends the existing pattern to allow finding matching dims from the back as well as the front.


lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -1099,97 +1099,111 @@ class CanonicalizeAtenViewPattern : public OpRewritePattern<AtenViewOp> {
     int64_t outRank = resultTy.getSizes().size();
 
     SmallVector<int64_t> sizes(selfTy.getSizes());
-    int64_t endMatchingDim = -1;
-    // input sizes vs. provided view sizes comparison loop
-    for (int64_t i = 0; i < std::min(outRank, inRank); i++) {
+    int64_t leftMatchEnd = -1;
+    // compare input sizes with provided dims from left
+    while (++leftMatchEnd < std::min(outRank, inRank)) {
```
rsuderman, line 1104: Better to still use a `for` loop

```
int64_t leftMatchEnd = 0;
for (; leftMatchEnd < std::min(outRank, inRank); leftMatchEnd++) {
```

lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
```diff
@@ -1099,97 +1099,111 @@ class CanonicalizeAtenViewPattern : public OpRewritePattern<AtenViewOp> {
...
+      int64_t dim;
+      // if the dim of the size int op doesn't match, fail
+      if (!matchPattern(sizeIntOp.getDim(), m_TorchConstantInt(&dim)) ||
+          dim != leftMatchEnd)
         break;
+    }
+
+    int64_t rightMatchEnd = -1;
+    // compare input sizes with provided dims from right
+    while (++rightMatchEnd < std::min(outRank, inRank) - leftMatchEnd) {
```
rsuderman, line 1132: See `for` comment above

rsuderman (approved):

PR #3860, opened 2024-11-08
[TOSA] Add promote type to unary ops and aten.cat lowering

justin-ngo-arm:
Change-Id: I2699bf9007723fe629edb1c524c10ef8142e0234


sjarus (approved):
Looks good to me . Going ahead and merging since this is a simple contained change.

PR #3869, opened 2024-11-13
[ONNX] Remove kernel shape and weight shape equivalence check from Onnx.Conv lowering

vivekkhandelwal1:
This commit removes the equivalence check for kernel shape and weight shape from the Onnx.conv lowering since those checks seem to be of no use (not sure why were they part of the lowering in the first place).


zjgarvey:
So I'm a bit curious about a few things:

1. Why do we need to check if the provided kernel shape matches in the first place? The kernel shape is ... what the kernel shape is?
2. What models did this come up in? What motivated this patch?

vivekkhandelwal1:
Hi @zjgarvey, the PR is updated. Can you please review it now?

zjgarvey (approved):

PR #3870, opened 2024-11-13
[ONNX] rework some reduction op lowerings

zjgarvey:
- Refactors more "onnx.ReduceXXX" patterns through helper function.
- Fixes bug with iterating unconditionally on `output_dim == 1` during `dimList` inference.

This change results in passes for the following 11 models:

crossvit_15_240
crossvit_15_dagger_240
crossvit_15_dagger_408
crossvit_18_240
crossvit_18_dagger_240
crossvit_18_dagger_408
crossvit_9_240
crossvit_9_dagger_240
crossvit_base_240
crossvit_small_240
crossvit_tiny_240

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -1066,26 +1066,25 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             binder.op, resultType, operand, vAlpha, vScale, vInputScale);
         return success();
       });
-  patterns.onOp("ReduceL1", 1,
-                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-                  Torch::ValueTensorType resultType;
-                  int64_t keepDims, noop_with_empty_axes;
-                  Value operand;
-                  if (binder.tensorOperandAtIndex(operand, 0) ||
-                      binder.tensorResultType(resultType) ||
-                      binder.s64IntegerAttr(keepDims, "keepdims", 1) ||
-                      binder.s64IntegerAttr(noop_with_empty_axes,
-                                            "noop_with_empty_axes", 0))
-                    return failure();
+  patterns.onOp(
+      "ReduceL1", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
+        Torch::ValueTensorType resultType;
+        int64_t keepDims, noop_with_empty_axes;
+        Value operand;
+        if (binder.tensorOperandAtIndex(operand, 0) ||
+            binder.tensorResultType(resultType) ||
+            binder.s64IntegerAttr(keepDims, "keepdims", 1) ||
+            binder.s64IntegerAttr(noop_with_empty_axes, "noop_with_empty_axes",
+                                  0))
+          return failure();
 
-                  Value data = rewriter.create<Torch::AtenAbsOp>(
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -117,38 +122,60 @@ LogicalResult reducedSumImpl(OpBinder binder,
...
+  }
+
   Value dimValueList = rewriter.create<Torch::PrimListConstructOp>(
       binder.getLoc(),
       Torch::ListType::get(Torch::IntType::get(binder.op->getContext())),
       axesList);
   Value keepDimBool =
       rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), keepDims);
-  Value dType = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
   // If we are using the ReducedSum as an intermediate op to be passed into
```
vinayakdsci, line 165:
```suggestion
  // If we are using the reduction op as an intermediate op to be passed into
```

As we are no longer reducing only sums, `ReducedSum` can be changed to `reduction op`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -117,38 +122,60 @@ LogicalResult reducedSumImpl(OpBinder binder,
...
+      return rewriter.notifyMatchFailure(
+          binder.op,
+          "no axes provided & no keepdim: expected result to be rank zero.");
+    if (keepDims == 1 &&
+        (resultType.getSizes().size() != inputType.getSizes().size() ||
+         llvm::any_of(resultType.getSizes(),
+                      [](int64_t size) { return size != 1; })))
+      return rewriter.notifyMatchFailure(
+          binder.op,
+          "no axes provided & keepdim: expected result to have 1 element.");
```
vinayakdsci, line 152:
```suggestion
          "no axes provided & keepdim: expected result to have all dimensions equal to 1.");
```

As in the if condition itself, result might not have just one element. 

Also nit: is it possible to also include that the shape should be the same as the input tensor? It might fail for that case too.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -77,32 +80,34 @@ LogicalResult reducedSumImpl(OpBinder binder,
...
+          for (auto i : reduceDims) {
+            axesList.push_back(rewriter.create<Torch::ConstantIntOp>(
+                binder.getLoc(), rewriter.getI64IntegerAttr(i)));
+          }
+        } else
+          binder.op->emitWarning(
+              "Number of inferred reduce dims, " +
+              std::to_string(reduceDims.size()) +
+              ", does not match the provided number of axes, " +
+              std::to_string(numAxes) + ".");
```
vivekkhandelwal1, line 97: Shouldn't we error out in this case?

vivekkhandelwal1 (approved):
Just a small doubt, except that everything else looks good.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
zjgarvey: Since the `axesList` will be empty, it will default to the next path below. I figured it would be better to return a warning that the above logic failed and grab the original axes from `axesVal`.

PR #3872, opened 2024-11-14
[TorchToLinalg] Use `linalg.transpose` instead of `generic` in `permuteTensor`

CoTinker:
This PR changes the lowering to use `linalg.transpose` instead of `linalg.generic` in `torch_to_linalg::permuteTensor`.

### DIFF
--- a/lib/Conversion/TorchToLinalg/Utils.cpp
+++ b/lib/Conversion/TorchToLinalg/Utils.cpp
@@ -578,6 +578,12 @@ LogicalResult torch_to_linalg::permuteTensor(Operation *op,
   int64_t inputRank = inType.getRank();
   Type elementType = inType.getElementType();
 
+  // Check for 0-D tensor.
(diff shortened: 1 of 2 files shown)


ubfx (commented):
I think this makes sense, but there seems to be a problem with 0D-tensors in the e2e tests?

CoTinker:
I'll take a look at it.

CoTinker:
I wonder if we should add a folder for `aten.permute` with 0-D input.

vivekkhandelwal1:
> I wonder if we should add a folder for `aten.permute` with 0-D input.

Yeah, that's a nice thing to do. You may do it in the same PR or in another one as you wish.

vivekkhandelwal1 (approved):
LGTM

CoTinker:
Thanks, I think it's better to open another PR for the folder.

PR #3877, opened 2024-11-15
[ONNX] reduce min opset versions for Greater and Gather

zjgarvey:
No appreciable difference between the lower opset version and the newer one. See
<https://onnx.ai/onnx/operators/onnx__Greater.html> and <https://onnx.ai/onnx/operators/onnx__Gather.html>


zjgarvey:
Resolves <https://github.com/nod-ai/SHARK-ModelDev/issues/946>

PR #3879, opened 2024-11-15
Add bazel configs for some C APIs and pybinds

### LINKED ISSUE #3787: Bazel configuration missing for Python bindings?
I'm interested in using the torch-mlir python bindings in a project that uses Bazel as its build system. Based on the [developer docs](https://github.com/llvm/torch-mlir/blob/main/docs/development.md#bazel-build), it seems like the Bazel configs for torch-mlir are mostly focused on supporting `torch-mlir-opt`, and indeed I don't see any CAPI or Python binding rules in the main [Bazel build file](https://github.com/llvm/torch-mlir/blob/main/utils/bazel/torch-mlir-overlay/BUILD.bazel). 

I think what's needed to do this is:
* The CAPI library (from `torch-mlir-c/Dialects.h`). This can be built with `mlir_c_api_cc_library` from `@llvm-project//mlir:build_defs.bzl`.
* The pybind11 extension library (from `python/TorchMLIRModule.cpp`). 
* A tablegen rule to generate the Python wrapper code for ops (`_torch_ops_gen.py`). Again, the MLIR project has examples.
* (Maybe) A convenience `py_library` that bundles the above and any MLIR Python dependencies.

There are examples of these in the MLIR project's [Bazel file](https://github.com/llvm/llvm-project/blob/main/utils/bazel/llvm-project-overlay/mlir/BUILD.bazel). 

Is there interest in this from the maintainers? I'd be willing to contribute these changes if so.

mmirgttam:
This PR improves bazel support by adding rules for building the following:
* C API binding for `torch-mlir-c/Registration.h`
* C API binding for `torch-mlir-c/Dialects.h`
* pybind11 extension module for `torch` dialect
* Tablegen'd Python wrapper classes for `torch` ops
* Filegroups for Python files

Closes #3787.

### DIFF
--- a/utils/bazel/torch-mlir-overlay/BUILD.bazel
+++ b/utils/bazel/torch-mlir-overlay/BUILD.bazel
@@ -2,7 +2,8 @@
 # See https://llvm.org/LICENSE.txt for license information.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library", "td_library")
+load("@llvm-project//mlir:build_defs.bzl", "mlir_c_api_cc_library")
+load("@llvm-project//mlir:tblgen.bzl", "gentbl_cc_library", "gentbl_filegroup", "td_library")
 load("@com_github_bazelbuild_buildtools//buildifier:def.bzl", "buildifier")
 
 package(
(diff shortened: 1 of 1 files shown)


tkocmathla:
@sjain-stanford are you the right person to review this? this is a follow up to our [discord conversation](https://discord.com/channels/636084430946959380/742573221882364009/1301010509415452692).

sjain-stanford (changes requested):
Thanks @tkocmathla . As it stands, there is no way to test these python interfaces. Can you please add some python test targets that depend on these (and hence are tested in CI)? You probably need a hermetic python interpreter setup in bazel as well.

PR #3883, opened 2024-11-19
Add support for the padding variations of conv op 

sahas3:
ConvOp defined with padding = "same"/"valid" produces the padding variant of the op, such as `conv2d.padding` for 2d convolution. This PR adds these conv variations to torch-mlir registry and a decomposition of these ops to `aten.convolution` to be able to go through the different pass pipelines.


sjarus:
Looks ok to me, will let @vivekkhandelwal1 do a final review and merge.

sahas3:
Hello @vivekkhandelwal1,  any thoughts on this PR? Thanks!

vivekkhandelwal1 (approved):
LGTM

PR #3887, opened 2024-11-22
convert to double before float materialization in scalarize shapes

zjgarvey:
Addresses a bug when trying to materialize a non fp64 attr to a constant float op in scalarize shapes.

### DIFF
--- a/lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
+++ b/lib/Dialect/Torch/Transforms/ScalarizeShapes.cpp
@@ -37,8 +37,8 @@ LogicalResult materializeFolds(ImplicitLocOpBuilder b,
 
     if (auto attr = dyn_cast<Attribute>(f)) {
       if (auto val = dyn_cast<FloatAttr>(attr)) {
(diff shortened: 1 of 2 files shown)


jinchen62:
@zjgarvey I'm a bit confused, why are we converting all `FloatAttr` to double.

zjgarvey:
> @zjgarvey I'm a bit confused, why are we converting all `FloatAttr` to double.

`Torch::ConstantFloatOp` has a value attribute that needs to be a `f64` or `double`. Not converting `FloatAttr`s to double and trying to create a constant float op will result in the error you saw:

```
error: 'torch.constant.float' op attribute 'value' failed to satisfy constraint: 64-bit float attribute
    %908 = torch.operator "onnx.Cast"(%907) {torch.onnx.to = 1 : si64} : (!torch.vtensor<[1],si64>) -> !torch.vtensor<[1],f32>
note: see current operation: %777 = "torch.constant.float"() <{value = 3.000000e+00 : f32}> : () -> !torch.float
```

jinchen62 (approved):
LGTM

PR #3888, opened 2024-11-22
Support stash_type attribute for onnx.LayerNormalization

jinchen62:
Fixes https://github.com/nod-ai/SHARK-ModelDev/issues/888

If stash_type is different from input_dtype/result_dtype:
1. convert x dtype to stash_type
2. calculate mean and var in stash_type since x is in stash_type already
3. convert back to result_dtype before stage two calculation
4. convert mean_dtype and var_dtype if they are different from stash_type

e2e test added in https://github.com/nod-ai/SHARK-TestSuite/pull/399

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -2543,7 +2543,34 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
             binder.s64IntegerAttr(stashType, "stash_type", 1))
           return failure();
 
+        std::optional<int64_t> stashTypeIntTorch =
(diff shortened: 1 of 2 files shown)


zjgarvey (approved):
I suppose I still don't get why this works since the two stages mentioned in the onnx docs aren't separated in the torch op. Technically, aren't we supposed to cast back to the original result type before the final mul and add? 

I think it is fine to merge this as-is, considering the examples we tested seem to give correct numerics. If we end up seeing numeric failures in models with layer normalization and weird dtypes, we can always fallback on the function expander on import. 

Can you mention the relevant e2e tests in the commit message?

jinchen62:
@zjgarvey Yeah I convert the dtype back before stage two if the dtype is different in decomposition. Added link of e2e test in commit message.

PR #3890, opened 2024-11-22
Fix output size computation for MaxPool2D for ceil_model = true.

sahas3:
This PR fixes the output size computation as per https://github.com/pytorch/pytorch/blob/d8c14838f164ee02b88b6e37471b71bb0373f865/torch/_meta_registrations.py#L3847

```
if ceil_mode:
        if (outputSize - 1) * stride >= inputSize + pad_l:
            outputSize -= 1
    return outputSize
```


lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -116,6 +117,34 @@ Value torch_to_linalg::getOutputDimForConvOps(OpBuilder &b, Location loc,
...
+    // Emit 'else' region of 'scf.if'
+    auto emitElseRegion = [&](OpBuilder &opBuilder, Location loc) {
+      opBuilder.create<scf::YieldOp>(loc, out);
+    };
+
+    // Emit 'scf.if' op
+    auto ifOp = b.create<scf::IfOp>(loc, reduceOutputDim, emitThenRegion,
+                                    emitElseRegion);
+
+    return castIntToIndex(b, loc, ifOp.getResult(0));
```
vivekkhandelwal1, line 145: You can also use `arith.Select` op in place of `scf.if`.

vivekkhandelwal1 (changes requested):

lib/Conversion/TorchToLinalg/Utils.cpp
```diff
@@ -527,8 +556,8 @@ Value torch_to_linalg::convertTensorToElementType(OpBuilder &b, Location loc,
                                                   Type elementType) {
   auto dtypePromoteBody = [&](OpBuilder &builder, Location loc,
                               ValueRange payloadArgs) {
-    Value elem =
-        convertScalarToDtype(builder, loc, payloadArgs[0], elementType);
+    Value elem = mlir::torch::Torch::convertScalarToDtype(
```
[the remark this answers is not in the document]
sahas3: `mlir::torch::Torch` was required otherwise `convertScalarToDtype` defined in `Arith/Utils.h` (included via `SCF.h`) was being used instead. However, with the change to use `arith.select` in place of `scf.if` I no longer need to include `SCF.h`, so this is reverted.

lib/Conversion/TorchToLinalg/Utils.cpp
sahas3: Fixed.

sjarus (approved):
Looks reasonable for TOSA target. Please check that there are no XPASSes on any TOSA target.

vivekkhandelwal1 (approved):

PR #3891, opened 2024-11-22
[RFC] Apply the trait `NoMemoryEffect` to most `ReadOnly` ops

zjgarvey:
In my understanding, the trait `ReadOnly` implies that the memory for input values does not get modified by the operation. 

This is similar to `NoMemoryEffect`, with a few exceptions. Ops like `prim.RaiseException`  and `prim.Print` are `ReadOnly`, but have effects like terminating the program and printing, respectively. 

There may be other `ReadOnly` ops with side effects. As a hypothetical example, there might be a comparison op that checks if two tensors are the same type, prints some debug info, and also returns a boolean value. In the event that such an op is discovered, I've added a keyword argument to the `emit_op` function to explicitly label an operation as `has_memory_effects=True` to avoid adding the `NoMemoryEffect` trait to the tablegen definition. 

My primary motivation for this change is to remove the need to have one-off `RemoveUnused` patterns for ops that we want to remove when dead, but adding this trait also has the very appealing effect of dramatically simplifying torch-IR for some models. Specifically for ONNX models, it is not uncommon to have patterns that consistently call "onnx.Shape" on the same exact tensor, only to extract a different element from the shape. That redundant IR doesn't get CSE'd until converting to a core mlir dialect like linalg/tensor/arith.

### DIFF
--- a/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
+++ b/include/torch-mlir/Dialect/Torch/IR/GeneratedTorchOps.td
@@ -22,7 +22,8 @@
 def Torch_AtenHardtanhOp : Torch_Op<"aten.hardtanh", [
     AllowsTypeRefinement,
     HasValueSemantics,
-    ReadOnly
+    ReadOnly,
+    NoMemoryEffect
   ]> {
   let summary = "Generated op for `aten::hardtanh : (Tensor, Scalar, Scalar) -> (Tensor)`";
   let arguments = (ins
@@ -71,7 +72,8 @@ def Torch_AtenHardtanh_Op : Torch_Op<"aten.hardtanh_", [
 def Torch_AtenEluOp : Torch_Op<"aten.elu", [
     AllowsTypeRefinement,
     HasValueSemantics,
-    ReadOnly
+    ReadOnly,
+    NoMemoryEffect
   ]> {
   let summary = "Generated op for `aten::elu : (Tensor, Scalar, Scalar, Scalar) -> (Tensor)`";
   let arguments = (ins
@@ -122,7 +124,8 @@ def Torch_AtenElu_Op : Torch_Op<"aten.elu_", [
 def Torch_AtenReluOp : Torch_Op<"aten.relu", [
     AllowsTypeRefinement,
     HasValueSemantics,
-    ReadOnly
+    ReadOnly,
+    NoMemoryEffect
   ]> {
   let summary = "Generated op for `aten::relu : (Tensor) -> (Tensor)`";
   let arguments = (ins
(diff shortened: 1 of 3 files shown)


zjgarvey:
I'm wondering if it is more appropriate to label most ops as `MemoryEffects::Allocate`? Perhaps `NoMemoryEffect` would imply that it also doesn't allocate memory for a new output tensor. 

Please give some feedback if you have some opinions on what would be the best way forward on this.

IanWood1:
> I'm wondering if it is more appropriate to label ops more carefully with specific `MemoryEffects`? Perhaps `NoMemoryEffect` would imply that it also doesn't allocate memory for a new output tensor, or write to the new output tensor?
> 
> Please give some feedback if you have some opinions on what would be the best way forward on this.


I looked into this a bit ago and one thing I found is that there are several ops that can throw exceptions that aren't as obvious as `prim.RaiseException` (e.g. [torch.aminmax](https://pytorch.org/docs/stable/generated/torch.aminmax.html#torch-aminmax)). But I'm not sure if torch-mlir's goal is to respect these errors so it may not be a problem.

kuhar (commented):
I skimmed it and it looks fine to me. I'd probably put all of these standard properties in a new def and give them a name instead of repeating everywhere. This should make it easier to update if you realize you need to tweak it later on. For example, see how `Pure` is defined here: https://github.com/llvm/llvm-project/blob/132de3a71f581dcb008a124d52c83ccca8158d98/mlir/include/mlir/Interfaces/SideEffectInterfaces.td#L144C1-L146C60

---

Edit:

Actually, I don't think you need `ReadOnly` on most/any of these. If they only operate on tensors, there's nor memory accessed. Printing, however, would be considered a side effect.

---

Edit 2:

I don't know where `ReadOnly` comes from, I don't see it in MLIR. Is this a torch thing?

kuhar:
> I'm wondering if it is more appropriate to label ops more carefully with specific `MemoryEffects`? Perhaps `NoMemoryEffect` would imply that it also doesn't allocate memory for a new output tensor, or write to the new output tensor?

Tensors are not considered memory. See https://mlir.llvm.org/docs/Rationale/SideEffectsAndSpeculation/.

IanWood1 (approved):
I think `operator.is_readonly() and len(operator.returns) != 0` makes sense as a heuristic. It's a bit concerning that there is no exact way to determine if an op has `NoMemoryEffect` but that may just be the nature of mapping torch -> mlir. This should be a good change to simplify the generated IR.

I'll approve since the rationale and implementation of the PR make sense to me. But this probably needs another set of eyes as it may lead to some difficult to find bugs if ops get mislabeled.

ibsidorenko:
Hi, @zjgarvey !
Thank you for this PR. Adding of trait `NoMemoryEffect` - this is something I’ve been dreaming about for a long time. 👍 
As I see this PR was approved, but still was not merged. Are there any blockers to merging it?

zjgarvey:
> Hi, @zjgarvey ! Thank you for this PR. Adding of trait `NoMemoryEffect` - this is something I’ve been dreaming about for a long time. 👍 As I see this PR was approved, but still was not merged. Are there any blockers to merging it?

@ibsidorenko Sorry for losing track of this comment. I'll clean this up and try to land this week.

PR #3892, opened 2024-11-26
Add TorchToArith rewrite pattern for AtenGtFloatOp

jinchen62:


zjgarvey (approved):
A lit test would be nice, but this is pretty straightforward. Thanks, Jinchen.

PR #3904, opened 2024-12-05
[MLIR][TORCH] Add support for 1-d group convolution

vivekkhandelwal1:
This commit adds the support for 1-d group convolution by transforming it into a 2-d group convolution which is already supported.

This commit also refactors the unsqueeze and squeeze tensor utility.


lib/Conversion/Utils/Utils.cpp
```diff
@@ -447,6 +447,120 @@ Value toPositiveValidDim(ConversionPatternRewriter &rewriter, Location loc,
...
+FailureOr<Value> squeezeTensor(PatternRewriter &rewriter, Operation *op,
+                               Value input, int64_t dim) {
+  Location loc = op->getLoc();
+  auto inputType = cast<RankedTensorType>(input.getType());
+  int64_t inputRank = inputType.getRank();
+
+  if (inputRank == 0) {
+    return rewriter.notifyMatchFailure(
+        op, "zero input rank should have been handled by the folder");
+  }
```
zjgarvey, line 506: The folder for what op?

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -850,6 +850,46 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
 
+    // Checks for valid group size
+    int64_t numGroups;
+    if (!matchPattern(op.getGroups(), m_TorchConstantInt(&numGroups)))
+      return rewriter.notifyMatchFailure(op,
+                                         "only constant group size supported.");
+    Value groups = castIntToIndex(rewriter, loc, adaptor.getGroups());
+
+    // Adding support for 1d group convolution by converting the 1d-conv to
+    // 2d-conv.
```
zjgarvey, line 861: Let's add "TODO: replace this logic when linalg has an appropriate op corresponding to 1-d grouped convolution".

lib/Conversion/Utils/Utils.cpp
vivekkhandelwal1: Sorry, that code was refactored through the op lowering hence that part was left as it is. I have changed it to handle that as well.

vivekkhandelwal1:
> If you used most of the logic from `squeeze.dim` and `unsqueeze` lowerings to linalg for these helper functions, should we use these util functions there as well?

I have already made those changes and replaced the code. Could you please re-check? Or, let me know if I am missing something.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -850,6 +850,48 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
...
+
+      Value cstZero = rewriter.create<arith::ConstantOp>(
+          loc, rewriter.getI64IntegerAttr(0));
+      paddingIntValues.push_back(cstZero);
+      outputPaddingIntValues.push_back(cstZero);
+      strideInts.push_back(1);
+      dilationInts.push_back(1);
+
+      inRank += 1;
+      numSpatialDims += 1;
```
zjgarvey, line 892: nit 
```suggestion
      inRank++;
      numSpatialDims++;
```

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -1280,13 +1315,24 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
         conv = torch_to_linalg::convertTensorToElementType(rewriter, loc, conv,
                                                            resultElementType);
       }
+
+      if (is1DGroupConv) {
+        // Squeezing the last dim of the result of conv.
+        auto squeezeInputInfo = squeezeTensor(rewriter, op, conv, /*dim=*/-1);
+        if (failed(squeezeInputInfo)) {
+          return rewriter.notifyMatchFailure(
+              op, "cannot generate unsqueeze tensor");
```
zjgarvey, line 1324:
```suggestion
              op, "cannot generate squeeze tensor");
```

zjgarvey (approved):
A few nits, then LGTM.

lib/Conversion/TorchToLinalg/Linear.cpp
zjgarvey: Maybe also rename variable to `squeezeOutputInfo`

zjgarvey: Same with the other instance of this.

PR #3907, opened 2024-12-06
Pin and update actions

marbre:
This pins and updates most actions. The PR is limited to those actions that seem actively maintained and updated. The actions left unpined should be reevaluated and eventually replaced with other actions. The rational for pinning actions is to follow the suggestions by OpenSSF Scorecard, see https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies.


ScottTodd (approved):
Workflow failures look to be preexisting and unrelated to the changes here.

PR #3913, opened 2024-12-09
Revert "[CI] Disable LTC build from CI"

antoniojkim:
Attempt to revert llvm/torch-mlir#3911 and run CI


vivekkhandelwal1 (commented):
Hi @antoniojkim, this PR can't be merged until the issue is fixed.

stellaraccident:
I believe that LTC needs its own testing bots. It shares practically no infra with the rest of the project, and while it is a fine thing to have in the repo if that is valuable to folks, the costs need to come from the people using it. One way to do that is by contributing a bot and test config tailored to what you need.

stellaraccident (changes requested):
Please only enable LTC with dedicated CI.

PR #3914, opened 2024-12-09
[onnx][importer] Merge character sanitization into regex

vinayakdsci:
Refactors https://github.com/llvm/torch-mlir/pull/3901 by merging illegal character sanitization into the regex and removing call to `replace()`.


vinayakdsci:
cc @zjgarvey, @ScottTodd.

ScottTodd (approved):

PR #3876, opened 2024-11-15
[Torch] Add decomposition for 1d torch.nonzero

AmosLewis:
Target model: 
`migraphx_onnx-model-zoo__gpt2-10`
`    %350 = torch.operator "onnx.NonZero"(%349) : (!torch.vtensor<[?],si64>) -> !torch.vtensor<[1,?],si64> `
```
module {
  func.func @main_graph(%arg0: !torch.vtensor<[?],i1>) -> !torch.vtensor<[1,?],si64>  attributes {torch.onnx_meta.ir_version = 9 : si64, torch.onnx_meta.opset_version = 20 : si64, torch.onnx_meta.producer_name = "pytorch", torch.onnx_meta.producer_version = "2.6.0"} {
    %0 = torch.operator "onnx.NonZero"(%arg0) : (!torch.vtensor<[?],i1>) -> !torch.vtensor<[1,?],si64> 
    return %0 : !torch.vtensor<[1,?],si64> 
  }
}
```
```
%176 = torch.aten.nonzero %175 : !torch.vtensor<[?],si64> -> !torch.vtensor<[1,1],si64>
%177 = torch.aten.transpose.int %176, %int0, %int1 : !torch.vtensor<[1,1],si64>, !torch.int, !torch.int -> !torch.vtensor<[1,1],si64>
```

The python implementation: [nonzero.py](https://gist.github.com/AmosLewis/f1b0dcad2c9ee2381f19d5836df952ff#file-nonzero-py) and [decompose ](https://gist.github.com/AmosLewis/7cbf94ef1f18daa35631c3fc08b314db)e2etest
To fix e2e test [error ](https://github.com/llvm/torch-mlir/actions/runs/11503867149/job/32022095672?pr=3721)in xida's previous draft https://github.com/llvm/torch-mlir/pull/3721

Here is the bug and reproducer mlir: https://gist.github.com/AmosLewis/92717dbe4847649afefc915425629124
```
Running AtenNonzero1DModule_one_nonzero...
mismatched size for broadcast
./build_tools/ci/test_posix.sh: line 12: 3770074 Aborted                 (core dumped) python -m e2e_testing.main --config=onnx -v --filter AtenNonzero1DModule_one_nonzero
```

Related iree issue to be fixed: https://github.com/iree-org/iree/issues/19481

PR #3916, opened 2024-12-11
[ONNX] Fix nonzero output type difference  between onnx and torch

AmosLewis:
The onnx output tensor has a shape of ((n, z)), where (n) is the number of dimensions in the input tensor and (z) is the number of non-zero elements2. This is different from PyTorch's default behavior, where the dimensions are reversed.

This help to debug decompose torch.nonzero https://github.com/llvm/torch-mlir/pull/3876

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
@@ -1093,18 +1093,35 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
         rewriter.replaceOp(binder.op, nllLoss);
         return success();
       });
-  patterns.onOp("NonZero", 13,
-                [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-                  Torch::ValueTensorType resultType;
-                  Value operand;
-                  if (binder.tensorOperand(operand) ||
-                      binder.tensorResultType(resultType)) {
-                    return failure();
-                  }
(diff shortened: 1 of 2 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1093,18 +1093,36 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(64), 0));
+        Value one = rewriter.create<Torch::ConstantIntOp>(
+            binder.getLoc(), rewriter.getType<Torch::IntType>(),
+            rewriter.getIntegerAttr(rewriter.getIntegerType(64), 1));
+        auto rawSize = resultType.getSizes();
+        SmallVector<int64_t> torchResultSize(rawSize.rbegin(), rawSize.rend());
+        auto torchResultType = Torch::ValueTensorType::get(
+            rewriter.getContext(), torchResultSize,
+            rewriter.getIntegerType(64, /*signed=*/true));
```
zjgarvey, line 1114: I'd prefer:
```suggestion
        auto torchResultType = rewriter.getType<Torch::ValueTensorType>(torchResultSize, resultType.getDtype());
```
Particularly in case the result dtype is allowed to be int32 in a future opset_version, for example.

zjgarvey (approved):
With a minor change, LGTM.

zjgarvey (approved):

PR #3917, opened 2024-12-13
[ONNX] Add per channel quantization support for Onnx.QLinearConv op

vivekkhandelwal1:
This commit extends the OnnxToTorch Lowering for Onnx.QLinearConv op by adding the support for per channel quantization for the weight argument.

Since the convolution operation in the downstream pipeline ("Linalg") does not support the per-channel quantization, hence we add the support by performing convolution over the dequantized input and weight and then quantizing the output.

Fixes https://github.com/nod-ai/SHARK-ModelDev/issues/894.

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -9,9 +9,12 @@
 
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Utils.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/ADT/ArrayRef.h"
+#include "llvm/ADT/STLExtras.h"
 #include "llvm/ADT/SmallVector.h"
+#include "llvm/Support/ErrorHandling.h"
 
 using namespace mlir;
 using namespace mlir::torch;
@@ -376,6 +379,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         input = makePerTensor(input, inputScale, inputZp);
         // The onnx's QLinearConv op expects per channel quantization only for
         // the weight tensor for axis = 0.
+        bool isPerChannelQuantization = false;
         auto weightTy = dyn_cast<Torch::ValueTensorType>(weight.getType());
         auto weightScaleTy =
             dyn_cast<Torch::ValueTensorType>(weightScale.getType());
@@ -384,31 +388,25 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           return failure();
         auto weightShape = weightTy.getSizes();
         auto weightScaleShape = weightScaleTy.getSizes();
-        Value weightScaleScalar = extract(weightScale);
-        if (weightScaleShape.size() == 1 &&
-            weightScaleShape[0] != Torch::kUnknownSize &&
-            weightScaleShape[0] == weightShape[0]) {
+        if (weightScaleShape.size() == 0 || llvm::all_of(weightScaleShape, 1)) {
+          weightZp = extract(weightZp);
+          weightScale = extract(weightScale);
+          weight = makePerTensor(weight, weightScale, weightZp);
+        } else if (weightScaleShape.size() == 1 &&
+                   weightScaleShape[0] != Torch::kUnknownSize &&
+                   weightScaleShape[0] == weightShape[0]) {
           Value axis = rewriter.create<Torch::ConstantIntOp>(
               binder.getLoc(), rewriter.getI64IntegerAttr(0));
           weight = makePerChannel(weight, weightScale, weightZp, axis);
+          isPerChannelQuantization = true;
         } else {
-          weightZp = extract(weightZp);
-          weight = makePerTensor(weight, weightScaleScalar, weightZp);
+          llvm_unreachable("Unidentified case for weight quantization");
         }
-        weightScale = weightScaleScalar;
-
-        auto outputTy = rewriter.getType<Torch::ValueTensorType>(
-            resultType.getOptionalSizes(),
-            rewriter.getIntegerType(32, /*issigned=*/true));
 
         // TODO(suderman): insert convolution operator.
         llvm::SmallVector<Value> newOperands = {input, weight};
         if (output)
           newOperands.push_back(output);
-
-        outputTy = rewriter.getType<Torch::ValueTensorType>(
-            resultType.getOptionalSizes(),
-            rewriter.getType<Torch::QInt32Type>());
 
         llvm::SmallVector<NamedAttribute> newAttributes;
         newAttributes.push_back(
@@ -420,20 +418,34 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           newAttributes.push_back(namedAttr);
         }
 
+        auto outputTy = rewriter.getType<Torch::ValueTensorType>(
+            resultType.getOptionalSizes(),
+            rewriter.getType<Torch::QInt32Type>());
+
         output = rewriter
                      .create<Torch::OperatorOp>(binder.getLoc(), outputTy,
                                                 newOperands, newAttributes,
                                                 binder.op->getRegions().size())
                      .getResult(0);
 
-        Value outScale = rewriter.create<Torch::AtenMulFloatOp>(
-            binder.getLoc(), rewriter.getType<Torch::FloatType>(), inputScale,
-            weightScale);
-        Value outZp = rewriter.create<Torch::ConstantIntOp>(
-            binder.getLoc(), rewriter.getType<Torch::IntType>(),
-            rewriter.getIntegerAttr(rewriter.getIntegerType(64), 0));
-        output = rewriter.create<Torch::Aten_MakePerTensorQuantizedTensorOp>(
-            binder.getLoc(), outputTy, output, outScale, outZp);
+        Value outScale, outZp;
+        if (isPerChannelQuantization) {
+          // outZp = rewriter.create<Torch::Ate
+          outScale = rewriter.create<Torch::AtenMulScalarOp>(
+              binder.getLoc(), weightScaleTy, weightScale, inputScale);
+          output = rewriter.create<Torch::Aten_MakePerChannelQuantizedTensorOp>(
+              binder.getLoc(), outputTy, output, outScale, outZp);
+        } else {
+          outZp = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getType<Torch::IntType>(),
+              rewriter.getIntegerAttr(rewriter.getIntegerType(64), 0));
+          outScale = rewriter.create<Torch::AtenMulFloatOp>(
+              binder.getLoc(), rewriter.getType<Torch::FloatType>(), inputScale,
+              weightScale);
+          output = rewriter.create<Torch::Aten_MakePerTensorQuantizedTensorOp>(
+              binder.getLoc(), outputTy, output, outScale, outZp);
+        }
+
         outputTy = rewriter.getType<Torch::ValueTensorType>(
             resultType.getOptionalSizes(), rewriter.getF32Type());
 
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -9,11 +9,9 @@
 
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Utils.h"
-#include "torch-mlir/Dialect/Torch/IR/TorchTypes.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/SmallVector.h"
-#include "llvm/Support/raw_ostream.h"
 
 using namespace mlir;
 using namespace mlir::torch;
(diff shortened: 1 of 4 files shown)


lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -318,24 +318,15 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
-              binder.op, "not supported for non per-tensor quantization");
+        Value input = operands[0];
+        Value inputScale = operands[1];
+        Value inputZp = operands[2];
+        Value weight = operands[3];
+        Value weightScale = operands[4];
+        Value weightZp = operands[5];
+        Value outputScale = operands[6];
+        Value outputZp = operands[7];
+        Value output = operands.size() == 9 ? operands[8] : nullptr;
```
vinayakdsci, line 329: @vivekkhandelwal1 this can be easily wrapped up in a helper function and then unpacked using structured bindings.

The following diff should work:

```diff
diff --git a/include/torch-mlir/Conversion/Utils/Utils.h b/include/torch-mlir/Conversion/Utils/Utils.h
index 264fb496..4a00c324 100644
--- a/include/torch-mlir/Conversion/Utils/Utils.h
+++ b/include/torch-mlir/Conversion/Utils/Utils.h
@@ -18,6 +18,21 @@ namespace mlir {
 namespace torch {
 namespace Torch {
 
+template <typename ElemType, int sz>
+std::optional<std::array<ElemType, sz>>
+takeFirstN(SmallVector<ElemType> &container) {
+  if (sz > container.size()) {
+    return std::nullopt;
+  }
+
+  std::array<ElemType, sz> firstNElems;
+  for (size_t i = 0; i < sz; ++i) {
+    firstNElems[i] = container[i];
+  }
+
+  return firstNElems;
+}
+
 LogicalResult verifyLinalgCompatibleTypes(Operation *op,
                                           PatternRewriter &rewriter);
 
diff --git a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
index 2a6f42a4..3f8361b1 100644
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -9,6 +9,7 @@
 
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Patterns.h"
 #include "torch-mlir/Conversion/TorchOnnxToTorch/Utils.h"
+#include "torch-mlir/Conversion/Utils/Utils.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/ADT/ArrayRef.h"
 #include "llvm/ADT/SmallVector.h"
@@ -318,14 +319,11 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
              binder.tensorOperands(operands, 9)) ||
             binder.tensorResultType(resultType))
           return failure();
-        Value input = operands[0];
-        Value inputScale = operands[1];
-        Value inputZp = operands[2];
-        Value weight = operands[3];
-        Value weightScale = operands[4];
-        Value weightZp = operands[5];
-        Value outputScale = operands[6];
-        Value outputZp = operands[7];
+
+        const auto firstEightOperands = Torch::takeFirstN<Value, 8>(operands);
+        auto [input, inputScale, inputZp, weight, weightScale, weightZp,
+              outputScale, outputZp] = firstEightOperands.value();
+
         Value output = operands.size() == 9 ? operands[8] : nullptr;
 
         auto extract = [&rewriter, &binder](Value v) {
```

Passing the size as a template parameter makes it known at compile-time, otherwise the unpacking would not be allowed by the compiler.

vivekkhandelwal1: Okay. There might be many ways of doing this but what are we trying to achieve via this change? Instead, it increases  the code size and I don't see it more optimal than the former one in any way.

vinayakdsci: Well, IMO splitting this logic into a function is useful as this makes it reusable. I can't see how exactly it would make the binary any slower. Writing 10 lines of code every time you need the first 10 elements of an array appears to me to warrant a function, but, of course, YMMV.

vivekkhandelwal1: Yeah, it's definitely good to have something in reusable form but that's beneficial when there are multiple uses of it. I can't find multiple uses of this except 1 (QLinearMatmul). And, I think the original author wrote it that way to make it more readable otherwise all of that is nothing more than just a single line of code. 

Anyway, sorry for the disagreement but I don't find moving this to a utility more helpful. So, I would like to keep it like that unless otherwise we have more people supporting that argument.

vinayakdsci: No, that is alright, of course. We can leave this, unless of course other reviewers think otherwise.

jinchen62 (approved):
LGTM

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -347,34 +338,61 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        // the weight tensor for axis = 0.
+        auto weightTy = dyn_cast<Torch::ValueTensorType>(weight.getType());
+        auto weightScaleTy =
+            dyn_cast<Torch::ValueTensorType>(weightScale.getType());
+        if (!weightTy || !weightScaleTy || !weightTy.hasSizes() ||
+            !weightScaleTy.hasSizes())
+          return failure();
+        auto weightShape = weightTy.getSizes();
+        auto weightScaleShape = weightScaleTy.getSizes();
+        Value weightScaleScalar = extract(weightScale);
```
zjgarvey, line 373: extract won't work if the weight scale isn't a single element. I'd put this in the else block below.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -347,34 +338,61 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        if (weightScaleShape.size() == 1 &&
+            weightScaleShape[0] != Torch::kUnknownSize &&
+            weightScaleShape[0] == weightShape[0]) {
```
zjgarvey, line 376: Additionally check that `weightShape[0] != 1` since we don't want to lower to per-channel when there is only one channel.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -347,34 +338,61 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          Value axis = rewriter.create<Torch::ConstantIntOp>(
+              binder.getLoc(), rewriter.getI64IntegerAttr(0));
+          weight = makePerChannel(weight, weightScale, weightZp, axis);
+        } else {
+          weightZp = extract(weightZp);
+          weight = makePerTensor(weight, weightScaleScalar, weightZp);
+        }
```
zjgarvey, line 383: A bit of a nit, but I'd prefer an `else if` here with the conditions for `makePerTensor`, and then an else branch with an unreachable, just to be very clear about what assumptions are being made in each case.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -347,34 +338,61 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
         // TODO(suderman): insert convolution operator.
-        llvm::SmallVector<Value> newOperands = {a, b};
-        if (c)
-          newOperands.push_back(c);
+        llvm::SmallVector<Value> newOperands = {input, weight};
+        if (output)
+          newOperands.push_back(output);
 
-        cTy = rewriter.getType<Torch::ValueTensorType>(
+        outputTy = rewriter.getType<Torch::ValueTensorType>(
```
zjgarvey, line 395: Okay, this is a bit subtle. The last optional input for this op is the int32 bias, assumed to be quantized via the product of input and weight scales. This implies that the quantization of the bias (and also the output of the convolution) is also per-channel if the weight was per-channel quantized. This part is fine, but we will need to case out the logic below.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
zjgarvey: I see you use this below to handle the quantization of the output, but this must also be per-channel if the weight is per-channel.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -388,36 +406,36 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        output = rewriter
+                     .create<Torch::OperatorOp>(binder.getLoc(), outputTy,
+                                                newOperands, newAttributes,
+                                                binder.op->getRegions().size())
+                     .getResult(0);
 
         Value outScale = rewriter.create<Torch::AtenMulFloatOp>(
-            binder.getLoc(), rewriter.getType<Torch::FloatType>(), aScale,
-            bScale);
+            binder.getLoc(), rewriter.getType<Torch::FloatType>(), inputScale,
```
zjgarvey, line 416: Will need to possibly be float x tensor mul.

zjgarvey (changes requested):
Thanks Vivek. I think you need to modify some of the output quantization handling in the per-channel case. Maybe store a bool that tracks if we are in the per-channel case so you can reuse it for the output.

It looks like this conversion automatically fuses the input and weight quantization with the convolution, so the only thing that fuse-quantized-ops is going to do is quantize the bias (which won't work currently in the per-channel case). I think it is fine, but we won't be able to check correctness e2e until we address the per-channel quantization, unfortunately.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
vivekkhandelwal1: I have updated the conditions.

vivekkhandelwal1:
Hi @zjgarvey, can you please review the patch now?

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
```diff
@@ -141,6 +141,65 @@ func.func @test_qlinearconv_bias(%arg0: !torch.vtensor<[1,1,7,7],ui8>, %arg1: !t
...
+// CHECK-SAME:                                                                %[[W_ZP:[0-9]+|[a-zA-Z$._-][a-zA-Z0-9$._-]*]]: !torch.vtensor<[64],si8>,
+// CHECK-SAME:                                                                %[[OUT_SCALE:[0-9]+|[a-zA-Z$._-][a-zA-Z0-9$._-]*]]: !torch.vtensor<[],f32>,
+// CHECK-SAME:                                                                %[[OUT_ZP:[0-9]+|[a-zA-Z$._-][a-zA-Z0-9$._-]*]]: !torch.vtensor<[],ui8>,
+// CHECK-SAME:                                                                %[[BIAS:[0-9]+|[a-zA-Z$._-][a-zA-Z0-9$._-]*]]: !torch.vtensor<[64],si32>) -> !torch.vtensor<[?,64,112,112],ui8> attributes {torch.onnx_meta.ir_version = 4 : si64, torch.onnx_meta.opset_version = 21 : si64, torch.onnx_meta.opset_versions = {ai.onnx.ml = 2 : si64, ai.onnx.preview.training = 1 : si64, ai.onnx.training = 1 : si64, com.microsoft = 1 : si64, com.microsoft.experimental = 1 : si64, com.microsoft.mlfeaturizers = 1 : si64, com.microsoft.nchwc = 1 : si64}, torch.onnx_meta.producer_name = "onnx.quantize", torch.onnx_meta.producer_version = "0.1.0"} {
```
zjgarvey, line 153: we don't need the attributes checked.

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
```diff
@@ -141,6 +141,65 @@ func.func @test_qlinearconv_bias(%arg0: !torch.vtensor<[1,1,7,7],ui8>, %arg1: !t
...
+  // CHECK:           %[[TRANSPOSED:.*]] = torch.constant.bool false
+  // CHECK:           %[[GROUPS:.*]] = torch.constant.int 1
+  // CHECK:           %[[CONV:.*]] = torch.aten.convolution %[[DEQUANT_INPUT]], %[[DEQUANT_WEIGHT]], %[[F32_BIAS]], %[[DILATION]], %[[PAD]], %[[KERNEL]], %[[TRANSPOSED]], %[[STRIDE]], %[[GROUPS]] : !torch.vtensor<[?,3,224,224],f32>, !torch.vtensor<[64,3,7,7],f32>, !torch.vtensor<[64],f32>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool, !torch.list<int>, !torch.int -> !torch.vtensor<[?,64,112,112],f32>
+  // CHECK:           %[[DTYPE:.*]] = torch.constant.int 13
+  // CHECK:           %[[QUANT:.*]] = torch.aten.quantize_per_tensor %[[CONV]], %[[OUTPUT_SCALE]], %[[OUTPUT_ZP]], %[[DTYPE]] : !torch.vtensor<[?,64,112,112],f32>, !torch.float, !torch.int, !torch.int -> !torch.vtensor<[?,64,112,112],!torch.quint8>
+  // CHECK:           %[[OUTPUT:.*]] = torch.aten.int_repr %[[QUANT]] : !torch.vtensor<[?,64,112,112],!torch.quint8> -> !torch.vtensor<[?,64,112,112],ui8>
+  // CHECK:           return %[[OUTPUT]] : !torch.vtensor<[?,64,112,112],ui8>
```
zjgarvey, line 197: nit: This lit test is a literal copy paste of the resulting IR and I'd much prefer the simplest possible test to check the weight dequantization logic and the fact that the conv is over f32.

zjgarvey: Especially if we need to eventually support per-channel quantized convs, this will just end up being scrapped anyway.

zjgarvey (approved):
Nice, this will get us the functionality we need for now. 

I have a nit about the test being too involved, but otherwise this looks good to me. Have you tested out any numerics?

AmosLewis (approved):

PR #3923, opened 2024-12-19
[ONNX] Add support for asymmetric padding for Onnx.AveragePool op

vivekkhandelwal1:
This commit adds support for the asymmetric padding for Onnx's AveragePool op.

This commit also extends the Torch->Linalg lowering of the pooling ops to consider asymmetric padding during the output dim computation.


lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -456,107 +456,47 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
-              binder.op, "strides list size does not match the number of axes");
-        }
-
-        SmallVector<Value> cstKernel, cstPadding, cstStridesDilations;
-        for (int64_t i : kernel) {
-          cstKernel.push_back(rewriter.create<Torch::ConstantIntOp>(
-              binder.getLoc(), rewriter.getI64IntegerAttr(i)));
-        }
-        // Onnx pads format: [x1_begin, x2_begin…x1_end, x2_end,…]
-        // Pytorch pads format: [x1, x2,...] or [x], assume begin==end for all
```
AmosLewis, line 516: add e2e tests in shark-testsuite if the change work. I don't think torch to linalg support this pattern.

lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
```diff
@@ -456,107 +456,47 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
...
-        if (autoPad != "NOTSET") {
-          // TODO: Add support for `auto_pad` != "NOTSET"
-          return rewriter.notifyMatchFailure(
-              binder.op, "unsupported conversion: auto_pad != NOTSET");
-        }
-
         Torch::ValueTensorType resultType;
         Value operand;
-        bool ceilMode, countIncludePad;
+        int64_t ceilMode, countIncludePad;
```
tuukkjs, line 461: Why change ceilMode and countIncludePad from bool to int64_t?

lib/Conversion/TorchOnnxToTorch/Utils.cpp
```diff
@@ -142,3 +142,117 @@ Value mlir::torch::onnx_c::createActivationByName(ImplicitLocOpBuilder &b,
...
+    padding.resize_for_overwrite(2 * spatialRank);
+    for (unsigned dimIdx = 0; dimIdx < spatialRank; dimIdx++) {
+      const int64_t dilatedKernelSize =
+          dilations[dimIdx] * (kernel[dimIdx] - 1) + 1;
+      int64_t totalPad =
+          ((inputShape[dimIdx + 2] + strides[dimIdx] - 1) / strides[dimIdx] -
+           1) *
+              strides[dimIdx] +
+          dilatedKernelSize - inputShape[dimIdx + 2];
+      totalPad = totalPad >= 0 ? totalPad : 0;
```
tuukkjs, line 203: Is `ceilMode` used in calculating the padding when using autopad? If not, why? I think the formulas in https://onnx.ai/onnx/operators/onnx__AveragePool.html differ depending on `ceilMode`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -1124,138 +1124,38 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
                                              "Unimplemented: unranked tensor");
-        int64_t rank = *maybeRank;
-        int64_t spatial = rank - 2;
+        unsigned rank = *maybeRank;
 
-        SmallVector<int64_t> kernel, padding, strides, dilations;
-        if (binder.s64IntegerArrayAttr(kernel, "kernel_shape", {}))
+        SmallVector<int64_t> kernel, padding, strides, dilations,
+            stridesDilations;
+        if (failed(checkAndGetOnnxPoolingOpParameters(
```
tuukkjs, line 1153: How about `countIncludePad = false`? If we pad using `AtenConstantPadNdOp `and after that do `AtenAvgPoolOp `don’t we lose the ability to support `countIncludePad = false`?

tuukkjs:
Looks very good! I made some comments since I have been working on similar changes. However, I am not very familiar with the project itself so some of my comments may be off.

In general, I would suggest to add more tests to test different cases of auto_pad and asymmetric and symmetric padding.

Also please note that I am out of office until January 7th and likely won't respond during that time. Maybe others can chip in. Previously we have been discussing some changes along these lines with @zjgarvey.

vivekkhandelwal1:
Hi, I have modified the implementation. Can this PR be reviewed again?

jinchen62 (approved):
LGTM. Make sure if @tuukkjs 's comment is addressed before merge.

vivekkhandelwal1:
@tuukkjs Can you please review the PR?

lib/Conversion/TorchOnnxToTorch/Utils.cpp
vivekkhandelwal1: Hi, this comment does not need to be addressed now, since the case for which this particular comment was raised is not supported now after the change in lowering strategy.

vivekkhandelwal1:
This PR can be merged now, since all the comments are addressed.

PR #3924, opened 2024-12-19
[torch] Update `torch.bmm` to use accumulator type

rsuderman:
Batch matmul was using the result type as the accumulator. Updated to use the preferred accumulator based on input type.


MaheshRavishankar:
Do we need a test?

rsuderman:
> Do we need a test?

There should be an existing testing. Getting the `f16` vs `f32` numerical comparison for accumulators would require a much larger matmul than we typically like to include in our e2e tests. For `torch-mlir` we only do numerical evaluations against the torch backend.

MaheshRavishankar (approved):

PR #3927, opened 2024-12-19
[cmake] Enable accepting external stablehlo project

shelkesagar29:
This MR enables `torch_mlir` project to accept path to external stablehlo and include those directories. This in turn enables `torch_mlir` to be part of bigger compiler project when `stablehlo` is already a
dependency.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,6 +43,12 @@ option(TORCH_MLIR_ENABLE_STABLEHLO "Add stablehlo dialect" ON)
 if(TORCH_MLIR_ENABLE_STABLEHLO)
   add_definitions(-DTORCH_MLIR_ENABLE_STABLEHLO)
 endif()
+# It is possible that both stablehlo and torch_mlir projects are used in some compiler project.
+# In this case, we don't want to use stablehlo that is downloaded by torch_mlir (in external/stablehlo)
+# folder but instead want to use stablehlo that is part of top level compiler project.
+# With TORCH_MLIR_USE_EXTERNAL_STABLEHLO enables, it is assumed that top level compiler project makes
+# stablehlo targets AND includes available (for example with `add_subdirectory` and `include_directories`).
+option(TORCH_MLIR_USE_EXTERNAL_STABLEHLO "Use stablehlo from top level project" OFF)
 
 option(TORCH_MLIR_OUT_OF_TREE_BUILD "Specifies an out of tree build" OFF)
 
@@ -142,7 +148,8 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/include)
 
 function(torch_mlir_target_includes target)
   set(_dirs
-    $<BUILD_INTERFACE:${MLIR_INCLUDE_DIRS}>
+    $<BUILD_INTERFACE:${MLIR_INCLUDE_DIR}>
+    $<BUILD_INTERFACE:${MLIR_GENERATED_INCLUDE_DIR}>
     $<BUILD_INTERFACE:${TORCH_MLIR_SOURCE_DIR}/include>
     $<BUILD_INTERFACE:${TORCH_MLIR_BINARY_DIR}/include>
   )
@@ -232,7 +239,8 @@ endif()
 # Getting this wrong results in building large parts of the stablehlo
 # project that we don't actually depend on. Further some of those parts
 # do not even compile on all platforms.
-if (TORCH_MLIR_ENABLE_STABLEHLO)
+# Only configure StableHLO if it isn't provided from a top-level project
+if (TORCH_MLIR_ENABLE_STABLEHLO AND NOT TORCH_MLIR_USE_EXTERNAL_STABLEHLO)
   set(STABLEHLO_BUILD_EMBEDDED ON)
   set(STABLEHLO_ENABLE_BINDINGS_PYTHON ON)
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo


shelkesagar29:
@powderluv can you please help review?

stellaraccident (approved):
@marbre for visibility (he was talking about possibly modernizing some of the cmake goo around this)

CMakeLists.txt
```diff
@@ -233,12 +241,17 @@ endif()
...
 if (TORCH_MLIR_ENABLE_STABLEHLO)
-  set(STABLEHLO_BUILD_EMBEDDED ON)
-  set(STABLEHLO_ENABLE_BINDINGS_PYTHON ON)
-  add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo
-    ${CMAKE_CURRENT_BINARY_DIR}/stablehlo
-    EXCLUDE_FROM_ALL)
-  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo)
+  if (NOT "${TORCH_MLIR_EXTERNAL_STABLEHLO_DIR}" STREQUAL "")
+    # Only include directories. It is assumed that stablehlo targets are made available by external project.
+    include_directories(${TORCH_MLIR_EXTERNAL_STABLEHLO_DIR})
```
marbre, line 246: I would have assumed that the external project also makes the includes available.

CMakeLists.txt
```diff
@@ -233,12 +241,17 @@ endif()
...
+  else()
```
marbre, line 247: Actually
```suggestion
  # Only configure StableHLO if it isn't provided from a top-level project
  if ("${TORCH_MLIR_EXTERNAL_STABLEHLO_DIR}" STREQUAL "")
```
should be sufficient. In that case I would probably replace the string with a boolean as it seems that the string isn't used elsewhere.

Depending on how the super project is set up (and pulls in torch-mlir and stablehlo) we could consider to add a `TORCH_BUILD_EMBEDDED` option, as some parts in the current out-of tree build config might be in responsibility of the super project as well.

CMakeLists.txt
shelkesagar29: Yes, this is fair assumption.

CMakeLists.txt
shelkesagar29: I have made suggested changes and added a boolean variable `TORCH_MLIR_USE_EXTERNAL_STABLEHLO`, set to OFF by default. As far as I can see, when `TORCH_MLIR_OUT_OF_TREE_BUILD` is enabled, MLIR package is searched by top level cmake.  Top level project can take care of this by passing `MLIR_DIR` and `LLVM_DIR` make options.

CMakeLists.txt
```diff
@@ -142,7 +148,8 @@ include_directories(${CMAKE_CURRENT_BINARY_DIR}/include)
 
 function(torch_mlir_target_includes target)
   set(_dirs
-    $<BUILD_INTERFACE:${MLIR_INCLUDE_DIRS}>
```
shelkesagar29, line 145: Using `MLIR_INCLUDE_DIRS`  creates an issue where `CMAKE_CURRENT_SOURCE_DIR` is added to include directories of target, leading to error as mentioned here https://stackoverflow.com/questions/76516562/cmake-why-interface-include-directories-can-be-populated-behind-my-back

One can confirm this by getting `INCLUDE_DIRECTORIES` target property of any target in Torch-MLIR.
This somehow created issue only when torch_mlir project is nested in some other project.

Change made here is really NFC since `MLIR_INCLUDE_DIRS ` contains both `MLIR_INCLUDE_DIR ` and `MLIR_GENERATED_INCLUDE_DIR ` but this does resolve source directory being added to include dir issue.

shelkesagar29:
@marbre can you double and give it a green light?

CMakeLists.txt
```diff
@@ -232,13 +239,14 @@ endif()
...
-    EXCLUDE_FROM_ALL)
-  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo)
+# Only configure StableHLO if it isn't provided from a top-level project
+if (TORCH_MLIR_ENABLE_STABLEHLO AND NOT TORCH_MLIR_USE_EXTERNAL_STABLEHLO)
+    set(STABLEHLO_BUILD_EMBEDDED ON)
+    set(STABLEHLO_ENABLE_BINDINGS_PYTHON ON)
+    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo
+      ${CMAKE_CURRENT_BINARY_DIR}/stablehlo
+      EXCLUDE_FROM_ALL)
+    include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo)
```
marbre, line 249:
```suggestion
  set(STABLEHLO_BUILD_EMBEDDED ON)
  set(STABLEHLO_ENABLE_BINDINGS_PYTHON ON)
  add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo
    ${CMAKE_CURRENT_BINARY_DIR}/stablehlo
    EXCLUDE_FROM_ALL)
  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/externals/stablehlo)
```
To keep it consistent the indent should be as before thus effectively only changing the if condition.

marbre (changes requested):

CMakeLists.txt
shelkesagar29: Done. Thanks for pointing it out.

CMakeLists.txt
marbre: Not following in depth here but this should be fine if the build passes.

marbre (approved):
Thanks!

PR #3928, opened 2024-12-19
[torch-mlir] add MPACT as an example torch-mlir based compiler

aartbik:
Rationale:
In addition to IREE and Blade, MPACT provides an MLIR-based example of a PyTorch compiler that uses TORCH-MLIR. It also illustrates propagating sparsity from sparse PyTorch into MLIR, a feature that is not widespread in DL compilers yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ Torch-MLIR is primarily a project that is integrated into compilers to bridge th
 
 * [IREE](https://github.com/iree-org/iree.git)
 * [Blade](https://github.com/alibaba/BladeDISC)
+* [MPACT](https://github.com/MPACT-ORG/mpact-compiler)
 
 While most of the project is exercised via testing paths, there are some ways that an end user can directly use the APIs without further integration:
 


reidtatge (approved):

stellaraccident (commented):
Is this project still active? The last substantive commit I see is from seven months ago. It is of a substantially different level of maturity and investment than the other two and I'd prefer that we only list projects that are active.

aartbik:
> Is this project still active? The last substantive commit I see is from seven months ago. It is of a substantially different level of maturity and investment than the other two and I'd prefer that we only list projects that are active.

I plan to actively work on this again. But regardless, I thought it would be nice illustration of a (perhaps still academic) way to propagate sparsity into an MLIR pipeline.

stellaraccident:
There isn't much there. Can we only use this page for advertising projects that are active and somewhat mature.

In the new year, we can refactor the docs to have sections for things that are more niche or up and coming.

aartbik:
> There isn't much there. Can we only use this page for advertising projects that are active and somewhat mature.
> 
> In the new year, we can refactor the docs to have sections for things that are more niche or up and coming.

Noted. I will remove.

PR #3932, opened 2024-12-23
Refreshes docs/development.md

bjacobgordon:
First PR, super stoked! 🥳

### Context
As part of Turbine Camp, I was referred to this doc to build Torch-MLIR on my Linux VM. It has a lot of potential, and I decided it to nudge it in a productive direction a bunch of times as I actually used the doc.

### Approach
I tried to make the commits here "atomic", i.e. each one is a singular unit of work such that the file "works better"/"makes more sense" post-commit than pre-commit. 

The majority of the commits in this PR is the story of how I "refactored" the doc step by step to look the way it does. I wrote them with Future Us in mind, so please don't squash haha.

Ready when you are, @marbre!

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,4 +1,4 @@
-# Checkout and build from source
+# Environment
 
 ## Check out the code
 
@@ -31,7 +31,6 @@ it with the following `apt` command on Ubuntu/Debian.
 sudo apt install python3-dev
 ```
 
-
 ## (Optional) Set up pre-commit
 
 This project uses [pre-commit](https://pre-commit.com/) in its CI. You can
@@ -45,15 +44,21 @@ pip install pre-commit
 pre-commit install
 ```
 
-## CMake Build
+# Building
+
+## With CMake
+
+### Configure for Building...
 
 Two setups are possible to build: in-tree and out-of-tree. The in-tree setup is the most straightforward, as it will build LLVM dependencies as well.
 
-### Building torch-mlir in-tree
+#### ...with LLVM "in-tree" using...
 
-The following command generates configuration files to build the project *in-tree*, that is, using llvm/llvm-project as the main build. This will build LLVM as well as torch-mlir and its subprojects.  On Windows, use the "Developer PowerShell for Visual Studio" to ensure that the compiler and linker binaries are in the `PATH` variable.
+The following commands generate configuration files to build the project *in-tree*, that is, using llvm/llvm-project as the main build. This will build LLVM as well as torch-mlir and its subprojects.  On Windows, use the "Developer PowerShell for Visual Studio" to ensure that the compiler and linker binaries are in the `PATH` variable.
 
-This requires `lld`, `clang`, `ccache`, and other dependencies for building `libtorch` / `PyTorch` wheels from source. If you run into issues because of these, try the [simplified build command](#simplified-build).
+##### ...Base + Optimization Options
+
+This requires `lld`, `clang`, `ccache`, and other dependencies for building `libtorch` / `PyTorch` wheels from source. If you run into issues because of these, try the [simplified build command](#base-options).
 
 ```shell
 cmake -GNinja -Bbuild \
@@ -85,7 +90,7 @@ cmake -GNinja -Bbuild \
   -DLIBTORCH_VARIANT=shared
 ```
 
-# Simplified build
+##### ...Base Options
 
 If you're running into issues with the above build command, consider using the following:
 
@@ -101,7 +106,7 @@ cmake -GNinja -Bbuild \
   externals/llvm-project/llvm
 ```
 
-#### Flags to enable MLIR debugging:
+#### Options to enable MLIR debugging
 
 * Enabling `--debug` and `--debug-only` flags (see [MLIR docs](https://mlir.llvm.org/getting_started/Debugging/)) for the `torch-mlir-opt` tool
 ```shell
@@ -109,7 +114,7 @@ cmake -GNinja -Bbuild \
   -DLLVM_ENABLE_ASSERTIONS=ON \
 ```
 
-#### Flags to run end-to-end tests:
+#### Options to run end-to-end tests
 
 Running the end-to-end execution tests locally requires enabling the native PyTorch extension features and the JIT IR importer, which depends on the
 former and defaults to `ON` if not changed:
@@ -118,7 +123,7 @@ former and defaults to `ON` if not changed:
   -DTORCH_MLIR_ENABLE_JIT_IR_IMPORTER=ON \
 ```
 
-### Building against a pre-built LLVM
+#### ...with LLVM "out-of-tree"
 
 If you have built llvm-project separately in the directory `$LLVM_INSTALL_DIR`, you can also build the project *out-of-tree* using the following command as template:
 ```shell
@@ -135,8 +140,7 @@ The same QoL CMake flags can be used to enable clang, ccache, and lld. Be sure t
 
 Be aware that the installed version of LLVM needs in general to match the committed version in `externals/llvm-project`. Using a different version may or may not work.
 
-
-### Build commands
+### Initiate Build
 
 After either cmake run (in-tree/out-of-tree), use one of the following commands to build the project:
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -122,6 +122,24 @@ Two setups are possible to build: in-tree and out-of-tree. The in-tree setup is 
 
 The following commands generate configuration files to build the project *in-tree*, that is, using llvm/llvm-project as the main build. This will build LLVM as well as torch-mlir and its subprojects.
 
+###### ...Base Options
+
+If you don't anticipate needing to frequently rebuild LLVM "in-tree", run:
+
+```shell
+cmake -GNinja -Bbuild \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DLLVM_ENABLE_ASSERTIONS=ON \
+  -DPython3_FIND_VIRTUALENV=ONLY \
+  -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
+  -DLLVM_TARGETS_TO_BUILD=host \
+  `# For building LLVM "in-tree"` \
+  externals/llvm-project/llvm \
+  -DLLVM_ENABLE_PROJECTS=mlir \
+  -DLLVM_EXTERNAL_PROJECTS="torch-mlir" \
+  -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR="$PWD"
+```
+
 ###### ...Base + Optimization Options
 This will build `libtorch` / `PyTorch` wheels from source and requires [the enablement mentioned earlier](#optional-enable-build-optimizations). If you encounter issues when you run this, try the [simplified build command](#base-options) instead.
 
@@ -154,24 +172,6 @@ cmake -GNinja -Bbuild \
   -DLIBTORCH_SRC_BUILD=ON \
   `# Set the variant of libtorch to build / link against. (shared|static and optionally cxxabi11)` \
   -DLIBTORCH_VARIANT=shared
-```
-
-###### ...Base Options
-
-If you don't anticipate needing to frequently rebuild LLVM "in-tree", run:
-
-```shell
-cmake -GNinja -Bbuild \
-  -DCMAKE_BUILD_TYPE=Release \
-  -DLLVM_ENABLE_ASSERTIONS=ON \
-  -DPython3_FIND_VIRTUALENV=ONLY \
-  -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
-  -DLLVM_TARGETS_TO_BUILD=host \
-  `# For building LLVM "in-tree"` \
-  externals/llvm-project/llvm \
-  -DLLVM_ENABLE_PROJECTS=mlir \
-  -DLLVM_EXTERNAL_PROJECTS="torch-mlir" \
-  -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR="$PWD"
 ```
 
 ##### ..."out-of-tree"
--- a/docs/development.md
+++ b/docs/development.md
@@ -169,23 +169,6 @@ cmake -GNinja -Bbuild \
   externals/llvm-project/llvm
 ```
 
-#### Options to enable MLIR debugging
-
-* Enabling `--debug` and `--debug-only` flags (see [MLIR docs](https://mlir.llvm.org/getting_started/Debugging/)) for the `torch-mlir-opt` tool
-```shell
-  -DCMAKE_BUILD_TYPE=RelWithDebInfo \ # or =Debug
-  -DLLVM_ENABLE_ASSERTIONS=ON \
-```
-
-#### Options to run end-to-end tests
-
-Running the end-to-end execution tests locally requires enabling the native PyTorch extension features and the JIT IR importer, which depends on the
-former and defaults to `ON` if not changed:
-```shell
-  -DTORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS=ON \
-  -DTORCH_MLIR_ENABLE_JIT_IR_IMPORTER=ON \
-```
-
 #### ...with LLVM "out-of-tree"
 
 If you have built llvm-project separately in the directory `$LLVM_INSTALL_DIR`, you can also build the project *out-of-tree* using the following command as template:
@@ -202,6 +185,23 @@ cmake -GNinja -Bbuild \
 The same QoL CMake flags can be used to enable clang, ccache, and lld. Be sure to have built LLVM with `-DLLVM_ENABLE_PROJECTS=mlir`.
 
 Be aware that the installed version of LLVM needs in general to match the committed version in `externals/llvm-project`. Using a different version may or may not work.
+
+#### Options to enable MLIR debugging
+
+* Enabling `--debug` and `--debug-only` flags (see [MLIR docs](https://mlir.llvm.org/getting_started/Debugging/)) for the `torch-mlir-opt` tool
+```shell
+  -DCMAKE_BUILD_TYPE=RelWithDebInfo \ # or =Debug
+  -DLLVM_ENABLE_ASSERTIONS=ON \
+```
+
+#### Options to run end-to-end tests
+
+Running the end-to-end execution tests locally requires enabling the native PyTorch extension features and the JIT IR importer, which depends on the
+former and defaults to `ON` if not changed:
+```shell
+  -DTORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS=ON \
+  -DTORCH_MLIR_ENABLE_JIT_IR_IMPORTER=ON \
+```
 
 ### Initiate Build
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -44,13 +44,13 @@
     python -m pip install -r requirements.txt -r torchvision-requirements.txt
     ```
 
-Also, ensure that you have the appropriate `python-dev` package installed
-to access the Python development libraries / headers. For example, you can install
-it with the following `apt` command on Ubuntu/Debian.
+1. Install the libraries and headers required for development
 
-```shell
-sudo apt install python3-dev
-```
+    - For Ubuntu or Debian, run:
+
+      ```shell
+      sudo apt install python3-dev
+      ```
 
 ## (Optional) Set up pre-commit
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -33,11 +33,13 @@
     python3 -m venv mlir_venv
     ```
 
-1. Activate the Python environment
+## Activate the Python environment
 
-    ```shell
-    source mlir_venv/bin/activate
-    ```
+```shell
+source mlir_venv/bin/activate
+```
+
+## Install Dependencies
 
 1. Get the latest version of pip
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -183,30 +183,33 @@ For workflows that demand frequent rebuilds, the following steps will allow you 
 ### Initiate Build
 
 1. [Configure the build](#configure-for-building) if you haven't already done so.
-1. Use one of the following commands to build the project:
-    - Build everything (including LLVM if in-tree)
+1. Append the base command:
+
+    ```shell
+    cmake --build build
+    ```
+
+1. **If you want to...**
+    - **...build _just_ torch-mlir (not all of LLVM)**, append:
 
       ```shell
-      cmake --build build
+       --target tools/torch-mlir/all
       ```
 
-    - Build just torch-mlir (not all of LLVM)
+      - NOTE: skipping this will result in _everything_ being built, which includes LLVM if configured as "in-tree"
+    - **...run unit tests**, append:
 
       ```shell
-      cmake --build build --target tools/torch-mlir/all
+       --target check-torch-mlir
       ```
 
-    - Run unit tests.
+    - **...run Python regression tests**, append:
 
       ```shell
-      cmake --build build --target check-torch-mlir
+       --target check-torch-mlir-python
       ```
 
-    - Run Python regression tests.
-
-      ```shell
-      cmake --build build --target check-torch-mlir-python
-      ```
+1. Execute the command with the desired targets, if any.
 
 ## Setup Python Environment to export the built Python packages
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -83,13 +83,22 @@ We recommend linting and formatting your commits _before_ the CI has a chance to
 
 ## With CMake
 
+### (Optional) Enable Build Optimizations
+
+For workflows that demand frequent rebuilds, the following steps will allow you to specify options during configuration that enable build optimizations.
+
+#### On Windows
+
+  1. Set up Developer PowerShell [for Visual Studio](https://learn.microsoft.com/en-us/visualstudio/ide/reference/command-prompt-powershell?view=vs-2022#start-in-visual-studio)
+  1. Ensure that the compiler and linker binaries are in the `PATH` variable.
+
 ### Configure for Building...
 
 Two setups are possible to build: in-tree and out-of-tree. The in-tree setup is the most straightforward, as it will build LLVM dependencies as well.
 
 #### ...with LLVM "in-tree" using...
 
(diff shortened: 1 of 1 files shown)


bjacobgordon:
@marbre The only thing I'm unsure about is [removing the libtorch flags](https://github.com/llvm/torch-mlir/pull/3932/commits/0a65cf26dbcf52709abcc8c4901786d45e9c326f)

stellaraccident:
Thanks! Marius/Jacob, please review and land as you see fit. I'll have a closer look when I'm back and can contribute any adjustments then.

docs/development.md
```diff
@@ -1,157 +1,215 @@
...
+          \
+          `# For building LLVM "in-tree"` \
+          externals/llvm-project/llvm \
+          -DLLVM_ENABLE_PROJECTS=mlir \
+          -DLLVM_EXTERNAL_PROJECTS="torch-mlir" \
+          -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR="$PWD"
+        ```
+
+        - NOTE: uses external/llvm-project/llvm as the main build, so LLVM will be built in additional to torch-mlir and its sub-projects.
+    1. **If building "out-of-tree" in separate directory**, append:
```
marbre, line 144: This is probably confusing as it mentions out-of-tree and separate directory in a single sentence without further explanation. The terminology is indeed not the best here, as LLVM out-of-tree means something different than CMake ouf-of-tree. I would rather leave as it was before and we can send a follow-up that addresses this.

marbre (commented):
Thanks for working through the documentation. I certainly understand that this needs improvement and should be restructured in some way. However, I am not sure if that is the structure we should aim for. Maybe we should split out some things, like the formatting, and land this separately. Next we could iterate on the docs, maybe in smaller chunks, before restructuring to much at once. Just thinking loud and other approaches have their benefits as well.

docs/development.md
```diff
@@ -1,15 +1,21 @@
 # Environment
 
-## Check out the code
+## Pull down the project
```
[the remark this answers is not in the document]
bjacobgordon: Ah yes, we should indeed avoid `pull`. And `checkout` refers to `git checkout`, so we should probably avoid that for the same reason!

1. "Clone the repository": "clone" only pertains to step 1, so let's find an umbrella term!
2. "Copy the repository": "repository" is "how the project is distributed", but we wanna convey "the project itself"
3. "Copy the project": and this is particular to our local machine so
4. "Copy the project locally"

What do you think of "Copy the project locally"?

docs/development.md
```diff
@@ -1,15 +1,21 @@
...
-git clone https://github.com/llvm/torch-mlir
-cd torch-mlir
-git submodule update --init --progress
-```
+1. ```shell
+   git clone https://github.com/llvm/torch-mlir
+   ```
+
+1. ```shell
+   cd torch-mlir
```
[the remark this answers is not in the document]
bjacobgordon: It does indeed!

GitHub offers the copy button by each snippet but using the multiline command might fail since there's some interstitial processes that take place. So, I ended up typing it out manually anyway line by line.

To make this quicker, I made each line separate. This way, they'd have independent copy buttons! It definitely felt easier to get through the docs when the copy button gave me the exact command I needed!

The hidden overhead for each new hire is now offloaded to the document itself.

I believe this is worth keeping, what do you think?

docs/development.md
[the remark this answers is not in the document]
marbre: I'd vote for 1. as the sections "Cloning a Project with Submodules" in the [docs](https://git-scm.com/book/en/v2/Git-Tools-Submodules) mentions `git submodule update --init --recursive` as an alternative if one didn't passed `--recurse-submodules` to `git clone`, but here we probably just don't want to get the submodules. As another example, the IREE [docs](https://iree.dev/building-from-source/getting-started/#quickstart-clone-and-build) also refer to this part as `clone` in the `clone and build` section thus `clone` is rather common.

docs/development.md
```diff
@@ -1,157 +1,215 @@
...
-Optionally, use `--depth=1` to make a shallow clone of the submodules.
-While this is running, you can already setup the Python venv and dependencies in the next step.
+1. Install Python VirtualEnvironment + MLIR variant
 
-## Setup your Python VirtualEnvironment and Dependencies
+    ```shell
+    python3 -m venv mlir_venv
+    ```
+
+## Activate the Python environment
```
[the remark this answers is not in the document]
bjacobgordon: 1. Will put "Create" instead of "Install", thanks!
2. I believe we should keep the separation for ergonomics. It didn't work when I copied the single big block into the command line. And later, we link the "Activate..." section rather than repeating that command elsewhere.

docs/development.md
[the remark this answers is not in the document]
bjacobgordon: Hmm so I'll think out loud with what you've taught me:
1. "Copy the project locally": both steps are about cloning of some sort, and cloning is a local thing
2. "Clone the project": This is done via the repo, so let's specify that instead
3. "Clone the repository": Ah okay, I understand why this is a better fit. Will change!

docs/development.md
```diff
@@ -87,6 +87,26 @@ We recommend linting and formatting your commits _before_ the CI has a chance to
 
 For workflows that demand frequent rebuilds, the following steps will allow you to specify options during configuration that enable build optimizations.
 
+#### On Ubuntu
+
+1. Install [Clang](https://clang.llvm.org/index.html)
+
+    ```shell
+    sudo apt install clang
```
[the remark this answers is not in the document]
bjacobgordon: Ah I see, how about "(Optional) Enable Quicker Rebuilds"?

docs/development.md
bjacobgordon: Okay, I'll reword it to be more like it was before!

docs/development.md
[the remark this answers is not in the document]
marbre: MLIR upstream has
> Using clang and lld speeds up the build, we recommend adding:

in the instructions (https://mlir.llvm.org/getting_started/). Thus maybe `(Optional) Speed up (re-)builds`?

docs/development.md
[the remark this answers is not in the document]
marbre: Seems I have missed that there is a link to the activate section.  If you prefer it, we can than have two separate comman blocks.

docs/development.md
```diff
@@ -1,157 +1,215 @@
...
-After either cmake run (in-tree/out-of-tree), use one of the following commands to build the project:
-```shell
-# Build just torch-mlir (not all of LLVM)
-cmake --build build --target tools/torch-mlir/all
+1. Install [LLD](https://lld.llvm.org/)
 
-# Run unit tests.
-cmake --build build --target check-torch-mlir
+    ```shell
+    sudo apt install lld
```
marbre, line 107: To shorten the instructions, what about having a combined
```shell
    sudo apt install clang ccache lld
```

docs/development.md
```diff
@@ -1,157 +1,215 @@
...
+1. Append (not "run") command with "common" options:
+
+    ```shell
+    cmake -GNinja -Bbuild \
+      `# Enables "--debug" and "--debug-only" flags for the "torch-mlir-opt" tool` \
+      -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+      -DLLVM_ENABLE_ASSERTIONS=ON \
+      -DPython3_FIND_VIRTUALENV=ONLY \
+      -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
+      -DLLVM_TARGETS_TO_BUILD=host
```
marbre, line 127: This "runs" and does not "appends" as there is no `\` at the last line. I still find the append / run a little confusing. Maybe having multiple blocks that "just work" is easier, even if it adds redundancy. I could image that @zjgarvey or @ScottTodd  might have an opinion on this.

docs/development.md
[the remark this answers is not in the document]
marbre: `cd torch-mlir` does not feel to deserve its own code block but it is okay if you prefer it.

docs/development.md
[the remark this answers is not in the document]
marbre: I think it would be fine to just mention it without needing to link back to the section.

docs/development.md
```diff
@@ -1,157 +1,215 @@
...
+      \
+      `# use clang`\
+      -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ \
+      `# use ccache to cache build results` \
+      -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+      `# use LLD to link in seconds, rather than minutes` \
+      `# if using clang <= 13, replace --ld-path=ld.lld with -fuse-ld=lld` \
+      -DCMAKE_EXE_LINKER_FLAGS_INIT="--ld-path=ld.lld" \
+      -DCMAKE_MODULE_LINKER_FLAGS_INIT="--ld-path=ld.lld" \
+      -DCMAKE_SHARED_LINKER_FLAGS_INIT="--ld-path=ld.lld"
```
ScottTodd, line 167: As of CMake 3.29, you can also use `-DCMAKE_LINKER_TYPE=lld` instead of these options: https://cmake.org/cmake/help/latest/variable/CMAKE_LINKER_TYPE.html

We have a tracking issue for recommending that in IREE at https://github.com/iree-org/iree/issues/18598

docs/development.md
ScottTodd: A page like https://llvm.org/docs/CMake.html may be useful to study and replicate, for consistency within LLVM.

* Start with the minimal command
* Call out frequently used variables
* List all other variables

In the IREE docs we have some recommended recipes at https://iree.dev/building-from-source/getting-started/. If the list of options gets longer and continues to branch based on usage style, I'd recommend looking at CMake presets or toolchain files.

docs/development.md
```diff
@@ -1,157 +1,215 @@
...
+    ```shell
+    cmake --build build
+    ```
+
+1. **If you want to...**
+    - **...build _just_ torch-mlir (not all of LLVM)**, append:
+
+      ```shell
+       --target tools/torch-mlir/all
+      ```
```
ScottTodd, line 197: Echoing some of the other comments about individual code blocks and appending, here I would spell out of the full command for easier copy/paste
```suggestion
1. **If you want to...**
    - **...build _just_ torch-mlir (not all of LLVM)**:

      ```shell
      ```
```

same with the other examples

docs/development.md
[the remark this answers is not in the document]
ScottTodd: +1 for not putting `cd torch-mlir` on its own, since that's the natural next step after cloning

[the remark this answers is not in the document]
bjacobgordon: I see what y'all mean! I'll `&&` it with the `clone` command.

docs/development.md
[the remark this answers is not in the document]
bjacobgordon: I'll combine it into the section above and just link that instead. The create and activate commands are now in the same block!

docs/development.md
bjacobgordon: Nice, will do!

docs/development.md
bjacobgordon: Nice, adding that!

docs/development.md
ScottTodd: Presets:

* https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html
* https://github.com/iree-org/iree/tree/main/build_tools/cmake/presets

Toolchain files can look similar (collections of settings in a file) but are load bearing for cross compilation:

* https://cmake.org/cmake/help/latest/variable/CMAKE_TOOLCHAIN_FILE.html
* https://github.com/iree-org/iree/blob/main/build_tools/cmake/riscv.toolchain.cmake

docs/development.md
bjacobgordon: I had tried that initially, but it didn't really convey that you could add all three targets, one on top of the other.

If I wanna run unit tests _and_ regression tests, I can copy-paste to get:
`cmake --build build --target check-torch-mlir --target check-torch-mlir-python`

From there, I can recall from command line history as much as I need!

But having them in separate blocks required me to click-drag-copy-paste.

What `cmake` wizardry do you recommend?

docs/development.md
bjacobgordon: @ScottTodd Would this allow us to build a macro to specify stuff like
- llvm in-tree: yes/no
- frequent rebuilds: yes/no
- enable tests: yes/no
such that it'll just use the relevant options given the flags?

docs/development.md
ScottTodd: > What `cmake` wizardry do you recommend?

For any high productivity setup (and not just validating that some set of options works or casually testing a project), I configure Visual Studio Code. My examples here are for IREE, but they should also apply for other projects like torch-mlir.

* JSON workspace/settings files let you track all of your options, toggling lines commented out as you want to flip between sets

    ```jsonc
    // iree.code-workspace
    {
	    "folders": [
		    {
			    "path": ".."
		    }
	    ],
	    "settings": {
		    "cmake.automaticReconfigure": false,
		    "cmake.buildDirectory": "${workspaceFolder}/../iree-build",
		    "cmake.configureArgs": [
			    "-DIREE_BUILD_PYTHON_BINDINGS=OFF",
			    // "-DIREE_BUILD_PYTHON_BINDINGS=ON",
			    "-DPython3_EXECUTABLE=${workspaceFolder}/.venv/bin/python",
			    //
			    "-DIREE_HAL_DRIVER_CUDA=ON",
			    "-DIREE_HAL_DRIVER_HIP=ON",
			    "-DIREE_HAL_DRIVER_LOCAL_SYNC=ON",
			    "-DIREE_HAL_DRIVER_LOCAL_TASK=ON",
			    "-DIREE_HAL_DRIVER_METAL=OFF",
			    "-DIREE_HAL_DRIVER_NULL=ON",
			    "-DIREE_HAL_DRIVER_VULKAN=ON",
			    //
			    "-DIREE_TARGET_BACKEND_VMVX=ON",
			    "-DIREE_TARGET_BACKEND_LLVM_CPU=ON",
			    "-DIREE_TARGET_BACKEND_LLVM_CPU_WASM=ON",
			    "-DIREE_TARGET_BACKEND_METAL_SPIRV=ON",
			    "-DIREE_TARGET_BACKEND_VULKAN_SPIRV=ON",
			    "-DIREE_TARGET_BACKEND_ROCM=ON",
			    "-DIREE_TARGET_BACKEND_CUDA=ON",
			    //
			    "-DIREE_HIP_TEST_TARGET_CHIP=gfx1100",
			    //
			    // ccache: https://ccache.dev/
			    "-DCMAKE_C_COMPILER_LAUNCHER=ccache",
			    "-DCMAKE_CXX_COMPILER_LAUNCHER=ccache",
			    //
			    "-DIREE_ENABLE_LLD=ON",
		    ],
		    "cmakeExplorer.buildDir": "${workspaceFolder}/../iree-build",
		    "cmakeExplorer.suiteDelimiter": "/",
		    "cmakeExplorer.extraCtestRunArgs": "--verbose",
	    }
    }
    ```

* GUI lets you select which target, including meta targets like `all`, to build
image

* Test explorer lets you choose which tests to run, including all tests under a given package/namespace
image

docs/development.md
ScottTodd: Sorta. Presets are a bit clunky when you get that far into the weeds though. They are good for high level switches that are mutually exclusive with one another.

bjacobgordon: @marbre 3 flags with 2 states each yields 8 separate combinations, meaning 8 separate (but closely related) code blocks. Because of this, I'm hesitant to go the "premade" route over the "composition" route. Maybe we could word it in a way that more effectively communicates how clusters of cmake options should be "tacked on"?

@ScottTodd Or maybe we just make a shell script for it until someone else has a chance to come along and make a decent preset for it?

docs/development.md
bjacobgordon: @ScottTodd That looks great! The development.md doc doesn't really specify anything for IDE config, probably because it's a "nice to have". Is that direction we wanna go? Or are we just wanting to get new contributors going with the "need to have" items?

docs/development.md
marbre: That is just one reason why I think the previous structure wasn't too bad. A rather complete command + a simplified one. All other options described separately and if you need those, you'd need to adjust your configuration command. In contrast the patch forces users to always copy paste multiple blocks to get a working configuration.

bjacobgordon: Ah I see what you mean! I've made an adjustment that combines the "common" options with the mandatory LLVM options. This way, devs will be able to copy of those blocks and immediately run it if build speed and testing are not relevant to them.

bjacobgordon:
Okay, @marbre, I think we're almost there! How's it look now?

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
-```
+1. Get the latest version of pip
 
-## CMake Build
+    ```shell
+    python -m pip install --upgrade pip
+    ```
 
-Two setups are possible to build: in-tree and out-of-tree. The in-tree setup is the most straightforward, as it will build LLVM dependencies as well.
+    - Some older pip installs may not be able to handle the recent PyTorch deps
```
marbre, line 39:
```suggestion
Some older pip installs may not be able to handle the recent PyTorch deps
```
I don't see the need for a bullet point here. Same at some places below.

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
-cmake --build build --target check-torch-mlir
+        ```shell
+        cmake -GNinja -Bbuild \
+          `# Enables "--debug" and "--debug-only" flags for the "torch-mlir-opt" tool` \
+          -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+          -DLLVM_ENABLE_ASSERTIONS=ON \
+          -DPython3_FIND_VIRTUALENV=ONLY \
+          -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
+          -DLLVM_TARGETS_TO_BUILD=host \
+          \
```
marbre, line 106:
```suggestion
```
The additional line looks odd to me.

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
-cmake --build build --target tools/torch-mlir/all
+1. [Activate the Python environment](#set-up-the-python-environment)
+1. Choose command with relevant LLVM options:
+    1. **If building "in-tree"**, run/append:
 
-# Run unit tests.
-cmake --build build --target check-torch-mlir
+        ```shell
+        cmake -GNinja -Bbuild \
+          `# Enables "--debug" and "--debug-only" flags for the "torch-mlir-opt" tool` \
```
marbre, line 100: ?

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
-```
+        ```shell
+        cmake -GNinja -Bbuild \
+          `# Enables "--debug" and "--debug-only" flags for the "torch-mlir-opt" tool` \
+          -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+          -DLLVM_ENABLE_ASSERTIONS=ON \
+          -DPython3_FIND_VIRTUALENV=ONLY \
+          -DMLIR_ENABLE_BINDINGS_PYTHON=ON \
+          -DLLVM_TARGETS_TO_BUILD=host \
+          \
```
marbre, line 125:
```suggestion
```

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
+          `# For building LLVM "out-of-tree"` \
```
marbre, line 126: Why the backticks?

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
+          -DMLIR_DIR="$LLVM_INSTALL_DIR/lib/cmake/mlir/" \
+          -DLLVM_DIR="$LLVM_INSTALL_DIR/lib/cmake/llvm/"
+        ```
+
+        - Be sure to have built LLVM with `-DLLVM_ENABLE_PROJECTS=mlir`.
```
marbre, line 131: I would rephrase to, make enable building MLIR.

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
+        - Be aware that the installed version of LLVM needs in general to match the committed version in `externals/llvm-project`. Using a different version may or may not work.
```
marbre, line 132: It's rather the MLIR commit than LLVM.

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
+    - [About MLIR debugging](https://mlir.llvm.org/getting_started/Debugging/)
+1. **If you anticipate needing to frequently rebuild LLVM**, append:
+
+    ```shell
+      \
```
marbre, line 137: ?

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
+      `# use LLD to link in seconds, rather than minutes` \
+      -DCMAKE_LINKER_TYPE=lld
+    ```
+
+    - This will build `libtorch` / `PyTorch` wheels from source and requires [the enablement mentioned earlier](#optional-enable-quicker-builds).
+    - If these options cause issues, just skip them for now.
+1. **If you need to enable local end-to-end tests**, append:
+
+    ```shell
+      \
```
marbre, line 151:
```suggestion
```

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
+      `# use clang`\
+      -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ \
+      `# use ccache to cache build results` \
+      -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
+      `# use LLD to link in seconds, rather than minutes` \
+      -DCMAKE_LINKER_TYPE=lld
+    ```
+
+    - This will build `libtorch` / `PyTorch` wheels from source and requires [the enablement mentioned earlier](#optional-enable-quicker-builds).
```
marbre, line 146: This seems misplaced here. Seems the part on `LIBTORCH_SRC_BUILD` that this refers to was lost?

docs/development.md
```diff
@@ -1,157 +1,191 @@
...
+    - NOTE: The JIT IR importer depends on the native PyTorch extension features and defaults to `ON` if not changed.
+1. Run assembled command (once you've appended the options pertaining to your workflow)
+
+### Initiate Build
+
+1. [Configure the build](#configure-for-building) if you haven't already done so.
+1. Append the base command:
+
+    ```shell
+    cmake --build build
```
marbre, line 165: To make copy pasting easier I would call with complete `cmake --build ..` commands here. Splitting this up prevents what you intended with your changes at other places to allow to easily copy.

marbre (commented):
Just a quick review, traveling, but leaving a few comments.

docs/development.md
bjacobgordon: This phrase was originally a comment in a command line script, so how do we communicate "you can gloss over this line"? Maybe prepending "NOTE:" instead of having the bullet point?

docs/development.md
bjacobgordon: I agree. Wanting to somehow communicate at-a-glance "this chunk of options is the key difference between the two otherwise code blocks". What's another way we could do that?

docs/development.md
bjacobgordon: I'm guessing this is not what you had in mind haha. I'll need a little more direction here, which parts seem odd?

docs/development.md
bjacobgordon: It's to ensure that the comments end in time for the `\` character treated to actually be used by the command line.

docs/development.md
bjacobgordon: Couldn't remember at first, but it looks like all I did was move it to a different spot haha. It's worded the same way it was on `main`.

Could we save rephrasing this for a subsequent PR?

docs/development.md
bjacobgordon: This is how it's worded on `main`, I think all I did was move it and add a bullet point.

Could we defer rephrasing this to a subsequent PR?

docs/development.md
bjacobgordon: This is what you had meant, right? All you need to get going is copying, pasting and running one of the "base" commands. Tacking this on before running is optional, right?

docs/development.md
bjacobgordon: CMake issued a warning upon fresh build:
> Manually-specified variables were not used by the project: LIBTORCH_CACHE LIBTORCH_SRC_BUILD LIBTORCH_VARIANT

So, I removed those but didn't remove this mention! I'll update it.

docs/development.md
bjacobgordon: Will make code blocks mutually exclusive

bjacobgordon:
Stacked onto #4048

bjacobgordon:
Stacked onto #4051

bjacobgordon:
Stacked onto #4053

bjacobgordon:
Stacked onto #4068

bjacobgordon:
Stacked onto #4074

bjacobgordon:
Inspired #4104, restacked

bjacobgordon:
Inspired #4110, restacked

bjacobgordon:
Inspired #4111, restacked.

bjacobgordon:
Inspired #4114, restacked

bjacobgordon:
Inspired #4114, restacked

bjacobgordon:
Inspired #4121, restacked

bjacobgordon:
Inspired #4136, restacked

bjacobgordon:
Inspired #4139, restacked

bjacobgordon:
Inspired #4149, restacked

bjacobgordon:
Inspired #4154, restacked

bjacobgordon:
Inspired #4160, restacked

bjacobgordon:
Inspired #4175, restacked

docs/development.md
```diff
@@ -95,6 +95,7 @@ sudo apt install clang ccache lld
 
 #### Configure for Building
 
+1. **If you haven't already**, [activate the Python environment](#set-up-the-python-environment)
```
marbre, line 98:
```suggestion
1. *If you haven't already*, [activate the Python environment](#set-up-the-python-environment)
```
maybe italic instead of bold?

marbre (approved):

docs/development.md
marbre: Well it looks like a very strong and not too kind invitation but I don't have too strong feelings about this. Probably depends on prior experience and I am rather used to _always_ create and use a venv thus the reminder could be more gentle without shouting at me ;) But as mentioned, no super strong feelings.

PR #3941, opened 2025-01-06
Use op.dtype to create aten.empty.memory_format during decomposition.

sahas3:
Prior to the change in this PR `torch-mlir-opt --convert-torch-to-linalg` was running into the following error:

```
error: 'tensor.cast' op operand type 'tensor<200x200x26xf32>' and result type 'tensor<200x200x26xf64>' are cast incompatible
    %1 = torch.aten.empty.memory_format %0, %none, %none, %none, %false, %none : !torch.list<int>, !torch.none, !torch.none, !torch.none, !torch.bool, !torch.none -> !torch.vtensor<[200,200,26],f64>
         ^
note: see current operation: %12 = "tensor.cast"(%11) : (tensor<200x200x26xf32>) -> tensor<200x200x26xf64
```

This is because when `dtype` of the `aten.empty.memory_format` is `none`, by default `f32` was being selected as the element type of the resulting tensor which doesn't match with the actual element type of the result.

### DIFF
--- a/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
+++ b/lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
@@ -7087,9 +7087,16 @@ public:
         Torch::ListType::get(Torch::IntType::get(op.getContext()));
     Value sizeList =
         rewriter.create<AtenSizeOp>(op.getLoc(), sizeListType, op.getSelf());
+
+    FailureOr<Value> dtype = getDtypeFromOp(rewriter, op);
+    if (failed(dtype)) {
+      return rewriter.notifyMatchFailure(
+          op, "could not determine dtype from the op.");
+    }
+
     rewriter.replaceOpWithNewOp<AtenEmptyMemoryFormatOp>(
-        op, op.getType(), sizeList, op.getDtype(), op.getLayout(),
-        op.getDevice(), op.getPinMemory(), op.getMemoryFormat());
+        op, op.getType(), sizeList, *dtype, op.getLayout(), op.getDevice(),
+        op.getPinMemory(), op.getMemoryFormat());
     return success();
   }
 };
@@ -7838,18 +7845,13 @@ class DecomposeAtenNewEmptyOp : public OpRewritePattern<AtenNewEmptyOp> {
   LogicalResult matchAndRewrite(AtenNewEmptyOp op,
                                 PatternRewriter &rewriter) const override {
     Value noneVal = rewriter.create<ConstantNoneOp>(op.getLoc());
-    Value dtype = op.getDtype();
-    if (isa<Torch::NoneType>(dtype.getType())) {
-      BaseTensorType tensorType = cast<BaseTensorType>(op.getSelf().getType());
-      if (!tensorType.hasDtype()) {
-        return rewriter.notifyMatchFailure(
-            op, "expected input tensor to have a dtype");
-      }
-      dtype =
-          getDtypeIntValueForType(rewriter, op.getLoc(), tensorType.getDtype());
+    FailureOr<Value> dtype = getDtypeFromOp(rewriter, op);
+    if (failed(dtype)) {
+      return rewriter.notifyMatchFailure(
+          op, "could not determine dtype from the op.");
     }
     rewriter.replaceOpWithNewOp<AtenEmptyMemoryFormatOp>(
-        op, op.getType(), op.getSize(), dtype, op.getLayout(), op.getDevice(),
+        op, op.getType(), op.getSize(), *dtype, op.getLayout(), op.getDevice(),
         op.getPinMemory(), /*memoryFormat=*/noneVal);
     return success();
   }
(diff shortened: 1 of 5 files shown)


sahas3:
Hi @ramiro050, I see that I am reverting [your change](https://github.com/llvm/torch-mlir/commit/f85f5799e45b1ea79ff46dd87c1c5428adae34e9#diff-88437d1f7c10e96d2664a7bb84af80a25776560589c9b95139839a3eeca67f88L229) here. It's not clear to me why the `resultElementType` was changed in that commit. I verified that the test you added passes with my change. Can you please take a look at this PR? Thanks!

sahas3:
Hello @ramiro050 and @vivekkhandelwal1 can you please review this PR when you have a chance? Thanks!

test/Conversion/TorchToLinalg/basic.mlir
```diff
@@ -354,3 +354,20 @@ func.func @torch.aten.transpose$basic(%arg0: !torch.vtensor<[4,3],f32>) -> !torc
...
+// CHECK:           return %[[RES]] : !torch.vtensor<[200,200,26],f64>
+func.func @torch.aten.empty.memory_format$noneDtype() -> !torch.vtensor<[200,200,26],f64> attributes {torch.assume_strict_symbolic_shapes} {
+    %int200 = torch.constant.int 200
+    %int26 = torch.constant.int 26
+    %false = torch.constant.bool false
+    %none = torch.constant.none
+    %0 = torch.prim.ListConstruct %int200, %int200, %int26 : (!torch.int, !torch.int, !torch.int) -> !torch.list<int>
+    %1 = torch.aten.empty.memory_format %0, %none, %none, %none, %false, %none : !torch.list<int>, !torch.none, !torch.none, !torch.none, !torch.bool, !torch.none -> !torch.vtensor<[200,200,26],f64>
+    return %1 : !torch.vtensor<[200,200,26],f64>
+}
```
vivekkhandelwal1, line 373: Hi @sahas3, the above test case is an invalid test since the tensor element type is f32 when none is specifed for the `aten.empty.memory_format` op.

I just tried it. See:

```
>>> a = torch.ops.aten.empty.memory_format([2, 3])
>>> a
tensor([[-3.4054e+29,  4.2543e-41, -3.4085e+29],
        [ 4.2543e-41, -3.4086e+29,  4.2543e-41]])
>>> a.dtype
torch.float32
```

vivekkhandelwal1: Also, the changes that you've done are not correct and hence this patch should not be merged.

sahas3: Great flag, @vivekkhandelwal1. I missed that `torch.ops.aten.empty.memory_format` produces `f32` for `dtype=none` in PyTorch as I assumed the `aten.empty.memory_format` op being produced during conversion of `ExportedProgram` to `torch` dialect was correct. Upon further investigation, I think the bug is instead in the `DecomposeComplexOps` pass where we decompose different torch ops to `aten.empty.memory_format`. 

I have pushed new changes, reverting the original change that addresses this issue. Can you please take another look at these new changes? Thanks!

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -7065,9 +7065,16 @@ class DecomposeAtenEmptyLikeOp : public OpRewritePattern<AtenEmptyLikeOp> {
         Torch::ListType::get(Torch::IntType::get(op.getContext()));
     Value sizeList =
         rewriter.create<AtenSizeOp>(op.getLoc(), sizeListType, op.getSelf());
+
+    FailureOr<Value> dtype = getDtypeFromOp(rewriter, op);
+    if (failed(dtype)) {
+      return rewriter.notifyMatchFailure(
+          op, "could not determine dtype from the op.");
+    }
+
```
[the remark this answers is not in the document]
sahas3: Hi @vivekkhandelwal1, my understanding is that `dtype` specifies the type of the returned tensor of any op. For ops like `empty_like` this `dtype` if not specified defaults to that of input tensor as per: https://pytorch.org/docs/stable/generated/torch.empty_like.html. So looking at the `torch.empty_like` op itself I think type of result tensor and input tensor will be the same if `dtype` is not specified.

On the other hand, for ops like `empty_strided`, `dtype` defaults to a global value as per https://pytorch.org/docs/stable/generated/torch.empty_strided.html. In this case, there's no input tensor but the result tensor should already have the type set correctly when it's in the `torch` dialect.

So for both cases, I think we can generalize the construction of `AtenEmptyMemoryFormatOp` based on the `dtype` of the result tensor when `dtype` for the op to be rewritten is not specified. We don't need to check for the input tensor `dtype` though I can add an assert that types of input and output tensors are same for ops like `empty_like`. The reason to generalize to using result tensor dtype is because ops like `empty_strided` don't have an input tensor and hence no `op.getSelf()` method on those ops.

I hope this clarifies the changes. I'll add a gist of this as comment as well for future reference.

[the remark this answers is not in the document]
vivekkhandelwal1: Thanks for the explanation.

vivekkhandelwal1 (approved):
LGTM

PR #3943, opened 2025-01-07
[torch-mlir] Support lowering of aten constraint ops

praveen-g-ctt:
1. aten::sym_constrain_range
2. aten::sym_constrain_range_for_size
3. aten::_assert_scalar

### DIFF
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -25,6 +25,7 @@
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/ADT/APSInt.h"
 #include <numeric>
+#include <string>
 #include <type_traits>
 
 using namespace mlir;
@@ -3564,6 +3565,98 @@ public:
 };
 } // namespace
 
+namespace {
+class ConvertSymConstrainRangeOp
+    : public OpConversionPattern<AtenSymConstrainRangeOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(AtenSymConstrainRangeOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    auto loc = op.getLoc();
+    auto min = op.getMin();
+    auto max = op.getMax();
+
+    auto minOp = min.getDefiningOp();
+    auto maxOp = max.getDefiningOp();
+
+    if (!minOp || !maxOp)
+      return op.emitError("Unimplemented: Non constant min/max values");
+
+    int64_t minValue = std::numeric_limits<int64_t>::min();
+    int64_t maxValue = std::numeric_limits<int64_t>::max();
+
+    Type operandType = rewriter.getI64Type();
+
+    if (!isa<Torch::ConstantNoneOp>(minOp))
+      if (!matchPattern(min, m_TorchConstantInt(&minValue)))
+        return rewriter.notifyMatchFailure(
+            op, "Expected min value to be constant integer");
+
+    if (!isa<Torch::ConstantNoneOp>(maxOp))
+      if (!matchPattern(max, m_TorchConstantInt(&maxValue)))
+        return rewriter.notifyMatchFailure(
+            op, "Expected max value to be constant integer");
+
+    if (maxValue < minValue) {
+      std::string errorMsg =
+          "Max must be greater than or equal to min, got min = " +
+          std::to_string(minValue) + ", max = " + std::to_string(maxValue);
+      return op.emitError(errorMsg);
+    }
+
+    min = getConstant(rewriter, loc, minValue, operandType);
+    max = getConstant(rewriter, loc, maxValue, operandType);
+
+    // Check min <= size <= max
+
+    // FIXME:: Skip the below checks if constraint ops are already inserted as
+    // part of symbol expr evaluation
+    auto checkMin = createLessThanOrEqual(rewriter, loc, operandType, min,
+                                          adaptor.getSize());
+    auto checkMax = createLessThanOrEqual(rewriter, loc, operandType,
+                                          adaptor.getSize(), max);
+    auto compareVal = rewriter.create<arith::AndIOp>(loc, checkMin, checkMax);
+
+    std::string assertMessage = "Invalid value range for size between [" +
+                                std::to_string(minValue) + ", " +
+                                std::to_string(maxValue) + "]";
+    rewriter.create<cf::AssertOp>(loc, compareVal,
+                                  rewriter.getStringAttr(assertMessage));
+
+    rewriter.eraseOp(op);
+    return success();
+  }
+};
+} // namespace
+
+namespace {
+class ConvertAssertScalarOp : public OpConversionPattern<Aten_AssertScalarOp> {
+public:
+  using OpConversionPattern::OpConversionPattern;
+  LogicalResult
+  matchAndRewrite(Aten_AssertScalarOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    auto assertCond = convertScalarToDtype(
+        rewriter, op.getLoc(), adaptor.getSelf(), rewriter.getI1Type());
+
+    std::string assertMessage;
+    if (!matchPattern(op.getAssertMsg(), m_TorchConstantStr(assertMessage)))
+      return rewriter.notifyMatchFailure(
+          op, "Assert message must be a constant string");
+
+    rewriter.replaceOpWithNewOp<cf::AssertOp>(op, assertCond, assertMessage);
+    return success();
+  }
+};
+} // namespace
+
 void mlir::torch::torch_to_linalg::populateUncategorizedPatternsAndLegality(
     TypeConverter &typeConverter, RewritePatternSet &patterns,
     ConversionTarget &target) {
@@ -3626,4 +3719,8 @@ void mlir::torch::torch_to_linalg::populateUncategorizedPatternsAndLegality(
   patterns.add<ConvertAtenLinalgDetOp>(typeConverter, context);
   target.addIllegalOp<AtenPolarOp>();
   patterns.add<ConvertAtenPolarOp>(typeConverter, context);
+  target.addIllegalOp<AtenSymConstrainRangeOp>();
+  patterns.add<ConvertSymConstrainRangeOp>(typeConverter, context);
+  target.addIllegalOp<Aten_AssertScalarOp>();
+  patterns.add<ConvertAssertScalarOp>(typeConverter, context);
 }
--- a/lib/Conversion/TorchToLinalg/Uncategorized.cpp
+++ b/lib/Conversion/TorchToLinalg/Uncategorized.cpp
@@ -21,6 +21,7 @@
 #include "torch-mlir/Conversion/TorchToLinalg/Utils.h"
 #include "torch-mlir/Conversion/Utils/Utils.h"
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
+#include "torch-mlir/Dialect/Torch/IR/TorchTypes.h"
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
 #include "llvm/ADT/APSInt.h"
@@ -3580,23 +3581,17 @@ public:
     auto min = op.getMin();
     auto max = op.getMax();
 
-    auto minOp = min.getDefiningOp();
-    auto maxOp = max.getDefiningOp();
-
-    if (!minOp || !maxOp)
-      return op.emitError("Unimplemented: Non constant min/max values");
-
     int64_t minValue = std::numeric_limits<int64_t>::min();
     int64_t maxValue = std::numeric_limits<int64_t>::max();
 
     Type operandType = getTypeConverter()->convertType(op.getSize().getType());
 
-    if (!isa<Torch::ConstantNoneOp>(minOp))
+    if (!isa<Torch::NoneType>(min.getType()))
       if (!matchPattern(min, m_TorchConstantInt(&minValue)))
         return rewriter.notifyMatchFailure(
             op, "Expected min value to be constant integer");
 
-    if (!isa<Torch::ConstantNoneOp>(maxOp))
+    if (!isa<Torch::NoneType>(max.getType()))
       if (!matchPattern(max, m_TorchConstantInt(&maxValue)))
         return rewriter.notifyMatchFailure(
             op, "Expected max value to be constant integer");
@@ -3621,37 +3616,13 @@ public:
         loc, arith::CmpIPredicate::sle, adaptor.getSize(), max);
     auto compareVal = rewriter.create<arith::AndIOp>(loc, checkMin, checkMax);
 
-    std::string assertMessage = "Invalid value range for size between [" +
+    std::string assertMessage = "Size constraint failed. Expected range: [" +
                                 std::to_string(minValue) + ", " +
                                 std::to_string(maxValue) + "]";
     rewriter.create<cf::AssertOp>(loc, compareVal,
                                   rewriter.getStringAttr(assertMessage));
 
     rewriter.eraseOp(op);
-    return success();
-  }
-};
-} // namespace
-
-namespace {
-class ConvertAssertScalarOp : public OpConversionPattern<Aten_AssertScalarOp> {
-public:
-  using OpConversionPattern::OpConversionPattern;
-  LogicalResult
-  matchAndRewrite(Aten_AssertScalarOp op, OpAdaptor adaptor,
-                  ConversionPatternRewriter &rewriter) const override {
-    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
-      return failure();
-
-    auto assertCond = convertScalarToDtype(
-        rewriter, op.getLoc(), adaptor.getSelf(), rewriter.getI1Type());
-
-    std::string assertMessage;
-    if (!matchPattern(op.getAssertMsg(), m_TorchConstantStr(assertMessage)))
-      return rewriter.notifyMatchFailure(
-          op, "Assert message must be a constant string");
-
-    rewriter.replaceOpWithNewOp<cf::AssertOp>(op, assertCond, assertMessage);
     return success();
   }
 };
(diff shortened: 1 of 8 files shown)


projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -6453,3 +6453,66 @@ def forward(self, x):
...
+class AtenSymConstrainRangeForSize(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([None, ([-1], torch.int, True)])
+    def forward(self, x):
+        a = x.item()
+        torch._check_is_size(a)
+        # max should be >= 2
```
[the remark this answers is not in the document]
praveen-g-ctt: Updated the comment in the test. This check is from the Aten native implementation. 

https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/Constraints.cpp#L66

test/Conversion/TorchToLinalg/constraints.mlir
```diff
@@ -0,0 +1,82 @@
...
+// CHECK:           cf.assert %[[VAL_10]], "Invalid value range for size between [0, 9223372036854775807]"
+// CHECK:           %[[VAL_11:.*]] = arith.constant 0 : i64
+// CHECK:           %[[VAL_12:.*]] = arith.constant 7 : i64
+// CHECK:           %[[VAL_13:.*]] = arith.cmpi ule, %[[VAL_11]], %[[VAL_5]] : i64
+// CHECK:           %[[VAL_14:.*]] = arith.cmpi ule, %[[VAL_5]], %[[VAL_12]] : i64
+// CHECK:           %[[VAL_15:.*]] = arith.andi %[[VAL_13]], %[[VAL_14]] : i1
+// CHECK:           cf.assert %[[VAL_15]], "Invalid value range for size between [0, 7]"
+// CHECK:           return %[[VAL_4]] : !torch.int
+
+module {
```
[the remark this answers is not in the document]
praveen-g-ctt: Done

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -3564,6 +3565,98 @@ class ConvertAtenPolarOp : public OpConversionPattern<AtenPolarOp> {
...
+        loc, arith::CmpIPredicate::sle, min, adaptor.getSize());
+    auto checkMax = rewriter.create<arith::CmpIOp>(
+        loc, arith::CmpIPredicate::sle, adaptor.getSize(), max);
+    auto compareVal = rewriter.create<arith::AndIOp>(loc, checkMin, checkMax);
+
+    std::string assertMessage = "Invalid value range for size between [" +
+                                std::to_string(minValue) + ", " +
+                                std::to_string(maxValue) + "]";
+    rewriter.create<cf::AssertOp>(loc, compareVal,
+                                  rewriter.getStringAttr(assertMessage));
```
zjgarvey, line 3628: If I remember correctly, this op was getting generated in some sdxl model we were importing to mlir? Are the `sym_constrain_range_for_size` ops intended to provide symbolic shape constrains similar to `symbolic_int` and `bind_symbolic_shape`? For example, is it desirable for us to try to canonicalize patterns like: 

```mlir
func.func @sym_constrain(%arg0 : !torch.vtensor<[?,?],f32>) -> !torch.vtensor<[?],f32> {
    %int10 = torch.constant.int 10
    %int7 = torch.constant.int 7
    %int1 = torch.constant.int 1
    %int0 = torch.constant.int 0
    %int-1 = torch.constant.int -1
    %none = torch.constant.none
    %dim0 = torch.aten.size.int %arg0, %int0 : !torch.vtensor<[?,?],f32>, !torch.int -> !torch.int
    %dim1 = torch.aten.size.int %arg0, %int1 : !torch.vtensor<[?,?],f32>, !torch.int -> !torch.int
    torch.aten.sym_constrain_range %dim0, %int1, %int10 : !torch.int, !torch.int, !torch.none
    torch.aten.sym_constrain_range %dim1, %int1, %int7 : !torch.int, !torch.int, !torch.int
    %list = torch.prim.ListConstruct %int-1 : (!torch.int) -> !torch.list<int>
    %view = torch.aten.view %arg0, %list : !torch.vtensor<[?,?],f32>, !torch.list<int> -> !torch.vtensor<[?],f32>
    return %view : !torch.vtensor<[?],f32>
}
```

into something like:

```mlir
func.func @sym_constrain(%arg0 : !torch.vtensor<[?,?,3],f32>) -> !torch.vtensor<[?],f32> {
    %int-1 = torch.constant.int -1
    %none = torch.constant.none
    %symbolic_int0 = torch.symbolic_int "s0" {min_val = 1, max_val = 10} : !torch.int
    %symbolic_int1 = torch.symbolic_int "s1" {min_val = 1, max_val = 7} : !torch.int
    torch.bind_symbolic_shape %arg0, [%symbolic_int0, %symbolic_int1], affine_map<()[s0, s1] -> (s0, s1, 3)> : !torch.vtensor<[?,?,3],f32>
    %list = torch.prim.ListConstruct %int-1 : (!torch.int) -> !torch.list<int>
    %view = torch.aten.view %arg0, %list : !torch.vtensor<[?,?,3],f32>, !torch.list<int> -> !torch.vtensor<[?],f32>
    return %view : !torch.vtensor<[?],f32>
}
```

I think it would be fine to have these kinds of ops lower to asserts as a backup, but IREE can take advantage of the symbolic shape ops by converting to IR like:

```mlir
    %c3 = arith.constant 3 : index
    %0 = hal.buffer_view.dim<%arg0 : !hal.buffer_view>[0] : index
    %1 = hal.buffer_view.dim<%arg0 : !hal.buffer_view>[1] : index
    %2 = hal.tensor.import wait(%arg1) => %arg0 : !hal.buffer_view -> tensor<?x?x3xf32>{%0, %1}
    %3 = util.assume.int %0<umin = 1, umax = 10> : index
    %4 = util.assume.int %1<umin = 1, umax = 7> : index
    %5 = arith.muli %3, %4 : index
    %6 = arith.muli %5, %c3 : index
    %7 = flow.tensor.reshape %2 : tensor<?x?x3xf32>{%3, %4} -> tensor<?xf32>{%6}
    %8 = hal.tensor.barrier join(%7 : tensor<?xf32>) => %arg2 : !hal.fence
    %9 = hal.tensor.export %8 : tensor<?xf32>{%6} -> !hal.buffer_view
    util.return %9 : !hal.buffer_view
```

Which can then be used to inform the compiler about appropriate intrinsics/tiling sizes for things like matmuls.

This comment doesn't necessarily need to be addressed here, but if we can get some context as to why these ops are introduced in examples, it might be helpful for improvements in the near future.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -3564,6 +3565,98 @@ class ConvertAtenPolarOp : public OpConversionPattern<AtenPolarOp> {
...
+                  ConversionPatternRewriter &rewriter) const override {
+    if (failed(verifyLinalgCompatibleTypes(op, rewriter)))
+      return failure();
+
+    auto loc = op.getLoc();
+    auto min = op.getMin();
+    auto max = op.getMax();
+
+    auto minOp = min.getDefiningOp();
+    auto maxOp = max.getDefiningOp();
```
zjgarvey, line 3584: You could instead check for `none` by:

```cpp
isa<Torch::NoneType>(min.getType())
```

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -3564,6 +3565,98 @@ class ConvertAtenPolarOp : public OpConversionPattern<AtenPolarOp> {
...
+
+    // FIXME:: Skip the below checks if constraint ops are already inserted as
+    // part of symbol expr evaluation
+    auto checkMin = rewriter.create<arith::CmpIOp>(
+        loc, arith::CmpIPredicate::sle, min, adaptor.getSize());
+    auto checkMax = rewriter.create<arith::CmpIOp>(
+        loc, arith::CmpIPredicate::sle, adaptor.getSize(), max);
+    auto compareVal = rewriter.create<arith::AndIOp>(loc, checkMin, checkMax);
+
+    std::string assertMessage = "Invalid value range for size between [" +
```
zjgarvey, line 3624: nit:

```suggestion
    std::string assertMessage = "Size constraint failed. Expected range: [" +
```

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -3564,6 +3565,98 @@ class ConvertAtenPolarOp : public OpConversionPattern<AtenPolarOp> {
...
+    std::string assertMessage;
+    if (!matchPattern(op.getAssertMsg(), m_TorchConstantStr(assertMessage)))
+      return rewriter.notifyMatchFailure(
+          op, "Assert message must be a constant string");
+
+    rewriter.replaceOpWithNewOp<cf::AssertOp>(op, assertCond, assertMessage);
+    return success();
+  }
+};
+} // namespace
```
zjgarvey, line 3658: I believe there is already a `torch.runtime.assert` op for this purpose. Did we also need this specific assert op to be supported, or did you add this op and it's lowering specifically to decompose `sym_constrain_range_for_size`?

zjgarvey: It would be preferable to use the existing op, since it has a canonicalization pattern.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -11455,6 +11455,56 @@ class DecomposeAtenSpecialExpm1Op
...
+    auto max = op.getMax();
+    auto minOp = min.getDefiningOp();
+    auto maxOp = max.getDefiningOp();
+
+    if (!minOp || !maxOp)
+      return op.emitError("Unimplemented: Non constant min/max values");
+
+    int64_t minValue, maxValue;
+
+    if (isa<Torch::ConstantNoneOp>(minOp)) {
```
zjgarvey, line 11477: Same comment as above (you can use `isa<Torch::NoneType>(min.getType()))`)

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
```diff
@@ -11455,6 +11455,56 @@ class DecomposeAtenSpecialExpm1Op
...
+      min = rewriter.create<Torch::ConstantIntOp>(loc, 0);
+    } else {
+      // Check if min value is a constant
+      if (!matchPattern(min, m_TorchConstantInt(&minValue)))
+        return rewriter.notifyMatchFailure(
+            op, "Expected min value to be constant integer");
+    }
+
+    if (!isa<Torch::ConstantNoneOp>(maxOp)) {
+      // Verify that max value is greater than 2
```
zjgarvey, line 11488: I think I have the same question as @mgehre-amd . Is it possible for the max value to be 0? This would just mean the symbolic value is actually 0, correct?

zjgarvey (changes requested):
Thanks for the PR!

I'd like to request a few small changes, and I'd also like to at least have a discussion for context on these ops and their relationship with `torch.symbolic_int` and `torch.bind_symbolic_shape`.

projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -6453,3 +6453,66 @@ def forward(self, x):
...
+class AtenAssertScalar(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([None, ([-1], torch.int, True)])
+    def forward(self, x):
+        a = x.item()
+        # The below checks introduces aten._assert_scalar op
+        torch._check_is_size(a)
```
vivekkhandelwal1, line 6511: Why do we need this check in the test?

projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
```diff
@@ -6453,3 +6453,66 @@ def forward(self, x):
...
+class AtenSymConstrainRangeForSize(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args([None, ([-1], torch.int, True)])
+    def forward(self, x):
+        a = x.item()
+        torch._check_is_size(a)
+        # max should be > 2
```
vivekkhandelwal1, line 6491: @praveen-g-ctt, I don't think this comment is needed here. It should only be part of the lowering.

lib/Dialect/Torch/Transforms/DecomposeComplexOps.cpp
praveen-g-ctt: @zjgarvey @mgehre-amd For the sym_constrain_range_for_size op, there is an assumption that the max value would be greater than 2 when size is used. The generic variant sym_constrain_range does not have this requirement. 

https://pytorch.org/docs/stable/generated/torch.fx.experimental.symbolic_shapes.constrain_range.html
https://github.com/pytorch/pytorch/blob/f7000350905be5073892e0b23df681c0281be0f0/torch/__init__.py#L2682

lib/Conversion/TorchToLinalg/Uncategorized.cpp
praveen-g-ctt: Done

lib/Conversion/TorchToLinalg/Uncategorized.cpp
praveen-g-ctt: @zjgarvey Thanks. The aten::_assert_scalar op gets generated as part of evaluating the sym_constrain_range_for_size op. Lowered the op to the existing torch.runtime.assert

projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
praveen-g-ctt: The _check_is_size generates the _assert_scalar op with an assertion message. I have modified the test case to invoke the aten op directly with a custom assertion message

lib/Conversion/TorchToLinalg/Uncategorized.cpp
praveen-g-ctt: _Are the sym_constrain_range_for_size ops intended to provide symbolic shape constrains similar to symbolic_int and bind_symbolic_shape?_

-> @zjgarvey Yes, the sym_constrain_range_for_size op is intended to specify the symbolic shape constrains for the scalar values. 

The constraints specified in this op is evaluated along with other expressions used in the symbol evaluation such as min/max values for dynamic shaped tensors. If the value is assumed to be within the range of the specified constrains, the **sym_constrain_range_for_size** node gets removed altogether while generating the graph.

```
def test_constraints():
    class Basic(nn.Module):
        def __init__(self):
            super().__init__()

        def forward(self, x, y):
            a = y.shape[0]
            torch.sym_constrain_range_for_size(a, max=11)
            return torch.broadcast_to(x, (a, -1))

    # Sample inputs
    x = torch.randn(1, 4)
    y = torch.randn(8)

    dim_0 = Dim("dim_0", max=10)
    dynamic_shapes = {
        "x": {},
        "y": {0: dim_0},
    }

    m = fx.export_and_import(
        Basic(),
        x,
        y,
        dynamic_shapes=dynamic_shapes,
        func_name="test_constraints",
        import_symbolic_shape_expressions=True,
    )
    print(m)
```


_This comment doesn't necessarily need to be addressed here, but if we can get some context as to why these ops are introduced in examples, it might be helpful for improvements in the near future._

-> Some of the checks such as **torch._check_is_size()** introduces sym_constrain_range_for_size op. I was not able to generate more examples which includes other operations involving symbolic constraints along with this. I guess we can investigate further regarding this with more examples.

zjgarvey: Revisiting this conversation after looking at the IR you shared with me, it looks like these ops were generated because `aten.nonzero` returns a dynamic-shaped result. But this nonzero op isn't really used. Uploaded IR to gist [here](https://gist.github.com/zjgarvey/db71a1d15079898a79820d465a61163c) for others to look at. I also don't understand why the fx graph doesn't remove these checks since the min and max values are `none`.

test/Dialect/Torch/decompose-complex-ops.mlir
```diff
@@ -228,3 +228,56 @@ func.func @torch.aten.fft_rfft$2d_first_dim(%arg0: !torch.vtensor<[36,23],f32>)
...
+// CHECK-LABEL:   func.func @torch.aten.sym_constrain_range_for_size(
+// CHECK-SAME:                                                       %[[VAL_0:.*]]: !torch.vtensor<[],si64>) -> !torch.int {
+// CHECK:           %[[VAL_1:.*]] = torch.constant.int 7
+// CHECK:           %[[VAL_2:.*]] = torch.constant.int 0
+// CHECK:           %[[VAL_3:.*]] = torch.constant.none
+// CHECK:           %[[VAL_4:.*]] = torch.aten.item %[[VAL_0]] : !torch.vtensor<[],si64> -> !torch.int
+// CHECK:           torch.aten.sym_constrain_range %[[VAL_4]], %[[VAL_2]], %[[VAL_3]] : !torch.int, !torch.int, !torch.none
+// CHECK:           torch.aten.sym_constrain_range %[[VAL_4]], %[[VAL_2]], %[[VAL_1]] : !torch.int, !torch.int, !torch.int
+// CHECK:           return %[[VAL_4]] : !torch.int
+func.func @torch.aten.sym_constrain_range_for_size(%arg0: !torch.vtensor<[],si64>) -> !torch.int {
```
zjgarvey, line 243: Although it is a bit of a nit, it might be more representative to have these lit tests use `%arg0 = !torch.vtensor<[?],f32>` and then apply the sym constrain op to the result of a `torch.aten.size.int` op on `%arg0`. 

You could also simplify these tests by just having `%arg0` be a `!torch.int` and then you don't need to test anything here other than the `sym_constrain_range_for_size` op. 

This is pretty minor, but it is a good idea to keep lit tests as simple as possible.

zjgarvey (approved):
Thanks for the work, Praveen. 

After looking at the example IR, I think it is totally fine to merge this as-is. In future patches, it would be good to add some folders (e.g., if min and max are both `%none`'s, just erase it). And if we see other examples where these ops could inform symbolic shape constraints in a useful way for codegen, we can make changes then.

test/Dialect/Torch/decompose-complex-ops.mlir
praveen-g-ctt: Thanks @zjgarvey. Updated the tests

praveen-g-ctt:
> Thanks for the work, Praveen.
> 
> After looking at the example IR, I think it is totally fine to merge this as-is. In future patches, it would be good to add some folders (e.g., if min and max are both `%none`'s, just erase it). And if we see other examples where these ops could inform symbolic shape constraints in a useful way for codegen, we can make changes then.

Sure Zach.  Will add folders in future patches.

vivekkhandelwal1 (approved):
LGTM

Issue #3453, opened 2024-06-12
convert-torch-onnx-to-torch generates invalid IR for onnx.Resize where scaling is in the first two dimensions


mgehre-amd:
The code in that pass seems to silently assume the the first two dimensions are not scaled, but ONNX has no such restriction.

With input
```
func.func @test_resize_middle(%arg0: !torch.vtensor<[1,36,42,384],f32>) -> !torch.vtensor<[1,72,84,384],f32> 
  attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
 %none = torch.constant.none
 %12 = torch.vtensor.literal(dense<[1.000000e+00, 2.000000e+00, 2.000000e+00, 1.000000e+00]> : tensor<4xf32>) : !torch.vtensor<[4],f32>
 %19 = torch.operator "onnx.Resize"(%arg0, %none, %12) {
  torch.onnx.coordinate_transformation_mode = "half_pixel",
  torch.onnx.mode = "nearest",
  torch.onnx.nearest_mode = "round_prefer_floor"} : (!torch.vtensor<[1,36,42,384],f32>, !torch.none, !torch.vtensor<[4],f32>) -> !torch.vtensor<[1,72,84,384],f32>
 return %19 : !torch.vtensor<[1,72,84,384],f32>
}
```
we get
```
module {
  func.func @test_resize_middle(%arg0: !torch.vtensor<[1,36,42,384],f32>) -> !torch.vtensor<[1,72,84,384],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
    %none = torch.constant.none
    %0 = torch.vtensor.literal(dense<[1.000000e+00, 2.000000e+00, 2.000000e+00, 1.000000e+00]> : tensor<4xf32>) : !torch.vtensor<[4],f32>
    %none_0 = torch.constant.none
    %int0 = torch.constant.int 0
    %false = torch.constant.bool false
    %true = torch.constant.bool true
    %str = torch.constant.str "nearest_half_pixel,round_prefer_floor"
    %int2 = torch.constant.int 2
[...]

PR #3945, opened 2025-01-07
fix(ONNX): avoids resizing unsupported dimensions 

bjacobgordon:
Prevents #3453

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2763,10 +2763,6 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
               binder.op, "unimplemented: cubic coeff must be -0.75");
         }
 
-        unsigned rank = dyn_cast<Torch::ValueTensorType>(operands[0].getType())
-                            .getSizes()
-                            .size();
-
         auto loc = binder.getLoc();
 
         Value cstFalse = rewriter.create<Torch::ConstantBoolOp>(loc, false);
@@ -2781,6 +2777,11 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             modeStr = modeStr + "_" + coordTfMode;
           modeStrValue = rewriter.create<Torch::ConstantStrOp>(loc, modeStr);
         }
+
+        unsigned rank = dyn_cast<Torch::ValueTensorType>(operands[0].getType())
+                            .getSizes()
+                            .size();
+
         // supported modes:
         // bilinear (half_pixel), bilinear with align_corners,
         // bilinear_pytorch_half_pixel, bilinear_asymmetric nearest
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2779,9 +2779,9 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         }
 
         Value inputTensor = operands[0];
-        unsigned rank = cast<Torch::BaseTensorType>(inputTensor.getType())
-                            .getSizes()
-                            .size();
+        auto inputTensorType =
+            cast<Torch::BaseTensorType>(inputTensor.getType());
+        unsigned rank = inputTensorType.getSizes().size();
 
         // supported modes:
         // bilinear (half_pixel), bilinear with align_corners,
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2778,7 +2778,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           modeStrValue = rewriter.create<Torch::ConstantStrOp>(loc, modeStr);
         }
 
-        unsigned rank = dyn_cast<Torch::ValueTensorType>(operands[0].getType())
+        unsigned rank = cast<Torch::ValueTensorType>(operands[0].getType())
                             .getSizes()
                             .size();
 
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2825,19 +2825,19 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
 
         int64_t assumedForemostSpatialDim = 2;
 
-        Value scalesValueList;
+        Value supportedScaleFactors;
         Value supportedSizes;
 
         Value noneVal = rewriter.create<Torch::ConstantNoneOp>(loc);
 
         if (numberOfOperands == 3) {
           Value scaleOperand = operands[2];
-          scalesValueList = createScalarSublist(
+          supportedScaleFactors = createScalarSublist(
               loc, scaleOperand, assumedForemostSpatialDim, rewriter);
           supportedSizes = noneVal;
         } else if (numberOfOperands == 4) {
           Value sizeOperand = operands[3];
-          scalesValueList = noneVal;
+          supportedScaleFactors = noneVal;
           supportedSizes = createScalarSublist(
               loc, sizeOperand, assumedForemostSpatialDim, rewriter);
         } else
@@ -2846,7 +2846,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         rewriter
             .replaceOpWithNewOp<Torch::Aten__InterpolateSizeListScaleListOp>(
                 binder.op, outputTensorType, inputTensor, supportedSizes,
-                scalesValueList, modeStrValue,
+                supportedScaleFactors, modeStrValue,
                 /* AnyTorchOptionalBoolType:$align_corners */ alignCorners,
                 /* AnyTorchOptionalBoolType:$recompute_scale_factor */ noneVal,
                 /*Torch_BoolType:$antialias*/ cstFalse);
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2782,7 +2782,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         auto inputTensorType =
             cast<Torch::BaseTensorType>(inputTensor.getType());
         auto sizesOfInputTensor = inputTensorType.getSizes();
-        unsigned rank = sizesOfInputTensor.size();
+        auto rank = sizesOfInputTensor.size();
 
         // supported modes:
         // bilinear (half_pixel), bilinear with align_corners,
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2831,9 +2831,9 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         Value noneVal = rewriter.create<Torch::ConstantNoneOp>(loc);
 
         if (numberOfOperands == 3) {
-          Value scaleOperand = operands[2];
+          Value proposedScaleFactors = operands[2];
           supportedScaleFactors = createScalarSublist(
-              loc, scaleOperand, assumedForemostSpatialDim, rewriter);
+              loc, proposedScaleFactors, assumedForemostSpatialDim, rewriter);
           supportedSizes = noneVal;
         } else if (numberOfOperands == 4) {
           Value sizeOperand = operands[3];
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2731,6 +2731,42 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                                           "round_prefer_floor") ||
             binder.f32FloatAttr(cubic_coeff_a, "cubic_coeff_a", -0.75))
           return failure();
+
+        int64_t const /*  */ batchDim = 0;
+        int64_t const /**/ channelDim = 1;
+
+        SmallVector<int64_t> nonResizableDims{
+            batchDim,
+            channelDim,
+        };
+
+        Value inputTensor = operands[0];
+        auto inputTensorType =
+            cast<Torch::BaseTensorType>(inputTensor.getType());
+        auto sizesOfInputTensor = inputTensorType.getSizes();
+        auto sizesOfOutputTensor = outputTensorType.getSizes();
+
+        auto unknownSize = Torch::kUnknownSize;
+
+        // Compile-time check for dimensions of static size
+        for (auto &eachDim : nonResizableDims) {
+          auto eachSizeOfInputTensor = sizesOfInputTensor[eachDim];
+          auto eachSizeOfOutputTensor = sizesOfOutputTensor[eachDim];
+
+          if (eachSizeOfInputTensor == unknownSize ||
+              eachSizeOfOutputTensor == unknownSize)
+            continue;
+          if (eachSizeOfInputTensor == eachSizeOfOutputTensor)
+            continue;
+
+          auto resizingIntentErrorMessage =
+              "unsupported: non-trivial intent to resize dimension: " +
+              std::to_string(eachDim);
+
+          return rewriter.notifyMatchFailure(binder.op,
+                                             resizingIntentErrorMessage);
+        };
+
         if (antialias != 0) {
           return rewriter.notifyMatchFailure(
               binder.op,
@@ -2778,10 +2814,6 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           modeStrValue = rewriter.create<Torch::ConstantStrOp>(loc, modeStr);
         }
 
-        Value inputTensor = operands[0];
-        auto inputTensorType =
-            cast<Torch::BaseTensorType>(inputTensor.getType());
-        auto sizesOfInputTensor = inputTensorType.getSizes();
         auto rankOfInputTensor = sizesOfInputTensor.size();
 
         // supported modes:
@@ -2823,7 +2855,9 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
 
         auto numberOfOperands = operands.size();
 
-        int64_t assumedForemostSpatialDim = 2;
+        Type boolType = rewriter.getType<Torch::BoolType>();
+
+        int64_t assumedForemostSpatialDim = 1 + nonResizableDims.back();
 
         Value supportedScaleFactors;
         Value supportedSizes;
@@ -2832,11 +2866,58 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
 
         if (numberOfOperands == 3) {
           Value proposedScaleFactors = operands[2];
+
+          Value scaleIdentity = rewriter.create<Torch::ConstantFloatOp>(
+              loc, rewriter.getF64FloatAttr(1.0));
+
+          // run-time scale factor check for dynamic sizes
+          for (auto &eachDim : nonResizableDims) {
+            Value eachProposedScaleFactor = extractTorchScalar(
+                loc, eachDim, proposedScaleFactors, rewriter);
+
+            Value eachScaleFactorIsIdentity =
+                rewriter.create<Torch::AtenEqFloatOp>(
+                    loc, boolType, eachProposedScaleFactor, scaleIdentity);
+
+            auto errorMessageForEachDim =
+                "Unsupported: non-trivial scale factor for dimension " +
+                std::to_string(eachDim);
+
+            rewriter.create<Torch::RuntimeAssertOp>(
+                loc, eachScaleFactorIsIdentity,
+                rewriter.getStringAttr(errorMessageForEachDim));
+          };
+
           supportedScaleFactors = createScalarSublist(
               loc, proposedScaleFactors, assumedForemostSpatialDim, rewriter);
           supportedSizes = noneVal;
         } else if (numberOfOperands == 4) {
           Value proposedSizes = operands[3];
+
+          // run-time target size check for dynamic sizes
+          for (auto &eachDimAsInt : nonResizableDims) {
+            Value eachDimAsValue =
+                rewriter.create<Torch::ConstantIntOp>(loc, eachDimAsInt);
+
+            Value eachSizeOfInputTensor = rewriter.create<Torch::AtenSizeIntOp>(
+                loc, inputTensor, eachDimAsValue);
+
+            Value eachProposedSize =
+                extractTorchScalar(loc, eachDimAsInt, proposedSizes, rewriter);
+
+            Value eachProposedSizeIsTrivial =
+                rewriter.create<Torch::AtenEqIntOp>(
+                    loc, boolType, eachProposedSize, eachSizeOfInputTensor);
+
+            auto errorMessageForEachDim =
+                "Unsupported: non-trivial resizing of dimension " +
+                std::to_string(eachDimAsInt);
+
+            rewriter.create<Torch::RuntimeAssertOp>(
+                loc, eachProposedSizeIsTrivial,
+                rewriter.getStringAttr(errorMessageForEachDim));
+          };
+
           supportedScaleFactors = noneVal;
           supportedSizes = createScalarSublist(
               loc, proposedSizes, assumedForemostSpatialDim, rewriter);
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2826,8 +2826,8 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
 
         int64_t assumedForemostSpatialDim = 2;
 
-        Value scalesValueList = noneVal;
-        Value sizesValueList = noneVal;
+        Value scalesValueList;
+        Value sizesValueList;
 
         if (numberOfOperands == 3) {
           Value scaleOperand = operands[2];
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2826,7 +2826,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
 
         int64_t assumedForemostSpatialDim = 2;
 
-        if (operands.size() < 4) {
+        if (operands.size() == 3) {
           Value scaleOperand = operands[2];
           scalesValueList =
               createScalarSublist(binder.getLoc(), scaleOperand,
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2836,10 +2836,10 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
               loc, proposedScaleFactors, assumedForemostSpatialDim, rewriter);
           supportedSizes = noneVal;
         } else if (numberOfOperands == 4) {
-          Value sizeOperand = operands[3];
+          Value proposedSizes = operands[3];
           supportedScaleFactors = noneVal;
           supportedSizes = createScalarSublist(
-              loc, sizeOperand, assumedForemostSpatialDim, rewriter);
+              loc, proposedSizes, assumedForemostSpatialDim, rewriter);
         } else
           return rewriter.notifyMatchFailure(binder.op, "unknown scaling mode");
 
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2778,7 +2778,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           modeStrValue = rewriter.create<Torch::ConstantStrOp>(loc, modeStr);
         }
 
-        unsigned rank = cast<Torch::ValueTensorType>(operands[0].getType())
+        unsigned rank = cast<Torch::BaseTensorType>(operands[0].getType())
                             .getSizes()
                             .size();
 
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2824,15 +2824,17 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
               rewriter.create<Torch::ConstantStrOp>(binder.getLoc(), modeStr);
         }
 
+        auto numberOfOperands = operands.size();
+
         int64_t assumedForemostSpatialDim = 2;
 
-        if (operands.size() == 3) {
+        if (numberOfOperands == 3) {
           Value scaleOperand = operands[2];
           scalesValueList =
               createScalarSublist(binder.getLoc(), scaleOperand,
                                   assumedForemostSpatialDim, rewriter);
           sizesValueList = noneVal;
-        } else if (operands.size() == 4) {
+        } else if (numberOfOperands == 4) {
           Value sizeOperand = operands[3];
           scalesValueList = noneVal;
           sizesValueList =
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2705,7 +2705,6 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         std::string mode, nearest_mode, coordTfMode;
         int64_t antialias, exclude_outside;
         float extrapolation_value, cubic_coeff_a;
-        Value noneVal = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
 
         if (auto attr = binder.op->getAttr("torch.onnx.axes")) {
           return rewriter.notifyMatchFailure(
@@ -2828,6 +2827,8 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
 
         Value scalesValueList;
         Value sizesValueList;
+
+        Value noneVal = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
 
         if (numberOfOperands == 3) {
           Value scaleOperand = operands[2];
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2832,17 +2832,15 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
               createScalarSublist(binder.getLoc(), scaleOperand,
                                   assumedForemostSpatialDim, rewriter);
           sizesValueList = noneVal;
-        } else {
+        } else if (operands.size() == 4) {
           Value sizeOperand = operands[3];
           scalesValueList = noneVal;
           sizesValueList =
               createScalarSublist(binder.getLoc(), sizeOperand,
                                   assumedForemostSpatialDim, rewriter);
-        }
-        if (isa<Torch::NoneType>(scalesValueList.getType()) &&
-            isa<Torch::NoneType>(sizesValueList.getType())) {
+        } else
           return rewriter.notifyMatchFailure(binder.op, "unknown scaling mode");
-        }
+
         rewriter
             .replaceOpWithNewOp<Torch::Aten__InterpolateSizeListScaleListOp>(
                 binder.op, resultType, operands[0], sizesValueList,
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2767,10 +2767,10 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
                             .getSizes()
                             .size();
 
-        Value cstFalse =
-            rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), false);
-        Value cstTrue =
-            rewriter.create<Torch::ConstantBoolOp>(binder.getLoc(), true);
+        auto loc = binder.getLoc();
+
+        Value cstFalse = rewriter.create<Torch::ConstantBoolOp>(loc, false);
+        Value cstTrue = rewriter.create<Torch::ConstantBoolOp>(loc, true);
         Value modeStrValue;
 
         Value alignCorners =
@@ -2779,8 +2779,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           std::string modeStr = "cubic";
           if (coordTfMode != "half_pixel")
             modeStr = modeStr + "_" + coordTfMode;
-          modeStrValue =
-              rewriter.create<Torch::ConstantStrOp>(binder.getLoc(), modeStr);
+          modeStrValue = rewriter.create<Torch::ConstantStrOp>(loc, modeStr);
         }
         // supported modes:
         // bilinear (half_pixel), bilinear with align_corners,
@@ -2806,8 +2805,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           // mode is apparently half_pixel, NOT pytorch_half_pixel
           if (coordTfMode != "half_pixel" && coordTfMode != "align_corners")
             modeStr = (modeStr + "_") + coordTfMode;
-          modeStrValue =
-              rewriter.create<Torch::ConstantStrOp>(binder.getLoc(), modeStr);
+          modeStrValue = rewriter.create<Torch::ConstantStrOp>(loc, modeStr);
         }
         if (mode == "nearest") {
           std::string modeStr = "nearest";
@@ -2817,8 +2815,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
             modeStr = (modeStr + "_") + coordTfMode;
           if (nearest_mode != "floor" && nearest_mode != "")
             modeStr = modeStr + "," + nearest_mode;
-          modeStrValue =
-              rewriter.create<Torch::ConstantStrOp>(binder.getLoc(), modeStr);
+          modeStrValue = rewriter.create<Torch::ConstantStrOp>(loc, modeStr);
         }
 
         auto numberOfOperands = operands.size();
@@ -2828,20 +2825,18 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         Value scalesValueList;
         Value sizesValueList;
 
-        Value noneVal = rewriter.create<Torch::ConstantNoneOp>(binder.getLoc());
+        Value noneVal = rewriter.create<Torch::ConstantNoneOp>(loc);
 
         if (numberOfOperands == 3) {
           Value scaleOperand = operands[2];
-          scalesValueList =
-              createScalarSublist(binder.getLoc(), scaleOperand,
-                                  assumedForemostSpatialDim, rewriter);
+          scalesValueList = createScalarSublist(
+              loc, scaleOperand, assumedForemostSpatialDim, rewriter);
           sizesValueList = noneVal;
         } else if (numberOfOperands == 4) {
           Value sizeOperand = operands[3];
           scalesValueList = noneVal;
-          sizesValueList =
-              createScalarSublist(binder.getLoc(), sizeOperand,
-                                  assumedForemostSpatialDim, rewriter);
+          sizesValueList = createScalarSublist(
+              loc, sizeOperand, assumedForemostSpatialDim, rewriter);
         } else
           return rewriter.notifyMatchFailure(binder.op, "unknown scaling mode");
 
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2778,7 +2778,8 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
           modeStrValue = rewriter.create<Torch::ConstantStrOp>(loc, modeStr);
         }
 
-        unsigned rank = cast<Torch::BaseTensorType>(operands[0].getType())
+        Value inputTensor = operands[0];
+        unsigned rank = cast<Torch::BaseTensorType>(inputTensor.getType())
                             .getSizes()
                             .size();
 
@@ -2843,7 +2844,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
 
         rewriter
             .replaceOpWithNewOp<Torch::Aten__InterpolateSizeListScaleListOp>(
-                binder.op, resultType, operands[0], sizesValueList,
+                binder.op, resultType, inputTensor, sizesValueList,
                 scalesValueList, modeStrValue,
                 /* AnyTorchOptionalBoolType:$align_corners */ alignCorners,
                 /* AnyTorchOptionalBoolType:$recompute_scale_factor */ noneVal,
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
@@ -2781,7 +2781,8 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         Value inputTensor = operands[0];
         auto inputTensorType =
             cast<Torch::BaseTensorType>(inputTensor.getType());
-        unsigned rank = inputTensorType.getSizes().size();
+        auto sizesOfInputTensor = inputTensorType.getSizes();
+        unsigned rank = sizesOfInputTensor.size();
 
         // supported modes:
         // bilinear (half_pixel), bilinear with align_corners,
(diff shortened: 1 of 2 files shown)


lib/Dialect/Torch/IR/TorchTypes.cpp
```diff
@@ -217,6 +217,31 @@ static bool isValidTorchDtype(Type dtype) {
   return false;
 }
 
+ArrayRef<std::optional<bool>>
+BaseTensorType::dimensionComparisonsAgainst(BaseTensorType that) const {
+  auto this_dimensions = /**/ getSizes();
+  auto that_dimensions = that.getSizes();
```
zjgarvey, line 223: `BaseTensorType` might not have sizes, and this will cause a crash when called. I would do:

```suggestion
  auto selfSizes = getOptionalSizes();
  auto otherSizes = other.getOptionalSizes();
```
Note also the variable and camel casing conventions. The variables `self` and `other` are used more typically than `this` and `that` in this codebase (easier to distiguish).

include/torch-mlir/Dialect/Torch/IR/TorchTypes.h
```diff
@@ -84,6 +84,10 @@ class BaseTensorType : public Type {
   /// Enable isa/dyn_cast for BaseTensorType.
   static bool classof(Type type);
 
+  /// The element-wise comparison of each dimension/size in `that` tensor
+  std::vector<std::optional<bool>>
```
zjgarvey, line 88: Use `SmallVector` instead of `std::vector`. The methods are the same, and it is better for small containers like this.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2686,12 +2686,11 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
       });
   patterns.onOp(
       "Resize", 11, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-        Torch::ValueTensorType resultType;
+        Torch::ValueTensorType outputTensor_blueprint;
```
zjgarvey, line 2689: I don't understand the name changing of this variable. This isn't a blueprint, it's the result type.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2718,6 +2717,32 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        Value inputTensor = operands[0];
+        Torch::ValueTensorType inputTensor_blueprint =
+            cast<Torch::ValueTensorType>(inputTensor.getType());
+
+        std::vector<std::optional<bool>> shapeComparison =
+            inputTensor_blueprint.shapeComparisonAgainst(
+                outputTensor_blueprint);
+
+        // Comparisons of the dimensions assumed to carry the batch and channel
+        auto shapeComparisonForFixedDimensions =
```
zjgarvey, line 2730: nit: change `Fixed` -> `Static`

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2718,6 +2717,32 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            ArrayRef(shapeComparison).take_front(2);
+
+        for (auto eachDimensionComparison : shapeComparisonForFixedDimensions) {
+          if (eachDimensionComparison == std::nullopt) {
+            return rewriter.notifyMatchFailure(
+                binder.op, "Sizes for batch and channel dimensions must be "
+                           "statically defined");
+          }
```
zjgarvey, line 2738: We definitely do not want to constrain this conversion to static batch and channel dims. This was the reason for needing to write asserts into the helper function `getValueList` in the dynamic case.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2718,6 +2717,32 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        std::vector<std::optional<bool>> shapeComparison =
+            inputTensor_blueprint.shapeComparisonAgainst(
+                outputTensor_blueprint);
+
+        // Comparisons of the dimensions assumed to carry the batch and channel
+        auto shapeComparisonForFixedDimensions =
+            ArrayRef(shapeComparison).take_front(2);
+
+        for (auto eachDimensionComparison : shapeComparisonForFixedDimensions) {
+          if (eachDimensionComparison == std::nullopt) {
```
zjgarvey, line 2734: Since you need to loop over the result of the shape comparison anyway, it would be more efficient to not define the helper function at all, and do 

```cpp
for (int64_t dim = 0; dim < 2; dim++) {
    if (inputSizes[dim] == Torch::kUnknownSize || outputSizes[dim] == Torch::kUnknownSize)
        continue; // you need to implement the runtime asserts, but at least still check the other dim if static.
    if (inputSizes[dim] != outputSizes[dim])
        return rewriter.notifyMatchFailure(...
```

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
zjgarvey: It might be fine to just put runtime asserts in right after the last match failure. Something like:

```cpp
Value inputDimZero = rewriter.create<Torch::AtenSizeIntOp>(loc, input, cstZero);
Value inputDimOne = rewriter.create<Torch::AtenSizeIntOp>(loc, input, cstOne);
Value outputDimZero = rewriter.create<Torch::AtenSizeIntOp>(loc, output, cstZero);
Value outputDimOne = rewriter.create<Torch::AtenSizeIntOp>(loc, output, cstOne);
Value cmpDimZero = rewriter.create<Torch::AtenEqIntOp>(loc, inputDimZero, outputDimZero);
Value cmpDimOne = ...
rewriter.create<Torch::RuntimeAssertOp>(loc, cmpDimZero, rewriter.getStringAttr("message"));
// same for DimOne
```

zjgarvey: By the way, if one of the two dims have input/output sizes that are static and equal, then these asserts will fold out, so there isn't a pressing need to check again for static dims.

zjgarvey (commented):
I think the main structural question is about the need for adding the `BaseTensorType` method. If it were useful elsewhere (I have some doubts, since we would need to know too much about the two tensor shapes prior to using it- namely that they are present, and they have the same rank), I would consider keeping it; however, the code is simplified here by not using it, and I suspect that the same would be true in other circumstances where it might be used.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: Thanks for pointing this out! There's actually an opportunity for even more clarity here.

Thinking out loud:
- "Static"/"Dynamic": whether the input/output shapes are anchored to specific values
- "Fixed"/"Changing": just now snapping to the fact that these are literally synonymous to the words above haha. Funny thing is that there's a more accurate word that'll convey my meaning
- "Scalable"/"Nonscalable": whether an input/output dimension is even eligible for scaling, regardless of whether they were defined at compile time.

A wording that might more accurately describe what I was thinking: `shapeComparisonForNonscalableDimensions`

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
zjgarvey: I was actually suggesting this for both readability _and_ (very modestly) compiler performance. The runtime performance won't be affected either way. 

In general, it is good to have runtime performance in mind at this level, but know that many things do indeed get optimized out later on (see, for example, my comment about the folding of runtime asserts for the static dim case). I tend to take a pessimistic view of what will and won't be optimized away, at least when I don't actually know- and will try to generate a cleaner pattern if it doesn't cost a huge amount in code complexity for the perceived benefit.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: Thanks for mentioning this!

My current mental model (correct me if I've misunderstood anything):
- `BaseTensorType`: assigned by value rather than by reference/pointer
  - an instance
     - can describe the shape of some tensor a.k.a. the size/number of elements along each dimension
     - can describe how the numerical elements within the tensor are embodied in memory, i.e. number of bits, integers vs floating points, signed vs unsigned etc.
     - does not actually contain the elements themselves
  - Subclasses: 
    - `ValueTensorType`: a `BaseTensorType` with `ValueTensorTypeStorage`
      - storage is passed by value ("value semantics"), so as an SSA, it's immutable and guaranteed to have no aliases
    - `NonValueTensorType`: a `BaseTensorType` with `NonValueTensorTypeStorage`
      - storage is not passed by value, a.k.a is passed by reference ("reference semantics"), so even as an SSA, it's mutable and "aliasable"

Challenges with current naming conventions (that you probably already know):
- "type": the most generic term there is for the nature of some instance
  - _so_ generic that it's hard-pressed to mean anything without context for surroundings or implementation details.
  - "types" are baked into the fabric of our understanding of information in general.
  - And every instance essentially must have a type to exist in memory, so saying "type" doesn't really add any information
- "base": Another generic term typically used to denote the beginning of some portion of an inheritance chain.
  - example:
     - ~`BaseTruck'~ `Truck`
        - `SemiTruck`
        - `PickupTruck`
- "result": TODO

Iterating on the name:
1. `resultType`: TODO

zjgarvey: Whatever matches the corresponding `Value` name is preferred. If we are using `outputTensor` for the output value variable name, then `outputTensorType` would be an appropriate name for the `Type` of `outputTensor`. 

Populating the code with analogy words like "summary" or "blueprint" would significantly increase the cognitive load, since these warrant broader interpretation. `Type` means `Type` (or a subclass of `Type`). Once a developer understands what `Type` is, the job is done. 

Maybe `<thing name><type subclass name>` would be an option, but it should even be clear what subclass of `Type` some particular thing is, either from casts or from the variable declaration. E.g., the line we are commenting on. Therefore, anything more than `<thing name>Type` is usually either superfluous or confusing.

bjacobgordon: Got it, will make the change!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: I revisited implementing raw loops for this, and I see what you mean. Should be way more intuitive now. The abstraction I had was definitely too much too soon, was just one of those cases where I had to do it anyway to see where the line was.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: Cool, thanks for the example! Will definitely build on that. I had misunderstood the context for the dynamic vs. static batch/channel dims, but it's clicked now.

include/torch-mlir/Dialect/Torch/IR/TorchTypes.h
bjacobgordon: I ended up removing this implementation entirely after what you recommended. 👍

lib/Dialect/Torch/IR/TorchTypes.cpp
bjacobgordon: Roger that! I removed this implementation after your other suggestion, but will keep that in mind for the future. 👍

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2706,7 +2705,7 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         }
 
         if (binder.tensorOperandsList(operands) ||
-            binder.tensorResultType(resultType) ||
+            binder.tensorResultType(outputTensor_type) ||
```
zjgarvey, line 2708: Nit: every cpp variable *should* be in camel case. It seems like some authors have chosen to name variables like `nearest_mode` to exactly match the ONNX attribute, but even this is probably not preferred. 

In cases like this where there isn't a justifiable reason to use a different convention, you should definitely stick to camel case.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2750,16 +2749,76 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
               binder.op, "unimplemented: cubic coeff must be -0.75");
         }
 
-        unsigned rank = dyn_cast<Torch::ValueTensorType>(operands[0].getType())
-                            .getSizes()
-                            .size();
+        Value inputTensor = operands[0];
+        Torch::ValueTensorType inputTensor_type =
+            cast<Torch::ValueTensorType>(inputTensor.getType());
```
zjgarvey, line 2754: Nit: naming should be camel case and I would use `auto` here, since the cast is explicit about what the type should be.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2750,16 +2749,76 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        ArrayRef<int64_t> inputTensor_sizes = inputTensor_type.getSizes();
+        ArrayRef<int64_t> outputTensor_sizes = outputTensor_type.getSizes();
```
zjgarvey, line 2756: Nit: camel case. I would also recommend `inputSizes` and `outputSizes` for brevity, since there is no ambiguity here between `inputTensorSizes` and `inputSizes`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2750,16 +2749,76 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            return "Unexpected intent to scale the channel dimension";
+          default:
+            return "Scalable dimension treated as non-scalable";
+          }
+        };
+
+        auto unknownSize = Torch::kUnknownSize;
+
+        // Compile-time check for dimensions of static size
+        for (auto eachDimension : nonScalableDimensions) {
```
zjgarvey, line 2779: This is a personal preference, but I would certainly prefer:
```suggestion
        // Compile-time check for dimensions of static size
        for (int64_t i = 0; i<2; i++) {
```

This is for two reasons: 

1. Although the torch op might have these specific dimension specifications, the ONNX op does not. We should say something a bit more generic like "unsupported: non-trivial scaling in the first two dimensions.". 
2. We might want to implement support for this path instead of writing a match failure in the future. With all of the variables and structures introduced here, I think this would cause more work for the future developer.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2750,16 +2749,76 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          auto eachInputSize = inputTensor_sizes[eachDimension];
+          auto eachOutputSize = outputTensor_sizes[eachDimension];
+
+          if (eachInputSize == unknownSize || eachOutputSize == unknownSize) {
+            continue;
+          } else if (eachInputSize == eachOutputSize) {
```
zjgarvey, line 2785: `else` is uneccessary because of the continue. Also check `!=` here so you don't need the additional `continue`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2750,16 +2749,76 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          Value eachOutputSizeAsValue = rewriter.create<Torch::ConstantIntOp>(
+              binderLocation, rewriter.getI64IntegerAttr(eachOutputSize));
+
+          Value eachSizeComparison = rewriter.create<Torch::AtenEqIntOp>(
+              binderLocation, eachInputSizeAsValue, eachOutputSizeAsValue);
+
+          rewriter.create<Torch::RuntimeAssertOp>(
+              binderLocation, eachSizeComparison,
+              rewriter.getStringAttr(errorMessageForScaling(eachDimension)));
+        };
```
zjgarvey, line 2813: Okay, this is were the problem with the CI likely is, and this is my fault for the poor suggestion earlier. 

The value `eachOutputSize` is `-1` (or whatever `Torch::kUnknownSize` is). This will always cause the assert to fail, since `eachInputSizeAsValue` will never be `-1`. 

We actually can't check the output size dynamically, since we aren't computing the original, so we really do need to check the provided size or scale values for non-default values instead, which might be a bit more complicated. Sorry for the misdirect.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2750,16 +2749,76 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            continue;
+          } else if (eachInputSize == eachOutputSize) {
+            continue;
+          }
+
+          return rewriter.notifyMatchFailure(
+              binder.op, errorMessageForScaling(eachDimension));
+        }
+
+        auto binderLocation = binder.getLoc();
```
zjgarvey, line 2793: please use `loc`. This is actually the original op's location (and not the binder), and `loc` is a fairly canonical variable name for the location of the op being rewritten.

zjgarvey (changes requested):
Sorry for the misdirect earlier, we need to perform the runtime asserts on the scales or sizes values instead of sizes, since we will not have access to the correct output sizes ahead of time.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: Right!
- `else` is procedurally unnecessary because, as is, `continue` will ensure the same procedural outcome.
- and checking size _in_equality to decide on the match failure is also procedurally equivalent.

Other considerations:
- Does the procedural structure immediately communicate which parts should absolutely _not_ be changed?
- Is the "purpose" of the code block as enunciated as possible?

Say we start with this:
```cpp
for (auto eachDimension : nonScalableDimensions) {
  auto eachInputSize = inputTensor_sizes[eachDimension];
  auto eachOutputSize = outputTensor_sizes[eachDimension];

  if (eachInputSize != unknownSize && eachOutputSize != unknownSize &&
      eachInputSize != eachOutputSize) {
        binder.op, errorMessageForScaling(eachDimension));
  }
}
```

And, after looking at this for a moment, we see that the "purpose" is the match failure guarded by a big ol' condition (that likes to play footsy with the linter). So, we extract the condition:
```cpp
for (auto eachDimension : nonScalableDimensions) {
  auto eachInputSize = inputTensor_sizes[eachDimension];
  auto eachOutputSize = outputTensor_sizes[eachDimension];

  auto eachPairOfKnownSizesAreInequal = eachInputSize != unknownSize &&
                                        eachOutputSize != unknownSize &&
                                        eachInputSize != eachOutputSize;

  if (eachPairOfKnownSizesAreInequal) {
        binder.op, errorMessageForScaling(eachDimension));
  }
}
```
and add a guard that makes the if statement a static condition:
```cpp
for (auto eachDimension : nonScalableDimensions) {
  auto eachInputSize = inputTensor_sizes[eachDimension];
  auto eachOutputSize = outputTensor_sizes[eachDimension];

  auto eachPairOfKnownSizesAreInequal = eachInputSize != unknownSize &&
                                        eachOutputSize != unknownSize &&
                                        eachInputSize != eachOutputSize;

  if (!eachPairOfKnownSizesAreInequal) {
    continue;
  }

      binder.op, errorMessageForScaling(eachDimension));
}
```

So now, the "purpose" of the for loop is no longer nested. Pyramid of doom avoided! Now to handle the double negatives in the condition:
```cpp
for (auto eachDimension : nonScalableDimensions) {
  auto eachInputSize = inputTensor_sizes[eachDimension];
  auto eachOutputSize = outputTensor_sizes[eachDimension];

  auto matchFailureIsUnnecessary =
      !(eachInputSize != unknownSize && eachOutputSize != unknownSize &&
        eachInputSize != eachOutputSize);

  if (matchFailureIsUnnecessary) {
    continue;
  }

      binder.op, errorMessageForScaling(eachDimension));
}
```
And distribute the "not":
```cpp
for (auto eachDimension : nonScalableDimensions) {
  auto eachInputSize = inputTensor_sizes[eachDimension];
  auto eachOutputSize = outputTensor_sizes[eachDimension];

  auto matchFailureIsUnnecessary =
      (eachInputSize == unknownSize || eachOutputSize == unknownSize ||
        eachInputSize == eachOutputSize);

  if (matchFailureIsUnnecessary) {
    continue;
  }

      binder.op, errorMessageForScaling(eachDimension));
}
```
extract the check for unknown sizes:
```cpp
for (auto eachDimension : nonScalableDimensions) {
  auto eachInputSize = inputTensor_sizes[eachDimension];
  auto eachOutputSize = outputTensor_sizes[eachDimension];

  auto eachPairHasUnknownSize =
      eachInputSize == unknownSize || eachOutputSize == unknownSize;
  auto matchFailureIsUnnecessary =
      eachPairHasUnknownSize || eachInputSize == eachOutputSize;

  if (matchFailureIsUnnecessary) {
    continue;
  }

      binder.op, errorMessageForScaling(eachDimension));
};
```
[WIP]

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: And personal preferences are worth sharing! They're heuristics for our engineering experiences, and they're worth putting out there because:
- they could help someone solve a problem that they've encountered in their own experience
- it's an opportunity to evaluate a preference, put them to words, recalibrate them, or maybe even drop them altogether when presented with new information.

1. Gotcha, so we want the error message to not reveal the rational. In that case, I'll have the message report the dimension index instead. Also gives us a chance to differentiate the messages between compile-time and run-time check.
2. I agree that it's work! In the case of adding support, work is unavoidable:
   - the work of deleting 3 variable declarations that contain cognitive load
     OR
   - the work of gathering the required cognitive load by
     - talking to other engineers
     - researching within the codebase
     - reading external docs
     - etc.
    OR
   - some other unforeseen work!

Predicted questions by future devs (especially if they're also cross-discipline):
- "Wait, why 2 loops? What's significant about the first two dimensions?"
  - *pings engineer*
- "Okay, so they shouldn't be scaled, but why? What do they represent?"
  - *pings engineer again*
- "Ah, gotcha, they're the batch and channels dims, the elements in which have nothing to do spatial dimensions"

With the dims explicitly declared and the implementation based on them, another engineer might be able to avoid having to recollect the CL we had already collected once before.

IanWood1:
I think the renaming of `loc` -> `opLocation` should probably be split into a separate PR so it can be reviewed separately, it adds ~600 lines of changes that aren't directly related to the functional changes. It would help to keep this PR focused and easier to review.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: Done and done

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -180,53 +180,170 @@ LogicalResult reduceOpImpl(OpBinder binder, ConversionPatternRewriter &rewriter,
...
+        opLocation, typeOfEveryScaleFactor, indexPathOfEachScaleFactor);
+
+    Value eachScaleFactorComparisonAsOp = rewriter.create<Torch::AtenEqFloatOp>(
+        opLocation, rewriter.getType<Torch::BoolType>(), eachScaleFactor,
+        scalingIdentityAsOp);
+
+    bool eachScaleFactorIsIdentity;
+
+    matchPattern(eachScaleFactorComparisonAsOp,
+                 Torch::m_TorchConstantBool(&eachScaleFactorIsIdentity));
```
zjgarvey, line 227: This will never succeed, since the `matchPattern` will not attempt to fold the chain of operations leading to the `AtenEqFloatOp` result. 

You could try `rewriter.createOrFold`, but this will still not work if the sizes/scales vector is still an `onnx.Constant` instead of a `ValueTensorLiteralOp`, or if the sizes/scales are generated from existing dynamic shapes. 

I think you should instead generate runtime asserts on the front two dimensions.

bjacobgordon: Ah okay, gotcha. I'll work on generating the runtime asserts!

bjacobgordon:
Okay, @zjgarvey, I think we're in business! Got green on the CI a few hours ago. Just wrapped up self-review.

I'm guessing now we'll:
- discuss how the diffs might need to change
- finalize those changes
- discuss which chunks of commits need to be in the PR stack
Something like that?

I kept the commits atomic and ordered such that it's easier to propagate changes to the head of the branch in case:
- we tweak a commit
- we rip out a commit
- we insert an early commit

Let me know what you think!

zjgarvey:
> Okay, @zjgarvey, I think we're in business! Got green on the CI a few hours ago. Just wrapped up self-review.
> 
> I'm guessing now we'll:
> 
>     * discuss how the diffs might need to change
> 
>     * finalize those changes
> 
>     * discuss which chunks of commits need to be in the PR stack
>       Something like that?
> 
> 
> I kept the commits atomic and ordered such that it's easier to propagate changes to the head of the branch in case an earlier commit needs to be inserted/tweaked/excised.
> 
> Let me know what you think!

Nice! 

At least for now, please exclude any commits which involve style changes not directly related to the fix content. E.g. widespread enforcement of naming preferences like the `loc -> opLocation` should be factored out into a separate PR as @IanWood1 suggested.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: - I opened #3975 so we can continue conversation there!
- This PR is stocked on top of that one, so it'll just abide by whatever outcome 👍

bjacobgordon:
> I think the renaming of `loc` -> `opLocation` should probably be split into a different PR

Done, now in #3975

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2718,6 +2790,43 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        Value inputTensor = operands[0];
+        auto typeOfInputTensor =
+            cast<Torch::BaseTensorType>(inputTensor.getType());
+
+        auto sizesOfInputTensor = typeOfInputTensor.getSizes();
+        ArrayRef<int64_t> sizesOfOutputTensor = typeOfOutputTensor.getSizes();
+
+        int64_t const dimensionAssumedToBeBatch = 0;
+        int64_t const dimensionAssumedToBeChannel = 1;
+        int64_t nonResizableDimensions[] = {
```
zjgarvey, line 2803: possibly important nit:

There might be a better way to leverage the llvm array-like structures. E.g.

```cpp
SmallVector<int64_t, 2> nonResizableDims{dimensionAssumedToBeBatch, dimensionAssumedToBeChannel};
```

This pre-allocates memory for two `int64_t`'s and has a few other benefits. See <https://llvm.org/docs/ProgrammersManual.html#llvm-adt-smallvector-h>

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2718,6 +2790,43 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+            dimensionAssumedToBeBatch,
+            dimensionAssumedToBeChannel,
+        };
+
+        auto unknownSize = Torch::kUnknownSize;
+
+        // Compile-time check for dimensions of static size
+        for (auto eachDimension : nonResizableDimensions) {
```
zjgarvey, line 2811: nit:

I'm not really sure about the choice to use `each` for the iteration variable. It might make reading the initial `for` statement a bit more natural in English, but it does make the body quite awkward to read. I've never seen anyone use this convention (in any repository), so I'm a bit curious what the motivation is? 

In my opinion, the reading comprehension difficulty is not in the setup of the `for` loop, but within the body, so it makes sense to try and maximize readability there instead. E.g., `currDim` or `iterDim` would be preferable in my opinion to `eachDim`, since `eachDim` is like an incomplete universal quantifier and not an individual "thing". My top preference: if there isn't some other loop-independent `dim` that we need to disambiguate with, then why not just use `dim` as the loop variable?

zjgarvey (changes requested):
Hey, Jacob! Thanks for factoring out the broader changes. The logic makes sense to me, but I have several comments since you made quite a few decisions that need to be addressed. 

I tried to label these comments to help prioritize importance. The code functions properly, so most of these are `nit` suggestions (I assume "nitpick"). Please try to address comments with priority reflective of the label. Unqualified nits warrant either a brief comment or small change without discussion. Medium nits at least require a discussion and likely require a code change. Important nits definitely require a code change. 

If I left any comments unqualified, try to prioritize them based on context.

test/Conversion/TorchOnnxToTorch/simple_ops_q_to_z.mlir
```diff
@@ -2256,7 +2256,7 @@ func.func @test_sce_mean_3d_log_prob(%arg0: !torch.vtensor<[3,5,2],f32>, %arg1:
 // CHECK-LABEL: func.func @test_resize_sizes_nearest
   func.func @test_resize_sizes_nearest(%arg0: !torch.vtensor<[1,1,2,4],f32>, %arg1: !torch.vtensor<[4],si64>) -> !torch.vtensor<[?,?,?,?],f32> attributes {torch.onnx_meta.ir_version = 7 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
     %none = torch.constant.none
-    // CHECK: torch.aten.__interpolate.size_list_scale_list %arg0, %4, %none_0, %str, %false, %none_0, %false : !torch.vtensor<[1,1,2,4],f32>, !torch.list<int>, !torch.none, !torch.str, !torch.bool, !torch.none, !torch.bool -> !torch.vtensor<[?,?,?,?],f32>
+    // CHECK: torch.aten.__interpolate.size_list_scale_list %arg0, %12, %none_0, %str, %false, %none_0, %false : !torch.vtensor<[1,1,2,4],f32>, !torch.list<int>, !torch.none, !torch.str, !torch.bool, !torch.none, !torch.bool -> !torch.vtensor<[?,?,?,?],f32>
```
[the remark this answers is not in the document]
bjacobgordon: I had the same thought! Will do.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2718,6 +2790,43 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+          auto eachSizeOfInputTensor = sizesOfInputTensor[eachDimension];
+          auto eachSizeOfOutputTensor = sizesOfOutputTensor[eachDimension];
+
+          if (eachSizeOfInputTensor == unknownSize ||
+              eachSizeOfOutputTensor == unknownSize) {
+            continue;
+          } else if (eachSizeOfInputTensor == eachSizeOfOutputTensor) {
+            continue;
+          }
```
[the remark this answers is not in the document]
bjacobgordon: Sure, will remove braces either way

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2718,6 +2790,43 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+
+        // Compile-time check for dimensions of static size
+        for (auto eachDimension : nonResizableDimensions) {
+          auto eachSizeOfInputTensor = sizesOfInputTensor[eachDimension];
+          auto eachSizeOfOutputTensor = sizesOfOutputTensor[eachDimension];
+
+          if (eachSizeOfInputTensor == unknownSize ||
+              eachSizeOfOutputTensor == unknownSize) {
+            continue;
+          } else if (eachSizeOfInputTensor == eachSizeOfOutputTensor) {
```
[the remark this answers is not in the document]
bjacobgordon: Ah yes, was thinking about how to explain this. I think the use of `else if` here is a consideration that's _almost_ defined in the LLVM coding standards, but not quite. It's more general though, so even though I think it scales well as a rule of thumb, it's not critical. I'll remove it for now and we can discuss elsewhere!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -180,53 +180,126 @@ LogicalResult reduceOpImpl(OpBinder binder, ConversionPatternRewriter &rewriter,
...
+  auto typeOfScaleFactors =
+      cast<Torch::BaseTensorType>(givenScaleFactors.getType());
 
+  Type typeOfSelectionFromScaleFactors =
+      typeOfScaleFactors.getWithSizesAndDtype(
+          ArrayRef<int64_t>{1}, typeOfScaleFactors.getOptionalDtype());
+
+  auto loc = binder.getLoc();
+
+  Value zeroAsOp =
```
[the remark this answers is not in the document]
bjacobgordon: Done!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -180,53 +180,126 @@ LogicalResult reduceOpImpl(OpBinder binder, ConversionPatternRewriter &rewriter,
...
+      rewriter.create<Torch::ConstantIntOp>(loc, rewriter.getI64IntegerAttr(0));
```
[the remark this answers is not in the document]
bjacobgordon: Great tip, will do!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -180,53 +180,126 @@ LogicalResult reduceOpImpl(OpBinder binder, ConversionPatternRewriter &rewriter,
...
+      loc, typeOfTargetSize, selectionFromTargetSizesAsOp);
+
+  Value originalSizeAsOp = rewriter.create<Torch::AtenSizeIntOp>(
+      loc, givenOriginalTensor, givenDimensionAsOp);
+
+  Type typeOfComparisonResult = rewriter.getType<Torch::BoolType>();
+
+  return rewriter.create<Torch::AtenEqIntOp>(loc, typeOfComparisonResult,
+                                             targetSizeAsOp, originalSizeAsOp);
+}
```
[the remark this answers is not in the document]
bjacobgordon: Great catch! Will DRY this up a bit.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -180,53 +180,126 @@ LogicalResult reduceOpImpl(OpBinder binder, ConversionPatternRewriter &rewriter,
...
-    ValueList = rewriter.create<Torch::PrimListConstructOp>(
-        binder.getLoc(), Torch::ListType::get(Torch::FloatType::get(context)),
-        itemList);
+    typeOfEveryTransformation = rewriter.getType<Torch::FloatType>();
+    typeOfEveryFilteredTransformation = Torch::FloatType::get(context);
+  }
+
+  SmallVector<Value> filteredTransformations;
+
+  auto numberOfUnsupportedDimensionsFromFront = 2;
```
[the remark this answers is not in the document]
bjacobgordon: Will do something to that effect!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -180,53 +180,126 @@ LogicalResult reduceOpImpl(OpBinder binder, ConversionPatternRewriter &rewriter,
...
+
+  Type typeOfComparisonResult = rewriter.getType<Torch::BoolType>();
+
+  return rewriter.create<Torch::AtenEqIntOp>(loc, typeOfComparisonResult,
+                                             targetSizeAsOp, originalSizeAsOp);
+}
+
+Value withUnsupportedDimensionsFilteredOut(
+    Value givenTransformationVector, OpBinder binder,
+    ConversionPatternRewriter &rewriter) {
```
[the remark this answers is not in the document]
bjacobgordon: Will find a better name after generalizing the function!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2806,26 +2910,63 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        Value filteredSizesAsOp = noneVal;
+
         if (operands.size() < 4) {
-          Value scaleOperand = operands[2];
-          scalesValueList = getValueList(binder, rewriter, scaleOperand);
-          sizesValueList = noneVal;
+          Value proposedScaleFactorsAsOp = operands[2];
+
+          // run-time scale factor check for dynamic sizes
+          for (auto eachDimension : nonResizableDimensions) {
```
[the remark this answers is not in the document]
bjacobgordon: Will do!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -180,53 +180,126 @@ LogicalResult reduceOpImpl(OpBinder binder, ConversionPatternRewriter &rewriter,
...
+  Type typeOfComparisonResult = rewriter.getType<Torch::BoolType>();
+
+  return rewriter.create<Torch::AtenEqIntOp>(loc, typeOfComparisonResult,
+                                             targetSizeAsOp, originalSizeAsOp);
+}
+
+Value withUnsupportedDimensionsFilteredOut(
+    Value givenTransformationVector, OpBinder binder,
+    ConversionPatternRewriter &rewriter) {
+  auto typeOfTransformationVector =
```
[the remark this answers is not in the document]
bjacobgordon: Yup, should be able to simplify!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2750,35 +2859,31 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
               binder.op, "unimplemented: cubic coeff must be -0.75");
         }
 
-        unsigned rank = dyn_cast<Torch::ValueTensorType>(operands[0].getType())
-                            .getSizes()
-                            .size();
+        auto loc = binder.getLoc();
```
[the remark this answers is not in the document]
bjacobgordon: We'll keep this as `auto` since the intellisense correctly shows `mlir::Location` as the inferred type.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: Nice, will do!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2750,35 +2859,31 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
         if (mode == "cubic") {
           std::string modeStr = "cubic";
           if (coordTfMode != "half_pixel")
             modeStr = modeStr + "_" + coordTfMode;
-          modeStrValue =
-              rewriter.create<Torch::ConstantStrOp>(binder.getLoc(), modeStr);
+          modeStrValue = rewriter.create<Torch::ConstantStrOp>(loc, modeStr);
         }
+
+        unsigned rankOfInputTensor = sizesOfInputTensor.size();
```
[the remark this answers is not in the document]
bjacobgordon: 1. Looked into this, looks like it already infers `size_t` which I believe is just an alias for `uint64_t`! Will use `auto` instead.
2. I usually optimize readability for where it's used (switch condition) rather than where it's defined. Definitions ought to have some redundancy since that ensures "semantic alignment". Will keep it as `rankOfInputTensor`.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2806,26 +2910,63 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+              proposedScaleFactorsAsOp, binder, rewriter);
         } else {
-          Value sizeOperand = operands[3];
-          scalesValueList = noneVal;
-          sizesValueList = getValueList(binder, rewriter, sizeOperand);
+          Value proposedSizesAsOp = operands[3];
+
+          // run-time target size check for dynamic sizes
+          for (auto eachDimension : nonResizableDimensions) {
+            auto eachSizeComparisonAsOp =
```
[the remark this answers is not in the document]
bjacobgordon: Will leave this as `auto`:
- this is a scenario where "the type [is]" abstracted away.
- the very next example appears to use auto in a for loop as well!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2686,12 +2759,11 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
       });
   patterns.onOp(
       "Resize", 11, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-        Torch::ValueTensorType resultType;
+        Torch::ValueTensorType typeOfOutputTensor;
```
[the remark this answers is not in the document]
bjacobgordon: I agree that it's redundant! I played around with a lot of options, I think I found a more concise way that leverages the fact that an instance of a meta type is itself a class.

And I've seen this done elsewhere in the codebase, so should be more consistent!

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
bjacobgordon: Great topic! I highly recommend keeping this as is. Discussion below:

## Motivations:
- to lean more towards readability: it's short and reads more like plain human language.
  - "Wait, does 'iter' mean 'iterator', 'iteration'?"
  - "I'm guessing 'curr' means "current", "currency'?"
  - "ah 'eachDim', like the dimension for each iteration of the loop"
- to record intended scope: "this is declared specifically for this loop"
- to make it easier to see when something doesn't need to be in the loop itself
  - "Wait, 'each' doesn't make sense here. This thing is the same for _every_ element, regardless of which loop we're on. Let's pull it out for performance/readability"
- to avoid "adjective collisions"
  - "current": used a ton in stateful apps, used to mean "some instance of state [probably cached from the database]".
- to make each line more semantically independent: bake-in the context of the loop to avoid having to go to the loop declaration for context
   - With `dim` alone, we'd have to go to the top of the loop to realize that it's the "subject" of each loop

## Also
In general, I've found that the best-written classes/typealiases are context-agnostic compared to their instances, which are only instantiated for a specific scope/purpose and therefore have inherent context. Hence, I add some kind of adjective to the name to capture that scope/purpose/context.

bjacobgordon:
Ready, @zjgarvey! This is stacked on #3987. Once that's merged, about ~20 commits will actually already be in `main` when it's time to re-review _this_ PR. 👍

zjgarvey:
> Ready, @zjgarvey! This is stacked on #3987. Once that's merged, about ~20 commits will actually already be in `main` when it's time to re-review _this_ PR. 👍

Sounds good. I think with a few small changes, we can merge #3987 . After that PR gets merged and this PR gets rebased, I'll re-review.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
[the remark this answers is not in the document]
zjgarvey: I meant the result of the function call. It isn't immediately obvious what the function returns from this scope, so using an explicit type would improve readability.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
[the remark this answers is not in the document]
zjgarvey: Yeah, using auto here is fine since it is a `size` of a container. 

I really don't care that much about the name change other than the fact that you only changed it based on an opinion that I don't agree with. We can keep it, but please try to minimize forcing your personal naming conventions on parts of the code that do not need functional restructuring. It requires unnecessary conversations that delay the progress of this PR.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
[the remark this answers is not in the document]
zjgarvey: Where else have you seen this in the codebase? I have never seen this convention used here. Personally, I think it is confusing to read and possibly incorrect. 

If you want to talk about consistency, notice that every single pattern in this source file contains the line `Torch::ValueTensorType resultType;`!

zjgarvey:
@bjacobgordon I revisited a few conversations and responded/unresolved some of them if you wanted to take a look.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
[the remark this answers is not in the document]
bjacobgordon: Oh, I gotcha! I agree, will do that in spots like this. (This particular one went away after extracting common logic)

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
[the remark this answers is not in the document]
bjacobgordon: I think I saw it under externals, which probably means that convention is not relevant like you were saying.

Good points, how about `outputTensorType` so it parallels the existing `inputTensorType`?

bjacobgordon:
Hey, @zjgarvey, I've rebased onto `main`!

To make it easier for you to re-review, I _could_ segment _this_ PR into a stack of _3_ PRs. It'll be really easy; the commits are already primed for it!

Would you like me to do that for you?

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -2820,33 +2850,84 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
...
+        Value supportedScaleFactors;
+        Value supportedSizes;
+
+        Value noneVal = rewriter.create<Torch::ConstantNoneOp>(loc);
+
+        if (numberOfOperands == 3) {
+          Value proposedScaleFactors = operands[2];
+
+          Value scaleIdentity = rewriter.create<Torch::ConstantFloatOp>(
+              loc, rewriter.getF64FloatAttr(1.0));
```
ScottTodd, line 2871: FYI this appears to have caused some test failures downstream in the IREE project on https://github.com/iree-org/iree/pull/19976. I did not bisect to this specific change or line of code, but this looked most relevant. These are the logs: https://github.com/iree-org/iree/actions/runs/13292751088/job/37117771168?pr=19976#step:8:50

````
 _ IREE compile and run: test_resize_downsample_scales_cubic_align_corners::model.mlir::model.mlir::cpu_llvm_sync _
[gw2] linux -- Python 3.11.11 /home/runner/work/iree/iree/venv/bin/python
Error invoking iree-compile
Error code: 1
Stderr diagnostics:
<unknown>:0: error: failed to legalize operation 'torch.constant.float'
<unknown>:0: note: see current operation: %6 = "torch.constant.float"() <{value = 1.000000e+00 : f64}> : () -> !torch.float


Stdout diagnostics:


Test case source:
  https://github.com/iree-org/iree-test-suites/blob/main/onnx_ops/onnx/node/generated/test_resize_downsample_scales_cubic_align_corners

Input program:
```
module {
  func.func @test_resize_downsample_scales_cubic_align_corners(%arg0: !torch.vtensor<[1,1,4,4],f32>, %arg1: !torch.vtensor<[4],f32>) -> !torch.vtensor<[1,1,3,3],f32> attributes {torch.onnx_meta.ir_version = 9 : si64, torch.onnx_meta.opset_version = 19 : si64, torch.onnx_meta.producer_name = "backend-test", torch.onnx_meta.producer_version = ""} {
    %none = torch.constant.none
    %0 = torch.operator "onnx.Resize"(%arg0, %none, %arg1) {torch.onnx.coordinate_transformation_mode = "align_corners", torch.onnx.mode = "cubic"} : (!torch.vtensor<[1,1,4,4],f32>, !torch.none, !torch.vtensor<[4],f32>) -> !torch.vtensor<[1,1,3,3],f32> 
    return %0 : !torch.vtensor<[1,1,3,3],f32>
  }
}

```

Compiled with:
  cd /home/runner/work/iree/iree/iree-test-suites/onnx_ops/onnx/node/generated/test_resize_downsample_scales_cubic_align_corners && iree-compile model.mlir --iree-hal-target-backends=llvm-cpu --iree-input-demote-f64-to-f32=false -o model_cpu_llvm_sync.vmfb
````

By default, IREE demotes f64 to f32 as 64 bits of precision is rarely needed in ML models and many hardware targets either do not support f64 at all or support it with significant performance penalties. The tests there do override that default by setting `--iree-input-demote-f64-to-f32=false` though.

Is f64 needed here, or would f32 work? I see lots of uses of `f64` in this file 🤔

ScottTodd: More context: some of the tests in the ONNX test suite _require_ f64, which is why we run the tests without f64 to f32 demotion: https://github.com/iree-org/iree/pull/18111.

zjgarvey: We dont need f64, this is a small bug with the changes. Will post a quick fix in a minute.

bjacobgordon: If I remember correctly when writing this, using f32 for `scaleIdentity` caused a test case or two within torch mlir to fail.

@zjgarvey Any insights here?

zjgarvey: Looks like a simple issue. `AtenEqFloatOp` doesn't have a lowering, but it should be easy to add. I'll post a PR shortly.

zjgarvey: https://github.com/llvm/torch-mlir/pull/4022

ScottTodd: Nice, thanks!

PR #3946, opened 2025-01-08
[TOSA] : Fix float to integer cast for `torch.ops.aten.to` lowering.

sahas3:
The behavior of float -> integer cast in PyTorch (though I haven't found the actual code implementing the cast) appears to be (based on the results produced in PyTorch):

1. round the float nearest to zero (similar to `arith.fptosi/ui`) 
2. then perform the conversion

Currently we only emit `tosa.cast` for this operation but as per the spec https://www.mlplatform.org/tosa/tosa_spec.html#_cast the rounding performed for float -> integer is round to nearest integer (not zero). Hence, the current TOSA lowering for `torch.ops.aten.to` produces incorrect answer.


justin-ngo-arm:
Hi @sahas3, for the Torch float to integer cast behavior, did you observe it to be the same for all Torch casting, or just `torch.aten.to`? This is because I plan to make `tosaCastTensorToType` the universal `tosa.cast` call within TorchToTosa (plan described in https://github.com/llvm/torch-mlir/pull/3948), and if the behavior you observed is specific to `aten.to` then I might need to re-evaluate my approach to unifying these calls.

sahas3:
> Hi @sahas3, for the Torch float to integer cast behavior, did you observe it to be the same for all Torch casting, or just `torch.aten.to`? This is because I plan to make `tosaCastTensorToType` the universal `tosa.cast` call within TorchToTosa (plan described in #3948), and if the behavior you observed is specific to `aten.to` then I might need to re-evaluate my approach to unifying these calls.

I think `torch.aten.to` or `torch.tensor.to` ops are the basis of all casts in PyTorch. I haven't seen different behavior in my experiments yet. Some additional tests also started passing with this PR indicating that the cast behavior is same for other cast like ops too. So I think using `tosaCastTensorToType` universally for all `tosa.cast` calls is the correct approach.

sjarus (commented):
If Torch mandates round towards zero, can that not be synthetically expressed here @sahas3 ?

sahas3:
> If Torch mandates round towards zero, can that not be synthetically expressed here @sahas3 ?

Hi @sjarus , can you elaborate more on what you mean by synthetically expressing rounding towards zero?

sjarus:
Looks like we had a misunderstanding earlier @sahas3 . Approving.

sjarus (approved):

PR #3948, opened 2025-01-08
[TOSA] Update tosa.cast check according to TOSA v1.0 spec

justin-ngo-arm:
* Update checkValidityOfCast function for tosa.cast according to the latest TOSA v1.0 spec: https://www.mlplatform.org/tosa/tosa_spec.html#_cast
* Clean up some dead code in TorchToTosa

Change-Id: I41209c698a694bca57ebf49ed3608cf89a0d8ba8


lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
```diff
@@ -274,16 +274,28 @@ static LogicalResult checkValidityOfCast(Type src, Type dest) {
...
+      (src.isInteger(32) && dest.isInteger(16)) ||
+      (src.isInteger(32) && dest.isInteger(8)) ||
       (src.isInteger(32) && dest.isInteger(1)) ||
       (src.isInteger(32) && dest.isF32()) ||
+      (src.isInteger(32) && dest.isF16()) ||
       (src.isInteger(32) && dest.isBF16()) ||
       // int16 -> *
       (src.isInteger(16) && dest.isBF16()) ||
+      (src.isInteger(16) && dest.isF32()) ||
+      (src.isInteger(16) && dest.isF16()) ||
```
sahas3, line 286: Can you also add 
1. signed 16 to signed 32
2. signed 16 to signed 8

lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
```diff
@@ -274,16 +274,28 @@ static LogicalResult checkValidityOfCast(Type src, Type dest) {
...
       (src.isInteger(8) && dest.isInteger(1)) ||
       (src.isInteger(8) && dest.isBF16()) ||
+      (src.isInteger(8) && dest.isF32()) ||
+      (src.isInteger(8) && dest.isF16()) ||
       // int1 -> *
       (src.isInteger(1) && dest.isInteger(64)) ||
+      (src.isInteger(1) && dest.isInteger(32)) ||
+      (src.isInteger(1) && dest.isInteger(16)) ||
+      (src.isInteger(1) && dest.isInteger(8)) ||
       (src.isInteger(1) && dest.isF32()) ||
```
sahas3, line 299: Not seeing this in the spec -- has this been removed?

sahas3 (changes requested):

lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
justin-ngo-arm: https://www.mlplatform.org/tosa/tosa_spec.html#_cast
This is still in the spec as I just checked.

lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
justin-ngo-arm: Thank you for catching this.

justin-ngo-arm:
I'm quoting my comment regarding unifying `tosa.cast` usage within TorchToTOSA from https://github.com/llvm/torch-mlir/pull/3949:

>  After some discussion with @sjarus, I've decided to combine tosaCastTensorToType and promoteType together into one function wrapped around the tosa.cast op creation process. The combined function will still be called tosaCastTensorToType as this name is more descriptive about what this function does. I plan to replace all usages of tosa.cast creation with this "new" function to make everything uniform.
> 
> The following are the differences between the current promoteType and tosaCastTensorToType functions:
> 
> - promoteType checks if the input and output types are same before performing the cast, which will prevent unnecessary cast operation if both types are the same.
> - tosaCastTensorToType has the cast validity check.
>
> My plan is to take the same-type check from promoteType and add it to tosaCastTensorToType, then remove the promoteType function altogether. As for the cast validity check, I will tighten it more so that it will strictly follow the TOSA v1.0 spec (i.e. removing all I64 and F64 casting as they are not allowed in TOSA). I will update that in https://github.com/llvm/torch-mlir/pull/3948. However, with this tightening, many e2e tests will fail. This is not desirable as these operations, although are not congruent with the spec,are still permissible. Therefore, after tightening up the checkValidityOfCast function, I will leave it out of the tosaCastTensorToType function for now and enable it later with a potential --strict mode that is defaulted to off. TOSA validation should flag illegal constructs based on each profile anyway, so this --strict mode is just another guard before that.
> 
> I will begin the new tosaCastTensorToType work in a separate PR from https://github.com/llvm/torch-mlir/pull/3948, so that it is easier to keep track of changes and progress.

projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -1704,6 +1704,21 @@
 # Write the TOSA set as a "passing" set as it is very early in development
 # and very few tests work yet.
 TOSA_PASS_SET = {
+    "AtenEyeMModuleInt2D_basic",
+    "AtenEyeModuleInt2D_basic",
+    "ElementwiseWhereScalarOtherStaticModule_basic",
+    "FullModuleFalsePinMemory_basic",
+    "FullModuleInt2D_basic",
+    "MaskedFillScalarFloatValueModule_basic",
+    "MaskedFillScalarFloatValueStaticModule_basic",
```
sahas3, line 1713: Clarification: all these tests are now passing because `checkValidityOfCast` is not enforced?

sahas3 (approved):

PR #3950, opened 2025-01-09
[MLIR][TORCH] Add support for `enable_gqa` flag in SDPA op

vivekkhandelwal1:


pashu123 (approved):
LGTM! Are there any reasons why we aren't decomposing this torch op into another set of torch ops?

vivekkhandelwal1:
> LGTM! Are there any reasons why we aren't decomposing this torch op into another set of torch ops?

AFAIK, the reason for not doing that is since we want the attention op as a single kernel. Hence, we just lower it to `tm_tensor.attention` and the rest is taken care of during the codegen.

PR #3951, opened 2025-01-09
Remove TOSA make_fx configuration

mgehre-amd:
We are not using this configuration downstream anymore, and it's not tested upstream.

I propose to remove it.


sjarus (approved):
Looks fine to me. We almost exclusively focus on fx_importer_tosa target . The make_fx_tosa was a stopgap construct that no longer has a use case.

PR #3955, opened 2025-01-13
[ONNX] improve regex matching in onnx-importer name sanitization

zjgarvey:
Instead of adding unsupported characters on a case-by-case basis, we should replace anything that isn't alphanumeric, `_`, or `.`.


jinchen62 (approved):
LGTM, thanks!

PR #3967, opened 2025-01-17
Migrate ci.yml to AKS cpubuilder cluster.

saienduri:
This commit contains the ccache and python changes neccessary to migrate the ci.yml workflow to an azure kubernetes cpubuilder cluster.
Timings before and after the infra change are in the same ballpark (~10 mins).
Previously, `test/python/compile.py` was unnecessarily excluding the python site packages, which needed to be changed for the cluster.
We also make use of `hendrikmuhs/ccache-action` now and use a key of `${{ github.job }}-${{ matrix.torch-version }}` which works as intended: https://github.com/llvm/torch-mlir/actions/runs/12828039126/job/35771248056#step:7:4657

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,20 +23,10 @@ jobs:
       matrix:
         torch-version: [nightly, stable]
     name: Build and Test (Linux, torch-${{ matrix.torch-version }}, assertions)
-    runs-on: torch-mlir-cpubuilder-manylinux-x86-64
+    runs-on: torch-mlir-cpubuilder-linux-x86-64-scale
     env:
       CACHE_DIR: ${{ github.workspace }}/.container-cache
     steps:
-      - name: Configure local git mirrors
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -50,11 +40,25 @@ jobs:
           restore-keys: |
             build-test-cpp-asserts-manylinux-${{ matrix.torch-version }}-v2-
 
+      - name: "Setting up Python"
```
marbre, line 43: Can't we use `actions/setup-python` on the cluster?

saienduri: We need `python3-dev` for torch-mlir, which doesn't come as part of the `actions/setup-python`, so easier to do manually. We didn't utilize it before either

marbre: Got it.

marbre (approved):

PR #3970, opened 2025-01-17
Update default Python versions

marbre:
Drops Python 3.8, which is EOL, see
https://devguide.python.org/versions/, and adds Python 3.12, which is the default e.g. on Ubuntu 24.04 LTS.

### DIFF
--- a/build_tools/python_deploy/build_linux_packages.sh
+++ b/build_tools/python_deploy/build_linux_packages.sh
@@ -16,7 +16,7 @@
 #   ./build_tools/python_deploy/build_linux_packages.sh
 #
 # Build specific Python versions and packages to custom directory:
-#   TM_PYTHON_VERSIONS="cp38-cp38 cp39-cp39" \
+#   TM_PYTHON_VERSIONS="cp39-cp39 cp310-cp310" \
 #   TM_PACKAGES="torch-mlir" \
 #   TM_OUTPUT_DIR="/tmp/wheelhouse" \
 #   ./build_tools/python_deploy/build_linux_packages.sh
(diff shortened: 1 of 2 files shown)


build_tools/python_deploy/build_linux_packages.sh
```diff
@@ -46,7 +46,7 @@ TM_RELEASE_DOCKER_IMAGE="${TM_RELEASE_DOCKER_IMAGE:-quay.io/pypa/manylinux2014_$
 # ./build_tools/docker/Dockerfile
 TM_CI_DOCKER_IMAGE="${TM_CI_DOCKER_IMAGE:-powderluv/torch-mlir-ci:latest}"
 # Version of Python to use in Release builds. Ignored in CIs.
-TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp38-cp38 cp310-cp310 cp311-cp311}"
+TM_PYTHON_VERSIONS="${TM_PYTHON_VERSIONS:-cp310-cp310 cp311-cp311 cp312-cp312}"
```
ScottTodd, line 49: > Drops Python 3.8, which is EOL, see
> https://devguide.python.org/versions/

Sounds good to me, though I did notice a revert in llvm/llvm-project recently due to something breaking on 3.8: https://github.com/llvm/llvm-project/commit/3f1486f08e0dd64136fb7f50e38cd618dd0255d2

ScottTodd (approved):

build_tools/python_deploy/build_linux_packages.sh
marbre: This was re-landed with llvm/llvm-project@f136c80, but I think even if upstream MLIR supports Python 3.8, Torch-MLIR does not need to include Python 3.8 in its defaults. The patch actually does not drops support Python 3.8 but also llvm/torch-mlir-release does not build releases for 3.8.

marbre: @ScottTodd as the project requires conversations to be resolved prior to merging, is it fine to mark as resolved?

ScottTodd: Yes

PR #3974, opened 2025-01-20
fix dangling arrayref by extending temporary object lifetime

annuasd:
In the following code, makeShapeTorchCompatible returns a SmallVector, while indexShape is an ArrayRef, which will lead to a lifetime bug.


lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -4025,7 +4025,7 @@ LogicalResult ConvertAtenOp<AtenIndexSelectOp>::matchAndRewrite(
     return rewriter.notifyMatchFailure(
         op, "Only RankedTensorType indices are currently supported");
 
-  auto inputShape = inputType.getShape();
+  auto inputShape = SmallVector<int64_t>(inputType.getShape());
```
IanWood1, line 4028: did you mean to change `indexShape` on line 4022?

annuasd: Yes. I made a mistake in previous commit.Already update commit.

PR #3975, opened 2025-01-20
refactor(ONNX): enforces `opLocation` naming convention

bjacobgordon:
1. `loc`: when searching "loc" across the codebase, this appears in contexts where it could mean "location", "locale", "lines of code", etc.
    - Avoiding abbreviations reduces this ambiguity.
2. `location`: Now the question to answer is "The location of what, exactly?" without having to ping a colleague or scrub the codebase. We see `auto location = binder.getLoc()`, so the (incorrect) inference is "ah the location of the binder, I guess", so we use that to modify the word "location"
3. `binderLocation`: Turns out that "binder" is short for "opBinder", meaning this is actually "location of _op_". So, we switch the "modifying noun"
4. `opLocation`: bakes in the understanding that probably required preceding engineers (like me) to spend their time (and others') clarifying.
    - Adding 7 characters to "loc" avoids time to grok, avoids unnecessary roping in of other engineers, and makes it easier to onboard other engineers.
    - Overall, this is easier to build upon from an organizational standpoint.

A longer, more specific name usually has less false positives in search vs a short generic one! This makes reversion easier than it was to do the conversion to begin with. On top of that, this particular name did not exist in the codebase prior to this commit, so the name was "available".

### DIFF
--- a/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
+++ b/lib/Conversion/TorchOnnxToTorch/DefaultDomainAtoF.cpp
@@ -24,7 +24,7 @@ LogicalResult windowFunctionImpl(OpBinder binder,
                                  Torch::ValueTensorType resultType,
                                  int64_t output_datatype, int64_t periodic) {
 
-  Location loc = binder.getLoc();
+  auto loc = binder.getLoc();
   ImplicitLocOpBuilder b(loc, rewriter);
 
   double isPeriodicFp = static_cast<double>(periodic);
@@ -355,7 +355,7 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
             binder.tensorResultTypeAtIndex(resultType, 0))
           return failure();
 
-        Location loc = binder.getLoc();
+        auto loc = binder.getLoc();
         Value cstFalse = rewriter.create<Torch::ConstantBoolOp>(loc, false);
         Value cstMomentum = rewriter.create<Torch::ConstantFloatOp>(
             loc, rewriter.getF64FloatAttr(momentum));
@@ -1290,7 +1290,7 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
       });
   patterns.onOp(
       "Conv", 1, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-        Location loc = binder.getLoc();
+        auto loc = binder.getLoc();
         Torch::ValueTensorType resultType;
         Value input, weight;
         int64_t group;
@@ -2325,7 +2325,7 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
                 });
   patterns.onOp(
       "Dropout", 12, [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-        Location loc = binder.getLoc();
+        auto loc = binder.getLoc();
         Torch::ValueTensorType resultType;
         int64_t numOperands = binder.op->getNumOperands();
         SmallVector<Value> operands;
@@ -2407,7 +2407,7 @@ void mlir::torch::onnx_c::populateDefaultDomainAtoF(
   patterns.onOp(
       "DynamicQuantizeLinear", 11,
       [](OpBinder binder, ConversionPatternRewriter &rewriter) {
-        Location loc = binder.getLoc();
+        auto loc = binder.getLoc();
         Value input;
         Torch::ValueTensorType resultType, scaleType, zeroPointType;
         if (binder.tensorOperand(input) ||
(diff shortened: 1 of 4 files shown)


bjacobgordon:
Okay, @zjgarvey, ready! We'll start here before getting to #3945?

zjgarvey:
In general, I disagree with the change from `loc` to `opLocation`, as I much prefer the shorter `loc`. The `mlir::Location` of a conversion pattern's target op is so ubiquitous that forcing the name to be any longer than `loc` would be a non-trivial developer burden. I have never experienced any ambiguity with the variable name `loc`; furthermore, the term `opLocation` itself passes the issue of ambiguity of "what is loc?" to "which op's location?". Attempting to resolve this ambiguity even further would result in an absurd variable name like `conversionRewritePatternTargetOpLocation`, when in reality, no ambiguity was ever present in `loc` to begin with. 

For cases outside of this subdirectory, there may be multiple `mlir::Locaton`'s in a single conversion pattern. E.g., when defining the payload of a linalg generic op, a location is passed as one of the lambda function arguments. Even still the name `loc` is re-used (even conflicting with the broader `loc`), because within that payload there is still no ambiguity over the location being used.

Because of these reasons, I would be unwilling to merge and subsequently enforce the conventions requested in this PR. However, I would be amenable to enforcing a convention like: 

_"In `lib/Conversion/TorchOnnxToTorch/`, the variable name `loc` is reserved for the `mlir::Location` of a conversion pattern's target op. E.g., `mlir::Location loc = binder.getLoc();`. If multiple instances of `binder.getLoc()` would be called in a single pattern, please either use an `ImplicitLocOpBuilder`, or store the location as a variable named `loc` for `rewriter.create<SomeOp>(loc, *args)`."_

bjacobgordon:
Thanks for taking the time to write that out, @zjgarvey! Love the detail here.

So, you want to keep using `loc` because you:
1. believe a longer name would require more work from developers
1. have never felt that the name was ambiguous
1. predict that lengthening the name to reduce ambiguity is a "slippery slope"

Do you feel that this is an accurate restatement/summary of your stance?

zjgarvey:
> Thanks for taking the time to write that out, @zjgarvey! Love the detail here.
> 
> So, you want to keep using `loc` because you:
> 
>     1. believe a longer name would require more work from developers
> 
>     2. have never felt that the name was ambiguous
> 
>     3. predict that lengthening the name to reduce ambiguity is a "slippery slope"
> 
> 
> Do you feel that this is an accurate restatement/summary of your stance?

The bullet point 3. is a bit of a mischaracterization. The general point is that it is inefficient to use a variable name to clarify the nature of what gets returned by a widely-used function. Fundamentally, there is zero ambiguity in `auto loc = binder.getLoc();`. If someone is new to this repository, they need to figure out what `binder.getLoc()` does anyway. 

This is a complaint I will likely raise elsewhere in <https://github.com/llvm/torch-mlir/pull/3945>. E.g. `Value someThingAsOp = rewriter.create<OpTy>...` doesn't need to specify `AsOp` unless there is already a variable called `someThing` of a different type. This is not to mention that the additional naming in this example is inaccurate since the instance of `OpTy` getting generated by `rewriter.create` is being implicitly cast to a `Value` via the `OneResult` trait. The point I'm reinforcing with this example is that the variable name should not be instructional about what `rewriter.create` does. This is an essential part of mlir and new developers need to figure out what it does anyway. `AsOp` or `AsValue` are only going to be filler words unless used to differentiate similar things, e.g. an `int64_t` from a corresponding `Torch::ConstantIntOp` result `Value`. These could both represent a `batchSize`, and you might need the name `batchSizeValue` in order to disambiguate. 

Feel free to continue this discussion here. However, I am going to close this for now as I do not believe it will be acceptable to merge this particular PR without substantial discussion or changes.

bjacobgordon:
@stellaraccident Right, that's what's interesting to me!

Evaluating standards is no easy lift, but it can be useful to poke and prod to see if they're still serving us in the long-term. Standards that go stale or lack context can become [indistinguishable from (maladaptive) traditions](https://scilithium.wordpress.com/2012/01/08/how-humans-out-sheep-the-sheep-five-monkeys-and-a-ladder-thought-experiment/).

i.e. Is this standard:
1. _implicitly_ upheld because of a common [psychological inertia](https://en.wikipedia.org/wiki/Psychological_inertia)?
  OR
2. _explicitly_ upheld because it's the decision we would have made again if we were back at the decision table with what we know now?
  OR
3. upheld for some other reason entirely?

From my perspective, the first point applies to this case, which leads me to ask:
What quick/easy adjustments can be made to the standard that reduce the amount of context that newcomers need to ingest?

stellaraccident:
It is an easy lift. I'm saying leave it as "loc" and not engaging in a linguistic argument.

PR #3979, opened 2025-01-22
Attempt to fix the Python wheels for Windows

marbre:


marbre:
Commit b17cf23 unfortunately broke the Windows builds, but this is fixed with the proposed patch, see https://github.com/marbre/torch-mlir-release/actions/runs/12907338009.

ScottTodd (approved):

PR #3990, opened 2025-01-29
[BUILD] Add nanobind to build-requirements

vivekkhandelwal1:


build-requirements.txt
```diff
@@ -5,6 +5,7 @@ setuptools
 cmake
 ninja
 packaging
+nanobind
```
marbre, line 8:
```suggestion
nanobind>=2.4, <3.0
```
Upstream is more specific about the versions and MLIR at HEAD needs nanobind>=2.4. Let's model this here as well.

marbre (changes requested):

marbre (approved):

PR #3991, opened 2025-01-29
Bump llvm to llvm/llvm-project@d4159e2a1d1d

pashu123:


lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
```diff
@@ -500,17 +511,5 @@ LogicalResult getConvOpsAccType(PatternRewriter &rewriter,
   return success();
 }
 
-// Temporary function to get TOSA const shape
-// TODO: Remove this function when getTosaConstShape is available in
-// externals/llvm-project/mlir/include/mlir/Dialect/Tosa/Utils/ConversionUtils.h
-Value getTosaConstShape(PatternRewriter &rewriter, Location loc,
```
praveen-g-ctt, line 506: Is this removed by mistake ?

pashu123: A function has already been added upstream.

pashu123: with the same signature.

lib/Conversion/TorchToTosa/TorchToTosa.cpp
```diff
@@ -3349,9 +3378,9 @@ LogicalResult ConvertAtenOp<AtenGeluOp>::matchAndRewrite(
         rewriter.create<tosa::PowOp>(op->getLoc(), resultType, self, three);
 
     // 0.044715 * x^3
-    auto inputPowThreeMul =
-        rewriter.create<tosa::MulOp>(op->getLoc(), resultType, magicNumber,
-                                     inputPowThree.getResult(), /*shift=*/0);
+    auto inputPowThreeMul = rewriter.create<tosa::MulOp>(
+        op->getLoc(), resultType, magicNumber, inputPowThree.getResult(),
+        /*shift=*/tosa::getTosaConstTensorSingleI32(rewriter, op, 0));
```
praveen-g-ctt, line 3383: The shift operand of tosa::MulOp expects int8 type 

https://github.com/llvm/llvm-project/blob/d4159e2a1d1d640077b2e5cde66b0a284049955f/mlir/include/mlir/Dialect/Tosa/IR/TosaOps.td#L840

lib/Conversion/TorchToTosa/TosaLegalizeUtils.cpp
praveen-g-ctt: Got it. We can remove the declaration too in ~/torch-mlir/include/torch-mlir/Conversion/TorchToTosa/TosaLegalizeUtils.h ?

PR #3995, opened 2025-01-31
[FxImporter] remove weakref finalizer of reftracker

egebeysel:
Fixes https://github.com/iree-org/iree-turbine/issues/281.

**_TL;DR:_**

The `weakref.finalize` objects cause the model parameters to be kept in memory in-between consecutive `aot.export` calls in the same process. We remove them to enable releasing the memory, this does not change the behavior of `RefTracker` or `RefMapping` classes anyhow else.

### DIFF
--- a/python/torch_mlir/extras/fx_importer.py
+++ b/python/torch_mlir/extras/fx_importer.py
@@ -2247,13 +2247,15 @@ class RefTracker:
         if existing:
             return existing
         info = RefMapping(referrent)
-        if referrent is not Empty:
-            weakref.finalize(referrent, self._ref_finalizer, ref_id)
+        # Finalizer is removed due to a memory leak
+        # See: https://github.com/iree-org/iree-turbine/issues/281
+        # if referrent is not Empty:
+        #    weakref.finalize(referrent, self._ref_finalizer, ref_id)
         self._refs[ref_id] = info
         return info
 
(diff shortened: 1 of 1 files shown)


egebeysel:
@stellaraccident could you review/approve and merge if everything looks alright :) Thanks!

penguin-wwy:
Could you comment out this section of code and add the issue link in the comments instead of deleting it directly?

stellaraccident:
> Could you comment out this section of code and add the issue link in the comments instead of deleting it directly?

This is a good idea: let's do that. I think this is safe to do, but this part was complicated. Let's plan for our future selves to have some context.

If changed as requested, then this lgtm

egebeysel:
> Could you comment out this section of code and add the issue link in the comments instead of deleting it directly?

> This is a good idea: let's do that. I think this is safe to do, but this part was complicated. Let's plan for our future selves to have some context.

Makes sense, updated it :+1:

egebeysel:
I don't think the failing CI is relevant to this specific change... Can we rerun?

penguin-wwy:
> I don't think the failing CI is relevant to this specific change... Can we rerun?

The CI issue has been fixed in #4008. You can update the code and push it again.

penguin-wwy (approved):

PR #3999, opened 2025-02-04
Support batch and classes for NonMaxSuppression

praveen-g-ctt:


praveen-g-ctt:
Following tests are passing 

https://github.com/iree-org/iree-test-suites/tree/main/onnx_ops/onnx/node/generated/test_nonmaxsuppression_two_batches
https://github.com/iree-org/iree-test-suites/tree/main/onnx_ops/onnx/node/generated/test_nonmaxsuppression_two_classes


The following test fails in IREE for cpu due to large tensor size used for stack allocation for sort operation. Tested for smaller tensor sizes and it is working as expected with correct results

https://github.com/nod-ai/SHARK-ModelDev/issues/891

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -3810,87 +3801,218 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+        auto numBatches =
+            rewriter.create<Torch::AtenSizeIntOp>(loc, scores, cst0);
+        auto numClasses =
+            rewriter.create<Torch::AtenSizeIntOp>(loc, scores, cst1);
+
+        // Create an empty tensor of shape (B*C*N, 3) to store the final result.
+        // We slice this to required elements at the end
+
+        // FIXME:: Currently empty tensors created with dynamic sizes are not
+        // fully supported. Uncomment the below lines once dynamic sizes for
```
praveen-g-ctt, line 3817: Issue is similar to https://github.com/nod-ai/SHARK-ModelDev/issues/876

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -3810,87 +3801,218 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+                loc, nmsTy, boxValue, scoreValue, iouThreshold);
+
+            // Compute NumOutputBoxes
+            Value numOutputBoxes =
+                rewriter.create<Torch::AtenSizeIntOp>(loc, result, cst0);
+            numOutputBoxes = rewriter.create<Torch::PrimNumToTensorScalarOp>(
+                loc, emptyTensorTy, numOutputBoxes);
+            Value maxBoxesPerClass =
+                rewriter.create<Torch::PrimNumToTensorScalarOp>(
+                    loc, emptyTensorTy, maxOutputBoxesPerClass);
```
jinchen62, line 3921: I would use tensor type with shape [1] instead of [] since those few arguments are coming with [1] and you do Minimum op with them.

jinchen62 (commented):
Left few comments, and I'm not quite clear about the nmsLoop part. @zjgarvey Would be nice to have your review too.

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -3810,87 +3801,218 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            auto minVal = rewriter.create<Torch::AtenMinimumOp>(
+                loc, numOutputBoxes.getType(), numOutputBoxes,
+                maxBoxesPerClass);
+            numOutputBoxes =
+                rewriter.create<Torch::AtenItemOp>(loc, intTy, minVal);
+
+            // Loop through the nms result
+            auto nmsLoop = rewriter.create<Torch::PrimLoopOp>(
```
zjgarvey, line 3929: I'm finding it difficult to parse this loop.

The result of the (per-batch per-channel) torchvision nms op has shape `<num_selected>`, and we need it to be `<num_selected x 3>`, where each triple is like `[batch_index, class_index, selected_box_index]`. Is the purpose of this loop to insert these elements into the final result? Is it possible to avoid using a loop for this and instead concatenate the nms result with some splat tensors, then insert that into the final result by keeping track of what the cumulative `num_selected` is?

test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
```diff
@@ -2034,53 +2034,74 @@ func.func @test_loop_forlike(%arg0: !torch.vtensor<[],si64>, %arg1: !torch.vtens
...
-  // CHECK:           %[[VAL_34:.*]] = torch.aten.size.int %[[VAL_33]], %[[VAL_24]] : !torch.vtensor<[1,1],si64>, !torch.int -> !torch.int
-  // CHECK:           %[[VAL_35:.*]] = torch.constant.int 2
-  // CHECK:           %[[VAL_36:.*]] = torch.prim.ListConstruct %[[VAL_34]], %[[VAL_35]] : (!torch.int, !torch.int) -> !torch.list<int>
-  // CHECK:           %[[VAL_37:.*]] = torch.constant.none
-  // CHECK:           %[[VAL_38:.*]] = torch.aten.zeros %[[VAL_36]], %[[VAL_37]], %[[VAL_37]], %[[VAL_37]], %[[VAL_37]] : !torch.list<int>, !torch.none, !torch.none, !torch.none, !torch.none -> !torch.vtensor<[1,2],si64>
-  // CHECK:           %[[VAL_39:.*]] = torch.prim.ListConstruct %[[VAL_38]], %[[VAL_33]] : (!torch.vtensor<[1,2],si64>, !torch.vtensor<[1,1],si64>) -> !torch.list<vtensor>
-  // CHECK:           %[[VAL_40:.*]] = torch.aten.cat %[[VAL_39]], %[[VAL_25]] : !torch.list<vtensor>, !torch.int -> !torch.vtensor<[1,3],si64>
-  // CHECK:           return %[[VAL_40]] : !torch.vtensor<[1,3],si64>
+// CHECK:           %[[VAL_5:.*]] = torch.constant.int 0
```
zjgarvey, line 2037: These lit tests are enormous. I'm not sure we want this to be so specific to the current implementation. 

It may be better to check for a few things like the loop bodies, extracts, inserts, etc.

zjgarvey (changes requested):
I'm mostly having difficulties parsing the `nmsLoop`. Could you give some details as to what the implementation there is trying to do?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -3810,87 +3801,218 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+            loc, numClasses.getType(), numBatches, numClasses);
+        numResults = rewriter.create<Torch::AtenMulIntOp>(
+            loc, numClasses.getType(), numResults, maxOutputBoxesPerClass);
+        auto finalResultType = resultType;
+        */
+
+        if (!scoreTensorType.toBuiltinTensor().hasStaticShape()) {
+          llvm_unreachable("Unimplemented: Encountered dynamic shaped tensors "
+                           "while lowering Onnx NonMaxSuppression op to torch");
+        }
```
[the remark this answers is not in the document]
zjgarvey: Or don't include the check at all if it isn't torch-mlir's fault that the IR won't compile in IREE.

test/Conversion/TorchOnnxToTorch/simple_ops_g_to_p.mlir
praveen-g-ctt: Updated the test to check for relevant ops

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -3701,63 +3701,58 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
               binder.op, "expected center_point_box attribute to be 0 or 1");
 
-        // TODO: Support multiple batches and classes
-        // Squeeze the boxes and scores tensor.
-        // In Onnx, the shape of boxes is [BxNx4] while the
-        // torchvision expects it to be of shape [Nx4]. Similarly, for
-        // the scores tensor shape in Onnx is [BxCxN] while the
-        // torchvision expects it to be of shape [N].
+        Value cst0 = rewriter.create<Torch::ConstantIntOp>(
+            loc, rewriter.getI64IntegerAttr(0));
```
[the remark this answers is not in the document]
praveen-g-ctt: Updated, Thanks

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -3810,87 +3801,218 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
+                &nmsClassLoop.getRegion(), nmsClassLoop.getRegion().begin(),
+                TypeRange({intTy, finalResultType, intTy, intTy}),
+                {loc, loc, loc, loc});
+
+            auto classIV = classLoopBody->getArgument(0);
+            auto currRes = classLoopBody->getArgument(1);
+            auto finalResIdx = classLoopBody->getArgument(2);
+            Value numResultValues = classLoopBody->getArgument(3);
+
+            auto scoreSelect = rewriter.create<Torch::AtenSelectIntOp>(
```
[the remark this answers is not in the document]
praveen-g-ctt: Updated, Thanks

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
```diff
@@ -3810,87 +3801,218 @@ void mlir::torch::onnx_c::populateDefaultDomainGtoP(
...
-        Value tensorList = rewriter.create<Torch::PrimListConstructOp>(
-            loc, listType, SmallVector<Value>{zeros, result});
-        rewriter.replaceOpWithNewOp<Torch::AtenCatOp>(binder.op, resultType,
-                                                      tensorList, cst1);
+          auto batchIV = batchLoopBody->getArgument(0);
+          auto currRes = batchLoopBody->getArgument(1);
+          auto finalResIdx = batchLoopBody->getArgument(2);
+          auto numResultValues = batchLoopBody->getArgument(3);
+
+          auto boxValue = rewriter.create<Torch::AtenSelectIntOp>(
```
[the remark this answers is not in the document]
praveen-g-ctt: @jinchen62 I had used AtenSelectIntOp as the selected dim is to be removed for all the usages and it gets decomposed in the subsequent DecomposeComplexOps pass. Please let me know if it would be better to use add + slice_tensor + squeeze in this change itself ?

lib/Conversion/TorchOnnxToTorch/DefaultDomainGtoP.cpp
praveen-g-ctt: @zjgarvey @jinchen62 Updated the comments for the nmsLoop part. This loop is used to insert the triplet [batch_index, class_index, selected_box_index] at the required indices element by element. 

" Is the purpose of this loop to insert these elements into the final result? Is it possible to avoid using a loop for this and instead concatenate the nms result with some splat tensors, then insert that into the final result by keeping track of what the cumulative num_selected is?"

-> Yes, I had already tried the approach with splat + concats as part of https://github.com/llvm/torch-mlir/pull/3981
I was running into runtime issues like segfault / invalid mem access due to handling of dynamic dims in IREE. 

The IR using concat + splat method is [here](https://gist.github.com/praveen-g-ctt/68511af88d6b7350ae04432a02dc71d5)


I made use of loops so that we can have a working solution initially and then update the logic once issues in IREE are fixed. Please let me know your thoughts on this!

PR #4002, opened 2025-02-05
[ONNX][TORCH] Add Onnx->Linalg lowering for RotaryEmbedding Op

vivekkhandelwal1:
This commit adds the Onnx->Linalg lowering for Onnx's RotaryEmbedding op (ref: https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#commicrosoftrotaryembedding) by registering a customized torch op named `OnnxVariantAtenRotaryEmbeddingOp`. This is done so that the Onnx's RotaryEmbedding op can be lowered to this op and this op can be lowered from Torch->Linalg.

The lowering has been adopted from the OnnxRuntime. Files for references:
1.) https://github.com/microsoft/onnxruntime/blob/e1e3f623f61816008e79dddc91a51ffe7f0ff5cf/onnxruntime/contrib_ops/cpu/bert/rotary_embedding.cc#L47-L93
2.) https://github.com/microsoft/onnxruntime/blob/94c69f55d480cb4a8dcbc161d29ef3acca9392a7/onnxruntime/contrib_ops/cpu/bert/rotary_embedding_helper.h


AmosLewis:
We need a test in https://github.com/nod-ai/SHARK-TestSuite/tree/main/alt_e2eshark/onnx_tests/operators to verify the numeric before merge

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -3628,6 +3628,334 @@ class ConvertSymConstrainRangeOp
...
+private:
+  struct RotaryParameters {
+    int64_t batchSize;
+    int64_t sequenceLength;
+    int64_t hiddenSize;
+    int64_t headSize;
+    int64_t rotaryEmbeddingDim;
+    int64_t numHeads;
+    int64_t maxSequenceLength;
+    int64_t positionIdsFormat;
```
AmosLewis, line 3646: in onnx, the int position_ids_format;   // Format of position ids - 0 is (1), 1 is (batch_size, sequence_length)
 But in current implementation, we only assume  position_ids : (batch_size, sequence_length), the position_ids_format  is not used, so we can delete it.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -3628,6 +3628,334 @@ class ConvertSymConstrainRangeOp
...
+    //    cos_cache    : (max_sequence_length, head_size / 2) or
+    //                   (max_sequence_length, rotary_embedding_dim / 2)
+    //    sin_cache    : (max_sequence_length, head_size / 2) or
+    //                   (max_sequence_length, rotary_embedding_dim / 2)
+
+    // For the `RotaryEmbedding` lowering to work, shapes of all the inputs are
+    // required to be statically known.
+
+    // TODO: Add a check for inputs to be floating-point
+
```
AmosLewis, line 3666: // TODO: Add 3d input support: (batch_size, sequence_length, hidden_size)

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -3628,6 +3628,334 @@ class ConvertSymConstrainRangeOp
...
+                  // orig_input = input[b][n][s][h]
+                  // rotated_input = input[b][n][s][j]
+                  // position_id = position_ids[b][s]
+                  // cos_emb = cos_cache[position_id][cache_idx]
+                  // sin_emb = sin_cache[position_id][cache_idx]
+                  // out[b][n][s][h] = orig_input * cos_emb
+                  //                             +
+                  //                  (rotated_input * sin_emb) * sign
+
+                  Value indexA = b.create<linalg::IndexOp>(loc, 0);
```
AmosLewis, line 3896: instead of ABCD, use "b, n, s, h" might be more intuitive

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -3628,6 +3628,334 @@ class ConvertSymConstrainRangeOp
...
+
+    parameters.batchSize = batchSize;
+    parameters.sequenceLength = sequenceLength;
+    parameters.hiddenSize = hiddenSize;
+    parameters.headSize = headSize;
+    parameters.numHeads = numHeads;
+    parameters.maxSequenceLength = maxSequenceLength;
+    parameters.rotaryEmbeddingDim =
+        rotaryEmbeddingDim > 0 ? rotaryEmbeddingDim : headSize;
+
```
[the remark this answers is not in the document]
vivekkhandelwal1: Yeah, we don't have them since we are not flattening the input tensors. The original implementation needs that because they use 1-d pointers to access the tensor.

vivekkhandelwal1:
> We need a test in https://github.com/nod-ai/SHARK-TestSuite/tree/main/alt_e2eshark/onnx_tests/operators to verify the numeric before merge

Actually, the test in SHARK-Testsuite is not working since the op comes from "com.microsoft" domain. Alhtough, I have verified the e2e correctness of lowering by manually generating the IR and then compiling and executing it with the IREE.

lib/Conversion/TorchToLinalg/Uncategorized.cpp
```diff
@@ -3628,6 +3628,332 @@ class ConvertSymConstrainRangeOp
...
+    RankedTensorType resultType =
+        cast<RankedTensorType>(typeConverter->convertType(op.getType()));
+    RankedTensorType inputType = cast<RankedTensorType>(input.getType());
+
+    if (inputType.getElementType() != resultType.getElementType())
+      return rewriter.notifyMatchFailure(
+          op, "Expected input and result to have the same element type");
+
+    int64_t interleaved, isPackedBatching, numHeads, rotaryEmbeddingDim;
+    if (!matchPattern(op.getInterleaved(), m_TorchConstantInt(&interleaved)))
```
zjgarvey, line 3804: Interleaved is treated like a bool below, is there a reason it is an int, or is that just the way that ONNX stores it?

test/Conversion/TorchToLinalg/basic.mlir
```diff
@@ -354,3 +354,78 @@ func.func @torch.aten.transpose$basic(%arg0: !torch.vtensor<[4,3],f32>) -> !torc
   %0 = torch.aten.transpose.int %arg0, %int0, %int1 : !torch.vtensor<[4,3],f32>, !torch.int, !torch.int -> !torch.vtensor<[3,4],f32>
   return %0 : !torch.vtensor<[3,4],f32>
 }
+
```
zjgarvey, line 357: I'm not sure about adding this lit test to `basic.mlir` when it's not really... basic.

This is also quite involved for a lit test, and there isn't a test for the interleaved path.

Perhaps converting the literal tensors to inputs would cut down on the size of this lit test?

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
```diff
@@ -1410,4 +1410,29 @@ def Torch_BindSymbolicShapeOp : Torch_Op<"bind_symbolic_shape", []> {
...
+    Torch_IntType:$interleaved,
+    Torch_IntType:$is_packed_batching,
+    Torch_IntType:$num_heads,
+    Torch_IntType:$rotary_embedding_dim,
+    Torch_FloatType:$scale
+  );
+  let results = (outs
+    AnyTorchTensorType:$result
+  );
+  let assemblyFormat = "$input `,` $position_ids `,` $cos_cache `,` $sin_cache `,` $interleaved `,` $is_packed_batching `,` $num_heads `,` $rotary_embedding_dim `,` $scale attr-dict `:` type($input) `,` type($position_ids) `,` type($cos_cache) `,` type($sin_cache)  `,` type($interleaved)  `,` type($is_packed_batching)  `,` type($num_heads)  `,` type($rotary_embedding_dim)  `,` type($scale) `->` type($result)";
```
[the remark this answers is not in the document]
vivekkhandelwal1: It works, but I didn't think of using it. Is there any advantage of using it?

lib/Conversion/TorchToLinalg/Uncategorized.cpp
vivekkhandelwal1: Yeah, I have kept it this way to match the semantics of Onnx op.

test/Conversion/TorchToLinalg/basic.mlir
vivekkhandelwal1: I have made the changes. I kept the constants since I used them for e2e testing. I didn't find any other file appropriate for adding this test so added it here.

lib/Conversion/TorchOnnxToTorch/DefaultDomainQtoZ.cpp
```diff
@@ -4555,4 +4555,48 @@ void mlir::torch::onnx_c::populateDefaultDomainQtoZ(
         rewriter.replaceOp(binder.op, primLoop);
         return success();
       });
+  patterns.onOp(
+      "RotaryEmbedding", 1,
```
zjgarvey, line 4559: Let's make a new source file for non-default domain ops. E.g., `ComMicrosoftDomain.cpp`

include/torch-mlir/Dialect/Torch/IR/TorchOps.td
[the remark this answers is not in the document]
zjgarvey: It's shorter.

zjgarvey:
Give me another day to consider this more. This is the first time we've added a custom torch op to try and support an onnx op, so I'd like to make sure we set a good precedent for how to approach this in the future. 

Do you know if the output shape of this op is guaranteed to be the same as `operand[0]`?

vivekkhandelwal1:
> Give me another day to consider this more. This is the first time we've added a custom torch op to try and support an onnx op, so I'd like to make sure we set a good precedent for how to approach this in the future.
> 
> Do you know if the output shape of this op is guaranteed to be the same as `operand[0]`?

Yeah, the output shape is same as the input operand. It's mention here: https://github.com/microsoft/onnxruntime/blob/main/docs/ContribOperators.md#outputs-80

zjgarvey:
I submitted a PR after playing around with the shape and dtype inference for unregistered ops. Take a look at <https://github.com/vivekkhandelwal1/torch-mlir/pull/4> and let me know what you think.

vivekkhandelwal1:
Hi @zjgarvey, that PR of yours is merged. Can you please do a final review now?

zjgarvey (approved):
Looks good to me. 

Some nits if you think it's worth doing now: some lit tests could be trimmed a bit (e.g. torch-onnx-to-torch is really just checking it lowers to the torch op, so it doesn't need to be so involved with explicit constants) and some verification could be moved to a verifier.

vivekkhandelwal1:
> Looks good to me.
> 
> Some nits if you think it's worth doing now: some lit tests could be trimmed a bit (e.g. torch-onnx-to-torch is really just checking it lowers to the torch op, so it doesn't need to be so involved with explicit constants) and some verification could be moved to a verifier.

I have simplified the lit test. Moving the changes to verifier can be done in a follow-up PR.

PR #4003, opened 2025-02-06
Missing Shape Inference for Prod

keshavj-cerebras:
Added required but missing Shape inference for `aten.prod`


prateekgu-cerebras (approved):
LGTM!

PR #4010, opened 2025-02-07
Generalize the PyTorch avg_pool linalg lowering algorithm for the case where count_include_pad = false.

ivangarcia44:
Currently the avg_pool2d PyTorch operation supports the cases where count_include_pad is true and false, but the avg_pool1d and avg_pool3d only the true case is supported (which is simpler).

The count_include_pad = false support for avg_pool2d was added by @AmosLewis in this change (reviewed by @rsuderman and @nirvedhmeshram) : https://github.com/llvm/torch-mlir/pull/3235

In this change I generalized the logic added above. I also did some refactoring to the original code to reduce the size of the functions and to avoid redundancy when possible.

@sahas3 @dixinzhou @rafaelubal


lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -829,6 +830,251 @@ class ConvertAtenMaxUnpool3dOp final
...
+  Value ODimDDim;
+  Value IDim0;
+  Value IDim;
+  Value IDim0KDim;
+  Value IDimPadDim;
+  Value IDim1;
+  Value IDim1IDims0;
+  Value IDim0Clamped;
+  Value IDim1Clamped;
+  Value IDim1_IDim0;
```
sahas3, line 875: Can you add some comments to explain what these variables are? For example, what's the difference between the similarly named variables `IDim1IDims0` and `IDim1_IDim0`.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -829,6 +830,251 @@ class ConvertAtenMaxUnpool3dOp final
...
+}
+} // namespace
+
+namespace {
+// This structure, used solely in PoolSizeCalculator, provides
+// the intermediate values for each dimension to compute the
+// divisor of the average pooling operator.
+struct PoolSizeValues {
+  int64_t SpatialDimsInt64;
+  int64_t DimSpatialInt;
```
sahas3, line 860: It'll be good to be consistent with the naming, one has `Int64` but the other has `Int`, on the other hand none of the other names has the type in them.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -829,6 +830,251 @@ class ConvertAtenMaxUnpool3dOp final
...
+}
+
+template <int NumOfDims>
+Value PoolSizeCalculator<NumOfDims>::getPoolSize(
+    OpBuilder &b, SmallVectorImpl<Value> &kernelSizeIntValues,
+    SmallVectorImpl<int64_t> &strideInts,
+    SmallVectorImpl<int64_t> &paddingInts) {
+  Value poolSize;
+
+  Value cstZero = rewriterHandle.create<arith::ConstantOp>(
```
sahas3, line 933: I think you can use `Opbuilder b` here instead of `rewriterHandle` -- no need to store it as a class property then.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -829,6 +830,251 @@ class ConvertAtenMaxUnpool3dOp final
...
+      location, rewriterHandle.getI64IntegerAttr(0));
+
+  for (int i = 0; i < NumOfDims; ++i) {
+    dims[i].IndexODim =
```
sahas3, line 937: Is there a need to store all these values for each iteration?

projects/pt1/python/torch_mlir_e2e_test/test_suite/pooling.py
```diff
@@ -1460,6 +1460,66 @@ def AvgPool3dStaticModule_basic(module, tu: TestUtils):
     module.forward(tu.rand(2, 2, 4, 4, 4, low=-1))
 
 
+class AvgPool3dCountIncludePadFalse(torch.nn.Module):
+
+    def __init__(self):
+        super().__init__()
+        self.ap2d = torch.nn.AvgPool3d(
```
sahas3, line 1467: ap3d is better suited than ap2d

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -829,6 +830,251 @@ class ConvertAtenMaxUnpool3dOp final
 };
 } // namespace
 
+namespace {
+// The following structures and the adsfdasf method
```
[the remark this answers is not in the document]
ivangarcia44: Yes, I usually do this as a place holder to copy paste the ID name in the comment. In this case it was getNumOfDims. Thanks!

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -829,6 +830,251 @@ class ConvertAtenMaxUnpool3dOp final
...
+namespace {
+// This is a helper class to create the pooling size value
+// used in the divisor of the average pooling operator.
+template <int NumOfDims> class PoolSizeCalculator {
+public:
+  PoolSizeCalculator(Value self, Value sumPool,
+                     ConversionPatternRewriter &rewriter, Location loc);
+
+  // The algorithm for computing the divisor with
+  // count_include_pad is manily based on pytorch
```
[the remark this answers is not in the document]
ivangarcia44: Corrected, thanks

lib/Conversion/TorchToLinalg/Pooling.cpp
ivangarcia44: These are intermediate variables for the average pooling divisor computation. Originally the code use height and width instead of Dim. For example IDim1IDims0 was named something like Ih1IH0 and Iw1Iw0. The original code did not had comments for the names but they point to the algorithm where this is based from in the PyTorch average pooling algorithm here (which do not have comments on the variables as well): https://github.com/pytorch/pytorch/blob/4a6dfbe4806b361c43210dfd56db64c4097c66bb/aten/src/ATen/native/cpu/AvgPoolKernel.cpp#L78

I modified the comments above all these variables to clarify to clarify this a little. 

// This structure, used solely in PoolSizeCalculator, provides
// the intermediate values for each dimension to compute the
// divisor of the average pooling operator.
// See the link below for the PyTorch implementation where
// this is derived from:
// https://github.com/pytorch/pytorch/blob/4a6dfbe4806b361c43210dfd56db64c4097c66bb/aten/src/ATen/native/cpu/AvgPoolKernel.cpp#L78
// Dim below stands for spatial dimension. Prior to the February
// 2025 change, these variables used "height" and "width" (or
// "h" and "w") in these intermediate variables instead of "Dim".

I think pad, clamping and index are obvious. I don't know what the "I" stands for, it could be input. The numbers I suppose are arbitrary to have different values on the temporary variables. 

For the difference between IDim1IDims0 and IDim1_IDim0, I renamed the later to be IDim1_IDim0_Clamped. The former uses IDim1 and IDim0, and the later uses the clamped versions of the. Adding Clamped on them should help.

lib/Conversion/TorchToLinalg/Pooling.cpp
ivangarcia44: Originally these were named wDim/hDim and dimH/dimW respectively. I should not have add the type in them. Removed the flipping of labels to differentiate them. I don't think this help. Instead I renamed them to indicate where these dimension sizes come from:

  int64_t DimSizeFromSelfType;
  int64_t DimSizeFromSumPoolType;
  
  You can see in the code snipped below where they are calculated from:
  
    const int64_t selfRank = selfType.getRank();
  RankedTensorType sumPoolType = cast<RankedTensorType>(sumPool.getType());
  const int64_t rank = sumPoolType.getRank();

  // Store dimensions in this order:
  // 0 => width, 1 => height, 2 => depth
  for (int i = 0; i < NumOfDims; ++i) {
    dims[i].DimSizeFromSelfType = toPositiveDim(-(i + 1), selfRank);
    dims[i].InputSpatialDimValues =
        getDimOp(rewriterHandle, location, self, dims[i].DimSizeFromSelfType);
    dims[i].DimSizeFromSumPoolType = toPositiveDim(-(i + 1), rank);
  }

lib/Conversion/TorchToLinalg/Pooling.cpp
ivangarcia44: Makes sense; I was able to localize most of these variables.

projects/pt1/python/torch_mlir_e2e_test/test_suite/pooling.py
ivangarcia44: Corrected typo here and in the other AvgPool3D case as well. The AvgPool1D tests were fine. Thanks

lib/Conversion/TorchToLinalg/Pooling.cpp
ivangarcia44: Did this correction, thanks.

projects/pt1/e2e_testing/xfail_sets.py
```diff
@@ -3386,6 +3390,8 @@
     "AtenKthvalueKeepDimModule_basic",
     "AtenKthvalueModule_basic",
     "AvgPool3dStaticModule_basic",
+    "AvgPool3dCountIncludePadFalse_basic",
+    "AvgPool3dCountIncludePadFalseWithoutPadding_basic",
```
sahas3, line 3394: Do the 1d tests not fail for `fx_importer_tosa` ? I thought the `count_pad = false` is only supported for `linalg_on_tensors` pipeline.

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -829,6 +830,224 @@ class ConvertAtenMaxUnpool3dOp final
...
+    SmallVector<AffineMap> &indexingMapsAvg,
+    SmallVector<utils::IteratorType> &iteratorTypesAvg) {
+  Location loc = op->getLoc();
+
+  Type resultElementType = cast<RankedTensorType>(resultType).getElementType();
+
+  Value divisor = kernelSizeIntValues[0];
+  for (uint32_t i = 1; i < kernelSizeIntValues.size(); ++i) {
+    divisor =
+        rewriter.create<arith::MulIOp>(loc, divisor, kernelSizeIntValues[i]);
```
sahas3, line 1023: `createOrFold` instead of `create` will be able to fold the op for constant kernels

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -829,6 +830,224 @@ class ConvertAtenMaxUnpool3dOp final
...
+  }
+  if constexpr (!std::is_same<OpTy, AtenAvgPool1dOp>()) {
+    divisor = isa<Torch::NoneType>(op.getDivisorOverride().getType())
```
sahas3, line 1026: will be good to add a comment that 2d/3d versions have optional divisor_override attribute but 1d does not

ivangarcia44: Added this comment:
  // Only average pooling 2D/3D have optional divisor override.

projects/pt1/e2e_testing/xfail_sets.py
ivangarcia44: The 1D tests do not fail with these commands:
python -m e2e_testing.main --config=fx_importer
python -m e2e_testing.main --config=fx_importer_tosa
bash build_tools/ci/test_posix.sh nightly

However, if the 3D lines above are removed, they will fail with this command:

python -m e2e_testing.main --config=fx_importer

I removed the new entries from the TOSA_PASS_SET. Everything passes without them. I added them originally while experiment what was causing the tests failure.

ivangarcia44:
Hi @rsuderman @AmosLewis , would it be ok to merge this change?

ivangarcia44:
Closing merge request. Will create a new one to avoid the merge issue.

ivangarcia44:
@AmosLewis  @rsuderman @sahas3 @dixinzhou 

Please review the new merge request here: https://github.com/llvm/torch-mlir/pull/4035

PR #4012, opened 2025-02-10
Lowering to linalg for AtenCol2ImOp

gdehame:
Added a lowering to linalg for the torch.aten.col2im operation.
    Added a unit test to verify the lowering.

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -24,6 +24,7 @@
 #include "torch-mlir/Dialect/Torch/IR/TorchOps.h"
 #include "torch-mlir/Dialect/Torch/Utils/TorchUpstream.h"
 #include "torch-mlir/Dialect/Torch/Utils/Utils.h"
+#include "torch-mlir/Dialect/TorchConversion/IR/TorchConversionOps.h"
 #include "llvm/ADT/APInt.h"
 
 #include <numeric>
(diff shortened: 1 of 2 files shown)


sahas3:
Hi @gdehame, thanks for the submission.

I'm yet to review the code -- just leaving one high-level feedback. How does this op appear in the IR through the `fx.export_and_import` API usage? With the source pytorch program, it'll be good to also add an e2e test for numerical verification as well.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -2789,6 +2790,269 @@ SmallVector<StringRef> ConvertSparseOperatorOp::legalizedNames = {
...
+                  ConversionPatternRewriter &rewriter) const override {
+    // Retrieve the hyperparameters
+    Value input = col2imOp.getSelf();
+    if (!(col2imOp.getOutputSize().getDefiningOp() &&
+          isa<Torch::PrimListConstructOp>(
+              col2imOp.getOutputSize().getDefiningOp())))
+      return failure();
+
+    Torch::PrimListConstructOp outputSizes = cast<Torch::PrimListConstructOp>(
+        col2imOp.getOutputSize().getDefiningOp());
```
sahas3:
```suggestion
    auto outputSizes = col2imOp.getOutputSize().getDefiningOp<Torch::PrimListConstructOp>();
    if (!outputSizes)
       return failure();
```

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -2789,6 +2790,269 @@ SmallVector<StringRef> ConvertSparseOperatorOp::legalizedNames = {
...
+    int height =
+        cast<Torch::ConstantIntOp>(outputSizes->getOperand(0).getDefiningOp())
+            .getValue();
+    int width =
+        cast<Torch::ConstantIntOp>(outputSizes->getOperand(1).getDefiningOp())
+            .getValue();
+    if (!(col2imOp.getPadding().getDefiningOp() &&
+          isa<Torch::PrimListConstructOp>(
+              col2imOp.getPadding().getDefiningOp())))
+      return failure();
```
sahas3: Here also similar suggestion about grabbing the def op at once and verifying it exists can be implemented as mentioned above.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -2789,6 +2790,269 @@ SmallVector<StringRef> ConvertSparseOperatorOp::legalizedNames = {
...
+
+    Torch::PrimListConstructOp paddings =
+        cast<Torch::PrimListConstructOp>(col2imOp.getPadding().getDefiningOp());
+
+    if (!(paddings.getNumOperands() == 2 &&
```
sahas3: Similar comment here and elsewhere too.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -2789,6 +2790,269 @@ SmallVector<StringRef> ConvertSparseOperatorOp::legalizedNames = {
...
+        cast<Torch::ConstantIntOp>(strides->getOperand(1).getDefiningOp())
+            .getValue();
+
+    // Create intermediate buffers
+    TensorType outputType =
+        cast<Torch::ValueTensorType>(col2imOp.getType()).toBuiltinTensor();
+    Type elementType = outputType.getElementType();
+    Value outputBuffer = tensor::EmptyOp::create(
+        rewriter, col2imOp->getLoc(),
+        ArrayRef<int64_t>{outputType.getDimSize(0), outputType.getDimSize(1),
```
sahas3, line 2926: I think assumption here is that `outputType` dims 0 and 1 are static but that is not explicitly checked anywhere.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -2789,6 +2790,269 @@ SmallVector<StringRef> ConvertSparseOperatorOp::legalizedNames = {
...
+    // The body of the linalg.generic op
+    auto body = [&](OpBuilder &b, Location loc, ValueRange args) {
+      Value acc =
+          (elementType.isInteger())
+              ? arith::AddIOp::create(b, loc, args[0], args[3]).getResult()
+              : (isa<mlir::FloatType>(elementType)
+                     ? arith::AddFOp::create(b, loc, args[0], args[3])
+                           .getResult()
+                     : complex::AddOp::create(b, loc, args[0], args[3])
+                           .getResult());
```
sahas3: These nested tertiary ops are hard to read (here and in other places). Can you break this down?

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -2789,6 +2790,269 @@ SmallVector<StringRef> ConvertSparseOperatorOp::legalizedNames = {
...
+                    verticalStride,
+            1 + ((paddedWidth - 1 - (kernelWidth - 1) * horizontalDilation)) /
+                    horizontalStride},
+        elementType);
+    assert(((isa<ComplexType>(elementType) &&
+             (cast<ComplexType>(elementType).getElementType().isInteger() ||
+              isa<mlir::FloatType>(
+                  cast<ComplexType>(elementType).getElementType()))) ||
+            isa<mlir::FloatType>(elementType) || elementType.isInteger()) &&
+           "Not implemented yet\n");
```
sahas3: This can be checked before starting to create any new ops in the IR and `failure()` can be returned for this unsupported case instead of an assert.

lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -2789,6 +2790,269 @@ SmallVector<StringRef> ConvertSparseOperatorOp::legalizedNames = {
...
+
+    TypedAttr init0 =
+        elementType.isInteger()
```
sahas3: integer/complex types are not being tested. I'd suggest adding e2e tests for testing those combinations

lib/Conversion/TorchToLinalg/DataMovement.cpp
gdehame: Done, thanks!

lib/Conversion/TorchToLinalg/DataMovement.cpp
gdehame: Done, thanks!

lib/Conversion/TorchToLinalg/DataMovement.cpp
gdehame: Done, thanks!

lib/Conversion/TorchToLinalg/DataMovement.cpp
gdehame: I added a check, thanks!

lib/Conversion/TorchToLinalg/DataMovement.cpp
gdehame: Done, thanks!

lib/Conversion/TorchToLinalg/DataMovement.cpp
gdehame: Done, thanks!

lib/Conversion/TorchToLinalg/DataMovement.cpp
gdehame: I started doing so and realized the integer case is bugged due to using `arith.constant` to generate a 0 constant of possibly (un)signed int which is not allowed.
I'll work on this further next week-end

gdehame:
> Hi @gdehame, thanks for the submission.
> 
> I'm yet to review the code -- just leaving one high-level feedback. How does this op appear in the IR through the `fx.export_and_import` API usage? With the source pytorch program, it'll be good to also add an e2e test for numerical verification as well.

This came up during an internship I was doing last year, I completely forgot about the exact setup and example. I'll try to recreate one during the week-end

lib/Conversion/TorchToLinalg/DataMovement.cpp
gdehame: I updated the PR with additional tests and fixed the lowering

gdehame:
> > Hi @gdehame, thanks for the submission.
> > I'm yet to review the code -- just leaving one high-level feedback. How does this op appear in the IR through the `fx.export_and_import` API usage? With the source pytorch program, it'll be good to also add an e2e test for numerical verification as well.
> 
> This came up during an internship I was doing last year, I completely forgot about the exact setup and example. I'll try to recreate one during the week-end

Here is an example:
```
from torch_mlir import fx
from torch_mlir.compiler_utils import OutputType
import torch
import numpy as np

fold = torch.nn.Fold((14, 30), (2, 2), (1, 1), (1, 1), (2, 2))

print(
    fx.export_and_import(
        torch.export.export(fold, (torch.tensor(np.zeros((1, 12, 128))),)),
        output_type=OutputType.LINALG_ON_TENSORS,
    )
)
```

PR #4015, opened 2025-02-10
refactor(ONNX): standardizes newest TorchOnnxToTorch helpers

bjacobgordon:
- extracts out two more helpers with potentially broad usage
- standardizes their function signatures for future use with ImplicitLocOpBuilder


bjacobgordon:
Hey, @zjgarvey, this is a small follow-up to #3987 where I had prematurely introduced these changes.

bjacobgordon:
@zjgarvey, thanks for looknig at this!

Skipping this as discussed:
- Helper functions would be better off in more centralized location
- Would have to document how to find and leverage them
- Would need to be higher up the priority list

PR #4016, opened 2025-02-11
[TOSA] TOSA updates for LLVM hash a854c26

justin-ngo-arm:
PR summary:
- [TOSA] Update tosa.mul's `shift` as input
- [TOSA] Update tosa.slice's start and size to !tosa.shape type
- [TOSA] Refactor: Use tosaCastTensorToType() function to create tosa.cast
- [TOSA] Update tosa.reshape's `shape` as a !tosa.shape input


vivekkhandelwal1:
Hi @justin-ngo-arm, I have just one small request for future reference. It would be slightly easier for me (or for others) if you could have one single commit comprising fixes corresponding to an LLVM hash.

PR #4020, opened 2025-02-12
feat(fx-importer): support for importing fp8 model parameters

UtkarshKunwar:
Essentially follows the same way how the importing of `bfloat16` is being handled. Related to #2896.


zjgarvey (approved):
Sorry it took a while to get around to this. It looks good to me.

I'd like to have some tests to make sure these dtypes import properly and don't regress with version bumps, but we might need to add the `ml_dtypes` dependency to a workflow file somewhere. If you think it is a good idea to spend some time adding tests, I think it would be best to add them before merging.

PR #4021, opened 2025-02-12
Add a CMake option to enable TOSA. Default to ON.

### LINKED ISSUE #4019: Make building TOSA support optional
Similar components in torch-mlir are already optional: https://github.com/llvm/torch-mlir/blob/a9a1355c98caddf30b220755b558f01fa1e5ee05/CMakeLists.txt#L37-L45

This would allow downstream projects that only depend on certain parts of torch-mlir to not pay the costs (binary size, build time, compiler support, etc.) for more than they use.

Making the component optional would involve plumbing across quite a few files: https://github.com/search?q=repo%3Allvm%2Ftorch-mlir%20TORCH_MLIR_ENABLE_STABLEHLO&type=code. Some refactoring could make that easier. For example, see how IREE has a plugin system: https://github.com/iree-org/iree/tree/main/compiler/plugins/input and top level options that choose which plugins are included: https://github.com/iree-org/iree/blob/99304ff016d0ee0f742b620fef59ae1cd9706899/CMakeLists.txt#L412-L431.

bjacob:
Fixes #4019.


bjacob:
FYI @justin-ngo-arm (somehow, I can't add you as reviewer).

justin-ngo-arm:
@bjacob This looks reasonable to me. You can add @sjarus as a reviewer as he has merge access to this repo.

sjarus (approved):
Looks reasonable to me with default ON.

lib/CMakeLists.txt
```diff
@@ -33,7 +32,11 @@ set(LinkedLibs
 )
 
 if(TORCH_MLIR_ENABLE_STABLEHLO)
-list(APPEND LinkedLibs StablehloLinalgTransforms StablehloPasses)
+  list(APPEND LinkedLibs StablehloLinalgTransforms StablehloPasses)
+endif()
+
+if (TORCH_MLIR_ENABLE_TOSA)
```
ScottTodd, line 38: nit: CMake style (fine to ignore if this is the only change)
```suggestion
if(TORCH_MLIR_ENABLE_TOSA)
```

lib/Dialect/TorchConversion/Transforms/Passes.cpp
```diff
@@ -18,7 +18,9 @@
 #include "torch-mlir/Conversion/TorchToSCF/TorchToSCF.h"
 #include "torch-mlir/Conversion/TorchToTMTensor/TorchToTMTensor.h"
 #include "torch-mlir/Conversion/TorchToTensor/TorchToTensor.h"
+#ifdef TORCH_MLIR_ENABLE_TOSA
```
ScottTodd, line 21: How about moving both `#ifdef`s to a separate include block? Right now they are in the middle of other includes, I guess to keep them alphabetical?

```c++
#include "torch-mlir/Conversion/TorchToTMTensor/TorchToTMTensor.h"
#include "torch-mlir/Conversion/TorchToTensor/TorchToTensor.h"
#include "torch-mlir/Dialect/Torch/Transforms/Passes.h"

#ifdef TORCH_MLIR_ENABLE_TOSA
#endif

#ifdef TORCH_MLIR_ENABLE_STABLEHLO
#include "stablehlo/transforms/Passes.h"
#include "torch-mlir/Conversion/TorchToStablehlo/TorchToStablehlo.h"
#endif
```

lib/InitAll.cpp
```diff
@@ -19,7 +19,9 @@
 #include "mlir/Dialect/SparseTensor/IR/SparseTensor.h"
 #include "mlir/Dialect/Tensor/IR/Tensor.h"
 #include "mlir/Dialect/Tensor/IR/TensorInferTypeOpInterfaceImpl.h"
+#ifdef TORCH_MLIR_ENABLE_TOSA
 #include "mlir/Dialect/Tosa/IR/TosaOps.h"
+#endif
```
ScottTodd, line 24: Similar comment here, could move this down next to
```c++
#ifdef TORCH_MLIR_ENABLE_STABLEHLO
#include "stablehlo/conversions/linalg/transforms/Passes.h"
#include "stablehlo/transforms/Passes.h"
#endif
```

ScottTodd (approved):

lib/Dialect/TorchConversion/Transforms/Passes.cpp
bjacob: I was being paranoid. A bunch of `#include` lines in C++ is like a house of cards.  But since you ask for it...

ScottTodd: If the ordering is load bearing, that's worth documenting too. I share your paranoia :D

PR #4024, opened 2025-02-13
Add -DPython_FIND_VIRTUALENV=ONLY to cmake configure instructions

giacs-epic:
Required because `externals/llvm-project/mlir/cmake/modules/MLIRDetectPythonEnv.cmake` also looks for `Python` other than `Python3`.


marbre (approved):
Seems there is now also some code upstream to warn about this
https://github.com/llvm/llvm-project/blob/f427890a1dcd7759240568f189df8fbbcf23669c/mlir/cmake/modules/MLIRDetectPythonEnv.cmake#L30-L38

I guess there could be other solutions that set the one depending on the other. This rather adds more band aid but this is probably better than leaving as is.

marbre:
Can you resolve the conflicts @giacs-epic?

marbre:
I just realized that #4004 seems more complete that this, thus rather going to land #4004. However, thanks for sending a PR @giacs-epic!

PR #4025, opened 2025-02-13
[onnx_importer] Fix constant bool tensor importing

giacs-epic:
A constant tensor of bools is imported as something like
```
%0 = torch.operator "onnx.Constant"() {torch.onnx.value = dense<[105, 108, 44]> : tensor<3xui8>} : () -> !torch.vtensor<[22],i1>
```
which leads to incompatible types errors
```
b.mlir:4:10: error: 'torch.vtensor.literal' op inferred type(s) '!torch.vtensor<[3],ui8>' are incompatible with return type(s) of operation '!torch.vtensor<[22],i1>'
    %0 = torch.operator "onnx.Constant"() {torch.onnx.value = dense<[105, 108, 44]> : tensor<3xui8>} : () -> !torch.vtensor<[22],i1> 
```

This PR fixes the translation by removing the `numpy.packbits()` processing.

Note about e2eshark tests: some tests (e.g. `test_sequence_map_identity_1_sequence_expanded`) contain scalar constant boolean tensors. Compilation of the constant part seems to work fine, but this is due to the dimension size of the tensor being 1, which means types are compatible.

### DIFF
--- a/projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
+++ b/projects/pt1/python/torch_mlir_e2e_test/test_suite/basic.py
@@ -2910,6 +2910,30 @@ def NumelZeroRankModule_basic(module, tu: TestUtils):
 # ==============================================================================
 
 
+class BoolTensorConstantModule(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    @export
+    @annotate_args(
+        [
+            None,
+        ]
+    )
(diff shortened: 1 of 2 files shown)


rkayaith:
I believe this should be fixed by https://github.com/llvm/torch-mlir/commit/5693ca7f5fae3bb91e6abde8b17ff75686def898
Let me know if you're still seeing issues!

PR #4029, opened 2025-02-18
Convert float exponents in `aten.pow` to int when possible

zjgarvey:
Pure floating point pow operations no-longer support negative base values (see <https://github.com/llvm/llvm-project/pull/126338>), but many models coming from ONNX use floating point representations of integers as the exponent. 

This change:

1. matches on constant rank-0 exponents and converts them to scalar constants.
2. matches on constant floating-point scalar exponents and converts them to ints if possible. 
3. lowers `Tensor(float)^int` cases to `math.fpowi` 

Addresses some remaining test failures related to <https://github.com/iree-org/iree/issues/19996>.

### DIFF
--- a/lib/Dialect/Torch/IR/TorchOps.cpp
+++ b/lib/Dialect/Torch/IR/TorchOps.cpp
@@ -12,6 +12,7 @@
 #include "llvm/Support/Debug.h"
 
 #include "mlir/Dialect/Func/IR/FuncOps.h"
+#include "mlir/Dialect/Utils/StaticValueUtils.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/BuiltinOps.h"
 #include "mlir/IR/PatternMatch.h"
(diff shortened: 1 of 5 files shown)


lib/Dialect/Torch/IR/TorchOps.cpp
```diff
@@ -2421,6 +2422,64 @@ OpFoldResult AtenReshapeOp::fold(FoldAdaptor adaptor) {
...
+  });
+}
+
+//===----------------------------------------------------------------------===//
+// AtenPowTensorTensorOp
+//===----------------------------------------------------------------------===//
+
+void AtenPowTensorTensorOp::getCanonicalizationPatterns(
+    RewritePatternSet &patterns, MLIRContext *context) {
+  // If the exponent is splat, convert to AtenPowTensorScalar.
```
zjgarvey, line 2456: Self-review: Although it would be a dumb situation, I should check to verify that the splat exponent isn't used for broadcasting.

praveen-g-ctt (approved):
LGTM

PR #4035, opened 2025-02-20
Extend avg pool count include pad false to1d3d

ivangarcia44:
Currently the avg_pool2d PyTorch operation supports the cases where count_include_pad is true and false, but the avg_pool1d and avg_pool3d only the true case is supported (which is simpler).

The count_include_pad = false support for avg_pool2d was added by @AmosLewis in this change (reviewed by @rsuderman and @nirvedhmeshram) : https://github.com/llvm/torch-mlir/pull/3235

In this change I generalized the logic added above. I also did some refactoring to the original code to reduce the size of the functions and to avoid redundancy when possible.

@sahas3 @dixinzhou @rafaelubal


ivangarcia44:
Link to old PR: https://github.com/llvm/torch-mlir/pull/4010

The review of @sahas3 can be found there.

ivangarcia44:
Including additional reviewers familiar with the Pooling.cpp file for review: @vivekkhandelwal1, @lingzhiz1998

In addition to @AmosLewis , @rsuderman, @sahas3, @dixinzhou, @rafaelubal

ivangarcia44:
Including @rafaelubalmw  in addition to the other reviewers: @vivekkhandelwal1, @lingzhiz1998, @AmosLewis , @rsuderman, @sahas3, @dixinzhou, @rafaelubal

sahas3 (approved):
LGTM

amd-vivekag:
Hi Ivan,

Your changes is causing a testcase failure: iree-test-suites/onnx_ops/onnx/node/generated/test_averagepool_2d_ceil/
Can you please fix it?

Thanks Vivek

lib/Conversion/TorchToLinalg/Pooling.cpp
```diff
@@ -829,226 +830,333 @@ class ConvertAtenMaxUnpool3dOp final
...
+          .create<linalg::GenericOp>(
+              loc, outputTensor.getType(), sumPool, outputTensor,
+              /*indexingMaps=*/indexingMapsAvg,
+              /*iteratorTypes=*/iteratorTypesAvg,
+              [&](OpBuilder &b, Location loc, ValueRange args) {
+                auto poolSize = poolSizeCalculator.getPoolSize(
+                    b, kernelSizeIntValues, strideInts, paddingInts);
+                // AtenAvgPool2/3dOp has an optional divisor_override
+                // attribute while AtenAvgPool1dOp does not.
+                if constexpr (avgPoolDims > 1) {
```
ScottTodd, line 1098: FYI this led to some downstream build failures on Windows: https://github.com/llvm/torch-mlir/issues/4085

ivangarcia44:
> Honestly, I think this PR should not have been merged. Apart from having some breaking changes, it has other issues, like incorrect code flow, and it does not adhere to standard LLVM contribution guidelines. For future reference, I would suggest that for such a PR making significant changes, we should wait for the original authors of the code to respond within a definite time frame.

For the next time I will. I apologize for the inconvenience. I waited for the review for almost a month, and I thought it was enough time to merge. I sent reminders various times during that month. Can you please point where is the incorrect code flow and LLVM guidelines violations? The fixes in the latest change cover various numerical correctness issues that were present prior any of my changes that were uncovered by a new set of E2E tests I added.

vivekkhandelwal1:
> > Honestly, I think this PR should not have been merged. Apart from having some breaking changes, it has other issues, like incorrect code flow, and it does not adhere to standard LLVM contribution guidelines. For future reference, I would suggest that for such a PR making significant changes, we should wait for the original authors of the code to respond within a definite time frame.
> 
> For the next time I will. I apologize for the inconvenience. I waited for the review for almost a month, and I thought it was enough time to merge. I sent reminders various times during that month. Can you please point where is the incorrect code flow and LLVM guidelines violations? The fixes in the latest change cover various numerical correctness issues that were present prior any of my changes that were uncovered by a new set of E2E tests I added.

Even if I don't go in much depth then you can see that atleast in this part of the code https://github.com/llvm/torch-mlir/blob/4fe75dd0a94934fc3e05dfc4bd2b49046b2cff0e/lib/Conversion/TorchToLinalg/Pooling.cpp#L902-L948, the basic things like variable names do not adhere to the LLVM Coding guidelines. 

I am not able to understand this piece of the code. It says `createAvgPoolValueCountIncludePadFalseCase` but returns even when that is the case, just because no padding is done, and then the next method is called `createAvgPoolValueCountIncludePadTrueCase` with a return, and then it's followed by another return, which can never be reached.  If that's the case, then it should have been described there; I have to spend quite some time to understand how this kind of control flow works.
https://github.com/llvm/torch-mlir/blob/4fe75dd0a94934fc3e05dfc4bd2b49046b2cff0e/lib/Conversion/TorchToLinalg/Pooling.cpp#L1043-L1054

Also, you said in the PR description:
```
I also did some refactoring to the original code to reduce the size of the functions and to avoid redundancy when possible.
```
But did it make the flow simpler? I don't think so. It's good that you did that, but the flow of the code should be such that any other contributor willing to contribute should be able to understand the code flow without putting a lot of effort, which is clearly not the case here because of the complexity being introduced in the code flow. Also, this might have enabled some new tests, but it should also error out clearly on the cases not supported and should not result in downstream failure.

In the end, I would like to say that I would not want to get into a debate on this and instead focus on what can be done ahead. I have to write the things above since you asked for them. I can find out more issues in the patch, but neither do I like doing that, nor do I want to. I believe in constructive feedback, and you should take it in that spirit only, and the focus should be on how we can address these issues.

ivangarcia44:
> > > Honestly, I think this PR should not have been merged. Apart from having some breaking changes, it has other issues, like incorrect code flow, and it does not adhere to standard LLVM contribution guidelines. For future reference, I would suggest that for such a PR making significant changes, we should wait for the original authors of the code to respond within a definite time frame.
> > 
> > 
> > For the next time I will. I apologize for the inconvenience. I waited for the review for almost a month, and I thought it was enough time to merge. I sent reminders various times during that month. Can you please point where is the incorrect code flow and LLVM guidelines violations? The fixes in the latest change cover various numerical correctness issues that were present prior any of my changes that were uncovered by a new set of E2E tests I added.
> 
> Even if I don't go in much depth then you can see that atleast in this part of the code
> 
> https://github.com/llvm/torch-mlir/blob/4fe75dd0a94934fc3e05dfc4bd2b49046b2cff0e/lib/Conversion/TorchToLinalg/Pooling.cpp#L902-L948
> 
> , the basic things like variable names do not adhere to the LLVM Coding guidelines.
> I am not able to understand this piece of the code. It says `createAvgPoolValueCountIncludePadFalseCase` but returns even when that is the case, just because no padding is done, and then the next method is called `createAvgPoolValueCountIncludePadTrueCase` with a return, and then it's followed by another return, which can never be reached. If that's the case, then it should have been described there; I have to spend quite some time to understand how this kind of control flow works.
> 
> https://github.com/llvm/torch-mlir/blob/4fe75dd0a94934fc3e05dfc4bd2b49046b2cff0e/lib/Conversion/TorchToLinalg/Pooling.cpp#L1043-L1054
> 
> Also, you said in the PR description:
> 
> ```
> I also did some refactoring to the original code to reduce the size of the functions and to avoid redundancy when possible.
> ```
> 
> But did it make the flow simpler? I don't think so. It's good that you did that, but the flow of the code should be such that any other contributor willing to contribute should be able to understand the code flow without putting a lot of effort, which is clearly not the case here because of the complexity being introduced in the code flow. Also, this might have enabled some new tests, but it should also error out clearly on the cases not supported and should not result in downstream failure.
> 
> In the end, I would like to say that I would not want to get into a debate on this and instead focus on what can be done ahead. I have to write the things above since you asked for them. I can find out more issues in the patch, but neither do I like doing that, nor do I want to. I believe in constructive feedback, and you should take it in that spirit only, and the focus should be on how we can address these issues.

The name of createAvgPoolValueCountIncludePadFalseCase  could be confusing because as you mentioned there are more conditions under which it can exit. Renaming this method could help as it has evolved over time while I discovered and fixed numerical correctness issues I found with the new tests. I will update this name and will outline the condition from this function to make the control flow easier to read. Although this could be perceived as a readability issue, it is not a control flow issue. I understand that a control flow issue is something that is numerically incorrect.

I still think the number of lines should be short. Various companies have this C++ coding standard, including Google (See "Write Short Functions" in this link: https://google.github.io/styleguide/cppguide.html#Function_Declarations_and_Definitions).

The original code (prior any of my changes) had way more than 100 lines of code, was hard to understand as you pointed above (in my old change set I just generalized the old algorithm, not modified it at its core), the variables did not conveyed any meaning and there was no comments explaining how the formula came about. 

For the new PR (https://github.com/llvm/torch-mlir/pull/4169) I had to rewrite it completely to fix an E2E numerical correctness bug I found as part of the new test suite. Now the code has more comments, the variables are better named, although the functions sizes still need more work but did not venture in further refactoring to avoid increasing the complexity of the change.

The reason why the IREE test failed with this change was because of a E2E test gap in the torch-mlir project. The IREE failed test is one of the tests I added in the new E2E test suite to avoid a situation like this in the future. The new E2E test suite reduces the testing gap of the operators, but does not eliminate it. I have found other bugs I fixed in the convolution operator due to lack of test coverage for them.

Please add add additional feedback in the PR below.

https://github.com/llvm/torch-mlir/pull/4169

Issue #4040, opened 2025-02-23
Python tests are failing in Windows build


sahas3:
There are two types of failures when running the python based tests on Windows:

1. End-to-end execution tests are [failing](https://github.com/sahas3/torch-mlir/actions/runs/13413554894/job/37468876174#step:10:75) due to the `signal.SIGALRM` (used for timeout handling in the following piece of testing code) is not available for Windows  https://github.com/llvm/torch-mlir/blob/a265d283357f572c5437f9217ea85fa9770d374c/projects/pt1/python/torch_mlir_e2e_test/framework.py#L310-L323
One possible solution is to use the https://pypi.org/project/pytest-timeout/ plugin with pytest which offers portability across different platforms.

2. Python API unit tests using LLVM's `lit` testing framework are [failing](https://github.com/sahas3/torch-mlir/actions/runs/13473693590/job/37650259773#step:8:6570) due to `pwd` module not being available in Windows.

Issue #3985, opened 2025-01-27
Add a build/test workflow for Windows


ScottTodd:
We've seen several downstream build breaks from torch-mlir due to missing test coverage. Having at least a nightly CI build using GitHub-hosted runners would provide earlier signal for build issues.

## Existing workflows

* https://github.com/llvm/torch-mlir/blob/main/.github/workflows/ci.yml only runs on Linux
* https://github.com/llvm/torch-mlir/blob/main/.github/workflows/buildRelease.yml runs on multiple platforms but was last triggered 1 year ago (https://github.com/llvm/torch-mlir/actions/workflows/buildRelease.yml)

## Expanding to new platforms

I see scripts used by ci.yml that could  be forked or generalized:

* https://github.com/llvm/torch-mlir/blob/main/build_tools/ci/build_posix.sh
* https://github.com/llvm/torch-mlir/blob/main/build_tools/ci/test_posix.sh

A workflow could also be added that runs commands directly. I've been moving https://github.com/iree-org/iree (a downstream user of torch-mlir) away from such scripts, instead opting to make the build system work better out of the box using default options, with things like ccache and the choice of compiler (e.g. gcc/clang) delegated to the user's choice or a github action that configures environment variables.

sahas3:
I gave this a shot by reusing  [https://github.com/llvm/torch-mlir/blob/main/build_tools/python_deploy/build_windows_ci.sh](https://github.com/llvm/torch-mlir/blob/main/build_tools/python_deploy/build_windows_ci.sh) which is used for building the windows wheels in [toch-mlir-release](https://github.com/llvm/torch-mlir-release) repo. The build worked fine but [testing fails](https://github.com/sahas3/torch-mlir/actions/runs/13413554894/job/37468876174#step:10:75) because `signal.SIGALRM` (used for timeout handling -- not much familiar with this piece of testing code) is not available for Windows. That's a different problem to solve to enable testing on Windows even outside of the CI workflow, so will create a separate issue for that. However, is there a benefit to still add a build only CI for now?

PR #4041, opened 2025-02-23
Add Windows build workflow in the CI.

sahas3:
1. Adds a workflow to trigger windows build in the CI similar to the existing one for Linux to address the build CI request captured in https://github.com/llvm/torch-mlir/issues/3985
2. Only non-python based LIT unit tests are enabled in the workflow due to failures reported in https://github.com/llvm/torch-mlir/issues/4040
3. The CI is only run against `torch-nightly` since python based tests are disabled. My understanding is that without the python based tests there's no need to run against the `stable` version too. If that's not the case, will enable `stable` as well. Tested that both `nightly` and `stable` versions work fine: https://github.com/sahas3/torch-mlir/actions/runs/13477124004


.github/workflows/ci.yml
ScottTodd, line 1: Want to tag https://github.com/llvm/torch-mlir/issues/3985 in the PR description?

.github/workflows/ci.yml
```diff
@@ -79,3 +79,62 @@ jobs:
...
         run: |
           bash build_tools/ci/check_generated_sources.sh
+
+  build-test-windows:
+    strategy:
+      fail-fast: true
+      matrix:
+        torch-version: [nightly]
+    name: Build (Windows, torch-${{ matrix.torch-version }})
+    runs-on: windows-latest
```
ScottTodd, line 89: The logs at https://github.com/llvm/torch-mlir/actions/runs/13487571031/job/37680706814?pr=4041 show this job taking 2 hours with an empty cache and the standard GitHub-hosted runners. I'd personally limit jobs running on `pull_request` triggers to 40-60 minutes and move longer running jobs to a lower frequency (`push` or `schedule` triggers). That might mean creating a new `ci_nightly.yml` workflow with different triggers.

Switching to larger self-hosted runners may be an option, but I'm not familiar with what runners are available in the llvm github organization.

Do any other torch-mlir contributors/maintainers have opinions? I don't personally work enough in this repository to know what others would be comfortable with.

build_tools/python_deploy/build_windows_ci.sh
```diff
@@ -14,9 +36,23 @@ cmake -GNinja -Bbuild \
   -DLLVM_EXTERNAL_PROJECTS="torch-mlir" \
   -DLLVM_EXTERNAL_TORCH_MLIR_SOURCE_DIR="$PWD" \
   -DPython3_EXECUTABLE="$(which python)" \
+  -DPython_EXECUTABLE="$(which python)" \
   -DTORCH_MLIR_ENABLE_STABLEHLO=OFF \
-  $GITHUB_WORKSPACE/externals/llvm-project/llvm
+  -DTORCH_MLIR_ENABLE_PYTORCH_EXTENSIONS=ON \
+  -DTORCH_MLIR_ENABLE_LTC=OFF \
+  -DTORCH_MLIR_ENABLE_JIT_IR_IMPORTER=ON \
+  $repo_root/externals/llvm-project/llvm
```
ScottTodd, line 44: nit: I would match the local style in this file and wrap variable expansions with `{}`, so `${repo_root}/externals/...` and similar in other locations

* https://google.github.io/styleguide/shellguide.html#variable-expansion

Without the braces, expressions like `$repo_root/build` could be seen as ambiguous, as the reader needs to know that `/` marks the end of the variable name.

sahas3: Done.

.github/workflows/ci.yml
sahas3: Good point, it makes sense to run this at a reduced frequency if there's no larger runner to run it on. I'll wait for a couple of days to make this change in case others have any opinion.

sahas3:
Hi @sjarus and @justin-ngo-arm, I had to modify the TOSA filecheck tests in this PR as looks like the emission order of back to back `tosa.const_shape` op is not deterministic across platforms. I am not sure if that's expected but since the current plan is to not run the Windows CI in all PRs (due to build taking too long, please see the unresolved discussion) this is something that might slip through.

sjarus:
> Hi @sjarus and @justin-ngo-arm, I had to modify the TOSA filecheck tests in this PR as looks like the emission order of back to back `tosa.const_shape` op is not deterministic across platforms. I am not sure if that's expected but since the current plan is to not run the Windows CI in all PRs (due to build taking too long, please see the unresolved discussion) this is something that might slip through.

Thanks! @justin-ngo-arm maybe we want to relax the checks to support such non determinism here ? Since we already have more robust functional testing , ordering restrictions in LIT tests are probably unnecessary.

justin-ngo-arm:
> > Hi @sjarus and @justin-ngo-arm, I had to modify the TOSA filecheck tests in this PR as looks like the emission order of back to back `tosa.const_shape` op is not deterministic across platforms. I am not sure if that's expected but since the current plan is to not run the Windows CI in all PRs (due to build taking too long, please see the unresolved discussion) this is something that might slip through.
> 
> Thanks! @justin-ngo-arm maybe we want to relax the checks to support such non determinism here ? Since we already have more robust functional testing , ordering restrictions in LIT tests are probably unnecessary.

Sounds good to me. I can use `CHECK-DAG` for `tosa.const_shape` in the future.

aartbik:
LG on sparse

sjarus (approved):
Oking the TOSA LIT changes.

.github/workflows/ci.yml
sahas3: Hi @stellaraccident, finally got around to adding the CI for Windows build as you had suggested in https://github.com/llvm/torch-mlir-release/pull/2#issuecomment-2431075695. So wanted to check if you've any thoughts on the frequency of running this CI as it takes a while to run and if you know whether there's a large windows runner available to schedule this new CI. Thanks!

ScottTodd: Here's an idea:

1. Let the tests run at the default frequency (`push` and `pull_request` triggers) and take the full ~2 hours
2. Wait to see if anyone complains :D

Commit volume may be low enough in this repo that it isn't yet worth doing more.

If it were me though, I'd probably stand up a `ci_windows.yml` that forks https://github.com/llvm/torch-mlir/blob/main/.github/workflows/ci.yml and has these triggers to start:
```yml
on:
  workflow_dispatch:
  workflow_call:
  schedule:
    - cron: '0 11 * * *'
```

Or `push` trigger instead of `schedule` given that there are only a few commits per day.

sahas3: Thanks @ScottTodd. I created a new `ci_windows.yml` and went with the `push` trigger. Looking at the history we don't have that many commits to the `main` branch daily so I agree that running the windows ci on push to `main` should be okay. We can always switch to running on a schedule if needed.

build_tools/python_deploy/build_windows_ci.sh
```diff
@@ -1,8 +1,30 @@
...
+  mkdir -p "${cache_dir}"
+  cache_dir="$(cd ${cache_dir} && pwd)"
+fi
+
+echo "Caching to ${cache_dir}"
+mkdir -p "${cache_dir}/ccache"
+mkdir -p "${cache_dir}/pip"
+
+export CCACHE_DIR="${cache_dir}/ccache"
+export CCACHE_MAXSIZE="350M"
```
ScottTodd, line 22: IDK if this will be enough cache space for the build... I guess we'll see after a few runs. Pretty low impact if there are cache misses given the run frequency we're starting that.

ScottTodd (approved):

ScottTodd:
Build times so far are _really_ good. Consistently getting high cache hit rates.

https://github.com/llvm/torch-mlir/actions/workflows/ci_windows.yml

Cold cache: 2 hours
Hot cache: 10 minutes!

This new workflow did miss a downstream Windows issue though: https://github.com/llvm/torch-mlir/issues/4085. I'll see if I can find what is different about the build (e.g. compiler flags tested, compiler version in the workflow, differences between upstream and downstream build environment).

PR #4042, opened 2025-02-24
[tosa] : Handle CHW input for avgpool2d.

sahas3:
This PR fixes two issues:
1. Avg(Max)Pool2d allows (C, H, W) as input which causes a failure when creating `tosa.avg_pool2d` or `tosa.max_pool2d` as those ops expects 4D (N, H, W, C) tensor. Fix is to add a N=1 dimension before creating the tosa ops.
2. Avg(Max)Pool2d also allows kernel/stride to be specified as a tuple of single int, in which case the value is repeated for both H and W dims. This is currently not handled as well causing a segv when trying to access `kernel[1]/stride[1]`. Fix is to expand `kernel/stride` to be size 2 by repeating the first element.

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -5764,6 +5764,35 @@ public:
                                           : nhwcToNchw4DTransposeDims);
   }
 
+  void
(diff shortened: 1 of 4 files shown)


sjarus (approved):
LGTM, thanks for this!

sjarus:
Could you please change the subject to [tosa] and use that as the prefix instead of '[Task]' @sahas3  ?

PR #4044, opened 2025-02-24
[TorchOnnxToTorch] Handle default value for Resize op attribute

praveen-g-ctt:
Add check for default value of the resize op attribute keep_aspect_ratio_policy = "stretch"


zjgarvey (approved):
Nice, LGTM

PR #4049, opened 2025-02-25
Integrate LLVM at b0b70577610ec65606b27ec8992ef0350e78a6a8

vivekkhandelwal1:
Update LLVM to https://github.com/llvm/torch-mlir/commit/b0b70577610ec65606b27ec8992ef0350e78a6a8

This commit replaces the use of `buffer-deallocation` pass with the `buffer-deallocation-pipeline` as specified here:  https://discourse.llvm.org/t/psa-bufferization-new-buffer-deallocation-pipeline/73375#migration-guide-1.

Also, this commit disables the failing sparsity tests.

TOSA Updates Summary:

  1. [TOSA] Update tosa.clamp's min/max pair
  
  * Update tosa.clamp's `min` and `max` according to TOSA 1.0 spec:
    https://www.mlplatform.org/tosa/tosa_spec.html#_clamp
  * Update LIT tests
  
  Signed-off-by: Justin Ngo <justin.ngo@arm.com>
  Change-Id: I6013c440c08f7c37f11ef4b5f46cbce1a3fae54f
  
  2. [TOSA] Update tosa.resize's scale, offset, border as inputs
  
  * In TOSA 1.0, tosa.resize's scale, offset, and border are inputs with
    !tosa.shape type
  * Update LIT tests

---------

Co-authored-by: Justin Ngo <justin.ngo@arm.com>

### DIFF
--- a/lib/Conversion/TorchToTosa/TorchToTosa.cpp
+++ b/lib/Conversion/TorchToTosa/TorchToTosa.cpp
@@ -870,25 +870,21 @@ LogicalResult ConvertAtenOp<AtenReluOp>::matchAndRewrite(
   Value self = adaptor.getSelf();
   auto selfTy = cast<TensorType>(self.getType());
 
-  // Maps to tosa.clamp which has both int and fp limits.
-  int64_t clampMin = 0;
-  Value clampIn = self;
   if (!selfTy) {
     return rewriter.notifyMatchFailure(op,
                                        "Only Tensor types supported in TOSA");
   }
 
-  // Rescale the clampIn for quantized types. TBD
+  // Rescale self for quantized types. TBD
   if (!isa<mlir::FloatType>(selfTy.getElementType())) {
     return rewriter.notifyMatchFailure(
         op, "Only floating-point datatype legalization currently supported");
   }
 
+  // Maps to tosa.clamp
   // Use default NaN Propagation mode "PROPAGATE" for tosa.clamp
   rewriter.replaceOpWithNewOp<tosa::ClampOp>(
-      op, getTypeConverter()->convertType(op.getType()), clampIn,
-      rewriter.getI64IntegerAttr(clampMin),
-      rewriter.getI64IntegerAttr(std::numeric_limits<int32_t>::max()),
+      op, getTypeConverter()->convertType(op.getType()), self,
       rewriter.getF32FloatAttr(0.0f),
       rewriter.getF32FloatAttr(std::numeric_limits<float>::max()),
       /*nan_mode=*/rewriter.getStringAttr("PROPAGATE"));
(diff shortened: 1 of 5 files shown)


sjarus (approved):

sjarus (approved):

aartbik:
does it make sense to simply remove the sparsity test completely? many tests are already disabled

(I offered to maintain the sparsity functionality in the past, but the feature was not considered high priority; I am really fine either way, but just commenting out tests does not seem the way forward).

vivekkhandelwal1:
> does it make sense to simply remove the sparsity test completely? many tests are already disabled
> 
> (I offered to maintain the sparsity functionality in the past, but the feature was not considered high priority; I am really fine either way, but just commenting out tests does not seem the way forward).

Hi @aartbik, I was also thinking the same that a majority of sparsity tests are now disabled. But I didn't remove them since I thought they might get enabled again in the future. Since, you're the only person AFAIK who maintains this so it's your call whether to keep them or remove them completely.

aartbik:
> Hi @aartbik, I was also thinking the same that a majority of sparsity tests are now disabled. But I didn't remove them since I thought they might get enabled again in the future. Since, you're the only person AFAIK who maintains this so it's your call whether to keep them or remove them completely.

I am perfectly okay either way (that is, either removing the tests completely or enabling them and maintaining them [which would be mostly on me]; but it sound we are both not okay with just commenting out tests).

It is really up to @stellaraccident . And no hard feelings whatever the outcome. The torch-mlir community has been extremely kind to me allowing the MLIR sparsifier team to add the torch-sparse to mlir bridge inside torch-mlir, and building the first end-to-end sparse compiler for PyTorch using MLIR was a privilege. But priorities change, and if keeping this slightly academic and experimental feature around takes too much engineering effort, I understand. I can keep this working in my private fork until such a time this makes sense to integrate again.

stellaraccident:
> > Hi @aartbik, I was also thinking the same that a majority of sparsity tests are now disabled. But I didn't remove them since I thought they might get enabled again in the future. Since, you're the only person AFAIK who maintains this so it's your call whether to keep them or remove them completely.
> 
> I am perfectly okay either way (that is, either removing the tests completely or enabling them and maintaining them [which would be mostly on me]; but it sound we are both not okay with just commenting out tests).
> 
> It is really up to @stellaraccident . And no hard feelings whatever the outcome. The torch-mlir community has been extremely kind to me allowing the MLIR sparsifier team to add the torch-sparse to mlir bridge inside torch-mlir, and building the first end-to-end sparse compiler for PyTorch using MLIR was a privilege. But priorities change, and if keeping this slightly academic and experimental feature around takes too much engineering effort, I understand. I can keep this working in my private fork until such a time this makes sense to integrate again.

If you want to keep working on it, I'm fine either with it existing in the repo or a fork. If in the repo, I'd like at least tests to be in their own location and flag guarded so it doesn't trip up folks on normal maintenance activities, and then it can have its own lifecycle to be ready for broader use when its ready. I think it's an interesting feature but it also requires a fair bit of understanding to operate and keep upgraded, and I'd rather have things organized so that you've got the time to take it where you think it can go without us being in a tug of ware on test commenting in/out and such. Main idea is that directories and flags for this level of thing are easy to co-exist with, and it is better if things that aren't quite ready yet have the space to get there without causing overhead to the folks keeping the project running who don't know much about it.

aartbik:
Sounds good. Vivek, you can of course submit this so the integration is not blocked, and then I will have a look whether I put it someplace else (and fix), or just remove altogether for the time being. Thanks both.

PR #4046, opened 2025-02-24
asan: Preload asan runtime into python interpreter

mgehre-amd:
Same as https://github.com/llvm/llvm-project/pull/123303 upstream

PR #4050, opened 2025-02-25
Fix sanitizers issues

mgehre-amd:
ubsan detected 'lib/Dialect/TorchConversion/Transforms/UnpackQuantTensor.cpp:109:21: runtime error: left shift of negative value -16' and asan detected `memcpy` with nullptr when we passed and empty vector with `.data() == nullptr` to mlirOperationStateAddResults.

Detected using https://github.com/llvm/torch-mlir/pull/4046


jinchen62 (approved):
LGTM, thanks

PR #4055, opened 2025-02-26
Add support for transposed grouped convolution in torch to linalg lowering

ivangarcia44:
The conversion of the convolutiong torch operation to linalg currently works for grouped convolution (number of groups > 1) and transposed convolution, but the conversion failed when both are used at the same time. This change set correct this.

The core of the changes are in the Linear.cpp. In transposed grouped convolution, the output filters is the one divided by the groups in the weights, not the input channel (see the "Variables" section in both links below for details). This was one of the fixes.

* https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html
* https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html

The other issue was that the weights expansion had to happen before the Channel/Filter dimension permutation/flip. This is because the expansion deals with adjacent dimensions, but in the final weights tensor the group and the input channel are not going to be adjacent. Once the dimensions are flipped, the expansion operation can't generate the expected dimension format. See the comment in the code for details.

@rsuderman 
@vivekkhandelwal1 
@zjgarvey
@penguin-wwy 
@ubfx 
@sahas3 
@dixinzhou 
@rafaelubalmw


ivangarcia44:
This PR got mixed with another one. Creating a new one.

ivangarcia44:
New PR: https://github.com/llvm/torch-mlir/pull/4056

PR #4056, opened 2025-02-26
Add support for transposed grouped convolution in torch to linalg lowering

ivangarcia44:
The conversion of the convolutiong torch operation to linalg currently works for grouped convolution (number of groups > 1) and transposed convolution, but the conversion failed when both are used at the same time. This change set correct this.

The core of the changes are in the Linear.cpp. In transposed grouped convolution, the output filters is the one divided by the groups in the weights, not the input channel (see the "Variables" section in both links below for details). This was one of the fixes.

https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html
https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html
The other issue was that the weights expansion had to happen before the Channel/Filter dimension permutation/flip. This is because the expansion deals with adjacent dimensions, but in the final weights tensor the group and the input channel are not going to be adjacent. Once the dimensions are flipped, the expansion operation can't generate the expected dimension format. See the comment in the code for details.

@rsuderman
@vivekkhandelwal1
@zjgarvey
@penguin-wwy
@ubfx
@sahas3
@dixinzhou
@rafaelubalmw


test/Conversion/TorchToLinalg/convolution.mlir
```diff
@@ -76,3 +76,74 @@ func.func @conv_broadcast(%arg0: !torch.vtensor<[1,80,3000],f32>, %arg1: !torch.
   %2 = torch.aten.convolution %arg0, %arg1, %arg2, %0, %0, %0, %false, %1, %int1 : !torch.vtensor<[1,80,3000],f32>, !torch.vtensor<[1024,80,3],f32>, !torch.vtensor<[1024],f32>, !torch.list<int>, !torch.list<int>, !torch.list<int>, !torch.bool, !torch.list<int>, !torch.int -> !torch.vtensor<[1,1024,3000],f32>
   return %2 : !torch.vtensor<[1,1024,3000],f32>
 }
+
+// CHECK-LABEL:   func.func @transposedConv2D(
+// CHECK-SAME:       %[[arg0:.*]]: !torch.vtensor<[1,2,5,7],f32>) -> !torch.vtensor<[1,4,10,14],f32>
+// CHECK:         = linalg.generic
+// CHECK-SAME:       outs(%[[VAR1:.*]] : tensor<4x2x3x3xf32>) {
+// CHECK:         %[[VAR2:.*]] = tensor.extract
```
sahas3, line 84: can you use meaningful variable names? See lines 61 - 69 in this file for example

test/Conversion/TorchToLinalg/convolution.mlir
```diff
@@ -76,3 +76,74 @@ func.func @conv_broadcast(%arg0: !torch.vtensor<[1,80,3000],f32>, %arg1: !torch.
...
+// CHECK-SAME:       : tensor<2x4x3x3xf32>
+// CHECK-NEXT:    linalg.yield %[[VAR3:.*]] : f32
+// CHECK-NEXT:    } -> tensor<4x2x3x3xf32>
+// CHECK:         %[[VAR4:.*]] = linalg.broadcast ins(%[[VAR5:.*]] : tensor<4xf32>) outs(%[[VAR6:.*]] : tensor<1x4x11x15xf32>) dimensions = [0, 2, 3]
+// CHECK:         %[[VAR7:.*]] = linalg.conv_2d_nchw_fchw {dilations = dense<1> : vector<2xi64>, strides = dense<1> : vector<2xi64>}
+// CHECK-SAME:        ins(%[[VAR8:.*]], %[[VAR9:.*]] : tensor<1x2x13x17xf32>, tensor<4x2x3x3xf32>) outs(%[[VAR10:.*]] : tensor<1x4x11x15xf32>) -> tensor<1x4x11x15xf32>
+// CHECK-NEXT:    %[[VAR11:.*]] = tensor.cast %[[VAR12:.*]] : tensor<1x4x11x15xf32> to tensor<1x4x?x?xf32>
```
sahas3, line 91: what does the rest of the IR look like? I am interested in seeing how the output tensor<1x4x11x15xf32> is reconciled with the original output torch.vtensor<[1,4,10,14],f32>

Is the e2e test capturing this config of the conv layer?

sahas3 (changes requested):

zjgarvey:
Cool! I'll take a look at this tomorrow. This is a tricky case to handle properly, so I'd like to go through the implementation carefully.

ivangarcia44:
Adding @Hanumanth04 to the review.

test/Conversion/TorchToLinalg/convolution.mlir
ivangarcia44: The rest of the IR looks like this:

    %cast_0 = tensor.cast %cast : tensor<1x4x?x?xf32> to tensor<1x4x10x14xf32>
    %12 = torch_c.from_builtin_tensor %cast_0 : tensor<1x4x10x14xf32> -> !torch.vtensor<[1,4,10,14],f32>
    return %12 : !torch.vtensor<[1,4,10,14],f32>

I added a check for the first of the 3 lines above, because after that it is trivial and the type reached to its final shape.

I did a similar update for the 3rd unit test (grouped transposed convolution, the case of the bug I am fixing). The second unit test (grouped convolution) did not needed this because its final shape is determined already.

I have an e2e test for this and it does numerical equivalence check. I had to fix the algorithm slightly to get numerical equivalence.

zjgarvey (approved):
Thanks, this looks good to me.

lib/Conversion/TorchToLinalg/Linear.cpp
```diff
@@ -955,7 +955,50 @@ class ConvertAtenConvolutionOp : public OpConversionPattern<AtenConvolutionOp> {
       if (isa<mlir::IntegerType>(inputDTy))
         pad = rewriter.create<arith::TruncIOp>(op.getLoc(), inputDTy, pad);
     }
+
+    // This code was moved earlier because in the grouped transposed convolution
+    // case we need to expand before doing the dimension permutation. For the
+    // grouped non-transposed convolution, we don't need to do filter/channel
+    // dimension flipping, we can just expand the group from the filter in place
+    // to have the group dimension in front:
```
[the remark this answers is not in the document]
ivangarcia44: I agree they don't need to know the previous state of the code. I removed the comment about moving the code. But still rewrote the paragraph to explain why the expandWeight needs to be executed earlier in the transpose case. Please let me know if I should make further changes for this. Thanks.

    // The expandWeight lambda function below is used to expand the group
    // dimension. For the normal case the group dimension is expanded out
    // of the output filter dimension:
    // expand F,C,H,W -> G,F/G,C,H,W
    //
    // Note that the group dimension has to be the first dimension. For the 
    // transposed convolution case, the group convolution is extracted out
    // of the input channel dimension. But note that the input channel 
    // dimension is interchanged with the output filter dimension (due to
    // the transposed operation). Because of this the group and input 
    // channel dimensions will not be adjacent and the expand_shape 
    // operation will not work.
    //
    // For this reason, in the transposed convolution case the expandWeight
    // lambda needs to be executed before this dimension flipping by doing
    // these two steps:
    //
    // Expansion:    C,F,H,W -> G,C/G,F,H,W
    //
    // Dimension interchange: G,C/G,F,H,W -> G,F,C/G,H,W
    //

sahas3 (approved):
LGTM.

PR #4057, opened 2025-02-27
Add runtime asserts for aten.reflection.pad_2d

praveen-g-ctt:
Add runtime asserts to check padding constraints of aten.reflection.pad_2d for dynamic dims

### DIFF
--- a/lib/Conversion/TorchToLinalg/DataMovement.cpp
+++ b/lib/Conversion/TorchToLinalg/DataMovement.cpp
@@ -406,8 +406,7 @@ public:
 
     auto verifyPadding = [&](int64_t padArgument, int64_t dim,
                              StringRef errorMessage) {
-      auto padValue =
-          rewriter.create<arith::ConstantIndexOp>(loc, padArgument);
+      auto padValue = rewriter.create<arith::ConstantIndexOp>(loc, padArgument);
       Value index = rewriter.create<arith::ConstantIndexOp>(loc, dim);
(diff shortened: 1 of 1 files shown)


lib/Conversion/TorchToLinalg/DataMovement.cpp
```diff
@@ -404,14 +404,24 @@ class ConvertAtenReflectionPad2dOp
...
-    assert(getVPadArgument(BOTTOM) < inputType.getShape()[vDim] &&
-           "Bottom padding too large");
+    auto verifyPadding = [&](int64_t padArgument, int64_t dim,
+                             StringRef errorMessage) {
+      auto padValue =
+          getConstant(rewriter, loc, padArgument, rewriter.getI64Type());
+      Value index = rewriter.create<arith::ConstantIndexOp>(loc, dim);
+      Value shapeDim = rewriter.create<tensor::DimOp>(loc, input, index);
+      shapeDim = rewriter.create<arith::IndexCastOp>(loc, rewriter.getI64Type(),
+                                                     shapeDim);
```
vivekkhandelwal1, line 414: You don't need to do this. Just change the type of `padValue` to be index. Then, you can compare both the values of index type.

vivekkhandelwal1 (approved):
Rest all, LGTM.

lib/Conversion/TorchToLinalg/DataMovement.cpp
praveen-g-ctt: Thanks Vivek. Updated the code.

PR #4060, opened 2025-02-27
[linalg] Use query shapes for attention broadcast

rsuderman:
When broadcasting the mask we need to use the query shapes and not the keys. This is due to the key for GQA having different batch dimensions than the expanded output.


zjgarvey:
The test change might not be correct. Does this require a torch version bump or something?

rsuderman:
> The test change might not be correct. Does this require a torch version bump or something?

Nope, I just messed up the causal arg.

PR #4062, opened 2025-02-28
Update shape_inference.cpp for randn.generator

keshavj-cerebras:
PR adds support for `aten.randn.generator`, the generator object (if passed) will not affect shape inference.


prateekgu-cerebras (approved):
LGTM!

PR #4063, opened 2025-02-28
[MLIR][TORCH] Add support for none result for convolution_backward op

vivekkhandelwal1:
This commit adds the support for handling the cases when any result of the convolution backward op is none.

This commit also adds the support for handling none results in SimplifyShapeCalculations pass.

### DIFF
--- a/lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp
+++ b/lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp
@@ -152,12 +152,16 @@ static LogicalResult refineShapeCalculateResult(ShapeCalculateOp op,
       sizes.push_back(kUnknownSize);
   }
 
-  auto originalResultType = cast<BaseTensorType>(result.getType());
-  auto impliedTypesFromShape = cast<BaseTensorType>(
-      cast<BaseTensorType>(originalResultType)
-          .getWithSizesAndDtype(ArrayRef(sizes),
-                                originalResultType.getOptionalDtype()));
-
+  // For the cases when the result is none, the implied type will remain the
+  // same as original result type.
+  Type impliedTypesFromShape = result.getType();
+  if (!isa<Torch::NoneType>(impliedTypesFromShape)) {
+    BaseTensorType originalResultType = cast<BaseTensorType>(result.getType());
+    impliedTypesFromShape =
+        cast<BaseTensorType>(originalResultType)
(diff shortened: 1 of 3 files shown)


lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp
```diff
@@ -152,12 +152,16 @@ static LogicalResult refineShapeCalculateResult(ShapeCalculateOp op,
...
 
-  auto originalResultType = cast<BaseTensorType>(result.getType());
-  auto impliedTypesFromShape = cast<BaseTensorType>(
-      cast<BaseTensorType>(originalResultType)
-          .getWithSizesAndDtype(ArrayRef(sizes),
-                                originalResultType.getOptionalDtype()));
-
+  // For the cases when the result is none, the implied type will remain the
+  // same as original result type.
+  Type impliedTypesFromShape = result.getType();
```
zjgarvey, line 157: I wonder if it is possible to resolve this by instead modifying the logic in the shape inference implementation for this op? It seems to be incorrect in not using the mask to determine output shapes.

zjgarvey (approved):
Looks good to me, but consider checking out if it would be possible to modify the shape inference implementation instead.

AmosLewis (approved):

lib/Dialect/Torch/Transforms/SimplifyShapeCalculations.cpp
vivekkhandelwal1: Actually, the return value of the shape function for this op is of type: `Tuple[List[int], List[int], List[int]]`, so I don't think we can do anything here.

vivekkhandelwal1: Ref: https://github.com/llvm/torch-mlir/blob/c7f8ac0b968b9f2e029b13db6fe77f3d75bc86c4/projects/pt1/python/torch_mlir/jit_ir_importer/build_tools/abstract_interp_lib_gen.py#L1941

vivekkhandelwal1: Also, isn't it better that a safety check is added to the `ShapeCalculateOp`?

zjgarvey: Yeah, that sounds good. Thanks for looking into it.

PR #4064, opened 2025-02-28
[ONNX] Fix import of boolean tensor constants

rkayaith:
Currently the shape and element type of boolean tensor attributes are incorrect, since they use the shape/dtype of the packed numpy array rather than of the original tensor value. Specifying the shape and element type explicitly resolves this.


zjgarvey (approved):
Thanks!

PR #4070, opened 2025-03-03
APInt fix

AaronStGeorge:
`ConvertAtenOp<ValueTensorLiteralOp>` triggers newly added assertions around `APInt` construction (see [llvm/llvm-project#106524](https://github.com/llvm/llvm-project/pull/106524)) when `isSigned == false` and we pass the result of `v.getSExtValue()` that doesn't fit within the specified bit-width. For example, if we build a 32-bit `APInt` using a value of `-1` (returned by `v.getSExtValue()`), that negative number is stored in a `uint64-t` representation as `0xFF FF FF FF FF FF FF FF`. Since we are treating it as unsigned, it obviously doesn't fit into 32-bits, causing the new assertion to fail.


sjarus (approved):
Looks ok. Adding @justin-ngo-arm to review against ongoing TorchToTosa work.

justin-ngo-arm:
I've tested this change against ongoing TOSA 1.0 update work and everything looks fine.

